PR #19, opened 2022-04-15
Predator Big SD Laugh (MOVIELIKE AF) + big SD fix

johnravioliUSA:
## About The Pull Request

Predators on the BIG SD setting will now laugh when they activate it, it is awesome.
A little bit of shitcode is changed to just add it in

[https://streamable.com/wygkja](url)
(p.s. You cannot big SD on the almayer, but you can still laugh maniacally as you blow up)

https://streamable.com/zvboiz

updated sound

## Why It's Good For The Game

Predators can do the movielike thing and automatically have a different laugh when self destructing on the big setting. It is good for the game since it's movielike AF, hence the name. Unironically, actually good for the game because it will be awesome

## Changelog

:cl: johnravioli
add: Big SD now has a laugh when you activate it
soundadd: New maniacal laugh to blow up to when fighting a Predator
fix: Sponsored by Betapower, Predator big SDs now work again.
/:cl:


code/modules/cm_preds/yaut_bracers.dm
```diff
@@ -751,6 +751,9 @@
 		var/turf/T = get_turf(M)
 		message_staff(FONT_SIZE_HUGE("ALERT: [usr] ([usr.key]) triggered their predator self-destruct sequence [A ? "in [A.name]":""] (<A HREF='?_src_=admin_holder;adminplayerobservecoodjump=1;X=[T.x];Y=[T.y];Z=[T.z]'>JMP</a>)"))
 		log_attack("[key_name(usr)] triggered their predator self-destruct sequence in [A ? "in [A.name]":""]")
+		var/obj/item/clothing/gloves/yautja/hunter/bracer = M.gloves
```
stanalbatross, line 754: why do you need this? it's not referenced anywhere.

code/modules/cm_preds/yaut_bracers.dm
```diff
@@ -751,6 +751,9 @@
...
+		if(explosion_type == 0 && is_ground_level(z))
```
stanalbatross, line 755: according to shaddeh this causes issues when trying to big SD

stanalbatross (changes requested):

stanalbatross (changes requested):
triio says give the SD sound a fadeout at the end because it sounds bad without one.

johnravioliUSA:
> triio says give the SD sound a fadeout at the end because it sounds bad without one.

its fixed

stanalbatross (approved):

PR #22, opened 2022-04-15
HPR suppressing fire

stanalbatross:
## About The Pull Request

MR by ghostsheet, I will make any changes needed and update + rebase it as needed.

This is the support version of the HPR, it's role will be to suppress the enemy with negative status effects rather than dealing sizable damage.

Suppressing Fire!

Suppressing stack by applying this amount of stack per bullet:

4 stacks to a human

2 stacks to a xeno

1 stack to a large xeno

Every 2 stack a small screen shake will apply, power scales on stacks

At 10 stack, daze will apply

At 20 stack, slow will apply

At 30 stack, super slow will apply

After 2 seconds stacks will start to dissipate completely over a few seconds

https://streamable.com/9arjdz

This is two of two proposal (!2131MR by ghostsheet, I will make any changes needed and update + rebase it as needed.
This is the support version of the HPR, it's role will be to suppress the enemy with negative status effects rather than dealing sizable damage.
Suppressing Fire!
Suppressing stack by applying this amount of stack per bullet:
4 stacks to a human
2 stacks to a xeno
1 stack to a large xeno
Every 2 stack a small screen shake will apply, power scales on stacks
At 10 stack, daze will apply
At 20 stack, slow will apply
At 30 stack, super slow will apply
After 2 seconds stacks will start to dissipate completely over a few seconds
https://streamable.com/9arjdz
This is two of two proposal ([!2131 (](https://github.com/cmdevs/colonial-warfare/-/merge_requests/2131)

## Why It's Good For The Game

HPR IFF is SHIT and this is pretty clearly evident. This moves the HPR more towards its intended role as an infantry support LMG, not a shitty smartgun that is less mobile.

Might need a design doc.

## Changelog

:cl: Ghostsheet, Stan_Albatross
add: HPR bullets will now give xenos suppressing fire stacks, which will daze, slow, and then super slow xenos after 15 or 30 shots, depending on their size.

balance: The HPR no longer has IFF when bipodded.
/:cl:

### DIFF
--- a/ColonialMarinesALPHA.dme
+++ b/ColonialMarinesALPHA.dme
@@ -291,6 +291,7 @@
 #include "code\datums\components\bonus_damage_stack.dm"
 #include "code\datums\components\cluster_stack.dm"
 #include "code\datums\components\footstep.dm"
+#include "code\datums\components\suppressing_stack.dm"
 #include "code\datums\components\toxin_buildup.dm"
 #include "code\datums\components\weed_damage_reduction.dm"
 #include "code\datums\components\xeno\shield_slash.dm"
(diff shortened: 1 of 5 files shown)


ghostsheet:
Remove holotargeting don't remove IFF

PR #26, opened 2022-04-15
Marker Spitter Strain/Metatarget Spitter

insertnamehere123:
## About The Pull Request

A spitter strain that debuffs targets by having them take 10% extra damage, which will begin ticking down after 5 seconds if they remain unmarked.

All xenos within 5 tiles of the hit target will get a shield to help them stay in the fight longer or escape if needed.

Has the viewrange of a boiler and can spit at distant targets, but the spit moves a tier slower than a base spitter.

Loses it's Spitter Frenzy and the acid spray, spit itself does less damage and loses it's DOT effect.

Adds a new xeno ammo with the above effects and IFF.

Github version of this: https://gitlab.com/cmdevs/colonial-warfare/-/merge_requests/2469

## Why It's Good For The Game

Gives the spitter a long ranged support oriented strain

## Changelog

:cl: Insertusername
rscadd: Marking Spit Ammo, Marker Spitter Strain.
tweak: New spitter strain.
/:cl:


stanalbatross:
DNM added until tested and design specifics decided upon

code/modules/mob/living/carbon/xenomorph/mutators/strains/spitter/marker.dm
```diff
@@ -0,0 +1,40 @@
+/datum/xeno_mutator/marker
+	name = "STRAIN: Spitter - Marker"
+	description = "You weaken your corrosive acid for pheremones to support your sisters and acid to weaken your foes"
```
stanalbatross, line 3:
```suggestion
	description = "You weaken your corrosive acid for pheromones to support your sisters and acid to weaken your foes"
```

code/modules/projectiles/ammo_datums.dm
```diff
@@ -2521,6 +2521,34 @@
...
+		var/shield_duration = 30 SECONDS
+		var/shield_decay = 10
+
+		if (!istype(X) || !X.check_state())
+			return
+
+		if(X.stat == DEAD)
+			continue
+
+		to_chat(X, SPAN_XENOBOLDNOTICE("THE MARKED ONE AND IT'S COHORTS MUST DIE!"))
```
stanalbatross, line 2548:
```suggestion
		to_chat(X, SPAN_XENOBOLDNOTICE("The marked host and its cohorts must die!"))
```
too over-the-top

stanalbatross (changes requested):
image
runtime during testing, occurred when moving around while zoomed out.

The zoomout also needs to be lower and there needs to be a better indication of what the spit actually does to the player.

PR #36, opened 2022-04-15
Fixes the light performance bug

### LINKED ISSUE #2: SGs don't have proper nightvision, they can't see xenos past 6 tiles.
See title.

Their glasses have darkness_view = 5 which is an issue, it should be darkness_view = 12

Watermelon914:
## About The Pull Request

Lummox said in a post that byond's built-in darkness makes view() calls O(n^3) instead of O(n^2), which has shown to be significant on CM. Profiler has shown that procs using view() take a lot more cpu time than they normally would, and players have described the game feeling a lot more sluggish earlier on ever since the lighting changes. As a result, I am reverting the changes to luminosity that were made 

Closes #2

## Why It's Good For The Game

Performance improvements


stanalbatross (approved):
zased

PR #38, opened 2022-04-15
SciAnnex minor fixes

Chainsawicus:
## About The Pull Request

A group of three telephones all intended to be density 0, had one in the line that was density 1, meaning people could not walk through them as intended since they're offset onto a wall.
Reagent tanks around LZ1 and LZ2 were on lower layers than nearby floor decals, this changes it so that they will show above floor decals properly.

## Why It's Good For The Game

Invisible walls are bad for the game, this fixes one.
Visual consistency & making sense is important for atmosphere, good for the game to have reagent tanks prioritised over floor decals.

## Changelog

:cl: ChainsawMullet

fix: Annex telephone no longer so dense it creates an invisible wall.
fix: Reagent tanks around LZ1 and LZ2 now properly display over floor decals.

/:cl:


Nanu308 (approved):
Mapping Good

PR #41, opened 2022-04-15
Set Squad Admin Tool

Geevies:
## About The Pull Request

Lets people with varedit perms set a human's squad.

## Why It's Good For The Game

Kinda neat for testing and gameplay I guess?

## Changelog

:cl:
admin: Add the ability for staff to set someone's squad.
/:cl:

### DIFF
--- a/code/modules/admin/player_panel/actions/physical.dm
+++ b/code/modules/admin/player_panel/actions/physical.dm
@@ -203,7 +203,7 @@
 	if(!selected_squad)
 		return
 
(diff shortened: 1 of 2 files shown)


Shad0vvs (commented):
We might want a return if they have no current squad.

Shad0vvs (changes requested):
We might want a return if they have no current squad so we don't log things that don't change anything.

Shad0vvs (approved):

PR #51, opened 2022-04-16
XvX Slash Damage Nerf

Geevies:
## About The Pull Request

Reduces the XvX Slash Damage modifier from x2.25 to x1.5.

## Why It's Good For The Game

I noticed that XvX tends to be either: The hives run into eachother and both die damn near instantly because slash damage is super high, or, the hives keep a 4 tile barrier between eachother because TTK is so low and no one wants to engage. With this change, there should be room to skirmish and still survive.

## Changelog

:cl:
balance: The XvX Slash Damage modifier has been lowered to x1.5 from x2.25.
/:cl:

### DIFF
--- a/code/__DEFINES/xeno.dm
+++ b/code/__DEFINES/xeno.dm
@@ -114,7 +114,7 @@
 
 #define XVX_UNIVERSAL_DAMAGEMULT 1.5 // Use to unilaterally buff every caste's DAMAGE against other xenos.
 
-#define XVX_SLASH_DAMAGEMULT 0.5 * XVX_UNIVERSAL_DAMAGEMULT // 0.75 | Applies to any abilities that uses brute damage or slash damage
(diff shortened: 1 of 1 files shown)


Watermelon914:
As the original developer on XvX, the slash damage is that high because it would often take an annoying amount of time to kill a xeno who could just run off and escape.

GoldenDarkness55:
Doesnt this make runners deal no damage? Ravs would be essentially unkillable aswell. I do agree TTK's a lil too fast now - 2.00 or 1.75 could work

stanalbatross (changes requested):
0.75 is likely too low, tune it up a bit.

Nanu308:
As Walter stated the opposite would become the issue where neither can kill each other due to the healthpools of xenos, put it at 1.5 instead @Geevies and we can work from that

Shad0vvs (approved):

PR #92, opened 2022-04-18
Taskboard Request: "Greenfire" Unification and Rework

Kaga-404:
## About The Pull Request

Per https://trello.com/c/AxLJ5ury/1-green-flame-rework

Point by point breakdown:

> 0) I wish to have all of these green fires made into one type

Previously, the variants of 'greenfire' were:
1. 'napalmb' reagent (Pyro Spec Type B Tanks):
- Its duration was much longer than the others at 50, but its intensity was equivalent to the other 'greenfires' at 15.
- These reagent values were used as a basis for the unification, and were not modified.

2. 'napalmgel' reagent (Incinerator Gel Tanks):
- napalmgel previously had an 'instant' duration, the same intensity as all the others (15), but also went 1 tile longer than most fuels (7 instead of 6).
- Outside of those properties, it had nothing special and functioned as a weaker, rapid-burning flamer fuel.
- During unification its only needed change was extending the duration to match Napalm B, described above.

3. flame_radius-generated greenfire (HIDP, Incendiary Mortar):
- flame_radius previously had default values equivalent to the HIDP's flame properties (14 duration, 15 intensity), but could be provided with parameters to change the generated fire's properties, while also automatically setting color by using one of the Napalm chemicals (UT/X/B) as a basis.
- flame_radius now has default values equivalent to UT Napalm, but otherwise still functions as before. HIDP and Incendiary Mortar have been updated to have their values as vars on themselves, which are pushed to flame_radius.
- The HIDP and Incendiary Mortar's direct values were brought in line with Napalm B as with napalmgel.

The intended sources of 'greenfire' were linked to the new 'Type B' properties, in addition to the changes noted here.

> 1) The new green fire should strip some armor while a mob/xeno is standing in the tile with the fire.

- "Type B" fire halves mob armor while on the tile, ignoring fire resistance.

> 2) The new green fire needs to create a slowdown to movement while a mob/xeno is standing in the tile with the fire/crossing it, think of sticky resin.

- "Type B" fire slows down humans, Xenos, and Yautja by the same amount when moving through it. Currently it is double the effectiveness of basic weeds but less than that of hive weeds.

> 3) The new fire needs to last for a long duration on the ground like current green pyro spec length, but it still needs to be able to be extinguished quickly from mobs on fire with it.

- All flames that fall under the "Type B" umbrella now have a base flamelevel of 55. This is exactly equivalent to the current Type B Pyro Spec flames.
- However, mobs that are lit on fire by it can be instantly put out by resisting or patting.
- Additionally, the flames on the ground can be instantly put out by an extinguisher or heavily reduced by acid. This was not specified within the notes, but follows logically in both how the compound would most likely function and from a gameplay perspective.

> 4) The tile’s damage should be high, but the dot on the mob not deal much dmg compared to normal fire.

- "Type B" fire still has a baseline damage value of 15, half that of normal UT-Napalm. This was semi-unified even previously, between the Pyro Spec, Gel, and Mortar/HIDP flame_radius.
- However, "Type B" fires now deal the same damage as UT-Napalm with each damage proc.

- Fires normally proc when moving across them, and "Type B" fires proc an additional time as well. This comes out to 30 damage per tile travelled plus 15 per Process().
- However, these fires only deal ~12 damage per mob fire stack proc, functionally identical to before but now with the instant extinguishing from the "Type B" properties. In comparison, standard UT-Napalm deals ~30 damage per proc.

## Why It's Good For The Game

Pre-approved Trello request.

Personally, this fleshes out and adds nuance to the fairly old fire code, as well as cleaning up a lot of the overall messiness from partial fire reworks over the years, which led to the current situation with greenfires.

## Changelog

:cl: Kaga
add: "Greenfire" has been unified into a single unique fire type. This is generally referred to as "Type B" Napalm/Napthal.
expansion: Flames with the "Type B" property behave as an area denial tool, remaining on the ground for a long time if not extinguished. Moving through them will slow you down (similar to weeds), deal heavy additional damage, and reduce your Armor while lingering inside.
expansion: Flames with the "Type B" property can be easily put out. They are instantly removed by extinguishers, and Xeno Acid has increased extinguishing effectiveness.
expansion: Fires applied to mobs by flames with the "Type B" property can be extinguished instantly by resisting and patting.
expansion: HIDP Grenades, Incendiary Mortar Shells, Gel Incinerator Tanks, and the Pyro Specialist's Type-B Tanks now all produce flames with "Type B" properties.
expansion: Gel Incinerator Tanks are now B-Gel Incinerator Tanks.
expansion: flame_radius now defaults to UT-Napalm (Flamer) values, instead of the old HIDP values. This only affects molotovs and the Whiskey Outpost Airstrike ability, which now generate using UT-Napalm flamer-equivalent fires instead of HIDP fires.
expansion: The Xenomorph Status Tab now displays your actual current armor, as modified by buffs and debuffs.
/:cl:


Nanu308:
Requested by me, will go into testing soontm

code/modules/mob/living/living_defense.dm
```diff
@@ -160,6 +160,9 @@
 		fire_reagent = new /datum/reagent/napalm/ut()
 
 	var/max_stacks = min(fire_reagent.durationfire, MAX_FIRE_STACKS) // Fire stacks should not exceed MAX_FIRE_STACKS for reasonable resist amounts
+	switch(fire_reagent.fire_type)
+		if(1)
```
stanalbatross, line 164: use the define here not just a number

code/modules/projectiles/guns/flamer/flamer.dm
```diff
@@ -571,6 +615,13 @@
 	for(var/i in loc)
 		if(++j >= 11) break
 		if(isliving(i))
+			switch(fire_variant)
+				if(FIRE_VARIANT_TYPE_B)
+					if(istype(i, /mob/living/carbon/Xenomorph))
```
stanalbatross, line 621: use isXeno() here

stanalbatross (changes requested):

stanalbatross (changes requested):
seeing how the armour HUD indicator is deprecated, it would be nice if that indicated how much of your current armour you have.

stanalbatross (changes requested):
messages about exoskeleton falling off should not appear for xenos without armour

Kaga-404:
> messages about exoskeleton falling off should not appear for xenos without armour

All xenos have an exoskeleton (technically referred to as a mesoskeleton by WY), and pieces of it bonding with the chemicals and having to be ripped off as they move through is what's 'slowing them down'.

There _is_ a check for whether a xeno has armor or not, but it specifically refers to _weakening_ the exoskeleton.

Kaga-404:
> seeing how the armour HUD indicator is deprecated, it would be nice if that indicated how much of your current armour you have.

Currently, Armor's numerical value is displayed in the Xeno Status tab, I just updated it to actually take into account buffs/debuffs. Previously, it only calculated based on integrity, which isn't currently used. Re-implementing the armor-integrity healthbar would be something for another patch, potentially.

stanalbatross:
> > messages about exoskeleton falling off should not appear for xenos without armour
> 
> All xenos have an exoskeleton (technically referred to as a mesoskeleton by WY), and pieces of it bonding with the chemicals and having to be ripped off as they move through is what's 'slowing them down'.
> 
> There _is_ a check for whether a xeno has armor or not, but it specifically refers to _weakening_ the exoskeleton.

should be fine then 



> > seeing how the armour HUD indicator is deprecated, it would be nice if that indicated how much of your current armour you have.
> 
> Currently, Armor's numerical value is displayed in the Xeno Status tab, I just updated it to actually take into account buffs/debuffs. Previously, it only calculated based on integrity, which isn't currently used. Re-implementing the armor-integrity healthbar would be something for another patch, potentially.

yeah, this can be done in a different PR if needed

rest of the code looks fine, we can throw this in for testing soon

Carnage068:
Wish I had read this PR before I got fucked over by green fire, I guess it would have explained a lot. The slowdown is nice, but does it really need to cut armor in half? Xenos with high armor already have to struggle with AP everywhere, so I feel like it's kind of excessive.

Nanu308:
TM into Merge soontm

Nanu308:
Might need further tweaks down the line but good to go into game for extended use and testing.

Will need to keep an eye on availability vs effectiveness.

code/game/objects/items/devices/binoculars.dm
```diff
@@ -448,13 +448,13 @@
...
+		flame_radius(cause_data, 3, target, , , , , )
 		explosion(target,  -1, 2, 3, 5, , , , cause_data)
-		flame_radius(cause_data, 3, target_2)
+		flame_radius(cause_data, 3, target_2, , , , , )
 		explosion(target_2,  -1, 2, 3, 5, , , , cause_data)
-		flame_radius(cause_data, 3, target_3)
+		flame_radius(cause_data, 3, target_3, , , , , )
 		explosion(target_3,  -1, 2, 3, 5, , , , cause_data)
-		flame_radius(cause_data, 3, target_4)
+		flame_radius(cause_data, 3, target_4, , , , , )
```
stanalbatross, line 457: what are these null args doing

code/game/objects/items/explosives/grenades/marines.dm
```diff
@@ -228,7 +232,7 @@
 	R.intensityfire = burn_level
 	R.rangefire = radius
 
-	new /obj/flamer_fire(T, cause_data, R, R.rangefire, null, flameshape, target)
+	new /obj/flamer_fire(T, cause_data, R, R.rangefire, null, flameshape, target, , , fire_type)
```
stanalbatross, line 235: just use fire_type = fire_type

stanalbatross (changes requested):
take a look back over at those args

code/game/objects/items/explosives/grenades/marines.dm
Kaga-404: Doesn't work, likely since it's because it's recursive and using the same var between both.

stanalbatross (approved):

PR #101, opened 2022-04-19
Fixes an old code comment, adds a new headset for VCs without squad comm access. 

Bram-KEP:
## About The Pull Request
Adds a new headset for vehicle crewman, identical to the Command one but without squad channels, as well as lowing it from VOICE_CRITICAL to VOICE_RAISED

## Why It's Good For The Game
VCs, despite being a corporal, often use their large text size and access to all 4 squad channels to spam them with callouts and orders, something that makes command jobs much harder, makes the marines understand orders less well, and in some cases the VC will spam orders directly contradicting commands.

## Changelog

:cl: Rem1
add: New headset for VCs, same as command one but without squad comm channel access. 
add: Lowers the text size when VCs speak on comms.
/:cl:


Bram-KEP:
(I assume I don't add the labels as I don't know how, 1st PR)

stanalbatross (approved):
Code is fine, DNM added for further discussion with maintainer team.

stanalbatross:
I for one am strongly opposed to this PR in its current state as it removes all the ways for VCs to communicate with standard Marines who are outside their vehicle, which is, if you've ever played VC, quickly apparent to be a very important part of communication for the job. 

To me it feels like this PR was made in knee-jerk reaction to two or three VC players spamming comms with , abcd messages, which should be dealt with in another way such as administratively.

Bram-KEP:
By the way, I have 16h as a VC, so I am not making this MR with no VC experience.

Bram-KEP:
> > This could be a reasonable statement if VCs needed communication with PFCs, something that in most situations they do not. they can contact engineers to get walls destroyed and advance with engi comms, contact medics for MEDAPC and medEVACs with medical comms, and ask command to order marines. this also favors usage of the tactics channel, as marines can buy it to coordinate with the APC, if that's needed.
> 
> It also stops all communication between marines and the APC, such as asking it to get out of the way, stop blocking the cades, etc. There are many occasions where communication is needed between the APC and squads and marines should not have to buy a tactics key solely to facilitate that. While I assume you think this will make APC drivers less annoying, without the information from squad comms, they will end up deaf to the cries of marines they are blocking/running over and end up being an even greater detriment.
> 
> Also they can just powergame command headsets anyway.

If every time the issue with something was "they could just powergame it" 3/4 of MRs and PRs would be closed. This goes back to my previous point that it readds a use for tactics encryption key, for communicating with the VCs, as well as still allowing communications in-between the marines and the APC, via Command, Medics, SLs, Engineers (for the cade issue) and promotes more intersquad cooperation.

Bram-KEP:
Also, currently APC drivers ignore the fact that marines yell for them to get out of the way and don't extract info from squad comms anyways, as APC is used like a battering ram.

Merrgear:
> Also, currently APC drivers ignore the fact that marines yell for them to get out of the way and don't extract info from squad comms anyways, as APC is used like a battering ram.

Removing the squad comms would only make this issue worse as now they cant hear when they are in the wrong place and dont realize. ,ABCD VCs are annoying yeah but they are a support tool that requires infantry to actually stay in communication with them BOTH WAYS to properly function. do you want to ask where medAPC is to be met with silence? do you want to get bodyblocked by the APC because you cant actually communicate with them to tell them to go away? on the VC side its annoying too as you cant tell the stupid engineers who didnt build APC accessible cades to fix their stuff when you have to run it over. MAYBE they could use tactics but no marine is going to spend points or go into req line just for a tactics key incase a VC is in the way. only part of this i actually agree is a good change is lowering the text size some as its very annoying but besides that VC comms are fine

Nanu308:
Not the way to go about it and neither a fan of it as there are quite a few ways to get around it with little to no issue.

PR #104, opened 2022-04-20
Synth eating

nauticall:
## About The Pull Request

"You sure you don't want some?"

What good is a synthetic at blending in if they can't even eat with humans?

This PR aims to give synthetics the ability to eat food. Until now whenever a synth tried to eat something, they'd be greeted by the clearly-taken-from-IPCs text "You have a monitor for a head, where are you gonna put that?". Does this make sense? Not really. But instead of just fixing that text I decided to go one further and let synths be able to consume stuff.

You might ask, is this realistic? Ask Ash and Bishop.

## Why It's Good For The Game

Synthetics were always stated in universe to be able to blend in almost flawlessly with humans, often not being able to find out if they're human or not except in certain conditions. A synth not being able to eat (you can only go "I'm not hungry" so many times) would eventually be a telltale sign that something's not right with them. So this PR aims to add immersion in that regard as well as just fostering more RP opportunities.

## Changelog

:cl:
add: Synthetics can now eat and drink.
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/food/snacks.dm
+++ b/code/game/objects/items/reagent_containers/food/snacks.dm
@@ -58,16 +58,15 @@
 		if(fullness > 540 && world.time < C.overeat_cooldown)
 			to_chat(user, SPAN_WARNING("[user == M ? "You" : "They"] don't feel like eating more right now."))
 			return
+		if(istype(M,/mob/living/carbon/human))
+			var/mob/living/carbon/human/H = M
+			if(H.species.flags & IS_SYNTHETIC)
+				fullness = 200 //Synths never get full
 
 		if(fullness > 540)
 			C.overeat_cooldown = world.time + OVEREAT_TIME
 
 		if(M == user)//If you're eating it yourself
-			if(istype(M,/mob/living/carbon/human))
-				var/mob/living/carbon/human/H = M
-				if(H.species.flags & IS_SYNTHETIC)
-					to_chat(H, SPAN_DANGER("You have a monitor for a head, where do you think you're going to put that?"))
-					return
 			if (fullness <= 50)
 				to_chat(M, SPAN_WARNING("You hungrily chew out a piece of [src] and gobble it!"))
 			if (fullness > 50 && fullness <= 150)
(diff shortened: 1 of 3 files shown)


code/game/objects/items/reagent_containers/food/snacks.dm
```diff
@@ -58,16 +58,15 @@
 		if(fullness > 540 && world.time < C.overeat_cooldown)
 			to_chat(user, SPAN_WARNING("[user == M ? "You" : "They"] don't feel like eating more right now."))
 			return
+		if(istype(M,/mob/living/carbon/human))
+			var/mob/living/carbon/human/H = M
+			if(H.species.flags & IS_SYNTHETIC)
+				fullness = 200 //Synths never get full
```
stanalbatross, line 64:
```suggestion
		if(isSynth(C))
			fullness = 200 //Synths never get full
```
would be a bit of a nicer check here

stanalbatross (commented):
also would be nice if you let them drink too

stanalbatross (changes requested):
now that synths can eat this stuff you should make it so that when `trans_to_ingest` is called (a proc that moves reagents that you eat into your bloodstream) it does nothing for synths - this will fix issues with them filling themselves up with nutriment and "tasting" stuff.

code/modules/reagents/Chemistry-Holder.dm
```diff
@@ -145,6 +145,8 @@
 
 /// Transfers to the reagents datum of an object
 /datum/reagents/proc/trans_to(atom/target, var/amount=1, var/multiplier=1, var/preserve_data=1, var/reaction = TRUE)
+	if(isSynth(target))
+		return
```
Shad0vvs, line 149:
```suggestion
```

code/modules/reagents/Chemistry-Holder.dm
```diff
@@ -169,6 +171,7 @@
 
 /// Transfers to object as ingestion
 /datum/reagents/proc/trans_to_ingest(var/atom/movable/target, var/amount=1, var/multiplier=1, var/preserve_data=1) //For items ingested. A delay is added between ingestion and addition of the reagents
+
```
Shad0vvs, line 174:
```suggestion
	if(isSynth(target))
		return
```

Shad0vvs (changes requested):
approve above

code/modules/reagents/Chemistry-Holder.dm
Shad0vvs, line 174:
```suggestion
```

Shad0vvs (approved):

PR #118, opened 2022-04-22
Changes the [ to \the [

stanalbatross:
## About The Pull Request

swaps out some improper instances of a non-swapping the and replaces them with the proper "\the".

Disclaimer. This was done using find & replace. There may be errors, but it does compile.

## Why It's Good For The Game

Less "the the"

## Changelog

:cl:
spellcheck: Fixed a few instances of "the" appearing twice in to_chat messages.
/:cl:


Concerned:
nah

This randomly throws backslashes across the code and (assuming \the [textstring] renders ingame as "textstring") probably makes a hell of a lot more messes than it fixes.

Geevies (changes requested):
you need to check that the thing inside the [ ] is an object, not a string, otherwise this won't work

stanalbatross:
> you need to check that the thing inside the [ ] is an object, not a string, otherwise this won't work

okay I will do this later when I am bored

PR closed for now

PR #132, opened 2022-04-24
CPR Can only be performed by one mob at a time

Varen:
## About The Pull Request

This adds a check to ensure that only one person can perform CPR on the same target at any one time.

## Why It's Good For The Game

There isn't a clear visual indicator in game on whether someone is performing CPR, suturing, or healing, and many players will attempt to help a downed marine by performing CPR all at the same time, while not understanding why it's failing despite the messages. 

This will help preventing getting in each other's way while CPRing someone, while also making the process make more sense: After all, how would two people perform CPR at the same time, they'd share a three-way kiss or something?

## Changelog

:cl: VarenRoth
qol: CPR can only be performed by one person at a time, avoiding multiple people causing CPR to fail.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_attackhand.dm
+++ b/code/modules/mob/living/carbon/human/human_attackhand.dm
@@ -1,4 +1,5 @@
 /mob/living/carbon/human/var/cpr_cooldown
+/mob/living/carbon/human/var/cpr_performer
 /mob/living/carbon/human/attack_hand(mob/living/carbon/human/M)
 	if(..())
 		return TRUE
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -26,10 +27,16 @@
...
 
 			if(M.head && (M.head.flags_inventory & COVERMOUTH) || M.wear_mask && (M.wear_mask.flags_inventory & COVERMOUTH) && !(M.wear_mask.flags_inventory & ALLOWCPR))
-				to_chat(M, SPAN_NOTICE(" <B>Remove your mask!</B>"))
+				to_chat(M, SPAN_NOTICE("<B>Remove your mask!</B>"))
 				return 0
 			if(head && (head.flags_inventory & COVERMOUTH) || wear_mask && (wear_mask.flags_inventory & COVERMOUTH) && !(wear_mask.flags_inventory & ALLOWCPR))
-				to_chat(M, SPAN_NOTICE(" <B>Remove [src.gender==MALE?"his":"her"] mask!</B>"))
+				to_chat(M, SPAN_NOTICE("<B>Remove [src.gender==MALE?"his":"her"] mask!</B>"))
+				return 0
+			if(cpr_performer != null)
```
stanalbatross, line 35:
```suggestion
			if(cpr_performer)
```

code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -1,4 +1,5 @@
 /mob/living/carbon/human/var/cpr_cooldown
+/mob/living/carbon/human/var/cpr_performer
```
fira, line 2: This should have the full type `/mob/living/carbon/human/var/mob/living/carbon/human/cpr_performer` and be cleared in `/mob/living/carbon/human/Destroy` to avoid even more cleanup problems.

Alternatively, do you really need to bother with all that ? There's probably simpler ways to go about this
For example, you could set a "CPR Start" cool down, preventing anyone from doing it during same duration. if do_after fails, clear it to allow someone else to.

fira (changes requested):

code/modules/mob/living/carbon/human/human_attackhand.dm
Varen: I've changed it to use a timer now instead, does it look fine to you now?

fira (approved):

stanalbatross (approved):

PR #134, opened 2022-04-25
QOL: Strip menu buttons rearrangement

JohnDoe50974:
## About The Pull Request
Bumps up (Exo)suit, Suit storage and Back slot buttons in the clickdrag strip menu on mobs to the top of the list.

## Why It's Good For The Game
The strip menu is essential for anyone playing any kind of medical role to use, it's impossible to do your job without it. The menu itself is functional, however the two most stripped item slots, Exosuit and Back, are near the bottom of the list, while slots that are much rarely touched like hands or head are at the top. Not very convenient when under pressure to revive multiple dead humans, right?
smenu1

The strip menu pop-up always opens up at the same size as well, and you have to scroll down even further if the person you're reviving has labeled their clothes and gear. Did I mention that using the mousewheel doesn't seem to work in this menu? At least it doesn't work for me. IIRC the strip menu resets itself back to top after an item has been removed as well, which is even less ideal.

In the interest of making medical roles more accessible and less janky, I believe this would be a good addition to the game. New strip menu pictured below.
smenu2
Now corpsmen or doctors don't need to fidget with the pop up's scroll bar just to find the Exosuit item slot to revive their patient.
Edit: I've now bumped Suit Storage button up as well, as it makes logical sense for it to be right next to the Exo(Suit) slot.

## Changelog
:cl: 50RemAndCounting
qol: Rearranged the Strip clickdrag menu items so that (Exo)suit, Suit storage and Back item slot buttons are always on the top of the body parts list.
/:cl:


Waseemq1235:
Exosuit storage is still at the bottom.. I don't know how to feel about it being so far from the exosuit.

stanalbatross (approved):

Geevies (approved):

PR #142, opened 2022-04-26
imrpove round-end notifs and gpanel integration

fira:
## Why It's Good For The Game

Round end notifications will work again without dubious shoestring serverside

Also exports additional data such as timing info that can be used by server staff

## Changelog

:cl:
server: added proper gpanel integration for game-end notifications and fixed some rough edges with it
/:cl:

### DIFF
--- a/code/controllers/subsystem/stats_collector.dm
+++ b/code/controllers/subsystem/stats_collector.dm
@@ -1,0 +1,17 @@
+/// Collects simple round statistics periodically
+SUBSYSTEM_DEF(stats_collector)
+	name      = "Round Stats"
+	wait      = 30 SECONDS
+	priority  = SS_PRIORITY_PAGER_STATUS
+	runlevels = RUNLEVELS_DEFAULT | RUNLEVEL_LOBBY
+	flags     = SS_KEEP_TIMING
+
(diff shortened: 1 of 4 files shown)


code/controllers/subsystem/stats_collector.dm
```diff
@@ -0,0 +1,17 @@
+/// Collects simple round statistics periodically
+SUBSYSTEM_DEF(stats_collector)
+	name      = "Round Stats"
+	wait      = 30 SECONDS
+	priority  = SS_PRIORITY_PAGER_STATUS
+	runlevels = RUNLEVELS_DEFAULT | RUNLEVEL_LOBBY
+	flags     = SS_KEEP_TIMING
```
stanalbatross, line 7: shouldn't this be indents between the vars and =s

fira: those aren't code indent, it's just spaces for alignment
tabs unfortunately don't work for alignment like this (despite it being what they were meant to be for..) since different viewers/platforms will display them as 2/4/8 spaces
i just took the habit of aligning multi assigns like that for readability

Geevies (approved):

PR #150, opened 2022-04-26
removes github tos breaching content

stanalbatross:
## About The Pull Request

Removes various words from the codebase that breach the github TOS

## Why It's Good For The Game

no more tg italian meatball fiasco


fira:
Do we even use these namelists at all ?.. 

Also they'll forever be in the git index ._.

Nanu308 (approved):
lgtm

Spircen:
I saw TG got taken down once cause something like this. Good thinking on doing this PR.

PR #155, opened 2022-04-26
Increases Resin Spike Count from 15 to 25

IsProbablyCatto:
## About The Pull Request

Changed resin spike count from 15 to 25.

## Why It's Good For The Game

Is a pretty unique building element, an increased cap lets the builder castes build more elaborate traps.

## Changelog

:cl:
balance: Changed Resin Spike Count From 15 to 25.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/resin_constructions.dm
+++ b/code/modules/mob/living/carbon/xenomorph/resin_constructions.dm
@@ -212,7 +212,7 @@ GLOBAL_VAR_INIT(resin_lz_allowed, FALSE)
 	desc = "Resin that harms any tallhosts when they walk over it."
 	construction_name = "resin spike"
 	cost = XENO_RESIN_SPIKE_COST
-	max_per_xeno = 15
+	max_per_xeno = 25
 
 	build_path = /obj/effect/alien/resin/spike
 


BKryth:
Why not just remove the cap, its just similar to sticky and it doesnt do that much damage, if spammed its as annoying as sticky spam.

Both sticky and spike resin dies in one click, only advantage spikes have are their sprites being really hard to click at times

Git-Nivrak:
> Why not just remove the cap, its just similar to sticky and it doesnt do that much damage, if spammed its as annoying as sticky spam.
> 
> Both sticky and spike resin dies in one click, only advantage spikes have are their sprites being really hard to click at times

Sticky should also have a limit, having to push through caves filled *everywhere* with sticky is cancer

stanalbatross:
> Why not just remove the cap, its just similar to sticky and it doesnt do that much damage, if spammed its as annoying as sticky spam.
> 
> Both sticky and spike resin dies in one click, only advantage spikes have are their sprites being really hard to click at times

It's more harmful than sticky, cannot be pushed over easily without clearing it, and notably is immune to the flamer, which is one of the most useful tools for clearing sticky. This makes it significantly stronger and therefore increasing its cap is a not insignificant buff to xeno defences. As such DNM on this PR for now, it will be further discussed with the dev team.

Nanu308:
Contributor Guidelines.

PR #162, opened 2022-04-27
Scrape Techs, TechTrees, and Refurbish Tech Impements

fira:
## About The Pull Request

This basically rips Tech objects and impelements from Techs/TechTrees and completely removes the later. Most of the objects have been refurbished as needed to be usable for events. Xeno Tech-Abilities are still in the game, but these don't work using items,, so could perhaps use a control panel to apply them or something...

## Why It's Good For The Game

By now, we've reused a lot of the components of Techwebs... but not the most important part. Techs. At this point I'd say it's foolish to think we'd do so in near future without a great deal of adapting anyway, so this is increasingly just being code (and runtime due to 2 Z-levels extra) bloat. 

## Changelog

:cl:
del: Removed Techs and Tech Tree system. We've done our best to find orphaned gadgets a new home in AdminLand...
add: Railgun Computers are back on the Almayer, but HC added safeties to prevent their unauthorized use.
add: Forget Spec Tokens. Added a 'Tank Coupon' for use in events. Use sparingly...
add: Added an standalone implant printer box, for 2 out of 4 special implants.
del: Removed RTO Backpack "Call Droppod" action, since there were none to call anyway.
/:cl:


code/modules/mob/living/carbon/human/human_abilities.dm
```diff
@@ -230,29 +230,14 @@ CULT
...
-		var/datum/tech/tech_to_use = i
-		list_of_techs += list("[tech_to_use.name]" = tech_to_use)
-
-	if(!list_of_techs.len)
-		to_chat(H, SPAN_WARNING("No droppods currently available."))
-		return
-
-	var/input = tgui_input_list(H, "Choose a tech to deploy at this location", "Tech deployment", list_of_techs)
+	to_chat(H, SPAN_WARNING("No droppods currently available."))
+	return
```
stanalbatross, line 234: Imo you should also remove the ability from RTO stuff (if I didn't notice it being removed elsewhere then whoops)

Carnage068:
wonderful

code/modules/cm_tech/implements/implants.dm
```diff
@@ -43,6 +7,43 @@
 	can_hold = list(/obj/item/device/implanter)
 	w_class = SIZE_SMALL
 
+
+/obj/item/storage/box/implant/picker
+	desc = "A sterile metal lockbox housing a printer for making hypodermic implant injectors."
+	foldable = FALSE
```
stanalbatross, line 13: reduce to 2 storage slots

code/modules/cm_tech/implements/railgun.dm
```diff
@@ -234,6 +192,10 @@ GLOBAL_DATUM(railgun_eye_location, /datum/coords)
 	if(!istype(H))
 		return
 
+	if(locked)
+		to_chat(H, SPAN_WARNING("Railgun Safeties prevent using the firing system."))
```
stanalbatross, line 196:
```suggestion
		to_chat(H, SPAN_WARNING("The railgun safeties prevent you from using the firing system!"))
```

code/modules/cm_tech/implements/tank.dm
```diff
@@ -1,33 +1,48 @@
...
+	icon_state = "pamphlet_written"
+	item_state = "pamphlet_written"
 
-/datum/tech/tank/on_unlock()
-	. = ..()
+/obj/item/tank_coupon/attack_self(mob/user)
+	if(QDELETED(src))
+		return
+	if(redeem_tank())
+		to_chat(user, SPAN_WARNING("[src] catches fire as it is read, resetting the ASRS Vehicle system!"))
```
stanalbatross, line 32:
```suggestion
		to_chat(user, SPAN_WARNING("\The [src] catches fire as it is read, resetting the ASRS Vehicle system!"))
```

stanalbatross (changes requested):

code/modules/cm_tech/implements/implants.dm
fira: It was intently 4 to allow VV'ing to more, this is an event tool only as is, but yeah, can lower down to 2

stanalbatross (approved):

PR #165, opened 2022-04-28
Fix shoes with no sprites for CL and Stowaway 

Spircen:
## About The Pull Request
There's no sprite for /obj/item/clothing/shoes since its an class from where others get atributes. So the sprite its missing for the item once you try to get it.

## Why It's Good For The Game
Fixes a weird direction error that creates some shoes with no sprite for the CL and to another outfit.

## Changelog
:cl:
fix: Now the uniform vendor for CL gives him a pair of laceup shoes instead of a missing object.
/:cl:


Agameofscones (approved):
LGTM

PR #166, opened 2022-04-28
Fixes omni-sentries being un-activatable after being repaired

### LINKED ISSUE #164: Unable to activate Omni-Sentries that have been repaired from the broken to working state
## Description

Fixing an omni-sentry when it is in the broken state, permanently has it turned off.

Well, it shouldn't. The omni-sentry should be turned on and not be turned off when repaired.

## Steps to reproduce

Launch Omni-Sentry
Attack it into the broken state.
Repair it back to a working state.
See that you're unable to turn it back on.

Git-Nivrak:
## About The Pull Request
Fixes a bug, as the title suggests

## Why It's Good For The Game
One less bug

## Changelog

:cl:
fix: Fixed omni-sentries being un-activatable after being repaired
/:cl:

Closes #164

### DIFF
--- a/code/modules/defenses/defenses.dm
+++ b/code/modules/defenses/defenses.dm
@@ -230,6 +230,9 @@
 	return TRUE
 
 /obj/structure/machinery/defenses/attack_hand(var/mob/user)
+	if(!attack_hand_checks(user))
+		return
+
 	if(isYautja(user))
 		to_chat(user, SPAN_WARNING("You punch [src] but nothing happens."))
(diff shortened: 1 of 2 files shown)


fira (commented):
This is all "defenses" behavior, not sentry-specific, and it's also where all the relevant logic is

So i'd wager it makes a ton more sense to do it there, in `/obj/structure/machinery/defenses/attack_hand` instead

fira (approved):

PR #176, opened 2022-04-30
Tweaks VC Radio Size 

Spircen:
## About The Pull Request
Reduces the Size of the Vehicle Crew from RADIO_VOLUME_CRITICAL to RADIO_VOLUME_RAISED

## Why It's Good For The Game
While VC needs the access to all the channels they are not over command but they have the same voice size to them so this should this makes eveyrthing more easy to identify. 

image

## Changelog
:cl:
del: VC no longer has the same size of voice to CIC personal. 
/:cl:


Geevies:
do their vendors also have a headset?

Spircen:
> do their vendors also have a headset?
image

No

Geevies (approved):

PR #177, opened 2022-04-30
Pref Pain Overlays

Geevies:
## About The Pull Request

Adds a preference option to disable generating pain overlays, because they tend to lag clients when they first get added, and randomly thereafter. Also changes the overlay strings into defines.

## Why It's Good For The Game

Lagging during combat, which is where you'll get pain overlays, isn't that much fun.

## Changelog

:cl:
add: Added a preference to disable pain overlays to the settings tab in character creation.
/:cl:


Watermelon914:
if these are causing lag, do a testmerge and remove them completely.
There is no reason for this to be a preference.

Watermelon914 (changes requested):
see above

PR #188, opened 2022-05-01
Github label now automatically gets applied. Features now cost only 5 GBP. Removes expansion tag

Watermelon914:
## About The Pull Request

Github labels now get automatically applied whenever something modifies the .github/ folder
Removes expansion tag (it isn't needed). Features now only cost 5 GBP

## Why It's Good For The Game

Expansion tag isn't really needed because it's hard to differentiate between an 'expansion' and a 'feature'. Additionally, the feature tag is just a trap for newcomers as it costs a whopping -10 GBP. All in all, this just means you need to fix 2 bugs to be net positive after making a feature.


Watermelon914:
@cmss13-devs/maintainers

fira:
dunno if we should remove it from add_label.py to maintain forward compat with upstream

you've seen how homebrewed changes are turning out with TGUI updates

Watermelon914:
> dunno if we should remove it from add_label.py to maintain forward compat with upstream
> 
> you've seen how homebrewed changes are turning out with TGUI updates

don't need to maintain forward compat with upstream

fira:
oh...

fira (approved):

.github/gbp.toml
```diff
@@ -3,16 +3,15 @@ reset_label = "GBP: Reset"
...
 "Atomic" = 2
 "Balance" = -5
 "Code Improvement" = 2
 "Fix" = 3
 "Grammar and Formatting" = 1
 "Hard Deletes" = 12
 "Logging" = 1
-"Feature" = -10
+"Feature" = -5
 "Tweak" = -3
```
fira, line 15:
```suggestion
```

PR #190, opened 2022-05-01
Sweeping perf changes to Area/Sound/Weather, targeted for Sorokyne

fira:
## About The Pull Request

Peformance on Soro always has been subpar but we never quite found why. It was clearly linked to the weather but despite so many attempts at fixing this, that still wasn't quite it. This is yet another attempt.

This PR targets several key points:
 * **Rendering method for Weather objects** - weather rendering has been done via `vis_contents` so far, to mirror an existing VFX over every turf indivdually. This now instead uses **overlays** on the **areas** as this is hopefully faster, and being the method the lighting system uses we are more familiar with it.
 * **Area handling for `/area/Entered`** - lighting subareas cause duplicate updates for Sound and Weather every time something steps (Xeno) between light levels. Having them properly handled (with `/area/var/master` and `/area/var/related`) cuts on that
 * **Duplicate sound updates** - In general, `update_ambience` was prone to sending empty updates. It's now much more aggressive at avoiding that.
 * **Proc Overhead Cleanup** - The Weather system is overly modular and does a lot of ping-pong with things that aren't actually being used otherwise. This cleans up a number of odd such calls/procs to lessen overhead.

## Why It's Good For The Game

Straighforward: more perf less lag less bloat

## Changelog

:cl:
refactor: Optimized several aspects of Sound Ambiences, Areas, and Weather
/:cl:

### DIFF
--- a/code/controllers/subsystem/weather.dm
+++ b/code/controllers/subsystem/weather.dm
@@ -30,6 +30,9 @@ SUBSYSTEM_DEF(weather)
 	// Holds necessary data for the current weather event
 	var/datum/weather_event/weather_event_instance
 
+	/// List of master areas to use for applying effects
+	var/list/area/weather_areas = list()
+
 /datum/controller/subsystem/weather/Initialize(start_timeofday)
 	// Set up our map delegate datum for supported maps
 	// The ONLY place where things should depend on map_tag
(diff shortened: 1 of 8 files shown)


Carnage068:
Running Sorokyne with less lag would be a dream ♥

stanalbatross:
testing will take place specifically on sorokyne

fira:
Don't merge this without a TM

fira:
Was TM'ed, apparently nobody could tell the difference, but it also was a short round
.... so probably that's not conclusive

stanalbatross:
I've got substantial feedback that this improved performance

stanalbatross (approved):

PR #198, opened 2022-05-01
Breaching Damage Nerf

Geevies:
## About The Pull Request

Removes the /turf path from the breaching damage boost list, meaning masterkeys (and other breaching damage weapons) are only useful for shooting down doors, girders, windows, window frames, grilles, barricades, and so forth.

## Why It's Good For The Game

Being able to destroy thick defenses in one click, as a PFC, is a little silly. With the upcoming breaching charges given to engineers, and the free availability to machetes, masterkeys no longer have a reason to be so powerful.

## Changelog

:cl:
balance: Masterkeys (and other breaching weapons) no longer deal bonus damage against turfs, meaning resin walls, and normal constructed walls. Other things like resin doors and airlocks are still affected as normal.
/:cl:

### DIFF
--- a/code/datums/elements/bullet_trait/damage_boost.dm
+++ b/code/datums/elements/bullet_trait/damage_boost.dm
@@ -1,5 +1,4 @@
 GLOBAL_LIST_INIT(damage_boost_breaching, typecacheof(list(
-	/turf,
 	/obj/structure/machinery/door,
 	/obj/structure/mineral_door,
 	/obj/structure/window_frame,
@@ -7,7 +6,7 @@ GLOBAL_LIST_INIT(damage_boost_breaching, typecacheof(list(
 	/obj/structure/surface,
 	/obj/structure/window,
 	/obj/structure/grille,
-	/obj/structure/barricade,
+	/obj/structure/barricade
 )))
 
 GLOBAL_LIST_INIT(damage_boost_pylons, typecacheof(list(


Bram-KEP:
Maybe do this PR AFTER the engi changes get merged..? they aren't even PRed rn, it's still a concept

Geevies:
> Maybe do this PR AFTER the engi changes get merged..? they aren't even PRed rn, it's still a concept

they should work faster then :)

GoldenDarkness55:
Machetes could use some buff vs structures if mkey's gone, they're pretty bad.

code/datums/elements/bullet_trait/damage_boost.dm
```diff
@@ -1,13 +1,12 @@
...
-	/turf,
 	/obj/structure/machinery/door,
 	/obj/structure/mineral_door,
 	/obj/structure/window_frame,
 	/obj/structure/girder,
 	/obj/structure/surface,
 	/obj/structure/window,
 	/obj/structure/grille,
-	/obj/structure/barricade,
+	/obj/structure/barricade
```
stanalbatross, line 9: this will also affect the XMB, make a separate one for masterkeys

stanalbatross (changes requested):

c4xmaniac:
Not a fan, mastery keys still require resource expenditure and space in your loadout for ammo to use the MK. Xeno structures are balanced around being easy and quickly made but weak, and the master key helps emphasize that. It takes 2-3 shells to destroy one thick resin wall.

BKryth:
This will make masterkeys literally useless, the reason its used a lot is because of its breaching capabilities and how easily accessible buckshot is

If the alternative to masterkey is a more limited item that is also role specific........then its not really an alternative, machete could be argued as an alternative but the huge drawback of it is its melee, you risk getting lunged breaking the door with a machete

neroisvcute:
Master key is only used to clear walls and doors. Removing this feature is basically removing the only useful this this attachment does.

stanalbatross (commented):
I would also rather just reduce its damage against walls (such as making it 1 shot and 2 bayo hits for a normal wall) than making it entirely useless against walls

JohnDoe50974:
Removing the bonus damage factor completely would virtually nullify the utility of the U7 underbarrel shotgun. The issue with this bonus damage remaining against doors and windows is that that removing these obstacles is already both fast and doable via other means. Airlocks can be forced open with a crowbar or a bayoneted rifle, resin doors take a pretty short time to destroy with bayonets already, and window frames can be quickly disposed of with a wrench. These tools are all much more common and readily available than the masterkey attachment, and if disposing of a wall is going to take 15 or more masterkey shots, players will stop using the attachment altogether, as it is redundant and inferior in these other use cases that do not involve taking down walls.

Are marines breaking down thick resin walls too fast? If so, I'd much rather that you keep the damage bonus vs. walls, but reduce it, rather than removing it outright. That way the U7 can both keep its current utility and justify its existence whilst not being a resin lawnmower at the same time. I'd say something like making the masterkey take 1-2 extra shots to take down a thick resin wall pointblank would be fine.

Diegoflores31:
-999999 gbp

PhantomEpicness:
Yeah sure the marines breach through thick resin walls too quickly, but removing the entire breaching ability over that is overkill. Especially considering the amount of walls Xenos can put out in an extremely short amount of time

ghostsheet:
Geeves I'd like the masterkey to still work on walls and doors like they've used to

Just change it so they don't work on resin structure as effectively

stanalbatross (approved):
code le good

Nanu308:
Fine with this going in for extended testing

PR #202, opened 2022-05-02
Aimed Shot size modifier fix.

Kaga-404:
## About The Pull Request

Fix for Aimed Shot since I found there's actually a MOB_SIZE_XENO_SMALL tier exclusively for Runners below MOB_SIZE_XENO. Also cleans up the way the damage is calculated, so that it doesn't call the damage proc multiple times based on size.

## Why It's Good For The Game

Makes the functionality less binary, while also slightly cleaning up the ability code.

## Changelog

:cl: Kaga
fix: Fixed an unintended oversight where mobs were taking damage calculated for runners from the Aimed Shot ability when used by an XMB or M42C sniper rifle. Normal-size Xenos now have their own tier for damage calculation.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1629,12 +1629,15 @@
 /datum/ammo/bullet/sniper/anti_materiel/on_hit_mob(mob/M,obj/item/projectile/P)
 	if(P.homing_target && M == P.homing_target)
 		var/mob/living/L = M
+		var/size_damage_mod = 0.8
 		if(isXeno(M))
 			var/mob/living/carbon/Xenomorph/target = M
+			if(target.mob_size >= MOB_SIZE_XENO)
+				size_damage_mod += 0.6
 			if(target.mob_size >= MOB_SIZE_BIG)
-				L.apply_armoured_damage(damage*1.2, ARMOR_BULLET, BRUTE, null, penetration)
-		L.apply_armoured_damage(damage*0.8, ARMOR_BULLET, BRUTE, null, penetration)
-		// 180% damage to all targets (225), 300% against Big xenos (375). -Kaga
+				size_damage_mod += 0.6
+		L.apply_armoured_damage(damage*size_damage_mod, ARMOR_BULLET, BRUTE, null, penetration)
+		// 180% damage to all targets (225), 240% (300) against non-Runner xenos, and 300% against Big xenos (375). -Kaga
 		to_chat(P.firer, SPAN_WARNING("Bullseye!"))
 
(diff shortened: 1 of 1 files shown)


fira:
Gonna sound stupid but why link this to Xenos at all ? The point of mob_size is that it's a common mob property

Kaga-404:
> Gonna sound stupid but why link this to Xenos at all ? The point of mob_size is that it's a common mob property

Holdover from how Aimed Shots were coded, could generalize it by removing the isXeno() check portion if needed.

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1629,12 +1629,15 @@
...
 		if(isXeno(M))
 			var/mob/living/carbon/Xenomorph/target = M
+			if(target.mob_size >= MOB_SIZE_XENO)
+				size_damage_mod += 0.6
 			if(target.mob_size >= MOB_SIZE_BIG)
-				L.apply_armoured_damage(damage*1.2, ARMOR_BULLET, BRUTE, null, penetration)
-		L.apply_armoured_damage(damage*0.8, ARMOR_BULLET, BRUTE, null, penetration)
-		// 180% damage to all targets (225), 300% against Big xenos (375). -Kaga
+				size_damage_mod += 0.6
+		L.apply_armoured_damage(damage*size_damage_mod, ARMOR_BULLET, BRUTE, null, penetration)
```
stanalbatross, line 1639: won't this do damage*0.8 and actually reduce the damage here?

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1654,14 +1657,17 @@
...
+		var/size_damage_mod = 0.5
 		if(isXeno(M))
 			var/mob/living/carbon/Xenomorph/target = M
+			if(target.mob_size >= MOB_SIZE_XENO)
+				size_damage_mod += 0.5
 			if(target.mob_size >= MOB_SIZE_BIG)
-				L.apply_armoured_damage(damage*1.5, ARMOR_BULLET, BRUTE, null, penetration)
-			L.apply_armoured_damage(damage*0.5, ARMOR_BULLET, BRUTE, null, penetration)
+				size_damage_mod += 1
+			L.apply_armoured_damage(damage*size_damage_mod, ARMOR_BULLET, BRUTE, null, penetration)
```
stanalbatross, line 1667: same here - if it doesn't pass any of the checks (i.e. is a runner) won't it just get *0.5?

stanalbatross (changes requested):

stanalbatross (approved):

PR #206, opened 2022-05-02
refactor: Fixes various internal issues with throwing objects launch_metadata

fira:
## About The Pull Request

(gitlab port)

This fixes a number of loose ends with launch_metadata used in throws that result in improperly cleared mob references. 

It also refactors launch to use a common `reset_throw` proc rather than several odds and ends in code doing it their own (partial) way. This should fix some rare observed race conditions when several sources handle it differently at same time.

## Why It's Good For The Game

 * Performance
 * Less references causing deletion failures for mobs
 * Less intermediate datums kept around for no reason

## Changelog

:cl:
refactor: Fixed various internal issues with throwing objects
/:cl:

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -11,21 +11,19 @@
 	var/mob/pulledby = null
 	var/rebounds = FALSE
 	var/rebounding = FALSE // whether an object that was launched was rebounded (to prevent infinite recursive loops from wall bouncing)
-
 	var/acid_damage = 0 //Counter for stomach acid damage. At ~60 ticks, dissolved
-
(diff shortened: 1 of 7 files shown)


fira:
TL;DR as of current if you throw an object, and that object is not deleted (explosion etc) and nobody else throws it, this keeps a lose reference to the mob preventing its deletion

MistChristmas:
I'm sure someone will come along with a more detailed report. But I'll quickly add the short version here. This breaks table flipping. Causing objects launched from tables including people to infinitely bash against walls, this causes quite the racket, and death, and gibbing.

stanalbatross:
see above

stanalbatross (changes requested):
fix the table thingy

PR #214, opened 2022-05-03
Balance: CO rifle meltable, CO tablet unmeltable

JohnDoe50974:
## About The Pull Request
Makes the m46c pulse rifle meltable again, makes command tablet unmeltable (and indestructible too. Is this in reference to mobs caught in explosions having their held items deleted? I'm not sure - I was told repeatedly that is a bug, not a feature.)

## Why It's Good For The Game
M46C being unmeltable is unjustifiable. The weapon already has an in-built magnetic harness and the weapon may be exclusively used by a role that should explicitly not be run-ahead-of-everyone frontlining and hosing down enemies with dummy strong bullets, which goes contrary to the weapon's current attributes which encourage doing just that.
On the other hand, something actually much more important like the command tablet is very bulky to carry around and can easily be lost and melted by comparison. Making it unmeltable should hopefully promote the importance of commanding and communicating and remind players that the Commanding Officer whitelist is not about burst PBing xenos half a screen ahead of the nearest squad rifleman and saying less than 20 messages throughout the round. :)

## Changelog
:cl: 50RemAndCounting
balance: The M46C pulse rifle is now able to be melted by xenomorphs again. The Command Tablet is now unmeltable.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -408,7 +408,6 @@
 							/obj/item/attachable/heavy_barrel,
 								)
 	flags_gun_features = GUN_AUTO_EJECTOR|GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER
-	//unacidable = TRUE
 	indestructible = TRUE
 	auto_retrieval_slot = WEAR_J_STORE
 
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -408,7 +408,7 @@
 							/obj/item/attachable/heavy_barrel,
 								)
 	flags_gun_features = GUN_AUTO_EJECTOR|GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER
-	unacidable = TRUE
+	//unacidable = TRUE
 	indestructible = TRUE
 	auto_retrieval_slot = WEAR_J_STORE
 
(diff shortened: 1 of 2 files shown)


JohnDoe50974:
I talked with Naut about making this PR beforehand and they okayed the changes.

c4xmaniac:
It’s unmeltable for the same reason the GL or RPG is unmeltable. A better change is to make it so that the CO rifle cannot be unlocked and is attached to the COs ckey even after the body is destroyed.

code/modules/projectiles/guns/rifles.dm
```diff
@@ -408,7 +408,7 @@
 							/obj/item/attachable/heavy_barrel,
 								)
 	flags_gun_features = GUN_AUTO_EJECTOR|GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER
-	unacidable = TRUE
+	//unacidable = TRUE
```
stanalbatross, line 411: remove instead of commenting out

commenting out stuff like this is bad practice

stanalbatross (changes requested):

code/modules/projectiles/guns/rifles.dm
JohnDoe50974: I thought the opposite was the case lol, no problem.

PalJohnson:
Tablet change is cool. I would just make the weapon CO locked and requires CO skill to use locking everyone out but CO. The issue with the gun is its being passed out or when a CO dies the weapon gets picked up and unlocked by a random grunt and now they have the same level of firepower as CO.

stanalbatross (approved):
requires balance approval to merge

Nanu308:
I'm fine with this trade-off, maybe CO's will be a bit more careful when near combat now.

PR #220, opened 2022-05-04
smartpistol auto-aim lock on ability

stanalbatross:
## About The Pull Request

Adds an ability for the smartpistol to "lock on" to a target, much like the sniper spec's ability. Once it is finished locking, it will produce a "stable" lock and all bullets fired from the gun will home into the target.

The lock is dropped when the weapon is unwielded, if the target goes over 7 tiles away, or after 15 seconds.

## Why It's Good For The Game

Smartpistol is underpowered right now, it does little damage and its utility is lowered by how inaccurate it can be. This should help to rectify that.

PLUS it's super cool.

## Changelog

:cl:
add: Added a new ability for the smart pistol - auto-aim. You can use it to lock onto a target and make your bullets automatically fire at it and home in at it while locked, after a short delay to lock onto the target. It can only be used while wielded.
add: the lock will drop on the pistol being unwielded, the target moving over 7 tiles away, or after 15 seconds. 
/:cl:


Lafrien:
But what are the differences between locking with already in-game mechanic from this built-in lock-on? Homing bullets? Also isn't the targets you'd lock-on with smart-pistol primarily "harassers" that are typically fast, how long would it take to lock?

Agameofscones:
Stan, remind me to resprite the smart pistol like I promised two years ago before you were a dev

Shad0vvs (changes requested):
>728 Exception has occurred: Cannot read null.l_hand
Having two smartpistols, selecting the aim of the one you will drop by pressing z on the other will cause above exception.

>Something is already locking on to Hal Joyce! Your gun's circuitry jams up and drops the lock!
Targeting someone of the same faction is the above message which is strange. Non iff targets have the same message and they aren't targetable.

You can spam middle mouse on multiple different targets forcing them to have the mark forever, preventing them from ever being locked on again.

>is too far away for the SU-6 Smartpistol's circuitry to get a lock!
If you start aiming at a target and un two hand the weapon (works most of the time, moreso when you're closer) you get the above after a bit(with the blank name)

Bram-KEP:
how long does it take to lock on?

ValoTheValo:
Finally lore accurate smart weapons

PR #222, opened 2022-05-05
Adds missing vehicle blockers in shivas

Git-Nivrak:
## About The Pull Request
Adds missing vehicle blockers in shivas around research and Med-Sec-Sci megastructure:
image
image

## Why It's Good For The Game
Hive locations should have vehicle blockers, That's how it is on every map and this MR makes it so it is on Shivas as well. (Usually the entire cave system is blocked off but since Shivas doesn't have one it ended up with none)

## Changelog

:cl:
add: Adds missing vehicle blockers in shivas around research and Med-Sec-Sci megastructure
/:cl:


Agameofscones (commented):
Put vehicle blockers under the destructible walls in the spiders caves (south research hab), and in the western interior of the mega-struct if they're not already there

Agameofscones (approved):
nice, let's get this out on the server

PR #224, opened 2022-05-05
kills and damage taken now carries over between evos

stanalbatross:
## About The Pull Request

when you evolve as a xeno your life kills and life damage taken is no longer reset to zero

## Why It's Good For The Game

consistency

## Changelog

:cl:
fix: when evolving/deevolving as a xeno, your life kills and life damage will now carry over between bodies
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
@@ -567,6 +567,12 @@
 		if(LIGHTING_PLANE_ALPHA_VISIBLE)
 			return XENO_VISION_LEVEL_NO_NVG
 
+/mob/living/carbon/Xenomorph/proc/get_life_kills()
+	return life_kills_total
+
+/mob/living/carbon/Xenomorph/proc/get_life_damage_taken()
+	return life_damage_taken_total
+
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
```diff
@@ -567,6 +567,12 @@
 		if(LIGHTING_PLANE_ALPHA_VISIBLE)
 			return XENO_VISION_LEVEL_NO_NVG
 
+/mob/living/carbon/Xenomorph/proc/get_life_kills()
+	return life_kills_total
+
+/mob/living/carbon/Xenomorph/proc/get_life_damage_taken()
+	return life_damage_taken_total
```
Geevies, line 574: don't use procs to read variables unless it's extremely likely to need an override in the future. procs have something called proc overhead so it's better to just read variables directly

Geevies (changes requested):

code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
stanalbatross: donezo

Geevies (changes requested):
do thing

Geevies (approved):

PR #228, opened 2022-05-06
QoL: viewing helmetless marines now gives a special status message in Overwatch console

JohnDoe50974:
## About The Pull Request
Adds a check in OW console code that writes (NO HELMET) in bolded orange letters if that marine does not wear an M10 marine helmet.

## Why It's Good For The Game
Trying to click any of the marine names on the squad monitor section will refresh the page and reset it back to the top. If a player's attempting to flip through marine cameras and clicks on a marine that doesn't wear a helmet, it's annoying to be forced to scroll all the way down and remember which person did you just click on, so you don't click on them again on accident and get your overwatch list scrolled back to top and then you scroll back down, accidentally click on that same name again and - you get the idea.
This is how it looks:
unknown

## Changelog
:cl: 50RemAndCounting
qol: Helmetless marines on Squad Overwatch Consoles now have a special status message indicating they're not wearing a helmet. Players no longer have to manually click on every marine's name to check if they're wearing a helmet or not.
/:cl:

### DIFF
--- a/code/modules/cm_marines/overwatch.dm
+++ b/code/modules/cm_marines/overwatch.dm
@@ -198,7 +198,6 @@
 				if(A)
 					area_name = sanitize(A.name)
 
-
 				switch(z_hidden)
 					if(HIDE_ALMAYER)
 						if(is_mainship_level(M_turf.z))
--- a/code/modules/cm_marines/overwatch.dm
+++ b/code/modules/cm_marines/overwatch.dm
@@ -198,6 +198,7 @@
 				if(A)
 					area_name = sanitize(A.name)
 
+
 				switch(z_hidden)
 					if(HIDE_ALMAYER)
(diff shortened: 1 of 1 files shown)


code/modules/cm_marines/overwatch.dm
```diff
@@ -198,6 +198,7 @@
 				if(A)
 					area_name = sanitize(A.name)
 
+
```
stanalbatross, line 201:
```suggestion
```
no need for an extra line here

stanalbatross (approved):
works fine but I would reccomend increasing the default width of the OW window and the columns

JohnDoe50974:
> works fine but I would reccomend increasing the default width of the OW window and the columns

Sure, but in a future PR. I'd rather like to see the overwatch screen overhauled to be much more usable.

PR #229, opened 2022-05-06
more medilathe items

stanalbatross:
## About The Pull Request

lets the medilathe print a few more things

## Why It's Good For The Game

lets the medilathe get used more

## Changelog

:cl:
balance: the medilathe can now print medevac stretchers, fixovein, and bone gel.
/:cl:

### DIFF
--- a/code/game/machinery/autolathe_datums.dm
+++ b/code/game/machinery/autolathe_datums.dm
@@ -391,6 +391,14 @@ datum/autolathe/recipe/medilathe/autoinjector/s30x6
 	name = "bonesetter"
 	path = /obj/item/tool/surgery/bonesetter
 
+/datum/autolathe/recipe/medilathe/bonegel
+	name = "bone gel"
+	path = /obj/item/tool/surgery/bonegel
+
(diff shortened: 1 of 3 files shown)


code/game/machinery/autolathe_datums.dm
```diff
@@ -399,6 +407,10 @@ datum/autolathe/recipe/medilathe/autoinjector/s30x6
 	name = "rollerbed"
 	path = /obj/item/roller
 
+/datum/autolathe/recipe/medilathe/surgerybed
```
Geevies, line 410: everything else seems fine, but i'm a bit concerned with the balance of adding even more surgical beds to the game. imo we should actually be reducing the amount

Geevies (changes requested):

code/game/machinery/autolathe_datums.dm
stanalbatross: Fairs, I'll remove them

Geevies (approved):

PR #230, opened 2022-05-06
Bipod Buff and Move and Look System Refactor

Geevies:
## About The Pull Request

Allows bipods to fire in a more intuitive 180 degree arc, rather than a weird 90 degree arc that isn't really telegraphed that well. This is done by adding more info to the face_atom proc return value by making use of the directional bitflags. NORTHEAST is equal to NORTH|EAST, likewise with NORTHWEST being NORTH|WEST. Thus, if you've deployed your bipod facing north and you fire NORTHEAST enough that your mob changes dir, it does a check to see if NORTH|EAST contains NORTH, and if it does, your bipod won't be dropped.

Additionally, replaced the event_handler / movement handler system with the signals system, which will hopefully help with lag problems.

## Why It's Good For The Game

Better bipods are cool, better movement handlers are better than better bipods.

## Changelog

:cl:
balance: Bipods can now fire in a near-180 degree arc rather than a 90 degree one.
refactor: Changed the movement handler code to use signals rather than a looping proc with a lot of datums. Some stuff may be broken, so make a Github Issue if you see one.
/:cl:


GoldenDarkness55:
Cant bipod just have a "locked" firing angle where it prevents you from firing outside of it entirely? Would be a cool payoff that u have to take it down and would be less stressful to actually fire it w/o worrying about weird angles and FF.

FlamermanGrills:
Another based MR from Mr.Geeves

stanalbatross (changes requested):
error when trying to overwatch a target as a queen via pressing alt and shift while in the qeye

image

happened after doing it twice

stanalbatross (commented):
rest seemed to be alright though, might want to check everywhere that it overrides

stanalbatross (approved):
no other issues detected

Nanu308:
Bipods may see a change at a later date, but fine for now.

PR #235, opened 2022-05-07
Performant Mob Verbs

Geevies:
## About The Pull Request

Replaces verbs on atoms and items with verbs on mobs because it's more performant or something like that. Also ripped out the Move to Top verb because I don't think anyone uses it?

## Why It's Good For The Game

More performance is always good.

## Changelog

:cl:
refactor: Tweaked how the Examine, Pickup, and Pull verbs work, they should still function as normal however.
del: Removed the move to top object verb.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/internal.dm
+++ b/code/game/objects/items/storage/internal.dm
@@ -9,7 +9,6 @@
 		CRASH("Internal storage was created without a valid master object! ([loc], [usr])")
 	master_object = loc
 	name = master_object.name
-	verbs -= /obj/item/verb/verb_pickup	//make sure this is never picked up.
 
 /obj/item/storage/internal/attack_hand()
 	return		//make sure this is never picked up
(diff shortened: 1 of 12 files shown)


code/game/objects/items/storage/internal.dm
```diff
@@ -9,7 +9,6 @@
 		CRASH("Internal storage was created without a valid master object! ([loc], [usr])")
 	master_object = loc
 	name = master_object.name
-	verbs -= /obj/item/verb/verb_pickup	//make sure this is never picked up.
```
stanalbatross, line 12: any way to rectify this via a check or whatever?

stanalbatross (changes requested):
this lets ghosts have the verbs in their right clicks - while it does nothing, it's quite annoying and cluttery and just adds more clutter to right click for admins.

code/game/objects/items/storage/internal.dm
Geevies: their locs will never be a turf, so yah it's no issue

PR #259, opened 2022-05-09
Nesting bug fix

Shad0vvs:
## About The Pull Request

We must double check so that the dead xeno 2000 zlevels away does not do things.

## Why It's Good For The Game

Bugfix

## Changelog

:cl:
fix: Properly fixes an unnesting bug and now also logs the unnest on the unnesting mob as well.

/:cl:

### DIFF
--- a/code/game/objects/structures/stool_bed_chair_nest/xeno_nest.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/xeno_nest.dm
@@ -78,14 +78,16 @@
 		if(H.stat != DEAD)
 			if(alert(user, "[H] is still alive and kicking! Are you sure you want to remove them from the nest?", "Confirmation", "Yes", "No") == "No")
 				return
-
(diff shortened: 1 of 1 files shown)


stanalbatross (approved):

code/game/objects/structures/stool_bed_chair_nest/xeno_nest.dm
```diff
@@ -78,14 +78,16 @@
 		if(H.stat != DEAD)
 			if(alert(user, "[H] is still alive and kicking! Are you sure you want to remove them from the nest?", "Confirmation", "Yes", "No") == "No")
 				return
-
+			if(!user.Adjacent(H) || user.stat || user.lying || user.is_mob_restrained())
```
Geevies, line 81: this should also check if buckled_mod is still exists

Geevies (changes requested):

Geevies (approved):

PR #268, opened 2022-05-10
Lifeboat Shuttle See-Through Obj/tiles

ghostsheet:
## About The Pull Request

Makes Part of the lifeboat see-through, allow them to see space turf or part of the docking port instead of 

This is in co-operation with @nauticall Sammy's lifeboat

We are going to be adding lifeboat in many small PRs so it's easier to review and have less merge conflicts

Before

After

## Changelog

:cl:
add: Lifeboat is now 10% more see through
/:cl:


Agameofscones (approved):
Fix the dropship turfs (same problem. default plating beneath) while you're at it.

Approved, LGTM

ghostsheet:
Can't do dropship, they're using old shuttle code.

Can only do this with new shuttle code

PR #271, opened 2022-05-11
adds medhud sunglasses

Waseemq1235:
## About The Pull Request
Adds extremely nerdy looking medhud "sun"glasses for the CMO, can choose between the sunglasses and traditional medhuds in their uniform vendor. Dark green = off | Light green = on
in-hand-sprites
off-right
on-right
on-front
off-front

## Why It's Good For The Game
Drip is always good! I gave this to the CMO only so that only the fancy officer gets fancy glasses. Plus, those glasses don't look like they would survive harsh combat, hence why I didn't give it to medics.

## Changelog

:cl:
add: CMO now gets nerdy-looking medhud sunglasses as an option in their uniform vendor.
/:cl:


code/modules/clothing/glasses/hud.dm
```diff
@@ -103,3 +103,9 @@
 	invisa_view = 2
 	toggleable = FALSE
 	actions_types = list()
+
+/obj/item/clothing/glasses/hud/health/sunmedhud
```
stanalbatross, line 107: the typepath can probably just be health/sunglasses, says it much more succinctly

stanalbatross (approved):
code is fine but this needs spriter review

PR #278, opened 2022-05-12
Remove resource files separation introduced with docker build

fira:
## About The Pull Request

Docker image building introduced a "proper" deployment stage (creating the image). This meant it became possible to bundle only the actually useful files in the image, through the RSC bundle. Regular icon/sound files would be missing on the image filesystem, rather than being duplicated, in and out of RSC bundle.

This PR undoes that by copying files anyway. This means the image gets about 100 MB bigger.

## Why It's Good For The Game

It's becoming increasingly apparent that while it was a good attempt, we're not fully ready to go that way. Mostly things like reboot sounds and instanciated map objects sneaked past this conversion and the later is a headache to deal with properly...

Also increases reproducibility of contributor testing to prevent even more of these slip by unnoticed

## Changelog

:cl:
server: removed resource files separation introduced with docker build (fixes things like invisible props on maps)
/:cl:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -91,8 +91,8 @@ COPY strings strings
 COPY nano nano
 COPY maps maps
 COPY html html
-COPY sound/dyn sound/dyn
-COPY icons/dyn icons/dyn
+COPY sound sound
+COPY icons icons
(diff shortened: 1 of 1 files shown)


Nanu308 (approved):
wholesome

PR #301, opened 2022-05-15
Support for flags by names in ooc

stanalbatross:
## About The Pull Request

Adds support for people to turn on a preference that shows their flag by their name in OOC chat, preference disabled by default.

Adds support for a config option that totally disables above system as well as the ip -> country system, config disables it by default.

The flag is based on where your IP is connecting from, IP-API is used to determine the country. The preference to have your flag is disabled by default, and can also be totally disabled on the server by commenting it out of the config.

## Why It's Good For The Game

Something to make OOC more engaging, can be turned off if it causes issues.

## Changelog

:cl:
add: added a preference to show your country flag by your name in ooc chat, turned off by default. If turned on it will show a flag based on where your IP is connecting from in OOC chat before your name.
config: added a config option to disable the above, turned on by default.
/:cl:


FransFeiffer (approved):
Flag images look fine, approved on that front.

Shad0vvs (approved):

PR #317, opened 2022-05-18
storage powercreep undoing

stanalbatross:
## About The Pull Request

tunes down the backpack storage powercreep that has been ongoing by removing logi backpacks (apart from for synths) and removing the free rocketbags in the OT room

## Why It's Good For The Game

because the logi backpack allows medics and everyone to bypass the 5 item satchel limit, it was powergamed every single round, and let medics carry the entire medbay around in their backpack as well as allowed for retarded loadouts like MD medic, dual pistol medic, etc etc.

This was a major irregualrity in the storage balance system which led to the bypassing of many many pre-intended aspects of the system as well as just general powercreep.

As such it and the free rocket bags are being removed in order to put storage back to the prior norm of 5 items for satchels 7 items for backpacks

if you wish to take 7 items, take a backpack.

They will remain for synths to allow for greater diversity in that aspect.

## Changelog

:cl:
balance: removed logi backpacks other than for synths
balance: removed the two free rocket backpacks in the OT room
/:cl:

### DIFF
--- a/code/datums/supply_packs/clothing.dm
+++ b/code/datums/supply_packs/clothing.dm
@@ -94,17 +94,6 @@
 	containername = "Lightweight IMP Backpack Crate"
 	group = "Clothing"
 
-/datum/supply_packs/backpack/logi
-	name = "Logistics IMP Backpack Crate (x4)"
-	contains = list(
-		/obj/item/storage/backpack/marine/satchel/big,
-		/obj/item/storage/backpack/marine/satchel/big,
-		/obj/item/storage/backpack/marine/satchel/big,
-		/obj/item/storage/backpack/marine/satchel/big
-	)
-	cost = RO_PRICE_NORMAL
-	containername = "Logistics IMP Backpack Crate"
-
 /datum/supply_packs/backpack/satchels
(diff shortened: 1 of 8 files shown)


JohnDoe50974:
Removing them from req altogether is pretty overkill. Keep them in the req and CO vendor at the very least - I'd rather COs have them instead of that silly XXL backpack in their locker that nobody ever uses. Plus logi IMP backpacks are insanely great for combat technicians, so I'd like at least a few marines to have a chance to get one if they really need it.

Unknownity:
the rocket backpacks in OT are for the guys putting their custom rockets inside and sending them to req/spec. where's the issue here exactly, if its being abused for other purposes you can simply limit what exactly you can put in @stanalbatross

Weuyn:
The OT rocket bag is incredibly useful for mixing chemicals and doing assemblies as things carried in inventory have no click delay. Will be sad to see it go.

Nanu308:
Many things have been powercrept, this is one of them unfortunately where on the original PR which added the backpacks it was a concern.

Nanu308 (approved):
Map fine

PR #318, opened 2022-05-18
Fixes unintended behavior with lunge stun fix

Git-Nivrak:
## About The Pull Request
As a side effect from the previous MR warriors could no longer free mobs they are lunging which was a main part of their gameplay (It was meant to just fix grabbing when not lunging)

## Why It's Good For The Game
Fixes an unintended removal of a feature

## Changelog

:cl:
fix: Fixes an unintended behavior removal as a side effect from the previous MR regarding this
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/Warrior.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Warrior.dm
@@ -79,10 +79,10 @@
 /mob/living/carbon/Xenomorph/Warrior/stop_pulling()
 	if(isliving(pulling) && lunging)
 		lunging = FALSE // To avoid extreme cases of stopping a lunge then quickly pulling and stopping to pull someone else
-		var/mob/living/L = pulling
-		L.SetStunned(0)
-		L.SetKnockeddown(0)
-	..()
+		var/mob/living/lunged = pulling
+		lunged.SetStunned(0)
+		lunged.SetKnockeddown(0)
+	return ..()
 
 /mob/living/carbon/Xenomorph/Warrior/start_pulling(atom/movable/AM, lunge)
 	if (!check_state() || agility)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/Warrior.dm
```diff
@@ -101,6 +111,11 @@
 			L.pulledby = src
 			visible_message(SPAN_XENOWARNING("\The [src] grabs [L] by the throat!"), \
 			SPAN_XENOWARNING("You grab [L] by the throat!"))
+			lunging = TRUE
+			addtimer(CALLBACK(src, .proc/stop_lunging), get_xeno_stun_duration(L, 2) SECONDS + SShuman.next_fire - world.time)
```
fira, line 115: This easily on the codebase's Most Wanted bugs/hacks list and it shouldn't be reused
It caused issues before (polio) and it still is (all stuns last 1.5 to 2.5x longer than tehy should due to stacking)

Thanksfully you shouldn't need it for this case, because you'll get the precise duration off the timer, without the need for tick compensation

fira (changes requested):

fira (approved):

code/modules/mob/living/carbon/xenomorph/castes/Warrior.dm
```diff
@@ -74,6 +76,14 @@
 /mob/living/carbon/Xenomorph/Warrior/throw_item(atom/target)
 	toggle_throw_mode(THROW_MODE_OFF)
 
+/mob/living/carbon/Xenomorph/Warrior/stop_pulling()
+	if(isliving(pulling) && lunging)
+		lunging = FALSE // To avoid extreme cases of stopping a lunge then quickly pulling and stopping to pull someone else
+		var/mob/living/L = pulling
+		L.SetStunned(0)
+		L.SetKnockeddown(0)
```
Watermelon914, line 84: L needs a more detailed variable name

code/modules/mob/living/carbon/xenomorph/castes/Warrior.dm
```diff
@@ -74,6 +76,14 @@
...
+	..()
```
Watermelon914, line 85:
```suggestion
	return ..()
```

Watermelon914 (changes requested):

PR #322, opened 2022-05-18
Fixes an exploit that allowed things to be spawned in null space & Slightly changes previous pull

CocaColaTastesGood:
## About The Pull Request
Fixes the above mentioned exploit alongside rewording the error message of my first pull.

## Why It's Good For The Game
Funny hellhound/queen in null space.

## Changelog

:cl:
fix: Fixes an exploit that allowed items to be spawned in null space.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/mark_menu.dm
+++ b/code/modules/mob/living/carbon/xenomorph/mark_menu.dm
@@ -109,8 +109,8 @@
 	switch(action)
 		if("choose_mark")
 			var/selected_type = text2path(params["type"])
-			if(!istype(selected_type, /datum/xeno_mark_define/)) // Hacky fix
-				message_staff("[key_name(usr)] has somehow entered an invalid path while choosing a mark. Possibly an attempted exploit.")
+			if(!istype(selected_type, /datum/xeno_mark_define)) // Hacky fix
+				return
 			var/datum/xeno_mark_define/x = new selected_type
 			var/datum/action/xeno_action/activable/info_marker/Xenos_mark_info_action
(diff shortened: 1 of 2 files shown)


CocaColaTastesGood:
Bad description. Players can use any path when creating a mark, allowing for that path to be spawned into null space. CM doesn't have things like singularity, but this can still be abused pretty easily.

code/modules/mob/living/carbon/xenomorph/mark_menu.dm
```diff
@@ -109,6 +109,8 @@
 	switch(action)
 		if("choose_mark")
 			var/selected_type = text2path(params["type"])
+			if(!istype(selected_type, /datum/xeno_mark_define/)) // Hacky fix
+				message_staff("[key_name(usr)] has somehow entered an invalid path while choosing a mark. Possibly an attempted exploit.")
```
Watermelon914, line 113: please get rid of the staff message and just early return here

Watermelon914: also remove the / at the end of your typepath, make it /datum/xeno_mark_define instead

CocaColaTastesGood: No staff message? I say possibly in the message, but the odds are that it's deliberate. The menu only shows subtypes of /datum/xeno_mark_define

Watermelon914: It's a useless message. Anyone who is able to check the code will see that it notifies staff members, so the only time this'd flag up is if someone found a bug. At that point, it should just runtime or send a log, not a staff message

PR #326, opened 2022-05-19
Warrior shield removal

Git-Nivrak:
## About The Pull Request
Removes shield slash from warrior and replaces it with healing

## Why It's Good For The Game
Shields bad, Also rebalances warrior to give it more durability at the frontline with the shield gone and encourage more risk & reward playstyle with healing being based on missing health.

## Changelog

:cl:
balance: Removed shielding slash from warrior, Health increased from 450 to 550, Armor reduced to 15 from 20, Warrior slashes now heal for (7 + 1 per enemy within 4 tiles, up to 3)% of missing health with a minimum of 15 and a maximum of 40.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/Warrior.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Warrior.dm
@@ -4,11 +4,11 @@
 
 	melee_damage_lower = XENO_DAMAGE_TIER_3
 	melee_damage_upper = XENO_DAMAGE_TIER_5
-	max_health = XENO_HEALTH_TIER_5
+	max_health = XENO_HEALTH_TIER_7
 	plasma_gain = XENO_PLASMA_GAIN_TIER_9
 	plasma_max = XENO_NO_PLASMA
 	xeno_explosion_resistance = XENO_EXPLOSIVE_ARMOR_TIER_4
-	armor_deflection = XENO_ARMOR_TIER_1
+	armor_deflection = XENO_ARMOR_MOD_MED
 	evasion = XENO_EVASION_NONE
 	speed = XENO_SPEED_TIER_7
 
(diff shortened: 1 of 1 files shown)


Carnage068:
The hell? Why'd you take off 5 armor?

Bram-KEP:
That PR title is very misleading. your PR doesn't simply remove shields, it gives warrior healing instead, removes 5 armor, and gives it +100 HP. This will have reward but a distinct lack of risk, with chungus probably being removed this is a straight buff.

Git-Nivrak:
> That PR title is very misleading. your PR doesn't simply remove shields, it gives warrior healing instead, removes 5 armor, and gives it +100 HP. This will have reward but a distinct lack of risk, with chungus probably being removed this is a straight buff.

Yeah no I am not going to write "Warrior shield removal and lifesteal and health  and armor changes" as the title, That's what the changelog is for. Removing 5 armor actually increases buck damage though it doesn't really matter since warrior was always 3 PBs to kill and it stays that way with the health buff- It will probably be 2 PBs if shotguns get buffed after the double shotgun MR gets merged. Also remember other sources of shields which warrior used to rely on are going to be gone such as Warden shield or fruits so you have to consider that as well. All in all I think warrior is going to stay the same since for example, You can't stack lifesteal ontop of your health like you can with shields.

Carnage068:
> > That PR title is very misleading. your PR doesn't simply remove shields, it gives warrior healing instead, removes 5 armor, and gives it +100 HP. This will have reward but a distinct lack of risk, with chungus probably being removed this is a straight buff.
> 
> Yeah no I am not going to write "Warrior shield removal and lifesteal and health and armor changes" as the title, That's what the changelog is for. Removing 5 armor actually increases buck damage though it doesn't really matter since warrior was always 3 PBs to kill and it stays that way with the health buff- It will probably be 2 PBs if shotguns get buffed after the double shotgun MR gets merged. Also remember other sources of shields which warrior used to rely on are going to be gone such as Warden shield or fruits so you have to consider that as well. All in all I think warrior is going to stay the same since for example, You can't stack lifesteal ontop of your health like you can with shields.

Just a question, but is Warden getting anything else? Or is it just losing shield and that's it?

Issue #311, opened 2022-05-18
[Bug] Dropping/Throwing a uniform while also wearing a uniform and armor instantly unequips and drops your armor.


Superwelder:
## Testmerges:

Not aware of any ongoing testmerges while observing this bug though it likely would also occur in a testmerge.

## Description

If you take someone's uniform off and pick it up, the moment your drop/throw the uniform away to free your hand up, your armor is removed followed by a message similar to "You drop the M3-L pattern light armor to the ground while unequipping USCM uniform".

## Difference between expected and actual behavior

Discarding a uniform you are not wearing should have no bearing on the armor you are wearing. Removing the uniform is pretty critical to obtaining useful gear and having your own armor and suit slot(usually a weapon) fall to the ground is incredibly inconvenient.

## Steps to reproduce

1. Ensure you are wearing a uniform with armor on top of it and have a second uniform ready to pick up.
![uniformbug1](https://user-images.githubusercontent.com/105765670/168953217-1100b02e-b77d-43ab-bdbe-96ad70110fb0.png)

2. Pick up the second uniform in a free hand.

![uniformbug2](https://user-images.githubusercontent.com/105765670/168953299-47a247b2-8d86-4d3f-bf23-a19f6314dc04.png)

3. Drop/Throw the second uniform and observe as your armor falls away along with anything attached to the armor like your gun.
![uniformbug3](https://user-images.githubusercontent.com/105765670/168953379-fbe0848f-e467-4801-b4f2-d44c6e96b35e.png)

## Issue bingo (not required to be all checked)

- [x] Issue could be reproduced at least once
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #335, opened 2022-05-21
Fixes throwing uniforms removing your armor

morrowwolf:
## About The Pull Request

Fixes the unintended behavior described in https://github.com/cmss13-devs/cmss13/issues/311

## Why It's Good For The Game

Bug bad UNGA

## Changelog

:cl: Morrow
fix: Fixes throwing uniforms removing your armor.
/:cl:

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -384,9 +384,12 @@ cases. Override_icon_state should be a list.*/
 
 /obj/item/proc/check_for_uniform_restriction(mob/user, obj/item/item)
 	SIGNAL_HANDLER
-	if(is_type_in_list(item, uniform_restricted))
-		user.drop_inv_item_on_ground(src)
-		to_chat(user, SPAN_NOTICE("You drop \the [src] to the ground while unequipping \the [item]"))
+	if(!istype(user, /mob/living/carbon/human))
+		return
+	var/mob/living/carbon/human/human_user = user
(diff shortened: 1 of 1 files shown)


MidwayAtom:
fixes #311

MidwayAtom:
closes #311

MidwayAtom:
Closes #311

MidwayAtom:
closes #311

MidwayAtom:
@morrowwolf Please link #311 for closure upon merge via keywords. Thank you R/S IM.

fira:
Closes #311 

Though I don't think this is a good solution at all

Between cutting things arbitrarily to human, and the fact that check looks to be in the wrong place

You also probably can't suppress drop here because then the unequip signal will still fire, etc

morrowwolf:
> Closes #311
> 
> Though I don't think this is a good solution at all
> 
> Between cutting things arbitrarily to human, and the fact that check looks to be in the wrong place
> 
> You also probably can't suppress drop here because then the unequip signal will still fire, etc

What other mobs that aren't in that human subtype would we be worried about with unequipping a uniform?  Predator/synths/monkeys are all under /mob/living/carbon/human.

Where would you like the check?

This functions as it should from testing in its current state.

Spircen:
I hope this PR its merged soon cause right now its a pain to check on Webbings or even a missclick on the front its a dead cause you lost your armor.

morrowwolf:
Other one looks better, closing this.

PR #336, opened 2022-05-21
Yautja Code Modernization

Geevies:
## About The Pull Request

Tweaks and redoes a lot of Yautja code to be more sane. There's too many individual changes to list, so I'll just put down the biggest changes here.

What I Can Remember:
 - Thrall weapons now have a human_adapted var, which makes them work the same way as if a Yautja would use them. For instance, the combi-stick is now actually usable.
 - Thralls can now actually pick their weapons when using radial menus because it used the wrong anchor initially.
 - Thralls now get the Claim Equipment verb because it wouldn't show up correctly all the time because it was getting attached to the bracer and the bracer stayed stationary.
 - Thralls now have 'Mentors', not 'Masters'. Please don't roleplay having a master.
 - Cloak cooldowns are now actually measured in seconds, instead of two seconds pretending to be a second.
 - Hard refs to linked bracers are now cleared when the bracer is destroyed.
 - The check random function has been completely revamped to not be copy pasted everywhere, and also made to not rely on usr. Same goes for the internal bracer procs.
 - Casters and blades now have a sane deployment system. Blades will now be the same blades you deploy, so your bloodstained and/or labeled blades will remain in that state as you use them.

## Why It's Good For The Game

Predcode is known for being preeeeeetty awful, this does a fair bit to fix that.

## Changelog

:cl:
add: Yautja thralls now refer to the thrallee as a mentor, not a master.
fix: Yautja thralls can now properly use their weaponry.
refactor: Refactored a whole lot of Yautja code, some things may be wonky.
/:cl:

### DIFF
--- a/code/datums/elements/yautja_tracked_item.dm
+++ b/code/datums/elements/yautja_tracked_item.dm
@@ -39,4 +39,6 @@
 		return TRUE
 	if(isHumanSynthStrict(carrier) && (carrier.hunter_data.honored || carrier.hunter_data.thralled) && !(carrier.hunter_data.dishonored || carrier.stat == DEAD))
 		return TRUE
+	if(istype(carrier, /mob/hologram/falcon))
+		return TRUE
 	return FALSE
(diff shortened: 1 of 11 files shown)


Unknownity:
"Yautja can now pick a hunter spear as their weapon."
I'd be against this, honor code states you cannot have multiple primary weapons, some people prefer getting the spears from the engine room and wearing them while using a whip/sword, etc.
if they became primary weapons then it would become a honor code violation

realforest2001:
Excellent all round, only issue is changing Thralls master to mentor. There's no mentorship, they're enslaved. Leave as is, please.

stanalbatross:
confirming that this was tested live and did not crash and burn

stanalbatross (approved):

PR #345, opened 2022-05-22
Xeno Language Understanding (Sort Of)

Geevies:
## About The Pull Request

Allows xenos to understand speech from languages they don't know, but with various letters taken out, making it pretty much impossible to decipher.

## Why It's Good For The Game

I love seeing people scream when xenos run at them, but I don't get that same experience when I'm playing xeno because all the words are swapped out for stock phrases. This should maybe alleviate that.

## Changelog

:cl:
balance: Xenos can now semi-understand human speech, but various letters are taken out and replaced with nothing to make it near impossible to decipher. This allows the vibe of the words to be maintained but the words themselves are incomprehensible.
/:cl:

### DIFF
--- a/code/modules/mob/hear_say.dm
+++ b/code/modules/mob/hear_say.dm
@@ -17,17 +17,7 @@
 		if (!speaker || (src.sdisabilities & DISABILITY_BLIND || src.blinded) || !(speaker.z == z && get_dist(speaker, src) <= world_view_size))
 			message = language.scramble(message)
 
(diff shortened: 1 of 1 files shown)


Carnage068:
Oh my god, thank fucking god. I was sick and tired of people just saying gibberish. It was incredibly dull.

v3nturis:
Is this similar to how xeno speech is partially understood by humans?

If one side understands the other better than I think it should be adjusted to be the same level of comprehension on both sides.

PR #386, opened 2022-05-25
Fixes multiple issues with laser designators

### LINKED ISSUE #372: Laser Designator Issues
## Description
A family of possible bugs with the designator/rangefinder/binoculars, such as interacting with them when you shouldn't be able to.

Mainly I am discussing laser designators, but all of this also applies to range finders and binoculars, as they all share some functionalities.

First of all, you shouldn't be able to range-find/laze when you're dead.

Also, you shouldn't be able to switch modes on the designator from far away. The switching sound location can also be far off from where it's supposed to be.

You can use designator functions without even holding a designator.

The following parts I am unsure about. I will describe what is possible, and I suppose you could tell me if it's a bug or not.

You can look through the designator, laze, and range-find while inside a container (e.g. tarp, locker, closet, coffin, even from inside the belly of a xeno). What makes sense is being able to look through it while inside of something, since that's basically just the binocular function. Doing the rest from there is a bit iffy.

You can look through a camera and range-find/laze.

You can look through binoculars out an APC window for an increased view (this makes a lot of sense, but I'm unsure if this is intended). You could also rangefind/laze in the window view.

### Range-finding when dead
1. Set designator to range-finding (green) mode.
2. Look through it.
3. "Use" it (press Z), exiting the large view.
4. Die without moving or doing anything else (can be tested by spawning dumb hostile xeno mob).
5. Now that you're dead, control-click somewhere to range-find.
6. Notice that you make the lazing sound and start the windup. This windup actually stops after a short delay, but you can spam this sound and windup while you're dead.

### Designator telekinesis
1. Notice a designator on the ground (this will work from very far away by looking through binoculars!).
2. Alt-click the designator.
3. Notice the designator change modes, even though you're too far away to be able to normally interact with it.
4. Also note that the designator-mode-switching "click" sound is loud and clear, meaning the sound is generated at your location instead of the designator's (likely a bug).

### Range-finding without holding a designator
1. Hold a designator (yes, I know you just read the above header, but bear with me).
2. Look through it.
3. While looking through it, either drop it or throw it away from you. This should cancel your view.
4. Control-click to start range-finding, notice the lazing sound is coming from the designator's location.

### Stealthy designator/rangefinder/binocular usage
1. Have a designator in hand.
2. Get in a container, e.g. a tarp, or wait for a xeno to eat you (without knocking you down though!).
3. Laze/range-find any suitable locations via control-click (after looking through designator of course). This is also useful to create a little bit of light where you're range-finding/lazing. If it doesn't work, look through it a few more times and toggle the modes via alt-click a few times. Sometimes it seems to need that in order for the laze/rangefind to work.

### Range-find via camera
1. Have designator in off-hand.
2. Go to DS camera console in hangar.
3. Look through Alamo cameras.
4. Switch hands to designator.
5. "Use" designator (i.e. look through it).
6. Control-click on camera view to start range-finding (perhaps lazing too; I didn't test lazing) and watch the dot appear in the camera view (which could be planetside).

### Increase view through APC window
1. Get in APC.
2. Have binoculars in inactive hand and have empty active hand.
3. Click APC window to look through.
4. Switch hands.
5. Use binoculars to increase view.

Git-Nivrak:
## About The Pull Request
As the title

## Why It's Good For The Game
Unintended / Exploitable behaviors

## Changelog

:cl:
fix: Fixed multiple issues with laser designators.
/:cl:

Closes #372

### DIFF
--- a/code/game/objects/items/devices/binoculars.dm
+++ b/code/game/objects/items/devices/binoculars.dm
@@ -116,9 +116,9 @@
 /obj/item/device/binoculars/range/proc/acquire_target(atom/A, mob/living/carbon/human/user)
 	set waitfor = 0
 
-	if(src.locs[1] != user)
+	if(src.loc != user)
 		return
-	if(A:vis_locs?.len > 0 && istype(A:vis_locs[1], /obj/screen/map_view)) // We have to use : here because vis_locs does not exist in areas, However the target will (hopefully) NEVER be an area.
+	if(get_dist(user, A) > user.client.view)
 		to_chat(user, SPAN_WARNING("You can't target through cameras."))
 		return
 	if(coord)
@@ -127,7 +127,7 @@
 	if(world.time < laser_cooldown)
 		to_chat(user, SPAN_WARNING("[src]'s laser battery is recharging."))
 		return
-	if(A.z != user.z || !istype(user.locs[1], /turf))
+	if(A.z != user.z || !istype(user.loc, /turf))
 		to_chat(user, SPAN_WARNING("You can't target from here!"))
 		return
 
@@ -229,7 +229,8 @@
 	toggle_bino_mode(usr)
 
 /obj/item/device/binoculars/range/designator/proc/toggle_bino_mode(mob/user)
-	if(user.action_busy || laser || coord || user.z != src.locs[1]?.z || get_dist(src.locs[1], user) > 1)
+	var/atom/src_loc = src.loc
+	if(user.action_busy || laser || coord || user.z != src_loc.z|| get_dist(src.loc, user) > 1)
 		return
 
 	mode = !mode
@@ -249,10 +250,10 @@
 /obj/item/device/binoculars/range/designator/acquire_target(atom/A, mob/living/carbon/human/user)
 	set waitfor = 0
 
-	if(src.locs[1] != user)
+	if(src.loc != user)
 		return
 
-	if(A:vis_locs?.len > 0 && istype(A:vis_locs[1], /obj/screen/map_view)) // We have to use : here because vis_locs does not exist in areas, However the target will (hopefully) NEVER be an area.
+	if(get_dist(user, A) > user.client.view)
 		to_chat(user, SPAN_WARNING("You can't target through cameras."))
 		return
 
@@ -264,7 +265,7 @@
 		to_chat(user, SPAN_WARNING("[src]'s laser battery is recharging."))
 		return
 
-	if(A.z != user.z || !istype(user.locs[1], /turf))
+	if(A.z != user.z || !istype(user.loc, /turf))
 		to_chat(user, SPAN_WARNING("You can't target from here!"))
 		return
 
--- a/code/game/objects/items/devices/binoculars.dm
+++ b/code/game/objects/items/devices/binoculars.dm
@@ -86,8 +86,6 @@
 		QDEL_NULL(coord)
 
 /obj/item/device/binoculars/range/clicked(mob/user, list/mods)
-	if(!istype(user.locs[1], /turf)) // Inside a xeno, tarp, tank etc..
-		return
 	if(!ishuman(usr))
 		return
 	if(mods["ctrl"])
@@ -123,8 +121,12 @@
 		to_chat(user, SPAN_WARNING("[src]'s laser battery is recharging."))
 		return
 	if(A.z != user.z)
-		to_chat(user, SPAN_WARNING("You can't lase through that!"))
+		to_chat(user, SPAN_WARNING("You can't target through that!"))
 		return
+	if(!istype(user.locs[1], /turf)) // Inside a xeno, tarp, tank etc..
+		to_chat(user, SPAN_WARNING("You can't target from here!"))
+		return
+
 
 	var/acquisition_time = target_acquisition_delay
 	if(user.skills)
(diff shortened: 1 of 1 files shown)


DoomLaser9:
I am testing this and finding quite a few issues.

This succeeds at blocking targeting from inside something (e.g. tarp), when using a rangefinder. However, it still allows targeting via a designator. Also for some reason the user message it gives when I try to use the rangefinder this way is `You can't target through that!` instead of `You can't target from here!`

It also succeeds at stopping you from alt-clicking to switch modes from afar. However, Alt-Clicking the designator never works now, even when it's in your hand (the only way to switch modes now is via right-clicking the designator, which is a real pain). I think alt-click should work even when the designator is on a tile next to you.

It fails to stop you from targeting through dropship cameras with the designator (both when dropship is in flight and when it's on the colony). Also keep in mind that the camera view could be on the same Z-level (e.g. looking at solaris ridge cameras via colony computers).

It does succeed in stopping camera view targetting with the rangefinder, but not when the Z-levels are the same.

This fails to stop you from targeting with a designator after dropping or throwing it, though I suspect it's because you misspelled `binoculars` as `bincoulars` in your code.

This succeeds at stopping you from targeting when dead, but not when stunned or unconscious (I just added a comment to the issue with more details on this).

Git-Nivrak:
> I am testing this and finding quite a few issues.
> 
> This succeeds at blocking targeting from inside something (e.g. tarp), when using a rangefinder. However, it still allows targeting via a designator. Also for some reason the user message it gives when I try to use the rangefinder this way is `You can't target through that!` instead of `You can't target from here!`
> 
> It also succeeds at stopping you from alt-clicking to switch modes from afar. However, Alt-Clicking the designator never works now, even when it's in your hand (the only way to switch modes now is via right-clicking the designator, which is a real pain). I think alt-click should work even when the designator is on a tile next to you.
> 
> It fails to stop you from targeting through dropship cameras with the designator (both when dropship is in flight and when it's on the colony). Also keep in mind that the camera view could be on the same Z-level (e.g. looking at solaris ridge cameras via colony computers).
> 
> It does succeed in stopping camera view targetting with the rangefinder, but not when the Z-levels are the same.
> 
> This fails to stop you from targeting with a designator after dropping or throwing it, though I suspect it's because you misspelled `binoculars` as `bincoulars` in your code.
> 
> This succeeds at stopping you from targeting when dead, but not when stunned or unconscious (I just added a comment to the issue with more details on this).

Thank you, Pretty sure this commit resolves all of these (tested them myself) but I'd  be nice if you can double check me on that since clearly you are great at finding bugs

DoomLaser9:
It appears the "mode switching" sound for the designator is still originating from the player instead of the designator's location (to test, toggle designator on one side of character, toggle it on another side, toggle it while holding, it all sounds the same). Although I'm not sure if this is really necessary to change.

Some unusual behavior I noticed. The following was tested by sending the Alamo to the surface and looking through the camera console in the hangar.
1. Targeting anything in the Alamo that's not the walls or floor (e.g. cockpit door, seats, crew attach point, intercom) via camera view (with either rangefinder or designator) results in the message `You can't target from here!` instead of `You can't target through cameras.`
2. First attempt at targeting via camera console results in looking up from the designator and facing north, instead of actually trying to target (second try then works). To reproduce, follow these steps:
   1. Send Alamo to surface.
   2. Stand next to right camera console in the hangar.
   3. Pick up designator from ground next to you.
   4. Switch hands.
   5. Click camera console.
   6. Click `Dropship Alamo #1`
   7. `X` to switch hands.
   8. `Z` to look through designator.
   9. Control-click the alamo floor. Notice the message `You look up from the laser designator.` and your character now faces north.
      1. Note that Control-clicking a non-floor/non-wall thing in the alamo instead seems to do nothing in this case.

Now for some "same-Z-level-camera-targeting" problems (for both designator and rangefinder), tested by standing to the left of the camera console on solaris ridge right below the hospital; these problems seem to have some similarities with the DS camera console problems:
1. Same issue as `2` above, test it by looking at `Filtration Plant #3`. Upon first-time opening of camera view, clicking floor makes you look up from designator, while clicking object with a "name" (that shows up on the bottom-left corner of dreamseeker window) does nothing.
2. Looking through the designator again while keeping the camera window open (avoiding the previously mentioned issue), Targeting the floor is properly blocked, giving the appropriate message. However, targeting an object (same kinds of objects as before), lets you successfully target them, including lazing them for CAS if the targeted object is outside a building (e.g. wooden barricade just below "right tile of 2-tile-wide operations airlock-door" on camera view of `Administration #3`).

Unrelated to cameras: Right-clicking a designator in-hand gives `Toggle Laser Mode` and `Toggle Rangefinder Display` options (this is normal/expected). However, right-clicking a designator on the ground, within a tile of you (or on same tile) does not give those options, even though you can change the mode via alt-click. I feel that, to maintain consistency, the laser mode toggle should be available via right-click when the designator is within one tile of you (might as well throw in the display toggle option, too). I'm not sure how difficult it would be to modify the right-click context menus like this. If this is implemented, it should also apply to the rangefinder's display toggle option as well.

Also, I'm not familiar enough with the codebase nor the DM language right now to give you a quick solution to the linter problems your code is creating.

Those were the problems I was able to find so far. I'm sure you're just falling in love with designators right now, right?

Shad0vvs:
Fix checks.

PR #394, opened 2022-05-27
Fix throw speeds of various items

DoomLaser9:
## About The Pull Request
Fixes the throw speeds of plasteel rods, sandbags, and wired glass tiles. Previously the throw speeds were set to 0 (they were the only objects I could find in the codebase with that value). Now they have been set to match the throw speeds of normal rods, empty sandbags, and glass tiles, respectively (all of which happen to be SPEED_VERY_FAST).

## Why It's Good For The Game
Currently the plasteel rods move extremely slowly when thrown. You could throw them, walk around ahead of them, and then be hit with the rods that you just threw. That doesn't seem very realistic, unless plasteel is so light that it just floats around on air.
The same applies to wired glass tiles and sandbags (full sandbags).

## Changelog

:cl: Laser9
fix: Fixed throw speeds of plasteel rods, wired glass tiles, and sandbags.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/rods.dm
+++ b/code/game/objects/items/stacks/rods.dm
@@ -53,7 +53,7 @@ GLOBAL_LIST_INIT(rod_recipes, list (
 	w_class = SIZE_MEDIUM
 	force = 9.0
 	throwforce = 15.0
-	throw_speed = 0
+	throw_speed = SPEED_VERY_FAST
 	throw_range = 20
 	matter = list("plasteel" = 3750)
 	max_amount = 60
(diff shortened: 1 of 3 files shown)


Superwelder:
I dunno if its related but sand bags, when filled, have the same thrown behavior. They travel so slowly you will actually hit yourself with them if you move in the direction you throw.

DoomLaser9:
> I dunno if its related but sand bags, when filled, have the same thrown behavior. They travel so slowly you will actually hit yourself with them if you move in the direction you throw.

The merge of this PR should be delayed until I can investigate this. It would be good to fix all affected items with one PR.

DoomLaser9:
It seems sandbags and wired glass tiles had the same issue. I have pushed a fix for all of them and updated the PR title and description.

PR #399, opened 2022-05-29
M56C MK.II

PhantomEpicness:
## About The Pull Request

Introduces the M56C MK.II, which can fire holotargetting rounds and fires 60% faster. Only available for COs and admin spawn.

`"The M56C is the latest and greatest replacement for the M56, featuring many improvements such as better ergonomics, 
target acquisition, efficiency, rate of fire and \"Smarter Holo-Targetting Bullets\" selectable with a flip of a switch. Backwards 
compatible with m56 magazines. 

\n Unfortunately they haven't fully replace the M56 in this sector due to shipping issues.\nYou may toggle firing restrictions by using a special action."`

Backward compatible with normal SG drums.

## Why It's Good For The Game
Requested by thwomper

## Changelog

:cl: TotalEpicness5
add: Adds a smartgun, the M56C MK.II, which sports 60% faster firing and can switch to holo-targetting ammo. Only available to COs
add: Adds m56C Holotargetting drums, which have smartgun holotargetting ammo
imageadd: Credit to thwomper for the M56C sprites!
/:cl:


hotimep:
INSTAMERGE NOW

c4xmaniac:
BASED NOW NOW NOW!!!

Chainsawicus:
Do COs really need more insanely strong frontline tools? They already have the protag rifle that incentivises them to suicide blindly as the vanguard of every push. Now they get a smartgun that fires nearly twice as fast (when the sg already has a ludicrous fire-rate) and makes everybody deal more damage. 

I feel like the CO should be less incentivised to spearhead every push, not more.

PhantomEpicness:
> Do COs really need more insanely strong frontline tools? They already have the protag rifle that incentivises them to suicide blindly as the vanguard of every push. Now they get a smartgun that fires nearly twice as fast (when the sg already has a ludicrous fire-rate) and makes everybody deal more damage.
> 
> I feel like the CO should be less incentivised to spearhead every push, not more.

this actually makes them not spearhead since SG should be more of a backliner.

MusselMain:
> Do COs really need more insanely strong frontline tools? They already have the protag rifle that incentivises them to suicide blindly as the vanguard of every push. Now they get a smartgun that fires nearly twice as fast (when the sg already has a ludicrous fire-rate) and makes everybody deal more damage.
> 
> I feel like the CO should be less incentivised to spearhead every push, not more.

The CO's special issue rifle is a risk/reward weapon, as it CAN be melted by acid, also their alternate choice was a standard Smartgun, however it was rarely used due to the fact that the M46C was significantly better, I think this will allow the smartgun to get some better use.
If you read what this change allows, this will allow CO's who intend to deploy to NOT frontline and NOT suicide rush like they do with the M46C (as even though the smartgun will have better firerate, the Holotargetting bullets aren't that great for damage but allow OTHER marines to do increased damage, what I'm saying is this is the support weapon that will prevent what you just described). Also incase you've never seen it, CO's have the highest Leadership skill possible (CO and XO have Elite level, whereas SO's have Expert, and SL's have Specially Trained), this means that there's a significant benefit to them being near the frontline, as their Move/Hold/Focus orders have a larger area of affect, last longer, and are more effective. While I do agree that Suicide rushing CO's are a problem, I also have to disagree with the statement that they're not supposed to be near the frontline.

hotimep:
from what i remember from discord convos this actually has ass damage right?

PhantomEpicness:
> from what i remember from discord convos this actually has ass damage right?

 holotargetting ammo has piss poor damage but its offset somewhat by the firerate.

GoldenDarkness55:
Whats the point of having RoF THAT high? Pretty sure u cant maintain current SG w/o an autoclicker? Would rather if CO rifle was nerfed and they kept normal SG rather than powercreeping them higher - why not give them a better armor aswell? lol

MusselMain:
> Whats the point of having RoF THAT high? Pretty sure u cant maintain current SG w/o an autoclicker? Would rather if CO rifle was nerfed and they kept normal SG rather than powercreeping them higher - why not give them a better armor aswell? lol

Since the APC and VC role are being removed, I really don't think the marines need any more nerfs

Pyrodes:
oh boy another CO protag gun :okbro:

silencer-pl:
>  which sports 60% faster firing

What. Why?

At most make it a custom skin for the existing smartgun. We need more enticement for COs to be on the frontline/backline and not on the ship or any other spot on the map like we need uh... Some smart analogy goes here.

I am all for COs having a SG, but not one that's such insanely buffed. CO's are not supposed to be killing machines and in the current meta we really don't need any more enticement for COs to ditch the CIC and go fight than we already have.

PhantomEpicness:
Further balance testing underway. Firerate buff may or may not stay

PhantomEpicness:
> > which sports 60% faster firing
> 
> What. Why?
> 
> At most make it a custom skin for the existing smartgun. We need more enticement for COs to be on the frontline/backline and not on the ship or any other spot on the map like we need uh... Some smart analogy goes here.
> 
> I am all for COs having a SG, but not one that's such insanely buffed. CO's are not supposed to be killing machines and in the current meta we really don't need any more enticement for COs to ditch the CIC and go fight than we already have.

Holotargetting ammo is extremely poor at killing things though I am testing it further to see if the effects from it are too powerful.

I'd rather have the power shafted to an SG since COs should not be spearheading and suicide charging and instead shadowing the frontline with orders and directing the marines if nesseaary. An weapon designed around shadowing the frontline is best for this (SG)

FransFeiffer (approved):
As far as sprites are concerned, this looks fine. 👌

Nanu308:
No, not a fan of this approach and secondly 60% is way too high a number.

PR #406, opened 2022-05-29
Fireman Carrying Skills

Geevies:
## About The Pull Request

Added fireman carrying to the following roles:
 - Hospital Corpsman
 - CLF Medic
 - CLF Leader
 - Freelancer Medic
 - Freelancer Leader
 - All UPP Roles
 - All Commando and Mercenary roles (UPP Commando, MARSOC, W-Y Whiteout, etc.)

Removed the police skill from UPP Privates and Engineers.

## Why It's Good For The Game

Logically, these roles would be trained to carry people around. Additionally, if removing the APC goes through, having a way to transport the wounded easier is nice.

## Changelog

:cl:
balance: Added fireman carrying to squad medics, CLF medics and leaders, freelancer medics and leaders, all UPP roles, and all Commando and Mercenary roles.
balance: Removed the police skill from UPP Privates and Engineers.
/:cl:

### DIFF
--- a/code/_onclick/human.dm
+++ b/code/_onclick/human.dm
@@ -90,10 +90,10 @@
 
 /mob/living/carbon/human/MouseDrop_T(atom/dropping, mob/user)
 	if(user != src)
-		return . = ..()
+		return ..()
 
 	if(pulling != dropping || grab_level != GRAB_AGGRESSIVE || !ishuman(dropping) || !(a_intent & INTENT_GRAB))
-		return . = ..()
+		return ..()
 
 	var/carry_delay = 3 SECONDS
 	var/list/carrydata = list("carry_delay" = carry_delay)
@@ -102,7 +102,7 @@
 
 	if(!skillcheck(src, SKILL_POLICE, SKILL_POLICE_SKILLED) && !(signal_flags & COMPONENT_CARRY_ALLOW))
 		to_chat(src, SPAN_WARNING("You aren't trained to carry people!"))
-		return . = ..()
+		return ..()
 
 	var/mob/living/carbon/human/target = dropping
 
@@ -116,7 +116,7 @@
 	SPAN_WARNING("You load [target] onto your back."))
 
 	if(pulling != dropping || !dropping || QDELETED(dropping))
-		return . = ..()
+		return ..()
 
 	grab_level = GRAB_CARRY
 
--- a/code/datums/skills.dm
+++ b/code/datums/skills.dm
@@ -134,6 +134,7 @@
 	// List of skill datums.
 	// Also, if this is populated when the datum is created, it will set the skill levels automagically
 	var/list/skills = list()
+	var/has_fireman_carrying = FALSE
 
 /datum/skills/New(var/mob/skillset_owner)
 	owner = skillset_owner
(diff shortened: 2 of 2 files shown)


morrowwolf:
GOOD CHANGE SPEEDMERGE

c4xmaniac:
Hospital corpsman shouldn’t have fireman carry. They get stretchers and should use them, fireman carry is a massive buff to medics - one that isn’t needed. Medic gameplay is already insanely easy, now you won’t even need to worry about transporting your patient when you decide to do something retarded like triage them on the front mid xeno push.

GoldenDarkness55:
Its fine for ERT roles but people were worried about this powercreep into medics happening when fireman carry was introduced. Unnecessary. Roller beds exist for a reason

MusselMain:
> Hospital corpsman shouldn’t have fireman carry. They get stretchers and should use them, fireman carry is a massive buff to medics - one that isn’t needed. Medic gameplay is already insanely easy, now you won’t even need to worry about transporting your patient when you decide to do something retarded like triage them on the front mid xeno push.

I agree, this allows medics to carry two patients at once (one on a roller or being dragged, the other on the shoulder), and will make tough retreats where medics are forced to choose which marine to save and which to leave less important, as you can save two instead of one

MusselMain:
> Hospital corpsman shouldn’t have fireman carry. They get stretchers and should use them, fireman carry is a massive buff to medics - one that isn’t needed. Medic gameplay is already insanely easy, now you won’t even need to worry about transporting your patient when you decide to do something retarded like triage them on the front mid xeno push.

Exactly, this allows medics to carry two marines at once, (one on roller or dragging, the other on shoulder) and this will mean medics don't have to choose which person they should save first.

stanalbatross:
Makes rollerbeds useless and is simple power creep

DNM

MusselMain:
> Makes rollerbeds useless and is simple power creep
> 
> DNM

Can we simply remove the Hospital corpsman from that list and keep it as is? ERT's have significantly limited resources, and so it makes sense for them to have this, (it also allows ERT's to have slightly better survivability, which is something they DESPERATELY need).

Diegoflores31:
roler beds become completely useless with this fr

Superreallycoolguy:
Absolutely don't merge this.

powercreep: the PR

like everyone says, medics have rollerbeds, they need to use them

silencer-pl:
Going to be unique and say that medics already have a way of fast transporting bodies and that's the roller bed. They don't need fireman.

No objections otherwise.

Bram-KEP:
I'm going to say medics having this is good, roller beds are clunky sometimes, and if people are worried it's powercreep, just make it so you can't fireman carry and drag someone on a roller at the same time.

IngloriouslyBad:
I think medics having an alternate way of transporting patients (after the shitbird runner or lurker 1 taps your rollerbed then points and runs away) is a GOOD change. Power gamers who use both bed and fireman carry are just sweatlords. Mechanically restrict them from doing both if you're that worried, but I really doint think it will be that big of an issue; unless you're upset about a medic being able to pull more than 1 person out of a complete retreat who would have perma'd if left behind.

Nanu308:
Powercreep and infringes on the reason for rollerbed's to exist and be utilized, currently i am opposed to this here, but this may change in the future.

PR #253, opened 2022-05-08
Japanese language OVERHAUL

stanalbatross:
## About The Pull Request

image

Rejoice, weebs of CM. Your favourite PR is finally here.

This PR fully overhauls the generation of Japanese language text in the game, making it 100% phonetically legal under Japanese sound rules and making it look natural and nice. Also adds it to CLF names, because why not. 

This code allows convincing-looking Japanese language text to be generated anywhere on demand!

Warning : I made all this code myself, no ports or anything.

## Why It's Good For The Game

MY IMMERSION

## Changelog

:cl:
add: Japanese-language text will now sound much more Japanese, as the entire system of generating it was reworked to make it consistent 100% with real Japanese.
add: added a 50% chance for CLF characters to get a randomly generated Japanese name.
/:cl:

PR #411, opened 2022-05-30
Fix for CLF names.

Segrain:
## About The Pull Request

Followup to #253. I have no idea how do you arrive from "randomly generated Japanese name" to "mash plausibly sounding syllables together and call it a name", but at the very least watch your `if` trees.

## Why It's Good For The Game

Is fix.

## Changelog

:cl:
fix: CLF males no longer have random chance of being named Unknown.
/:cl:


MusselMain:
Makes sense, I use to see 3 different CLF unknowns in a single ERT and was very confused by it.

Shad0vvs (approved):

Issue #392, opened 2022-05-27
Coffin Deconstruction Produces Metal


DoomLaser9:
## Description

Deconstructing coffins/crates (constructed crates) always produces 1 metal sheet, even though they're mostly constructed from wood.

## Difference between expected and actual behavior
It shouldn't be possible to turn wood into metal (even though it's an abysmal rate of 5 wood = 1 metal).

A coffin/crate deconstructed should produce wood (and probably more than one plank; after all, violently destroying a wooden barricade produces 3 wood planks).
A pred coffin should probably produce runed sandstone bricks (it matches the visuals and the "stone" open/close sound).

## Steps to reproduce

1. Construct coffin from 5 wood.
2. Open coffin.
3. Hit coffin with active, fueled welder.
4. Observe that coffin is gone and in its place is 1 metal sheet.
Note: metal is also produced by the constructed wooden crate and pred coffin.

## Issue bingo (not required to be all checked)

- [x] Issue could be reproduced at least once
- [ ] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #419, opened 2022-06-02
Coffin material fix

insertnamehere123:
## About The Pull Request

Fixes a bug with coffin deconstruction as stated in issue #392 by changing the material returned in the deconstruction of coffins and wooden crates from metal to wood.

## Why It's Good For The Game

Consistency.

## Changelog

:cl: Insertusername

fix: Coffins and wooden crates now return wood when deconstructed

/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/coffin.dm
+++ b/code/game/objects/structures/crates_lockers/closets/coffin.dm
@@ -16,7 +16,7 @@
 	if(istype(item, /obj/item/tool/crowbar))
 		new /obj/item/stack/sheet/wood(src.loc)
 		for(var/mob/in_view in viewers(src))
-			in_view.show_message(SPAN_NOTICE("\The [src] has been wrenched apart by [user] with a [item]."), 3, "You hear wood breaking.", 2)
+			in_view.show_message(SPAN_NOTICE("\The [src] has been wrenched apart by [user] with [item]."), 3, "You hear wood breaking.", 2)
 		qdel(src)
 		return
 	if(isrobot(user))
--- a/code/game/objects/structures/crates_lockers/closets/coffin.dm
+++ b/code/game/objects/structures/crates_lockers/closets/coffin.dm
@@ -12,6 +12,46 @@
 	else
 		icon_state = icon_opened
 
+/obj/structure/closet/coffin/attackby(obj/item/W, mob/living/user)
+	if(src.opened)
+		if(istype(W, /obj/item/grab))
+			if(isXeno(user)) return
+			var/obj/item/grab/G = W
+			if(G.grabbed_thing)
+				src.MouseDrop_T(G.grabbed_thing, user)      //act like they were dragged onto the closet
+			return
+		if(W.flags_item & ITEM_ABSTRACT)
+			return 0
+		if(istype(W, /obj/item/tool/weldingtool))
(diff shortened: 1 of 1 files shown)


Segrain (changes requested):
Firstly, this is way too much copypasting. If all you want to do is change which type of material is produced upon welding, then that can be done by adding a variable to represent precisely that.

Secondly, _why_ is it even welding? How does somebody weld a wooden coffin shut or cut it apart with a welder? Might as well change that to something more reasonable instead.

insertnamehere123:
In hindsight, a crowbar would probably work better than a blowtorch

insertnamehere123:
> Firstly, this is way too much copypasting. If all you want to do is change which type of material is produced upon welding, then that can be done by adding a variable to represent precisely that.
> 
> Secondly, _why_ is it even welding? How does somebody weld a wooden coffin shut or cut it apart with a welder? Might as well change that to something more reasonable instead.

Made some changes, is it better now?

code/game/objects/structures/crates_lockers/closets/coffin.dm
```diff
@@ -12,6 +12,18 @@
...
+	if(istype(W, /obj/item/tool/crowbar))
+		var/obj/item/tool/crowbar/WT = W
+		new /obj/item/stack/sheet/wood(src.loc)
+		for(var/mob/M in viewers(src))
+			M.show_message(SPAN_NOTICE("\The [src] has been wrenched apart by [user] with [WT]."), 3, "You hear wood breaking.", 2)
+		qdel(src)
+		return
+	if(isrobot(user))
+		return
+	user.drop_inv_item_to_loc(W,loc)
```
Watermelon914, line 25: Give W, M, WT more descriptive variable names

Watermelon914 (changes requested):

code/game/objects/structures/crates_lockers/closets/coffin.dm
insertnamehere123: I've made the requested changes, the variables should be a bit better now

 `/obj/structure/closet/coffin/attackby(obj/item/Prybar, mob/living/user)
	if(istype(Prybar, /obj/item/tool/crowbar))
		var/obj/item/tool/crowbar/CrB = Prybar
		new /obj/item/stack/sheet/wood(src.loc)
		for(var/mob/Mob in viewers(src))
			Mob.show_message(SPAN_NOTICE("\The [src] has been wrenched apart by [user] with [CrB]."), 3, "You hear wood breaking.", 2)
		qdel(src)
		return
	if(isrobot(user))
		return
	user.drop_inv_item_to_loc(Prybar,loc)
`

Segrain (changes requested):
Compare your procedure to the parental `closet/attackby()`. It checks whether the closet is opened, it has interaction with grabs and abstract items, and so on and on and on. It is not even that the differences themselves are that great - it is the fact that they exist. Let the parent procedure handle all of that and let the child only handle that in which it is different.
One way to do this would be to give the parent closet type a variable representing it material. Then modify _its_ `attackby()` so that special welder interactions - both for welding it shut and for cutting it apart - only happen if it is made out of metal, and add an interactiong with a crowbar so that it can pry open closet into wood planks if its material is wood. Then all you have to do for the coffin itself is adjust its variable to mark it as wooden instead.

PR #425, opened 2022-06-04
Unskilled Mortars

Geevies:
## About The Pull Request

Removed the engineer skill requirement from mortars.
Inspired by: https://github.com/tgstation/TerraGov-Marine-Corps/pull/6494

## Why It's Good For The Game

Mortars should be fairly easy to operate, you just dial the values to what gets called in, then drop the shell into the tube. This will also allow PFC fobbits to have some fun if the engineer has to deal with a cade breach.

## Changelog

:cl:
balance: You no longer need engineer skill to operate or deploy a mortar.
/:cl:

### DIFF
--- a/code/modules/cm_marines/equipment/mortar/mortars.dm
+++ b/code/modules/cm_marines/equipment/mortar/mortars.dm
@@ -49,9 +49,6 @@
 	if(isYautja(user))
 		to_chat(user, SPAN_WARNING("You kick [src] but nothing happens."))
 		return
-	if(!skillcheck(user, SKILL_ENGINEER, SKILL_ENGINEER_TRAINED))
-		to_chat(user, SPAN_WARNING("You don't have the training to use [src]."))
-		return
 	if(busy)
 		to_chat(user, SPAN_WARNING("Someone else is currently using [src]."))
 		return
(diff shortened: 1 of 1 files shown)


c4xmaniac:
A PFC wouldn’t know how to operate a mortar, the army(11-C) and marines (0341) have a special MOS for infantry that makes them mortar operators, there’s more to using mortars than dialing in coords. They can always buy a pamphlet to let them use it.

Sargeantmuffinman:
I disagree on this one. PFC pamphlets are there for a reason. And there's _usually_ one RTO or engineer sitting on the mortar at all times. It also dilutes the importance of those two roles.

MusselMain:
> I disagree on this one. PFC pamphlets are there for a reason. And there's _usually_ one RTO or engineer sitting on the mortar at all times. It also dilutes the importance of those two roles.

I personally think that while there shouldn't be a engineer requirement, I think that not every schmuck should be allowed to use the mortar, engineers have a very important role to play, however not every engineer follows it, and I've seen countless Bravo engineers rushing the frontline like some TF2 engineer, who get capped or killed nearly instantly meaning even less people to repair or use the mortar. This will also give the marines a slight chance when the front collapses, as FOB sieges can be very demanding for engineers, who are well equipped for fixing breaches, this way some PFC who's lost their gun or has been injured can still contribute to the hold without having a engineer have to babysit the mortar

tempelers:
This is another PR to make SL and RTO less special. Mortar should be restricted to engineering. A rifleman should never do mortar, not should a medic. They have their own specialized roles within the game. Making mortar less exclusive is like making CAS or OB less exclusive.

Nanu308:
Not a fan of this change, CM has enough players overall and skill requirements are needed. Besides the point of yes, albeit "easy" to operate they are easy after being trained into utilizing them as others have already pointed out.

PR #429, opened 2022-06-04
GAU Minor rework

Tisx1:
## About The Pull Request

For a while now the GAU has been referred to as the rubber bullet CAS gun due to its ineffectiveness. Imo this was due to the GAU's impacts being explosions and not bullet impacts. This MR intends to fix that

By using create_shrapnel, on each impact turf the game will create a single 0 range shrapnel that will represent the 30mm shell. The intention here is to make the 30mm GAU which is a ballistic gun using ballistic ammo, use the game's ballistic damage formulas instead of the explosive formulas. I believe the ballistic damage formulas are better for the smaller AoE that each GAU shell impact creates. 

The explosive power of each impact has been dialed back one tier to compensate for the new shrapnel. 

Also 2 new ammo datums in ammo_datums; they are simply the 2 types of ammo for the GAU. The values placed are those which I believe would be fair, but if you disagree and your tags are cooler than mine feel free to change.  

Also added in are 3 sound files to aid in "refreshing" the GAU. 

## Why It's Good For The Game

The biggest problem currently with GAU balance is the explosive damage model, due to how small each GAU impact is it has been difficult to balance the weapon, one tier up on EXPLOSIVE_THRESHOLD and suddenly you go from cute firecrackers to OHGODOHFUCK CAS JUST GIBBED 4 PEOPLE AND TOKE OUT HALF OF THE HIVE'S RESIN BUILDINGS.

With the ballistics damage model devs/contribs can fine tone the GAU much much more than before, hopefully making the weapon much easier to balance and hopefully not leave it to the wayside 

Updated sounds can be found below. They speak for themselves in terms of adding atmosphere to the game. 

## Changelog

:cl: Tisx
balance: Changed GAU impacts from a purely explosive damage model to a more ballistic one for greater fine tuning. 
spellcheck: Renamed GAU ammo types; standard is now PGU-100 Multi-purpose. HV is now PGU-105 Anti-Tank
balance: Current damage numbers for GAU are; 150 damage 10 AP for multi-purpose. 95 damage and 40 AP (2x the bullets however) for Anti-tank
balance: Point cost for both ammo types have been changed to be more equivalent with each other. 
qol: Getting directly hit by a GAU shell will now display a message (same as getting hit by any other bullet) 
soundadd: Added in some new sounds to the GAU to spice things up a bit.

/:cl:

[compressed.zip](https://github.com/cmss13-devs/cmss13/files/8905090/compressed.zip)


Sargeantmuffinman:
So, did some testing with the new GAU's ammunition. These were all against thick resin walls and against no xenos or marines.



4 GAUs with normal bullets:
https://user-images.githubusercontent.com/104228030/172031560-3715f2b5-0c7e-412e-a041-842d24ebbc36.mp4

4 GAUs with High Velocity bullets:
https://user-images.githubusercontent.com/104228030/172031578-6678bd0e-04da-4551-ad62-cd9406d7728d.mp4


1 GAU with normal bullets:
https://user-images.githubusercontent.com/104228030/172031586-e26b3046-5bae-4c39-86b9-21726743bdee.mp4

1 GAU with High Velocity bullets:
https://user-images.githubusercontent.com/104228030/172031594-7b628355-39aa-4ca7-941d-2f8462115ac3.mp4

Initial thoughts on this is that it brings the GAU up to the standard of all other CAS ammunition.
Normal 30mm bullets against thick resin walls might be a tad bit too much.
High velocity on the other hand fits perfectly against shredding resin.

I love it all. Probably needs maybe a damage tweak for regular 30mm against thick resin.
Haven't tested it against normal resin, probably shreds it like paper.

GoldenDarkness55:
Should probably reduce damage by half. It rips t3s and walls apart, easily becoming the best available CAS weapon/ammunition.

DesertFox2:
> Should probably reduce damage by half. It rips t3s and walls apart, easily becoming the best available CAS weapon/ammunition.

Wouldn't precisely say slice it in half, might actually be just a little too much. I say this very much deserves a test merge, but more importantly to be adjusted while its tested to find whats best.

A good band-aid on the CAS Question, however. Good PR

thwompa:
i think it'd be awesome if it did a lot of damage to walls but not nearly as much to xenos, so it could be used to soften up defenses in a flashy way

Bram-KEP:
I think it's perfect, just needs slightly reduced damage on thick resin walls. fully agree with sgtmuffin.

GoldenDarkness55:
>more DMG than missiles to t3s
>more DMG than missiles to walls
>more than half the price of missiles
>more range than missiles
>can be fired several times without reloading
>>>>>>>>>>>Fine/perfect
Lmao
Only other gun this competes with is laser - which deals no damage to walls, doesn't stun and a big chunk of it's damage is contained in DoT burn.
The DMG needs to go down by around half, unless you want this to become the only viable option with builds like 3 GAU 1 laser

fira (approved):
Makes sense to me

Tisx1:
> Status on TM ?

Not yet done.

stanalbatross:
I have seen this in action. It is FAR too strong. Able to take out crushers, ammo is super cheap, quite literally the best cas weapon and it makes rockets nigh useless.

sose96:
Very strong, I think it should be made more expensive and only toned down a little however. As bringing GAU back to an economy option will just end up with it being unused once more.

GoldenDarkness55:
If you raise damage by 30 but lower armor pen by 10, isnt this a buff? lol 75 cost increase irrelevant aswell.
HV probably still too strong even after that nerf.

Sargeantmuffinman:
Okay from the CAS perspective of this.
It is very strong, so damage needs to be toned down a bit on both normal and HV.
Cost wise, needs to be bumped up.
I suggest mirroring the prices of mini-rockets and incendiary rockets for the 30mm and HV ammo.

code/modules/cm_marines/dropship_ammo.dm
```diff
@@ -109,35 +110,38 @@
...
 		if(!debriscooldown)
 			U.ceiling_debris_check(1)
 			debriscooldown = 6
 		debriscooldown--
 		new /obj/effect/particle_effect/expl_particles(U)
+	sleep(11) //speed of sound simulation
+	playsound(impact, 'sound/effects/gau.ogg',100,1,60)
 
 
 /obj/structure/ship_ammo/heavygun/highvelocity
```
stanalbatross, line 133: maybe change paths?

stanalbatross (approved):
was TMed a few times, can be done again if needed

code/modules/projectiles/ammo_datums.dm
```diff
@@ -2856,6 +2856,27 @@
...
+
+/*
+//========
+					CAS 30mm impacters
+//========
+*/
+/datum/ammo/bullet/shrapnel/gau  //for the GAU to have a impact bullet instead of firecrackers
+	name = "30mm Multi-Purpose shell"
+
+	damage = 150 //More damaging, but 2x less shells and low AP
```
Nanu308, line 2868: Drop down to 115, will most likely still be quite powerful but not an instant destroyer of all.

code/modules/projectiles/ammo_datums.dm
```diff
@@ -2856,6 +2856,27 @@
...
+	penetration = ARMOR_PENETRATION_TIER_2
+	accuracy = HIT_ACCURACY_TIER_MAX
+	max_range = 0
+	shrapnel_chance = 100 //the least of your problems
+
+/datum/ammo/bullet/shrapnel/gau/hv
+	name = "30mm Anti-Tank shell"
+
+	damage = 95 //Standard AP vs standard. (more AP for less damage)
```
Nanu308, line 2877: With the amount of AP it has, drop down to 80.
We can tweak these numbers further down the line after seeing it in active gameplay some more.

Tisx1: done

code/modules/projectiles/ammo_datums.dm
Tisx1: done

stanalbatross:
still in testing on the server, haven't gotten a good chance to see it in action yet, probably fine tho

PR #434, opened 2022-06-05
Floor Refactor

Geevies:
## About The Pull Request

Refactors floor turfs to not use floor_tiles to handle their interactions, but do it themselves. Meaning each turf no longer has an object spawned inside it.

Also enables floor deconstruction with tool flags, such as REMOVE_CROWBAR, BREAK_CROWBAR, and REMOVE_SCREWDRIVER. Floors that can be removed and has these flags can be examined to see how to remove whatever it is that's on them.

## Why It's Good For The Game

Less lag, more features (turf deconstruction).

## Changelog

:cl:
refactor: Refactored floor code, everything should function as expected, floors can now be deconstructed.
/:cl:


PurpleCIoud:
Its finaly here, +1

Shad0vvs (approved):

Diegoflores31:
this means floor tiles can be extracted ?

fira (approved):

PR #439, opened 2022-06-06
Surgery in interiors tweak

JeserTheFox:
## About The Pull Request

Disallowing surgery in interiors outside of interior medical tiles, in water and in shuttles.

## Why It's Good For The Game

Disallows surgery in vehicles by default.

## Changelog

:cl: Jeser
balance: Surgery in vehicle interiors is now only allowed if the patient is located on interior medical tile. Surgery is impossible on water and shuttle tiles.
/:cl:

### DIFF
--- a/code/modules/surgery/surgery_initiator.dm
+++ b/code/modules/surgery/surgery_initiator.dm
@@ -9,6 +9,10 @@ proc/initiate_surgery_moment(obj/item/tool, mob/living/carbon/target, obj/limb/a
 	var/list/available_surgeries = list()
 	var/list/valid_steps = list() //Steps that could be performed, if we had the right tool.
 
+	if(user.z == GLOB.interior_manager.interior_z && !istype(target.loc, /turf/open/shuttle/vehicle/med))
+		to_chat(user, SPAN_WARNING("You can't perform surgery under these bad conditions!"))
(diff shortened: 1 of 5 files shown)


Git-Nivrak:
Good change though I still think APC should be removed instead of bandage fixed for the 90000 time

tempelers:
Good PR. This'll stop the five docs at the FOB plus the synth from all filling up the APC to do surgery.

code/modules/surgery/surgery_initiator.dm
```diff
@@ -9,6 +9,10 @@ proc/initiate_surgery_moment(obj/item/tool, mob/living/carbon/target, obj/limb/a
 	var/list/available_surgeries = list()
 	var/list/valid_steps = list() //Steps that could be performed, if we had the right tool.
 
+	if(user.z == GLOB.interior_manager.interior_z && !istype(target.loc, /turf/open/shuttle/vehicle/med))
```
Geevies, line 12: a better solution might be a var on turfs that goes like "supports_surgery" true or false. check if the loc is a turf, then check if it supports surgery

Geevies (changes requested):

Geevies (approved):

PR #440, opened 2022-06-06
APCs balance tweaks

JeserTheFox:
## About The Pull Request

1. Added transport without FPWs interior.
2. Reduced reload time on FPWs to 10 seconds.
3. Replaced walls with void in some parts of APC interiors to prevent weeds spreading to non-walkable tiles.
4. Tweaked MED APC interior to not have medical interior turfs, removed Operating table, slight rearranging.
5. MED APC is dedicated mobile triage instead of surgery vehicle, doubled revivable dead bodies slots from 4 to 8.

## Why It's Good For The Game

Changes MED APC role from readily available surgery to dedicated mobile triage and medical resupply vehicle, shifting prioritized surgery locations from front to FOB and Almayer.

## Changelog

:cl: Jeser
add: Added transport without Firing Ports Weapons interior.
fix: Replaced floor with void in some parts of APC interiors to prevent weeds spreading to non-walkable tiles.
balance: Reduced reload time on Firing Ports Weapons to 10 seconds.
balance: MED APC interior no longer supports surgery, operating table was removed. Back half of interior was rearranged to improve triage capability.
balance: MED APC's revivable bodies slots were increased from 4 to 8 in accordance to the dedicated triage role.
/:cl:


c4xmaniac:
The MED APC still going to be a dedicated surgery vehicle because the doctors will just use the surgery table instead. The only thing that changes is that the surgery will take slightly longer, and you can't remove larvae in the APC. The same thing happens in the CMD APC, and the transport APC.

AGasCanister:
>Reduced reload time on FPWs to 10 seconds.

We turning this into Battlefield 4!! Prepare the C4s

PhantomEpicness:
local chad dev comes out of a warzone just to save APC

Carnage068:
Based as fuck, good going, my guy.

Nanu308 (approved):
Mapping LGTM

Diegoflores31:
i would recomend removing the Body scanner .. it makes no sense inside the apc if there is no surgery. it would be better if instead was changed maybe for an sleeper or just outright removed since localizing organ damage makes no sense if you cant fix it and FOB already has the crate body scanner so . it just eats 2 tiles from the already cramped space

PR #450, opened 2022-06-09
Acid Blood Tweaking 2

FoxyStalin:
Lowers Acid Damage again. Higher than it's old 12,  lower than it's current 25

## Why It's Good For The Game
Acid Blood at it's current damage is very punishing to marines for...doing what they're intended to do?
It's very possible to take more damage from someone shooting a xeno that's on you than you would take from  just letting the xeno slash you.

Acid damage was at this high level initially, and it was nerfed back then because it's just frustrating to play against.  Time for history to repeat itself.

## Changelog

:cl: FoxyStalin
balance: Lowered Acid Blood damage down to 15
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
@@ -220,7 +220,7 @@
 	var/list/tackle_counter
 	var/evolving = FALSE // Whether the xeno is in the process of evolving
 	/// The damage dealt by a xeno whenever they take damage near someone
-	var/acid_blood_damage = 25
+	var/acid_blood_damage = 15
 	var/nocrit = FALSE
 
 


BKryth:
The author of this PR:

MusselMain:
> The author of this PR: 

I have to disagree a bit. The damage for acid blood is kind of goofy, in the three years since I started playing on this server, I've had many close calls fighting a lone xeno where It's been an incredibly close fight (often coming down to me being very injured with my bootknife/gun VS a nearly critted xeno), and honestly I find it a bit problematic that I've had at least 9 times where I successfully crit said xeno then begin stabbing them (as I'm either out of ammo or have lost my firearms) to finish them off, only to hardcrit from 50 acid damage as I'm splashed twice by the xeno I'm trying to kill, I'm then either captured by said xeno, or die to my wounds. I get that this is a defense mechanism and is lore accurate, however as it currently is, 2 splashes deal around a quarter of your health and if you were already injured this could be a deathblow. I want to emphasize that I don't think this feature should be removed, but from my experience 25 is pretty high.

Nanu308:
Read the contributor guidelines.

PR #465, opened 2022-06-12
Testmerges-New Armor Set

cnaper-dodo:
Different variations of armor are good.But here are the benefits of these variations..It is practically not there.Everyone uses light, medium, and rarely M4.Heavy and b12 are not used, because their indicators are much better than usual, and you have to be disabled.So I'm going to add a new set of armor, with its advantages and disadvantages.
A set of heavy assault armor.
It is an expensive and very heavy set of armor.Due to too heavy weight, it is recommended to wear no more than 2 hours, due to the risk of developing deviations in the back and joints.
armor characteristics:
Armor Melee = 45
Armor Bullet = 30
Armor Bomb = 25
Armor Internal Damage = 25
Armor Bio = 25
Armor Slowdown = 0.15
Once every 10 minutes gives 5 units of brut damaga, due to too much mass.
helmet characteristics:
Armor Melee = 40
Armor Bullet = 35
Armor Bomb = 5
Armor Internal Damage = 25
Armor Bio = 5
due to too much mass, it gives 5 units of brut damaga every 10 minutes
due to the small viewing slit, he is able to see only from the front.
I propose to make it purchasable in cargo / in SL vendors and engineers for points.For engineers, it will be a convenient armor for repairing barricades in the thick of things.For SL, it is an opportunity to lead people without fear that they will fall behind

73248453-0a2cfed6-caa2-44be-893f-6fcdafc13623.png)

.


Bram-KEP:
According to ur commits u got something messing with chem code??

Diegoflores31:
literally CO armor

silencer-pl:
There are so many hidden changes in this MR its not even funny, like goddamn :X

MusselMain:
I'm not sure, but it seems like you have the armor slowdown currently set to a value that makes it more maneuverable then CLF armor, I might be reading that wrong but that seems a bit crazy

thwompa:
so many things wrong here. closing until everything here is legible and coherent.

Unknownity:
Why do you have the research warcrimes commits inside your PR bro???? I'm supposed to post that!!

PR #472, opened 2022-06-14
Rebalances M46c, adds new cosmetic SG variant

JohnDoe50974:
## About The Pull Request
Rebalances the m46c, reducing its damage output among other things, details below. Adds a co-exclusive cosmetic variant of the smartgun that's ID-lockable for mostly fluff reasons, alongside a new weapon case that replaces the standard M56B case in CO vendor. M56C sprites were made by Thwomper and originally appeared in [!399](https://github.com/cmss13-devs/cmss13/pull/399) where they were reviewed and approved. I've reused them here with Thwomper's permission. 
All of the changes here were discussed with Nanu, Nauticall and the CO council beforehand and were agreed upon.

## Why It's Good For The Game
M46C in its current state encourages maximally aggressive play when a CO is deployed, running ahead of advancing marines to burst PB xenos as much as possible, taking advantage of whoppingly high maximum burst damage potential, coupled with burst PB mechanic and so on. This behavior goes against the CO Code of Conduct. This rebalancing will greatly reduce the M46C's damage output ceiling without compromising its role as an unique rifle built to outperform the M41A Mk1 and Mk2, alongside making use of its IFF singlefire more desirable, encouraging COs to use the weapon more like they would a smartgun.
As for the M56C, while the original PR for it presented an absolutely ridiculously overpowered weapon, myself and CO council have agreed that Thwomper's sprites for the weapon are cool and that it'd be a shame to let them go to a waste because of a doomed PR. Alongside giving COs a SG drum belt, which was previously unobtainable for COs, this will hopefully contribute towards breaking the players' stigmatization of COs using smartguns over the M46C, and make it a more appealing primary weapon choice overall compared to the M46C.

## Changelog

:cl: 50RemAndCounting
add: Added the M56C, a mostly-cosmetic smartgun variant exclusive to the CO. Can be ID locked. Added a M56C equipment case with an extra drum belt to replace the standard M56B case.
balance: Slightly reduced M46C's damage per shot, burstfire shot amount and accuracy. Barrel Charger can no longer be installed on the M46C. Removed IFF mode's damage reduction.
imageadd: added M56C sprites, made by Thwomper.
/:cl:


Bram-KEP:
Not a CO, but could make the M65C slightly better then base, just not stupidly OP

c4xmaniac:
I do not like this change for a number of reasons. For one, its unnecessary. The M46C serves as a good litmus test to see if you are actually worthy of holding the CO WL, if you run forwards to PFC+++ everyone notices and you will get capped/killed frequently.

I also do not think this nerf is going to achieve the desires results you're seeking.

1. You want the M46C to stop overshadowing the SG in an attempt to remove the stigma attached to forgoing the protag rifle for another SG.
2. At the moment, the M46C and the SG offer two very different things. The M46C offers reliable damage, far better ergonomics, ease of use, and takes incen mags. The SG is bulky, takes your backpack slot, and takes your eye slot that is used for the medhud, but gives you far better situational awareness with nightvision and 1 more tile of sight.
3. This nerf removes the burst advantage of the m46c, it removes the bc, and it lowers the damage of the rifle as a whole, but also removes the IFF damage reduction. In effect, you make the m46c a mk1 with IFF. 

Now the question is not which play style do I want, it is instead do I want night vision or a medhud. Its an obviously inferior choice.

If, instead, you decide to remove the IFF damage reduction from this PR, why would you take the m46c over the SG? It becomes a question of do I take the mk1 or do I take the SG, and the game has already answered this question with a maximum of 1 SG per squad, whereas you can get large numbers of mk1s from req and multiple red alert cabinets littered around the almayer with them. 


The biggest issue for this change is that it removes the current choice, which is a **real** choice. Do I take a major damage dealer that is difficult to use without endangering my own life, or do I take a far easier to use weapon with extra situational awareness options while losing important slots like the eye and backslot. 

If you want to nerf the m46c, do so in a way that ensures it and the SG are still genuine side grades, not direct competition. As is, this nerf removes player choice and further decreases the desire to choose the SG because it turns the m46c into an SG that does more damage.

JohnDoe50974:
> The M46C serves as a good litmus test to see if you are actually worthy of holding the CO WL, if you run forwards to PFC+++ everyone notices and you will get capped/killed frequently.

I agree with this sentiment, but had this actually been happening recently there would be no consistent community backlash against the M46C and people doing this wouldn't keep their whitelists. This hasn't happened. People keep noticing and (I assume) keep reporting but nothing has happened to the offenders that I know of.

> This nerf removes the burst advantage of the m46c, it removes the bc, and it lowers the damage of the rifle as a whole, but also removes the IFF damage reduction. In effect, you make the m46c a mk1 with IFF.
This is nonsense, the M46C still does more damage per shot and burst damage than the Mk1 or Mk2. Mk1 is very limited when it comes to its allowed attachment list, which is just not true for the M46C, which has incredible modularity not only from its attachment selection but from the fact that it can accept any M41A magazine type. This point is an oversimplification.

> Now the question is not which play style do I want, it is instead do I want night vision or a medhud. Its an obviously inferior choice.
More nonsense. Picking between the smartgunner sight and the medhud isn't an either-or. I genuinely don't understand what point are you trying to make here.

> Do I take a major damage dealer that is difficult to use without endangering my own life
You shouldn't be making decisions which lead you to constantly consciously endangering your own life on the frontlines, THAT'S THE WHOLE POINT. It toes the line between acceptable and unacceptable in the CO code of conduct and was one of the main driving points behind the cancerous "deployer mains vs CIC mains" CO slapfights in the Discord. Also, major damage dealer is an understatement. There's quite nothing else in this game that can deal so much damage in so little time as the M46C, as I've elaborated in my original post. It used to be tolerated because the average CO used to be pants on head terrible at fighting, which isn't the case anymore.

> If you want to nerf the m46c, do so in a way that ensures it and the SG are still genuine side grades, not direct competition.
This makes no sense. Both primary choices should be considered equally viable picks, how is that bad?

Chainsawicus:
This change has been needed for a long time, the M46 has received half-measured changes to make sure its power only stays in the CO's hands, but that has never addressed that it has too much power in the first place. It is, by far, the strongest weapon in the game in its current state, it can kill a queen faster than an M2C can, which is a restricted, stationary weapon that has an overheat mechanic to balance out how insane its damage out is. That is insane, and it's good that it's finally being stopped.

stanalbatross:
> This change has been needed for a long time, the M46 has received half-measured changes to make sure its power only stays in the CO's hands, but that has never addressed that it has too much power in the first place. It is, by far, the strongest weapon in the game in its current state, it can kill a queen faster than an M2C can, which is a restricted, stationary weapon that has an overheat mechanic to balance out how insane its damage out is. That is insane, and it's good that it's finally being stopped.

and even the m2c itself is overpowered in many cases!

Geevies (approved):

silencer-pl:
Versatility, not outright power should be the way to go. Sort of like the COs themselves.

We've arrived at a point where players literally have meltdowns and accuse COs of griefing if they don't deploy or deploy with the SG, Which should tell something to everyone involved.

senorsapo:
if COs are being shitters its councils job,  shouldn't cuck the role out of laziness

code/modules/projectiles/guns/specialist.dm
```diff
@@ -721,6 +721,66 @@
...
+		to_chat(usr, SPAN_NOTICE("[icon2html(src)] You pick up [src], registering yourself as its owner."))
+	..()
+
+/obj/item/weapon/gun/smartgun/co/verb/toggle_lock()
+	set category = "Weapons"
+	set name = "Toggle Lock"
+	set src in usr
+
+	if(usr != linked_human)
+		to_chat(usr, SPAN_WARNING("[icon2html(src)] Action denied by [src]. Unauthorized user."))
```
stanalbatross, line 759:
```suggestion
		to_chat(usr, SPAN_WARNING("[icon2html(src)] Action denied by \the [src]. Unauthorized user."))
```

code/modules/projectiles/guns/specialist.dm
```diff
@@ -721,6 +721,66 @@
...
+		return
+
+	is_locked = !is_locked
+	to_chat(usr, SPAN_NOTICE("[icon2html(src)] You [is_locked? "lock": "unlock"] [src]."))
```
stanalbatross, line 763:
```suggestion
	to_chat(usr, SPAN_NOTICE("[icon2html(src)] You [is_locked? "lock": "unlock"] \the [src]."))
```

code/modules/projectiles/guns/specialist.dm
```diff
@@ -721,6 +721,66 @@
...
+	playsound(loc,'sound/machines/click.ogg', 25, 1)
+
+/obj/item/weapon/gun/smartgun/co/proc/name_after_co(var/mob/living/carbon/human/H, var/obj/item/weapon/gun/smartgun/co/I)
```
stanalbatross, line 766: confusing proc name

JohnDoe50974: Yell at whoever made the original proc, I copy pasted this from the M46C code.

code/modules/projectiles/guns/specialist.dm
```diff
@@ -721,6 +721,66 @@
 		if(!auto_fire)
 			STOP_PROCESSING(SSobj, src)
 
+//CO SMARTGUN
+/obj/item/weapon/gun/smartgun/co
+	name = "\improper M56C 'Command' smartgun"
```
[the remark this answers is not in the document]
JohnDoe50974: Changed.

FransFeiffer (approved):
As far as sprites are concerned, they do be fine.

Nanu308:
balance be good, lowers the power by a hefty margin which is needed

stanalbatross (approved):

PR #473, opened 2022-06-14
Fixes tooltips crashing the admin supply pod interface

### LINKED ISSUE #52: The tgui window for the "config/launch supplypod" admin command bluescreens after 5-15 seconds of being open.
## Reproduction:

Title pretty much says it all. 

5 - 15 seconds after opening the config/launch supplypod admin command window, regardless of what action (or no action) is taken, the window tgui bluescreens with the following text:

> A fatal exception has occurred at 002B:C562F1B7 in TGUI. The current application will be terminated. Send the copy of the following stack trace to an authorized Nanotrasen incident handler at https://github.com/tgstation/tgstation. Thank you for your cooperation. 
> 
>  UnhandledRejection: TypeError: Unable to get property 'x' of undefined or null reference
>    at c.fn (http://127.0.0.1:59820/tmp27264/tgui.bundle.js:1:23127)
>    at y.forceUpdate (http://127.0.0.1:59820/tmp27264/tgui.bundle.js:1:2082)
>    at Anonymous function (http://127.0.0.1:59820/tmp27264/tgui.bundle.js:1:2221)
>    at W (http://127.0.0.1:59820/tmp27264/tgui.bundle.js:1:119055)
>    at Anonymous function (http://127.0.0.1:59820/tmp27264/tgui.bundle.js:1:2189)
>    at Anonymous function (http://127.0.0.1:59820/tmp27264/tgui.bundle.js:1:28900)
>    at ie (http://127.0.0.1:59820/tmp27264/tgui.bundle.js:1:117743)
>    at Anonymous function (http://127.0.0.1:59820/tmp27264/tgui.bundle.js:1:119408)
>    at r (http://127.0.0.1:59820/tmp27264/tgui.bundle.js:1:69501)
> User Agent: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.2; WOW64; Trident/7.0; .NET4.0C; .NET4.0E)
> State: {"ckey":"silencerpl","interface":"PodLauncher","window":{"key":"[0x21023716]-main","size":null,"fancy":1,"locked":0}}
> 
> Nanotrasen (c) 2525-2562 

There is also a giant black rectangle covering the error window, see attached screenshot
image_2022-04-16_120625289

silencer-pl:
## About The Pull Request
What it says on the tin, closes #52 

Turns out, the current tgui implementation only accepts "cardinal" directions for tooltipPosition, so "top", "bottom", "left" and "right" work, but "bottom-left" for instance made it crash.

Problem fixed by removing all instances of these and making them stick to either top or bottom, depending on case.

I am somewhat annoyed at myself it took me this long to even take a look at.

## Why It's Good For The Game
Not going to lie, the supplypod launcher is an extremely versatile, customizable tool with a plethora of event uses. Want to simulate misses impacting the ship? Got it. Want to use them as a spawner for custom mobs? Yep, it does that. Want to use it to gib that one marine you don't like? Can do that too, but please don't. It sucks having to tiptoe around the "bad" buttons only to ruin all your works by one bad move, not to mention one of the broken buttons was fairly crucial to operation.

Robust tools lead to happy admins which means happy players, which means happy devs. Or something. Everyone's happy.

## Changelog

:cl:silencer_pl
fix: The admin supply pod launcher no longer crashes when it tries to display some of its tooltips
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/PodLauncher.js
+++ b/tgui/packages/tgui/interfaces/PodLauncher.js
@@ -466,7 +466,7 @@ const ReverseMenu = (props, context) => {
               Where the pods
               go after being recalled`}
             tooltipOverrideLong
-            tooltipPosition="bottom-right"
+            tooltipPosition="bottom"
             onClick={() => {
               if (data.target_mode === TARGET_MODE_DROPOFF) {
                 act("set_target_mode", {
@@ -607,7 +607,7 @@ class PresetsPage extends Component {
               color="transparent"
               icon="trash"
               tooltip="Deletes the selected preset"
-              tooltipPosition="bottom-left"
+              tooltipPosition="bottom"
               onClick={() => this.deletePreset(presetIndex)} />
           </>
         )}>
@@ -726,7 +726,7 @@ const Timing = (props, context) => {
           Reset all pod
           timings/delays`}
           tooltipOverrideLong
-          tooltipPosition="top-left"
+          tooltipPosition="top"
           onClick={() => {
             act('set_delays', {
               "drop_time": 0,
@@ -819,7 +819,7 @@ const Damage = (props, context) => {
           color="transparent"
           tooltip="Reset damage"
           tooltipOverrideLong
-          tooltipPosition="top-left"
+          tooltipPosition="top"
           onClick={() => {
             act('set_damage', {
               "damage": 0,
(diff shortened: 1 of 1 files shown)


silencer-pl:
Closes #52

fira (approved):
Well good catch, this was a known issue after the TGUI update that came with the Github move, seems it never got fixed :(( Thanks

PR #512, opened 2022-06-19
fixes moving while putting on armour just equipping it

### LINKED ISSUE #350: Fast Unstoppable Armor Equipping
## Description

Losing the held armor or moving during the equipping windup results in equipping the armor after a much shorter delay.

When equipping certain kinds of armor, there is a significant windup. Currently, moving during that windup results in the armor being equipped after a much shorter amount of time, instead of it canceling (note that there's a message that you stop putting the armor on, but that contradicts what actually happens). For example, you could be running around armor-less while holding heavy armor, and then equip it near instantly if there's any danger, bypassing the windup. This also happens if you drop or throw the armor while equipping it.

Also, throwing the armor this way at another mob would deal brute damage to the mob immediately before the armor teleports onto your body.

## Steps to reproduce

1. Hold armor in hand (it must be the type that has a windup for equipping it).
2. Start equipping it.
3. Move/drop/throw immediately after starting the equip windup.
4. Observe the armor be near-instantly equipped onto your body.

stanalbatross:
## About The Pull Request

if you move while putting on armour it will cancel rather than equipping it

honestly how was this a bug for so long

closes #350 

## Why It's Good For The Game

also some consistency with the fail messages

## Changelog

:cl:
fix: moving while putting on armour will now actually stop you putting it on, rather than just putting it on your immediately.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/internal.dm
+++ b/code/game/objects/items/storage/internal.dm
@@ -56,26 +56,24 @@
 				if("r_hand")
 					if(master_item.time_to_unequip)
 						user.visible_message(SPAN_NOTICE("[user] starts taking off \the [master_item]."))
-						spawn(0)
-							if(!do_after(user, master_item.time_to_unequip, INTERRUPT_ALL, BUSY_ICON_GENERIC))
-								to_chat(user, SPAN_NOTICE("You stop taking off \the [master_item]."))
-							else
-								user.drop_inv_item_on_ground(master_item)
-								user.put_in_r_hand(master_item)
-							return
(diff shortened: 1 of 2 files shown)


Geevies (approved):
i get the feeling an olddev put that spawn(0) in for a good reason, but i can't see why that would be in this case

fira (approved):

fira:
> i get the feeling an olddev put that spawn(0) in for a good reason, but i can't see why that would be in this case

We'll find out and fix it then ,i guess!

PR #520, opened 2022-06-19
Cryo Jelly and Latejoin RP

Geevies:
## About The Pull Request

Anyone waking up next to a cryosleep chamber will now be covered in cryo jelly, which makes them flammable. This can be washed off at the showers. Additionally, latejoin roles that start equipped will deposit all their items into their backpack, since waking up fully equipped doesn't really make sense.

## Why It's Good For The Game

This is good for lore accuracy and pre-drop RP.

## Changelog

:cl: Geeves
add: Anyone that spawns next to a cryotube will now be hungry and covered in flammable cryo jelly, this can be washed off at the showers.
add: Any equipped role that spawns next to a cryotube will now drop their equipment into their backpack after spawning.
/:cl:

### DIFF
--- a/code/game/jobs/role_authority.dm
+++ b/code/game/jobs/role_authority.dm
@@ -522,7 +522,7 @@ var/global/players_preassigned = 0
 	if(!ishuman(M))
 		return
 
-	var/mob/living/carbon/H = M
+	var/mob/living/carbon/human/H = M
 
 	H.job = J.title //TODO Why is this a mob variable at all?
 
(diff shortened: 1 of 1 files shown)


stanalbatross:
DNM until the lore team looks at it

Lumdor:
al-apone-aliens

PhantomEpicness:
suggestion:
You must shower with soap, soap has a 10 percent chance of dropping

Nanu308:
Put Ckey next to the CL at the top, so people can see who added it ingame

Nanu308:
Pending commentary by lore contributors

TopHatPenguin:
> What if the backpack doesn’t have enough space for all the gear

Backpacks can be forced to store as many items as needed, so should be fine. (Donor gear usually breaks the storage limitations of a backpack, appearing outside the ui)

Wintermote:
Why would the cryo jelly be flammable? Is that necessary? 

I think that will cause problems for new players unless it is conveyed to them appropriately. Imagine spawning in the game, you're covered in flammable jelly, you don't even know what being covered in the jelly means, you don't know where the showers even are on the ship, and you deploy groundside like this and then get set on fire.

And players already get a huge wall of text when they spawn in telling them what squad they're in that they already mostly ignore.

I would just have the jelly go away by itself after a few minutes and not have it be flammable, since it's intended for RP reasons and not for gameplay.

PR #536, opened 2022-06-20
Fixes double evolution readiness notification and sound

### LINKED ISSUE #532: Visual and Audio bug concerning Droonis.
## Description
When drone gets the notification to evo, it plays twice.

Shouldn't play twice.

## Steps to reproduce
Get enough evo points as drone.

Maciekkub:
## About The Pull Request

The evolution notification and sound will now only play once the treshold is at or above max level.

Personally I cannot find a good reason for the threshold - 1 in code but I removed the equality check. If the preferred alternative is to have it operate a >= on unmodified threshold then it is not an issue to have it changed due to the scale of the changes. 

closes #532 

## Why It's Good For The Game

You should not hear the evolve hiss twice nor have a double notification in chat about it.

## Changelog
:cl: Maciekkub
fix: Sound and notification now only play when you reach max threshold or above. 
/:cl:


Nanu308:
Put Ckey next to the CL at the top, so people can see who added it ingame

stanalbatross (approved):

PR #539, opened 2022-06-21
ravs no longer get empower stacks off the queen eye

### LINKED ISSUE #535: Queen Eye acts as an enemy for the purpose of the base ravager ability Empower.
## Description

Using Empower as a ravager scans for all enemies in close proximity and buffs the ravager based on each enemy therein. When using Empower in proximity to the queen eye, Empower counts the eye as an enemy and buffs the ravager with it. Attached below are two instances of using empower, one with the queen eye in proximity and one without-
With the queen eye:
Empower With Queen Eye Bug

Without the queen eye:
Empower No Queen Eye Empower

The wiki claims Empower counts enemies for the purpose of buffing and one would hope the queen's eye is not an enemy for the ravager using Empower.

## Steps to reproduce

1. As base ravager, move into proximity of the queen's eye.
2. Use the empower ability.
3. Notice the queen's eye is counted as an enemy for Empower.

stanalbatross:
## About The Pull Request

closes #535 

## Why It's Good For The Game

bugfix

## Changelog

:cl:
fix: Ravagers will no longer count holograms such as the queen eye as enemies for the purposes of empowering.
/:cl:

### DIFF
--- a/code/__DEFINES/typecheck/mobs_generic.dm
+++ b/code/__DEFINES/typecheck/mobs_generic.dm
@@ -4,7 +4,6 @@
 #define isRemoteControllingOrAI(M)	((M && M.client && M.client.remote_control) || (istype(M, /mob/living/silicon/ai)))
 
 #define isbrain(A) (istype(A, /mob/living/brain))
-#define ishologram(A) (istype(A, /mob/hologram))
 #define isrobot(A) (istype(A, /mob/living/silicon/robot))
 #define isanimal(A) (istype(A, /mob/living/simple_animal))
 #define iscorgi(A) (istype(A, /mob/living/simple_animal/corgi))
--- a/code/__DEFINES/typecheck/mobs_generic.dm
+++ b/code/__DEFINES/typecheck/mobs_generic.dm
@@ -4,6 +4,7 @@
 #define isRemoteControllingOrAI(M)	((M && M.client && M.client.remote_control) || (istype(M, /mob/living/silicon/ai)))
 
 #define isbrain(A) (istype(A, /mob/living/brain))
+#define ishologram(A) (istype(A, /mob/hologram))
 #define isrobot(A) (istype(A, /mob/living/silicon/robot))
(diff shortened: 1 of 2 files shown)


fira:
Could probably just have loop typecast to /mob/living for this, avoids adding ishologram() that will certainly be misused in the future

code/__DEFINES/typecheck/mobs_generic.dm
```diff
@@ -4,6 +4,7 @@
 #define isRemoteControllingOrAI(M)	((M && M.client && M.client.remote_control) || (istype(M, /mob/living/silicon/ai)))
 
 #define isbrain(A) (istype(A, /mob/living/brain))
+#define ishologram(A) (istype(A, /mob/hologram))
```
fira, line 7: Can you remove it too ? That was my point above, don't want to encourage code behavior around snowflake typechecking, it always bites us back later on

code/modules/mob/living/carbon/xenomorph/abilities/ravager/ravager_powers.dm
```diff
@@ -43,6 +43,8 @@
 	for(var/mob/M as anything in mobs_in_range)
 		if(empower_targets >= max_targets)
 			break
+		if(!istype(M, /mob/living))
+			break
```
fira, line 47: This should still be `continue` as before here not `break`, otherwise that's gonna screw up the count

IMO just replace loop above to `for(var/mob/living/M in mobs_in_range)` instead so it's done auto

fira (changes requested):
other concerns aside, this is not gonna work properly with `break`, guess typo'ed during change

code/modules/mob/living/carbon/xenomorph/abilities/ravager/ravager_powers.dm
stanalbatross: replaced

Geevies (approved):

PR #561, opened 2022-06-25
Smartgun verbs -> action buttons

stanalbatross:
## About The Pull Request

replaces SG verbs with action buttons

image

image

also moved SG code into its own file (holy shit 700 lines of sg code in the spec guns file!!!)

## Why It's Good For The Game

action buttons are cool and easy to use

verbs are shit

## Changelog

:cl: Stan_Albatross, Nauticall.
qol: Converted the smartgun-specific verbs into action buttons. Sprites by Nauticall.
qol: added helpful messages telling you how to open a smartgun's feed cover if you try to do something that is blocked by it being closed.
/:cl:


JohnDoe50974:
The issue with action buttons is they clutter the HUD, which could especially become an issue if you need to keep them on-screen because of the order abilities and such. Why not keep the verbs? Given the sheer amount of available options just right clicking the weapon and selecting the desired verb can be a much handier avenue of handling the weapon.

NexusStorm1:
I agree with John Doe as these buttons will make the HUD extremely crowded, especially since SGs are often made aSL and the order buttons make the HUD already cluttered enough. Also, I believe the removal of verbs would remove macro support? I have around 160 hours in the role and one of the most exciting parts of SG is being able to quickly switch between ammo modes depending on the xeno. Without this capability SG gameplay will become more dull and less interactive. I think the verbs should be kept and the new buttons could be a toggleable setting for people who prefer them.

stanalbatross:
> I agree with John Doe as these buttons will make the HUD extremely crowded, especially since SGs are often made aSL and the order buttons make the HUD already cluttered enough. Also, I believe the removal of verbs would remove macro support? I have around 160 hours in the role and one of the most exciting parts of SG is being able to quickly switch between ammo modes depending on the xeno. Without this capability SG gameplay will become more dull and less interactive. I think the verbs should be kept and the new buttons could be a toggleable setting for people who prefer them.

you can still use unique_action to swap ammo types

hud clutter should not be an issue as these are only 6 new icons - it won't force a second row even if you get aSL. Also, verbs cause lag and as such action buttons are always preferable. I can put the verbs back but will move them to the "weapon" tab rather than smartgun.

Shad0vvs (approved):

PR #563, opened 2022-06-25
Fixes restocking defibs/power cells cheese

stanalbatross:
## About The Pull Request

You can no longer restock defibs and power cells if they are not at full charge.

## Why It's Good For The Game

OKAY HOLD YOUR HORSES SALTY MARINE MAINS I SEE YOU EYEING THAT COMMENTS SECTION

This is 100% required in the current state of the game. Medics have it too easy, with incredible ease in resupplying, and shipside is less relevant than ever. Only through making items use their actual INTENDED mechanics to recharge (shock horror, the restocking feature was not intended to make medical vendors an infinite use recharger!) can we slow down gameplay a bit and force medics to put THOUGHT into their supplies/loadout plus cut down on medvendor cheese. 

Also it makes no sense that vendors magically recharge stuff when you put it back into them. It's just the same stuff.

## Changelog

:cl:
fix: You can no longer restock defibs and power cells if they are not at full charge.
balance: increased the rate of charge for power cells (including power cells inside other items such as defibrillators).
/:cl:

### DIFF
--- a/code/game/machinery/vending/cm_vending.dm
+++ b/code/game/machinery/vending/cm_vending.dm
@@ -954,8 +954,11 @@ IN_USE						used for vending/denying
 
 			if(istype(item_to_stock, /obj/item/device/defibrillator))
 				var/obj/item/device/defibrillator/D = item_to_stock
+				if(!D.cell)
+					to_chat(user, SPAN_WARNING("\The [item_to_stock] needs a cell in it to be restocked!"))
+					return
 				if(D.dcell.charge < D.dcell.maxcharge)
-					to_chat(user,SPAN_WARNING("\The [item_to_stock] needs to be fully charged to restock it!"))
+					to_chat(user, SPAN_WARNING("\The [item_to_stock] needs to be fully charged to restock it!"))
 					return
 
 			if(istype(item_to_stock, /obj/item/cell))
--- a/code/game/machinery/vending/cm_vending.dm
+++ b/code/game/machinery/vending/cm_vending.dm
@@ -954,7 +954,7 @@ IN_USE						used for vending/denying
 
 			if(istype(item_to_stock, /obj/item/device/defibrillator))
 				var/obj/item/device/defibrillator/D = item_to_stock
-				if(!D.cell)
+				if(!D.dcell)
 					to_chat(user, SPAN_WARNING("\The [item_to_stock] needs a cell in it to be restocked!"))
 					return
 				if(D.dcell.charge < D.dcell.maxcharge)
--- a/code/game/machinery/vending/cm_vending.dm
+++ b/code/game/machinery/vending/cm_vending.dm
@@ -952,6 +952,18 @@ IN_USE						used for vending/denying
 	for(R in (stock_listed_products))
 		if(item_to_stock.type == R[3] && !istype(item_to_stock,/obj/item/storage))
 
+			if(istype(item_to_stock, /obj/item/device/defibrillator))
+				var/obj/item/device/defibrillator/D = item_to_stock
+				if(D.dcell.charge < D.dcell.maxcharge)
+					to_chat(user,SPAN_WARNING("\The [item_to_stock] needs to be fully charged to restock it!"))
+					return
+
+			if(istype(item_to_stock, /obj/item/cell))
+				var/obj/item/cell/C = item_to_stock
+				if(C.charge < C.maxcharge)
+					to_chat(user,SPAN_WARNING("\The [item_to_stock] needs to be fully charged to restock it!"))
+					return
+
 			if(item_to_stock.loc == user) //Inside the mob's inventory
 				if(item_to_stock.flags_item & WIELDED)
 					item_to_stock.unwield(user)
--- a/code/game/machinery/vending/vending.dm
+++ b/code/game/machinery/vending/vending.dm
@@ -664,6 +664,9 @@
 
 			if(istype(item_to_stock, /obj/item/device/defibrillator))
 				var/obj/item/device/defibrillator/D = item_to_stock
+				if(!D.cell)
+					to_chat(user, SPAN_WARNING("\The [item_to_stock] needs a cell in it to be restocked!"))
+					return
 				if(D.dcell.charge < D.dcell.maxcharge)
 					to_chat(user,SPAN_WARNING("\The [item_to_stock] needs to be fully charged to restock it!"))
 					return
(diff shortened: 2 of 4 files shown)


Shad0vvs:
You're insane.

Lafrien:
So now FoB will be filled with charged defibs on floor and blinking chargers. I don't see much change, except more clunkyness with all items around the floor at FoB. But it surely was wierd to have vendor recharging them magically.

Unknownity:
> So now FoB will be filled with charged defibs on floor and blinking chargers. I don't see much change, except more clunkyness with all items around the floor at FoB. But it surely was wierd to have vendor recharging them magically.

this is mainly a nerf to power cells as they're never actually charged through normal means

Nanu308:
Look into increasing charge rate for items as well with this PR

Superwelder:
This change is a start but it is only a half-step. In my experience, defibrillator charge is almost never the limiting resource for a medic in terms of dishing out healing. By the time you actually run out of defibrillator charge, there is almost always another defib on offer. Be it from a dead medic or vending them from the vendor. I have had to recharge my defibrillator in a vendor maybe 3 times in the last 3 months while I have taken a dead medic's defib dozens of times in the same span of time. What is really the medic 'cheese' is being able to split any stack of of kits/splints and manufacture ridiculous numbers of said kits/splints instantaneously. That coupled with any marine being able to fill almost any useful injector causes a post-scarcity medical economy in CM. The seeds of this problem were sown when medvendors were made wrench-able way back when. It does not help that currently the synthetics, RTOs, survivors, and of course engineers can all hack said medvendors for anyone to use.

Wintermote:
I don't think we need to fix duping of minor healing items like splints; these are things that you can already get 20 of them per player very easily, but you'd also never want to - and if we reduce the number of splints available so that each individual one actually matters, it becomes a very serious balance change. I'm not sure it'd be a good idea.

And this is also mostly true of defibs - getting your hands on an extra defib is not usually a problem.

However, what IS a problem, is trying to recharge defibs or batteries instead of hotswapping through a vendor - nobody has the time to waste 5-20 minutes waiting for their defib or battery to recharge, so the actual effect would be disabling defib and battery replenishment entirely, not merely preventing them from being swapped through the vendors.

Additionally, there just aren't any real portable chargers for marines to use. It's already common enough to have ZERO chargers in the fob at all, let alone having to go up to the ship to stick a defib in a charger.

IMO - This needs to go a whole lot further than simply disabling swapping in vendors. Fixing this requires a redo of how batteries and charging work in their entirety.

There needs to be:
1. A way to swap batteries out of defibs and anything else with 'charges'
2. A way to recharge batteries in a timely manner (so that players are not standing around waiting for even 3+ minutes)
4. A way to keep all these batteries organized and available to players
5. A way to carry all these batteries without destroying your ability to carry other gear

I am altogether more of the opinion that charging batteries is an unnecessary "realism" headache for gameplay and we are better off without any of this stuff.

But if it really needs to be done, just changing the vendors is only going to create a huge balance problem. It needs more work than just that.

code/game/machinery/vending/cm_vending.dm
```diff
@@ -952,6 +952,18 @@ IN_USE						used for vending/denying
 	for(R in (stock_listed_products))
 		if(item_to_stock.type == R[3] && !istype(item_to_stock,/obj/item/storage))
 
+			if(istype(item_to_stock, /obj/item/device/defibrillator))
+				var/obj/item/device/defibrillator/D = item_to_stock
+				if(D.dcell.charge < D.dcell.maxcharge)
```
Geevies, line 957: future-proof this check, add a "does this defib even have a cell", because i bet someone in three years is gonna add a defib with swappable cells

code/game/machinery/vending/cm_vending.dm
```diff
@@ -952,6 +952,18 @@ IN_USE						used for vending/denying
...
+					to_chat(user,SPAN_WARNING("\The [item_to_stock] needs to be fully charged to restock it!"))
```
Geevies, line 958:
```suggestion
					to_chat(user, SPAN_WARNING("\The [item_to_stock] needs to be fully charged to restock it!"))
```

code/game/machinery/vending/cm_vending.dm
```diff
@@ -952,6 +952,18 @@ IN_USE						used for vending/denying
...
+					return
+
+			if(istype(item_to_stock, /obj/item/cell))
+				var/obj/item/cell/C = item_to_stock
+				if(C.charge < C.maxcharge)
+					to_chat(user,SPAN_WARNING("\The [item_to_stock] needs to be fully charged to restock it!"))
```
Geevies, line 964:
```suggestion
					to_chat(user, SPAN_WARNING("\The [item_to_stock] needs to be fully charged to restock it!"))
```

code/game/machinery/vending/vending.dm
```diff
@@ -661,6 +661,19 @@
 				if(W.tape)
 					to_chat(user,SPAN_WARNING("Remove the tape first!"))
 					return
+
+			if(istype(item_to_stock, /obj/item/device/defibrillator))
+				var/obj/item/device/defibrillator/D = item_to_stock
+				if(D.dcell.charge < D.dcell.maxcharge)
```
Geevies, line 667: ditto

code/game/machinery/vending/vending.dm
```diff
@@ -661,6 +661,19 @@
...
+					to_chat(user,SPAN_WARNING("\The [item_to_stock] needs to be fully charged to restock it!"))
```
Geevies, line 668:
```suggestion
					to_chat(user, SPAN_WARNING("\The [item_to_stock] needs to be fully charged to restock it!"))
```

code/game/machinery/vending/vending.dm
```diff
@@ -661,6 +661,19 @@
...
+					return
+
+			if(istype(item_to_stock, /obj/item/cell))
+				var/obj/item/cell/C = item_to_stock
+				if(C.charge < C.maxcharge)
+					to_chat(user,SPAN_WARNING("\The [item_to_stock] needs to be fully charged to restock it!"))
```
Geevies, line 674:
```suggestion
					to_chat(user, SPAN_WARNING("\The [item_to_stock] needs to be fully charged to restock it!"))
```

Geevies (changes requested):

PR #564, opened 2022-06-25
Type 23 nerf

stanalbatross:
## About The Pull Request

very very strong shotgun gets a bit less strong

## Why It's Good For The Game

this thing can nigh oneshot a human and is impossible to play against. nerf needed.

plus it's so strong marines should not get it from black market crates (which has started becoming a problem like with the pounceproof ERT armours that used to be in there)

## Changelog

:cl:
balance: removed the type 23 from the black market crate pool
balance: tweaked down the damage of all type 23 ammo (8g shells)
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/largecrate.dm
+++ b/code/game/objects/structures/crates_lockers/largecrate.dm
@@ -321,7 +321,6 @@
 					/obj/item/weapon/gun/rifle/hunting = /obj/item/ammo_magazine/rifle/hunting,
 					/obj/item/weapon/gun/smg/mp5 = /obj/item/ammo_magazine/smg/mp5,
 					/obj/item/weapon/gun/rifle/m16 = /obj/item/ammo_magazine/rifle/m16,
-					/obj/item/weapon/gun/shotgun/type23 = /obj/item/ammo_magazine/handful/shotgun/heavy/buckshot,
 					/obj/item/weapon/gun/rifle/type71 = /obj/item/ammo_magazine/rifle/type71,
 					/obj/item/weapon/gun/smg/fp9000 = /obj/item/ammo_magazine/smg/fp9000
 				)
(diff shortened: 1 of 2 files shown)


Geevies (approved):

Diegoflores31:
I dont know what to think about this . Its a weapon with very bas ROF . 4 bullets only . Impossible to get ammo . And now will underperform a normal m37??? Kinda lame nerf

c4xmaniac:
If you let the shotgunner with a shotgun that carriers 4 shells, has an insanely slow ROF, and rare ammo come close to you - you should probably get 1 shot as a marine considering the normal m37 already 1 shots your or almost 1 shots you. Completely unneeded nerf. Also I don't particularly care about it getting removed from the loot box, but you again overexaggerate how common the gun is because req needs to invest large sums of its budget into loot boxes to get a chance at the shotgun.

MusselMain:
This is the rare case where I have to say: literal skill issue, this thing already has only 4 shots, a rate of fire so slow that T3's can tank a shot from you and kill you before you fire again, and buckshot is only effective at PB, using Buckshot at anything farther then two tiles causes it to do pitiful damage, AND YOU WANT TO NERF IT??? This is a very situational weapon, that's at it's CURRENT STATE is barely viable in anything but ambushes, I understand it getting removed from Black market, but don't nerf it, you'll be fucking over the UPP and Admin Merc's that are stuck using a very 'quirky' weapon that'll now be completely worse then a standard M37 (imagine spawning with only the Type 23, and with only buckshot, that's a role in this game and they have very clear and obvious weaknesses), It's a shotgun that is already really bad against most xenos, in it's current state, so I'd recommend nerfing longer range shells slightly, but DON'T touch the close range ammo, it's incredibly situational, and can be easily avoided by actually using your brain while in combat (avoid hugging corners, travel in groups, use your range to kill the shotgunners). As I mentioned, don't nerf the PB ammo, It'll just make the UPP even worse when they encounter Xenos.

PR #565, opened 2022-06-25
allows construction in the brig

### LINKED ISSUE #552: Unable to Place Foldable Chairs in the Brig
## Testmerges:
Could be related to #416 

## Description
Can't place folding chairs anywhere in lower brig, it prevents some cool RP scenarios, but otherwise doesn't affect much.

You should be able to place chairs.

## Steps to reproduce
1. Pick up any foldable chair in the brig
2. Place down the foldable chair in the brig
 

stanalbatross:
## About The Pull Request

closes #552 

## Why It's Good For The Game

It makes no sense to dissallow construction in the brig and especially no sense to only dissallow it on some of the tiles
I don't think this was ever intended and if it was it's dumb.

## Changelog

:cl:
fix: fixed some tiles in the brig on the Almayer not allowing construction.
/:cl:

### DIFF
--- a/maps/map_files/USS_Almayer/USS_Almayer.dmm
+++ b/maps/map_files/USS_Almayer/USS_Almayer.dmm
@@ -3266,9 +3266,7 @@
 /turf/open/floor/plating/plating_catwalk,
 /area/almayer/lifeboat_pumps/north1)
 "ajj" = (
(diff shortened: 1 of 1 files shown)


Nanu308 (approved):
seems like an oversight yeah, lgtm

PR #578, opened 2022-06-27
Removes the ability to crowbar grass

### LINKED ISSUE #574: The grass Tiles from the garden on the Almayer are "growing" out of control
## Description
If you play any role that can pickup a crowbar, MT, engineer Etc. You can duplicate the grass tiles found in the memorial garden. Not only that but placing one down also causes you to not be able to remove it forever. This means if you place a grass tile underneath let's say, the Telecoms APC, you will never be able to repair it post hijack. 

Crowbaring grass tiles does not remove it but duplicates. What needs to happen is when it gets crowbarred it removes that tile

(It won't)

- [ x] Issue could be reproduced at least once
- [x ] Issue happened in a recent (less than 7 days ago) round
- [ x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Maciekkub:
## About The Pull Request

overrides the default flag_tools to null to prevent using crowbar on grass tiles, mainly affecting the almayer grass.

closes #574 

## Why It's Good For The Game

I tried having the grass tiles on the almayer be destroyed when crowbared instead, but for reasons unknown to me the grass tiles would still be present and you would get messages about removing the grass tile but the grass tile would not be ridden of. 

Like I said earlier, it would probably be preferable to have the grass tile destroyed like a normal floor tile would be, but I had no luck finding the actual reason behind the grass still being present after using crowbar to remove them. Based on that I have assumed that the grass there is not to be touched and removing the ability to crowbar it entirely would be a viable solution. 
## Changelog

:cl:
fix: You cannot crowbar grass tiles on Almayer preventing duplication. 
/:cl:

### DIFF
--- a/code/game/turfs/floor_types.dm
+++ b/code/game/turfs/floor_types.dm
@@ -417,6 +417,7 @@
 	name = "grass patch"
 	icon_state = "grass1"
 	tile_type = /obj/item/stack/tile/grass
+	tool_flags = null
 
 /turf/open/floor/grass/Initialize(mapload, ...)
 	. = ..()


MusselMain:
Grass tiles must be made of astroturf or something

stanalbatross (approved):

PR #581, opened 2022-06-28
Increases ethanol speech slurring threshold (no more raging alcoholic RP from 2 sips of whiskey)

Bram-KEP:
## About The Pull Request
Increases the amount of alcohol absorbed before speech starts slurring. triples it, to be exact. If this is too much, not opposed to only making it double.

## Why It's Good For The Game
Current cap is extremely low, nobody speaks like a drunkard after 2 sips of whiskey irl. This increases it, allowing for more RP, say the CL giving a survivor a drink without the survivors speech becoming incomprehensible. 

## Changelog

:cl: Rem1
balance: The alcohol threshold before speech slurring starts has been tripled. 
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry_reagents/alcohol.dm
+++ b/code/modules/reagents/chemistry_reagents/alcohol.dm
@@ -26,7 +26,7 @@
 	var/adj_sleepy = 0
 	var/slurr_adj = 3
 	var/confused_adj = 2
-	var/slur_start = 300 //(formerly 180) amount absorbed after which mob starts slurring
+	var/slur_start = 900 //(formerly 180) amount absorbed after which mob starts slurring
 
 /datum/reagent/ethanol/on_mob_life(mob/living/M, alien)
 	//This is all way too snowflake to accurately transition to the property system so it stays here


Bram-KEP:
oh my god this costs GBP

PhantomEpicness:
Through my testing of the alcohol system, one beer can get you into the thousands given enough time. The system needs a rework in general.

MusselMain:
This is needed, my first time drinking I had one can of beer and was absolutely smashed the rest of the op, this was a issue as I was a PO who went into crit in the cockpit from that one beer.

Nanu308:
This would be a bandaid on top of an bleeding issue with the current system, preferably this should really be a refactor rather than just a number change. Take the offer of Epicness and check out his code for it would be my recommendation.

Bram-KEP:
> This would be a bandaid on top of an bleeding issue with the current system, preferably this should really be a refactor rather than just a number change. Take the offer of Epicness and check out his code for it would be my recommendation. 

See if I knew how to code I would, sadly I don't lol. I talked with epicness in DMs and he said the refractor was nearly done and he planned to possibly work on it again. So hopefully he will be able to PR it. Until then a bandaid fix is probably good?

Wintermote:
"Good enough" fixes tend to stick around for months/years, especially if the problem is not really apparent to the players afterwards.

Not saying it shouldn't be done but I could easily see a refactor getting forgotten and nobody looks at the alcohol code for years because there doesn't seem to be any more problems in the gameplay anymore.

Nanu308:
Closed for now, as stated "Good enough fixes" is not the way to go about it when it's not a critical issue.

PR #582, opened 2022-06-28
Xeno Deathcheck

Geevies:
## About The Pull Request

Xenos now check for death on updatehealth, meaning they die when you shoot them.

## Why It's Good For The Game

Delayed deaths on a laggy server isn't very great.

## Changelog

:cl:
code: Xenos now check for death on updatehealth, meaning they die when you shoot them.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -148,24 +148,7 @@
 
 	updatehealth()
 
-	if(health <= crit_health - warding_aura * 20) //dead
-		if(prob(gib_chance + 0.5*(crit_health - health)))
-			INVOKE_ASYNC(src, .proc/gib, last_damage_data)
-		else
-			death(last_damage_data)
-		return
-
-	else if(health <= 0) //in crit
-		if(hardcore)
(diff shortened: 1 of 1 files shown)


senorsapo:
this has been pending for decades thank you

stanalbatross (approved):

Zangeiti:
SPEEDMERGE.. NOW!!!

JeserTheFox (approved):

Superwelder:
After having played with this in a few rounds, it is extremely clunky. It does not feel natural at all to just freeze when seeing a fellow xeno or yourself reach critical health. It feels like being pushed into a trapper boiler trap. You just sit there for multiple ticks unable to move or act then fall down. Once, when it happened to me as lurker, I actually froze mid-lunge when I got hit with a rifle round to crit so I did not finish the leap which makes zero sense as momentum would carry an object in motion and not stop mid-air. No the system was not perfect before but it felt much more natural and appropriate for the actions you take before death.

Lumdor:
This should have been tested. An extreme nerf to the Xenos because you freeze in place when you crit. No more grace period of a few seconds from when you hit crit to save yourself. The sprites also freeze standing up in place when you do go crit and make the whole experience clunky as the person above stated.

PR #586, opened 2022-06-29
Adds consequences of being above maximum blood level

### LINKED ISSUE #364: Infinite Blood
## Description

You can have a blood level over 100%, with no apparent limit on it.

You shouldn't be able to be a walking blood bank.

## Steps to reproduce

1. Acquire fork/spoon/knife.
2. Acquire ridiculous amounts of food.
3. Use fork/spoon/knife to force-feed yourself all the food.
4. Use health analyzer/scanner to see that your blood level keeps increasing past 100%.

Note: Use of a fork/spoon/knife can be substituted by extreme toileting sessions (to generate hunger).

Maciekkub:
## About The Pull Request

- If our blood level is above the high threshold, then we will start losing blood slowly.
- When the blood level is above the maximum threshold, vision becomes blurry and we take temporary oxygen damage
- Health scanner will now show if the blood level is above high threshold and the advice to transfuse the blood away will be shown.
- Changed the code for health anazyler blood-related popups to match their corresponding levels instead of magic numbers.
- Added new threshold for maximum blood of around 120% of standard blood levels and replaced the existing references to a high blood level.

Closes #364 

## Why It's Good For The Game

You should not be able to manipulate your blood levels to counter bloodloss by stacking up on blood before deployment.
I found it extremely difficult to reach the above safe levels of blood and believe that unless done on purpose it should not happen regularly. 
Adds realism, closes issue.

## Changelog

:cl: Maciekkub
add: Upon blood levels exceeding 120%, vision will now become blurry and the player will take temporary oxygen damage, similar to having low blood. Player will also be notified about feeling dizzy.
add: Health analyzer will now display if the blood level is too high and the advised course of action. 
code: Changed health analyzer code from magic numbers to already defined thresholds. 
code: added new threshold of 120% and changed existing references to the old MAXIMUM to new HIGH
/:cl:


JohnDoe50974:
This is too harsh for something that is too easy to mess up, -1

Maciekkub:
Both of you are quite right about that, the level needs to be way higher. Currently as things stand everyone will be blind and take damage at round start due to nutrition levels starting low and blood starting at 100%. I'll adjust the levels.

Nanu308:
Headdev here

Look into having two different set levels if its attainable similar to how we are doing current OD thresholds. For example everything above 103/5% or such numbers give the first set of "debuffs" while the second "Extreme overdose" gives a lot more severe issues.

I think this would be a good solution, if it does not impact performance etc. in an adverse way.

Maciekkub:
> Headdev here
> 
> Look into having two different set levels if its attainable similar to how we are doing current OD thresholds. For example everything above 103/5% or such numbers give the first set of "debuffs" while the second "Extreme overdose" gives a lot more severe issues.
> 
> I think this would be a good solution, if it does not impact performance etc. in an adverse way.

Hello Nanu,

As per your suggestions I added another level of "blood overdose", somewhere in between of HIGH and MAXIMUM. The level % at which the first effects start is 110% and the debuffs are minor. I agree with you that even as high as 105% blood level should have an effect on us, however like has already been mentioned in this thread, a marine getting out of cryo and eating will easily attain around 107%, which would basically cause the entire marine force to sit in medbay for blood tranfusions and delay deployment - in addition to being quite annoying. Now that I think about it, it could be done that the blood would go down on a lower level, so the effects would be present, but they would disappear over time so that there is no need to queue up at medbay.  I however am quite satisfied with a compromise of 110%.

As for performance, I do not think that it is going to be impacted in a negative way, due to the fact blood could have been stacked up anyway, nonetheless of these changes. However just to make sure, I moved the checks under the switch statement.

Weuyn:
Could you include the HGN research property in this. ATM its very very lethal if its over potency 3 and the marine its inside has full blood.

Maciekkub:
> Could you include the HGN research property in this. ATM its very very lethal if its over potency 3 and the marine its inside has full blood.

Hello Weuyn,

This PR does not affect chemicals, their potency and interactions with blood.
No idea what HGN is either way.

Wintermote:
I really don't think this is needed

What is it fixing - ??

Players aren't, and largely can't, go out of their way to maximize their blood volume, and even if you're sitting on 120% blood the effect on gameplay flow is basically non-existent.

This is a solution in search of a problem imo

Geevies:
seems fine, i would just suggest preventing mundane things every person does, like eating food, from generating blood over the "damage" threshold. look into capping the amount that nutriment provides, for instance.

stanalbatross:
> i can imagine doctors like : " dont worry marine i will help you " and then cutting you 20 times with his knife to take the excess blood lol.

this is soul

also @Maciekkub join the cm discord so you can get ping spammed about reviews

Maciekkub:
> seems fine, i would just suggest preventing mundane things every person does, like eating food, from generating blood over the "damage" threshold. look into capping the amount that nutriment provides, for instance.

My problem here is I have gone through the code searching by "nutri" or "blood" and I was not able to find the proc which handles the blood addition. Searching by "reagent" yields around 15k results or so. Is there an easier way to find what I am looking for here?

Geevies:
> > seems fine, i would just suggest preventing mundane things every person does, like eating food, from generating blood over the "damage" threshold. look into capping the amount that nutriment provides, for instance.
> 
> My problem here is I have gone through the code searching by "nutri" or "blood" and I was not able to find the proc which handles the blood addition. Searching by "reagent" yields around 15k results or so. Is there an easier way to find what I am looking for here?

looks like it's cuz nutriment has this HEMOGENIC property. i'd look into tweaking this then, maybe make it so that if the hemogenic value is only 1, then it wont go over a certain threshold, but stronger concoctions can
image

Geevies:
> > looks like it's cuz nutriment has this HEMOGENIC property. i'd look into tweaking this then, maybe make it so that if the hemogenic value is only 1, then it wont go over a certain threshold, but stronger concoctions can image
> 
> I had a look and fiddled around with it and I believe the way the chem properties are handled is by calling their respective process x times depending on their assigment in the list which is defined at the chemical level. If we have got Iron with its PROPERTY_HEMOGENIC = 3 in a list, it will in fact just invoke the hemogenic proc three times.
> 
> For instance from the above snippet of code you provided, I believe the if statement is never executed, since potency is always 1 as set in the function's elements. I am not sure how DM supports default assignment but I have found that potency is always 1 during my testing.
> 
> I tried having a hemogenic local variable to be able to manipulate the "potency" of the chemical and to bypass the limit for weak chemicals, but it did not seem to care whatsoever about it.
> 
> It might be the case that PROPERTY_HEMOGENIC might be defined as maybe PROPERTY_HEMOGENIC_WEAK and then HIGH respectively to check for limit bypass, but I do not know how the chemicals are able to call the hemogenic proc by just having their name in the "properties" list at their definition level. I tried looking for properties' definition but there were no results.
> 
> In short, I do not quite know how to handle this.

gotcha, i'll do some testing of my own when i have time, which is probably gonna be this saturday

Don2137:
This is an extremely harsh and punishing solution, affecting an unclear and hard to accurately control system, trying to solve an issue that doesn't exist. Nobody out there is BLOODMAXXXXING predeployment, eating iron up to OD is only enough for like 8 minutes, and you need to get that from chemistry in the first place.

A better way would be to tweak the OD values of the various hemogenic reagents but still, this is not an issue that exists or has ever existed.

Joelampost:
This will make pred even funnier to play. with thwei giving preds blood until it runs out with a hard cap of 125% blood level this will make preds unable to heal at all due to the oxy damage being infinitely stacked due to thewi giving blood at a level of hemogenic 9.

Weuyn:
HMG is what I meant, not HGN, oops. Yeah the hemogenetic one Geeves made a screenshot. It murders the fuck out of you if your blood goes over the threshold.

Maybe fix forks from allowing you to nutrientmaxx on food? Thats how the infinite blood is being generated, right?

There just needs to be the same line of code that says 'you don't feel like eating right now' applied to forcefeeding and forks.

PR #596, opened 2022-07-01
[Bugfix] Fix core cooldown on hijack launch

TeDGamer:
## About The Pull Request
Simply a core can be rebuilt once hijack starts. This fixes the issue of having to wait 5 minutes into hijack to rebuild the core.

You can't relocate core after 20 minutes into the game so you need to destroy it but then wait 5 minutes. So this PR makes it so that on launch, AFTER the special structures are destroyed can you rebuild the hive core

## Why It's Good For The Game
Build pool early, get larva surge early.

Also it's a bugfix.

## Changelog

:cl: TeDGamer
fix: Hive core cooldown is reset on launching the dropship after all special structures are destroyed.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
+++ b/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
@@ -737,10 +737,6 @@
 		last_locked = world.time
 		if(almayer_orbital_cannon)
 			almayer_orbital_cannon.is_disabled = TRUE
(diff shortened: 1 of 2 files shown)


PhantomEpicness:
My solution was reseting it when it disabled OB but I this is better.

stanalbatross (approved):

JeserTheFox (approved):

PR #598, opened 2022-07-01
[Code] Human status check after health updates

TeDGamer:
## About The Pull Request
Keeps this consistent with the [xeno deathcheck](582). Also it seems like originally updatehealth was in the handle_status_updates and happened before the rest of the code so it basically goes back to that.

recalculate_move_delay is set to true in updatehealth so if we change the order, setting  that in handle_status_updates is no longer neccessary. 

## Why It's Good For The Game
Consistency for code's sake and people will die when they should die, not based off of the health from the previous life cycle

## Changelog

:cl: TeDGamer
code: Humans handle status updates (such as death) after their health is updated in current life cycle
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -76,10 +76,10 @@
 	//Handle temperature/pressure differences between body and environment
 	handle_environment() //Optimized a good bit.
 
+	updatehealth()
+
 	//Status updates, death etc.
 	handle_regular_status_updates() //Optimized a bit
-
-	updatehealth()
 
 	update_canmove()
 
--- a/code/modules/mob/living/carbon/human/life/handle_regular_status_updates.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_regular_status_updates.dm
@@ -9,10 +9,6 @@
 		blinded = 1
 		silent = 0
 	else //ALIVE. LIGHTS ARE ON
-		//updatehealth() // moved to Life()
-
-		recalculate_move_delay = TRUE
-
 		if(health <= HEALTH_THRESHOLD_DEAD || (species.has_organ["brain"] && !has_brain()))
 			death(last_damage_data)
 			blinded = 1


hotimep:
rip the comedic one-liner window when you're about to die

Superwelder:
Perhaps someone can explain how this will work to me because I am worried this merge will force so, so many marine deaths to be at exactly 200 damage total from xenos and friendly fire. If you no longer have that delay in death, short though it may be, won't the dying human stop taking damage instantly from say, multiple xenos slashing or those extra bullets from a burst fire if you die to the first and fall to the ground? At first glance that may not seem like a problem but it makes the medic's job a hell of a lot easier if you can just assume everyone that dies can instantly be defibbed alive not counting something extreme like an OB, CAS, or fire. Reviving someone at 270 brute damage is a lot more nuanced than 200.

TeDGamer:
> Perhaps someone can explain how this will work to me because I am worried this merge will force so, so many marine deaths to be at exactly 200 damage total from xenos and friendly fire. If you no longer have that delay in death, short though it may be, won't the dying human stop taking damage instantly from say, multiple xenos slashing or those extra bullets from a burst fire if you die to the first and fall to the ground? At first glance that may not seem like a problem but it makes the medic's job a hell of a lot easier if you can just assume everyone that dies can instantly be defibbed alive not counting something extreme like an OB, CAS, or fire. Reviving someone at 270 brute damage is a lot more nuanced than 200.

So basically it used the code used to do this.

Handle status update ( death and all that shiz) -> Update your health

What I've done is 

Update your health -> Handle status update

This was originally how it was years back but it seems like the order got jumbled along the way. This basically makes it so you use  the health from the current life cycle ( I believe these are called in BYOND 2 seconds time ) instead of using the health from the previous life cycle.

Previously: Life Cycle 4444 -> Handle status update -> _**Update health -> Life Cycle 4445 -> Handle Status Update**_
So you would be using life cycle 4444's health for the status update in 4445. I just simply made it so it's like this

Now: Life Cycle 4444 -> _**Update health -> Handle Status Update**_
Granted this does kill marines faster by ~2 BYOND seconds so you would lose out on some slash damage

Wintermote:
I think it's altogether a very good change. This is making cause->effect more consistent for players.

PR #600, opened 2022-07-01
Balance: synthskin arms no longer malf below a 10 combined damage threshold

JohnDoe50974:
## About The Pull Request
Essentially a rehearse of this Gitlab merge request, but with a more sensible threshold value, and applies to synthetics only: https://gitlab.com/cmdevs/colonial-warfare/-/merge_requests/2102
If the combined brute/burn damage on a synthskin arm or hand is at or below 10 points, the malfunction proc will not trigger.

## Why It's Good For The Game
Prevents situations where punching a synth once or throwing a piece of paper at one's arm causes their arms to malf and drop held items until smartpack charge/welder/cable coil was expended to repair it, due to synths lacking the natural hp regeneration that human mobs have, which is an unreasonably expensive process both time and resources-wise for something so comparatively meaningless. A marine wearing light armor receives infinitesimal damage from another human's punch, and said damage amount heals itself almost immediately, which leads to comical situations where synths appear more fragile than your average human. More importantly, this balance change does not affect regular prosthetic limbs' balance, which was what Nanu requested to be done in the previous MR after closing it.

In short, it will make synthetic gameplay less frustrating and greatly reduce fussing over, say 0.534568 points of brute damage causing a synth to drop what they were holding, which feels completely nonsensical and unrealistic.

## Changelog

:cl: 50RemAndCounting
balance: synthskin arms and hands will no longer malfunction and drop items if the combined brute and burn damage on them does not exceed 10 points.
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -1131,12 +1131,10 @@ treat_grafted var tells it to apply to grafted but unsalved wounds, for burn kit
 	return ((status & LIMB_BROKEN) && !(status & LIMB_SPLINTED))
 
 /obj/limb/proc/is_malfunctioning()
-{
 	if(status & LIMB_ROBOT)
 		return ((status & LIMB_ROBOT) && prob(brute_dam + burn_dam))
 	else if(status & LIMB_SYNTHSKIN && (brute_dam + burn_dam) > 10)
 		return (status & LIMB_SYNTHSKIN) && (prob(brute_dam + burn_dam))
-}
 
 //for arms and hands
 /obj/limb/proc/process_grasp(var/obj/item/c_hand, var/hand_name)
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -1132,9 +1132,9 @@ treat_grafted var tells it to apply to grafted but unsalved wounds, for burn kit
 
 /obj/limb/proc/is_malfunctioning()
 	if(status & LIMB_ROBOT)
-		return ((status & LIMB_ROBOT) && prob(brute_dam + burn_dam))
+		return prob(brute_dam + burn_dam)
 	else if(status & LIMB_SYNTHSKIN && (brute_dam + burn_dam) > 10)
-		return (status & LIMB_SYNTHSKIN) && (prob(brute_dam + burn_dam))
+		return prob(brute_dam + burn_dam)
 
 //for arms and hands
 /obj/limb/proc/process_grasp(var/obj/item/c_hand, var/hand_name)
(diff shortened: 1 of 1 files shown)


code/modules/organs/limbs.dm
```diff
@@ -1131,7 +1131,12 @@ treat_grafted var tells it to apply to grafted but unsalved wounds, for burn kit
 	return ((status & LIMB_BROKEN) && !(status & LIMB_SPLINTED))
 
 /obj/limb/proc/is_malfunctioning()
-	return ((status & (LIMB_ROBOT|LIMB_SYNTHSKIN)) && prob(brute_dam + burn_dam))
+{
+	if(status & LIMB_ROBOT)
+		return ((status & LIMB_ROBOT) && prob(brute_dam + burn_dam))
+	else if(status & LIMB_SYNTHSKIN && (brute_dam + burn_dam) > 10)
+		return (status & LIMB_SYNTHSKIN) && (prob(brute_dam + burn_dam))
+}
```
stanalbatross, line 1139:
```suggestion
	if(status & LIMB_ROBOT)
		return ((status & LIMB_ROBOT) && prob(brute_dam + burn_dam))
	else if(status & LIMB_SYNTHSKIN && (brute_dam + burn_dam) > 10)
		return (status & LIMB_SYNTHSKIN) && (prob(brute_dam + burn_dam))
```
also I don't think you need this **((status & LIMB_ROBOT)**

stanalbatross (changes requested):

code/modules/organs/limbs.dm
JohnDoe50974: How is the game going to know to trigger the malf proc on LIMB_ROBOT limbs? I'd rather not touch it as that part was the original proc.

stanalbatross: kk understandable

code/modules/organs/limbs.dm
```diff
@@ -1131,7 +1131,10 @@ treat_grafted var tells it to apply to grafted but unsalved wounds, for burn kit
 	return ((status & LIMB_BROKEN) && !(status & LIMB_SPLINTED))
 
 /obj/limb/proc/is_malfunctioning()
-	return ((status & (LIMB_ROBOT|LIMB_SYNTHSKIN)) && prob(brute_dam + burn_dam))
+	if(status & LIMB_ROBOT)
+		return ((status & LIMB_ROBOT) && prob(brute_dam + burn_dam))
+	else if(status & LIMB_SYNTHSKIN && (brute_dam + burn_dam) > 10)
+		return (status & LIMB_SYNTHSKIN) && (prob(brute_dam + burn_dam))
```
stanalbatross, line 1137:
```suggestion
	if(status & LIMB_ROBOT)
		return prob(brute_dam + burn_dam)
	else if(status & LIMB_SYNTHSKIN && (brute_dam + burn_dam) > 10)
		return prob(brute_dam + burn_dam)
```

stanalbatross (changes requested):

stanalbatross (approved):

PR #604, opened 2022-07-02
Hijack anti-delay : scouts and snipers.

stanalbatross:
## About The Pull Request

stops scouts and snipers from delaying the round end with cloak

## Why It's Good For The Game

stops delay

## Changelog

:cl:
balance: if a scout or a sniper is the last marine alive during hijack, their cloaking abilities will no longer function and they will be forcibly decloaked. 
/:cl:

### DIFF
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -869,6 +869,7 @@ var/list/squad_colors_chat = list(rgb(230,125,125), rgb(255,230,80), rgb(255,150
 	var/camouflage_break = 5 SECONDS
 	var/camouflage_enter_delay = 4 SECONDS
 	var/aiming_time = 1.25 SECONDS
+	var/can_camo = TRUE
 
 	var/aimed_shot_cooldown
 	var/aimed_shot_cooldown_delay = 2.5 SECONDS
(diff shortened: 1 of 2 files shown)


morrowwolf:
When do we get this for lurkers?

Bram-KEP:
Was going to say the same thing, this should apply to lurkers as well

stanalbatross:
> When do we get this for lurkers?

when you code it

BKryth:
LRP, no soul, plus where the balance tag 🤨

JohnDoe50974:
Cool now help Nauticall complete their lifeboats PR so hijack becomes completely undelayable

MusselMain:
> > When do we get this for lurkers?
> 
> when you code it

Yeah, if it's coded in, it should be some sort of something like the Sniper/Scout's Cloaking battery has died, or in the case of the lurker, the lurker is too weak from the lack of a hive to utilize their cloaking abilities.

PhantomEpicness:
Just make it end the round automatically like venting and save the trouble

Diegoflores31:
literally just make the round end instead ... also you should do it for both sides if its a PR about anti delay lol

SexmasterWasTaken:
> > When do we get this for lurkers?
> 
> when you code it

If you're not gonna do it for both sides, don't do it at all. This is quite frankly a bad response to a genuine question.

Wintermote:
> > > When do we get this for lurkers?
> > 
> > 
> > when you code it
> 
> If you're not gonna do it for both sides, don't do it at all. This is quite frankly a bad response to a genuine question.

This is a balance consideration, applying a balance change to 1 side and not the other is not good. I have to agree, either apply it equally or don't apply it at all. If nobody is going to take the time to do it for the other team it shouldn't be done. 

Because what happens with changes like this is 6 months later someone finally does it for xenos and someone says, "They didn't do it when it was first implemented so it obviously isn't intended and we shouldn't do it" and the change gets reverted

Geevies (changes requested):

Geevies (approved):

JeserTheFox (approved):

PR #617, opened 2022-07-04
Check for Simple mob

### LINKED ISSUE #365: Autoinjector Bricking
## Description

Using an autoinjector on a mouse results in the injector being unrefillable (i.e. "bricked") due to being "already loaded".

Examining a single-use autoinjector after using it on a mouse yields: "It is empty" and "It is currently loaded". You can't use the injector because it's empty, but you can't refill it because some part of the code still thinks it's "full". This shouldn't happen.

This also applies to multi-use injectors.

Note that you can still restock the glitched-out injector into a vendor and get a new working one.

## Steps to reproduce

1. Get mouse.
2. Notice mouse is in urgent need of medical attention.
3. Inject mouse with first-aid autoinjector.
4. Try and fail to refill the injector at a vendor.
5. Enjoy your novelty Schrodinger's Injector!

BKryth:
## About The Pull Request

Prevents people from injecting simple mobs and breaking them

## Why It's Good For The Game

closes #365 

I'm not fixing allowing injectors to work on animals because that is going to be a bitch thing to deal with, if anyone wants to add that feature down the line I put a comment on to removing the type check to allow them to inject them

## Changelog

:cl:
fix: prevents injecting animals and breaking injectors
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/hypospray.dm
+++ b/code/game/objects/items/reagent_containers/hypospray.dm
@@ -170,6 +170,13 @@
 
 	if(!istype(M))
 		return
+
+	// Check if animal mob, would break the injector childs if used on animals
+	// Remove this if you want to make chems work on simple mobs like rats and crabs, they break injectors if allowed to
+	// Fix that first if you want to remove this tldr; if you want to make rats injectable for rp, implement feature first :)
+	if(isanimal(M))
+		return
+
 	if(!M.can_inject(user, TRUE))
 		return
 
--- a/code/game/objects/items/reagent_containers/hypospray.dm
+++ b/code/game/objects/items/reagent_containers/hypospray.dm
@@ -171,12 +171,6 @@
 	if(!istype(M))
 		return
 
-	// Check if animal mob, would break the injector childs if used on animals
-	// Remove this if you want to make chems work on simple mobs like rats and crabs, they break injectors if allowed to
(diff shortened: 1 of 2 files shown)


code/game/objects/items/reagent_containers/hypospray.dm
```diff
@@ -170,6 +170,13 @@
 
 	if(!istype(M))
 		return
+
+	// Check if animal mob, would break the injector childs if used on animals
+	// Remove this if you want to make chems work on simple mobs like rats and crabs, they break injectors if allowed to
+	// Fix that first if you want to remove this tldr; if you want to make rats injectable for rp, implement feature first :)
+	if(isanimal(M))
+		return
```
stanalbatross, line 178: handle this by overriding the can_inject proc on animals, and make it give an error message too

stanalbatross (changes requested):

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -368,3 +368,8 @@
 	message = capitalize(trim_left(message))
 
 	..(message, null, verb, nolog = !ckey)	//if the animal has a ckey then it will log the message
+
+/mob/living/simple_animal/can_inject(var/mob/user, var/error_msg)
+	if(error_msg)
```
stanalbatross, line 373:
```suggestion
	if(user && error_msg)
```

stanalbatross (approved):

PR #622, opened 2022-07-04
Xeno Crit Grace Period + Fix

Geevies:
## About The Pull Request

Gives xenos a grace period of 1 second after reaching crit health before falling into crit. Also fixes them getting rooted in place if they fall into crit.

## Why It's Good For The Game

Was too punishing going from a variable grace period caused by Life ticks to absolutely no grace period at all, plus the buggy crit root thingy.

## Changelog

:cl:
add: Xenos now have a one second grace period before falling into crit.
add: The sound environment system now affects all heard sounds, so open spaces will echo, so will interiors. Xenos in their crit grace period now have a Call of Duty near death soundscape.
/:cl:


Wintermote:
So this basically just undoes the fix applied earlier?

I don't think this is right, it's returning us to unintended behaviour - being able to fight/run away while actually at negative hp and in crit. The whole point of entering crit is that you don't get to run away.

Your "grace period" is not fighting at low hp and going to negative in the first place.

JackieEstegado:
If you are gonna give an extra second of mobility to xenos when going crit then you should atleast restrict it by not allowing ability usage during that time, so runners and lurkers don't just pounce off-screen while crit. Maybe even add a small slowdown.

AGasCanister:
give them a funny sound when in their grace period!!
https://www.youtube.com/watch?v=DcgwjseQldM

c4xmaniac:
A second is a pretty big grace period especially with byond ticks on high pop, it’ll be more like 2-3. I’m not sure why it was fixed and now it’s being unfixed, if you go into crit you should be punished for it, not given an extra amount of time to just leap away to safety.

DesertFox2:
I'm against this, this just returns us to having ballsy runners, lurkers, and other xenos who can simply just dash offscreen for what is effectively 2-3 beyond seconds. Why are we undoing the previous fix?

stanalbatross:
The previous change showed how important the randomness of lifeticks was to xeno durability, this merely adds back a less random, more consistent version of it to undo the nerf to xeno durability in that PR. Xeno durability is going to get buffed back up to the prior level either way

Wintermote:
I don't understand what is random about "if you hit negative hp you go into crit immediately"

This change doesn't take away randomness - it introduces it with byond tick lag. The exact opposite of what you're arguing.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -453,26 +453,39 @@ updatehealth()
...
 
+/mob/living/carbon/Xenomorph/proc/handle_crit()
+	if(stat == DEAD || gibbing)
+		return
+
+	sound_environment_override = SOUND_ENVIRONMENT_NONE
+	stat = UNCONSCIOUS
+	blinded = 1
+	see_in_dark = 5
+	if(isXenoRunner(src) && layer != initial(layer)) //Unhide
```
stanalbatross, line 483: do for larvas as well since they can hide

stanalbatross (changes requested):
make warding pheros buff the length of the period (cool idea)

stanalbatross:
document sound stuff in cl too

RadiationFan:
Xeno survivability is pretty important to game balance, especially when talking about t3s, who are the primary beneficiaries of delayed deaths. I like the idea of warding giving significantly higher grace periods. Perhaps xenos will remember to turn on warding now.

Nanu308:
This by itself is fine, timers will need adjustment especially with warding as byond ticks/lags can greatly change the intended timer of crit to go into effect.

stanalbatross (approved):

PR #631, opened 2022-07-05
[Balance] Ignited carbon humans take variable damage based on fire reagent now

TeDGamer:
## About The Pull Request
Changes the flat magic number of 5 damage per second from fire to depend on the fire type.

Previously:
Carbon humans took a flat 5 burn damage while they were on fire no matter the reagent

Now:
Carbon humans take damage based on the fire reagent:
Welder Fire ( mini-flamer): 2 dps ( much much less dangerous, feel free to use this more )
Napalm B-Gel (Green fire):  3 dps ( much less dangerous now )
UT-Napthal Fuel (Orange fire): 6 dps ( slightly more dangerous now)
Napalm X (Blue fire): 8 dps ( more dangerous now!!! )
Incidendary OB ( white fire): 14 dps (MUCH MUCH MORE DANGEROUS NOW. DO NOT WALK IN)

I still use the magic number 5 as a divider of damage due to how otherwise flames would be too dangerous for carbon humans
(5 was the only number that would nicely divide into all the burn levels which are seperated by 5 ).

## Why It's Good For The Game
Rejoice marines, you can run into green fire, pull a burnt marine without suffering as much damage as you would previously.

Also, I hate magic numbers considering we have an actual working flamer reagent system in the game and this just hasn't been updated. Flamer system doesn't really get a lot of love anymore.

Plus buffs pyros for HvH when using better flame reagents.

## Changelog

:cl: TeDGamer
balance: Fire damage dealt to carbon humans on fire is based on the fire reagent.
/:cl:

### DIFF
--- a/code/__DEFINES/weapon_stats.dm
+++ b/code/__DEFINES/weapon_stats.dm
@@ -317,3 +317,5 @@ Fire Variant = Markers for special fire types that behave outside of chemfire co
 #define FIRE_VARIANT_DEFAULT 0
 ///"Type B" Armor Shredding Greenfire: Burn Time T5, Burn Level T2, Slows on Tile, Increased Tile Damage, Easier Extinguishing.
 #define FIRE_VARIANT_TYPE_B 1
+// Lowers burn damage to humans
+#define HUMAN_BURN_DIVIDER 5
--- a/code/modules/mob/living/carbon/human/life/handle_fire.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_fire.dm
@@ -12,5 +12,5 @@
 	if((1 - thermal_protection) > 0.0001)
 		bodytemperature += BODYTEMP_HEATING_MAX
 		recalculate_move_delay = TRUE
-		var/dmg = armor_damage_reduction(GLOB.marine_fire, fire_reagent.burn_level / HUMAN_BURN_DIVIDER)
+		var/dmg = armor_damage_reduction(GLOB.marine_fire, fire_reagent.intensityfire / HUMAN_BURN_DIVIDER)
 		apply_damage(dmg, BURN)
--- a/code/modules/mob/living/carbon/human/life/handle_fire.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_fire.dm
@@ -12,5 +12,5 @@
 	if((1 - thermal_protection) > 0.0001)
 		bodytemperature += BODYTEMP_HEATING_MAX
 		recalculate_move_delay = TRUE
-		var/dmg = armor_damage_reduction(GLOB.marine_fire, R.burn_level / HUMAN_BURN_DIVIDER)
+		var/dmg = armor_damage_reduction(GLOB.marine_fire, fire_reagent.burn_level / HUMAN_BURN_DIVIDER)
 		apply_damage(dmg, BURN)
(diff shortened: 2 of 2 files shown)


Wintermote:
Green fire was already so weak that it wasn't anything to worry about running into it. I don't think that needed any changes downwards. 

However, regular fire and everything above that already did a LOT of damage to marines, and now this is making it worse. It is incredibly easy for marines to stack so much fire damage they need 2-3 medics to revive them. OB's in particular already do this to marines, and this is changing it to make the damage stack higher and faster.

This is not a good change imo; fire is highly lethal to marines already with the exception of green fire, and marines lack a quick and effective way of treating burn damage.

In general fire's issue right now is that it does a LOT of damage to marines, and hardly anything to xenos. It's bad enough that xenos regularly run into/through fire and purposefully drag marines into it, now this is making that tactic even more effective - when fire is meant to be one of the direct counters to xenomorphs in the lore.

This is just going to cause more issues than it solves imo.

TeDGamer:
> Green fire was already so weak that it wasn't anything to worry about running into it. I don't think that needed any changes downwards.
> 
> However, regular fire and everything above that already did a LOT of damage to marines, and now this is making it worse. It is incredibly easy for marines to stack so much fire damage they need 2-3 medics to revive them. OB's in particular already do this to marines, and this is changing it to make the damage stack higher and faster.
> 
> This is not a good change imo; fire is highly lethal to marines already with the exception of green fire, and marines lack a quick and effective way of treating burn damage.
> 
> In general fire's issue right now is that it does a LOT of damage to marines, and hardly anything to xenos. It's bad enough that xenos regularly run into/through fire and purposefully drag marines into it, now this is making that tactic even more effective - when fire is meant to be one of the direct counters to xenomorphs in the lore.
> 
> This is just going to cause more issues than it solves imo.

Green fire was weak to walk into. But you were taking 5 damage per second while on fire. 

The other fires are still dangerous to walk into and this doesn't affect that. This only affects the damage you take while being ignited and on fire.

Most of the fire damage comes from being on fire. I wouldn't say it hardly does anything to xenos considering I considered it in the PR to divide the damage by 5 times for marines.

This makes it so being on welder fire vs green fire vs regular fire vs blue fire has a difference. Welder fire isn't going to crit you especially with how it should be around 2 dps instead of 5 dps.

It makes it so being on blue fire and white fire is more dangerous as it was already dangerous before walking through it. These are only accessible to small parts ( incidendary OB ( limited to 2 + tech ) + pyro spec ( initial set + points + OT) ) so you really shouldn't notice much of a difference. 

However, you should notice a difference with weaker flames. Sources of weaker flames:
- Welders
- Underbarrel flamer
- Incidendary grenade( green )
- Pyro spec of course
- CAS Banshee
- Incidendary mortar

I doubt you will notice the 1 extra damage from regular flamers but I could be wrong in that being game changing. I do think this PR will actually benefit marines more as green fire is also getting a rework.

kiVts:
wait doesnt banshee gives of blue flame?

TeDGamer:
> wait doesnt banshee gives of blue flame?

You're right. I misread that.

stanalbatross (approved):
code approved x2

Nanu308:
@stanalbatross will be left in charge to approve/deny/request changes for Balance on this PR as needed.

stanalbatross (approved):
I am fine with this balance wise

doesn't make much sense they should all do the same damage

PR #641, opened 2022-07-08
removes the mixed metal plas crate

stanalbatross:
## About The Pull Request

removes the mixed metal plas crate from req

## Why It's Good For The Game

did you know req gets more money now? and cades are cheaper? so having another way to avoid weimar republic inflation is not good for req 

tweaks down their metal spam ability a bit

also should make sandbags used more

## Changelog

:cl:
balance: removed the "mixed metal and plasteel" crate from req
/:cl:

### DIFF
--- a/code/datums/supply_packs/engineering.dm
+++ b/code/datums/supply_packs/engineering.dm
@@ -39,15 +39,6 @@
 	containername = "plasteel sheets crate"
 	group = "Engineering"
 
-/datum/supply_packs/plas_metal
-	name = "mixed metal sheets (x30 metal, x20 plasteel)"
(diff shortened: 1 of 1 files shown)


JohnDoe50974:
How bout you remove the infinite free req money instead god motherfucking damn!!!!

lcarapaica:
That is happening in the IO TM, along with this and the previous engineer buffs to barricade costs should be looking good.

lcarapaica:
> How bout you remove the infinite free req money instead god motherfucking damn!!!!

That is happening in the IO TM, along with this and the previous engineer buffs to barricade costs should be looking good but not overly easy / obnoxious to deal with as xeno.

PR #645, opened 2022-07-09
[Bugfix/QOL] Allows hive core repair again, Prevent XvX quick core destruction, Allow for core repair on "undamaged" core that lost health

TeDGamer:
## About The Pull Request
Core can be repaired again. It was due to the early return that it never called it super function.

Core can only be destroyed with windup by the same hive member. Other hive xenos should still be able to destroy the core albiet with slashing as before instead of new wind up destruction.

Core and clusters can now be repaired when health is less than max outside of being damaged.
A core / cluster is only considered damaged when a weed tile is destroyed ( so even with no health loss you can repair it).
Since repairing repairs the health, I thought it makes sense to make it so if the health is also less than max, you should be able to repair it as well.

## Why It's Good For The Game
Bug fixing for allowing xenos to repair core again. 

Bug fixing also to make sure only the same hive member can destroy the core to prevent XvX quick core destruction. Doesn't affect the slow clawing, just the new destruction of core mechanic that uses a simple wind up.

QOL change to allow core to be fixed when health is less than max and not just when one weed tile is removed ( which is when damaged is set to true ). 

## Changelog

:cl: TeDGamer
fix: Hive core can be repaired again
fix: Xenos must be in the same hive to destroy the core quickly
qol: Hive core and cluster can be repaired when health is less than max
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/special/pylon_core.dm
+++ b/code/modules/cm_aliens/structures/special/pylon_core.dm
@@ -47,7 +47,7 @@
 /obj/effect/alien/resin/special/pylon/proc/do_repair(mob/living/carbon/Xenomorph/M)
 	if(!istype(M))
 		return
-	if(!damaged)
+	if(!damaged && health >= maxhealth)
 		to_chat(M, SPAN_XENONOTICE("\The [name] is in good condition, you don't need to repair it."))
 		return
 
--- a/code/modules/cm_aliens/structures/special/pylon_core.dm
+++ b/code/modules/cm_aliens/structures/special/pylon_core.dm
@@ -47,13 +47,14 @@
 /obj/effect/alien/resin/special/pylon/proc/do_repair(mob/living/carbon/Xenomorph/M)
 	if(!istype(M))
 		return
-	if(!damaged && health >= maxhealth)
+	var/can_repair = damaged || health < maxhealth
+	if(!can_repair)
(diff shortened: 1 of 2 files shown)


code/modules/cm_aliens/structures/special/pylon_core.dm
```diff
@@ -121,25 +122,24 @@
...
 /obj/effect/alien/resin/special/pylon/core/attack_alien(mob/living/carbon/Xenomorph/M)
-	if(M.a_intent != INTENT_HARM || !M.can_destroy_special())
-		return
-	if(!hardcore && last_attempt + 6 SECONDS > world.time)
-		to_chat(M,SPAN_WARNING("You have attempted to destroy \the [src] too recently! Wait a bit!")) // no spammy
-		return XENO_NO_DELAY_ACTION
-
-	else if(warn && world.time > HIVECORE_COOLDOWN_CUTOFF)
-		if((alert(M, "Are you sure that you want to destroy the hive core? (There will be a 5 minute cooldown before you can build another one.)", , "Yes", "No") == "No"))
+	if(M.a_intent == INTENT_HARM && M.can_destroy_special() && M.hivenumber == linked_hive.hivenumber)
```
stanalbatross, line 125:
```suggestion
	if(M.a_intent != INTENT_HELP && M.can_destroy_special() && M.hivenumber == linked_hive.hivenumber)
```

stanalbatross (changes requested):

stanalbatross (approved):

PR #647, opened 2022-07-10
bugfix check if mob is synth when on fire

BKryth:
## About The Pull Request

Stops synths from screaming, including whiteout, when on fire

## Why It's Good For The Game

Synths dont feel pain, why are they screaming, HELLO GEEVES, THANKS FOR THE FREE GBP

## Changelog

:cl:
fix: Synth type of mob no longer scream repeatedly when on fire
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -48,7 +48,7 @@
 			if(!stat && getToxLoss() >= 45 && nutrition > 20)
 				vomit()
 
-			if(on_fire)
+			if(on_fire && !isSynth(src))
 				INVOKE_ASYNC(src, /mob.proc/emote, pick("pain", "scream"))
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/life.dm
```diff
@@ -48,7 +48,7 @@
 			if(!stat && getToxLoss() >= 45 && nutrition > 20)
 				vomit()
 
-			if(on_fire)
+			if(on_fire && !isSynth(src))
```
Geevies, line 51: check if the species feels pain instead of doing a synthcheck

Geevies (changes requested):

Geevies:
if you wanna be super cool, make this be a proc on the species level so we can define which screams each species should do

Geevies (approved):

PR #678, opened 2022-07-13
energy gun + flash fixes

### LINKED ISSUE #594: Changing item sprites don't work (MP tools)
## Description
When an item has different sprites depending on its status, it doesn't seem to actually work.
image
image

Taser should visually drain itself when shot, it don't. Same with a burned out flash, it appears like as if it still works.

stanalbatross:
## About The Pull Request

closes #594 

## Why It's Good For The Game

qol when using these tools plus bugfix

## Changelog

:cl:
add: energy weapons, provided they have a charge meter, now tell you how many shots they have left on examine and when you fire them
fix: taser overlays now update properly again
fix: flashes now visually burn out, added a modern sprite for burnt out flashes
code: rewrote most of the code for flashes. It was really, really bad.
add: flashes now tell you how many flashes they have left when you use them, when you examine them, and they tell you when they recharge a flash
add: you can now only use a flash twice per second.
/:cl:


MusselMain:
Many thanks, these always sucked to work with as a MP, and were more complicated, gud change.

code/modules/projectiles/guns/energy.dm
```diff
@@ -134,7 +147,7 @@
 /obj/item/weapon/gun/energy/taser/able_to_fire(mob/living/user)
 	. = ..()
 	if (. && istype(user)) //Let's check all that other stuff first.
-		if(!skillcheck(user, SKILL_POLICE, SKILL_POLICE_SKILLED))
+		if(skilllock && !skillcheck(user, SKILL_POLICE, SKILL_POLICE_SKILLED))
```
Shad0vvs, line 150:
```suggestion
		if(skilllock && !skillcheck(user, SKILL_POLICE, skilllock))
```

Shad0vvs (changes requested):
Synthflash uses normal flash animation so it transforms in your hand, otherwise good to go.

Shad0vvs (approved):

PR #693, opened 2022-07-17
Fix for reflective walls.

Segrain:
## About The Pull Request

Thanks to walls tracking damage from zero up to maximum instead of health from maximum down to zero, nobody ever noticed that the multiplier meant to reduce the damage of incoming projectile actually reduced the damage taken so far every time the wall got shot, meaning it is literally impossible to take a reflective wall down by shooting. Fixed at last.
Numbers involved will still need a look from balance perspective at some point. A hundred bullets to take down a single wall might be a bit unreasonable. Reflection itself will need a look too, but that is a general accuracy issue in need of a separate fix.

## Why It's Good For The Game

Is fix.

## Changelog

:cl:
fix: Reflective walls now actually take damage from the bullets they did not reflect.
/:cl:

### DIFF
--- a/code/game/turfs/walls/wall_types.dm
+++ b/code/game/turfs/walls/wall_types.dm
@@ -1087,36 +1087,34 @@
 
 	var/brute_multiplier = 0.3
 	var/explosive_multiplier = 0.3
+	var/reflection_multiplier = 0.5
 
 /turf/closed/wall/resin/reflective/bullet_act(obj/item/projectile/P)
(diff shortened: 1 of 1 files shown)


Carnage068:
Just when I thought they couldn't get any worse. But hey, if it's a bug, it's a bug.

stanalbatross:
we can look at buffing reflective walls later if they are too weak with this fixed

PR #694, opened 2022-07-17
Warden minor rework

stanalbatross:
## About The Pull Request

replaces the warden shield ability with an ability called "retrieve" that can retrieve a xeno from far away like an oppressor hook

shaddeh gave me the idea for this thank you @Shad0vvs 

balance details NOT FINAL

## Why It's Good For The Game

shields removal design doc

shields r bad, this is better. makes you engage your brain a bit more.

## Changelog

:cl:
del: removed the ability of wardens to shield other xenos
add: added a new ability to wardens : "retrieve". Works like an oppressor hook but for allies - it can hook and pull an ally to you with a short windup. It costs 100 internal health reserves, and can pull any size xeno while they are resting, but big xenos cannot be pulled while they are standing. The pulled xeno will go through mobs, and will land on the tile behind you if it is not dense.
add: added "xeno primary action five" and support for it as a macro
/:cl:


stanalbatross:
drafting because we're not final

FoxyStalin:
This is going to be really funny to grief with.

GoldenDarkness55:
Maybe let them speed up xenos instead? Feels more in line with the heal/cc removal.

GoldenDarkness55:
> a simple point, click, and you're sped up is kinda boring. I wanted wardens to have a fun and engaging ability instead of a simple click once to win

Would be better to re-do the heal and CC removal alongside then, cause otherwise this kinda sticks out. Maybe just merge it all into this hook-save giving it a small heal and cc remov?

Still not a fan of abilities like this though - it really feels like it's been taken straight out of a MOBA.

stanalbatross:
> > a simple point, click, and you're sped up is kinda boring. I wanted wardens to have a fun and engaging ability instead of a simple click once to win
> 
> Would be better to rework the heal and CC removal alongside then, cause otherwise this kinda sticks out. Maybe just merge it all into this hook-save giving it a small heal and cc remov?

No because then warden would be completely reliant on it and its cooldown to be useful at all

> Still not a fan of abilities like this though - it really feels like it's been taken straight out of a MOBA.

Great, I don't care

Carnage068:
> No because then warden would be completely reliant on it and its cooldown to be useful at all

Yeah I feel like castes/strains relying entirely on one ability to function just generally isn't good game design. It's very binary.

RadiationFan:
> > No because then warden would be completely reliant on it and its cooldown to be useful at all
> 
> Yeah I feel like castes/strains relying entirely on one ability to function just generally isn't good game design. It's very binary.

Quite a few xenos have only one ability that makes the rest of the kit function, though
Lurker-Invis
Runner-Pounce
Base Rav-Dash
Hedgehog-Shield
Warrior-Lunge
Boiler- Bombard
Crusher-Charge
Queen-Scree
And if it is not a single ability, it is rarely more than two.
Spitter-Acid spit>Vomit combo
Boxer-I don't know the abilities names off the top of my head but you get the point.
The only xenos that really have multiple ways of doing things are builders and drone castes, burrower, carrier ETC. 
All that stuff being said, removing shields is not really going to help balance the game that much, or make the game better, it will just make xenos squishier, which is pretty pointless as you will just end up with fights that are even more lopsided than they already are. If xenos are going to have all shields removed eventually, marine damage will have to come down quite a bit, because even now with shields as prevalent as they are, xenos still get melted in seconds, whether t1 or queen.

stanalbatross:
> The changelog's usage of only is pretty confusing because it is at conflict with the statement it can pull any size xeno while they are resting "...It costs 100 internal health reserves, and can pull any size xeno while they are resting, but can **only** pull non-big xenos while they are standing..." probably should say "also" instead of "only".
> 
> There are likely some situations the retrieve is useful (I'd imagine to sort of pre-save someone or in a tight choke), but it also feels at odds with Rejuvenate. It would be a lot better to just combine this skill with Rejuvenate and maybe allow you to just remove Protect & Toggle Protection Type and have everything be its own skill.

updated cl entry, also a good idea

Carnage068:
> Quite a few xenos have only one ability that makes the rest of the kit function, though Lurker-Invis Runner-Pounce Base Rav-Dash Hedgehog-Shield Warrior-Lunge Boiler- Bombard Crusher-Charge Queen-Scree And if it is not a single ability, it is rarely more than two. Spitter-Acid spit>Vomit combo Boxer-I don't know the abilities names off the top of my head but you get the point. The only xenos that really have multiple ways of doing things are builders and drone castes, burrower, carrier ETC. All that stuff being said, removing shields is not really going to help balance the game that much, or make the game better, it will just make xenos squishier, which is pretty pointless as you will just end up with fights that are even more lopsided than they already are. If xenos are going to have all shields removed eventually, marine damage will have to come down quite a bit, because even now with shields as prevalent as they are, xenos still get melted in seconds, whether t1 or queen.


Uh, no??? Sorry, I don't want to be rude like I usually am, but you REALLY don't play a whole lot of Xeno, do you? Runners are incredibly fast and are amazing at avoiding gunfire or simply running people down. The fact that Acid Runners don't have a pounce, are significantly slower, yet STILL can somewhat fulfill the role regular runners do is proof enough that they're not completely reliant on the ability to do anything meaningful.
You say that Runners are completely reliant on their pounce, but then you say Lurker is reliant on stealth even though it has a pounce of its own that covers even more distance, stuns for way longer, and helps them get out of situation they wouldn't be in if they had the movement speed of a Runner? What?
Base Rav isn't overly reliant on its dash. If it was, it would be dogshit and boring. It has decent movement speed for a T3, a potentially huge shield, and fantastic slash damage. If it was overly reliant on its dash to function, every time a Base Rav charged in and didn't get the cooldown reset on its leap, it would die. That RARELY happens. I honestly have no idea what you're talking about, and again, this kind of shows that you aren't very familiar with any of the xeno castes.
Hedgehog is similar, it's not completely reliant on its shield. That's just wrong.
Warriors you might be right about, but that's probably the only thing you said that's correct.
Crusher isn't reliant on its charge. It's reliant on its single target bullshit permaslow to kill people because of its dogshit movement speed. If it's reliant on anything, it's probably the goddamn shield.
Yeah, Boiler's only good ability is the acid cloud, but that's because they made everything else awful.
Take away Queen's screech, she still has neuro. Screech and neuro are both 50% of her entire kit, so saying she's just reliant on ONE of them and not both is just further evidence that you never play xenomorph.

And no, marines don't need their damage nerfed. I have no idea what you're talking about when you say Queen is being melted, because last time I checked Queen is completely fine. T1s besides Defender are SUPPOSED to get melted. You can't frontline with a fucking sentinel.

MistChristmas:
Just gonna throw my two cents in the hat as a resident zero brain warden main when I play prae. Cause shields be powerful. Taking extra hits to kill can turn dead beno into dead marines. Even more so because Prae's shield also effects themselves. Giving you a two for one. So this does kinda reduce overall xeno combat power - eh ish. I can totally fore-see Retrieve saving xeno's but I don't think it's powerful enough to replace shielding. Could think about maybe buffing the xeno that is getting retrieved during the wind-up a little? Or maybe just increase some of the strains base stats to compensate. Have to see. Though I do think the ability itself is a nice thing to have. And fits a Warden sort of theme.

Personally I wouldn't mind having shields in the bin. Though i'd prefer that AP also goes in the bin and maybe the armour break returns in some form.

Carnage068:
> Just gonna throw my two cents in the hat as a resident zero brain warden main when I play prae. Cause shields be powerful. Taking extra hits to kill can turn dead beno into dead marines. Even more so because Prae's shield also effects themselves. Giving you a two for one. So this does kinda reduce overall xeno combat power - eh ish. I can totally fore-see Retrieve saving xeno's but I don't think it's powerful enough to replace shielding. Could think about maybe buffing the xeno that is getting retrieved during the wind-up a little? Or maybe just increase some of the strains base stats to compensate. Have to see. Though I do think the ability itself is a nice thing to have. And fits a Warden sort of theme.
> 
> Personally I wouldn't mind having shields in the bin. Though i'd prefer that AP also goes in the bin and maybe the armour break returns in some form.

Prae's shield affecting themselves was something done fairly recently. They didn't have it before and were fine without it. They can live without it again. Then again, before their shields affected themselves, they had their health reserves dump into their own health whenever they went into crit, which was arguably better, but at least it was easier for marines to tell what was going on. Even still, Warden still has the ability to heal other Xenos, and considering they also get half of that health, I think that's good enough. Like, the pull ability works REALLY well with the heal, and the pull just has a lot of utility in general.

AP doesn't really need to be axed, although I can see where you're coming from. The problem with AP is that it's just too goddamn strong right now, and regular ammo is too weak in comparison. Like, if you shoot a Boiler with AP, it will still deal a good amount of damage- Regular just does more. If you shoot a Hedgehog with regular ammo, it barely does jack shit, meanwhile AP pretty much just melts it. Regular ammo isn't mandatory for killing castes with lower armor, so why is AP mandatory for killing castes with higher armor? It's not good balance, imo.
AP ignores like, 100% of armor unless the caste has 45 armor or higher, which is pretty much the point where regular ammo stops doing anything. Ideally AP should start slowly falling off after 25 armor, and the castes that can get 45 armor or higher should get their armor slightly reduced to compensate.

Agameofscones (approved):
action sprites are serviceable

Listaless:
Dropping some test play feedback:
Pull can be avoided by ally's which means if a ally is moving when you are attempting to rescue them it will miss, which honestly makes the ability fairly bad, while making the pull instant would likely solve it, I think it would make it too strong, so instead have it lock on to the target then pull them back after the current duration, which would make the ability more reliable to use.

Aspaceman12:
i really like this feature it spices up warden and is more engaging that just healing and spitting

stanalbatross:
> Or maybe just increase some of the strains base stats to compensate

will think about 

> if a ally is moving when you are attempting to rescue them it will miss

it's _mainly_ intended to be used on crit xenos, if the xeno stays in the green target area they will get saved

tbh not a fan of making it track, might just increase the target area

Carnage068:
> it's _mainly_ intended to be used on crit xenos, if the xeno stays in the green target area they will get saved

Kind of an issue imo, considering Xenos don't really stay in crit in the vast majority of situations, and instead just instagib. It's a well designed ability (way better than just a flat shield) but its main intended purpose is really damn ineffective currently. Most of the time the Warden player is going to have to just predict what tile a low health Xeno will be walking on BEFORE they even get into crit range, which wouldn't be a problem if it weren't for just how difficult that often is, even for more experienced players, so if you're gonna do anything, you should probably reduce the windup or (like you said) increase the target area. Either that, or give Warden something that makes Xenos last longer when they're in crit, but I have a feeling marines would hate that.

code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_powers.dm
```diff
@@ -867,3 +867,150 @@
...
+
+	var/throw_target_turf = get_step(X.loc, facing)
+	var/turf/behind_turf = get_step(X.loc, reversefacing)
+	if(!(behind_turf.density))
+		throw_target_turf = behind_turf
+
+	X.frozen = TRUE
+	X.update_canmove()
+	if(windup)
+		if(!do_after(X, windup, INTERRUPT_NO_NEEDHAND, BUSY_ICON_HOSTILE, null, null, FALSE, 1, FALSE, 1))
```
Shad0vvs, line 977:
```suggestion
		if(!do_after(X, windup, INTERRUPT_NO_NEEDHAND, BUSY_ICON_HOSTILE, null, null, FALSE, 1, FALSE, 1))
```
Can possibly do this better/more readable.

Shad0vvs (approved):

code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_powers.dm
stanalbatross: improved

Nanu308:
Mhm yes reverse oppressor my beloved

Mirvra:
Only tested it two rounds so far.

But currently it feels way too expensive for what it does.
100 health reserve pool is expensive as heck just to pull someone, and if you have warriors that keep demanding you heal them then you'll find yourself without reserve pool when you actually have to use it... Have to ask why the ability uses health reserve pool? Why not just make it an incredibly expensive plasma ability? Or is the ability really meant to make you tell low health players ''no I won't heal you because I want to stay at full reserve to be ready in case someone goes crit''

Secondly, as @Carnage068 said, the amount of time you have to determine IF you need to use it, WHERE, you need to use it, and WHEN you need to use it is stupidly short. Even with Warding pheremones you just don't have the time if the xeno is already at crit, especially since they can still walk a step or two before actually going into pain crit.


Side note: I just noticed the wiki says Warden heal should only cost 100 health reserve and plasma... but I could've sworn using it consumes way more than that? ... Doesn't using the health reserve consume a larger portion of it based on how much you have? And thus it makes it so you'll never have enough in the pool to pull someone in case you healed someone recently and haven't been out spitting?

PR #695, opened 2022-07-17
[Bugfix] Accounts for darkness when checking if source can see target

TeDGamer:
## About The Pull Request
can_see now accounts if the target is in darkness and if the source doesn't have nightvision.

see_in_dark seems to be at least 12 ( for base xenos nightvision and nvgs ) so it checks that and if the lighting lumcount of the tile is 0 ( I wasn't sure how to check for darkness otherwise with the new plane master )

I use the lumcount being 0 to be equal to darkness. 

Originally, the luminosity should be 0 when there is no light. Watermelon914 tried to fix this in their planemaster [PR ](https://github.com/cmss13-devs/cmss13/commit/1f27a282e57908ce3fddd841225b28ca563b62cc)but for some reason it lead to performance problems which lead to the [revert](https://github.com/cmss13-devs/cmss13/commit/0fc7bd68aa8c780e6ef28c55c98fd38cc809f033)

This however allows anybody with no nightvision to right click dark tiles to see what's there. I believe this is leading to a higher accuracy than normal since opacity isn't updated properly due to luminosity never being set to 0 in the lightning controller anymore.

IF you revert the code to the if else code, it prevents right clicking into darkness which I think restores the opacity count:
```
       if(light <= 0)
		light = 0
		luminosity = 0
	else
		luminosity = 1
```

## Why It's Good For The Game
Accuracy of blind firing into darkness shouldn't be as high as it is now. Can't revert the code for luminosity to prevent right clicking the darkness due to performance issues mentioned in the revert PR but this should be a stopgap due to firing at a target with no light should mean you can't see the target unless you have nightvision.

## Changelog

:cl: TeDGamer
fix: can_see now takes into account the darkness of the target tile and whether you have nightvision
/:cl:

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -861,7 +861,7 @@
 	if(ismob(source))
 		var/mob/M = source
 		has_nightvision = M.see_in_dark >= 12
-	if (!has_nightvision && target_turf.lighting_lumcount == 0)
+	if(!has_nightvision && target_turf.lighting_lumcount == 0)
 		return FALSE
 
 	while(current != target_turf)
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -858,7 +858,7 @@
 	var/turf/target_turf = get_turf(target)
 	var/steps = 0
 	var/has_nightvision = FALSE
-	if (ismob(source))
+	if(ismob(source))
 		var/mob/M = source
 		has_nightvision = M.see_in_dark >= 12
 	if (!has_nightvision && target_turf.lighting_lumcount == 0)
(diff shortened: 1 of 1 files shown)


code/__HELPERS/unsorted.dm
```diff
@@ -857,6 +857,12 @@
 	var/turf/current = get_turf(source)
 	var/turf/target_turf = get_turf(target)
 	var/steps = 0
+	var/has_nightvision = FALSE
+	if (ismob(source))
```
stanalbatross, line 861:
```suggestion
	if(ismob(source))
```

code/__HELPERS/unsorted.dm
```diff
@@ -857,6 +857,12 @@
...
+		var/mob/M = source
+		has_nightvision = M.see_in_dark >= 12
+	if (!has_nightvision && target_turf.lighting_lumcount == 0)
```
stanalbatross, line 864:
```suggestion
	if(!has_nightvision && target_turf.lighting_lumcount == 0)
```

stanalbatross (changes requested):

stanalbatross (approved):

PR #75, opened 2022-04-17
Telebaton

realforest2001:
## About The Pull Request
Gives the telebaton a 1s stun on humans when used in help intent. This is something in most other codebases and can confuse newer players that use one.

## Why It's Good For The Game
Gives the CO and Synthetics a readily available non-lethal defence method, and makes it harder to accidentally break a leg.

## Changelog

:cl:
add: Telebaton now stuns when not on harm intent.
/:cl:

PR #697, opened 2022-07-18
telebaton stun redux

stanalbatross:
## About The Pull Request

PR #75 all over again

## Why It's Good For The Game

makes it consistent with other servers + way for synths and COs to defend themselves against angry marines

does NOT affect xeno combat

## Changelog

:cl:
add: the telebaton can now deal 50 stam damage when used on a human if not on harm intent
/:cl:


morrowwolf:
Good change +1

JohnDoe50974:
Thanks :3

stanalbatross:
> Make it deal stamina dmg, also, it can perma stun potentially a single target as it has no max energy...
> 
> Probably fine, but keep it in mind.

done

morrowwolf:
So while two hits with a stunbaton will likely knock someone down given it is checking against energy armor (which most marines don't have) the telebaton is checking against melee defense which most marines do have a good bit of. It may be good to increase the stunforce higher to be more effective as it's three hits with just a regular USCM uniform to knock over and more with any armor. 125 stunforce will reliably knock over a marine in normal armor in two hits. 100 is three hits with standard gear/armor.

Having the extension and collapsing use danger and notice respectively rather than the same feels funny to me but designer choice and this PR doesn't explicitly touch that.

Otherwise, LGTM.

code/game/objects/items/weapons/swords_axes_etc.dm
```diff
@@ -75,14 +83,28 @@
...
 
 	if(blood_overlay && blood_color)
 		overlays.Cut()
 		add_blood(blood_color)
 	return
 
+/obj/item/weapon/melee/telebaton/proc/stun(mob/living/carbon/human/target, mob/living/user)
+	if(target.check_shields(src, 0, "[user]'s [name]"))
+		return FALSE
+	user.visible_message(SPAN_DANGER("[user] knocks down [target] with \the [src]!"),\
```
JeserTheFox, line 97: This always says "knocks down" even if you don't actually knock down target. Either need to be made more universal or do a check post-hit whether target is stunned/lying and show respective message.

morrowwolf (approved):
LGTM

PR #704, opened 2022-07-18
Rebalance: Removes greenfire slowdown

JohnDoe50974:
## About The Pull Request
Opened after discussing current greenfire balance with the development team. The slowdown effect has been axed, nothing else was touched.

## Why It's Good For The Game
Armor reduction is a good niche to give greenfire more use cases, but coupling it with the slowdown was overkill.  The armor reduction is a fair and significant punishment for xenos that walk on through the greenflamed tiles, but the slowdown messes hard with an essential cornerstone of xeno gameplay - the ability to engage or disengage from fights. Why is this so bad? Because this rather significant slowdown is applied per-flamed tile, and greenfire blankets a wide area - other flame types simply flame in a straight line. This in essence creates a significantly large Area-of-effect debuff, which is out of pocket even for squad specialist weapons, in my opinion. This makes RPG/grenade launcher/scout hi-impact ammo (none of which simultaneously apply an on-fire effect, mind you!) (chain)stuns look weak in comparison. By removing said slowdown effect but keeping the rest in, I believe greenfire will be actually manageable by xenos whilst still significantly punishing them for stepping into it.

## Changelog

:cl: 50RemAndCounting
balance: Type-B fire (greenfire) no longer slows mobs down that walk over it.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/flamer/flamer.dm
+++ b/code/modules/projectiles/guns/flamer/flamer.dm
@@ -513,8 +513,7 @@
 				var/mob/living/carbon/human/H = M
 				if (HAS_TRAIT(M, TRAIT_SUPER_STRONG))
 					resist_modifier = 0.25
-				H.next_move_slowdown = H.next_move_slowdown + (3 * resist_modifier)
-				to_chat(H, SPAN_DANGER("The viscous napalm clings to your limbs as you struggle to move through the flames!"))
+				to_chat(H, SPAN_DANGER("The viscous napalm clings to your body!"))
 			else if(isXeno(M))
 				var/mob/living/carbon/Xenomorph/X = M
 				if(!X.armor_deflection_debuff) //Only applies the xeno armor shred reset when the debuff isn't present or was recently removed.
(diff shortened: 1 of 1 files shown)


c4xmaniac:
Might be better to keep the slow down but just nerf its intensity, green fire armor reduction still isn't that relevant and I see xenos ignore green fire all the time like they used to.

Unknownity:
> if there is not slowdown xenos can currently just walk on the fire with shaman shield . pull marine inside and GG.
> 
> why ? because green fire does Piss damage to xenos but roasts marines pretty easily . making it just reduce armor is not scary at all for xenos and they will just shrug it off like the did before.

just like xenos shrugged off toxin rounds right.

Carnage068:
> 

MUH, BUT MY GAME CONTENT-REMOVAL ADDICTION, WHAT WILL I DO IF I CAN'T REMOVE MORE LINES OF CODE FROM THE GAME?!

Nanu308:
Make a new PR where you rather lower the intensity than an outright removal, however greenfire and how easily accessible is also something you can look at with @stanalbatross

PR #705, opened 2022-07-18
no more double headbutt

### LINKED ISSUE #449: Defender double headbutt
## Description

Defender can double headbutt a target if they click on the target fast enough.

Well, it should only do it once and not twice.

## Steps to reproduce

Be a defender. Use the headbutt ability and middle or shift click twice in rapid succession on a target.
This will cause it to activate twice on the target instead of just once.

https://user-images.githubusercontent.com/104228030/172740009-ca41b651-5a07-4350-a5a4-9cff2d0f10bd.mp4

I used base defender for this but this probably works with the steel crest strain as well.

Diegoflores31:
## About The Pull Request

fixes double headbutt
- from what i tested the issue was basically that cooldown was applied too late and you could basically attack twice while the X.throwforce part was procesing.

## Why It's Good For The Game

because i hate defenders :) 
closes #449 

## Changelog

:cl:

fix: defender cant double headbutt anymore

/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/defender/defender_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/defender/defender_powers.dm
@@ -66,12 +66,11 @@
 		return
 
 	if(!X.crest_defense)
+		apply_cooldown()
 		X.throw_atom(get_step_towards(H, X), 3, SPEED_SLOW, X)
 
 	if(!X.Adjacent(H))
 		return
-
-	apply_cooldown()
 
 	H.last_damage_data = create_cause_data(X.caste_type, X)
 	X.visible_message(SPAN_XENOWARNING("[X] rams [H] with its armored crest!"), \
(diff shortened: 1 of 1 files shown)


TeDGamer:
This seems like a nerf as well as a potential fix.

This is due to applying the cooldown when you don't actually headbutt the person but throw yourself at them.

The original code, the headbutt didn't count unless it connected.

Diegoflores31:
UPDATE :cooldown now gets refunded if you are unable to hit the target (aka not adjacent after dash is done)

GoldenDarkness55:
Warrior has the same issue with fling as headbutt sometimes btw, mb out of scope for this MR but possibly same fix?

Chainsawicus:
Concerned about how this will affect the responsiveness and playability of Defender, should be testmerged because it majorly changes how its most important ability is functioning.

Diegoflores31:
> Warrior has the same issue with fling as headbutt sometimes btw, mb out of scope for this MR but possibly same fix?

make an open an issue and put a video of it and i can maybe get it fixed too

stanalbatross (approved):

JeserTheFox (approved):

PR #711, opened 2022-07-19
fixes bell going through walls

stanalbatross:
## About The Pull Request

bell will no longer trigger on enemies behind walls / blocked from seeing the bell by walls

## Why It's Good For The Game

inconsistency bug fix

## Changelog

:cl:
fix: the R1NG bell tower can no longer affect enemies through walls
/:cl:

### DIFF
--- a/code/modules/defenses/bell_tower.dm
+++ b/code/modules/defenses/bell_tower.dm
@@ -114,9 +114,11 @@
 	if(M.get_target_lock(faction))
 		return
 
-	if(!(src in view(linked_bell))) //anti exploit -> stops placing walls after making tripwires saving them
-		qdel(src)
-		return
+	var/list/turf/path = getline2(src, linked_bell, include_from_atom = TRUE)
+	for(var/turf/PT in path)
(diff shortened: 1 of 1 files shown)


Segrain:
Questionable approach. So you do not want somebody to place a bell and then wall it off, fine. What if somebody places the bell in front of a wall and then takes that wall down? There will be no tripwires on that side, and the bell will not work at all.

stanalbatross:
> Questionable approach. So you do not want somebody to place a bell and then wall it off, fine. What if somebody places the bell in front of a wall and then takes that wall down? There will be no tripwires on that side, and the bell will not work at all.

mmm, can't think of a good way to handle this other than doing the path check every time someone walks over a tripwire, and not qdeling the wires when walked over out of sight

DylonBR:
bell F tier now. mini sentries are the new frontline meta

Diegoflores31:
Tested on private server and it doesnt look as bad as it appears on paper. 👍

Shad0vvs (approved):

PR #717, opened 2022-07-19
Full-Auto Smartguns

Geevies:
## About The Pull Request

Smartguns can now only fire in full-auto mode. They no longer scatter over time when firing in full auto mode.

## Why It's Good For The Game

Prevents people from breaking their mice by spam-clicking in single fire mode to get more DPS with no scatter.

## Changelog

:cl:
balance: Smartguns now only have a full-auto firing mode, but full-auto no longer causes scatter over time.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_helpers.dm
+++ b/code/modules/projectiles/gun_helpers.dm
@@ -699,7 +699,7 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
 			flags_gun_features |= GUN_BURST_ON
 			return
 
-		to_chat(usr, SPAN_NOTICE("[src] can only be fired in bursts!"))
+		to_chat(usr, SPAN_NOTICE("\The [src] can only be fired in bursts!"))
 		return
 
 	if(flags_gun_features & GUN_FULL_AUTO_ONLY)
@@ -711,7 +711,7 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
 			RegisterSignal(usr.client, COMSIG_CLIENT_LMB_DRAG, .proc/full_auto_new_target)
 			return
 
-		to_chat(usr, SPAN_NOTICE("[src] can only be fired in full auto mode!"))
+		to_chat(usr, SPAN_NOTICE("\The [src] can only be fired in full auto mode!"))
 		return
 
 	playsound(usr, 'sound/weapons/handling/gun_burst_toggle.ogg', 15, 1)
--- a/code/modules/projectiles/gun_helpers.dm
+++ b/code/modules/projectiles/gun_helpers.dm
@@ -695,11 +695,23 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
 
 	if(flags_gun_features & GUN_BURST_ONLY)
 		if(!(flags_gun_features & GUN_BURST_ON))
-			stack_trace("[src] has GUN_BUST_ONLY flag but not GUN_BURST_ON.")
+			stack_trace("[src] has GUN_BURST_ONLY flag but not GUN_BURST_ON.")
 			flags_gun_features |= GUN_BURST_ON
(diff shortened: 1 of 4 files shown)


code/modules/projectiles/guns/smartgun.dm
```diff
@@ -52,7 +52,7 @@
 						/obj/item/attachable/burstfire_assembly,
```
Segrain, line 52:
```suggestion
```
No burst - no need for BFA.

Segrain (changes requested):

code/modules/projectiles/gun_helpers.dm
```diff
@@ -695,13 +695,25 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
...
+	if(flags_gun_features & GUN_FULL_AUTO_ONLY)
+		if(!(flags_gun_features & GUN_FULL_AUTO_ON))
+			stack_trace("[src] has GUN_FULL_AUTO_ONLY flag but not GUN_FULL_AUTO_ON.")
+			flags_gun_features |= GUN_FULL_AUTO_ON
+			RegisterSignal(usr.client, COMSIG_CLIENT_LMB_DOWN, .proc/full_auto_start)
+			RegisterSignal(usr.client, COMSIG_CLIENT_LMB_UP, .proc/full_auto_stop)
+			RegisterSignal(usr.client, COMSIG_CLIENT_LMB_DRAG, .proc/full_auto_new_target)
+			return
+
+		to_chat(usr, SPAN_NOTICE("[src] can only be fired in full auto mode!"))
```
stanalbatross, line 714:
```suggestion
		to_chat(usr, SPAN_NOTICE("\The [src] can only be fired in full auto mode!"))
```

code/modules/projectiles/gun_helpers.dm
```diff
@@ -695,13 +695,25 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
 
 	if(flags_gun_features & GUN_BURST_ONLY)
 		if(!(flags_gun_features & GUN_BURST_ON))
-			stack_trace("[src] has GUN_BUST_ONLY flag but not GUN_BURST_ON.")
+			stack_trace("[src] has GUN_BURST_ONLY flag but not GUN_BURST_ON.")
 			flags_gun_features |= GUN_BURST_ON
 			return
 
 		to_chat(usr, SPAN_NOTICE("[src] can only be fired in bursts!"))
```
stanalbatross, line 702:
```suggestion
		to_chat(usr, SPAN_NOTICE("\The [src] can only be fired in bursts!"))
```

stanalbatross (changes requested):

MusselMain:
Holy shit is this a marine buff?? (but for real this is a good change, I use to get sticky keys warnings constantly while playing SG).

Chainsawicus:
image

fa_delay should be increased to match fire_delay, otherwise you're just nerfing the Smartgun.

hotimep:
smartgun

NexusStorm1:
If you're just removing single fire, then it's a HUGE nerf for smartgun. Full auto is only 300 RPM while single fire can reach up to 600 RPM, so this is technically a 50% DPS nerf. I think most people who click normally can reach 7-8 cps, so why not make the full-auto RPM reflect that (420-480 RPM)?

stanalbatross (changes requested):
make full auto fire faster, to a similar level as what can be reached while naturally clicking in single fire (which I believe is around 450-500 rpm)

PhantomEpicness:
Good. Currently, it simply encourages scroll wheel SG gaming. Why was full auto nerfed in the first place? Literally, no reason to do it.

stanalbatross (approved):
if sg is too strong after this it can be tweaked down

PR #723, opened 2022-07-20
clf spec skill consistency

### LINKED ISSUE #709: CLF Specs can't use C4
I would like to point out that CLF Specs spawn with C4 and are unable to use it. I don't think they have Engineering skills. Not even a little.

stanalbatross:
## About The Pull Request

closes #709 , consistency issue

## Why It's Good For The Game

lets clf specs use the c4 they spawn with

## Changelog

:cl:
fix: CLF specs now have engi skill level 1 to use the c4 they spawn with, and have melee weapons skill level 1 to be in line with USCM specs
/:cl:


Segrain (approved):
Skill system still needs an overhaul - this is far from the only issue - but this will do for now.

PR #725, opened 2022-07-21
Greenfire rebalance, try #2

JohnDoe50974:
## About The Pull Request
Try 2 on green fire rebalance. A significant amount of flame-related code was refactored by Geeves and myself. The slowdown intensity has been reduced from the slowdown multiplier value of 3 to the multiplier value of 1. The code refactor includes an exception for normally fire-immune xenomorphs too, so they are no longer affected by the greenfire's effects.

## Why It's Good For The Game
The rebalanced greenfire will allow xenomorphs to still be seriously punished for walking on greenfire without almost instantly game-ending them. Escape will still be difficult, but no longer borderline impossible. Fire-immune xenos having their immunity properly apply to greenfire as well ensures that their roles in a combat situation are not unduly infringed upon, as queens and ravagers (as is any xeno caste without a pounce/leap/dash ability tbh, but slow T3s especially so) were the most vulnerable to getting flamed.
Also, code refactors will help improve server performance and possibly help prevent server crashes and code instability. More atom proccalls and bitflags, less for() loops and if statements.

## Changelog

:cl: Geeves, 50RemAndCounting
balance: Type-B fire/greenfire now no longer affects fire-immune xenomorphs. Greenfire slowdown effect has been reduced.
refactor: refactored flamer fire handling code.
/:cl:


Kaga-404:
type_b_debuff_xeno_armor() already checks and calculates for FRENZY and NO_IGNITE, and given the baseline slowdown's already being reduced from 3 to 1 in this patch both Ravagers and Queens will feel very little effect from the slowdown once modifiers are applied. It's explicitly meant to encourage targets to minimize time spent on it in addition to deterring them from crossing at all if possible, which is the role of the armor shred (also modified by the same calculation), which this entirely bypasses for _reasons_.

Excellent refactor on the overall flamer and fire end, though, and one sorely needed for future updates to flamer/fire-related features. Consider also taking a look at updating the ignition system as well, if possible.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -291,3 +291,26 @@
 
 	var/list/overlap = iff_tag.faction_groups & access_to_check
 	return length(overlap)
+
+/mob/living/carbon/Xenomorph/handle_flamer_fire(obj/flamer_fire/fire, var/damage, var/delta_time)
+	. = ..()
+	if(caste && (caste.fire_immunity & (FIRE_IMMUNITY_NO_DAMAGE|FIRE_IMMUNITY_NO_IGNITE)))
+		return
```
Kaga-404, line 298: See above, already automatically compensated for when type_b_debuff_xeno_armor() is called.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -291,3 +291,26 @@
...
+	switch(fire.fire_variant)
+		if(FIRE_VARIANT_TYPE_B)
+			if(!armor_deflection_debuff) //Only adds another reset timer if the debuff is currently on 0, so at the start or after a reset has recently occured.
+				reset_xeno_armor_debuff_after_time(src, delta_time*10)
+			fire.type_b_debuff_xeno_armor(src) //Always reapplies debuff each time to minimize gap.
+
+/mob/living/carbon/Xenomorph/handle_flamer_fire_crossed(obj/flamer_fire/fire)
+	. = ..()
+	if(caste && (caste.fire_immunity & (FIRE_IMMUNITY_NO_DAMAGE|FIRE_IMMUNITY_NO_IGNITE)))
+		return
```
Kaga-404, line 308: Ditto.

Kaga-404:
I'll dig through the code for sticky resin at a later point and see how this matches up, but that's around the desired baseline slow from what I recall from my discussions. The present values were already considered fine as they were and the original plan was to look into _reducing availability_ rather than modifying effects, but I'm willing to see how a reduced slow adjustment performs on its own as effectively a long-term live test.

JohnDoe50974:
> I'll dig through the code for sticky resin at a later point and see how this matches up, but that's around the desired baseline slow from what I recall from my discussions. The present values were already considered fine as they were and the original plan was to look into _reducing availability_ rather than modifying effects, but I'm willing to see how a reduced slow adjustment performs on its own as effectively a long-term live test.

I personally viewed reducing its effects as a priority, due to the slowdown effect in particular being blatantly overpowered with nothing else in the marine arsenal being quite comparable to it at all. I do not see availability as an issue, pyrotechnician spec can acquire very large amounts of greenfire and regular incinerator units are hard enough to obtain that I don't believe the free ASRS flame tanks matter.

Kaga-404:
> I personally viewed reducing its effects as a priority, due to the slowdown effect in particular being blatantly overpowered with nothing else in the marine arsenal being quite comparable to it at all. I do not see availability as an issue, pyrotechnician spec can acquire very large amounts of greenfire and regular incinerator units are hard enough to obtain that I don't believe the free ASRS flame tanks matter.
Fulfilling the previously unoccupied niche was the intended effect, correct. Pyro Spec's newfound frontal-based utility is intended, as the flames have to propagate and require a fire-delay before extinguishing (don't ask me why extinguishers share the cooldown with the gun itself, it just works), preventing immediate chasedown even if a target is caught on the edge.

I'm not sure if I'm reading this right, but it seems like Sticky Resin applies a movement delay of _8?_ That's... Interesting. I'll test to confirm if it's relative to the 3 Type-B's applying at some point.

That aside, my comments stand as previously mentioned. I'm willing to see how a reduced slow functions, but clean out the unnecessary fire flag checks marked above. Good to have the data on how it'll perform with just a reduced baseline slow given we might be getting a flamer nozzle at some point, which is going to make me do a full pass over on Type B.

Reduced availability also means clipping down how many ways there are to obtain cheap HIDPs/Incendiary Mortars, such as from packets or ASRS.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
stanalbatross: this should be moved down to before the slowdown check so that it doesn't affect NO_IGNITE xenos

stanalbatross (changes requested):

JohnDoe50974:
Due to the recent fluctuations in balance considerations, I've concluded that nerfing greenfire would be out of place and tilt the balance too harshly towards xenos. Unless said fluctuations get resolved, I'm closing this pull request until further notice.

Paperbell:
I think using sticky resin to justify the strength of green fire isn't a good way to balance it. Sticky resin needs to be made one cell at a time. While cutting it away with a bayonet is slow, every marine can do it and every round has a lot of marines. In combat, marines can fight well without stepping on the sticky resin because they have guns. A lot of sources of green fire can cover large areas and with more freedom. The resin whisperer hivelords need weeds to create the sticky resin at range, and green fire can instantly overwrite anything they do. A lot of xeno castes don't have a way to extinguish green fire, and the ones that do can only use their sprays after waiting through a cooldown.

I don't play much xeno but if a lot of marines were forced to use an smg that could vaporize things but could only do damage within a range of 2, we would probably lose a lot if the queen could create 5x5 areas of sticky resin and weeds with an ability that could be used once every 5 seconds, even with pyro and larger numbers.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
JohnDoe50974: I just rebased my branch to latest Master iteration and these two lines do not exist. Where are these from, exactly?

Carnage068:
How did I only notice this now? This is a great nerf- I thought someone was just going to delete the slow entirely, but it's much better to just make it less extreme. Green fire will still be good, it just won't instantly fuck over any Xeno without a leap. I just hope it actually gets merged lol

JohnDoe50974:
Too many issues and developers forced me to expand the scope way too much beyond what I was interested in doing. If you want to do this yourself, look at Geeves' commit and start from there.

PR #726, opened 2022-07-21
xm88 heavy rifle (marine r4t)

stanalbatross:
## About The Pull Request

adds the marine version of the r4t, has a semi-similar stacks mechanic

## Why It's Good For The Game

adds more variety to marine weapons, adds a new more skill-based gun that rewards landing direct clicks on moving targets

is atmospheric and soulful

## Changelog

:cl: Stan_Albatross, Thwomper
add: added the xm88 heavy rifle to req, it is a rifle that must manually chamber each shot. Hitting direct hits will grant "stacks", which increase its ap value and damage up until 40 ap. Sprites by thwomper.
fix: trying to fire a weapon with an internal magazine without a chambered round will no longer display a message saying how many rounds are left in the magazine
/:cl:


sose96:
Looking forward to the test merge

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1386,7 +1386,7 @@
 	name = "lever-action bullet"
 
 	damage = 80
-	penetration = ARMOR_PENETRATION_TIER_1
+	penetration = null
```
Segrain, line 1389:
```suggestion
	penetration = 0
```

code/modules/projectiles/guns/lever_action.dm
```diff
@@ -327,6 +347,142 @@ their unique feature is that a direct hit will buff your damage and firerate
...
+	lever_message = initial(lever_message)
+	wield_delay = initial(wield_delay)
+	cur_onehand_chance = initial(cur_onehand_chance)
+	if(in_chamber)
+		var/obj/item/projectile/P = in_chamber
+		P.ammo = GLOB.ammo_list[/datum/ammo/bullet/lever_action/xm88]
+	floating_penetration = FLOATING_PENETRATION_TIER_0
+	//these are init configs and so cannot be initial()
+	lever_delay = FIRE_DELAY_TIER_3
+	fire_delay = FIRE_DELAY_TIER_1
```
Segrain, line 478: This is very different from the delay in `set_gun_config_values()`. Intended?

Segrain (changes requested):

Geevies:
EB Broken Sprite
image

Refers to lever action tube when reloading
image

code/modules/projectiles/gun_attachables.dm
```diff
@@ -951,6 +951,17 @@ Defined in conflicts.dm of the #defines folder.
 	zoom_offset = 7
 	dynamic_aim_slowdown = SLOWDOWN_ADS_NONE
 
+/obj/item/attachable/scope/mini/xm88
+	name = "XS-9 targeting relay"
+	desc = "An ARMAT XS-9 optical interface. Unlike a traditional scope, this rail-mounted device features no telescoping lens. Instead, the firearm's onboard targeting system relays data directly to the optic for the system operator to reference in realtime. "
```
XSlayer300, line 956:
```suggestion
	desc = "An ARMAT XS-9 optical interface. Unlike a traditional scope, this rail-mounted device features no telescoping lens. Instead, the firearm's onboard targeting system relays data directly to the optic for the system operator to reference in realtime."
```

code/modules/projectiles/guns/lever_action.dm
```diff
@@ -327,6 +347,142 @@ their unique feature is that a direct hit will buff your damage and firerate
...
+
+#define FLOATING_PENETRATION_TIER_0	0
+#define FLOATING_PENETRATION_TIER_1	1
+#define FLOATING_PENETRATION_TIER_2	2
+#define FLOATING_PENETRATION_TIER_3	3
+#define FLOATING_PENETRATION_TIER_4	4
+
+/obj/item/weapon/gun/lever_action/xm88
+	name = "\improper XM88 heavy rifle"
+	desc = "An experimental man-portable anti-material rifle originally developed by ARMAT Battlefield Systems for the government of the state of Greater Brazil for use in the Favela Wars (2161 - Ongoing) against mechanized infantry. Chambered in .458 SOCOM, the platform features an onboard computerized targeting system, sensor array, and an electronic autoloader; these features work in tandem to reduce and render inert armor on the users target with successive hits. The Almayer was issued a small amount of XM88s while preparing for Operation Swamp Hopper with the USS Nan-Shan.\nIt has a special property - when you obtain multiple direct hits in a row, its armour penetration and damage will increase."
```
XSlayer300, line 365:
```suggestion
	desc = "An experimental man-portable anti-material rifle originally developed by ARMAT Battlefield Systems for the government of the state of Greater Brazil for use in the Favela Wars (2161 - Ongoing) against mechanized infantry. Chambered in .458 SOCOM, the platform features an onboard computerized targeting system, sensor array, and an electronic autoloader; these features work in tandem to reduce and render inert armor on the users target with successive hits. The Almayer was issued a small amount of XM88s while preparing for Operation Swamp Hopper with the USS Nan-Shan. It must be manually chambered for every shot.\nIt has a special property - when you obtain multiple direct hits in a row, its armour penetration and damage will increase."
```

I dunno how does an "electronic autoloader" work on a weapon that you must use the chambering button to reload.

Bram-KEP:
Fun fact, the scope does give a zoom effect, despite the scopes description saying it doesn't

Bram-KEP:
Examining it will say the follwing: This gun is levered via Unique-Action, but it has a bonus feature: Hitting a target directly will grant you a firerate and damage buff for your next shot during a short interval. Combo precision hits for massive damage.

Which is false as well.

code/modules/projectiles/guns/lever_action.dm
```diff
@@ -327,6 +347,142 @@ their unique feature is that a direct hit will buff your damage and firerate
...
+	lever_sound = lever_super_sound
+	lever_message = "<b><i>You quickly press the [lever_name]!<i><b>"
+	last_fired = world.time - buff_fire_reduc //to shoot the next round faster
+	damage_mult = initial(damage_mult) + BULLET_DAMAGE_MULT_TIER_4
+
+	if(floating_penetration < floating_penetration_upper_limit)
+		floating_penetration++
+
+	for(var/slot in attachments)
+		var/obj/item/attachable/AM = attachments["slot"]
```
Segrain, line 432:
```suggestion
		var/obj/item/attachable/AM = attachments[slot]
```
This took me embarrassingly long to notice.

code/modules/projectiles/gun.dm
```diff
@@ -1110,7 +1110,7 @@ and you're good to go.
 		var/obj/item/projectile/projectile_to_fire = load_into_chamber(user) //Load a bullet in or check for existing one.
 		if(!projectile_to_fire) //If there is nothing to fire, click.
 			click_empty(user)
-			break
+			return
```
Segrain, line 1113:
```suggestion
			flags_gun_features &= ~GUN_BURST_FIRING
			return
```
This broke guns _a lot_. It still looks wrong somehow, but at least this should stop guns freaking out dozen times per round.

Segrain (changes requested):

Geevies:
loading belt broken on fiorina (reported by discord user)
image

code/modules/projectiles/guns/lever_action.dm
stanalbatross: I imagine it automatically takes the round out of the magazine and chambers it with only a button press

but I did not write this desc so 🤷

code/modules/projectiles/guns/lever_action.dm
stanalbatross: er probably not

Mister-moon1:
image
image
these bullets come from the ammo box
image
these are the proper bullets that come with the xm88, the 45-70 version are non compatable and have the same texture, they are stored in the 458. ammo box which needs to be fixed

Mister-moon1:
The gun as of the time that the LV-624 round started at 2:05 AM BST, the gun is incapable of accepting 458. ammo types or any other form of ammo, instead it results in the standard
image
see https://streamable.com/emhuro for what the standard message is and what happens when you try to load it, taking out the ammo with the gun then attempting to load it with the same ammo also doesnt work

Mister-moon1:
after looking over messages its shown that this infact began after an event just before the round in question, one at trijent dam at 1:14 AM BST so i believe that this is a correlation to the issue

Mister-moon1:
ive been noticing that even if you do click on a sprite and it does connect, alot of the time it doesnt register as a precise shot, therefore not giving a chain or ending the current one, i think this should be replaced with successive shot rather than direct hits since it leads to the stacks system barely ever coming into play

Mister-moon1:
also stacks do not occur on PB shots

The-Sun-In-Splendour:
> 

The entire point of the gun is that if you hit a target that you direct clicked on it makes the gun stronger. That's the entire gimmick.

Mister-moon1:
Issue being that its very inconsistent, many times have i direct clicked a sprite only for no stacks and then there are times when im not even close but i get a stack anyway, even though that is the gimmick there are issues with the server, lag etc that makes using it as intended difficult

Sargeantmuffinman:
Yeah. Echoing what Moon1 said. The stacks are inconsistent. I also don't get feedback on when I lost a stack.

Apple-107:
> Yeah. Echoing what Moon1 said.

I echo this too, playtesting on the LV round at 8PM AEST, I thought my experience was a bit marred by being from Australia (and having the ping to go with it) but it's clear some others have the same issue. Direct clicking is inconsistent.

I did a small write-up on the discord and was suggested to comment here, and it looks like moon1 already has the same idea; make it for consecutive hits, rather than consecutive _direct_ hits

>The entire point of the gun is that if you hit a target that you direct clicked on it makes the gun stronger.

This would still be a present gimmick, just significantly more consistent than the direct hit mechanic. Even if you have to decrease the bonus for balancing or whatever it may be, I'd rather a gimmick that is consistent rather than never used because it isn't. You'd still have to be careful to keep the stack up!

I love this gun, I really do. Powerful one-shot damage at a slightly longer range than a shotgun would. But from my experience so far I almost ignored the gimmick because of its inconsistency.

stanalbatross:
> I also don't get feedback on when I lost a stack.

added

AGasCanister:
image
seems like you cannot get streaks with the XM88 scope

RoboticVodka:
Though it may be met with a resounding "skill issue" personally I'd suggest changing it simply to consecutive hits and not direct clicks.

Apple-107:
> Though it may be met with a resounding "skill issue"
There's a lot of people with skill issues here in this chat then.
Including me!

BKryth:
2nd round using it

It feels inconsistent as when i sprite click xeno, i check and dont see "bullseye" despite me shooting it with sprite click. Honestly id rather have it be on hit than on click since that way it would be more consistent, hell, maybe just make it lose the stacks if no successive hits after X seconds

theres also a lack of feedback, the chat and sound is just lacking on that feedback

in terms of usefulness, its so shit, the damage is so much weaker compared to other guns, its fun to use but a m4a3 might do more damage. My suggestion is increased the bullet capacity and/or give it base AP and just remove the entire gimmick thing of it doing increased damage.

stanalbatross:
Going to improve firerate and maybe stacks grace period

nauticall:
sprites are fine 👍

nauticall (approved):

stanalbatross:
> Though it may be met with a resounding "skill issue" personally I'd suggest changing it simply to consecutive hits and not direct clicks. I feel like that way it may find a lot more utility.

there was a bug causing the fire delay after direct hits to be longer than intended

fixed now

DylonBR:
I know its not a sniper but why not let us use the 4x scope?

Ticker175:
Very well balanced weapon, the bolt action thing feels a little clunky but all in all it’s a good addition. Works well as a spotter weapon

Nanu308:
Approved, albeit we'll see how it ends up performing when merged plus any issues that stem from it.

morrowwolf (approved):
Code LGTM

PR #727, opened 2022-07-22
Makes the RTO phone accessible via button rather than clicking the bag

### LINKED ISSUE #626: The USCM Radio Telephone Pack has extremely odd item removal behavior.
## Description

When items are placed in the telephone pack, if you do not click directly on the item's sprite when you intend to remove something, a different item entirely is taken from the pack. Also, if you click around the sprite of the first item(left-most in the line), nothing happens.

The telephone pack storage should behave in the same way a satchel would. With a satchel, clicking anywhere in the item's designated space removes said item from the satchel into your hand. The problem is most annoying for small or very thin objects like a crowbar since you have to pixel hunt to click on the object to avoid withdrawing something else entirely.

## Steps to reproduce

1.  Place multiple items into the telephone pack.
2.  With a free hand click on the space around an object.
3. Notice the discrepancy between how you remove objects from a telephone pack and a standard satchel.

CometBlaze:
## About The Pull Request

As it says on the tin, RTOs now access their phone by clicking a button like one would do to turn on the armour light, zoom in, etc...

## Why It's Good For The Game

Makes the radiopack behave like a regular satchel and generally less clunky to use. Also hopefully fixes #626.

## Changelog

:cl:CometBlaze
qol: Using the radiopack's phone is now done via a button similarly to turning on the armour light. Hanging up is still done by clicking the radiopack
fix: Grabbing items from the radiopack no longer requires pixel hunting
/:cl:


stanalbatross (changes requested):
image
 fix deletion of old sprite

code/game/objects/items/storage/backpack.dm
```diff
@@ -433,21 +433,22 @@ GLOBAL_LIST_EMPTY_TYPED(radio_packs, /obj/item/storage/backpack/marine/satchel/r
 		internal_transmitter.phone_id = "[user]"
 
 	internal_transmitter.enabled = TRUE
+	give_action(user, /datum/action/human_action/activable/rto_pack/use_phone)
 
 /obj/item/storage/backpack/marine/satchel/rto/dropped(mob/user)
 	. = ..()
 	internal_transmitter.phone_id = "[src]"
 	internal_transmitter.enabled = FALSE
+	remove_action(user, /datum/action/human_action/activable/rto_pack/use_phone)
```
stanalbatross, line 442: use the actions_types var on the backpack itself

stanalbatross (changes requested):

code/game/objects/items/storage/backpack.dm
```diff
@@ -433,21 +433,22 @@ GLOBAL_LIST_EMPTY_TYPED(radio_packs, /obj/item/storage/backpack/marine/satchel/r
...
 	if(user.back == src)
 		internal_transmitter.attack_hand(user)
 	else if(internal_transmitter.get_calling_phone())
-		if(internal_transmitter.attached_to && internal_transmitter.attached_to.loc != internal_transmitter)
-			return . = ..()
 		internal_transmitter.attack_hand(user)
-	else
-		. = ..()
+	for(var/datum/action/A in actions)
+		A.update_button_icon()
```
stanalbatross, line 450: why? your action doesn't use update_button_icon at all

code/modules/mob/living/carbon/human/human_abilities.dm
```diff
@@ -116,6 +116,18 @@
 /datum/action/human_action/smartpack/repair_form/cooldown_check(var/obj/item/storage/backpack/marine/smartpack/S)
 	return S.repairing
 
+/datum/action/human_action/activable/rto_pack/use_phone
+	name = "Use Phone"
+	action_icon_state = "phone"
```
stanalbatross, line 121: put this in /New() like how this does it
`/datum/action/item_action/specialist/fire_shield/New(var/mob/living/user, var/obj/item/holder)
	..()
	name = "Activate Fire Shield"
	button.name = name
	button.overlays.Cut()
	var/image/IMG = image('icons/obj/items/clothing/cm_suits.dmi', button, "pyro_armor")
	button.overlays += IMG`

also you should probably change your phone to just the phone itself (not the background) and then apply it via overlay onto the button

stanalbatross (changes requested):

stanalbatross (changes requested):
odd bug - if you spawn in an RTO via create-humans, you must take their backpack off and put it back on for this to work

not caused by code in this PR but would be nice to fix regardless

stanalbatross (approved):

PR #745, opened 2022-07-25
Pred bug fixes 

Joelampost:
## About The Pull Request
Fixed three pred bugs.

## Why It's Good For The Game
It fixes three bugs

## Changelog 

:cl: Joe Lampost
fix: Made the predator self destruct actually gib the pred and kill them instead of the fake gib that was added.
fix: Removed anchored items from showing up on the gear tracker
fix: Fixed preds being able to charge their combi sticks on simple mobs including their own smart discs
/:cl:

### DIFF
--- a/code/modules/cm_preds/yaut_bracers.dm
+++ b/code/modules/cm_preds/yaut_bracers.dm
@@ -610,7 +610,7 @@
 	if(istype(T) && exploding)
 		victim.apply_damage(50,BRUTE,"chest")
 		if(victim)
-			victim.gib() // Gibs them but does not drop the limbs so the equipment isn't dropped
+			victim.gib() // kills the pred
 			qdel(victim)
 		var/datum/cause_data/cause_data = create_cause_data("yautja self destruct", victim)
 		if(explosion_type == 0 && is_ground_level(T.z))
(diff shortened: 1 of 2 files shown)


Shad0vvs (approved):

code/modules/cm_preds/yaut_bracers.dm
```diff
@@ -406,6 +406,8 @@
 		var/atom/loc = get_true_location(I)
 		if(is_honorable_carrier(recursive_holder_check(I)))
 			continue
+		if(I.anchored)
```
Geevies, line 409: this should be above the honorable carrier check because it'll make the processing a bit faster

Geevies (changes requested):

code/modules/cm_preds/yaut_weapons.dm
```diff
@@ -475,6 +475,9 @@
 	if(target == user || target.stat == DEAD)
 		to_chat(user, SPAN_DANGER("You think you're smart?")) //very funny
 		return
+	if(target == user || isanimal(target))
+		to_chat(user, SPAN_DANGER("You think you're smart?"))
+		return
```
Geevies, line 480:
```suggestion
	if(isanimal(target))
		return
```

Geevies (changes requested):
they could have a valid reason to attack as simple mob, don't harass them over it

PR #759, opened 2022-07-27
abovehead emote improvements

stanalbatross:
## About The Pull Request

image

abovehead emotes now use small chat + a small icon, which is much more readable than italics

adds a pref to turn off seeing them if you care that much

had to default define an arg as langchat in here, idk why but it was making langchat break

abovehead emotes no longer inherit size / boldness from your default langchat size / boldness (ie for SLs)

## Why It's Good For The Game

makes abovehead emotes way easier to read and makes them able to be told apart from italic spoken words

## Changelog

:cl:
qol: abovehead emotes now use smaller chat + a small blue asterisk before them, which makes them much easier to read
qol: added a setting to turn off seeing abovehead emotes
qol: added abovehead chat settings to the setup settings section
fix: abovehead emotes no longer inherit size and boldness from their user
/:cl:


Agameofscones (approved):
sprites lgtm

code/modules/mob/emote.dm
```diff
@@ -52,7 +52,9 @@
 						M.show_message(message, m_type)
 						viewers.Add(M)
 				O.show_message(message, m_type)
-			langchat_speech(input, viewers, GLOB.all_languages, skip_language_check = TRUE, additional_styles = list("langchat_italic"))
+				if(!(O?.client?.prefs.toggles_langchat & LANGCHAT_SEE_EMOTES))
```
Watermelon914, line 55: O is definitely always valid here, so remove the ?. after O.
Also this will probably runtime if you try to do a bit operation with a null value, so consider storing it as a variable first and checking the validity of the variable before doing the bit operation

code/modules/mob/emote.dm
```diff
@@ -72,7 +74,9 @@
 						M.show_message(message, m_type)
 						hearers.Add(M)
 				O.show_message(message, m_type)
-			langchat_speech(input, hearers, GLOB.all_languages, skip_language_check = TRUE, additional_styles = list("langchat_italic"))
+				if(!(O?.client?.prefs.toggles_langchat & LANGCHAT_SEE_EMOTES))
```
Watermelon914, line 77: same thing as above here

code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -308,6 +312,10 @@
 					for(var/mob/M in heard)
 						if(M.ear_deaf)
 							heard -= M
+							continue
+						if(!(M?.client?.prefs.toggles_langchat & LANGCHAT_SEE_EMOTES))
```
Watermelon914, line 316: same as above (why are there so many of these?)

code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -359,6 +367,10 @@
...
```
Watermelon914, line 371: same as above

Watermelon914 (changes requested):

PR #792, opened 2022-08-01
AGM-Smoke grenades

stanalbatross:
## About The Pull Request

adds smoke AGMs

## Why It's Good For The Game

more opportunities for clever smoke nade use

## Changelog

:cl:
add: added AGM-S smoke grenades, they work like smoke grenades but in AGM form, and have a smaller radius. 
/:cl:


AntoineXYZ:
I see radius is 2, is it 2 tiles on each + a center one? If so, isn't that still a tad too big?
I'd think a grenade like would be perfect "choke" smoker without actually choking the whole front line, something like a 3x3 cross?

Agameofscones (approved):
sprites good

morrowwolf (approved):
LGTM

PR #803, opened 2022-08-03
Adds vehicles to follow options

JeserTheFox:
## About The Pull Request

Adds vehicles to follow options.

## Why It's Good For The Game

Fixed missing options since folloew/orbit rework.

## Changelog

:cl: Jeser
add: Vehicles appear in Follow window now and can be followed from there.
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/Orbit.tsx
+++ b/tgui/packages/tgui/interfaces/Orbit.tsx
@@ -53,6 +53,7 @@ type OrbitData = {
   ghosts: OrbitList[],
   misc: OrbitList[],
   npcs: OrbitList[],
+  vehicles: OrbitList[],
 }
 
 type BasicSectionProps = {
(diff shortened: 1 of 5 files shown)


DylonBR:
another xeno metagaming buff smh

tgui/packages/tgui/interfaces/Orbit.tsx
```diff
@@ -287,6 +289,13 @@ export const Orbit = (props: any, context: any) => {
           source={misc}
           searchText={searchText}
         />
+
+        <BasicSection
+          title="Vehicles"
+          source={vehicles}
+          searchText={searchText}
+        />
+
```
stanalbatross, line 298: 1. move this under humans since it's more relevant

2. replace with

```suggestion

         <Section title={`Vehicles - (${vehicles.length})`}>
          {vehicles
            .filter(searchFor(searchText))
            .sort(compareNumberedText)
            .map(thing => (
              <OrbitedButton
                key={thing.name}
                color="blue"
                thing={thing} />
            ))}
        </Section>
```
this code should work I apologise if it does not

this will make it so numbers display properly etc

stanalbatross (changes requested):

stanalbatross (approved):

PR #817, opened 2022-08-06
High-Capacity Laser Battery Point Increase

Geevies:
## About The Pull Request

Increases the cost of HC Laser Batteries for CAS to 200 (up from 100).

## Why It's Good For The Game

Brings the cost of the ammo in line with other CAS ammunition, and makes other CAS weaponry more viable by doing so.

## Changelog

:cl:
balance: Increases the cost of HC Laser Batteries for CAS to 200 (up from 100).
/:cl:

### DIFF
--- a/code/modules/cm_marines/dropship_ammo.dm
+++ b/code/modules/cm_marines/dropship_ammo.dm
@@ -201,7 +201,7 @@
 	ammo_used_per_firing = 10
 	max_inaccuracy = 1
 	warning_sound = 'sound/effects/nightvision.ogg'
-	point_cost = 100
+	point_cost = 250
 	fire_mission_delay = 4 //very good but long cooldown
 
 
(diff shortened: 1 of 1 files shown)


FoxyStalin:
'Makes other Weapons more viable'  as if Lasers isn't the second worst option other than GAU (Until the GAU Buff gets in)

c4xmaniac:
Laser killer PR, time of the Xwing is over. It was only viable due to how much you could use because it’s not particularly strong.

code/modules/cm_marines/dropship_ammo.dm
```diff
@@ -201,7 +201,7 @@
 	ammo_used_per_firing = 10
 	max_inaccuracy = 1
 	warning_sound = 'sound/effects/nightvision.ogg'
-	point_cost = 100
+	point_cost = 250
```
stanalbatross, line 204:
```suggestion
	point_cost = 200
```
doesn't need to be 250

stanalbatross (changes requested):

stanalbatross (approved):

Weuyn:
The laser having such cheap ammo is a relic of the time the hangar had a bunch of ammo depots for every other weapon type lying around.

PR #821, opened 2022-08-07
bodyscanner no longer sends you to the nighmare dimension

### LINKED ISSUE #790: Body Scanner Causing Players to Disappear
## Description
When multiple bodies are placed in a body scanner simultaneously, one of them vanishes into the ether. The mob is still locally placed inside the body scanner, and can interact with their inventory, such as using their radio or printing from a recorder, but cannot use local chat.

Only tested on marines, was reproduced multiple times shortly prior to the hijack that took place during Sortie Falling-Scalpel which began at approximately 4:18 PM, CST, on 7/31/2022. First marine to be disappeared this way during the round was CMP Anthony Hammonds.

A body scanner should either not allow multiple bodies, or not banish one of the bodies shoved into it into the ether.

## Steps to reproduce
1. Have two marines grab two other marines adjacent to a body scanner.
2. Place both other marines in the body scanner simultaneously.
3. Open the body scanner.
4. Only one body will emerge, the other will be gone forever.

Diegoflores31:
## About The Pull Request

PR adds a check after you try to get inside of the bodyscanner so you cant enter while its occupied . sending the previous user into the nightmare realm

## Why It's Good For The Game

Closes #790 

## Changelog

:cl:
fix: Bodyscanner no longer dissapears marines if they hop inside at the same time.
/:cl:

### DIFF
--- a/code/game/machinery/adv_med.dm
+++ b/code/game/machinery/adv_med.dm
@@ -78,6 +78,9 @@
 	if(isXeno(M))
 		return
 	if(do_after(usr, 10, INTERRUPT_NO_NEEDHAND, BUSY_ICON_GENERIC))
+		if (src.occupant)
+			to_chat(usr, SPAN_NOTICE(" <B>The scanner is already occupied!</B>"))
+			return
 		to_chat(usr, SPAN_NOTICE("You move [M.name] inside \the [src]."))
 		M.forceMove(src)
 		occupant = M
--- a/code/game/machinery/adv_med.dm
+++ b/code/game/machinery/adv_med.dm
@@ -65,10 +65,10 @@
 	if (usr.stat || !(ishuman(usr)))
 		return
 	if (src.occupant)
-		to_chat(usr, SPAN_NOTICE(" <B>The scanner is already occupied!</B>"))
+		to_chat(usr, SPAN_BOLDNOTICE("The scanner is already occupied!"))
 		return
 	if (usr.abiotic())
-		to_chat(usr, SPAN_NOTICE(" <B>Subject cannot have abiotic items on.</B>"))
+		to_chat(usr, SPAN_BOLDNOTICE("Subject cannot have abiotic items on."))
 		return
 	go_in_bodyscanner(usr)
 	add_fingerprint(usr)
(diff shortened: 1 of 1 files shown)


code/game/machinery/adv_med.dm
```diff
@@ -78,6 +78,9 @@
 	if(isXeno(M))
 		return
 	if(do_after(usr, 10, INTERRUPT_NO_NEEDHAND, BUSY_ICON_GENERIC))
+		if (src.occupant)
```
JeserTheFox, line 81:
```suggestion
		if (occupant)
```

code/game/machinery/adv_med.dm
```diff
@@ -78,6 +78,9 @@
...
+			to_chat(usr, SPAN_NOTICE(" <B>The scanner is already occupied!</B>"))
```
spookydonut, line 82: dont think you're meant to use <b>

Diegoflores31:
suggestions applied

code/game/machinery/adv_med.dm
```diff
@@ -78,6 +78,9 @@
 	if(isXeno(M))
 		return
 	if(do_after(usr, 10, INTERRUPT_NO_NEEDHAND, BUSY_ICON_GENERIC))
+		if(occupant)
+			to_chat(usr, SPAN_NOTICE(" <B>The scanner is already occupied!</B>"))
```
stanalbatross, line 82:
```suggestion
			to_chat(usr, SPAN_NOTICE("<B>The scanner is already occupied!</B>"))
```
use BOLDNOTICE

stanalbatross (changes requested):

code/game/machinery/adv_med.dm
```diff
@@ -78,6 +78,9 @@
 	if(isXeno(M))
 		return
 	if(do_after(usr, 10, INTERRUPT_NO_NEEDHAND, BUSY_ICON_GENERIC))
+		if(occupant)
+			to_chat(usr, SPAN_NOTICE("<B>The scanner is already occupied!</B>"))
```
stanalbatross, line 82:
```suggestion
			to_chat(usr, SPAN_BOLDNOTICE("The scanner is already occupied!"))
```

PR #822, opened 2022-08-08
Surgery table no longer removes huggers

### LINKED ISSUE #804: You can remove a facehugger on a person if you connect them to an operating table
## Description

As per title.

Well it shouldn't happen.

## Steps to reproduce

Get a facehugged patient and connect them to the surgery table.
The facehugger will pop off the patient but will still be alive.

The patient does get their face scrawled though and it might cause some internal bleeding. 

https://user-images.githubusercontent.com/104228030/182676987-e4adc8dc-a561-4960-8538-779af56fd518.mp4

Diegoflores31:
## About The Pull Request

bug bad >:)

## Why It's Good For The Game

Closes #804 

## Changelog

:cl:
fix: Surgery anestetic system no longer removes Facehuggers and other unremovable masks
/:cl:

### DIFF
--- a/code/game/machinery/OpTable.dm
+++ b/code/game/machinery/OpTable.dm
@@ -103,10 +103,8 @@
 	if(H.wear_mask)
 		var/obj/item/mask = H.wear_mask
 		if(mask.flags_inventory & CANTSTRIP)
-			var/obj/item/clothing/mask/facehugger/hugger = mask
-			if(hugger.stat != DEAD)
-				to_chat(user, SPAN_DANGER("You can't remove their mask!"))
-				return
+			to_chat(user, SPAN_DANGER("You can't remove their mask!"))
+			return
 		H.drop_inv_item_on_ground(mask)
 	var/obj/item/clothing/mask/breath/medical/B = new()
 	if(!H.equip_if_possible(B, WEAR_FACE, TRUE))
--- a/code/game/machinery/OpTable.dm
+++ b/code/game/machinery/OpTable.dm
@@ -103,8 +103,10 @@
 	if(H.wear_mask)
 		var/obj/item/W = H.wear_mask
 		if(W.flags_inventory & CANTSTRIP)
-			to_chat(user, SPAN_DANGER("You can't remove their mask!"))
-			return
+			var/obj/item/clothing/mask/facehugger/hugger = W
+			if(hugger.stat != DEAD)
+				to_chat(user, SPAN_DANGER("You can't remove their mask!"))
+				return
 		H.drop_inv_item_on_ground(H.wear_mask)
 	var/obj/item/clothing/mask/breath/medical/B = new()
 	if(!H.equip_if_possible(B, WEAR_FACE, TRUE))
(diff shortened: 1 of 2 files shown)


Diegoflores31:
DO NO MERGE THIS I DID AN OPPSIE

code/game/machinery/OpTable.dm
```diff
@@ -100,7 +100,7 @@
 	if(!anes_tank)
 		to_chat(user, SPAN_WARNING("There is no anesthetic tank connected to the table, load one first."))
 		return
-	if(H.wear_mask && !H.drop_inv_item_on_ground(H.wear_mask))
+	if(H.wear_mask && NODROP)
```
morrowwolf: This is just checking a random bit flag. It needs to check if NODROP is active in the mask's flags

morrowwolf (changes requested):

code/game/machinery/OpTable.dm
Diegoflores31: outdated version >:)

code/game/machinery/OpTable.dm
```diff
@@ -100,10 +100,14 @@
 	if(!anes_tank)
 		to_chat(user, SPAN_WARNING("There is no anesthetic tank connected to the table, load one first."))
 		return
-	if(H.wear_mask && !H.drop_inv_item_on_ground(H.wear_mask))
-		to_chat(user, SPAN_DANGER("You can't remove their mask!"))
-		return
-
+	if(H.wear_mask)
+		var/obj/item/W = H.wear_mask
```
morrowwolf, line 104: Name this variable better

code/game/machinery/OpTable.dm
```diff
@@ -100,10 +100,14 @@
...
+		if(W.flags_inventory & CANTSTRIP)
+			var/obj/item/clothing/mask/facehugger/hugger = W
```
morrowwolf, line 106: There should be an explicit check for if the mask is a facehugger before assigning it. While CANTSTRIP has a comment saying it's only used for facehuggers this may change in the future and there are items with CANTSTRIP already.

code/game/machinery/OpTable.dm
```diff
@@ -100,10 +100,14 @@
...
+			if(hugger.stat != DEAD)
+				to_chat(user, SPAN_DANGER("You can't remove their mask!"))
+				return
+		H.drop_inv_item_on_ground(H.wear_mask)
```
morrowwolf, line 110: Use the above assigned variable as the argument here.

morrowwolf (changes requested):

code/game/machinery/OpTable.dm
Diegoflores31: if those items cannot be stripped by normal means they should not be able to be stripped by the surgery table anyways so. think of it as a double bugfix

Diegoflores31: the reason why i used the CANSTRIP instead of (ISTYPE) is because in the future some mask with that flag may be added and instead of having to make an exception for every single one we better just do this

morrowwolf: This needs to be rethought if it's going to be generalized.

spookydonut (approved):

Diegoflores31:
CANSTRIP flag check for the surgery table was kept and the variable for CANTSTRIP was modified on huggers so they only maintain it while the hugger is alive

PR #824, opened 2022-08-08
supply drop console tgui

stanalbatross:
## About The Pull Request

image

code was partially ported from tgmc, due to differences between their supply drop console and ours a few things had to be changed

makes it use tgui

slight refactor to some parts of the code

image

## Why It's Good For The Game

no more shit ui

## Changelog

:cl:
ui: converted the supply drop console to TGUI
fix: mortar interface numberinputs now work properly
/:cl:


code/game/supplyshuttle.dm
```diff
@@ -125,99 +125,112 @@ var/datum/controller/supply/supply_controller = new()
...
-		return
-	if(!href_list["operation"])
+/obj/structure/machinery/computer/supply_drop_console/ui_data(mob/user)
+	. = ..()
+	check_pad()
+	.["busy"] = busy
+	.["can_pick_squad"] = can_pick_squad
+	.["current_squad"] = current_squad
+	.["launch_cooldown"] = drop_cooldown
+	.["next_fire"] = COOLDOWN_TIMELEFT(src, next_fire)
```
spookydonut, line 157: same deal here as other PR, send expiry time not remaining duration

code/game/supplyshuttle.dm
```diff
@@ -125,99 +125,112 @@ var/datum/controller/supply/supply_controller = new()
...
+			var/new_x = text2num(params["set_x"])
+			if(!new_x)
+				return
+			x_supply = new_x
+
+		if("set_y")
+			var/new_y = text2num(params["set_y"])
+			if(!new_y)
+				return
+			y_supply = new_y
```
spookydonut, line 180: do some sanity validation here for positive and in bounds, though just 0 <= x < 255 is probably fine

code/game/supplyshuttle.dm
```diff
@@ -125,99 +125,112 @@ var/datum/controller/supply/supply_controller = new()
...
+
 		if("pick_squad")
 			if(can_pick_squad)
 				var/list/squad_list = list()
 				for(var/datum/squad/S in RoleAuthority.squads)
 					if(S.active && S.faction == faction)
 						squad_list += S.name
 
-				var/name_sel = tgui_input_list(usr, "Which squad would you like to claim for Overwatch?", "Overwatch", squad_list)
+				var/name_sel = tgui_input_list(usr, "Which squad pad would you like to use to send supplies?", "Supply drop console", squad_list)
```
spookydonut, line 189: do this inside the tgui ;)

spookydonut (changes requested):

code/game/supplyshuttle.dm
stanalbatross: done

code/game/supplyshuttle.dm
stanalbatross: not done - the values input here are obfuscated so they can be negative and out of bounds with relation to real coords

actual sanitisation takes place in the launch proc

	var/x_coord = deobfuscate_x(x_supply)
	var/y_coord = deobfuscate_y(y_supply)
	var/z_coord = SSmapping.levels_by_trait(ZTRAIT_GROUND)
	if(length(z_coord))
		z_coord = z_coord[1]
	else
		z_coord = 1 // fuck it

	var/turf/T = locate(x_coord, y_coord, z_coord)
	if(!T)
		to_chat(usr, "[icon2html(src, usr)] [SPAN_WARNING("Error, invalid coordinates.")]")
		return

as such I want to keep it the same as the old system and not sanitise values here, so that if invalid coords are given then req can't tell.

morrowwolf (changes requested):
operable check

stanalbatross:
> operable check

was already in for some reason

spookydonut (approved):

PR #834, opened 2022-08-10
VP78 Buffs (Stan said I could)

spartanbobby:
## About The Pull Request

Okay, so I buffed the magazine count of the VP78 to be 18 like it is in lore, I reduced its falloff to be 5 (one more than the MK2), gave it a premounted lasersight and upped its PFC vendor cost from 10 to 15 as stan requested

"It is fitted with a laser sight/flashlight unit underneath the barrel as standard, although it is not usable in the game; instead, [Rookie](https://avp.fandom.com/wiki/Rookie) (or multiplayer Marines) uses his [TNR Shoulder Lamp](https://avp.fandom.com/wiki/TNR_Shoulder_Lamp) for illumination, as with any other weapon."

Feed system: 18-round detachable box magazine

https://avp.fandom.com/wiki/VP78_Pistol

## Why It's Good For The Game

Gives the VP alittle bit of a boost when compared to the other objectively better and more readily available sidearms 

## Changelog

:cl:SpartanBobby
balance: VP78 Now costs 15 points in PFC vendor
balance: VP78 Now has the lore accurate 18 magazine capacity instead of 14
balance: VP78 Now has the lore-accurate pre-attached laser sight
balance: VP78 Now has falloff 5 from falloff 7, one more than the MK2
/:cl:

### DIFF
--- a/code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
+++ b/code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
@@ -138,7 +138,7 @@ GLOBAL_LIST_INIT(cm_vending_clothing_marine, list(
 		list("Rebreather", 0, /obj/item/clothing/mask/rebreather, MARINE_CAN_BUY_MASK, VENDOR_ITEM_REGULAR),
 
 		list("RESTRICTED FIREARMS", 0, null, null, null),
-		list("VP78 Pistol", 10, /obj/item/storage/box/guncase/vp78, null, VENDOR_ITEM_REGULAR),
+		list("VP78 Pistol", 15, /obj/item/storage/box/guncase/vp78, null, VENDOR_ITEM_REGULAR),
 		list("SU-6 Smart Pistol", 15, /obj/item/storage/box/guncase/smartpistol, null, VENDOR_ITEM_REGULAR),
 		list("M41AE2 Heavy Pulse Rifle", 30, /obj/item/storage/box/guncase/lmg, null, VENDOR_ITEM_REGULAR),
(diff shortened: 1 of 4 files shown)


Don2137:
I'm all for VP buffs but this doesn't really fix how woefully bad the gun is. It needs either more damage per bullet or a faster fire rate to be even remotely viable, nevermind good (especially when 88 mod exists). Right now even the smartpistol is better.

Sargeantmuffinman:
Yeah you're gonna need to bump the fire rate to something faster.
Not as fast as the M4A3 god no.
But bumping the fire rate to 150 RPM on single fire would definitely help.

Diegoflores31:
Yeah fire rate needs to be at least faster than revolver

MusselMain:
> Yeah fire rate needs to be at least faster than revolver

Two words: Burst Fire, I've used the VP on burst constantly and it gets the job done.

Don2137:
> > Yeah fire rate needs to be at least faster than revolver
> 
> Two words: Burst Fire, I've used the VP on burst constantly and it gets the job done.

Still slower. 124 rpm on VP burst, 150 on revolver. Less damage per bullet too.

Diegoflores31:
Yeah thats what i was saying .. i understand its a "heavy" pistol but i think beign a restricted it should have better stats than a m44 lmao
Buest fire doesnt makes it good . Mod88 or m4a3 overperform this weapon in every single aspect

Diegoflores31:
Also if you ever played before any avp with the VP you should be able to see that the firerate is Way better than here in cm

GoldenDarkness55:
Nice. I agree with others that it should get +1 to RoF tier on top of this

Issue #708, opened 2022-07-18
Fix Trijent Dam water, specifically the issues with the "Truck"  [Suggestion]


Zenith00000:
## Summary
I've been playing CM for years now, Survivor main. I've never really had issues with the game mechanics and such, but I've been recently banned because of "Rule 5 - Mechanical Exploits" due to driving the Truck on Trijent Dam into the "acid water". Now I know what you're thinking, "Just don't drive the Truck into the water" and yes that would be the obvious thing to do. However this isn't just about me specifically. 

Simply having vehicles take damage when driving over the water would effectively fix this issue and prevent anyone having the same troubles as I've had with salty Xenos and such.

_Also this is my first time using this website, filing a Bug report or such, I'm not even sure if this is the right place to post it, as the forums just redirect me to another page. Please forgive me if this isn't correct._

## Details

My suggestion is really just to save the problems with other Survivor players doing the same thing and being warned/banned over something that shouldn't even be a problem in the first place. It's been years since Trijent Dam has been a thing and the problem with the water not damaging the APC, Longstreet, Trucks or any vehicles has always been a problem. I understand if it's a difficult or seemingly insignificant thing to implement, but it would be much appreciated by Survivor players and the Xeno players even more so. 

## Implementation

I'm a total smooth brain when it comes to these things, and I've no idea how you'd implement these things, but if the acid water works the same way as let's say lava does on the other SS13 servers, then it should hopefully not be too difficult to make it damage vehicles or such, but like I've said I've no idea how this shit works and I'm just a smooth brain consumer.

JeserTheFox:
water not damaging tank and APC is intended, as Tank is, well, tank, and APC is designed to transport marines over dangerous areas. As for trucks, gonna make them receive wheel damage over toxic water, easily done.

Zenith00000:
> water not damaging tank and APC is intended, as Tank is, well, tank, and APC is designed to transport marines over dangerous areas. As for trucks, gonna make them receive wheel damage over toxic water, easily done.

Thanks a lot man. This will only be a minor thing but prevent some problems in the future and such. Appreciate it.

PR #835, opened 2022-08-10
Vehicle environmental damage

JeserTheFox:
## About The Pull Request

1. Reworked vehicle locomotion module acid damage handling to be better and part of vehicle itself. Added acid damage sound to it.
2. Toxic trijent river now deals damage to civillian vehicles. Cleaned up code a bit
3. Some fixes in vehicle spawners. Fix for vehicle speed speed.

dreamseeker_Gisoy3VXOt

Fixes [!708](https://github.com/cmss13-devs/cmss13/issues/708)

## Why It's Good For The Game

Prevents using vans to succesfully avoid xenos in water, while leaving opportunity to drive on other side of broken bridge to ditch pursuit at the cost of damaged/losing wheels.
Fixes some bugs.

## Changelog

:cl: Jeser
balance: Vehicles now can receive damage from various environmental acidic sources like toxin rivers in addition to acid spray. Colony Vans receive wheels damage when driving through toxins river on Trijent Dam now.
fix: Fixed some vehicle spawners behaviour and vehicle speed.
/:cl:

### DIFF
--- a/code/modules/vehicles/hardpoints/wheels/locomotion.dm
+++ b/code/modules/vehicles/hardpoints/wheels/locomotion.dm
@@ -32,17 +32,32 @@
 /obj/item/hardpoint/locomotion/on_uninstall(var/obj/vehicle/multitile/V)
 	deactivate()
 
-//unique proc for locomotion modules, taking damage from acid spray on ground
-/obj/item/hardpoint/locomotion/proc/handle_acid_spray(var/obj/effect/xenomorph/spray/acid)
-	var/take_damage = acid.damage_amount
-	//First we check source of acid. Due to traps generating 3x3 acid spray field and triggering only when at least 4 tiles
-	//of vehicle enter the spray spawn area, it deals a huge amount of damage. But simply nerfing damage will also nerf it for
-	//acid spraying castes like spitters and praetorians, which is not ideal.
-	if(acid.cause_data.cause_name == "resin acid trap")
-		take_damage = round(take_damage / 3)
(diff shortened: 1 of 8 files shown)


MusselMain:
How strong will this affect be? I'd rather not be driving a decently damaged van across a half destroyed bridge and then have the wheels die on me, I see how this can obviously be abused, but if it's too strong it'll prevent anyone from passing the river via vehicle.

JeserTheFox:
> How strong will this affect be? I'd rather not be driving a decently damaged van across a half destroyed bridge and then have the wheels die on me, I see how this can obviously be abused, but if it's too strong it'll prevent anyone from passing the river via vehicle.

I literally attached a gif to show the effect...

code/modules/vehicles/hardpoints/wheels/locomotion.dm
```diff
@@ -32,17 +32,32 @@
...
+	else if(istype(A, /obj/effect/blocker/toxic_water))
+		//multitile vehicles are, well, multitile and will be receiving damage for each tile, so damage is low per tile.
+		take_damage = 10
+
 	//then we check whether this locomotion module is acid-resistant
 	if(acid_resistant)
 		take_damage = take_damage / 2
 	health -= take_damage
+
+	if(!(world.time % 3))
```
stanalbatross, line 59: what does this actually do? never seen a % operator like that before

stanalbatross (approved):

PR #842, opened 2022-08-11
Jeser/spec gear box fix

JeserTheFox:
## About The Pull Request

1. Moved both spec kits and sets lists to global_lists.dm
2. Spec_kit boxes now support allowing any specific or any at all roles to activate them by editing allowed_roles_list in spec_kit's code or in VVs.
3. Upon examining, spec_kit boxes now show user all roles that can activate examined spec_kit box.
4. Fixed several exploits with them.
5. Fixed ID assignment.

изображение
изображение

## Why It's Good For The Game

Qol, bug and exploit fixes.

## Changelog

:cl: Jeser
add: Specialist kit (the one that allows you to select a specialization and receive a respecting equipment box) now supports specifying what roles are able to activate it or removing restrictions at all by admins for events.
qol: Specialist kit shows what roles exactly can use it upon examination now.
fix: Fixed a couple of bugs and exploits related to specialist kit.
imageadd: Specialist kit has it's own sprite now!
/:cl:


code/modules/cm_marines/equipment/kit_boxes.dm
```diff
@@ -148,52 +148,72 @@
...
 			qdel(src)
-	else
-		to_chat(user, SPAN_NOTICE("This box is not for you, give it to a squad rifleman!"))
+			return
 
-/obj/item/spec_kit/proc/select_and_spawn(mob/living/carbon/human/user)
-	if(!istype(user))
-		to_chat(user, SPAN_WARNING("You can't use \the [src]!"))
-		return
+	for(var/allowed_role in allowed_roles_list)
```
Geevies, line 192: this should be made into /proc/can_use(var/mob/living/carbon/human/user) so you don't have to do select_and_spawn above and below here, something like

/proc/can_use(var/mob/living/carbon/human/user)
   if(!length(allowed_roles_list))
       return TRUE
   loop through list
      matches 
          return TRUE
   return FALSE

Geevies (changes requested):

code/modules/cm_marines/equipment/kit_boxes.dm
JeserTheFox: Done

PR #848, opened 2022-08-13
Hardy weeds nerf

stanalbatross:
## About The Pull Request

slight rebalance + added code for stopping construction of all structures

## Why It's Good For The Game

hardy weeds circumvent the desires of map designers and allow for the weeding of intended unweedable areas such as LV LZ direct flanks

this strikes a compromise by letting them spread but not be planted on unweedable surfaces

as such this ensures that mappers' intentions in map design cannot be circumvented by them

## Changelog

:cl:
balance: hardy weeds can no longer be planted in unweedable areas. They will still spread into not normally weedable areas, but cannot be planted there.
/:cl:

### DIFF
--- a/code/modules/cm_aliens/weeds.dm
+++ b/code/modules/cm_aliens/weeds.dm
@@ -1,3 +1,7 @@
+#define BLOCK_NOTHING 				0
+#define BLOCK_SPECIAL_STRUCTURES	1
+#define BLOCK_ALL_STRUCTURES		2
+
 /obj/effect/alien/weeds
 	name = "weeds"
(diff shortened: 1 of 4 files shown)


Carnage068:
Justified, tbh. The same thoughts were running through my head, but I couldn't personally think of a way they could be balanced without removing them from the game. Big brain PR lol

Geevies (approved):

Weuyn:
could you make them delete regular weeds when spreading, very annoying to put them anywhere already weeded. Have to exploit with resin surge to delete regular weed nodes then plant

PR #855, opened 2022-08-14
CL Trackers and Tracking Tweaks

Geevies:
## About The Pull Request

Slightly refactors how headset tracker settings are managed, it's now a list on the headset rather than a proc override on each one. You can now add tracker options to encryption headsets and it'll be applied to the headset upon channel recalculation.

WY-encryption keys, like the ones handed out by the Corporate Liaison, now enable CL tracking. Additionally, all non-HvH PMC headsets now automatically track the CL.

Headsets can now only track personnel wearing uniforms, with their sensors set to the Location/Maximum setting.

## Why It's Good For The Game

Headset tracking has always been a pain, so hopefully this unified system solves it forever. As for the CL tracking, it's just kinda neat I guess. It's useful for PMCs since securing the CL is one of their main objectives.

As for tracking only if the uniform's sensors is maxed out, I considered that maybe someone doesn't want to be tracked, or perhaps a hostage taker doesn't want that someone to be tracked.

## Changelog

:cl:
add: Headset tracking HUDs now bring up a list of options rather than cycling through them.
add: The WY encryption key now enables you to track the CL using the SL tracker. Non-HvH PMC headsets automatically track the CL.
balance: You can now only track someone wearing a uniform with their sensors set to Location/Maximum.
/:cl:


c4xmaniac:
I’m sure the PMCs will head back once they realize via their tracker that the CL isn’t on the ship :D

morrowwolf (changes requested):
When someone takes a key out it does not stop them from continuing to have the tracker that should be lost.

A headset without misc_tracking enabled will not gain a tracker from a key adding a tracker.

Geevies:
> When someone takes a key out it does not stop them from continuing to have the tracker that should be lost.
> 
> A headset without misc_tracking enabled will not gain a tracker from a key adding a tracker.

fixed the first one, i don't think it's necessary to change the second one in this PR, because it probably needs a headset wide refactor for that

code/game/objects/items/devices/radio/headset.dm
```diff
@@ -164,20 +171,29 @@
...
 		if(key.translate_hive)
 			translate_hive = TRUE
 		if(key.syndie)
 			syndie = TRUE
 
+	var/list/tracking_stuff = list()
+	for(var/thing in tracking_options)
+		tracking_stuff += tracking_options[thing]
+	if(!(locate_setting in tracking_stuff))
+		locate_setting = tracking_stuff[1]
```
morrowwolf, line 195: If I'm reading this correctly I believe the following:
If a headset has only the tracker from a key and that key is removed the locate_setting will not be in tracking_stuff and tracking_stuff will be empty and throw a runtime.

Probably needs a check for if there are any tracking options. Also, these variable names could probably be better :eyes:

Let me know if I'm misreading.

morrowwolf (changes requested):

nauticall (approved):
sprites look good!

morrowwolf (approved):
Code LGTM

PR #862, opened 2022-08-14
Improvements for the OOC Country Flags system.

Kaga-404:
## About The Pull Request

As it says on the tin, an attempt to fix the runtimes from Country Flags and make it retry on failure instead of leaving some people without flags indefinitely. Maybe in the future we can make it request when they OOC chat without a flag but that's a whole different matter.

## Why It's Good For The Game

Less runtimes the better.

## Changelog

:cl: Kaga
fix: OOC Country Flags should now retry failed people after a minute, as well as be less likely to runtime.
/:cl:

### DIFF
--- a/code/modules/client/country_flags.dm
+++ b/code/modules/client/country_flags.dm
@@ -4,7 +4,7 @@
 		spawn if(src)
 			src.country = ip2country(address, src)
 
-/proc/ip2country(ipaddr, client/origin, var/failed = 0)
+/proc/ip2country(ipaddr, client/origin)
 	if(!origin || origin?.country)
 		return //null source, or already has a flag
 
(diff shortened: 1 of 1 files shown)


code/modules/client/country_flags.dm
```diff
@@ -2,20 +2,29 @@
 	. = ..()
 	if(CONFIG_GET(flag/ooc_country_flags))
 		spawn if(src)
-			src.country = ip2country(address)
+			src.country = ip2country(address, src)
+
+/proc/ip2country(ipaddr, client/origin, var/failed = 0)
```
stanalbatross, line 7: what's the point of failed

code/modules/client/country_flags.dm
```diff
@@ -2,20 +2,29 @@
 	. = ..()
 	if(CONFIG_GET(flag/ooc_country_flags))
 		spawn if(src)
-			src.country = ip2country(address)
+			src.country = ip2country(address, src)
```
stanalbatross, line 5: if origin is having its country modified in the proc then this src.country is probably not needed

stanalbatross (changes requested):

code/modules/client/country_flags.dm
stanalbatross: it's returning a big list so it probably will save on load

stanalbatross (approved):

PR #866, opened 2022-08-15
Nerfs the AGMF

Nanu308:
## About The Pull Request
Nerfs the AGMF from 400+dmg on avg with a maxcap of 750 to around 200 on avg with a maxcap of 400dmg.

## Why It's Good For The Game
Balance because being 1shot by a tool that is easily accessible and can be used by every marine if purchased is unfun.

## Changelog

:cl: Nanu

balance: Nerfs the AGMF from 400+dmg on avg with a maxcap of 750 down to around 200 dmg on avg with a maxcap of 400dmg via turning the shrapnel count down to 16 from 30.

/:cl:

### DIFF
--- a/code/game/objects/items/explosives/grenades/marines.dm
+++ b/code/game/objects/items/explosives/grenades/marines.dm
@@ -138,7 +138,7 @@
 	item_state = "grenade_m74_airburst_f_active"
 	explosion_power = 0
 	explosion_falloff = 25
-	shrapnel_count = 12
+	shrapnel_count = 16
 	det_time = 0 // Unused, because we don't use prime.
 	hand_throwable = FALSE
 	falloff_mode = EXPLOSION_FALLOFF_SHAPE_LINEAR
--- a/code/game/objects/items/explosives/grenades/marines.dm
+++ b/code/game/objects/items/explosives/grenades/marines.dm
@@ -138,7 +138,7 @@
 	item_state = "grenade_m74_airburst_f_active"
 	explosion_power = 0
 	explosion_falloff = 25
-	shrapnel_count = 30
+	shrapnel_count = 12
 	det_time = 0 // Unused, because we don't use prime.
 	hand_throwable = FALSE
 	falloff_mode = EXPLOSION_FALLOFF_SHAPE_LINEAR


stanalbatross (approved):

c4xmaniac:
Epic change AGMFs were cancer

Sernoh:
I agree that AGM-Fs deserve a damage nerf, but this is the type of nerf that delete their usage from the game.
AGMs require a hard read, direct line of sight, and you have nine to start with over the course of the round.

With 140-180 damage, why would you use them over a Mou53 (which outputs 240-360), or spam clicking the M41 for one second and dealing 165-220 ?

The problem with AGM-Fs is their RNG and dealing upwards of 750 damage in lucky situations. That aspect of AGMs should be nerfed, not AGM-Fs as a whole.

stanalbatross:
might make agms a bit too useless imo...

Alardunx:
AGM-F throw 30 fragments in a 60* degree, or in other words, 1 fragment per 2 degree (On average, the create_shrapnels create them in a random position, so it's technically possible for all of them to spawn at the same angle)

On average, though, that tends to be a 8/14/8 shrapnel spread (left tile, center tile, right tile), at 17-33 damage (25+/-8) per shrapnels, or in other word

136-264 / 238-462 / 136-264, presuming the hit is directly in front of the Xenomorph (Only the middle value is actually relevant here)

Damage decrease, roughly, by 50% per tile missed, and a direct impact is a guaranteed 5 shrapnels and no more.

However, there is currently a bug that afflicts Shrapnels spawning and has been around for a while (And is also the reason for HEFA insta-gibbing people), and it is the fact that Shrapnel spawning codes check for entities on the tile when spawning to ignore them for the guaranteed collision, but not for entities that are currently traveling to the tile where they are spawning.

So if you're moving to a tile while the AGM-F is detonating, you're eligible for all 30 shrapnels that are checking for collision, rather than the expected 5 or up to 14, and that is what's gibbing people.

While a reduction in shrapnels might be viable, a drop to 12 reduces the center hit from 14 shrapnels (~350) to 5.5 shrapnels (~137), which is quite low for a grenade, unless their utility is significantly boosted.

(Also note that Shrapnels individually slows, so reducing their count dramatically lower the utility provided by the grenade).

Carnage068:
https://media.discordapp.net/attachments/382970842662961153/1008913371418075226/unknown.png

Alardunx:
> https://media.discordapp.net/attachments/382970842662961153/1008913371418075226/unknown.png

That's 12 shrapnels, while also under fire by other things, not entirely sure what we're looking at here (204-396, 300 average)

joookschad:
> might make agms a bit too useless imo...
> 
> just make damage more consistent and reduce the upper bound?

wow thats almost like removing iff on the mk1/mk2 GL could make the damage consistent and lower the skill bar so the damage nerf is excusable

stanalbatross:
> lower the skill bar

not desirable

stanalbatross:
> However, there is currently a bug that afflicts Shrapnels spawning and has been around for a while (And is also the reason for HEFA insta-gibbing people), and it is the fact that Shrapnel spawning codes check for entities on the tile when spawning to ignore them for the guaranteed collision, but not for entities that are currently traveling to the tile where they are spawning.

seems to be the crux of the issue here, 800 damage AGMs would not be happening without this

Alardunx:
> > However, there is currently a bug that afflicts Shrapnels spawning and has been around for a while (And is also the reason for HEFA insta-gibbing people), and it is the fact that Shrapnel spawning codes check for entities on the tile when spawning to ignore them for the guaranteed collision, but not for entities that are currently traveling to the tile where they are spawning.
> 
> seems to be the crux of the issue here, 800 damage AGMs would not be happening without this

For those who are a bit confused about what this means

image

This is how many shrapnels you can expect at max depending on your tile position when the projectile detonate.

SexmasterWasTaken:
If you're gonna nerf AGMFs, buff the M15 frag grenade because its literally useless right now.

Nanu308:
I've increase the shrap count to 16x. As it currently to me still seems too easy to hit mobs as they travel into a tile which allows all 30x shrap to hit and almost instantly kill every single xeno.

This will most likely be the final adjustment, however, if someone finds a good way to implement a lower/higher cap on dmg. taken from shrapnel originating from a specific source, we can buff the shrapnel count back and instead rely on that method to hinder players from immediately being nuked.

PR #871, opened 2022-08-15
Fixes welded locker

### LINKED ISSUE #393: Xenos Unaffected by Welded Closets
## Description

Welding a closet or coffin makes no difference when a xeno tries to open it.

It should be much more difficult for a xeno to smash open a welded closet/coffin. Sometimes even an unwelded closet can take several hits from a xeno before it opens. So welding it should definitely have an impact.

## Steps to reproduce

1. Weld closet (looks like locker).
2. Smash closet as xeno. Notice it's no more difficult than an unwelded closet.

TheBoondock:
## About The Pull Request
The attackby code for xeno attacks on lockers does not check for if the locker is welded shut or not this pr fixes that by adding in the missing check and make it harder to smash open welded lockers (still rng tho, just lower chances)
Fixes #393

## Why It's Good For The Game
Bug begone!

## Changelog

:cl:
fix: fixed a xeno being able to smash open welded lockers just as easily as normal closed lockers
/:cl:


code/modules/mob/living/carbon/xenomorph/attack_alien.dm
```diff
@@ -889,10 +889,14 @@
...
-			M.visible_message(SPAN_DANGER("[M] smashes [src] open!"), \
-			SPAN_DANGER("You smash [src] open!"), null, 5, CHAT_TYPE_XENO_COMBAT)
+		if(!opened)
+			var/difficulty = 70	//if its just closed we can smash open quite easily
+			if(welded)
+				difficulty = 30 // if its welded shut it should be harder to smash open
+			if(prob(difficulty))
+				break_open()
+				M.visible_message(SPAN_DANGER("[M] smashes [src] open!"), \
+				SPAN_DANGER("You smash [src] open!"), null, 5, CHAT_TYPE_XENO_COMBAT)
```
stanalbatross, line 899:
```suggestion
				M.visible_message(SPAN_DANGER("[M] smashes \the [src] open!"), \
				SPAN_DANGER("You smash \the [src] open!"), null, 5, CHAT_TYPE_XENO_COMBAT)
```

TheBoondock:
I think ideally we should have it a set number of slashes or use integrity system so rng can be discarded but that's out of scope

stanalbatross (approved):

PR #876, opened 2022-08-16
pushups

stanalbatross:
## About The Pull Request

adds pushups

prevents crawling without 4/8 hands/feet arms/legs

## Why It's Good For The Game

allows for new roleplay opportunities via game mechanics supporting it (pushups)

prevents immersion breaking limbless blobs crawling

## Changelog

:cl:
add: added pushups, you can do pushups with a verb in the IC tab
fix: you can no longer crawl without enough limbs to do so
/:cl:

### DIFF
--- a/ColonialMarinesALPHA.dme
+++ b/ColonialMarinesALPHA.dme
@@ -1559,6 +1559,7 @@
 #include "code\modules\mob\living\carbon\human\death.dm"
 #include "code\modules\mob\living\carbon\human\emote.dm"
 #include "code\modules\mob\living\carbon\human\examine.dm"
+#include "code\modules\mob\living\carbon\human\exercise.dm"
 #include "code\modules\mob\living\carbon\human\human.dm"
 #include "code\modules\mob\living\carbon\human\human_abilities.dm"
 #include "code\modules\mob\living\carbon\human\human_attackhand.dm"
--- a/code/modules/mob/living/carbon/human/exercise.dm
+++ b/code/modules/mob/living/carbon/human/exercise.dm
@@ -1,0 +1,122 @@
+/*
+Verbs related to getting fucking jacked, bro
+*/
+
+/mob/living/carbon/human/verb/pushup()
+	set name = "Do Pushup"
+	set desc = "Makes you do a pushup."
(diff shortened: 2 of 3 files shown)


FoxyStalin:
I feel like you should still be able to crawl with just arms.

stanalbatross:
> I feel like you should still be able to crawl with just arms.

If you have JUST arms you can crawl

but you need both hands

1 arm with hand and 1 arm with no hand == you can't crawl

neeshacark:
I mean you can totally crawl without using hands, you can use your forearms or elbows.  The only thing that should stop you from crawling is losing arms.  Maybe not totally either, maybe losing one arm slows your speed down and losing both stops you entirely.  But you like - really don't need hands to crawl.

hotimep:
private 1000 pushups

anclandor:
I feel like its possible to crawl without any hands, or feet.

PhantomEpicness:
soul

AGasCanister:
The push ups is a cool addition, but it's very limited. A marine with endurance 1 will always do 18 push ups no matter what. There is really no variety nor randonmess. There will always be a fixed number of push ups. So I thought of two suggestions that could make push ups fun and competitive,
1. There should be a chance to recover stamina from a push up, reducing the stamina cost of a push up, or simply negate the stamina cost of a push up, which, as I said before, randomizes the number of push ups that a marine can do. Take it as folding chair stacking, determined with luck
2. We have a trait system that's used in languages, I recommend expanding it in a way, like adding a trait that you can spawn with or get by either obtaining EXP for every push up you do, or by simply doing 10, 15 push ups in a row in order to level it up. Of course, the cost should increase so you don't get many levels super fast. This trait could increase the chances of recovering, reducing or negating a stamina push up cost. A speed increase would be interesting to see, so you could differenciate between the low level marine starting doing push ups against the marine that spent ten minutes doing push ups. Maybe some text changes could work too to make it more interesting and duel preds/synths in a push up battle.

Now in order to increase your push up limit or trait, I thought of using bicaridine or adding a new steroids chemical to give it a lore/IC excuse for the accelerated muscle regeneration.

literallyme

PR #880, opened 2022-08-16
The Sandbag Buff No One Asked For

morrowwolf:
## About The Pull Request

This changes unfilled sandbag stacks to be small items.

## Why It's Good For The Game

Currently we have a situation where sandbags are pretty much outclassed in every way by metal/plasteel. You have to take a stack (equal to the size of metal/plasteel stacks) and then dig (also a small item you have to carry) for a while before they're even usable and _then_ you can't even store them effectively. Even further down the line, the final product can only be repaired with more sandbags. Meanwhile, metal/plasteel is repaired by welding and welding fuel is easy to carry lots of and easily found.

Sandbags sort of share the same niche with portacades with quickly placing them to fill holes in lines but we can do better and I hope this is step one to seeing some neat sandbag usage.

## Changelog

:cl: Morrow
balance: Unfilled sandbag stacks are now small items.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/sandbags.dm
+++ b/code/game/objects/items/stacks/sandbags.dm
@@ -6,7 +6,7 @@
 	icon = 'icons/obj/items/marine-items.dmi'
 	icon_state = "sandbag_stack"
 	item_state = "sandbag_stack"
-	w_class = SIZE_MEDIUM
+	w_class = SIZE_SMALL
 	force = 2
 	throwforce = 0
 	throw_speed = SPEED_VERY_FAST


Lafrien:
Sandbags being normal-sized while unfilled is not the main problem here. You can bring lots of sandbags to the groundside and easily fill them there.
The problem is - after sandbags being filled you are unable to fit them into the backpack, forcing you to always have them in hands. What that means? You exchange your hand slot for debatable defensive structure while you walk from FoB to frontline.
You may want to say "Well why not just carry them unfilled and then fill them on the spot", to which i will say: some maps have frontlines in spots, where no "sand" will be. And even if you are lucky enough, you will be forced to fill them on the spot during the fight, which can end quickly in being pushed by friendly or enemy forces, leaving you with half filled sandbags that you still will be forced to carry in hand.

My suggestion for making them actually good? Make filled sandbags normal-sized item to fit in backpacks/satchels.

stanalbatross (approved):

Issue #832, opened 2022-08-09
View Objectives interface sometimes BSODs when viewing the data tab


CometBlaze:
## Description

At seemingly random the data tab of the View Objectives interface will receive a corrupted entry with null display colour which causes the tab to BSOD. This is limited to an individual but it can be transmitted to EVERYONE else should the "infected" person use the data upload computer inside the IO lab. 
There has also been one instance where the retrieve tab was the one to BSOD, also due to a corrupt entry with null colour.

## Steps to reproduce

1. Spawn as IO
2. Do your thing
3. ????
4. Check the data tab and verify it isn't working

## Issue bingo (not required to be all checked)

- [x] Issue could be reproduced at least once
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Segrain:
Several separate issues at work here that can cause indistinguishable results. Going to do something about at least some of them.

Sabitix:
Ok i got this bug in my last round so it happened right after i uploaded a shit ton of data at once onto the terminal at once

Mark9013100:
![image](https://user-images.githubusercontent.com/115238768/196964789-039b415d-e88e-47a0-be93-ef89ee64542f.png)

Mark9013100:
update: last time I had the data tab BSOD, it was after reading about 60 or so intel items, the last item was a tech manual.

ThePiachu:
Got that during my recent IO game. It has something to do with you reading the intel I think, since I got it, but uploading all my intel to other IOs they didn't get it and they could read the passwords I couldn't.

ThePiachu:
It happened while I was testing stuff offline. I read all the paper scraps, reports, folders and things were fine. I was decoding a few disks, and then when I started readiny my first Manual it BSOD:

![image](https://user-images.githubusercontent.com/964559/199687262-fec81f3f-6302-42ea-ad7b-a2d2e2f1a0af.png)

ThePiachu:
Or it could be because I was cracking disks as well.

ThePiachu:
Okay, I've narrowed it down to two suspects:

1) Terminals don't have colour, so TGUI errors out trying to reference it

OR more likely:

2) Some Nightmare Inserts delete some derminals and the Objective never reflects that.

Seeing as in my test I have objectives for terminals that are null and not at the location, I think number 2 is it.

PR #882, opened 2022-08-16
Fix for disappearing targets of objectives.

Segrain:
## About The Pull Request

_One_ cause of #832 is objective items being destroyed by chance OB, stray grenade, IO's own C4 and whatever else tends to explode in the course of the round. Given that intel related things tend to be in general indestructible by acid, bomb and whatever else, it is probably much saner to extend that protection to all relevant items, so that their destruction does not leave objectives hanging and IOs bluescreening.
This is not the only cause, so not marking the bug report for closure yet.

## Why It's Good For The Game

Is fix.

## Changelog

:cl:
fix: IO objectives tied to items no longer bluescreen due to said items getting exploded.
/:cl:


morrowwolf (approved):
CODE LGTM

PR #893, opened 2022-08-18
fixes incredibly strong phone line

### LINKED ISSUE #366: RTO Phone Dragging
## Description

Normally-immovable objects can be dragged via phone cord, plus cord visiblity issues.

Lockers and disposal units shouldn't be able to be dragged around the ship. Tables also shouldn't be pulled apart (i.e. square with phone placed on it detaches from rest of table) just because you're walking away with the RTO backpack.

Phones can also drag crates around, but that actually seems realistic. As well as the ability to have a chain of people holding each others phones being dragged by a single person I suppose.

Some weirdness happens when a xeno devours an RTO, when another person is holding the phone from the RTO's backpack. The tether visually disappears, but it's still there (i.e. the "phone holder" is still tethered to the xeno who ate the RTO). However, the phone holder doesn't seem to get pulled, until the RTO is regurgitated perhaps a tile or two away.

Going into a locker while holding an RTO phone also makes the cord disappear.

## Steps to reproduce

1. Be RTO.
2. Call whatever phone.
3. Place your phone handset into a locker, disposals, on a table, in a crate, or on/in a similar container.
4. Run away.
5. See your phone handset dragging the object behind you (the table tile will separate from the rest of the table).

TheBoondock:
## About The Pull Request
In proc Moved() of tether codes any obj is converted into a subtype of atom/movable/ therefore any and all objects are moveable. This pr simply add an early return if the the tether is connected to a strcuture (which most are anchored)
I have attempted to allow it o move unanchored objects but was unable to do so advices are welcomed for such implementation
Fixes parts of #366

## Why It's Good For The Game
Bus gone is good!

## Changelog

:cl: The_Boondock
fix: due to unintended materials being used in phone lines resulting in them being incredibly strong, the mistakes have been found and all current phone lines are recalled and replaced with weaker one to save on cost, they will now break if pulling a heavy objects.
/:cl:

### DIFF
--- a/code/datums/effects/tether.dm
+++ b/code/datums/effects/tether.dm
@@ -42,7 +42,9 @@
 		return
 
 	if(isStructure(tethered.affected_atom))//we are attached to a structure, shouldnt move it (too heavy)
-		return
+		var/obj/structure/BigBoi = tethered.affected_atom
+		if(BigBoi.anchored)
+			return
 
 	var/atom/movable/A = tethered.affected_atom
 	if (get_dist(affected_atom, A) <= range)
--- a/code/datums/effects/tether.dm
+++ b/code/datums/effects/tether.dm
@@ -42,8 +42,8 @@
 		return
 
 	if(isStructure(tethered.affected_atom))//we are attached to a structure, shouldnt move it (too heavy)
-		var/obj/structure/Big_boi = tethered.affected_atom
-		if(Big_boi.anchored)
+		var/obj/structure/anchored_object = tethered.affected_atom
+		if(anchored_object.anchored)
 			return
 
 	var/atom/movable/A = tethered.affected_atom
--- a/code/datums/effects/tether.dm
+++ b/code/datums/effects/tether.dm
@@ -42,8 +42,8 @@
 		return
 
 	if(isStructure(tethered.affected_atom))//we are attached to a structure, shouldnt move it (too heavy)
-		var/obj/structure/BigBoi = tethered.affected_atom
-		if(BigBoi.anchored)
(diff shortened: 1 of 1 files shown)


code/datums/effects/tether.dm
```diff
@@ -41,6 +41,11 @@
 	if (isnull(tethered))
 		return
 
+	if(isStructure(tethered.affected_atom))//we are attached to a structure, shouldnt move it (too heavy)
+		var/obj/structure/BigBoi = tethered.affected_atom
+		if(BigBoi.anchored)
```
stanalbatross, line 46: snake_case

stanalbatross (changes requested):

code/datums/effects/tether.dm
TheBoondock: Wut?

stanalbatross: Variable names must use snake_case

code/datums/effects/tether.dm
```diff
@@ -41,6 +41,11 @@
 	if (isnull(tethered))
 		return
 
+	if(isStructure(tethered.affected_atom))//we are attached to a structure, shouldnt move it (too heavy)
+		var/obj/structure/Big_boi = tethered.affected_atom
```
morrowwolf, line 45: nitpick but snake_case is not capitalized throughout. Also this variable could be named way better.

morrowwolf (changes requested):

code/datums/effects/tether.dm
morrowwolf: Oh and make sure to put your ckey next to the :cl: tag

PR #894, opened 2022-08-18
Fix unbreakable power APCs after basic repairs

### LINKED ISSUE #890: Power APCs cannot be broken by most xenos after repair
## Description

Once an APC has had its wires cut by a xeno (not fully broken requiring new frame and circuity), and its repaired, the working APC will give the xeno the message like `You aren't big enough to further damage the Main Hallway APC` because the beenhit var is still set to 10 even after repairs. If you have a xeno able to damage it further, they can of course damage it further and break it.

It is possible this is intended behavior, but seems weird. I would expect any working APC to have its wires exposed and cut again.

## Steps to reproduce

- Slash an APC with a defender w/o crest down until wires are cut
- Mend wire w/ wirecutters
- Screwdriver to close panel
- Attempt to slash w/o crest down

- [X] Issue could be reproduced at least once (Experienced in round and confirmed locally on latest master)

Drulikar:
## About The Pull Request
Resolves #890 

## Why It's Good For The Game
Don't break my immersion: If a power APC is fixed, it should be able to be deactivated again. Additionally, if wires are exposed, they now can be immediately cut by a xeno.

## Changelog

:cl:
fix: Fixed power APCs being unbreakable (atleast not by small/med xenos) after basic repairs have been made (mending wires and closing panel)
fix: Xenos now cut all power APC wires, not just all - 1 because of an off-by-one error
fix: If a power APC's wires are exposed some reason, now they can just immediately be slashed
code: Power APC cut proc now can take a boolean argument to hide messages (And xeno slashing now doesn't produce the redundant messages)
/:cl:

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -631,6 +631,7 @@ GLOBAL_LIST_INIT(apc_wire_descriptions, list(
 				update_icon()
 		else
 			wiresexposed = !wiresexposed
+			beenhit = wiresexposed ? XENO_HITS_TO_EXPOSE_WIRES_MIN : 0
 			user.visible_message(SPAN_NOTICE("[user] [wiresexposed ? "exposes" : "unexposes"] [src]'s wiring."),
 			SPAN_NOTICE("You [wiresexposed ? "expose" : "unexpose"] [src]'s wiring."))
 			playsound(src.loc, 'sound/items/Screwdriver.ogg', 25, 1)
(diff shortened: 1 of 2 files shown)


code/modules/power/apc.dm
```diff
@@ -890,22 +891,22 @@ GLOBAL_LIST_INIT(apc_wire_descriptions, list(
 	var/wireFlag = getWireFlag(wire)
 	return !(apcwires & wireFlag)
 
-/obj/structure/machinery/power/apc/proc/cut(wire, mob/user)
+/obj/structure/machinery/power/apc/proc/cut(var/wire, mob/user, var/withMessage = TRUE)
```
stanalbatross, line 894: snake_case

stanalbatross (approved):

PR #895, opened 2022-08-18
Dancer buff: moar mobility, moar evasion

TheBoondock:
## About The Pull Request
Adds the dash ability to the dancer kit granting the strain more agility to perform its niche, it also slightly buff its evasion timer from slash to 5 seconds 

## Why It's Good For The Game
So dancer, its purpose is a damage dealer, praetorians trade away their health and armour to gain the speed and damage granted by this strain. This strain's playstyle is supposed to be a get in, choose an unfortunate marine to nuke him with your abilities and slashes. Its a quick get in deal damage get out playstyle, of course others play it differently but i believe this is the inteded playstyle (theres also a high risk high reward situation with it punishing you for missing clicks). However,c urrent state of dancer is quite abysmal, its a tier 3 caste yet it doesnt function well on frontline thus many has resort to sideline fight where theres 1 to 3 marines. I believe this problem stems from the fact that dancer is too squishy to survive focus fire long enough to do meaningful damage and get out alive. This problem can be addressed in 2 ways, one is buffing its health which is lame or the other more fun way, which is giving it more mobility. So thats what i did, giving dancer a quick dashing ability that ranges to 6 tiles and functions with the tagging system, meaning if you dive into a target that has a tag you will slash and knock them down instead. This should hopefully allows dancer to get into the fight, deal their damage and get out before all the focus fire kills it. 

Argument 1: "But theres already dodge to help it get into the fight"
Counter-argument: dodge serves as a counter against block as well as a good speed buff to disengage in battle, keyword here is disengage because thats what it is imo. It allows you to pass through mobs and grants a speed buff to run away without being bodyblocked. The dash will give it something to engage while dodge will be used to disengage

Argument 2: "But this makes it too agile"
Counter-argument: Yes its supposed to be an agile caste since it is losing its health pool in exchange for this speed and agility 

Argument 3: "Why does it need the evasion buff then?"
Counter-argument: Because 2.5 seconds isnt enough window for it imo, what with time dilation and all this 2.5 seconds could be gone in a flash (not to mention players regularly experience crashes/freezes)

Of course all of this can be change at a request of a maintainer, so discuss your balance concerns

## Changelog

:cl:
balance: dancer evasion buff now last 5 seconds instead of 2.5
add: Through numerous battle the dancer caste has evolved a new ability to aid its fight against the tall horde. It can now dash  up to 6 tiles away and if it dashes on a human that has been tagged before, it will knock the human down and slash them.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_abilities.dm
@@ -208,6 +208,23 @@
 
 	var/click_miss_cooldown = 15
 
+/datum/action/xeno_action/activable/pounce/dancer
+	name = "Dive"
+	action_icon_state = "prae_dash"
+	ability_name = "Dive"
+	macro_path = /datum/action/xeno_action/verb/verb_dive
+	action_type = XENO_ACTION_CLICK
+	ability_primacy = XENO_PRIMARY_ACTION_4
+	xeno_cooldown = 40
+	plasma_cost = 50
+
+	// Config options
+	distance = 6
+	knockdown = FALSE
+	slash = FALSE
+	freeze_self = FALSE
+	throw_speed = SPEED_VERY_FAST //tier 3 agile xeno
+
 ////////// BASE PRAE
 
 /datum/action/xeno_action/activable/pounce/base_prae_dash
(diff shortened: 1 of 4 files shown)


PalJohnson:
Would probably mark it as DRAFT until the TODO is finished otherwise PR might be reviewed with content that is already here

Carnage068:
my god yes please it has needed something like this for so long LOL

TheBoondock:
nevremind i think im happy with the way it is and the to-do list isnt necessary anymore

Nanu308:
Needs to be tested on live.

PR #905, opened 2022-08-20
Cryomarine Stuff

Geevies:
## About The Pull Request

The Reserves squad is now called Foxtrot. Their brown color has been brightened a bit, the radio color as well. They also no longer break the crew manifest.

You can now call the reserves in-game via the tech tree, it's a Tier 3 tech that costs 10 points, but increases by 5 each time it's called. The squad contains a squad leader, medics, engineers, and privates.

The tech points required to go up a tier has been reduced to 5 from 15.

## Why It's Good For The Game

Foxtrot is a cooler name than Reserves, and exists within the actual games. The brown was a bit muddy and made it look like they didn't have a squad at all, so the brighter one resolves that.

Calling the cryomarines via the tech tree is cool, and keeps people in the game in longer games.

The tech points required to transition was too much, I've been told, so knocking it down to 5 seems fine to me.

## Changelog

:cl:
fix: The Reserves squad is now called Foxtrot. Their brown color has been brightened a bit, the radio color as well. They also no longer break the crew manifest.
add: You can now call the reserves in-game via the tech tree, it's a Tier 3 tech that costs 10 points, but increases by 5 each time it's called. The squad contains a squad leader, medics, engineers, and privates.
balance: The tech points required to go up a tier has been reduced to 5 from 15.
/:cl:


hotimep:
foxtrot squad will be real in 3 minutes

code/__DEFINES/job.dm
```diff
@@ -8,7 +8,7 @@
 #define SQUAD_MARINE_3					"Charlie"
 #define SQUAD_MARINE_4					"Delta"
 #define SQUAD_MARINE_5					"Echo"
-#define SQUAD_MARINE_CRYO				"Reserves"
+#define SQUAD_MARINE_CRYO				"Foxtrot"
```
stanalbatross, line 11: IMO this change just makes the reserves squad more confusing 
it should still be called reserves
or just put the cryorines into random squads

stanalbatross (approved):

PhantomEpicness:
INCREDIBLY based

Wintermote:
Cryomarines in the past were way too expensive for what they delivered.

How many marines is this spawning? Ideally you can get 2 purchases with a good budget supply from intel, but if we're talking 4 marines each time like it used to be, that simply isn't going to be enough to matter.

Marines need somewhere in the range of 8-12 for the number of reinforcements to matter. Especially if there's only a handful of specialists like medics or engineers in the group. It should also scale depending on total players. An extra 12 or 15 marines when there were only 30 marines to start with is way too much, and 6 marines for a round that started with 60+ is way too little

Imo the sweet spot is a 15-20% reinforcement amount. That should be enough to give marines a bit more staying power or recover and counter attack when necessary.

Anyway, good PR. It is, imo, one of the most important changes for making intel collection not feel underwhelming/useless. Yeah more ammo for CAS/Mortar is nice, but you really do not feel that compared to knowing you got some more marines back into the round. The hope of IO's collecting enough intel for cryomarines imo keeps some people spectating the round after they die instead of quitting and coming back in an hour.

FoxyStalin:
The Cryomarine's headset is a little bit bugged, in order to get the SL tracker you have to take your headset off and put it back on.

Diegoflores31:
would be good if there was a lil bit more marines but appart from that very nice stuff . this will make long rounds less boring

morrowwolf (approved):
Code LGTM, needs architect approval

morrowwolf (changes requested):
Wait I lied, fix the headset bug where they need to re-equip the headset for tracking to work.

PR #908, opened 2022-08-20
Cluster Mortar Ammo

Geevies:
## About The Pull Request

Adds cluster mortar ammo, which is basically just a weaker version of the OB.

## Why It's Good For The Game

It's neat.

## Changelog

:cl:
add: Added cluster mortar ammo, which is like the OB but smaller.
/:cl:


JohnDoe50974:
No way, CoH British Artillery in CM

code/modules/cm_marines/equipment/mortar/mortar_shells.dm
```diff
@@ -59,6 +59,33 @@
...
+/obj/item/mortar_shell/cluster/detonate(var/turf/T)
+	start_cluster(T)
+
+/obj/item/mortar_shell/cluster/proc/start_cluster(turf/target)
+	set waitfor = 0
+
+	var/range_num = 7
+	var/list/turf_list = list()
+
+	for(var/turf/T in RANGE_TURFS(range_num, target))
```
stanalbatross, line 79: /t

stanalbatross (changes requested):

code/modules/cm_marines/equipment/mortar/mortar_shells.dm
```diff
@@ -59,6 +59,33 @@
...
+	var/total_amount = 8
+	var/instant_amount = 3
+
+/obj/item/mortar_shell/cluster/detonate(var/turf/T)
+	start_cluster(T)
+
+/obj/item/mortar_shell/cluster/proc/start_cluster(turf/target)
+	set waitfor = 0
+
+	var/range_num = 7
```
stanalbatross, line 76: might want to consider reducing range ig

Geevies: i think making it concentrated on a small area would make it too much of a choke-buster, the bigger range makes it more balanced in that way. maybe we should TM and see?

Ticker175:
I can say that when this was first test merged I didn’t notice until these funny new blue shells appeared from the sky. Req said it was a gift from ares and I was very excited to give them a go, I did and it does what it says on the tin it’s a low damage large area cluster mortar, effective at clearing low-moderate strength walls/obstacles or severely damaging teir 1s and some tier 2s

Very happy with this addition I also think it’s very balanced at the current and needs no tweaking.

also there was some comment on stats, from practical experience the area of the explosions is approx 30x30 (Centered on the splash tile), damage is about enough to bring a hivey down to a quarter health or kill a drone, duration is around 8-12 seconds and amount of individual shells is approx 20

Ticker175:
So an update to my previous comment. It seems that these shells have been severely nerfed and are almost useless in the current form. They have had their amount of individual explosions reduced significantly, the explosive power has been significantly been reduced and the damage is pathetic

Wintermote:
I haven't noticed these doing much of anything. Might be too weak.

Ticker175:
As I said previous cluster shells in their current state are useless. They arnt even effective to remove weeds! They desperately need a buff

c4xmaniac:
Damage is too low in its current state, no one uses it over incen or He atm. 

Also could you add shrapnel mortar shells to req as apart of this PR? It’s not necessarily apart of it obviously, but they’ve been missing from req for over a year.

PR #915, opened 2022-08-21
some tgui fixes

stanalbatross:
## About The Pull Request

- fixes not being able to open door tguis
- fixes wires/panel window not opening when APCs have their wires exposed/unexposed

## Why It's Good For The Game

fix nge

## Changelog

:cl:
fix: fixed not being able to open an airlock's wires UI if the airlock had no power
fix: fixed APCs not automatically reopening their UI if one was already open when their covers were closed/opened
/:cl:

### DIFF
--- a/code/game/machinery/computer/camera_console.dm
+++ b/code/game/machinery/computer/camera_console.dm
@@ -60,6 +60,11 @@
 		user.set_interaction(src)
 	tgui_interact(user)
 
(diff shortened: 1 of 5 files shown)


morrowwolf (changes requested):
Just from a design standpoint I'm not sure if interacting with the door via cutting wires should do anything if there is no power. Pulsing I can kinda see as you are applying electricity (if I understand correctly).

morrowwolf (approved):
Stan says design good. CODE LGTM

PR #920, opened 2022-08-22
Readjust AGM-F

Alardunx:
## About The Pull Request

Reducing the cone to keep with the reduced shrapnel count.

## Why It's Good For The Game

Attempting to keep AGMF relevant while reducing the unexpected burst from the glitch triggering and doing omega damage.

## Changelog

:cl:
balance: Reduced AGM-F cone from 60 degree to 40 degree. Along with the shrapnel count reduction from a previous change, this has essentially two effects: The situation which caused targets to take all Shrapnels can now at most do ~133% expected damage (down from 250%) and the side-hits are even weaker than they were before. Directly in front of the target still deals damage similar to what it used to be (~8-13 shrapnels)
/:cl:

image

### DIFF
--- a/code/game/objects/items/explosives/grenades/marines.dm
+++ b/code/game/objects/items/explosives/grenades/marines.dm
@@ -144,7 +144,7 @@
 	falloff_mode = EXPLOSION_FALLOFF_SHAPE_LINEAR
 	shrapnel_type = /datum/ammo/bullet/shrapnel/jagged
 	var/direct_hit_shrapnel = 5
-	var/dispersion_angle = 60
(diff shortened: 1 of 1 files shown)


DylonBR:
good PR, very cool

Shad0vvs (approved):

Sernoh:
Compared to pre-nerf AGM-Fs, this change retains its old power when firing dead-on. However when firing at oblique angles, only a fraction of shrap will hit and you deal minuscule damage. It's much harder to use compared to before.

Great MR, I only wish 750-damage nuke bug was fixed and we had old old AGM-Fs.

PR #927, opened 2022-08-22
Fix for ghost pills.

### LINKED ISSUE #897: "Ghost pills" occasionally spawning after feeding a real pill to someone.
## Testmerges:

A single PR is gona be difficult to associate with this bug. I have known of it for a year or more but can not intentionally cause the bug to occur.

## Description

Sometimes, seemingly more often at high pop when the lag is at its highest, just after administering a pill to a marine, a "ghost pill" is created on the ground where you were standing. This pill has 0u worth of chems and cannot be interacted with in any way via a person's hands. You cannot drag it nor pick it up though the pill does appear in the right click and alt + click menus. I have observed this 'ghost pill' from all types of chems so long as they are stored in pill form: custom such as IA or KD and basic such as kelotane or tramadol. 

Feeding someone a pill should delete said pill instead of leaving a useless object to clutter the ground. Attached are 3 separate examples of this 'ghost pill' in a single round on 8/18/2022 but if you keep an eye out, you will likely notice these 'ghost pills' in almost all rounds:

Ghost Pill 1
-------------------
Ghost Pill 2
-------------------
Ghost Pill 3

## Steps to reproduce

I have tried in vain to replicate this bug over quite some time. The ghost pills SEEM to occur most frequently when the target of the pill walks out of the tile you are feeding just as you finish feeding the pill but the ghost pill does not always spawn from this. Combine that with exceptional lag spikes and that is my best guess as to the origin of these ghost pills. 

Segrain:
## About The Pull Request

Trying to feed the pill a second time during the `do_after` (which is more likely to happen on high lag) resulted in it being properly dropped and qdeleted on first feed, but then dropped again on second feed resolving anyway, resulting in the ghost pill on the ground, which escapes deletion. Checking for `QDELETED` prevents it for now, and one day I will get to actually updating all the `do_after`s to run proper checks for items being used.
Also removed the unnecessary checks in text messages: all of that is happening in the `else` of `if(M==user)`, no need to check for it again.

## Why It's Good For The Game

Closes #897.

## Changelog

:cl:
fix: Feeding pills to somebody while lagging and clicking a lot no longer leaves ghost pills on the floor.
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/pill.dm
+++ b/code/game/objects/items/reagent_containers/pill.dm
@@ -84,22 +84,25 @@
 			return
 
 		user.affected_message(M,
-			SPAN_HELPFUL("You <b>start feeding</b> [user == M ? "yourself" : "[M]"] a pill."),
+			SPAN_HELPFUL("You <b>start feeding</b> [M] a pill."),
 			SPAN_HELPFUL("[user] <b>starts feeding</b> you a pill."),
-			SPAN_NOTICE("[user] starts feeding [user == M ? "themselves" : "[M]"] a pill."))
+			SPAN_NOTICE("[user] starts feeding [M] a pill."))
 
 		var/ingestion_time = 30
 		if(user.skills)
 			ingestion_time = max(10, 30 - 10*user.skills.get_skill_level(SKILL_MEDICAL))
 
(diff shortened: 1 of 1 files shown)


morrowwolf (approved):
CODE LGTM

PR #938, opened 2022-08-24
bugfix: move delay after stopping to pull a mob

Git-Nivrak:
## About The Pull Request
If you grab and pull a mob then stop pulling it and try to move you will have the same move delay as if you are grabbing the mob even though you are not grabbing it for one tile - this fixes it

## Why It's Good For The Game
bugs are bad for the game so if there is less bad there is more good for the game

## Changelog

:cl:
fix: Fixed a bug which would make you have increased movement delay after stopping to pull a mob for one tile
/:cl:

### DIFF
--- a/code/modules/mob/mob_verbs.dm
+++ b/code/modules/mob/mob_verbs.dm
@@ -226,8 +226,6 @@
 		pulling = null
 
 		grab_level = 0
-
-
 		if(client)
 			client.recalculate_move_delay()
 			// When you stop pulling a mob after you move a tile with it your next movement will still include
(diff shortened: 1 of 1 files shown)


code/modules/mob/mob_verbs.dm
```diff
@@ -226,8 +226,13 @@
 		pulling = null
 
 		grab_level = 0
+
+
```
stanalbatross, line 230: why

stanalbatross (changes requested):

code/modules/mob/mob_verbs.dm
Git-Nivrak: soul

stanalbatross (approved):

PR #939, opened 2022-08-25
M2C Ammo type fix

Mister-moon1:
## About The Pull Request
Fixes the type of ammo the M2C box mags uses, changing it from smartgun ammo to machinegun auto ammo

## Why It's Good For The Game
The M2C doesnt use smartgun ammo it uses 10x28mm tungsten

## Changelog

:cl:
Fix: Ammo type of M2C changed to be the correct ammo
/:cl:

### DIFF
--- a/code/modules/cm_marines/smartgun_mount.dm
+++ b/code/modules/cm_marines/smartgun_mount.dm
@@ -27,7 +27,7 @@
 	flags_magazine = NO_FLAGS //can't be refilled or emptied by hand
 	caliber = "10x28mm"
 	max_rounds = 700
-	default_ammo = /datum/ammo/bullet/machinegun/auto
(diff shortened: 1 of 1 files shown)


Mister-moon1:
i was initially mistaken due to spaghetti code but i found that the m2c should be using /datum/ammo/bullet/machinegun/auto as its default ammo type and not /datum/ammo/bullet/machinegun, this has now been corrected

stanalbatross (changes requested):
also needs the caliber var set or it makes () caliber rounds

I can confirm this will not affect the damage of the gun

Mister-moon1:
It should bare a caliber type now and shouldnt show up as () anymore if it all goes correctly

stanalbatross (approved):

PR #943, opened 2022-08-26
Fix area names having their first letter sometimes truncated during sanitization

Drulikar:
## About The Pull Request
Instead of: 
image
How about: 
image

Some reason replacetext in code\__HELPERS\text.dm is also including the first char we want, usage of replacetext_char has no effect on this behavior, replacetextEx doesn't seem to match the special char, and html_encode is already removing the characters we don't want. Let me know if the default sanitization should just be changed, or if there's a different solution to this that would be preferred. I figured turf names are much harder to exploit so we can probably be less aggressive on the sanitization there, but I do not know what exploit was entailed in https://github.com/cmss13-devs/cmss13/commit/e0027fbde74cc4c6ed9caac81cbcc75f24679faf to know if this change would undo that fix.

## Why It's Good For The Game
Correct area names that don't have their first letter truncated.

## Changelog

:cl: Drathek
fix: Fixed strings of area names (e.g. "ombat Information Center") by using strip_improper instead.
/:cl:

### DIFF
--- a/code/__HELPERS/text.dm
+++ b/code/__HELPERS/text.dm
@@ -25,37 +25,41 @@
 	return t
 
 //Removes a few problematic characters
-/proc/sanitize_simple(var/t, var/list/repl_chars = list("\n"=" ","\t"=" ","�"=" "))
+/proc/sanitize_simple(var/text, var/list/repl_chars = list("\n"=" ","\t"=" ","�"=" "))
 	for(var/char in repl_chars)
-		t = replacetext(t, char, repl_chars[char])
-	return t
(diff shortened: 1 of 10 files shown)


Geevies:
might be worth creating a sanitize_area_name proc instead. why do we even need to sanitize area names though, don't we have full control of it?

stanalbatross:
> might be worth creating a sanitize_area_name proc instead. why do we even need to sanitize area names though, don't we have full control of it?

station blueprints can change it I think, if they weren't dummied out

stanalbatross (approved):

PR #957, opened 2022-08-28
limb printer update

stanalbatross:
## About The Pull Request

gives some overhauling to the limb printer such as new features and tgui

## Why It's Good For The Game

severly unmaintained and old machine recieving a fresh lick of paint codewise, also makes gathering metal for it more of a task rather than a background nonissue

## Changelog

:cl:
ui: the limb printer now uses tgui
ui: the signaller tgui now uses sliders rather than knobs
add: gave the limb printer some love. It now has a unique sprite when printing, can eject metal, has a max metal limit of 500, and starts with 125 metal instead of 1000. It can no longer accept meat (was not used for anything) and its backend code is much better now.
/:cl:


stanalbatross:
> Is it still going to cost 125 metal to print a limb? If so, this means that between the four limb printers in medbay field docs could only print ~4 limbs at round start. Given that xenos target feet/legs and delimbs are quite common this feels like it may be a balance adjustment as it will require more metal to be diverted to medbay for limb printing instead of FOB as rounds go on.
> 
> Not saying it's a bad change, just that if the above is accurate, maybe a balance tag should be applied to this PR?

true, my intention with reducing the metal count a bit was to require docs to scav a bit for metal (they only get 4 limbs round-start instead of 32)

monkeysfist101 (changes requested):
Stray pixel in the top left on one animation frame. Fix it and the sprites will be good.
image

Agameofscones (approved):
sprites and mapping are GOOD

stanalbatross:
> Stray pixel in the top left on one animation frame. Fix it and the sprites will be good. image

icondiffbot error, there wasn't one in the actual dmi (I checked)

code/game/machinery/autodoc.dm
```diff
@@ -626,10 +627,13 @@
 		return // no
 	if(istype(W, /obj/item/stack/sheet/metal))
 		var/obj/item/stack/sheet/metal/M = W
+		for(var/i in 1 to M.amount)
+			stored_metal += 100
+			stored_metal = sanitize_integer(stored_metal, 0, max_metal, 0)
+			M.use(1)
+			if(stored_metal == max_metal)
+				break
```
spookydonut, line 635: do the maths properly this is awful

code/game/machinery/bioprinter.dm
```diff
@@ -1,73 +1,162 @@
...
-		"synthetic left leg (125 - Metal)" =  list(/obj/item/robot_parts/leg/l_leg,  0, LIMB_METAL_AMOUNT, LIMB_PRINTING_TIME),
-		"synthetic right leg (125 - Metal)" = list(/obj/item/robot_parts/leg/r_leg,  0, LIMB_METAL_AMOUNT, LIMB_PRINTING_TIME)
-		)
+/obj/structure/machinery/bioprinter/Initialize()
+	. = ..()
+	if(isnull(products))
+		products = list()
+		for(var/i in subtypesof(/datum/bioprinter_recipe))
+			var/datum/bioprinter_recipe/bp_recipe = new i
+			products += bp_recipe
```
spookydonut, line 26: why not use a GLOB ?

spookydonut (changes requested):

spookydonut (approved):

PR #877, opened 2022-08-16
reverts blowtorches no longer causing eye damage ever

stanalbatross:
## About The Pull Request

welding screen var turned to OFF

closes #859 

## Why It's Good For The Game

**gameplay**

cutting out an entire game mechanic solely because of a little quirk on the sprite is not good. There was no reason to remove the need for welding protection and it makes the entire system deprecated and useless - as well as nullifying several other interesting mechanics such as synth welding protection

you should also be more vulnerable while welding and not able to see everything around you with ease.

also, stuff like this 
image
 should not be possible

**realism**

a tiny little glass screen on a welder CANNOT protect you from the massive light caused by the torch. The light will go _around_ the screen and directly into your eyes.

I know that in alien isolation it works like that, but due to gameplay concerns this should not be a thing. If it really matters then change the sprite to not have the transparent gap in the screen. (their blowtorch also works off replaceable cans while ours does not - completely different item)

## Changelog

:cl:
del: blowtorches no longer always protect you from eye damage
add: added "old blowtorches", which have a weaker flame that only lets them weld vents and doors. However, this makes their welding screens effective and as such they do not hurt your eyes.
/:cl:

PR #964, opened 2022-08-28
Fix for wrenching vehicles.

Segrain:
## About The Pull Request

Followup to #877. No, a wrench does not need to be a blowtorch to work.

## Why It's Good For The Game

Is fix.

## Changelog

:cl:
fix: Wrench once again can be properly used in fixing vehicles' hulls.
/:cl:

### DIFF
--- a/code/modules/vehicles/multitile/multitile_interaction.dm
+++ b/code/modules/vehicles/multitile/multitile_interaction.dm
@@ -20,9 +20,6 @@
 
 	// Are we trying to repair the frame?
 	if(iswelder(O) || HAS_TRAIT(O, TRAIT_TOOL_WRENCH))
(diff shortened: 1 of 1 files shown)


code/modules/vehicles/multitile/multitile_interaction.dm
```diff
@@ -20,9 +20,6 @@
 
 	// Are we trying to repair the frame?
 	if(iswelder(O) || HAS_TRAIT(O, TRAIT_TOOL_WRENCH))
-		if(!HAS_TRAIT(O, TRAIT_TOOL_BLOWTORCH))
```
stanalbatross, line 23: still needs to check if the blowtorch is weak/strong to prevent blowtorches without the trait from repairing vehicles

stanalbatross (changes requested):

stanalbatross (approved):

PR #953, opened 2022-08-27
New synthvivor loadouts & Fiorina spelling fixes

JohnDoe50974:
## About The Pull Request
Adds three new synth survivor loadouts to the pool the game randomly selects from upon spawning - the firefighter, the UA military synthvivor with an unique dogtag variant and carrying misc low-grade spare ammo and medicine, and a PMC support synthetic with PMC ID and IFF. Also fixes code that misspells "Fiorina" as "FLorina"

## Why It's Good For The Game
The synthvivor loadout randomizer system has been received warmly, so further expansion on it is welcome.

## Changelog

:cl: 50RemAndCounting
add: Added a few new synth survivor loadouts that are picked randomly upon spawning as one.
spellcheck: fixed all occurences in code and ingame where Fiorina is spelled as FLorina
fix: fixed synthetics still having the CO headset as their replacement headset in the vendor.
/:cl:

PR #977, opened 2022-09-01
Readds comm channel prefixes to radio interfaces

JohnDoe50974:
## About The Pull Request
Renamed all the marine channels + the CL-specific WY channel to include the proper prefix the player needs to type to use said channel, like it was when the radio interface still used NanoUI. Other channels were not touched, as to my knowledge the ERT channels all use :h anyways.
image

## Why It's Good For The Game
Turns out having IC radio channel prefixes viewable was massively important to many players, as I've seen a LOT of complaints about this, both from newer and older players. The amount of info new and occasional players get bombarded it can be overwhelming, so helpful hints and pointers like this one are always welcome, case in point TGUI radios.

## Changelog

:cl: 50RemAndCounting
qol: Radio channel prefixes have been readded! You can left click on any radio headset to open up its interface and view the exact preset you need to input in the Say bar in order to use that channel.
/:cl:


code/__DEFINES/job.dm
```diff
@@ -3,12 +3,12 @@
 #define GET_DEFAULT_ROLE(title) (RoleAuthority?.default_roles[title] ? RoleAuthority.default_roles[title] : title)
 
 // Squad name defines
-#define SQUAD_MARINE_1					"Alpha"
-#define SQUAD_MARINE_2					"Bravo"
-#define SQUAD_MARINE_3					"Charlie"
-#define SQUAD_MARINE_4					"Delta"
-#define SQUAD_MARINE_5					"Echo"
-#define SQUAD_MARINE_CRYO				"Reserves"
+#define SQUAD_MARINE_1					":a Alpha"
```
morrowwolf, line 6: These are used *all over the place* in the code. Find a better way to do this.

code/__DEFINES/radio.dm
```diff
@@ -21,27 +21,27 @@
 
 //Radio channels
 #define RADIO_CHANNEL_ALMAYER "Almayer"
-#define RADIO_CHANNEL_ALAMO "Alamo"	
-#define RADIO_CHANNEL_COMMAND "Command"
+#define RADIO_CHANNEL_ALAMO "Alamo"
+#define RADIO_CHANNEL_COMMAND ":v Command"
```
morrowwolf, line 25: These aren't used outside of radios and the define (as far as I can tell) *but* this should be done better.

tgui/packages/tgui/constants.js
```diff
@@ -185,7 +185,7 @@ export const RADIO_CHANNELS = [
     color: '#1ecc43',
   },
   {
-    name: 'Common',
+    name: ':h Common',
```
morrowwolf, line 188: This also feels funny.

morrowwolf (changes requested):
My recommendation is to work directly in the TGUI code for headsets/radios and create a new constant or similar. This PR right now isn't the way to do this.

Also in the future atomize your PRs. This should probably be two different pull requests.

tgui/packages/tgui/constants.js
JohnDoe50974: This is how I'd prefer to do it, since logically the radio channel names should be pulled from this list, but they're pulled from the defines instead.

JohnDoe50974:
Moved the synth headset bugfix to https://github.com/cmss13-devs/cmss13/pull/953 as I'll need more time to figure this out and the linked PR is ready for review.

morrowwolf:
> Moved the synth headset bugfix to #953 as I'll need more time to figure this out and the linked PR is ready for review.

Brother, just move it to its own PR so it can get added easier.

Wintermote:
This looks really good. It's not obvious to most players imo, that you can shift+click to examine the headset and see what channels it has. This feels like a more elegant and obvious solution.

PR #980, opened 2022-09-01
Fourth unholster keybind and unholstering refactor

morrowwolf:
## About The Pull Request

Refactors the unholstering code to be less bad. Now should cycle through slots without things like secondary and tertiary both picking up the same pouch weapon when you have a knife in your boot you want to draw with tertiary.

Also adds quaternary unholster for the insane people (like myself) who take way too many guns. This is left unbound by default.

Order for checks is suit, belt, back, pockets, uniform, shoes. This has not been changed from before.

## Why It's Good For The Game

Better QoL is good.

## Changelog

:cl: Morrow
qol: Adds fourth unholster keybind option
refactor: Makes the unholster code better.
/:cl:


JohnDoe50974:
Blessed PR

PR #984, opened 2022-09-01
ammo boxes can no longer be carried on your back

stanalbatross:
## About The Pull Request

this PR removes the ability to carry ammo boxes on your back.

The ability of marines to do this has significantly simplified ammo transport and storage for them to the point that supply dumps such as hydro and the fob are rarely returned to for restocking

in order to make the murderball less self-sufficient, the ability of marines to be walking ammo dumps must be nerfed somewhat, and as such I am making this PR

## Why It's Good For The Game

This PR is good for the game because it will make the role of req more important in ensuring that marines will not be able to easily transport around all the ammo they could ever use in a round on their backs. It will also make the backline more populated and interesting as marines will more frequently need to resupply, as they cannot carry 600 rounds / like 15 mags on their backs at all times. This will reduce the individual self-sufficiency of marines and make them require the assistance of req and logistics more, rather than only needing to ask for a cardboard sheet at the start of the round and not worrying at all about resupply from that point forth.

The game is improved by increased interaction between shipside and groundside, increased importance of req, and increased movement around the map as well as increased importance of forwards firebases such as hydro.

As such this PR is good for the game.

## Changelog

:cl:
balance: ammo boxes can no longer be carried on your back
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_boxes.dm
+++ b/code/modules/projectiles/ammo_boxes.dm
@@ -245,7 +245,6 @@
 /obj/item/ammo_box/magazine/shotgun
 	name = "shotgun shell box (Slugs x 100)"
 	icon_state = "base_slug"
-	flags_equip_slot = SLOT_BACK
 	overlay_ammo_type = ""
 	overlay_gun_type = "_shells"
 	overlay_content = "_slug"
@@ -304,7 +303,6 @@
 
 /obj/item/ammo_box/magazine/ap
 	name = "magazine box (AP M41A x 10)"
-	flags_equip_slot = SLOT_BACK
 	overlay_ammo_type = "_ap"
 	overlay_content = "_ap"
 	magazine_type = /obj/item/ammo_magazine/rifle/ap
@@ -314,7 +312,6 @@
 
 /obj/item/ammo_box/magazine/le
 	name = "magazine box (LE M41A x 10)"
-	flags_equip_slot = SLOT_BACK
 	overlay_ammo_type = "_le"
 	overlay_content = "_le"
 	magazine_type = /obj/item/ammo_magazine/rifle/le
@@ -324,7 +321,6 @@
 
 /obj/item/ammo_box/magazine/ext
 	name = "magazine box (Ext M41A x 8)"
-	flags_equip_slot = SLOT_BACK
 	overlay_ammo_type = "_ext"
 	num_of_magazines = 8
 	magazine_type = /obj/item/ammo_magazine/rifle/extended
@@ -334,7 +330,6 @@
 
 /obj/item/ammo_box/magazine/incen
 	name = "magazine box (Incen M41A x 10)"
-	flags_equip_slot = SLOT_BACK
 	overlay_ammo_type = "_incen"
 	overlay_content = "_incen"
 	magazine_type = /obj/item/ammo_magazine/rifle/incendiary
@@ -344,7 +339,6 @@
 
 /obj/item/ammo_box/magazine/explosive
 	name = "magazine box (Explosive M41A x 10)"
-	flags_equip_slot = SLOT_BACK
 	overlay_ammo_type = "_expl"
 	overlay_content = "_expl"
 	magazine_type = /obj/item/ammo_magazine/rifle/explosive
@@ -357,7 +351,6 @@
 /obj/item/ammo_box/magazine/m39
 	name = "magazine box (M39 x 12)"
 	icon_state = "base_m39"
-	flags_equip_slot = SLOT_BACK
 	overlay_ammo_type = "_reg"
 	overlay_gun_type = "_m39"
 	overlay_content = "_hv"
@@ -409,7 +402,6 @@
 /obj/item/ammo_box/magazine/l42a
 	name = "magazine box (L42A x 16)"
 	icon_state = "base_l42"
-	flags_equip_slot = SLOT_BACK
 	overlay_gun_type = "_l42"
 	num_of_magazines = 16
 	magazine_type = /obj/item/ammo_magazine/rifle/l42a
@@ -459,7 +451,6 @@
 /obj/item/ammo_box/magazine/M16
 	name = "magazine box (M16 x 12)"
 	icon_state = "base_m16"
-	flags_equip_slot = SLOT_BACK
 	overlay_ammo_type = "_reg"
 	overlay_gun_type = "_m16"
 	num_of_magazines = 12
@@ -535,7 +526,6 @@
 /obj/item/ammo_box/magazine/m4a3
 	name = "magazine box (M4A3 x 16)"
 	icon_state = "base_m4a3"
-	flags_equip_slot = SLOT_BACK
 	overlay_ammo_type = "_reg"
 	overlay_gun_type = "_m4a3"
 	num_of_magazines = 16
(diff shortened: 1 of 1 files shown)


FoxyStalin:
I don't think being able to carry around ammo boxes is why we don't see Hydro supply drops anymore.
We don't see Hydro Supply drops for the same reason we don't see Bar holds on Solaris anymore,  fortifying those places isn't worth the material drain because cades are so easy to breech unless the entire marine force is holding them.

And even then, Req has the ability to give frontline supply drops any time they can get an SO to listen to them.  People call for resupply at Containers or Filt Entrance all the time. I don't think this is going to be that big of a deal, but, it's another nerf that feels like it's without a point.

Wintermote:
> I don't think being able to carry around ammo boxes is why we don't see Hydro supply drops anymore. We don't see Hydro Supply drops for the same reason we don't see Bar holds on Solaris anymore, fortifying those places isn't worth the material drain because cades are so easy to breech unless the entire marine force is holding them.
> 
> And even then, Req has the ability to give frontline supply drops any time they can get an SO to listen to them. People call for resupply at Containers or Filt Entrance all the time. I don't think this is going to be that big of a deal, but, it's another nerf that feels like it's without a point.

Seriously, this is a PR change being made without actually understanding the problem.

I don't even think this "problem" exists to begin with. If anything, marines are already in the position that they can't carry enough useful ammo to be combat effective for more than 3-4 engagements.

In fact, getting ammo for the M41 in particular is ridiculous right now and it is almost never actually available in fob or elsewhere and THAT is why people are trying to carry as much ammo on themselves as possible. You HAVE to. If you try to rely on fob supplies you're just going to find it completely empty and forced to go back up to the ship for ammo - something I've had to do way too often already. That's before even getting to resupply areas being difficult to keep safe in the first place.

This isn't fixing anything, not even what its purported to fix.

Kenju22:
I'm curious to see how it plays out. As is nobody bothers going to FOB for supplies and FOB has no reason to send people with supplies because marines just carry a squads worth of ammo individually.

Plus it does open the door to possibilities, each marine carrying less ammos means one could argue for incendiary ammo being purchasable with points (even if expensive) or directly from Req (again even if expensive) instead of rolling the dice with Surplus ammo crate.

At best, we might see RTO gain more value with their ability to get ammo delivered to the front line.

4hands44:
This wont fix the problem. Mainly because the problem isn't there. The Reason nobody goes back to FOB or hydro for resupply is because there ain't any ammo there to begin with.  

With the nerfs to other weapons marines have less reason to return to FOB(or forward positions) because its all m39 and Shotty ammo, or AP, which doesn't work when the m4a1mk2 or HPR with HV ammo has become the weapon setup 70% of marines are using. Rather than remove the ability to carry more ammo alone. The solution is to fix the economy issue with REQ that was created by making the Mk2 and HPR the weapons of choice for everyone, without increasing the availability of their main ammo (10x24 HV).

This then all compounds on top of the fact forward mini FOB rally bases are not viable with current cade stats-Xeno dmg ratios, unless the entire platoon agrees to frontline Cadehug. Meaning Hydro (or map equivalent) I'snt worth trying to stay around and wait for ammo, since it's never going to hold unless REQ sends all the metal there instead of the LZ FOB. Instead unga or die is the current meta. Thus sending ammo to the front without need ends in it being wasted/lost when the point gets overrun early because all the marines pushed hard to avoid a battle of attrician they will be unlikely to win. 



TL:DR

Nobody goes back to FOB for ammo, not cause they can carry all the ammo they ever need, but because there isn't any for the guns they have been moved towards using with current meta/stat alterations made over the last year and a half. Combined with not nonstop ungaing towards hive in current meta results in FOB sieges marines usually can't win, thus retreating for resupply is sub-viable most of the time, even if there was any ammo to use.


Hell today I couldn't find any ammo for the mk2 in all of FOB for 20 minutes because it had already been spent, and REQ was broke/couldn't send any. Carrying too much ammo isn't a thing currently.

FlamermanGrills:
Yeah like other people have said before the reason people are carrying as much ammo as they can at the moment is because they don't won't to roll the dice of whether or not M41 ammo/mags are at the FOB/local hold point. If you wait 10/ 15 minutes into the round to check for ammo at FOB all your gonna find is M39 and pistol ammo no one wants while you now have to either go and refill on the ship or pray that req sends down more of your type of ammo.

kiVts:
I must remind you that, during highpop, the only way marines can win, is by murderball in most cases.(other cases include sadar, stims, apc,tank. Literally any way marines can outsmart(expect early tank, that was just busted) xenos will get nerfed one way or another) 
balance literally circles around it.

But I wont bother argumenting, let them have their "perfect cm" bullshit balancing and lets see the world burn, I brought popcorn 🍿

Lagomorphica:
You can also carry 7.5 magazines worth of ammo in a single armor slot without having to give up your entire backpack, in exchange for having to spend some downtime repacking your magazines (about like, 5 seconds), using nothing but your roundstart vendors in squad prep.

Maybe for shotguns this alters how they play but for the M41A MK2, M41A MK1, L42A, and any other gun firing standard 10x24mm rounds it's not really changing anything. Maybe you get less AP ammo for them, I dunno. Frontline supply drops seem common enough in my experience.

I also don't think it'll make forward firebases more common. The issue with them is that nobody wants to sit back and guard them and it's a waste of materials. For a proper Hydroponics forward base that will survive any significant contact, you also have to prevent the xenomorphs being able to just circumvent and go around it, ala cading all of Hydroponics Road - typically the most that is done is a small triage area for medics to fall back too or cading the cave entrances so that xenos can't push aggressively.

You would also need Bravo to regularly patrol this area and prevent it from being breached because each time a cade is damaged its 4 metal, and the xenomorphs can just acid it and run away.

Don2137:
Idk what you people are smoking I rarely see fobs without rifle ammo. And you can make a box yourself pre drop anyway if it bothers you so much?

Anyway this change doesn't mean much when you can just refill from HPR mags.

Weuyn:
No ammo dumps at hydro because there's barely enough metal for the LZ as is

remuluson2:
Do you know what fuel murderballs? SG ammo.

SG has SO MUCH AMMO. That they can push all the way to the hive.

Don2137:
> Do you know what fuels murderballs? SG ammo.
> 
> SG has SO MUCH AMMO. That they can push all the way to the hive.

Also true but unrelated to the pr

Unknownity:
personal anecdotal experience, the usage of being able to carry ammo boxes on your back from my point of view has only ever been used to carry ammo for your HPR during the time when their ammo boxes couldn't fit in your standard pouches/belts so you had much less ammo available and the ammo box helped with that. Outside of that, the only other time of me seeing marines with carrying around ammo boxes was MOU users but they carried the boxes in their hands, not on their back.

so TLDR, this feature is honestly so niche that you actually rarely ever see it being used, most players probably don't even know that you can do it. I seriously doubt carrying ammo boxes on your back actually have to do anything with forward bases not being used (they are used) People just have too much ammo, to begin with then.
(lets not ignore that this will basically change nothing since people will just now carry the boxes with their hands now, which most still do or use the APC to bring the ammo to the front)

spartanbobby:
No I dont think so pal

RoboticVodka:
This is a blatant removal of soul. You're trading off a large amount of space/another weapon for this box you have to you place on the ground to use anyways. If you believe people are somehow actually using this as some kind of meta strategy(which no one is, basically only people LARPing with the HPR carry this and they don't even need the ammo most of the time) they'll just carry the ammo box in hand and drop it a tile away from where they're fighting.

I have never legitimately seen someone use the ammo box to some great effectiveness, or to otherwise ignore the entire round. You are removing soul also because I like to carry around the box on my back and supply my squad ammo as an ammo bearer(soulful interaction btw).

Strobia:
rip psoule.

stanalbatross:
gonna close will address via nerfs to hpr mags

PR #999, opened 2022-09-05
Pressurized pouch change and addition

Don2137:
## About The Pull Request
This PR changes the contents of the "Revival Mix" pressurized pouch (peridaxon replaced with tricordrazine) and adds a new pouch containing tricordrazine.

## Why It's Good For The Game
Pressurized pouches are very cool and soulful but generally rarely used due to the current inventory meta as well as some other issues. I would like to try to partially change that, both by resolving one issue and by adding a new option.

**Revival mix:** Revival mix is a good idea on paper but there is a single issue with it: peridaxon. You use the mix on a dead guy, heal him up to full, a minute later he's passed out on the floor because he had organ damage and neither of you could find that out since peridaxon prevents organ damage symptoms.

There's several solutions to this issue but I believe replacing it with tricordrazine is the best one. It fits with the "revival" theme (by virtue of healing all damage types), can be used as a bootleg peridaxon alongside dexalin/dylovene (thus acting as an ok enough replacement), and it gives medics greater access to a chemical they seldom use. Also, the changed chemical ratios mean you can now double inject revival mix without ODing (currently this is how all injectors and pills aside from emergency work).

**Tricord pouch:** Tricordrazine is a niche/situational/generalist chemical that's occasionally used by medics to supplement other reagents. It's somewhat limited due to only being available in injectors, and having access to a pouch filled with it allows players to specialize into this kind of playstyle much easier (and it's a lot more interesting than adding standard tricord pills). It also doesn't change any medical balance, since tricord itself is already perfectly balanced.

## Changelog

:cl: Don
add: Medical personnel now have access to pressurized pouches containing tricordrazine in their vendors.
balance: Revival Mix pressurized pouches now come with tricordrazine instead of peridaxon; the contents are now evenly split between the three reagents.
/:cl:


stanalbatross:
No linebreaks in changelog without prefix

code/game/objects/items/storage/pouch.dm
```diff
@@ -842,12 +842,18 @@
 
 /obj/item/storage/pouch/pressurized_reagent_canister/revival/Initialize()
 	. = ..()
-	inner.reagents.add_reagent("adrenaline", inner.volume/2.5)
-	inner.reagents.add_reagent("inaprovaline", inner.volume/2.5)
-	inner.reagents.add_reagent("peridaxon", inner.volume/5)
+	inner.reagents.add_reagent("adrenaline", inner.volume/3)
+	inner.reagents.add_reagent("inaprovaline", inner.volume/4)
+	inner.reagents.add_reagent("tricordrazine", inner.volume/3)
```
stanalbatross, line 847: don't tweak around the other numbers w/o documenting in cl

moltoretardo:
I think that perixadon is perfect to stabilize a patient after he has been revived.
Looking for the signs of organ damages is one of the last thing that requires a tiny bit of judgement as a medic, I'd say let's not dumbify it even more.

Superwelder:
> Pressurized pouches are very cool and soulful but generally rarely used due to the current inventory meta as well as some other issues.

This is just not true. The revival pouch remains one of the most popular pocket slot choices for all medics due to covering 3 separate needs:epi to assist the defib, inaprov to stop choke out, and peridaxon usually to stop oxy damage accumulation and allow faster shake ups. The reason the other pouches are not used is because they suck. They cover a single need but require vast space relative to the other options and are easy OD bait. For instance, would you rather have a pill bottle with 15u bicard x 16 pills for a total of 240u bica that you can store in your belt and replace easily or a 300u bicard pocket slot filling pouch that is not easy to replace? The revival pouch is more like 3 useful pill bottles in a pocket slot which is fair and not far off from using a medical pouch to store pill bottles anyways. Tricord in the revival pouch would be trivial. We are talking about a grand total of  ~10 damage healed by that quantity of tricord. Not to mention tricord OD is far more serious than peri OD.
> Revival mix: Revival mix is a good idea on paper but there is a single issue with it: peridaxon. 

This is also not true. The only issue I have come across with the revival pouch has been multiple medics ODing the patient on epinephrine with multiple shots but even then, epi is one of the quickest chems to metabolize out of OD in the game. Peridaxon serves a very strong role in getting the patient stable as fast as possible. It is on the medic to understand the chems they are putting into people. In this case, understanding that giving peridaxon before or just after reviving would abate any organ damage effects. If anything, this is a strong nerf to the revival pouch due to medics being forced to now go through the process of giving peridaxon pills before the other chems and kits they apply so their patient does not die on them.

Don2137:
> I think that perixadon is perfect to stabilize a patient after he has been revived. Looking for the signs of organ damages is one of the last thing that requires a tiny bit of judgement as a medic, I'd say let's not dumbify it even more.

I'm not sure how you're supposed to look for signs of organ damage when peridaxon removes them. The buildup is the only indication you'll get, there's no other sign. And the patient certainly won't spend a full minute next to you waiting for it to run out.

Alardunx:
> > I think that perixadon is perfect to stabilize a patient after he has been revived. Looking for the signs of organ damages is one of the last thing that requires a tiny bit of judgement as a medic, I'd say let's not dumbify it even more.
> 
> I'm not sure how you're supposed to look for signs of organ damage when peridaxon removes them. The buildup is the only indication you'll get, there's no other sign. And the patient certainly won't spend a full minute next to you waiting for it to run out.

For the record, back when I implemented the Revival Mix, the Peridaxon was meant to be a bit of a poisoned gift. While it helps get the marine back up and running immediately, the Peridaxon's hiding of the organs wound means marine have to fallback or gamble. You could always assume they've organ damage if they've broken bone but...

The peridaxon quantity was previously reduced, but some amount of Peridaxon is still handy in getting the marine to atleast walk under his own power ASAP.

Could always reduce it further, but I don't necessarily agree with no peridaxon.

(The poisoned gift bit is because it's meant to reduce the pouch power a little bit, since large amount of Epinephrine are a little difficult to acquire under normal circumstances)

Don2137:
Well if the original contributor says it's supposed to be a poisoned gift then there's not much left to discuss. I'll leave it up to nanu to decide.

Alardunx:
> Well if the original contributor says it's supposed to be a poisoned gift then there's not much left to discuss. I'll leave it up to nanu to decide.

For the record, you can extract the container from the revival pouch at a chemmaster, thus allowing you to flush the peridaxon and put something else.

That applies to any Pressurized Reagent pouches

Superwelder:
Alardunx, do you remember what the revival pouch had before you swapped it to peridaxon? I remember there was a different chem and nobody used the pouch because it ODed very often but the peridaxon was a great replacement.

Alardunx:
> Alardunx, do you remember what the revival pouch had before you swapped it to peridaxon? I remember there was a different chem and nobody used the pouch because it ODed very often but the peridaxon was a great replacement.

It always had peri, but the ratio were changed from 3/3/3 to 5/5/2.5, if my memory is correct.

It used to OD on peri a lot otherwise

Superwelder:
Ah yes I remember now. Was rough since peri lasts a hot second and back then we did not have graft or surgical line so we spammed defib to save massive damage patients, often requiring multiple shots.

lcarapaica:
Revival is the most used medic pouch, peri fits well within the "revival" aspect of it while trico doesn't, don't see a need for this change, medics having to balance their loadouts to fit trico has always been one of the very few  judgements medics have to make.

Wintermote:
I don't think it's a good idea to change up the revival pouch

Better would be to do something to the bicard/kelo pouches so I have some reason to actually use them.

Alardunx:
> I don't think it's a good idea to change up the revival pouch
> 
> Better would be to do something to the bicard/kelo pouches so I have some reason to actually use them.

I mean, we could improve them to use Meralyn and Dermaline respectively. *shrug*

stanalbatross (approved):

morrowwolf:
Merging in 24 hours if GA does not weigh in.

Superwelder:
There is no PR Feedback page for this particular PR in the Discord as it predates the migration so I will comment on my experience here. After having played multiple rounds with this I can say the changes in this PR are far, far inferior to having peridaxon in the revival pouch. Patients are far slower to treat in combat and the extra step of feeding a peridaxon pill slows the process substantially. The most common organ damage medics come across is minor heart damage from defibbing that causes a consistent ~20 oxy damage. The peridaxon inside the revival pouch helped enormously with this problem. The tricord amounts to almost no healing to the point where the pouch would be better served with its removal. Please consider reverting this change for medic sanity.

PR #1007, opened 2022-09-05
Airlock Shock fixes and consistency

morrowwolf:
## About The Pull Request

Alright so I investigated the shocking issue.

First, areas without real power sources that still have pseudo power would not shock people. This fixes that.

Second, there is a probability attached to whether mobs get shocked when messing with shocked stuff. Xenos had a 70% chance to get shocked when attempting to open shocked airlocks but 100% chance when bumping into the door to open it. I've made this a consistent 100% across the board for sanity's sake.

## Why It's Good For The Game

Bugs bad, consistency good

## Changelog

:cl: Morrow
fix: Shocked doors without a real power source will now shock correctly.
fix: Xeno shock chances made consistent across ways to open airlocks.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
+++ b/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
@@ -508,7 +508,7 @@
 /obj/structure/machinery/door/airlock/attack_alien(mob/living/carbon/Xenomorph/M)
 	var/turf/cur_loc = M.loc
 	if(isElectrified())
-		if(shock(M, 70))
+		if(shock(M, 100))
 			return XENO_NO_DELAY_ACTION
 
 	if(!density)
--- a/code/modules/power/power.dm
+++ b/code/modules/power/power.dm
@@ -302,47 +302,57 @@
 	if(istype(power_source,/obj/structure/cable))
 		var/obj/structure/cable/Cable = power_source
 		power_source = Cable.powernet
-
-	var/datum/powernet/PN
-	var/obj/item/cell/cell
-
-	if(istype(power_source,/datum/powernet))
-		PN = power_source
-	else if(istype(power_source,/obj/item/cell))
-		cell = power_source
-	else if(istype(power_source,/obj/structure/machinery/power/apc))
-		var/obj/structure/machinery/power/apc/apc = power_source
-		cell = apc.cell
-		if (apc.terminal)
-			PN = apc.terminal.powernet
-	else if (!power_source)
-		return 0
-	else
-		log_admin("ERROR: /proc/electrocute_mob([M], [power_source], [source]): wrong power_source")
-		return 0
-	if (!cell && !PN)
-		return 0
-	var/PN_damage = 0
-	var/cell_damage = 0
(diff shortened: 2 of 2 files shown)


gohdatrice:
As a xeno player I will point out that having a 100% chance of getting shocked does change the balance and affect gameplay. Now if you are stuck in a room with shocked doors and you can't reach the APC you are completely fucked until someone comes and melts them, whereas previously you could roll your chances on clawing the door open. I have been in that situation before.
Basically what I'm saying is this is more than a "fix", this is removing a feature. I'm no developer but I'm not even sure the chance of clawing open the door was ever a bug in the first place.

Wintermote:
> As a xeno player I will point out that having a 100% chance of getting shocked does change the balance and affect gameplay. Now if you are stuck in a room with shocked doors and you can't reach the APC you are completely fucked until someone comes and melts them, whereas previously you could roll your chances on clawing the door open. I have been in that situation before. Basically what I'm saying is this is more than a "fix", this is removing a feature. I'm no developer but I'm not even sure the chance of clawing open the door was ever a bug in the first place.

I don't think it's a big issue. The situations where a xeno gets trapped in a room with shocked doors and isn't already about to be killed by marines is something exceptionally rare. Most rooms either have an APC accessible in that room, or have a vent or window to get out of the room too.

Additionally there's a whole bunch of xeno castes that can either acid an opening through a wall to get in/out, or can just outright break down the wall/aren't affected by the shocks. So simple communication and teamwork can save you in the worst case scenario.

The only situation I can see the 30% chance of opening without getting shocked being nice to have is when trying to sneak into a secured area, which seems to me like it completely defeats the point of shocking the door to begin with.

stanalbatross (approved):

gohdatrice:
At the very least I suggest you change the changelog to mention that you've removed the possibility to claw open shocked doors. "Xeno shock chances made consistent across ways to open airlocks" does not communicate very well that you've completely removed that feature. It feels like you're trying to hide what you've done from people. Something like "Removed the ability for xenos to sometimes claw open shocked doors" would be way more clear. Changes like this shouldn't be secret or you'll have xenos bashing their heads against shocked doors over and over without realising that doesn't work anymore.

gohdatrice:
I know I have no influence here but I genuinely want to know how anyone thinks this is an acceptable changelog. It does not come even close to actually conveying what you have changed. Nobody reading this changelog in-game would know that you have removed the ability for xenos to claw open doors.
Now it makes me wonder, how many other changelogs are bullshit? Do I have to read the github page for every single Morrow change now or else I'll have no idea what secret changes you're doing that aren't conveyed by the changelog? Why even have a changelog in game if you can just write whatever bullshit you want instead of actually conveying what you've changed?
My guess is that you're doing this to increase your fix score or whatever it is, I'm not sure how this works. But I don't think you should be sacrificing changelog clarity just for github points.
Again, the reason I'm posting this is because I genuinely have no idea how anyone thinks this changelog accurately reflects the changes made. If there's something I'm missing that makes this make sense then please explain it to me.

PR #471, opened 2022-06-14
Redesigns MP armory.

PhantomEpicness:
## About The Pull Request
This PR is primarily as a QoL as it is mainly meant to fix the issue of this: 
image
Having 9 morbillion items spill out the second you open the racks and lag your game to shit trying to select the things you want is horrible. Plus the sheer item density of these racks is likely a minor contributor to late-game lag. 

So I replaced them with vendors that _**SHOULD**_ have the exact same content of the racks. A light redesign of the armory was also required as its equally as bad as the racks.
image

I also went about and put a door here so MPs can take a shortcut through the holding cells instead of going around everything
image

 The M81 (M74's long lost shitty cousin) also can shoot riot slugs now, which I'm pretty sure is it's loreful intention instead of bullying warriors.

## Why It's Good For The Game

Much needed QoL & redesign. Replaces a shitty system.

## Changelog

:cl: TotalEpicness
tweak: M81 Riot Grenade Launcher (M74's long lost brother) can now shoot baton slugs.
qol: Redesigned brig armory.
qol: Added a door to Holding cells as a shortcut to processing.
balance: rebalanced something
fix: Fixed the massively overloaded & lag causing racks in MP armory by replacing them with a vendor.
/:cl:

PR #1010, opened 2022-09-06
Redesigns MP Armory

PhantomEpicness:
## About The Pull Request
funny github wont let me reopen!!!!

This PR is primarily as a QoL as it is mainly meant to fix the issue of this: 
image
Having 9 morbillion items spill out the second you open the racks and lag your game to shit trying to select the things you want is horrible. Plus the sheer item density of these racks is likely a minor contributor to late-game lag. 

So I replaced them with vendors that _**SHOULD**_ have the exact same content of the racks. A light redesign of the armory was also required as its equally as bad as the racks.
image

I also went about and put a door here so MPs can take a shortcut through the holding cells instead of going around everything
image

 The M81 (M74's long lost shitty cousin) also can shoot riot slugs now, which I'm pretty sure is it's loreful intention instead of bullying warriors.

## Why It's Good For The Game

Much needed QoL & redesign. Replaces a shitty system.

## Changelog

:cl: TotalEpicness
tweak: M81 Riot Grenade Launcher (M74's long lost brother) can now shoot baton slugs.
qol: Redesigned brig armory.
qol: Added a door to Holding cells as a shortcut to processing.
fix: Fixed the massively overloaded & lag causing racks in MP armory by replacing them with a vendor.
/:cl:


morrowwolf:
Can you take a minute to talk about how you've addressed issues brought up in the previous PR? (https://github.com/cmss13-devs/cmss13/pull/471)

PhantomEpicness:
All MPs should be able to access the vendors themselves.

silencer-pl:
So a few things.

First, the bug you're fixing here was already fixed. I don't think the lockers contain excess M39s anymore.

Second, The entire idea behind a holding cell is that it can be only opened and entered from one direction (so only one MP is needed to effectively guard it during processing). Putting another door on it makes it more of a room than a holding cell. Having to go literally to the next room on either side seems like a little price to pay for the cells and processing area not exposed from multiple sides. All that door is doing is providing an additional easy escape avenue to potential JBers AND lets people easily infiltrate the inner brig from the medbay area. Seems like its creating multiple security hazards for the sake of saving 5 seconds.

I also repeat my criticism from the previous PR - putting in vendors with an unbreakable APC (unbreakable why?) seems like its going to stand out in the middle of the brig like a sore thumb and seems like a completely unnecessary workaround considering the lockers already do their job well.

morrowwolf:
Let me break down this PR real quick.

First, you have negative GBP and this doesn't feel like a pure fix.
Second, I'm gonna quote contributor guidelines regarding what we don't want with mapping: "Major remaps without prior approval". So essentially, get dev approval before re-opening regarding mapping changes to the brig.
Third, I feel like you're making this fix more problematic than it needs to be. A simple reduction to the amount of stuff per locker would work. The change of theme for brig red alert room as well as an APC that will always have power post-hijack is jarring.
Fourth, atomize your PR. Split these changes into their own PRs so they can be judged individually on their merits.

As it stands, I will be closing this unless overruled. I do hope we see you soon and I don't want to discourage you from contributing.

PR #1021, opened 2022-09-08
fix non-intel items not being destroyed by spawn pool / eggmorpher

### LINKED ISSUE #1017: Sentries in bursted bodies get deployed when pooled
## Description
When a body that was impregnated is bursted and pooled, if it had a sentry in their backpack, that sentry will get deployed once the body finishes pooling leaving a sentry set up right on top of the pool. If its a flamer sentry, this is basically just death for the pool if the sentry gets damaged enough to perform its explosion.

Sentries in pooled body inventories should just delete when pooled - not deploy (a vat of whatever that's able to disintegrate a body and most of its equipment in seconds isn't going to melt a sentry too?)

## Steps to reproduce
(Maybe can be optimized but I don't know what vars are all needed or proc set up a body to be bursted)
(Edited because it seems deployment of the sentry must have occurred before somereason)
1. Be a marine (RTO equipped in my case)
2. Spawn a handheld flamer sentry
3. Deploy that sentry
4. Undeploy it with a access tuner
5. Put the sentry in backpack
6. Get hugged
7. Change ckey to xeno
8. Allow marine to burst
9. Pool their body
10. Sentry:
need a sentry here

- [X] Issue could be reproduced at least once (confirmed in localhost from master on 3Sep)

stanalbatross:
## About The Pull Request

only unacidable INTEL items should not be deleted

everything else should still be deleted 

closes #1017 

## Why It's Good For The Game

bugfix + random spec gear should not be getting saved from the spawn pool. Intel shit is only saved because it is meant to NEVER be destroyed by xenos

## Changelog

:cl:
fix: only unacidable intel items are saved from the spawn pool now, this fixes a number of issues with unacidable items such as sentries not being deleted properly
fix: the spawn pool and egg morpher will no longer pile all intel items on top of them
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/special/egg_morpher.dm
+++ b/code/modules/cm_aliens/structures/special/egg_morpher.dm
@@ -89,11 +89,11 @@
 				to_chat(user, SPAN_XENOWARNING("\The [src] is full of children."))
 				return
 			if(user)
-				visible_message(SPAN_XENOWARNING("[user] slides [F] back into [src]."), \
-					SPAN_XENONOTICE("You place the child back into [src]."))
+				visible_message(SPAN_XENOWARNING("[user] slides [F] back into \the [src]."), \
+					SPAN_XENONOTICE("You place the child back into \the [src]."))
 				user.temp_drop_inv_item(F)
 			else
-				visible_message(SPAN_XENOWARNING("[F] crawls back into [src]!"))
+				visible_message(SPAN_XENOWARNING("[F] crawls back into \the [src]!"))
 			stored_huggers = min(huggers_to_grow_max, stored_huggers + 1)
 			qdel(F)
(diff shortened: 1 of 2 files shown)


JediKing112:
add dog tags to the list of not being deleted, would be cool for a little RP and knowing if people got capped or cryo'd or whatever happened to them,

Mirvra:
> add dog tags to the list of not being deleted, would be cool for a little RP and knowing if people got capped or cryo'd or whatever happened to them,

This is a really good idea that I don't think anyone's ever really thought of? It might clutter up the pool a bit... But on the other hand as Jedi said it would make for some nice RP to know which marines got pooled, especially if the tags get brought back to memorial.

morrowwolf (approved):
Code LGTM

PR #1043, opened 2022-09-12
Predalien larva sprite fix

morrowwolf:
## About The Pull Request

Forest says this will fix predalien larva not having correct sprites. I was unable to verify this locally.

## Why It's Good For The Game

Bug bad.

## Changelog

:cl: Morrow
fix: Predalien larva sprite works now
/:cl:


morrowwolf:
I assume it's because the sprite is on the server and not locally accessible but I'm not 100% sure.

The runtime should be fixed now.

Carnage068:
If the issue is that the sprite is invisible, it probably IS because it’s not locally accessible. Same shit would happen if you took the Acid Runner strain on Xenoid.

PR #1047, opened 2022-09-12
Cause Data Checks and Gibbing Runtime Fixes

Drulikar:
## About The Pull Request
Adds more type checking to cause_data usage. Runtimes can still occur for cause_data actions, but all should now be manual messages that do not end the proc call.

## Why It's Good For The Game
Less runtimes and more death reason data. E.g. these queen gib runtimes are now gone: 
image

## Changelog

:cl: Drathek
fix: Fixed runtimes caused by gibbing.
code: Added more checks for cause_data usage.
/:cl:


fira (approved):
Looks good to me. About time this was straightened up, thanks for that! I think some of these were fixed in the past but lost in Gitlab PR Hell.

PR #1049, opened 2022-09-12
ID locked item QoL

PhantomEpicness:
## About The Pull Request
ID locked items now unlock when swiped with their owner's ID card after a confirmation prompt.

## Why It's Good For The Game

Removes stupid situations where a spec doesnt want an item, but takes it out of the box, leading to the item being useless. This has notably screwed over many Sniper Spec spotters as their entire kit is useless because the spec touched it.

Yes I know command IDs override, but CO/XO have better things to do than to go all the way down to alpha prep to unlock a VP78 for a PFC.

 Yes I know people usually throw the box, but sometimes that doesn't happen, either by accident or indecision. Currently, it is unintuitive and dumb to be unable to do anything at all, even as the owner, to locked gear.

Now the owner (whoever it is locked to) can swipe their ID to remove the ID locking altogether. 

Edit: A concern was raised that someone could steal someone else's ID and steal their shit

Stealing IDs is already against the rules so this is a non-issue. Added a log_admin just in-case

## Changelog

:cl: TotalEpicness5
balance: ID locked items (ex: items in spec box) can now be unlocked by their owner swiping their ID on it.
/:cl:


DylonBR:
certified balance label moment

code/game/objects/items.dm
```diff
@@ -245,6 +245,14 @@ cases. Override_icon_state should be a list.*/
 // Due to storage type consolidation this should get used more now.
 // I have cleaned it up a little, but it could probably use more.  -Sayu
 /obj/item/attackby(obj/item/W, mob/user)
+	if(istype(W,/obj/item/card/id))
+		if(((flags_item & MOB_LOCK_ON_EQUIP)) && (user == locked_to_mob))
```
stanalbatross, line 249: flag check before istype please

code/game/objects/items.dm
```diff
@@ -245,6 +245,14 @@ cases. Override_icon_state should be a list.*/
...
+			if (alert(user, "Would you like to disable the ID lock for \the [src]. This cannot be undone!", "Confirm", "Yes", "No") == "Yes")
+				to_chat(user, SPAN_WARNING("You unlock \the [src]."))
+				src.flags_item &= ~MOB_LOCK_ON_EQUIP
+				locked_to_mob = null
+				log_admin("[user] has removed the ID lock on \the [src]. Owner of the item was [locked_to_mob] (CKEY: ([user.ckey]))")
```
stanalbatross, line 254: needs improving, this makes it look like the owner's ckey is the user's. Also include JMP link.

stanalbatross (changes requested):

code/game/objects/items.dm
```diff
@@ -245,6 +245,17 @@ cases. Override_icon_state should be a list.*/
 // Due to storage type consolidation this should get used more now.
 // I have cleaned it up a little, but it could probably use more.  -Sayu
 /obj/item/attackby(obj/item/W, mob/user)
+	if(((flags_item & MOB_LOCK_ON_EQUIP)))
+		to_chat_spaced(world,"B")
+		if(user == locked_to_mob && istype(W,/obj/item/card/id))
+			to_chat_spaced(world,"A")
```
stanalbatross, line 251: 💢

code/game/objects/items.dm
```diff
@@ -245,6 +245,17 @@ cases. Override_icon_state should be a list.*/
...
+			if (alert(user, "Would you like to disable the ID lock for \the [src]. This cannot be undone!", "Confirm", "Yes", "No") == "Yes")
```
stanalbatross, line 252:
```suggestion
			if (alert(user, "Would you like to disable the ID lock for \the [src]? This cannot be undone!", "Confirm", "Yes", "No") == "Yes")
```

stanalbatross (changes requested):

PR #1061, opened 2022-09-15
Cryomarine Stuff

morrowwolf:
## About The Pull Request

Re-opening: https://github.com/cmss13-devs/cmss13/pull/905

With permission from Geeves I will be re-opening and maintaining some closed pull requests while he is on break.

## Why It's Good For The Game

"Foxtrot is a cooler name than Reserves, and exists within the actual games. The brown was a bit muddy and made it look like they didn't have a squad at all, so the brighter one resolves that.

Calling the cryomarines via the tech tree is cool, and keeps people in the game in longer games.

The tech points required to transition was too much, I've been told, so knocking it down to 5 seems fine to me."

## Changelog

:cl: Geeves
fix: The Reserves squad is now called Foxtrot. Their brown color has been brightened a bit, the radio color as well. They also no longer break the crew manifest.
add: You can now call the reserves in-game via the tech tree, it's a Tier 3 tech that costs 10 points, but increases by 5 each time it's called. The squad contains a squad leader, medics, engineers, and privates.
balance: The tech points required to go up a tier has been reduced to 5 from 15.
/:cl:


morrowwolf:
I approved all of Geeves' code in the old PR. Just need someone to look over the fix for HUDs.

code/modules/gear_presets/uscm.dm
```diff
@@ -43,6 +43,16 @@
...
 	H.hud_set_squad()
 
+	if(H.wear_l_ear)
+		if(istype(H.wear_l_ear, /obj/item/device/radio/headset/almayer/marine))
+			var/obj/item/device/radio/headset/almayer/marine/equipped_headset = H.wear_l_ear
+			equipped_headset.add_hud_tracker(H)
+	if(H.wear_r_ear)
+		if(istype(H.wear_r_ear, /obj/item/device/radio/headset/almayer/marine))
+			var/obj/item/device/radio/headset/almayer/marine/equipped_headset = H.wear_r_ear
+			equipped_headset.add_hud_tracker(H)
```
stanalbatross, line 53: you can't wear two headsets at once so for optimisation if the left one has a headset you should not check right

Wintermote:
Issue that's been noticed since testmerge, there needs to be some kind of recompensation for cryomarines immediately going back into the cryotubes after being selected as a cryomarine.

I don't know why people do it, but they roll for cryomarine, get cryomarine, and then immediately go back into the cryotube, so we're seeing less than half the cryomarines actually going back into the fight.

Cancan1203:
> Issue that's been noticed since testmerge, there needs to be some kind of recompensation for cryomarines immediately going back into the cryotubes after being selected as a cryomarine.
> 
> I don't know why people do it, but they roll for cryomarine, get cryomarine, and then immediately go back into the cryotube, so we're seeing less than half the cryomarines actually going back into the fight.

I think there needs to be a punishment for those who do this, since I believe they think its a faster way to end the round by denying cryo similar to larva suiciding but thats not really anything to do with this PR and a coding fix is probably easier than enforcing a rule like that.

ZAB909:
> I want to roll standard squad marine
> roll medic
>cryo

Its that simple

>Want to help heal people
>roll engie
>cryo

awesome

morrowwolf:
> > I want to roll standard squad marine
> > roll medic
> > cryo
> 
> Its that simple
> 
> > Want to help heal people
> > roll engie
> > cryo
> 
> awesome

These are options that can be set in your preferences regarding what you roll for in an ERT. This is no excuse.

Wintermote:
>but thats not really anything to do with this PR and a coding fix is probably easier than enforcing a rule like that.

I agree, it needs a mechanic change. I don't think relying on moderator enforcement is going to work very well.

stanalbatross (approved):

stanalbatross:
Fine for now, should make Intel more impactful and help Marines maybe turn around long xeno rolls

Costs subject to tweaking if needed

PR #1085, opened 2022-09-18
Sedative rework + Tranq gun buff

PhantomEpicness:
## About The Pull Request
Reworks sedative so it's actually consistent and works as a sedative. It also is now balanced as its consistent and uses stamina damage to incapacitate rather than sleep due to sleep causing someone to be stunned for an unpredictably long period of time.

Also buffs tranquilizer gun to use a sedative chemical as currently it is shit as seen by Warfan's UPP event on 9/17, the UPP agents were robusted because the tranq gun dealt 30 stamina damage so it didn't even knock people down on one hit and even if they mag dumped them, they would go up in less than a minute

With this rework it now is more realistic in the following ways:

- Each round injects 4u of tranquilizer, lasting no more than one minute and twenty seconds before the victim starts to awake again.
- It takes a few seconds to kick in, meaning the victim can fight back in the precious few seconds they have left.
- It has an OD, causing oxyloss, critical OD causes brain damage.
- Medical can do something about it by flushing the victim's chems. This provides more RP opportunities.

## Why It's Good For The Game
Reworks an untouched and very inefficient chemical property. Makes tranq guns actually tranquilize people

## Changelog

:cl: Totalepicness5
add: Reworked sedative chemical property. It now consistently incapacitates people. 
balance: Tranquilizer pistol now actually injects people with a sedative tranquilizer now.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -314,9 +314,15 @@
 	name = "light pistol bullet"
 
 /datum/ammo/bullet/pistol/tranq
-	name = "tranquilizer bullet"
+	name = "tranquilizer dart"
 	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_IGNORE_RESIST
-	stamina_damage = 30
+	stamina_damage = 15
+	var/amount = 4
+
+/datum/ammo/bullet/pistol/tranq/on_hit_mob(mob/M, obj/item/projectile/P, mob/user) //Special effects when hitting mobs.
+	M.reagents.add_reagent("tranq", amount, , , user)
+	to_chat(M,SPAN_XENOHIGHDANGER("\The [src] injects something into you...."))
+
 
 //2020 rebalance: is supposed to counter runners and lurkers, dealing high damage to the only castes with no armor.
 //Limited by its lack of versatility and lower supply, so marines finally have an answer for flanker castes that isn't just buckshot.
--- a/code/modules/reagents/chemistry_reagents/toxin.dm
+++ b/code/modules/reagents/chemistry_reagents/toxin.dm
@@ -350,3 +350,15 @@
 	reagent_state = SOLID
 	chemclass = CHEM_CLASS_UNCOMMON
 	properties = list(PROPERTY_TOXIC = 1)
+
+/datum/reagent/toxin/traquilizer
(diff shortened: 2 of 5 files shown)


code/modules/projectiles/ammo_datums.dm
```diff
@@ -314,9 +314,15 @@
...
 /datum/ammo/bullet/pistol/tranq
-	name = "tranquilizer bullet"
+	name = "tranquilizer dart"
 	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_IGNORE_RESIST
-	stamina_damage = 30
+	stamina_damage = 15
+	var/amount = 4
+
+/datum/ammo/bullet/pistol/tranq/on_hit_mob(mob/M, obj/item/projectile/P, mob/user) //Special effects when hitting mobs.
+	M.reagents.add_reagent("tranq", amount, , , user)
```
stanalbatross, line 323: use named args instead of null ones

code/modules/projectiles/ammo_datums.dm
```diff
@@ -314,9 +314,15 @@
...
+	to_chat(M,SPAN_XENOHIGHDANGER("\The [src] injects something into you...."))
```
stanalbatross, line 324: ? span_warning

code/modules/reagents/chemistry_reagents/toxin.dm
```diff
@@ -350,3 +350,15 @@
 	reagent_state = SOLID
 	chemclass = CHEM_CLASS_UNCOMMON
 	properties = list(PROPERTY_TOXIC = 1)
+
+/datum/reagent/toxin/traquilizer
+	name = "Tranquilizer"
+	id = "tranq"
+	description = "A synthetic central nervous system depressant that is designed to quickly succumb victims into unconciousness."
```
stanalbatross, line 357: > succumb victims into unconciousness

this does not make sense

stanalbatross (changes requested):

Mrrpip:
Won't it just get purged by Tricordazine anyhow?

PhantomEpicness:
> Won't it just get purged by Tricordazine anyhow?
> 
> Edit:(Meaning they can premed or purge it during the few seconds of non-KO)

I dont think purge chems are standard issue, tricord doesn't purge chems either

PR #1093, opened 2022-09-19
M39 minor rework

Mister-moon1:
## About The Pull Request
Changes the damage fall off to be that of the m41a

## Why It's Good For The Game
The m39 sees little to 0 use outside of very specific circumstances and even when it is used, it's very weak compared to the other weapons marines have at their disposal

## Changelog

:cl:
balance: Decreases m39 damage falloff
/:cl:

PR #1098, opened 2022-09-19
M39 rework the second coming

Mister-moon1:
## About The Pull Request
The M39 long awaited rework to attempt to get it to compete with the other primary weapons has arrived with quite the amount of community talk about what to do with it.

## Why It's Good For The Game
M39 in recent times has always been considered a noob trap seeing very little use outside of very specific circumstances, this PR should help it become more mainstream as a close range DPS gun

## Changelog

:cl:
balance: M39 fire delay buff
balance: M39 damage fall off buff
balance: M39 elite variant match M39 fire rate
balance: M39 mag size changed to be able to fit in pistol mag pouches
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -731,7 +731,7 @@
 	effective_range_max = 4
 	penetration = ARMOR_PENETRATION_TIER_1
 	shell_speed = AMMO_SPEED_TIER_6
-	damage_falloff = DAMAGE_FALLOFF_TIER_5
+	damage_falloff = DAMAGE_FALLOFF_TIER_7
 	scatter = SCATTER_AMOUNT_TIER_6
 	accuracy = HIT_ACCURACY_TIER_3
 
--- a/code/modules/projectiles/guns/smgs.dm
+++ b/code/modules/projectiles/guns/smgs.dm
@@ -72,8 +72,8 @@
 
 /obj/item/weapon/gun/smg/m39/set_gun_config_values()
 	..()
-	fire_delay = FIRE_DELAY_TIER_SMG
-	burst_delay = FIRE_DELAY_TIER_SMG
+	fire_delay = FIRE_DELAY_TIER_10
+	burst_delay = FIRE_DELAY_TIER_10
 	burst_amount = BURST_AMOUNT_TIER_3
 	accuracy_mult = BASE_ACCURACY_MULT
 	accuracy_mult_unwielded = BASE_ACCURACY_MULT - HIT_ACCURACY_MULT_TIER_5
@@ -112,7 +112,7 @@
 
 /obj/item/weapon/gun/smg/m39/elite/set_gun_config_values()
 	..()
-	fire_delay = FIRE_DELAY_TIER_SMG
+	fire_delay = FIRE_DELAY_TIER_10
 	accuracy_mult = BASE_ACCURACY_MULT + HIT_ACCURACY_MULT_TIER_7
 	accuracy_mult_unwielded = BASE_ACCURACY_MULT
 	scatter = SCATTER_AMOUNT_TIER_9
--- a/code/modules/projectiles/magazines/smgs.dm
+++ b/code/modules/projectiles/magazines/smgs.dm
@@ -14,7 +14,7 @@
 	caliber = "10x20mm"
 	icon_state = "m39_HV"
 	max_rounds = 48
-	w_class = SIZE_MEDIUM
+	w_class = SIZE_SMALL
 	gun_type = /obj/item/weapon/gun/smg/m39
 	default_ammo = /datum/ammo/bullet/smg/m39
 


Mister-moon1:
PR will remain a draft until Nanu has reviewed it

kiVts:
oh god the reowkr oh god in dyinf hepa

Mirvra:
Two things:

1, Personally I feel like the M39 is more of a large sidearm to be used by Medics or Engineers, why does it need to be a Primary and compete with the MUCH larger weapons? The only marines I could see lore-wise using M39 as a Primary are basically hot drop marines, or some kind of ODST equivalent. PMC's have the upgraded version because they're not quite as front-line combat intended (outside of the heavy SG).

2, Please be careful not to bring the M39's power level back to where it was ages ago... where it would let you run full speed while firing, straight up deleting xenos without any effort.

Wintermote:
Should pick some sort of niche you want the gun to fill, and then adjust stats towards that, rather than just adjusting stats and seeing where it ends up..

Mister-moon1:
Do note that the actual damage of the m39 has not been touched and the DPS increase from fire rate is only around +18%, also ive tested it, even with stock the accuracy still isnt great and you will miss a majority of shots unless you touch skin

At ranges the m4 should outclass it in every scenario, but where the m39 can shine is close range DPS

Faunenschnitt:
Despite the M39 might no be sufficient right now for your needs, it was often too strong after buffing it and I think we should rather reduce the damage of other weapons then keep buffing all weapons to the current best ones (power creep).

History of buffing M39 since end of 2021

https://github.com/cmss13-devs/cmss13/commit/bc2fc6c38c6bc5c48c85e531dc01e484bc888f9e

https://github.com/cmss13-devs/cmss13/commit/c3d8209727e0a256d54c1168ca04a723f29cc3a7

(e.g. It went from damage_falloff = DAMAGE_FALLOFF_TIER_1 (december 21) to DAMAGE_FALLOFF_TIER_7 (this PR))

(Also I do not thing that the next buff attempt should be made immediately after having the last one closed in a single day
At least reference it: https://github.com/cmss13-devs/cmss13/pull/1093)

Mister-moon1:
The community has talked about it for quite the amount of hours and we agreed that this was the best course of action for the m39, also it will still be better to pick an m41a over an m39 for versatility and mid range combat, just that the m39 has been decided on being for close range dps

One of the other things other people agreed on though is that guns should not be nerfed to make others look better

MusselMain:
> Two things:
> 
> 1, Personally I feel like the M39 is more of a large sidearm to be used by Medics or Engineers, why does it need to be a Primary and compete with the MUCH larger weapons? The only marines I could see lore-wise using M39 as a Primary are basically hot drop marines, or some kind of ODST equivalent. PMC's have the upgraded version because they're not quite as front-line combat intended (outside of the heavy SG).
> 
> 2, Please be careful not to bring the M39's power level back to where it was ages ago... where it would let you run full speed while firing, straight up deleting xenos without any effort.

The issue is that currently it's actually worse than most sidearms in a gunfight, with the only advantage being firerate, (I once was a PMC and got attacked by a disgruntled SO armed with a M4A3, I BARELY beat him, despite wearing armor, and having a SMG that's SUPPOSED to be good at close range, and despite having the XO help me put the rogue SO down, which is clearly a issue).

As a SMG it's supposed to be at least decent in close range, but in it's current state, every single primary the standard USCM PFC starts with is not only better that the SMG in the area it's supposed to accel at, but it also fill their own areas they were designed for, meaning the SMG has almost no purpose aside from being easier to run around with, and even then, scouts nowadays realize that the Shotgun is way better at that job, as it won't miss half as often at 3 tiles.

MusselMain:
> Do note that the actual damage of the m39 has not been touched and the DPS increase from fire rate is only around +18%, also ive tested it, even with stock the accuracy still isnt great and you will miss a majority of shots unless you touch skin
> 
> At ranges the m4 should outclass it in every scenario, but where the m39 can shine is close range DPS

I agree, I personally think that the M39 shouldn't be missing half of it's 48 round magazines at 3 tiles, (as I literally had a easy fight where I had significant advantages, and barely won said gunfight because the M39 missed 20-25 rounds at 2 to 3 tiles) but I also think it shouldn't be anywhere near as versatile as the M41A, the only thing that needs improving is the M39's performance at close range, because if a weapon designed specifically for close quarters combat, is actually worse at that job than almost all of the standard PFC's other primary options, you know it's a major issue (For example: it wouldn't make sense for the M44 to do less damage than it's faster sidearm counterparts, as that would eliminate it's niche/role).

Mister-moon1:
If its so good, why does noone use it? Also other guns are in a good point balance wise right now, even if the accuracy is good with a fully optimised m39 its still a peashooter compared to other guns

LoJoFan:
> If its so good, why does noone use it? Also other guns are in a good point balance wise right now, even if the accuracy is good with a fully optimised m39 its still a peashooter compared to other guns

Because they are afraid to experiment.

Here are some things that might work for you : 

- Use the optimal attachments to reduce scatter and increase accuracy (I'm not going to help you here. Look at the wiki)
- Use the burst fire mode
- Move optimally and choose your engagements wisely

I've used it a fair share and I can safely say that the M39 in its current state is fine. As long as you stay within 4 tiles of whatever you're fighting, you are going to shred them.

Chainsawicus:
> If its so good, why does noone use it? Also other guns are in a good point balance wise right now, even if the accuracy is good with a fully optimised m39 its still a peashooter compared to other guns

Symptom of a community that refuses to accept when nerfs are necessary, ever since it got nerfed down from being ludicrously overpowered people have been adamant that it's useless because their toy got touched. In its current state it serves as a close quarters AP weapon that does more DPS than the m41 whilst having zero slowdown, allowing you to dodge around and move quickly through tight spots.

The gun's perfectly fine as is and needs nothing more. If all your shots are missing, stop using it like an m41/l42 and get closer, and if that doesn't work, aim better, because it won't be missing within 4 tiles.

Nanu308:
I'll be honest, I've reviewed this and the ideasguys thread and i do not really agree with most inputs made other than that pistols are indeed strong and should recieve a nerf.

Besides that, i believe the M39 is fine in it's current spot.

PR #1102, opened 2022-09-20
Anti pred slime

Joelampost:
## About The Pull Request
-stops preds form changing their SD size mid SD (this only works if someone tries to change it from small to big)
-reduces the amount of health gained from the first step of the medicomp by 10 and gave plus 5 to the following steps 
-stops preds stacking multiple mtend wound surgeries  

## Why It's Good For The Game
less slime good

## Changelog

/:cl: Joe lampost
fix: stops preds form changing their SD size mid SD (this only works if someone tries to change it from small to big)
fix: reduces the amount of health gained from the first step of the medicomp by 10 and gave plus 5 to the following steps 
fix stops preds stacking multiple mtend wound surgeries  
/:cl:

### DIFF
--- a/code/modules/cm_preds/yaut_bracers.dm
+++ b/code/modules/cm_preds/yaut_bracers.dm
@@ -623,7 +623,7 @@
 
 	exploding = 1
 	var/turf/T = get_turf(src)
-	if(explosion_type == 0 && victim.stat == CONSCIOUS && is_ground_level(T.z))
+	if(explosion_type == SD_TYPE_SMALL && victim.stat == CONSCIOUS && is_ground_level(T.z))
 		playsound(src, 'sound/voice/pred_deathlaugh.ogg', 100, 0, 17, status = 0)
 
 	playsound(src, 'sound/effects/pred_countdown.ogg', 100, 0, 17, status = 0)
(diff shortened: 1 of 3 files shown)


code/modules/cm_preds/yaut_bracers.dm
```diff
@@ -643,12 +643,17 @@
 	set category = "Yautja.Misc"
 	set src in usr
 
+	if(explosion_type == 1 && exploding)
```
stanalbatross, line 646: why the fuck are these integers make them defines

stanalbatross (changes requested):

code/modules/cm_preds/yaut_bracers.dm
```diff
@@ -643,12 +648,17 @@
...
 
+	if(explosion_type == SD_TYPE_BIG && exploding)
+		to_chat(usr, SPAN_WARNING("Why would you want to do this?"))
+		return
+
 	if(alert("Which explosion type do you want?","Explosive Bracers", "Small", "Big") == "Big")
 		explosion_type = 0
 		log_attack("[key_name_admin(usr)] has changed their Self Destruct to Large")
 	else
 		explosion_type = 1
```
stanalbatross, line 659: lampost when I make you add in defines you know you have to add it for the entire thing right?

stanalbatross (changes requested):
also there is no code in here that stops stacking surgeries

Joelampost:
> also there is no code in here that stops stacking surgeries

it only works if you target chest now so it kinda does

stanalbatross (changes requested):

stanalbatross (changes requested):
you have fucked up

previously 0 was BIG and 1 was SMALL

swap the numbers and change every define you changed

PR #1110, opened 2022-09-22
greenfire tweaks without bad rebase

stanalbatross:
## About The Pull Request
Try 3 on green fire rebalance. A significant amount of flame-related code was refactored by Geeves and myself. The slowdown intensity has been reduced from the slowdown multiplier value of 3 to the multiplier value of 1. The code refactor includes an exception for normally fire-immune xenomorphs too, so they are no longer affected by the greenfire's effects.

## Why It's Good For The Game
The rebalanced greenfire will allow xenomorphs to still be seriously punished for walking on greenfire without almost instantly game-ending them. Escape will still be difficult, but no longer borderline impossible. Fire-immune xenos having their immunity properly apply to greenfire as well ensures that their roles in a combat situation are not unduly infringed upon, as queens and ravagers (as is any xeno caste without a pounce/leap/dash ability tbh, but slow T3s especially so) were the most vulnerable to getting flamed.
Also, code refactors will help improve server performance and possibly help prevent server crashes and code instability. More atom proccalls and bitflags, less for() loops and if statements.

## Changelog

:cl: Geeves, 50RemAndCounting, Stan_Albatross
balance: Type-B fire/greenfire now no longer affects fire-immune xenomorphs. Greenfire slowdown effect has been reduced.
refactor: refactored flamer fire handling code.
balance: floor acid is now twice as effective at extinguishing greenfire
/:cl:


JohnDoe50974:
Cool

DylonBR:
Pyro back to being just a retreat and siege delayer spec

usnpeepoo:
Finally

Wintermote:
> Pyro back to being just a retreat and siege delayer spec

Pyro's blue fire is still dangerous. It's red and green flame that's sucked for years now. Applying a slow was a step in the right direction, but it's still the case that damage output is so low xenos will literally walk through fire and set themselves on fire, just to kill/grab/drag people

Fire doesn't work to suppress movement if xenos are not appreciably punished for moving through fire tiles

Paperbell:
thank you nanu for unprofanity

Mirvra:
Hey Stan, ''(as is any xeno caste without a pounce/leap/dash ability tbh, but slow T3s especially so)'' Are xenos immune to fire while leaping/dashing? Because I could've sworn they instantly get applied damage from every single fire tile they cross while jumping over it... Or is it only certain abilities that avoids the fire proc?

I'm just asking because I know I've died instantly to green fire in the past by trying to dash through it as Vanguard, but I haven't tried it on other xenos so am unsure if this is a feature we have? Or a possible bug if we're supposed to be immune?

Geevies:
greenfire tweaks without fucked up rebase

PR #1111, opened 2022-09-22
queen eye fixes

stanalbatross:
## Changelog

:cl:
fix: the queen eye can no longer be put in nests
fix: when emitting pheromones, if you are using the queen eye, the radial menu will appear from the eye itself
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
@@ -335,7 +335,7 @@
 					return
 			else
 				var/static/list/phero_selections = list("Help" = image(icon = 'icons/mob/radial.dmi', icon_state = "radial_help"), "Frenzy" = image(icon = 'icons/mob/radial.dmi', icon_state = "phero_frenzy"), "Warding" = image(icon = 'icons/mob/radial.dmi', icon_state = "phero_warding"), "Recovery" = image(icon = 'icons/mob/radial.dmi', icon_state = "phero_recov"))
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
```diff
@@ -335,7 +335,7 @@
 					return
 			else
 				var/static/list/phero_selections = list("Help" = image(icon = 'icons/mob/radial.dmi', icon_state = "radial_help"), "Frenzy" = image(icon = 'icons/mob/radial.dmi', icon_state = "phero_frenzy"), "Warding" = image(icon = 'icons/mob/radial.dmi', icon_state = "phero_warding"), "Recovery" = image(icon = 'icons/mob/radial.dmi', icon_state = "phero_recov"))
-				pheromone = lowertext(show_radial_menu(src, src, phero_selections))
+				pheromone = lowertext(show_radial_menu(src, src.client.eye, phero_selections))
```
Zonespace27, line 338:
```suggestion
				pheromone = lowertext(show_radial_menu(src, src?.client.eye, phero_selections))
```
assuming clients is always a risk

stanalbatross: yeah forgot

Wintermote:
Does the queen eye still trigger motion detectors?

That always seemed pretty weird to me

Diegoflores31:
i think other way would be changing all the "ismob" and make a new variable called " isliving " or smth

PR #1112, opened 2022-09-22
Fixes ABR fire delay

silencer-pl:
## About The Pull Request
Turns out giving the ceremonial rifle that takes regular l42 ammo fire delay that's drastically lower than the base gun is not a great idea. Especially if the drill variant has a fire delay effectively 3 times lower of its normal counterpart.

## Why It's Good For The Game
Bugged OP guns bad.

## Changelog

:cl:Silencer_pl
fix: Fixed fire delay values for the ABR rifle that made them fire several times faster than intended.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -1273,7 +1273,7 @@
 	wield_delay = WIELD_DELAY_FAST
 	starting_attachment_types = list(/obj/item/attachable/stock/carbine/wood/drill)
 	map_specific_decoration = FALSE
-	
+
 /obj/item/weapon/gun/rifle/l42a/abr40/set_gun_config_values()
 	..()
 	fire_delay = FIRE_DELAY_TIER_7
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/guns/rifles.dm
```diff
@@ -1274,9 +1274,6 @@
 	starting_attachment_types = list(/obj/item/attachable/stock/carbine/wood/drill)
 	map_specific_decoration = FALSE
 
-/obj/item/weapon/gun/rifle/l42a/abr40/set_gun_config_values()
-	..()
-	fire_delay = FIRE_DELAY_TIER_7
```
morrowwolf, line 1279: This can stay, it's slower than base L42A.

morrowwolf (changes requested):

code/modules/projectiles/guns/rifles.dm
PhantomEpicness: Yeah the base ABR-40 is intended to be a surv weapon, though it is unused currently as I am seeking another sprite for it.

silencer-pl: Fixed

morrowwolf:
We appreciate the quick PR. Reverting was decided on.

PR #988, opened 2022-09-03
ABR-40 Ceremonial Rifle

PhantomEpicness:
## About The Pull Request
Adds the ABR-40 ceremonial rifle, a sleek polished rifle based on the L42A meant for ceremonies, drills, and execution. Credit to Blithering for the sprites

> An ABR-40 battle rifle, the L42A's predecessor, with an authentic polished wood furniture for ceremonial usage. Its frame is shined to perfection and you can even see your reflection in the bolt. Lacks the threaded barrel and underbarrel rail for attachments, with only the bayonet lug being the only place to attach anything. Too cumbersome to be fired with one hand, and the Sergeant Major would kill any marine caught trying.

Statistically, it sports a slightly faster fire rate, though it can only mount a bayonet, has worse wield delay, spawns with a half magazine, and movement slowdown due to a non-detachable stock, making it more of a side grade to the L42. It also cannot be fired while one-handed as it is shouldered while doing so.

The rifles can be found in the new room, execution chamber (replaced l42s there), and the SEA gear vendor. They have their own "ABR-40" magazines, which are essentially just half-size L42A mags, can be found in execution and purchasable in SEA vendor. L42 is compatible with ABR-40 mags and visa versa.

image
image
image
image
image

I also added M41A and ABR-40 blank magazines, which can be found in the SEA vendor and the new ceremonial room near the memorial. 
image

Two boxes of blank magazines are in the ceremonial room and blank magazines can be ordered in SEA gear vendor. Additionally, a box can also be found in maint

In addition, this PR also adds the ABR-40, which is in code only as of right now. ABR-40 is L42,but with worse fire delay and wield delay.

Also gave the SEA uniform vendor service jackets and dress shoes as a request from a mentor.

Both blank and ABR-40 magazines are buyable at cheap prices at ASRS

## Why It's Good For The Game
soul

## Changelog

:cl: Totalepicess, Blithering
add: Added M41A and L42A blank ammo
add: Added the ABR-40 Ceremonial Rifle, found in SEA vendor, execution and ceremonial room. Sprites by Blithering
add: Added a new ceremonial room to the Almayer, next to the memorial. 
add: Added service jacket to SEA vendor
/:cl:

PR #1113, opened 2022-09-22
Revert "ABR-40 Ceremonial Rifle"

morrowwolf:
Reverts cmss13-devs/cmss13#988

Multiple bugs need sorted as well as serious design concerns. Let's bring this back to the drawing board.


Segrain (approved):

Tristan63:
Having used the weapon ingame I think I can offer some advice on the best way to balance it. You could simply give it the same stats as the L42 but with the irremovable stock plus rails and lug restrictions, its aesthetic alone makes it unique enough to warrant its addition. Perhaps aswell limiting the amount onboard to just a handful in req, such as 5 or so kits that include blank magazines. Another thing that you could do is remove its ability to auto eject, and auto cycle. Some fun manual inputs for taking the old bess out.

PR #1135, opened 2022-09-27
Fix for explosive ammunition.

### LINKED ISSUE #1114: Explosive Mateba Rounds do not work when direct sprite clicking at range.
## Testmerges:
N/A

## Description

As says in the description - explosive mateba rounds do not seem to produce any result when direct clicking a sprite at range. PBs work as expected, as does clicking behind a sprite.

What happens: No projectile or explosion seems to be generated.

What is expected: Impact and a sizeable explosion.

## Steps to reproduce

1. Get a Mateba. Any Mateba will do.
2. Spawn explosive rounds and load
3. Direct click a sprite at range

Segrain:
## About The Pull Request

Hitting the turf should only happen _after_ we checked that we are not hitting anything on that turf. Otherwise we would never hit a literal immobile dummy, or anybody else for that matter.

## Why It's Good For The Game

Closes #1114.

## Changelog

:cl:
fix: Direct hits with explosive ammunition actually work as intended now.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -701,8 +701,7 @@
 
 /datum/ammo/bullet/revolver/mateba/highimpact/explosive/on_hit_turf(turf/T, obj/item/projectile/P)
 	..()
-	if(T.density)
-		cell_explosion(T, 120, 30, EXPLOSION_FALLOFF_SHAPE_LINEAR, P.dir, P.weapon_cause_data)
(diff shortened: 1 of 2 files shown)


GoldenDarkness55:
Wont this mess things up? Sprite clicks were meant to produce tile hit to stun for 5 sec with AP rockets, whereas clicking past would hit the mob with a shorter stun time. This MR would make direct clicks count as a mob hit instead, messing that up, right?

Shad0vvs (approved):

PR #1138, opened 2022-09-27
Update health scanner UI

StarkSilence:
## About The Pull Request

Makes the health scanner UI more readable

## Why It's Good For The Game

The current TGUI health scanner UI is difficult to read quickly and this helps alleviate that

## Changelog

- Damage total backgrounds no longer change if damage is not zero
- Both brute and burn damage is shown for each limb that is damaged in any way
- Unsalved and Unbandaged tags were removed, and the damage number is bold italicized instead
- Fracture and Splint tags colors changed
- Chemicals moved below limb status and made vertical
- Blood is no longer a health bar

image

:cl:
qol: Simplified TGUI Health Scanner interface for readability.
/:cl:


Misttyy:
🙏

If you could figure out why IB isn't updating or showing up after it's fixed that'd p much bring the health scanner back to normal imo

StarkSilence:
Stan already has a PR for that so I don't think I need to do anything

stanalbatross:
One thing - I do not think unbandaged and unsalved should be removed as these were included in the old ui as (t) and (b)

StarkSilence:
I know it's not as verbose but having the damage numbers bold italicized instead makes it much faster to kit/bandage/salve spam since you don't have to look over to check for those.

StarkSilence:
I like the alternating background colors idea, I played with dividers but that solution seems much cleaner.
Icons could work but using the tags is more consistent for now, I think that would have to be tested in another PR.
I thought about using multiple columns but having one column makes it easier to read since you only have to move your eyes vertically.

Mirvra:
> One thing - I do not think unbandaged and unsalved should be removed as these were included in the old html ui as (t) and (b)

I agree. One thing that the new Tgui has helped with is to make it easier to tell which areas of a patient you can treat with Surgical line or Graft to bring their damage down for revive.

Managed to revive a marine with 600+ damage last night by finding every single burn/bruise and fixing it.

StarkSilence:
Two other solutions to Unsalved and Unbandaged:

Brackets:
brackets

Letters:
letter

Mirvra:
Personally Prefer the Brackets, though I could see new players having a harder time understanding it... But I'll still vote for Brackets.

stanalbatross:
Decrease window width a bit too

Wintermote:
Can we please put medication advice above blood type, body temperature, and pulse? The information about those 3 things is nearly irrelevant, especially for combat medics. Meanwhile the information about what drugs to give someone is absolutely /critical/.

It would be nice not to have to scroll down the UI window to see what medication to give at a glance.

Wintermote:
> Personally Prefer the Brackets, though I could see new players having a harder time understanding it... But I'll still vote for Brackets.

Brackets are way better imo. 
Also, I don't see how [Bleeding] would be harder to understand than 'B' which is completely non descriptive and meaningless without additional meta context (e.g. you already know B stands for bleeding)

Misttyy:
> Stan already has a PR for that so I don't think I need to do anything

My bad, I didn't know he had a PR up for that! Great work getting to this stuff quickly.

StarkSilence:
With alternating background colors and brackets:
image

gohdatrice:
Seems like it breaks when scanning a xeno, you get the blue screen thing. Keep in mind I've only seen this with the scan health feature as ghost but I imagine it happens if you scan a xeno with a health scanner in game

StarkSilence:
That seems to happen on the main branch when I test it locally too so I'm not sure that's related to this change

gohdatrice:
Oh yeah I got confused, this was a bug I found on the main server so it wouldn't be this since I don't think this is merged. I'm not sure where the PR is for the original health scan change so hopefully whoever's in charge of that ends up reading this

Mirvra:
> > Personally Prefer the Brackets, though I could see new players having a harder time understanding it... But I'll still vote for Brackets.
> 
> Brackets are way better imo. Also, I don't see how [Bleeding] would be harder to understand than 'B' which is completely non descriptive and meaningless without additional meta context (e.g. you already know B stands for bleeding)
> 
> Abbreviations might be necessary, but it's better to be descriptive so newer players can hamfist their way through figuring it out.

The B doesn't refer to Bleeding. It refers to Burns, which means you can treat the patient with either Surgical Graft or burn kits. T stands for Trauma I think, which means you can surgical line or Trauma kit them. They're completely separate from Bleeding.

It's just a way to see which areas of the patients body you can still Kit or suture to bring damage down for revives and such.

Katskan:
Minor thing but the old UI had burns at a much more yellow color, as far as I'm aware the same color code was used for the new stuff but there's a clear visual change to it. The orange is also a little less visually distinct and negatively impacts rapid recognition as well as it no longer does the medicine color matching what it treats, as the newer more orange color matches tricord more than it does kelotane

tgui/packages/tgui/interfaces/HealthScan.js
```diff
@@ -198,18 +178,10 @@ export const HealthScan = (props, context) => {
...
-                    good: [0.9, Infinity],
-                    average: [0.7, 0.9],
-                    bad: [-Infinity, 0.7],
-                  }}>
-                  <Box>
-                    {Math.round(blood_amount/5.6)}%, {blood_amount}cl
-                  </Box>
-                </ProgressBar>
+                label={"Blood Type " + blood_type}>
+                <Box color={blood_amount/560 > 0.9 ? "green" : blood_amount/560 > .7 ? "orange" : "red"}>
```
stanalbatross, line 182:
```suggestion
                const bloodpct = blood_amount/560
                <Box color={bloodpct > 0.9 ? "green" : bloodpct > .7 ? "orange" : "red"}>
```

stanalbatross (approved):
happy with this, will be merged, next I will try to address vertical height issues (it's big) and the screen position not saving

PR #1145, opened 2022-09-29
Re-use Health Scanner UI

Drulikar:
## About The Pull Request
Scanners now store the health_status datum so they can alter the target_mob of the scan and re-use the interface rather than create a new interface. The TGUI is keyed by the usr and src object it was created for, so having a single health_status datum is what allows it to be re-used.

## Why It's Good For The Game
Helps with UI clutter and may also be more performant.

## Changelog

:cl: Drathek
qol: Health scanners now re-use their interface when scanning again.
qol: Opening the archived scan for a health scanner now has a window named "Stored Health Scan" and closes the "Health Scan" window (and vice versa).
fix: Fixed health scanners sometimes incorrectly considering a mob to be permadead when they are alive.
code: Mobs no longer have a health display datum and instead each scanner does.
/:cl:


stanalbatross (changes requested):
image
seemingly advice does not update between re-scans?
there seems to be some weirdness with this and advice in general, in some instances it doesn't give any entirely

stanalbatross (approved):
good

PR #1148, opened 2022-09-30
Xeno Ability check refractor

PhantomEpicness:
## About The Pull Request
This PR refractors Xeno ability checking so that checking obvious conditions like if xenos are dead, marked for deletion, busy or otherwise occurs at the parent rather than the abilities themselves, saving the headache of future coders having to include all of them so dumb bullshit doesn't happen.

It also fixes a bug that allowed xeno_action/activable abilities to skip their usual checks that other types like /onclick would fail.

In essence, abilities can no longer even be selected, let alone even activated, if you are in a state that you shouldn't be allowed to activate abilities.

This has potential issues as well as balance implications so please add a TM tag.

## Why It's Good For The Game
massive coding QoL
Coders no longer have to deal with the clusterfuck of copy + pasting code every time they want to make a new ability or deal with xenos using them while dead. 

## Changelog

:cl: totalepicness
fix: Fixed being able to select an ability despite being dead,crit,incapacitated, etc
refactor: Refractored Xeno ability selection and usage checks
/:cl:


PhantomEpicness:
**This is not in a merge-ready state.** 

While it is functional, I still need to refractor all Xeno abilities to now call their parents and remove the existing, now redundant checks from them among other things.

I made the PR as I need further dev input on what to check for on all abilities and what exceptions need to be made if necessary.

code/modules/mob/living/carbon/xenomorph/abilities/xeno_action.dm
```diff
@@ -42,8 +42,12 @@
 // Any strain or caste-specific state should be stored on behavior_delegate objects
 // which use_ability invocations can modify using typechecks and typecasts where appropriate.
 /datum/action/xeno_action/proc/use_ability(atom/target)
+	var/mob/living/carbon/Xenomorph/X = owner
 	if(!owner)
 		return FALSE
+	if (!X.check_state() || X.action_busy || QDELETED(X) || !action_cooldown_check() || !istype(X))
+		to_chat(X,SPAN_WARNING("You can't do that now USEABILITY."))
```
stanalbatross, line 49: better tochat please

code/modules/mob/living/carbon/xenomorph/abilities/xeno_action.dm
```diff
@@ -64,7 +68,8 @@
 	var/mob/living/carbon/Xenomorph/X = owner
 	if(!istype(X))
 		return FALSE
-	if(X && !X.is_mob_incapacitated() && !X.dazed && !X.lying && !X.buckled && X.plasma_stored >= plasma_cost)
+	if(X && !X.is_mob_incapacitated() && !X.dazed && !X.lying && X.check_state() && !X.buckled && X.plasma_stored >= plasma_cost)
+		to_chat(X,SPAN_WARNING("You can't do that now CAN USE."))
```
stanalbatross, line 72: better tochat

code/modules/mob/living/carbon/xenomorph/abilities/xeno_action.dm
```diff
@@ -127,7 +132,8 @@
 /datum/action/xeno_action/activable
 
 /datum/action/xeno_action/activable/can_use_action()
-	return TRUE
+	. = ..()
+	return
```
stanalbatross, line 136: why does this even need to be a thing if behaviour is all handled in parent

stanalbatross (changes requested):

PR #1150, opened 2022-10-01
surgical line fix

stanalbatross:
## Changelog

:cl:
fix: surgical line and synthgraft no longer have nobludgeon flags, so they will work as normal
/:cl:


Shad0vvs (approved):
shitty fix

PR #1155, opened 2022-10-02
Logistics Stuff

Geevies:
## About The Pull Request

Implements packable storage containers that can hold up to 4 crates at a time. These containers occupy a 2x2 tile area and cannot be moved by conventional means. However, a powerloader can lift and transport the container. To pack the container, just drag the crates onto it, then click it. To unpack, just click it. 

Added flatbed trucks, one spawns in the Req bay. Powerloaders can lift the storage container onto the flatbed when hit from the back, which loads the container onto the flatbed. The flatbed can then transport the crates anywhere, but a powerloader is needed to unload the flatbed.

Added powerloader parts. When a powerloader is broken, the wreck is no longer movable, and has to be disassembled with a wrench. This gives you some spare parts which can be combined with other parts to make a new powerloader. A complete powerloader can also be disassembled into parts, but this needs a minimum engi level of trained to do so. To start assembling the loader, just click on the floor with a single part, then load more parts into the assembly. Additionally, the powerloader's health has been raised from 200 to 300.

Cargo Technicians and Requisition Officers now get Vehicles 1.

Lastly, removed all of Req's drop pads.

## Why It's Good For The Game

It makes logistics more interesting than just dropping a package to the front, as you now have to manually package, load, transport, unload, unpackage, repackage, load, bring it back, repeat.

## Changelog

:cl:
add: Implemented storage containers that can hold 4 crates at a time. Click to package/unpackage, can be lifted with a powerloader.
add: Added flatbed trucks, which can hold the storage containers in the back.
add: Added powerloader parts. Various stages of the powerloader can be disassembled with a wrench for parts, and 5 parts can be used to make a new loader. Click the ground with a part to begin assembly.
balance: Buffed the powerloader from 200 health to 300.
balance: CTs and ROs now get vehicles 1.
del: Removed all of Req's drop pads.
/:cl:


AGasCanister:
what 8 hours of foxhole does to a mf

Skatuk:
Make the van go the same speed as the colony van's. but yeah you've played too much foxhole.

GDSPathe:
Ran the new req for a one-hour Solaris game. This is coming from someone who prefers to run a solo req, and who mostly plays low-pop, so your YMMV in regards to feedback.

Honestly feels like adding a bit of busy work for req while eliminating potential coordination with the ground-side marines. Moving crates from req was not that tedious during the round itself, I could see it getting old fast given the relatively slow speed of the truck. 

Prior to the PR, you could supply marines without needing coords by shuttling crates to and from the Hangar, it was tedious and devoid of any need to coordinate with marines. The PR effectively turns that into the core req cycle. 

Whereas with drop pads, req had the potential to coordinate closely with marines to deliver ammo and material to different locations, the new system dumps that for a request -> load -> drive -> unload cycle that seems less satisfying and offers less room for marine interaction then the old cycle of: request -> load -> communicate -> drop.

SexmasterWasTaken:
Make the van faster, it gets outrun by marines in armour.

Also, make it easier to load and unload the van, this is way too tedious to just supply the front for it to be kept like this. Maybe not having the weird 2x2 loading block, and instead being able to manually put crates on it, but only like 2 or 3. Having to manually unload and load them 1 by 1.  Because needing a power-loader to unload this just makes it really annoying, especially for providing supplies to the front.

Personally I think it's a more interesting than just sitting in req and waiting for more money and ASRS to buy stuff, but it is definitely a hinderance for the frontline.

Bram-KEP:
this shit is epic af

DylonBR:
Aw hell naw they done added foxhole to CM

JohnDoe50974:
RIP frontline drops. Though the significant ease in shuttling crates to and fro req is very welcome.

DorfGuy1:
This is a terrible idea, and is a huge nerf to the overall balance, love the fox hole ideas though, just make it better, because this is not the way. We've added static comms, wich is another backline thingns marines need to baby sit. This is not foxhole. And as cool as logistics is, this is a terrible idea, because it subtracts competent, or at least intelligent marines from the fob, front and comms. It also diminishes the roll of overwatch. If you really wanted to do somethign more interesting than drops you'd do a little cargo dropship or something, but as it stands this is a bad idea, its not like anyone used the drops to move up to the front, and the ROs will streight up tell you to go fuck yourself if you ask for things to be delivered to anywhere other than fob.

tyeagg:
from gameplay i think you should atleast readd launch pad to launch exclusively to fob and truck will just be transport to frontlines because waiting to go up and down the dropship in a truck is not fun

harryob:
CleanShot 2022-10-03 at 15 35 47
the flatbed (and the driver and any occupants) get totally destroyed if the ASRS pad moves up while it's on this corner at all. it's kinda finny though

AmoryBlaine:
The majority of supplies go from Req, to the Shuttle, to the FOB at the LZ. Very little is actually moved to the frontline since most rounds don't last long enough to build up a frontline that would need materials delivered to it. That rapidity has nothing to do with supply drops to the front since those are usually just extra ammo crates, stuff that usually gets dragged to the front from the FOB anyways, so this won't slow down rounds and make the truck viable, it just means that supplies will never actually make it to the front when necessary.

ghostsheet:
Aww come on guys!

If this was about to happen why did this get denied?

https://gitlab.com/cmdevs/colonial-warfare/-/merge_requests/2210

stanalbatross:
> Aww come on guys!
> 
> If this was about to happen why did this get denied?
> 
> https://gitlab.com/cmdevs/colonial-warfare/-/merge_requests/2210

"about to happen" (ten months later)

do you understand the concept of time?

ghostsheet:
Woah did it really take 10 months to get this PR approved for testing? That's quite the development time

Well it's an interesting PR so far, however it has quite a few factors that makes it less than engaging to use.
Adding another multi-tile vehicles on the field is already a little sketchy,
I think it would be improved if this was the power loader aspect was removed, I think it really adds unncecssary bulk, and prevents delivery cargo to the frontline, which I'm assuming is a big part of this. (The risk of logisitic? backline xeno intercepting supplies)
and plus, right now it seems to be easier and faster to push the crate onto the dropship than to use the vehicle + power loader combo sadly

Sorta wanna see how an easier cargo truck would do

Wintermote:
The only thing I see this actually being useful for is hauling intel crates to the intel room, but only if you don't need a powerloader and if the truck can somehow go up the stairs and its faster than hauling/shoving two crates at once.

I don't think this PR was well thought out. Transporting stuff to the front needs to be relatively painless and quick, powerloaders on the frontline are simply not going to cut it, they will get destroyed before you even do anything with them because they're huge targets, easy to destroy by xenos or friendly fire, and very slow.

Here's how I'd do this instead:
1. Keep the supply drops as they are
2. Turn the truck into a powered ruggedized pallet jack 
3. Stack crates by hand on a wood pallet, maybe wrap them in plastic too if you want to be autistic and accurate; but mind its going to be a pain to deal unless marines have easy access to alternative methods to strapping things down (cable coils?)
4. Powered ruggedized pallet jack moves wrapped pallets between dropship and req, or fob and frontline at a speed at least as fast as a marine in normal armor

5. Consider how well this works and whether further changes can or should be done to remove crate drops and only using pallet jacks/trucks to transport things

Making logistics harder and more time consuming for marines isn't going to solve anything. It's already bad enough that people simply haul their own ammo crates/backpacks by hand, making it worse will just encourage more of that and remove requisition as a job entirely.

Consider also that its very easy and fast for a marine to run back to fob to resupply; like it takes maybe 2-3 minutes. If your logistics process is slower than a marine running back to fob to grab more ammo and then back to the front, you need to reconsider your logistics process. This btw, is also why nobody uses medevac instead of just walking/hobbling back to fob - the medevac + shipside surgery + going back down is 10x slower than just walking back.

Overall, the idea is promising, the actual implementation is lacking

PR #1170, opened 2022-10-04
New Xeno T3: Vicar

usnpeepoo:
## About The Pull Request

Adds a new T3 caste to the Xeno roster. It's called Vicar, evolves from Carrier, based off the Shaman strain. Uses the King sprite from TGMC, credits to Manzehino who sprite it originally. Also credits to ZSpitfire for the ability icons.
It's a frontliner, weeding T3 which has the special ability of buffing xenos or debuffing marines.

It has a reserve system, similar to Warden but instead of hitpoints it is called "acid reserves". You can only gain acid reserves off slashing, and your 2 special abilities use this reserve system. The max amount of acid reserves is 150, and you gain 10 for each slash.

**First ability, Roar:**
Has two modes, a buffing and debuffing roar.
The buffing roar gives every xeno within view/6 tiles 15 extra armor, and 10 extra slash damage. Its duration is 15 seconds, it has the shaman shield overlay on the xeno/Vicar so it can be seen by everyone. It cannot be stacked.
The debuffing roar has a range of 3 tiles, and only affects humans. The debuffing when used, causes a 0.5s knockdown, 4 second slow, and 5 second welder vision.
The roar uses up the entire reserve, and has a 50 second cooldown.

**Second ability, Rooting Slash:**
2 tile range ability, that roots and deals 15 AP damage to chest. The root is 2 seconds long.

**Third Ability, Dash:**
3 tile dash, deals 15 normal slash damage, and gives 10 reserve (if the slash hits).

**4th, and final ability Acid Throw:**
Use up 70 points of your acid reserve to create an 3x3 tile that causes marines caught in it to be blinded/welder vision for 2 seconds. This is similar to boilers acid mine, but instead of dealing any damage/cade damage it simply blinds marines.

It can weed.

**Health is 650, armor is 20, slash damage is 30.**

## Why It's Good For The Game

Xeno content, and a well rounded support T3 that doesn't just heal, or give shields. It's debuffing abilities are meant to support other xenos, or allow the Vicar to fight with an advantage. I've also wanted to make a t3 for the carrier/drone line that isn't just the queen.

Design doc to be made, will explain the abilities/caste in further detail.

## Changelog

:cl:Usnpeepoo ZSpitfire 
add: A new T3 Caste to Xenos. It's called Vicar, evolves from carrier and is based off the Shaman strain. It can weed.
add: A acid reserve system for Vicar, it's maximum is 150 and it gains 10 points for each slash. This system is used to enable the Roar, and Acid Throw
add: A new ability to Vicar, Roar. It' has two modes, a buffing and debuffing roar, both of which can be toggled. The buffing roar gives every xeno in view/range of the Vicar 15 extra armor, and 10 more slash damage for 15 seconds. There is a green overlay when this occurs allowing anyone to see the ability active. It cannot be stacked. Roar on both modes use up 150 acid reserves and has a 50 second cooldown.
add: The second mode of Roar is a debuffing one, marines within 3 tiles of the Vicar when this ability is used get knocked down for 0.5s, slowed for 4 seconds, and welder vision for 5 seconds.
add: 2nd ability of Vicar, rooting slash. 2 tile range, roots for 2 seconds, 15 AP damage.
add: 3rd ability of Vicar, Dash. It is a 3 tile dash, that when it hits a marine gives the Vicar 10 extra reserves and slashes the marine for 15 normal damage.
add: 4th ability of Vicar, Acid Throw. It is similar to trapper boilers acid mines, but uses up 70 acid reserves and does not deal damage. The effects when a marine gets hit by it are 2 secs of blindness/2 seconds for welder vision. 21 second cooldown.
/:cl:


usnpeepoo:
Permission to make this draft PR was given by Stan.

stanalbatross:
the name and sprites are a bit mid but we can come up with something better later

code/__DEFINES/xeno.dm
```diff
@@ -653,3 +657,6 @@
 #define NOT_WEEDABLE 0
 #define SEMI_WEEDABLE 1
 #define FULLY_WEEDABLE 2
+
+// Royal
+#define ROYAL_SCREECH_BUFFED  	1
```
carlarctg, line 662: probably would be wise to give this a different name since it might get confused with ROYAL_SCREECH_BUFF
may as well move it up next to those royal_screech defines so they're all together
```suggestion
#define ROYAL_SCREECH_ENHANCED  	1
```

code/modules/mob/living/carbon/xenomorph/abilities/royal/royal_powers.dm
```diff
@@ -0,0 +1,219 @@
...
+		Hu.update_xeno_hostile_hud()
+
+		addtimer(CALLBACK(GLOBAL_PROC, .proc/unroot_human, H), get_xeno_stun_duration(H, root_duration))
+		to_chat(H, SPAN_XENOHIGHDANGER("[X] has pinned you to the ground! You cannot move!"))
+
+	apply_cooldown()
+	..()
+	return
+
+/datum/action/xeno_action/activable/cleave/proc/remove_bufff()
```
carlarctg, line 160: extra f on the proc name

usnpeepoo: it conflicts with another var, i'll change it to remove_enhancement or some shit

code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
```diff
@@ -269,6 +269,8 @@
 	var/icon_xeno
 	var/icon_xenonid
 
+	var/screech_status_flags = 0
```
carlarctg, line 272: as i understand it this flag is used as a single variable that tells you if the xeno is buffed by a royal rn or not. this would definitely be better off as a variable that holds bitflags that indicate various buffs or not the xeno might have for the future

such as 

`var/xeno_effect_flags = ROYAL_ROAR_ENHANCED`
and checked with
`if(X.xeno_effect_flags & ROYAL_ROAR_ENHANCED)`

this would also let it be used for other possible buffs either by refactoring preexisting buffs or the shitty pheromone variables to also be bitflags, also helpful for future castes or strains

`#define ROYAL_ROAR_ENHANCED (1<<1)`

code/__DEFINES/xeno.dm
carlarctg: note this suggestion won't replace all instances of the former name

ilikesaying:
frontline weeding t3 caste that uses TGMC King sprites but also has TGMC Gorger abilities

GoldenDarkness55:
2s on the root seems a bit too long, a death sentence on top of turning the marines into shields on an ability they can't reliably dodge.
The CD also looks oppressor tier. I get the balance direction but giving "you died" abilities long CDs doesn't make them any better to play against. Also introduces situations where support caste have healed you but there's no point going in as ur still waiting on cooldowns, making them less impactful.

Scree CD seems similarly excessive by 10-20s, I guess it's a way to force spending blood res on the other ability instead? I think it should be in line with other caste (vguard, crusher, warden etc) where if you're doing well enough to keep stacking at a high speed you get rewarded w/o hard limiters. Altho you'd have to rethink this weird design triangle so players don't just keep blood for scree - maybe it's fine as-is.
Not a fan of armor stack, kind of a shield reskin and possibly unholy on some caste while having questionable effects on the rest (armor vs AP ammo pen value). Makes non-AP casuals even more irrelevant, giving the few robustos even more monopoly on round results - arguably worse for the game than the shield system.
Could consider movement/slash speed bonuses or extra on slash effects like debuffs/buffs/CDR/lifesteal to go with the dmg buff, but up to where devs wanna take the game I guess.

Cool design in general though, seems like a fun ability set.

Agameofscones:
Add your name to the changelog, find the original artist to give proper credit.

usnpeepoo:
> frontline weeding t3 caste that uses TGMC King sprites but also has TGMC Gorger abilities

Aside from the sprite, this caste is nothing like gorger. The only thing similar is the blood usage which will be changed.

usnpeepoo:
> New mini-queen screech xeno yaaaaay
> 
> 
> 
> This feels like a xeno that's entire job is already covered by other castes, right down to the mini-screech.

It's not the queen screech

Carnage068:
> Not a fan of armor stack, kind of a shield reskin and possibly unholy on some caste while having questionable effects on the rest (armor vs AP ammo pen value). Makes non-AP casuals even more irrelevant, giving the few robustos even more monopoly on round results - arguably worse for the game than the shield system.

i completely agree on this lmao. unless it's a backline supporter (which it's clearly not), giving xenos 15 armor would be pretty bad for the game. a small AOE heal (say like, 75-125 health) would be a lot neater. if it's supposed to be a generalist it would still be making its single target healing way weaker than something like Warden, but eh, wtf do i know.

anclandor:
So... this new xeno caste, the royal, is a frontline caste.  Would be okay to call it a battle royal?  I can't believe they are adding battle royal to cm.

Agameofscones (approved):
Sprites are fine

stanalbatross (changes requested):
name must follow "verb-er" format or otherwise have a good reason for being called what it is (i.e. praetorians are guards to the queen, burrowers burrow, etc) (reminder : names are made based on what marines would call them after contact with the caste in question)

carlarctg:
> name must follow "verb-er" format or otherwise have a good reason for being called what it is (i.e. praetorians are guards to the queen, burrowers burrow, etc) (reminder : names are made based on what marines would call them after contact with the caste in question)

> A vicar (/ˈvɪkər/; Latin: vicarius) is a representative, deputy or substitute; anyone acting "in the person of" or agent for a superior (compare "vicarious" in the sense of "at second hand"). Linguistically, vicar is cognate with the English prefix "vice", similarly meaning "deputy". The title appears in a number of Christian ecclesiastical contexts, but also as an administrative title, or title modifier, in the Roman Empire. In addition, in the Holy Roman Empire a local representative of the emperor, perhaps an archduke, might be styled "vicar".

> Vicars exercise authority as the agents of the bishop of the diocese. Most vicars, however, have ordinary power, which means that their agency is not by virtue of a delegation but is established by law. Vicars general, episcopal vicars, and judicial vicars exercise vicarious ordinary power; they each exercise a portion of the power of the diocesan bishop (judicial for the judicial vicar, executive for the others) by virtue of their office and not by virtue of a mandate.

TL;DR: A 'vicar', is a representative, often of faith, who definitionally is a minion, underling, agent of a higher power, like a bishop, emperor, or archduke.
Their authority, if religious, comes not from said religion and its mandates but simply by the local law.

These attributes and the definition perfectly translate to the role of this caste: As a drone caste, the closest to the Queen or Empress, they offensively assist her, fulfilling a similar role with the screeching, weeds, and ranged attacks while being significantly lesser, not just in might, but in definitional power: A Vicar is no more powerful than any other T3, unlike the Queen, who has a psychic mandate to lead and control the entire hive, standing in contrast to, just as an example, TGMC's Shrike, their version of a mandated hive princess.

usnpeepoo:
Damn this shit broke again

PR #1187, opened 2022-10-07
animations fixes + misc.

### LINKED ISSUE #1183: Cloak Item Equip Visible
## Description
Cloaked entities still have the same item equip visible, which enables other entities to see them plain as day just because they move something out of a bag or off an inventory slot. Additionally mobs which can see through walls using meson scanners can see items being picked up through walls. 

Cloaked entities shouldn't be made instantly visible with a weird floating item, including if they're through a wall or not. 

## Steps to reproduce
Activate a scout/hunter cloak, then pick up an item. You can see the little item pickup animation plain as day, even though the person is invisible. 

stanalbatross:
## About The Pull Request

some fixing

closes #1183 

## Changelog

:cl:
fix: pickup animations now inherit the alpha level of the person picking them up
fix: pickup animations can no longer be seen through walls
fix: fixed a rogue submodule in the git
code: changed all instances of var/mob/m in viewers to as anything in viewers, because viewers only gets mobs.
/:cl:

### DIFF
--- a/code/game/machinery/robot_fabricator.dm
+++ b/code/game/machinery/robot_fabricator.dm
@@ -134,6 +134,6 @@ Please wait until completion...</TT><BR>
 			else return //Someone's doing href fuckery if this gets here.
 		return
 
-	for (var/mob/M in viewers(1, src))
+	for (var/mob/M as anything in viewers(1, src))
 		if (M.client && M.interactee == src)
 			src.attack_hand(M)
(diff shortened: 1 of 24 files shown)


morrowwolf:
R O G U E SUBMODULE

Shad0vvs:
submodule...

Shad0vvs (approved):

PR #1197, opened 2022-10-07
CO Sniper

Geevies:
## About The Pull Request

Adds a CO sniper to the CO vendor, it's blue, doesn't have IFF. That's about it.
Also adds sniper ammo to the CO vendor, prices should be balanced. Also a bipod, nice.
The rifle grants nightvision only when you aim down the scope.

## Why It's Good For The Game

It's a sniper rifle that lets the CO hang back and command, but it also doesn't have IFF, so it isn't busted OP.

## Changelog

:cl:
add: Added a CO sniper and its ammo plus a bipod to the CO gear vendor. The rifle grants nightvision when aiming down the scope.
/:cl:


code/game/machinery/vending/vendor_types/crew/commanding_officer.dm
```diff
@@ -26,6 +27,8 @@ GLOBAL_LIST_INIT(cm_vending_gear_commanding_officer, list(
 		list("SPECIAL AMMUNITION", 0, null, null, null),
 		list("M41A Incendiary Magazine", 65, /obj/item/ammo_magazine/rifle/incendiary, null, VENDOR_ITEM_REGULAR),
 		list("M41A Rubber Shot Magazine", 10, /obj/item/ammo_magazine/rifle/rubber, null, VENDOR_ITEM_REGULAR),
+		list("M42A Incendiary Magazine", 45, /obj/item/ammo_magazine/sniper/incendiary, null, VENDOR_ITEM_REGULAR),
+		list("M42A Flak Magazine", 30, /obj/item/ammo_magazine/sniper/flak, null, VENDOR_ITEM_REGULAR),
```
stanalbatross, line 31: price diff seems somewhat arbitrary, considering these three ammo types are very similar in power

stanalbatross (changes requested):

stanalbatross (changes requested):
needs sprites for on your back, in suit slot, etc.

Geevies:
> needs sprites for on your back, in suit slot, etc.

the normal m42a has an empty state as its suit slot, so i just copied that. added the back one tho

code/game/machinery/vending/vendor_types/crew/commanding_officer.dm
stanalbatross: by this I meant make them all 30 points or whatever, marksman is as strong as flak due to aimed shot

code/modules/projectiles/guns/specialist.dm
```diff
@@ -220,6 +226,26 @@
...
+	if(user.job == GET_DEFAULT_ROLE(JOB_CO))
+		return TRUE
+	return ..()
+
+/obj/item/weapon/gun/rifle/sniper/M42A/co/set_bullet_traits()
+	return
+
+/obj/item/weapon/gun/rifle/sniper/M42A/co/handle_zoom_vision(var/mob/living/carbon/human/H)
+	H.see_in_dark = 12
+	H.lighting_alpha = LIGHTING_PLANE_ALPHA_MOSTLY_INVISIBLE
```
stanalbatross, line 247: why is this on the gun not the scope 💀 

also I don't get this feature, really. It makes it better to look down your rifle scope to get binoc vision than down actual binocs, which is weird. 

The lack of iff also makes it somewhat useless other than when - I guess - pushing a wide open area?

because snipers suck in fob defence (can't finish off kills)

Superreallycoolguy:
I know this has been a meme between you and COs for a bit, but I think it really needs to stay just a joke. This item honestly seems worthless to the CO weapon roster. It doesn't fit a niche or a way to improve the core values of CO. It can not even backline properly, it seems like a genuine hindrance to CO, making him less of a commander and a psuedo shitty sniper spec. I am not sure what the intentions of this weapon was but I just fully believe it doesn't fit something that CO needs. 

The SG is a great support/CAS coordinator with its backline and support potential, the Night Vision is a boon too! While the M46C still fulfills that easily flexible rifle to use in pushes or as a vanguard weapon. Both weapons have defined roles and do their own unique thing to change up the game for CO while still being an active groundside commander. I just don't believe this rifle does that or fits its own spot

Im sure if you talked to COs or the Council about this weapon, something could have been formulated to make it unique and exciting to play with, as well as enhanced the ground side CO experience, maybe give it its own niche. But as it stands, I do not think this weapon should be added to CO in the slightest. The infamous CO sniper should stay as a meme until something more can come from it than an IFF-less (why) sniper rifle for CO to use.

c4xmaniac:
No IFF sniper is funny and was a good joke in the CO chat but it's not usable. You can't use it to support a push and you can't use it period because marines are always in front of you. The reason the m46c has an IFF toggle is because you can't (and shouldn't) always be at the head of a push, so the iff mode exists to allow the CO to trade damage for actually being able to fire. The SG allows you to always be firing with its constant IFF, but doesn't do the burst damage that the m46c can. This sniper just cannot be fired, and when it can, it doesn't meet the expectations one would have of a very difficult and rare to fire weapon.

I disagree with Karl that the CO sniper as a concept being a bad idea, I do think it has a purpose because it lets you get a far greater sense of perspective of what's going on than just what your binos allow, which is in of itself an interesting trade compared to the SG which allows you to always have night vision and +1 tile of vision. 

Though I agree with Karl that if you had wanted to make a CO sniper, you should have spoken with the council. There could have genuinely been cool stuff added like maybe a laser range finder built into the scope or something, but as it stands this weapon should not be added.

4hands44:
the m456c Can already do sniper as good or better than an l42 with a scope. This isnt needed. plus why would a CO have a dedicated sniper rifle? It doesn't make much sense.

Drevenshek:
this is completely pointless meme PR. Not having IFF  makes it literally useless for a backline CO, it does nothing the M46C or Smartgun can't do. Why did you close 4hands CO PR which was actually a good and useful addition to CM then open this lmao

Fortelian:
A sniper rifle without IFF... any CO who actually wants to use this is going to spend more time trying to shoot xenos than command their marines. A pair of binoculars gets the job done better. Like stated above, this will only negatively impact COs and distract them from their actual job. If you really want to get this added then talk to the CO council before trying to make it a thing.

JohnDoe50974:
Ignore the critics and just merge this. Do it for the vine!

GoldenDarkness55:
Less guns more commanding please. About to transcend pfc++ and become a spec+ role

LoJoFan:
this is awesomesauce

morrowwolf:
This needs more discussion IMO. I personally feel a sniper goes against what we're looking for for the CO roll.

MusselMain:
Cool idea, but as someone who's played Sniper spec for years, having no IFF is incredibly crippling to an already incredibly selective weapon (getting kills as a Sniper spec requires a fair amount of factors), this could be fixed if it was a burst sniper that shot two or three shots at a time, making the aimed mode much more powerful but also more risky, as you lose the IFF.

stanalbatross:
PR author stated in DMs that this was 100% a meme PR with no intention of actually being merged so it will be closed

PR #1205, opened 2022-10-08
XM88 buff

stanalbatross:
## About The Pull Request

The XM has been underperforming when compared to similar AP dealing weapons like the mod 88 and L42 with AP ammo so this PR seeks to make it a bit better. It'll be better at chasing, making it on a similar level to the L42, and it will have a bit of inherent AP so it'll hit a bit harder when starting its streak.

## Why It's Good For The Game

Makes a somewhat weak kit gun more viable and competitive

## Changelog

:cl:
balance: reduced the XM88 stock slowdown a bit
balance: gave XM88 10 ap ammo by default
balance: shifted the XM88 stack ap tiers up by 10 each to compensate for this
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1443,10 +1443,6 @@
 	accurate_range = 14
 	handful_state = "boomslang_bullet"
 
-/// unused
-/datum/ammo/bullet/lever_action/xm88/pen10
-	penetration = ARMOR_PENETRATION_TIER_2
-
 /datum/ammo/bullet/lever_action/xm88/pen20
 	penetration = ARMOR_PENETRATION_TIER_4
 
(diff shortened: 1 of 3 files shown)


code/modules/projectiles/ammo_datums.dm
```diff
@@ -1437,12 +1437,13 @@
...
-	penetration = 0
+	penetration = ARMOR_PENETRATION_TIER_2
 	accuracy = HIT_ACCURACY_TIER_1
 	shell_speed = AMMO_SPEED_TIER_6
 	accurate_range = 14
 	handful_state = "boomslang_bullet"
 
+/// unused
 /datum/ammo/bullet/lever_action/xm88/pen10
 	penetration = ARMOR_PENETRATION_TIER_2
```
Geevies, line 1448:
```suggestion
```

Geevies (changes requested):

Geevies:
as for the balance, yeah i don't see an issue with it, i don't see the XM88 frag out on the field when i've played, so this change should be interesting to see. as usual, if its an issue, we can roll it back

PR #1210, opened 2022-10-08
Fix labeling of living mobs

DoomLaser9:
## About The Pull Request

Living mobs are not supposed to be able to be labeled. However, some mobs can be picked up and then labeled. This fixes that by adding a check for mob holders.

## Why It's Good For The Game

Bugfix.

## Changelog

:cl: Laser9
fix: Living mobs that have been picked up can no longer be labeled.
/:cl:


harryob (approved):
major jones fix

PR #1211, opened 2022-10-08
Cat meat fix

DoomLaser9:
## About The Pull Request

This adds real_name values for cats. Jones also has a custom real_name now. I did not give the Runtime cat mob a custom real_name as it is used for several custom-named cats in maps, which do not change real_name. I used `corgi.dm` as a reference for this.

Previously the gibber would create `-meat` items from cats. Now it properly creates `cat-meat` (and `Jones-meat`).

## Why It's Good For The Game

Bugfix. (also Jones-burger)

## Changelog

:cl: Laser9
fix: Processing cats with the gibber now properly names the resulting meat.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/cat.dm
+++ b/code/modules/mob/living/simple_animal/friendly/cat.dm
@@ -1,6 +1,7 @@
 //Cat
 /mob/living/simple_animal/cat
 	name = "cat"
+	real_name = "cat"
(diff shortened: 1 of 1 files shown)


harryob (approved):
code will work, psychopathic results however

stanalbatross (approved):
might as well give one to /kitten as well.... for kitten-meat

PR #1217, opened 2022-10-09
TGUI Clothing Vendors

mullenpaul:
## About The Pull Request

Migrates the clothing vendor to use tgui. Updates existing tgui vendors to new style feedback.

## Why It's Good For The Game

This would mean two of the three big cm vending machine types are working with tgui.

image

image

image

## Changelog

:cl:
refactor: uscm clothing vendors now use tgui
refactor: added alt text for icons on cm vendor
refactor: updated existing tgui vendors to new style
qol: when using cm vendor ui you can close by pressing escape
refactor: vendor icons now use spritesheet for asset transfer
refactor: cm vendors now inform UI about what kind of vendor it is (clothing/ sorted)
refactor: updated uscm role specific vending machines to use global item list via get_listed_products
refactor: clothing item list for synth snowflake now global list
refactor: uscm dress vendor rewritten with custom tgui code because of how item assignment is made
/:cl:


stanalbatross:
image
example image - font size is almost exactly the same

mullenpaul:
There are issues with the synth vendors, these issues need to be resolved:

- Its completely fucked the synthetic equipment rack. Only capable of vending clothing at 12 points each even though its supposed to hold the surgical webbing, tool webbing, etc
- The clothing one IS the snowflake one. The one with 45 points is for synth equipment such as tool webbing or surgical webbing, your belt choice, welding helmet, etc.
and your experimental tool token 
- The middle clothing one should NOT use points, it should give you one selection from each category. The right-most one is for gear and uses points, the left snowflake one uses points

code/game/machinery/vending/cm_vending.dm
```diff
@@ -95,6 +97,71 @@ IN_USE						used for vending/denying
 			return
 	return
 
+// Vendor Icon Procs
+
+/obj/structure/machinery/cm_vending/proc/preload_assets()
+	var/datum/asset/simple/dynamic_icons/dyn = get_asset_datum(/datum/asset/simple/dynamic_icons)
+	for (var/list/i in product_icon_list)
```
stanalbatross, line 104: I believe you can use `as anything` here since it's always going to be lists in the list

code/game/machinery/vending/cm_vending.dm
```diff
@@ -95,6 +97,71 @@ IN_USE						used for vending/denying
...
+		var/filename = i["href"]
+		if(filename)
+			dyn.update(filename)
+
+/obj/structure/machinery/cm_vending/proc/build_icons(var/list/items, var/name_index=1, var/type_index=3)
+	for (var/list/item in items)
```
stanalbatross, line 110: same

code/game/machinery/vending/cm_vending.dm
```diff
@@ -95,6 +97,71 @@ IN_USE						used for vending/denying
...
+		// initial item count setup
+		var/item_name = item[name_index]
+		// icon setup
+		var/typepath = item[type_index]
+		if (item_name == null || item_name == "" || typepath == null)
```
stanalbatross, line 115: `var == null` can be changed to `!var`

code/game/machinery/vending/cm_vending.dm
```diff
@@ -95,6 +97,71 @@ IN_USE						used for vending/denying
...
+		if(!(item_name in product_icon_list))
+			product_icon_list[item_name] = list(
+				"href" = "",
+				"desc" = ""
+			)
+		if(!r)
+			continue
+		var/asset_name = generate_asset_name(r)
+		var/key = "[asset_name].png"
+		if(asset_name != null && asset_name != "" && r != null)
```
stanalbatross, line 155: `var != null` can just be `var`

code/game/machinery/vending/cm_vending.dm
```diff
@@ -766,6 +774,218 @@ IN_USE						used for vending/denying
...
+					return
+
+			vend_succesfully(L, H, T)
+		if("cancel")
+			SStgui.close_uis(src)
+			return TRUE
+
+	add_fingerprint(usr)
+
+/obj/structure/machinery/cm_vending/clothing/tgui_interact(mob/user, datum/tgui/ui)
```
stanalbatross, line 954: do we need autoupdate here? all vends are clientside only (the stock can't change due to other people) so I would consider turning it off as it'd likely help with stress/strain of asset transport

code/game/machinery/vending/cm_vending.dm
```diff
@@ -766,6 +774,218 @@ IN_USE						used for vending/denying
...
+		ui = new(user, src, "VendingSorted", name)
+		ui.open()
+
+/obj/structure/machinery/cm_vending/clothing/attack_hand(mob/user)
+	if(stat & TIPPED_OVER)
+		if(user.action_busy)
+			return
+		user.visible_message(SPAN_NOTICE("[user] begins to heave the vending machine back into place!"),SPAN_NOTICE("You start heaving the vending machine back into place."))
+		if(do_after(user, 80, INTERRUPT_NO_NEEDHAND, BUSY_ICON_FRIENDLY))
+			user.visible_message(SPAN_NOTICE("[user] rights the [src]!"),SPAN_NOTICE("You right the [src]!"))
```
stanalbatross, line 966:
```suggestion
			user.visible_message(SPAN_NOTICE("[user] rights \the [src]!"),SPAN_NOTICE("You right \the [src]!"))
```

stanalbatross (changes requested):

code/game/machinery/vending/cm_vending.dm
mullenpaul: If we don't have the ui auto update on use, then the user isn't informed about what they have gotten. For example, the buttons don't grey out after vending and it looks as though I can vend another

stanalbatross: ah, makes sense

code/game/machinery/vending/cm_vending.dm
mullenpaul: This seems to cause an issue

code/game/machinery/vending/cm_vending.dm
mullenpaul: This seems to cause an issue

mullenpaul:
Ticket has been updated and should be good to review.

mullenpaul:
All existing issues resolved. There might be some code cleanup to get on with but ready for TM.

code/game/machinery/vending/cm_vending.dm
```diff
@@ -465,6 +481,16 @@ IN_USE						used for vending/denying
 	update_icon()
 	return
 
+//-----------TGUI PROCS------------------------
+/obj/structure/machinery/cm_vending/ui_static_data(mob/user)
+	. = list()
```
stanalbatross, line 486: `.`-hijacking is generally discouraged, would be better to do var/list/data = list() and then return data

code/game/machinery/vending/cm_vending.dm
```diff
@@ -766,6 +724,218 @@ IN_USE						used for vending/denying
...
+		return
+
+	var/mob/living/carbon/human/H = usr
+	switch (action)
+		if ("vend")
+			if(stat & IN_USE)
+				return
+			var/has_access = can_access_to_vend(usr)
+			if (!has_access)
+				return TRUE
```
stanalbatross, line 850: isn't this already being handled in UI status?

code/game/machinery/vending/cm_vending.dm
```diff
@@ -766,6 +724,218 @@ IN_USE						used for vending/denying
...
+			var/idx=params["prod_index"]
+
+			var/list/topic_listed_products = get_listed_products(usr)
+			var/list/L = topic_listed_products[idx]
+			var/cost = L[2]
+
+			if((!H.assigned_squad && squad_tag) || (!H.assigned_squad?.omni_squad_vendor && (squad_tag && H.assigned_squad.name != squad_tag)))
+				to_chat(H, SPAN_WARNING("This machine isn't for your squad."))
+				vend_fail()
+				return
```
stanalbatross, line 861: do checks in UI itself not ui act

code/game/machinery/vending/cm_vending.dm
```diff
@@ -766,6 +724,218 @@ IN_USE						used for vending/denying
...
+						vend_fail()
+						return
+					else
+						H.marine_snowflake_points -= cost
+				else
+					if(H.marine_points < cost)
+						to_chat(H, SPAN_WARNING("Not enough points."))
+						vend_fail()
+						return
+					else
```
stanalbatross, line 888: same here, aren't these checks already being done in the ui?

code/game/machinery/vending/cm_vending.dm
```diff
@@ -766,6 +724,218 @@ IN_USE						used for vending/denying
...
+	if(!ishuman(user))
+		vend_fail()
+		return
+
+	var/has_access = can_access_to_vend(user)
+	if (!has_access)
+		return
+
+	user.set_interaction(src)
+	tgui_interact(user)
```
stanalbatross, line 936: unreachable code

code/game/machinery/vending/vendor_types/antag/antag_clothing.dm
```diff
@@ -9,14 +9,21 @@
 
 	listed_products = list()
 
-/obj/structure/machinery/cm_vending/clothing/antag/ui_interact(mob/user, ui_key = "main", var/datum/nanoui/ui = null, var/force_open = 0)
+/obj/structure/machinery/cm_vending/clothing/antag/get_listed_products(var/mob/user)
+	if(!user)
+		var/list/all_equipment = list()
+		var/list/presets = typesof(/datum/equipment_preset)
+		for(var/i in presets)
+			var/datum/equipment_preset/eq = new i
```
stanalbatross, line 17: shouldn't we be deleting this eq

stanalbatross (changes requested):

code/game/machinery/vending/vendor_types/antag/antag_clothing.dm
mullenpaul: Err, the return is just a list of lists. I don't think we need to delete this?

code/game/machinery/vending/cm_vending.dm
mullenpaul: Gotcha.

code/game/machinery/vending/cm_vending.dm
mullenpaul: As discussed in discord - the front and back end checks serve different purposes

code/game/machinery/vending/cm_vending.dm
mullenpaul: As discussed in discord - the front and back end checks serve different purposes

code/game/machinery/vending/cm_vending.dm
mullenpaul: I'm not sure how this is unreachable - I use several early-fail checks to back out of the proc. If none of these checks fail, we update the UI.

code/game/machinery/vending/cm_vending.dm
mullenpaul: Yeah - I'm still not 100% sure about how all ui interactions are done.

code/game/machinery/vending/vendor_types/antag/antag_clothing.dm
mullenpaul: I just misread this. Good catch.

PR #1220, opened 2022-10-10
Adds Xeno QOL improvement by allowing for snow layers to be autocleared

Youbar:
## About The Pull Request

Bungled the first PR (merged unrelated code from what should have been a separate branch into it) so I've resubmitted this one.

To clear snow on Shivas Snowball and the likes, Xenos have to manually click for each layer of snow removed. This PR makes it so that one single click will continue to clear additional layers if they exist, thus serving as a small QOL boost.

## Why It's Good For The Game

Serves as a QOL boost for Xenos without fundamentally impacting game balance, similar to the automatic filling of sandbags mechanism.

## Changelog

:cl: Youbar
qol: Xenos can now clear several layers of snow with one click
/:cl:

### DIFF
--- a/code/game/turfs/auto_turf.dm
+++ b/code/game/turfs/auto_turf.dm
@@ -193,7 +193,7 @@
 	if(M.a_intent == INTENT_HELP || !bleed_layer)
 		return XENO_NO_DELAY_ACTION
 
-	M.visible_message(SPAN_NOTICE("[M] starts clearing out [src]."), SPAN_NOTICE("You start clearing out [src]."), null, 5, CHAT_TYPE_XENO_COMBAT)
+	M.visible_message(SPAN_NOTICE("[M] starts clearing out \the [src]."), SPAN_NOTICE("You start clearing out \the [src]."), null, 5, CHAT_TYPE_XENO_COMBAT)
 	playsound(M.loc, 'sound/weapons/alien_claw_swipe.ogg', 25, 1)
 
 	while(bleed_layer > 0)
@@ -206,8 +206,8 @@
 			to_chat(M, SPAN_WARNING("There is nothing to clear out!"))
 			return XENO_NO_DELAY_ACTION
 
-		M.visible_message(SPAN_NOTICE("[M] clears out [src]."), \
-		SPAN_NOTICE("You clear out [src]."), null, 5, CHAT_TYPE_XENO_COMBAT)
+		M.visible_message(SPAN_NOTICE("[M] clears out \the [src]."), \
+		SPAN_NOTICE("You clear out \the [src]."), null, 5, CHAT_TYPE_XENO_COMBAT)
 
 		var/new_layer = bleed_layer - 1
 		changing_layer(new_layer)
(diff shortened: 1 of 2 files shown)


code/game/turfs/auto_turf.dm
```diff
@@ -193,21 +193,25 @@
 	if(M.a_intent == INTENT_HELP || !bleed_layer)
 		return XENO_NO_DELAY_ACTION
 
-	M.visible_message(SPAN_NOTICE("[M] starts clearing out the [name]."), SPAN_NOTICE("You start clearing out the [name]."), null, 5, CHAT_TYPE_XENO_COMBAT)
+	M.visible_message(SPAN_NOTICE("[M] starts clearing out \the [src]."), SPAN_NOTICE("You start clearing out \the [src]."), null, 5, CHAT_TYPE_XENO_COMBAT)
```
Geevies, line 196:
```suggestion
	M.visible_message(SPAN_NOTICE("[M] starts clearing out \the [src]..."), SPAN_NOTICE("You start clearing out \the [src]..."), null, 5, CHAT_TYPE_XENO_COMBAT)
```

code/game/turfs/auto_turf.dm
```diff
@@ -193,21 +193,25 @@
...
 	playsound(M.loc, 'sound/weapons/alien_claw_swipe.ogg', 25, 1)
-	xeno_attack_delay(M)
-	if(!do_after(M, 12, INTERRUPT_ALL, BUSY_ICON_FRIENDLY))
-		return XENO_NO_DELAY_ACTION
 
-	if(!bleed_layer)
-		to_chat(M, SPAN_WARNING("There is nothing to clear out!"))
-		return XENO_NO_DELAY_ACTION
+	while(bleed_layer > 0)
+
```
Geevies, line 200:
```suggestion
```

code/game/turfs/auto_turf.dm
```diff
@@ -193,21 +193,25 @@
...
+		xeno_attack_delay(M)
+		if(!do_after(M, 12, INTERRUPT_ALL, BUSY_ICON_FRIENDLY))
+			return XENO_NO_DELAY_ACTION
+
+		if(!bleed_layer)
+			to_chat(M, SPAN_WARNING("There is nothing to clear out!"))
+			return XENO_NO_DELAY_ACTION
+
+		M.visible_message(SPAN_NOTICE("[M] clears out \the [src]."), \
```
Geevies, line 209: i don't thiiiiiiink you need a message for each clear, since the changing turf should be enough of a visual indicator

Geevies (commented):
not 100% on this, since the intent of snow is to be a pain in the ass and adding more QoL makes it less of a PITA. might be better to keep manual clearing terrible and make spitting acid and crusher charging clear snow instead

Youbar:
> not 100% on this, since the intent of snow is to be a pain in the ass and adding more QoL makes it less of a PITA. might be better to keep manual clearing terrible and make spitting acid and crusher charging clear snow instead

Snow maps are already disliked by Xenos and the prospect of clearing snow tends to turn the match into a throw. The time cost involved also eats into how long hivelords and drones can spend prepping defenses and puts Xenos at a comparative disadvantage to marines. The argument could also be made that frontliners like spitters and chargers should be frontlining and not facilitating logistics (not to exclude the addition of that functionality as it is something I would like to develop).

Geevies (approved):
yebb

PR #1225, opened 2022-10-10
Removes the +FUN admin permission, gives smod more event verbs

silencer-pl:
## About The Pull Request
This PR removes all references to the +FUN admin permission by removing all R_FUN references from the code. +FUN is a legacy TG permission and there is currently zero verbs or functions in the code that require it alone. The classification is also somewhat confusing and completely redundant with current perms as they are.

Additionally, two Admin Verbs previously only available to Administrators have now been made available for Senior Moderators as well: Distress Beacon (to call ERTs, needed for the upcoming policy changes related to the Investigation ERTs I put in) and Toggle Remote Control (Which is an extremely harmless tool - it allows admins to open and/or bolt doors as well as break lights) 

## Why It's Good For The Game
Considering that there literally wasn't a single instance of the +FUN permission in the code that would not also have another permission, the group was literally redundant. It was also clearly a legacy TG thing. Organizing things by MOD/ADMIN and additionally DEBUG and SPAWN makes way more sense.

Senior Moderators are about to get permission to use certain ERTs as reactive events without the need for asking for auth. It would be nice if they had access to what they will have permissions to use, hence the Distress Beacon use (senior mods already can send ERTs via Create Human anyway so its not like this gives them anything other than access to the fancy pre-prepared ERTs with nice flavor text and objectives.

The primary use of Toggle Remote Control is closing and locking of doors, something that again is an indispensable tool and that senior moderators can already do via varedit. There is no reason why smods should not have access to it.

## Changelog

:cl:Silencer_pl
admin: Removed the redundant legacy +FUN fun admin permission and scrubbed all references to it.
admin: Senior Moderators now have access to the Distress Beacon and Toggle Remote Control verbs. 
/:cl:


code/__HELPERS/type2type.dm
```diff
@@ -207,7 +207,6 @@
 	if(rights & R_BUILDMODE)	. += "[seperator]+BUILDMODE"
 	if(rights & R_ADMIN)		. += "[seperator]+ADMIN"
 	if(rights & R_BAN)			. += "[seperator]+BAN"
-	if(rights & R_FUN)			. += "[seperator]+FUN"
```
stanalbatross, line 210: also remove from example config

code/global.dm
```diff
@@ -7,7 +7,6 @@
 #define R_BUILDMODE		(1<<0)
 #define R_ADMIN			(1<<1)
 #define R_BAN			(1<<2)
-#define R_FUN			(1<<3)
```
stanalbatross, line 10: shift all the other ones below up one (i.e. R_SERVER (1<<3)

stanalbatross (changes requested):

code/modules/admin/admin_verbs.dm
```diff
@@ -287,8 +287,6 @@ var/list/roundstart_mod_verbs = list(
 	if(CLIENT_HAS_RIGHTS(src, R_BAN))
 		add_verb(src, admin_verbs_ban)
 		add_verb(src, admin_verbs_teleport) // ???
-	if(CLIENT_HAS_RIGHTS(src, R_FUN))
-		add_verb(src, admin_verbs_fun)
```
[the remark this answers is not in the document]
silencer-pl: Moved to R_ADMIN

code/global.dm
silencer-pl: There was one number already missing, so I had originally left this alone. Rearranged all of them to be in sequence now.

stanalbatross (approved):

PR #1230, opened 2022-10-10
Tailstab Hotfix

realforest2001:
## About The Pull Request
An imperfect fix to tailstab going through doors/windows & blast doors among other things. If it's solid and you can't throw an object over it, you can't stab through it.

## Why It's Good For The Game
Stops cheese.

## Changelog

:cl:
fix: Fixes tailstab going through blocked tiles.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
@@ -858,12 +858,12 @@
 			to_chat(stabbing_xeno, SPAN_WARNING("There's something blocking your strike!"))
 			return FALSE
 		for(var/obj/path_contents in path_turf.contents)
-			if(path_contents.density && !path_contents.throwpass)
+			if(path_contents != targetted_atom && path_contents.density && !path_contents.throwpass)
 				to_chat(stabbing_xeno, SPAN_WARNING("There's something blocking your strike!"))
 				return FALSE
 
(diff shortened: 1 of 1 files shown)


Carnage068:
agreed. cheese is disgusting- it smells awful. it must be stopped.

code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
```diff
@@ -857,6 +857,11 @@
 		if(path_turf.density)
 			to_chat(stabbing_xeno, SPAN_WARNING("There's something blocking your strike!"))
 			return FALSE
+		for(var/obj/path_contents as obj in path_turf.contents)
+			if(path_contents.density && !path_contents.throwpass)
```
Geevies, line 861: not so sure about this throwpass check, is there a better way to check if it blocks `(PASS_MOB_THRU_XENO|PASS_OVER_THROW_MOB|PASS_TYPE_CRAWLER)` specifically?

Geevies (changes requested):

code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
realforest2001: I don't know. I can't tell why the barrier code isn't working. The only thing I could get to work reliably is that combination.

Geevies: hmm, in that case maybe filter out stuff that's ONBORDER like barricades (check the flag) so you can at least still stab over those in accordance to the other bits of code

morrowwolf (approved):
I approve the base code and Geeves approves my tiny change, let's get this boy out there.

PR #1234, opened 2022-10-12
Removes Warrior Shield, Implements Lifesteal and Cooldown on Slash Reductions (wit no bad rebase)

usnpeepoo:
## About The Pull Request
Warrior slash shield has been removed, in place for it is lifesteal from slashes and a slight cooldown reduction per slash. Warrior also gains 50 more health. The lifesteal is comparable to shield, but requires 4 slashes to gain. The %ge of health gained increases with more marines near the warrior by 3 tiles up to a maximum of 9 (2 marines = 2% more lifesteal). This increased health shows a glow, and a cue in chat when it happens. Warrior also gains a very slight cooldown reduction per slash (0.30 seconds worth).
Huge thanks to Nivrak for allowing me to reuse their lifesteal PR, and helping me with the visual cue for it.

## Why It's Good For The Game
Warriors shield is unhealthy for current CM gameplay, and this PR hopes to fix it through something similar to berserkers system of life steal. With shields, warriors have an invisible (to marines) extra health bar, that is permanent and always recoupable. By removing it, and implementing life steal and slash cooldown reduction warriors no longer have the points mentioned above, rather they are rewarded more intuitively for hitting abilities, finding opportunities to hit their slashes a lot more, and being rewarded for doing so without giving them a shield. By allowing warrior to heal and acquire stronger heals by being more risky with their abilities, and also giving them less cooldowns to initiate fights with per slash, warriors are now able to fit the role of a frontline-ambusher caste that can continue fighting through the use of its abilities while dealing with the issue with shields being in that role.

This PR also aims to be in line with the xeno shield design doc, and seeks to bring one of the goals to light by making warrior buff itself in a different way than simply shields.
You can read the design doc here, https://github.com/cmss13-devs/cmss13-design-docs/blob/main/xeno/shields.md.

## Changelog
🆑Usnpeepoo, Nivrak
balance: Removed shielding slash from warrior. Warrior slashes now heal for (7 + 1 per enemy within 3 tiles, up to 2)% of missing health with a minimum of 20 and a maximum of 40. There is now a visual, and chat cue for when a warrior gains extra health per marine.
balance: Warriors health is reverted back to 500, up from 450.
balance: Warriors slash now reduces cooldowns from all abilities by 0.30 seconds worth.
/🆑

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/Warrior.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Warrior.dm
@@ -5,7 +5,7 @@
 	melee_damage_lower = XENO_DAMAGE_TIER_3
 	melee_damage_upper = XENO_DAMAGE_TIER_5
 	melee_vehicle_damage = XENO_DAMAGE_TIER_5
-	max_health = XENO_HEALTH_TIER_5
+	max_health = XENO_HEALTH_TIER_6
 	plasma_gain = XENO_PLASMA_GAIN_TIER_9
 	plasma_max = XENO_NO_PLASMA
 	xeno_explosion_resistance = XENO_EXPLOSIVE_ARMOR_TIER_4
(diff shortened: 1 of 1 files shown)


usnpeepoo:
pr 1234 this is so awesome

Wintermote:
The biggest problem with the shield is that it protects against stuns, which is criminally overpowered in this game that depends so heavily on stuns for killing things.

The lifesteal just gives more uptime in active combat which imo is good for warriors who should be one of the main combat castes. It'll probably be better balanced than ravagers, which I don't agree with having lifesteal + so much damage reduction and escape-from-jail abilities.

morrowwolf:
Merging in 18 hours unless architects weigh in.

PR #1243, opened 2022-10-13
Fixed the issue with plastic explosives cause_data and added door type

aceroyuno:
## About The Pull Request

Fixes the issue of plastic explosives calling handle_explosion with null cause data.
And added a new poddoor/almayer type, this one not destroyable by the APC but still destroyable by C4 and acid.

## Why It's Good For The Game

It's a bug fix and an new door type for mappers to use.

## Changelog

:cl: Aceroy
fix: Fixed the issue of plastic explosives cause_data being deleted in the process of exploding
add: Added a new poddoor type that is immune to APC but still can be destroyed by explosives and acid.
/:cl:

### DIFF
--- a/code/game/machinery/doors/poddoor.dm
+++ b/code/game/machinery/doors/poddoor.dm
@@ -274,6 +274,7 @@
 /obj/structure/machinery/door/poddoor/almayer
 	icon = 'icons/obj/structures/doors/blastdoors_shutters.dmi'
 	openspeed = 4 //shorter open animation.
+	var/bumpable = TRUE
 	tiles_with = list(
 		/obj/structure/window/framed/almayer,
 		/obj/structure/machinery/door/airlock)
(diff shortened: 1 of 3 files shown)


code/game/machinery/doors/poddoor.dm
```diff
@@ -274,6 +274,7 @@
 /obj/structure/machinery/door/poddoor/almayer
 	icon = 'icons/obj/structures/doors/blastdoors_shutters.dmi'
 	openspeed = 4 //shorter open animation.
+	var/bumpable = TRUE
```
Geevies, line 277: this variable isn't super clear, call it something like `vehicle_resistant` and then swap the current logic around

Geevies (changes requested):

code/game/machinery/doors/poddoor.dm
aceroyuno: var name has been changed from bumpable to vehicle_resistant

PR #1250, opened 2022-10-14
Vehicle broadcaster

JeserTheFox:
## About The Pull Request

1. Added Use Broadcaster verb to armored vehicles, has 5 seconds recharge.
2. Added vehicle specific long langchat proc that needs finishing.
3. Added categories to vehicle verbs.

## Why It's Good For The Game

VCs finally get a way to communicate with marines directly around them.

## Changelog

:cl: Jeser
add: Added Use Broadcaster verb to armored vehicles. Allows VCs to directly address marines near vehicle. Has 5 seconds recharge.
qol: Vehicle verbs now have more compact names and received categories.
/:cl:

### DIFF
--- a/code/datums/langchat/langchat.dm
+++ b/code/datums/langchat/langchat.dm
@@ -151,6 +151,48 @@
 	else
 		addtimer(CALLBACK(src, /atom.proc/langchat_drop_image, language), timer, TIMER_OVERRIDE|TIMER_UNIQUE|TIMER_NO_HASH_WAIT)
 
+//used for broadcasting from a vehicle
+/mob/proc/langchat_vehicle_broadcast(message, var/list/listeners, language, var/obj/vehicle/multitile/V)
+
+	//uncomment before finishing
+	//if(!V || interactee != V)
+	//	return
+
+	langchat_drop_image()
+	langchat_make_image()
+
+	var/text_left = null
+	var/text_to_display = message
+
+	if(length(message) > LANGCHAT_LONGEST_TEXT)
+		text_to_display = copytext_char(message, 1, LANGCHAT_LONGEST_TEXT - 5) + "..."
+		text_left = "..." + copytext_char(message, LANGCHAT_LONGEST_TEXT - 5)
+	var/timer = 6 SECONDS
+	if(text_left)
+		timer = 4 SECONDS
+	text_to_display = "<span class='center [langchat_styles] langchat_announce langchat'>[text_to_display]</span>"
+
+	langchat_image.maptext = text_to_display
+	langchat_image.maptext_width = LANGCHAT_WIDTH * 2
+
+	langchat_listeners = listeners
+	for(var/mob/M in langchat_listeners)
+		if(langchat_client_enabled(M) && !M.ear_deaf && M.say_understands(src, language))
+			M.client.images += langchat_image
+
+	//uncomment before finishing
+	//if(!isVehicleMultitile(interactee))
+	//	return
+
+	langchat_image.loc = interactee
+	langchat_image.maptext_y += 32
+
+	animate(langchat_image, pixel_y = langchat_image.pixel_y + LANGCHAT_MESSAGE_POP_Y_SINK, alpha = LANGCHAT_MAX_ALPHA, time = LANGCHAT_MESSAGE_POP_TIME)
+	if(text_left)
+		addtimer(CALLBACK(src, /mob.proc/langchat_long_speech, text_left, listeners, language), timer, TIMER_OVERRIDE|TIMER_UNIQUE|TIMER_NO_HASH_WAIT)
+	else
+		addtimer(CALLBACK(src, /mob.proc/langchat_drop_image, language), timer, TIMER_OVERRIDE|TIMER_UNIQUE|TIMER_NO_HASH_WAIT)
+
 /** Displays image to a single listener after it was built above eg. for chaining different game logic than speech code
 This does just that, doesn't check deafness or language! Do what you will in that regard **/
 /atom/proc/langchat_display_image(var/mob/M)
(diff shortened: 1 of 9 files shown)


code/datums/langchat/langchat.dm
```diff
@@ -151,6 +151,43 @@
 	else
 		addtimer(CALLBACK(src, /atom.proc/langchat_drop_image, language), timer, TIMER_OVERRIDE|TIMER_UNIQUE|TIMER_NO_HASH_WAIT)
 
+//used for broadcasting from a vehicle
+/mob/proc/langchat_vehicle_broadcast(message, var/list/listeners, language, var/obj/vehicle/multitile/V)
```
stanalbatross, line 155: better var name neded for V

code/datums/langchat/langchat.dm
```diff
@@ -151,6 +151,43 @@
...
+	var/timer = 6 SECONDS
+	if(text_left)
+		timer = 4 SECONDS
+	text_to_display = "<span class='center [langchat_styles] langchat_announce langchat'>[text_to_display]</span>"
+
+	langchat_image.maptext = text_to_display
+	langchat_image.maptext_width = LANGCHAT_WIDTH * 2
+
+	langchat_listeners = listeners
+	for(var/mob/M in langchat_listeners)
```
stanalbatross, line 178: better var name needed for M

code/modules/vehicles/multitile/multitile_verbs.dm
```diff
@@ -327,12 +327,87 @@
...
+/obj/vehicle/multitile/proc/use_broadcaster()
+	set name = "Use Broadcaster"
+	set desc = "Allows to send a loud message to the exterior via hull broadcaster."
+	set category = "Vehicle.General"
+
+	var/mob/living/user = usr
+	if(!istype(user))
+		return
+
+	var/obj/vehicle/multitile/V = user.interactee
```
stanalbatross, line 340: var name for V

code/modules/vehicles/multitile/multitile_verbs.dm
```diff
@@ -327,12 +327,87 @@
 	if(honk_sound)
 		playsound(loc, honk_sound, 75, TRUE, 15)	//heard within ~15 tiles
 
+//Finally, a broadcaster!
+/obj/vehicle/multitile/proc/use_broadcaster()
+	set name = "Use Broadcaster"
+	set desc = "Allows to send a loud message to the exterior via hull broadcaster."
+	set category = "Vehicle.General"
```
stanalbatross, line 334: /proc and it's using set? doesn't check out (also this would be better as an action button anyway)

code/modules/vehicles/multitile/multitile_verbs.dm
```diff
@@ -327,12 +327,87 @@
...
+			O.show_message("\The <B>[V]</B> broadcasts, [FONT_SIZE_LARGE("\"[message]\"")]",2) // 2 stands for hearable message
+
+		for(seat in V.seats)
+			if(ishuman(V.seats[seat]))
+				recipients += V.seats[seat]
+
+		user.langchat_vehicle_broadcast(message, recipients, user.get_default_language(), V)
+
+		V.broadcaster_available = FALSE
+		addtimer(VARSET_CALLBACK(V, broadcaster_available, TRUE), 5 SECONDS)
```
stanalbatross, line 403: use COOLDOWN_DECLARE system here rather than callback timers for cooldowns

stanalbatross (changes requested):

code/datums/langchat/langchat.dm
JeserTheFox: 1. Why?
2. It's how it is in other two procs.

code/modules/vehicles/multitile/multitile_verbs.dm
JeserTheFox: Well, that's how vehicle controls work.

code/modules/vehicles/multitile/multitile_verbs.dm
JeserTheFox: Done.

code/modules/vehicles/multitile/multitile_verbs.dm
JeserTheFox: Done

code/datums/langchat/langchat.dm
JeserTheFox: Done

fira (approved):
Looks good to me. We could TM it quick

fira:
This had issues during TM'ing, both VCs started having their langchat displaying much much higher than normal. For one of them, it was like 10 turfs above, so it couldn't be seen at all. I'm guessing this is because of the broadcaster

code/datums/langchat/langchat.dm
```diff
@@ -151,6 +151,43 @@
...
+	langchat_image.maptext = text_to_display
+	langchat_image.maptext_width = LANGCHAT_WIDTH * 2
+
+	langchat_listeners = listeners
+	for(var/mob/Listener in langchat_listeners)
+		if(langchat_client_enabled(Listener) && !Listener.ear_deaf && Listener.say_understands(src, language))
+			Listener.client.images += langchat_image
+
+	langchat_image.loc = interactee
+	langchat_image.maptext_y += 32
```
fira, line 183: this might need to be reset

fira (changes requested):

code/datums/langchat/langchat.dm
JeserTheFox: Resetting didn't work, had to figure out other way.

fira: that doesn't look correct to me because the maptext offset will revert after first message if it's a long message segmented

one solution i could see is to mark the langchat_image as "expendable" and if it is, delete it completely after display. that way a new one is created from scratch next time.

fira (changes requested):

PR #1256, opened 2022-10-14
Slur Blocker

realforest2001:
## About The Pull Request
Prevents slurs being used in-game. Changes may yet be needed, pending review for code quality. Potentially will need regex integration however this is beyond my ability at the moment.
Made at request of Management.

## Why It's Good For The Game
More welcoming place, and reduces the amount of time staff will need to deal with people using slurs.

## Changelog

:cl:
add: Slurs are now blocked ingame and messages containing slurs will not be sent. List of slurs is determined by server config file.
/:cl:

### DIFF
--- a/code/__HELPERS/logging.dm
+++ b/code/__HELPERS/logging.dm
@@ -37,6 +37,8 @@
 	GLOB.STUI.processing |= STUI_LOG_ADMIN
 
 /proc/check_slur(var/text, var/mob/source, var/method = "Speech")
+	if(!CONFIG_GET(flag/log_slurs))
+		return FALSE
 	var/list/TriggerWords = file2list("config/banned_words.txt")
 	for(var/Trigger in TriggerWords)
 		if(findtext(text,Trigger))
--- a/code/__HELPERS/logging.dm
+++ b/code/__HELPERS/logging.dm
@@ -36,6 +36,13 @@
 	GLOB.STUI.admin.Add("\[[time_stamp()]]ADMIN: [text]")
 	GLOB.STUI.processing |= STUI_LOG_ADMIN
 
+/proc/check_slur(var/text, var/mob/source, var/method = "Speech")
+	var/list/TriggerWords = file2list("config/banned_words.txt")
+	for(var/Trigger in TriggerWords)
+		if(findtext(text,Trigger))
+			log_slur(source, Trigger, text)
+			return TRUE //Blocks message being sent.
+
 /proc/log_slur(var/mob/origin, var/word, var/message)
 	admin_log.Add("[key_name(origin)] had a message flagged for banned word '[word]'. Message reads: '[message]'")
 	if (CONFIG_GET(flag/log_slurs))
--- a/code/__HELPERS/logging.dm
+++ b/code/__HELPERS/logging.dm
@@ -36,6 +36,13 @@
 	GLOB.STUI.admin.Add("\[[time_stamp()]]ADMIN: [text]")
 	GLOB.STUI.processing |= STUI_LOG_ADMIN
 
+/proc/log_slur(var/mob/origin, var/word, var/message)
+	admin_log.Add("[key_name(origin)] had a message flagged for banned word '[word]'. Message reads: '[message]'")
+	if (CONFIG_GET(flag/log_slurs))
+		diary << html_decode("\[[time_stamp()]]SLUR: [key_name(origin)] had a message flagged for banned word '[word]'. Message reads: '[message]'[log_end]")
+	GLOB.STUI.admin.Add("\[[time_stamp()]]SLUR: [key_name(origin)] had a message flagged for banned word '[word]'. Message reads: '[message]'")
+	GLOB.STUI.processing |= STUI_LOG_ADMIN
+
 /proc/log_asset(text)
 	asset_log.Add(text)
 	if (CONFIG_GET(flag/log_asset))
--- a/code/__HELPERS/logging.dm
+++ b/code/__HELPERS/logging.dm
@@ -36,18 +36,20 @@
 	GLOB.STUI.admin.Add("\[[time_stamp()]]ADMIN: [text]")
 	GLOB.STUI.processing |= STUI_LOG_ADMIN
 
-/proc/check_slur(var/text, var/mob/source, var/method = "Speech")
-	if(!CONFIG_GET(flag/log_slurs))
+
+GLOBAL_LIST_INIT(SlurWords, file2list("config/slur_words.txt"))
+/proc/check_slur(var/mob/source, var/text, var/method = "Speech")
+	if(!CONFIG_GET(flag/block_slurs))
 		return FALSE
-	var/list/TriggerWords = file2list("config/banned_words.txt")
-	for(var/Trigger in TriggerWords)
+	for(var/Trigger in GLOB.SlurWords)
 		if(findtext(text,Trigger))
 			log_slur(source, Trigger, text)
+			to_chat(source, SPAN_WARNING("Your message, '[text]' has been blocked due to containing the word '[Trigger]'."))
 			return TRUE //Blocks message being sent.
 
 /proc/log_slur(var/mob/origin, var/word, var/message)
 	admin_log.Add("[key_name(origin)] had a message flagged for banned word '[word]'. Message reads: '[message]'")
-	if (CONFIG_GET(flag/log_slurs))
+	if (CONFIG_GET(flag/block_slurs))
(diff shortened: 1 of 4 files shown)


johnravioliUSA:
👎

carlarctg:
code looks good, but may want to add an ingame toggle just for consistency's sake (and if it somehow turns to zero)

now.... please do tell me players who get their slur message blocked _arent_ bwoinked and noted for it. that's not code related but really seems like a step too far

PhantomEpicness:
We turning into a Christian SS-13 server.

Should use it to block netspeak instead of blocking swear words. Can't wait to get banned for saying "fuck"

Chainsawicus:
There is no such thing as an effective and non-intrusive chat censor, this will either be entirely overzealous or completely pointless. 

This implementation immediately seems to lean towards being overzealous, especially since you included a chat message sent to all staff online so they can seemingly punish people for something that the game already blocked from being posted.

The only impact this will have on the game is a negative one.

realforest2001:
In regards to punishments on this, while it will log to the server file, at the very least to explain why a message was flagged, it does not actively tell staff mid round that someone said something. Nor will there be persecution for blocked words, as with the discord if something is AUTOMATICALLY removed due to the filters in place, punishment is not furthered because it's not something staff have to act on.
Based upon the discussions in management I see no intent to punish people twice over, and should it be the case where I am wrong about that, I will not partake or enable it as I believe that to be wrong.

As for the if this should or should not be done, it's not open to widespread discussion. There is a discord post for discussion that should be used regardless however.

code/__HELPERS/logging.dm
```diff
@@ -36,6 +36,22 @@
 	GLOB.STUI.admin.Add("\[[time_stamp()]]ADMIN: [text]")
 	GLOB.STUI.processing |= STUI_LOG_ADMIN
 
+/proc/check_slur(var/text, var/mob/source, var/method = "Speech")
+	if(!CONFIG_GET(flag/log_slurs))
+		return FALSE
+	var/list/TriggerWords = file2list("config/banned_words.txt")
+	for(var/Trigger in TriggerWords)
+		if(findtext(text,Trigger))
```
stanalbatross, line 44: gotta make sure this won't trigger false positives for things like the word "sniggers"

code/__HELPERS/logging.dm
```diff
@@ -36,6 +36,22 @@
 	GLOB.STUI.admin.Add("\[[time_stamp()]]ADMIN: [text]")
 	GLOB.STUI.processing |= STUI_LOG_ADMIN
 
+/proc/check_slur(var/text, var/mob/source, var/method = "Speech")
+	if(!CONFIG_GET(flag/log_slurs))
+		return FALSE
+	var/list/TriggerWords = file2list("config/banned_words.txt")
```
Geevies, line 42: this should be cached so it doesn't read file2list all the time, and additionally be a regex. i gave it some thought and the best way to handle the scunthorpe problem would be to make the regex check for whitespace, common punctuation, and line ends surrounding the word. that way it won't trigger for everything and also be more performant than looping for a list

Geevies: additionally, instead of being "banned_words.txt", call it "slur_words.txt" for clarity, since we might want to ban other words in the near future winkwinknudgenudge

code/__HELPERS/logging.dm
```diff
@@ -36,6 +36,22 @@
 	GLOB.STUI.admin.Add("\[[time_stamp()]]ADMIN: [text]")
 	GLOB.STUI.processing |= STUI_LOG_ADMIN
 
+/proc/check_slur(var/text, var/mob/source, var/method = "Speech")
+	if(!CONFIG_GET(flag/log_slurs))
```
Geevies, line 40: this is called log_slurs but it prevents checking slurs altogether. either rename it or remove it

code/__HELPERS/logging.dm
```diff
@@ -36,6 +36,22 @@
 	GLOB.STUI.admin.Add("\[[time_stamp()]]ADMIN: [text]")
 	GLOB.STUI.processing |= STUI_LOG_ADMIN
 
+/proc/check_slur(var/text, var/mob/source, var/method = "Speech")
```
Geevies, line 39: imo it'd be better to do it like check_slur(var/mob/source, var/text, var/method) as it's more natural to code in that order

code/__HELPERS/logging.dm
```diff
@@ -36,6 +36,22 @@
...
+	if(!CONFIG_GET(flag/log_slurs))
+		return FALSE
+	var/list/TriggerWords = file2list("config/banned_words.txt")
+	for(var/Trigger in TriggerWords)
+		if(findtext(text,Trigger))
+			log_slur(source, Trigger, text)
+			return TRUE //Blocks message being sent.
```
Geevies, line 46: return the full text here as an error message but with the slur word underlined so the user understands what exactly was blocked. that way if there's a false positive it can be reported, and also silent failing isn't good UX.

code/modules/mob/living/say.dm
```diff
@@ -99,6 +99,9 @@ var/list/department_radio_keys = list(
 
 	message = process_chat_markup(message, list("~", "_"))
 
+	if(check_slur(message, usr, "Say"))
```
Geevies, line 102: use src isntead of usr here, and also make sure that it interacts correctly with forcesay, so the person being forced to say something isn't marked as the culprit

Geevies (changes requested):
this should be applied to other user inputs as well, such as character names, flavour text, labellers, and APC names. lastly, i think it'd be good for transparency if the list is published somewhere public.

code/__HELPERS/logging.dm
realforest2001: will probably rename it to block_slurs when I make next changes

PhantomEpicness:
> There is no such thing as an effective and non-intrusive chat censor, this will either be entirely overzealous or completely pointless.
> 
> This implementation immediately seems to lean towards being overzealous, especially since you included a chat message sent to all staff online so they can seemingly punish people for something that the game already blocked from being posted.
> 
> The only impact this will have on the game is a negative one.

True, look at roblox.

They have an extremely aggressive chat filter, yet people still get away with saying the worst of slurs on their platform.

silencer-pl:
Just to clarify on the policy level: The intent behind an automatic slur blocker is so that people are just told "no" when using a word that we do not want to see used on the server and no further punishment is enacted on them. Only cases of bypassing the filter will be "additionally" sanctioned and that's not even the right word here really.

Anyway yes, if you get told no and your message is blocked by the filter, no further staff action will be taken against you.

Carnage068:
damn that probably would have saved me a few months of being banned on top of a perma, funny how this is only implemented after im gone smh!!!

carlarctg:
why isnt carnage git banned for the love of god he's ingame banned and discord banned all he does now is post incoherent comments on prs as someone who was literally perma banned for a thousand instances of toxicity played only one caste ever and has not played for several months his opinions are utterly meaningless

code/__HELPERS/logging.dm
realforest2001: Regex I don't know how to integrate into DM.
File renamed.
As for applying to other inputs I'll do so later when I've time :)

stanalbatross:
@Carnage068 you have been blocked from the github due to repeated abusive/irrelevant/insulting/useless comments (dating from far before this PR) and being ingame + discord permabanned for harassment etc.

At such a time that you are unbanned from the game and discord your unblocking may be considered.

code/__HELPERS/logging.dm
harryob: @realforest2001 https://www.byond.com/docs/ref/#/proc/regex

code/__HELPERS/logging.dm
```diff
@@ -36,6 +36,24 @@
 	GLOB.STUI.admin.Add("\[[time_stamp()]]ADMIN: [text]")
 	GLOB.STUI.processing |= STUI_LOG_ADMIN
 
+
+GLOBAL_LIST_INIT(SlurWords, file2list("config/slur_words.txt"))
```
Watermelon914, line 40: use the configuration datum instead.

Watermelon914 (changes requested):

code/__HELPERS/logging.dm
Watermelon914: please use regex for this instead

Watermelon914: If you need an example of how to do this, please check https://github.com/tgstation/tgstation/blob/6ce946a0566cb6c19bffb588c331107ef54e04ee/code/controllers/configuration/configuration.dm#L448

PR #1257, opened 2022-10-15
Prevents floodlight from being wrenchable

Youbar:
## About The Pull Request

Floodlights can currently be used to form an impenetrable barricade for Xenos (apart from melting them via boiler/spitter - given that they don't block bullets, a machine gun emplacement can make doing so needlessly difficult at particular chokes). No other object in the game is capable of doing so with such a simple setup, and so floodlights should either be unmovable, or movable but slashable. I am happy to accomodate either in this PR.

Another alternative is that the admin team revises their ruling about floodlight barricades not being an exploit.

image

## Why It's Good For The Game

Stops exploits such as

image

## Changelog

:cl: Youbar
del: Floodlights can no longer be unanchored
/:cl:

### DIFF
--- a/code/game/machinery/floodlight.dm
+++ b/code/game/machinery/floodlight.dm
@@ -77,14 +77,6 @@
 	if(!ishuman(user))
 		return
 
-	if (HAS_TRAIT(W, TRAIT_TOOL_WRENCH))
-		if (!anchored)
-			anchored = 1
-			to_chat(user, "You anchor the [src] in place.")
-		else
-			anchored = 0
-			to_chat(user, "You remove the bolts from the [src].")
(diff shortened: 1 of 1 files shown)


Rhynostar:
Cope

Wintermote:
It's an issue but this is entirely the wrong way to 'fix' it.

There are floodlights all over the place that are meant to be moved and are placed specifically so they can be used and moved.

Youbar:
I'll do a rewrite so that non-luminous floodlights don't BB then. Never realised these things dropped in crates.

PhantomEpicness:
this is stupid
make them destructable

PR #1259, opened 2022-10-15
Aghost via Ghost Verb

Geevies:
## About The Pull Request

If you have mod level perms you can now Aghost as an option via the Ghost verb. Also the ghost verb is now a TGUI alert thing.

## Why It's Good For The Game

Q O L

## Changelog

:cl:
admin: The ghost verb alert component now allows Aghosting as well as regular ghosting.
/:cl:

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -351,7 +351,13 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 			mind.player_entity.update_panel_data(round_statistics)
 		ghostize(TRUE)
 	else
-		var/response = alert(src, "Are you -sure- you want to ghost?\n(You are alive. If you ghost, you won't be able to return to your body. You can't change your mind so choose wisely!)","Are you sure you want to ghost?","Ghost","Stay in body")
+		var/list/options = list("Ghost", "Stay in body")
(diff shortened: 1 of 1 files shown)


code/modules/mob/dead/observer/observer.dm
```diff
@@ -351,7 +351,13 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 			mind.player_entity.update_panel_data(round_statistics)
 		ghostize(TRUE)
 	else
-		var/response = alert(src, "Are you -sure- you want to ghost?\n(You are alive. If you ghost, you won't be able to return to your body. You can't change your mind so choose wisely!)","Are you sure you want to ghost?","Ghost","Stay in body")
+		var/list/options = list("Ghost", "Stay in body")
+		if(check_rights(R_MOD))
+			options = list("Aghost") + options
```
stanalbatross, line 356: grrrrr strings being used instead of defines grrr

stanalbatross (changes requested):

code/modules/mob/dead/observer/observer.dm
Geevies: using defines here doesn't really make sense, as it's only being used in one place

fira (approved):

PR #1273, opened 2022-10-16
big runtime fix batch 1

stanalbatross:
## About The Pull Request

fixes ~10 runtimes

https://cdn.discordapp.com/attachments/992551201558253778/1031308274538139688/unknown.png

as seen in this image

## Why It's Good For The Game

runtimes r bad

## Changelog

:cl:
fix: fixed some runtimes that were causing fancy storage items' examine text to break, among other things.
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -185,7 +185,8 @@ directive is properly returned.
 
 /atom/proc/examine(mob/user)
 	var/list/examine_strings = get_examine_text(user)
-	to_chat(user, examine_block(examine_strings.Join("\n")))
+	if(examine_strings)
+		to_chat(user, examine_block(examine_strings.Join("\n")))
 
 /atom/proc/get_examine_text(mob/user)
 	. = list()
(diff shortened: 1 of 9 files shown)


code/game/atoms.dm
```diff
@@ -185,7 +185,8 @@ directive is properly returned.
 
 /atom/proc/examine(mob/user)
 	var/list/examine_strings = get_examine_text(user)
-	to_chat(user, examine_block(examine_strings.Join("\n")))
+	if(examine_strings)
```
Geevies, line 188: there won't ever be a case something calls parent but doesn't run examine strings, so not doing an if check so it runtimes so we can see that there's something wrong is good here, so don't put an if statement here

Geevies (changes requested):

code/game/atoms.dm
stanalbatross: done with log debug because runtimes bad grrrrr

morrowwolf (changes requested):
We kind of want it to runtime so we can track it. I think I agree with Fira that putting an active runtime behind a vague debug log isn't what we want.

PR #1278, opened 2022-10-17
Fixes Ledges Dropping Bodies

Misttyy:
## About The Pull Request

Changes it so you can fireman carry and pull bodies over ledges, windows and etc.

## Why It's Good For The Game

This beforehand was an unintentional bug wherein bodies were dropped going over windows due to how the code was written. This fixes that and removes a bit of jank from the game. Code done by Geeves originally, updated by Misty.

## Changelog
:cl: Geevies, Misty
add: Makes it so you can fireman carry/drag bodies over ledges.
fix: Refactored code in living mobs and structures to fix a bug wherein people couldn't carry over ledges or windows.
:cl:

### DIFF
--- a/code/game/objects/structures.dm
+++ b/code/game/objects/structures.dm
@@ -148,9 +148,9 @@
 	user.forceMove(TT)
 
 	var/list/grabbed_things = list()
-	for(var/obj/item/grab/G in list(user.l_hand, user.r_hand))
-		grabbed_things += G.grabbed_thing
-		G.grabbed_thing.forceMove(user.loc)
+	for(var/obj/item/grab/grab in list(user.l_hand, user.r_hand))
+		grabbed_things += grab.grabbed_thing
+		grab.grabbed_thing.forceMove(user.loc)
 	user.forceMove(TT)
 	for(var/atom/movable/thing as anything in grabbed_things) // grabbed things aren't moved to the tile immediately to: make the animation better, preserve the grab
 		thing.forceMove(TT)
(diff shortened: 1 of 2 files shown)


Misttyy:
Will add missing line tomorrow when I'm home from Seattle.

code/game/objects/structures.dm
```diff
@@ -147,6 +147,14 @@
 
 	user.forceMove(TT)
 
+	var/list/grabbed_things = list()
+	for(var/obj/item/grab/G in list(user.l_hand, user.r_hand))
```
stanalbatross, line 151: varname

stanalbatross (changes requested):

fira (approved):

code/game/objects/structures.dm
```diff
@@ -147,6 +147,14 @@
 
 	user.forceMove(TT)
 
+	var/list/grabbed_things = list()
+	for(var/obj/item/grab/grab in list(user.l_hand, user.r_hand))
+		grabbed_things += grab.grabbed_thing
+		grab.grabbed_thing.forceMove(user.loc)
+	user.forceMove(TT)
```
morrowwolf, line 154: I may be missing something but why is this forceMove necessary? It seems like it's doubled with the one above.

morrowwolf (changes requested):

Misttyy:
All fixed, tested on local and works.

ThePiachu (changes requested):
This works inconsistently:
- If you are crossing a ledge with a body, sometimes it drags the body onto your tile, sometimes it leaves be body behind. Maybe it's tied to directions or something?
- Dragging a body with you over a window frame doesn't drag the body to the window frame
- When you successfully drag a body over the ledge it moves 2 spaces, it should probably move only 1 space and with the next move you should be able to drag it over the ledge
- If you start moving over the ledge by holding the button down you will get moved 2 spaces which is jarring and probably not an intended behaviour

Issue #106, opened 2022-04-20
Predator Skinning Synths


FrogZeke:
## Description
Predator skinning does not have a issynth check.  This allows predators to skin synthetics, producing meat and a skin.

## Difference between expected and actual behavior
It has been stated by the Pred Council that synthetics are nothing more than semi-sentient forklifts.  The purpose of skinning is taking meat and a trophy.  Preds cannot skin the APC.  The actual behavior is pred removing synth fillets and creating a skin from a dead synth.

## Steps to reproduce
Use pred tools to butcher a dead synth.

## Issue bingo (not required to be all checked)
- [ ] Issue could be reproduced at least once
- [ ] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Diegoflores31:
what if you could skin tham and get Synth meat instead ?

FrogZeke:
Synthmeat is part of SS13, a combination of blood and coagulants.  In Alienverse lore, food was recycled on ships, so it would be possible to exist.  The issue is that synthetics themselves are not of made of synthmeat or blood.  Skinning a synthetic would not actually kill them, outside of many deep cuts that drained the white cooling fluid.  Aside from the technical aspects of synth construction, Yautja do not skin sentry turrets for meat or trophies. The Yautja Council has determined that synthetics are the equivalent of semi-sentient forklifts, equipment to be ignored or simply destroyed if it interferes with the hunt.  It is not something to take as a trophy regardless.

FrogZeke:
![image](https://user-images.githubusercontent.com/22293891/165671966-25b0392a-6962-418e-a67b-d179db17e6d7.png)
To quote the Synth Senator as this is a WL change.

FrogZeke:
> Seems to be working fine. Have you seen a synth being flayed in game?

Yes, Nathan (Discord name: InsaneRed) was the one flayed, you can confirm with him.  We have multiple species of synthetics.  In the case I saw he was a synthvivor.

Issue #61, opened 2022-04-16
you can dupe smart-disks


stanalbatross:
## Reproduction:

place an active smartdisk on a table, it will duplicate it.

PR #1287, opened 2022-10-18
Pred bug fix no.2

Joelampost:
## About The Pull Request
Resolves: https://github.com/cmss13-devs/cmss13/issues/106
Resolves: https://github.com/cmss13-devs/cmss13/issues/61

## Why It's Good For The Game
Because fix

## Changelog

:cl: Joe lampost
fix: Stops duplication of smart discs 
fix:Stops the skinning of synths 
/:cl:


harryob (changes requested):

code/game/objects/structures/tables_racks.dm
```diff
@@ -305,6 +305,12 @@
 			to_chat(user, SPAN_WARNING("You slice at the table, but only claw it up a little."))
 		return
 
+	if(istype(W, /obj/item/explosive/grenade))
+		var/obj/item/explosive/grenade/I = W
```
Watermelon914, line 309: this variable name is not detailed whatsoever

Diegoflores31:
i think it would be better to rather say that the synthethic is not worth it or something .. i really doubt a pred cant cut thru the synth skin .

Joelampost:
> i think it would be better to rather say that the synthethic is not worth it or something .. i really doubt a pred cant cut thru the synth skin .

tools break

code/game/objects/structures/tables_racks.dm
```diff
@@ -305,6 +305,12 @@
 			to_chat(user, SPAN_WARNING("You slice at the table, but only claw it up a little."))
 		return
 
+	if(istype(W, /obj/item/explosive/grenade))
+		var/obj/item/explosive/grenade/detonating_grenade = W
+		if(detonating_grenade.active)
+			to_chat(user, SPAN_WARNING("It's too late for that!"))
+		return
```
stanalbatross, line 312:
```suggestion
			return
```
needs an extra indent here or it will ALWAYS return

code/modules/cm_preds/yaut_procs.dm
```diff
@@ -72,14 +72,18 @@
...
+		
 	if(isXenoLarva(T) || isXenoFacehugger(T))
 		to_chat(src, SPAN_WARNING("This tiny worm is not even worth using your tools on."))
 		return
 
 	if(is_mob_incapacitated() || lying || buckled)
 		return
 
+	if(isSynth(T))
+		to_chat(src, SPAN_WARNING("You would break your tools if you did this."))
```
stanalbatross, line 84:
```suggestion
		to_chat(src, SPAN_WARNING("You would break your tools if you did this!"))
```

stanalbatross (changes requested):

stanalbatross (approved):

Issue #1286, opened 2022-10-18
Health Scanner - Medication Advice is not updating correctly, and advises to OD patients.


blackdragonTOW:
## Description
Health Scanners will tell users to administer drugs that would OD the patient.

## Difference between expected and actual behavior
Medication Advice does not display "Administer a single dose of X" when a dose of X would put the patient over the OD limit of said drug
![Capture](https://user-images.githubusercontent.com/31581761/196359939-7180ac58-9a58-41cb-b0e6-020b8727b81d.PNG)

## Steps to reproduce
1. Wound yourself with around 80 Brute
2. Slam 30u of Bicaridine
3. Wait a few moments to metabolize the Bica and start reduction of Brute damage
4. Scan yourself with a Health Analyzer 
5. ---> The Medication Advice will recommend that you give yourself more bica even when in the high 20s already

## Issue bingo (not required to be all checked)
- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #1288, opened 2022-10-19
Fixing medical advise

aceroyuno:
## About The Pull Request
Fixes and closes https://github.com/cmss13-devs/cmss13/issues/1286

## Why It's Good For The Game
Less ODs

## Changelog

:cl: Aceroy
fix: Health Analyzers now accurately recommend when to administer dosages
/:cl:

### DIFF
--- a/code/modules/mob/living/living_healthscan.dm
+++ b/code/modules/mob/living/living_healthscan.dm
@@ -372,8 +372,8 @@ GLOBAL_LIST_INIT(known_implants, subtypesof(/obj/item/implant))
 						"icon" = "syringe",
 						"colour" = "purple"
 						))
-					if(!chemical_lists["inaprovaline"])
-						if(chemicals_lists["inaprovaline"]["amount"] < "don't know what to put here")
+					if(!chemicals_lists["inaprovaline"])
+						if(chemicals_lists["inaprovaline"]["amount"] < 30)
 							advice += temp_advice
 					else
 						advice += temp_advice
@@ -388,8 +388,8 @@ GLOBAL_LIST_INIT(known_implants, subtypesof(/obj/item/implant))
 						"icon" = "syringe",
 						"colour" = "white"
 						))
-					if(!checmicals_lists["tramadol"])
-						if(chemicals_lists["tramadol"]["amount"] < 30)
+					if(!chemicals_lists["tramadol"])
+						if(chemicals_lists["tramadol"]["amount"] < 15)
 							advice += temp_advice
 					else
 						advice += temp_advice
--- a/code/modules/mob/living/living_healthscan.dm
+++ b/code/modules/mob/living/living_healthscan.dm
@@ -372,7 +372,7 @@ GLOBAL_LIST_INIT(known_implants, subtypesof(/obj/item/implant))
 						"icon" = "syringe",
 						"colour" = "purple"
 						))
-					if(!chemicals_lists["inaprovaline"])
+					if(chemicals_lists["inaprovaline"])
 						if(chemicals_lists["inaprovaline"]["amount"] < 30)
 							advice += temp_advice
 					else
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/living_healthscan.dm
```diff
@@ -299,67 +300,106 @@ GLOBAL_LIST_INIT(known_implants, subtypesof(/obj/item/implant))
...
+					else
+						advice += temp_advice
+				if(human_target_mob.health < 0)
+					temp_advice = list(list(
 						"advice" = "Administer a single dose of inaprovaline.",
 						"icon" = "syringe",
 						"colour" = "purple"
 						))
-
+					if(!chemicals_lists["inaprovaline"])
```
stanalbatross, line 375:
```suggestion
					if(chemicals_lists["inaprovaline"])
```

code/modules/mob/living/living_healthscan.dm
```diff
@@ -299,67 +300,106 @@ GLOBAL_LIST_INIT(known_implants, subtypesof(/obj/item/implant))
...
-				if(has_pain && chemicals_lists["tramadol"] < 3 && !chemicals_lists["paracetamol"])
-					advice += list(list(
+				if(has_pain && !chemicals_lists["paracetamol"])
+					temp_advice = list(list(
 						"advice" = "Administer a single dose of tramadol.",
 						"icon" = "syringe",
 						"colour" = "white"
 						))
+					if(!chemicals_lists["tramadol"])
+						if(chemicals_lists["tramadol"]["amount"] < 15)
```
stanalbatross, line 392:
```suggestion
						if(chemicals_lists["tramadol"]["amount"] < 3)
```
will prevent asking for more tram with literally every single scan + less ODs

code/modules/mob/living/living_healthscan.dm
```diff
@@ -620,6 +660,8 @@ GLOBAL_LIST_INIT(known_implants, subtypesof(/obj/item/implant))
 				advice += "<span class='scanner'>Administer a single dose of kelotane.</span>\n"
 			if(H.getBruteLoss(1) > 30 && reagents_in_body["bicaridine"] < 3)
 				advice += "<span class='scanner'>Administer a single dose of bicaridine.</span>\n"
+			if(H.getBruteLoss(1) > 30 && reagents_in_body["bicaridine"] > 3)
+				advice += "<span class='scanner'>Do not Administer a single dose of bicaridine.</span>\n"
```
stanalbatross, line 664: ?

stanalbatross (changes requested):

code/modules/mob/living/living_healthscan.dm
aceroyuno: this is left over code from the nano ui that I was confused about why it was being called. It can be removed once the TGUI port has been finished.

code/modules/mob/living/living_healthscan.dm
```diff
@@ -620,6 +660,8 @@ GLOBAL_LIST_INIT(known_implants, subtypesof(/obj/item/implant))
 				advice += "<span class='scanner'>Administer a single dose of kelotane.</span>\n"
 			if(H.getBruteLoss(1) > 30 && reagents_in_body["bicaridine"] < 3)
 				advice += "<span class='scanner'>Administer a single dose of bicaridine.</span>\n"
+			if(H.getBruteLoss(1) > 30 && reagents_in_body["bicaridine"] > 3)
+				advice += "<span class='scanner'>Administer a single dose of bicaridine.</span>\n"
```
stanalbatross, line 664: you have not fixed this.

stanalbatross (changes requested):

code/modules/mob/living/living_healthscan.dm
aceroyuno: duplicate line removed

code/modules/mob/living/living_healthscan.dm
```diff
@@ -299,67 +300,106 @@ GLOBAL_LIST_INIT(known_implants, subtypesof(/obj/item/implant))
...
+						advice += temp_advice
+				if(human_target_mob.health < 0)
+					temp_advice = list(list(
 						"advice" = "Administer a single dose of inaprovaline.",
 						"icon" = "syringe",
 						"colour" = "purple"
 						))
-
+					if(chemicals_lists["inaprovaline"])
+						if(chemicals_lists["inaprovaline"]["amount"] < 5)
```
harryob, line 376: can't you use ['od'] instead of hardcoding the od amounts? or use the same handling that the ['od'] variable takes for dynamically getting the OD amount?

aceroyuno: so, ["od"] is not what we are checking for, [5] or less is the amount that prompts recommending another dosage. _Idealy_ we could find a way to dynamically test for this using some sort of for loop setup. But it would take a complete refactor to do so.

However, to solve what you are asking for. we could have a ["recommend_dose"] variable that is gotten from the chemical. But so much of this is already hardcoded, it hardly makes a difference.

stanalbatross (approved):

PR #1292, opened 2022-10-19
Fixed resin spikes deleting on evolution

carlarctg:
## About The Pull Request

Title

Instead of deleting built_structures on the old xeno, evolution now simply transfers over the list to the new xeno and deletes the old list. This means the list carries over 

additionally tail stab now gives a message if it misses because yorue aiming for a missing limb. the tail has a spike so its fine and nots cope creep

## Why It's Good For The Game

its a bugfix

## Changelog

:cl:
fix: Fixed resin spikes deleting on evolution, allowing drones who evolve into hivelord (or vice versa) to keep them.
fix: Tail stab now has a msg if you aim for a nonexistent limb
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Evolution.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Evolution.dm
@@ -141,10 +141,6 @@
 
 	new_xeno.built_structures = built_structures.Copy()
 
-	for(var/i in built_structures)
-		var/list/L = built_structures[i]
-		QDEL_NULL_LIST(L)
-
 	built_structures = null
 
 	new_xeno.visible_message(SPAN_XENODANGER("A [new_xeno.caste.caste_type] emerges from the husk of \the [src]."), \
@@ -304,10 +300,6 @@
 	var/mob/living/carbon/Xenomorph/new_xeno = new xeno_type(get_turf(src), src)
 
 	new_xeno.built_structures = built_structures.Copy()
-
-	for(var/i in built_structures)
-		var/list/L = built_structures[i]
-		QDEL_NULL_LIST(L)
 
 	built_structures = null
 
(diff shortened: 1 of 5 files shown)


carlarctg:
wha!!!

ihatethisengine:
But what if let's say hivelord builds available reflective walls and acid pillars and then de-evolves?

carlarctg:
then they wasted a t2 slot and 500 evo points to make two situational buildings?

code/modules/mob/living/carbon/xenomorph/Evolution.dm
```diff
@@ -139,12 +139,15 @@
 	else
 		new_xeno.plasma_stored = new_xeno.plasma_max*(plasma_stored/plasma_max) //preserve the ratio of plasma
 
+	for(var/i in src.built_structures)
+		new_xeno.built_structures += built_structures[i]
```
Geevies, line 143: can't you just do `built_structures.Copy()`? also QDEL_NULL_LIST it once you're done

code/modules/mob/living/carbon/xenomorph/Evolution.dm
```diff
@@ -139,12 +139,21 @@
...
```
Geevies, line 143: when i said 'can't you just do' i meant remove this code and do the copy instead

code/modules/mob/living/carbon/xenomorph/Evolution.dm
```diff
@@ -139,12 +139,21 @@
...
+
+	new_xeno.built_structures = built_structures.Copy()
+
+	for(var/i in built_structures)
+		var/list/L = built_structures[i]
+		QDEL_NULL_LIST(L)
```
Geevies, line 149: are you sure that this is right, because in theory this deletes the spikes. did you test it?

code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
```diff
@@ -672,12 +672,6 @@
 	QDEL_NULL(mutators)
 	QDEL_NULL(behavior_delegate)
 
-	for(var/i in built_structures)
-		var/list/L = built_structures[i]
-		QDEL_NULL_LIST(L)
-
-	built_structures = null
```
Geevies, line 679: it should still null out

Geevies (changes requested):

PR #1293, opened 2022-10-19
Fixes flashbangs permadeafening 

memesky117:
## About The Pull Request

Flashbangs no longer cause ear damage, stopping marines from getting permanently deafened when there's no solution to it past ahealing.

## Why It's Good For The Game

Flashbangs only purpose is to stun crowds of marines and give MPs time to escape/take people in. Ear damage only causes round permanent damage to marines and can only be healed by aheals, this stops that.

## Changelog

:cl: memesky

fix: Flashbangs no longer cause ear damage.

/:cl:


Geevies (changes requested):
this isn't a fix

memesky117:
I'll look into fixing the root cause (I am NOT losing GBP for this)

PR #1295, opened 2022-10-19
adds DMAPI v6.0.5 for tgs support, adds a verb to check current server revision

harryob:
## About The Pull Request

this adds DMAPI support to the codebase, allowing deeper integration with TGS if cm ever migrated to using it. it also adds a verb to check the current revision data (courtesy of tg) to help with bug reports, but this can be removed if preferable.

## Why It's Good For The Game

modernised and maintained hosting infrastructure might be nice to have the option to use. this might be all horribly incompatible with our existing infrastructure (somehow) so do let me know

## Changelog

:cl:
server: added (deeper) support for tgs with DMAPI support, alongside a verb to check the current server revision
/:cl:


fira (approved):
Looks good to me

fira:
The verb won't serve much purpose so long this is running without TGS, but I don't think it's a big concern

PR #1298, opened 2022-10-20
Fixes TOUCH effect of Toxic trait doing nothing

Diegoflores31:
## About The Pull Request

Research warcrimes did an oppsie and Toxin damage was only being applied if you were wearing a mask ( yeah the one that SHOULD protect you btw)

## Why It's Good For The Game

bug bad .
PROOF that it was intended the way i am fixin : 
image

## Changelog

:cl:
fix: Toxic trait now applies Toxin damage on touch as intended
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry_properties/prop_negative.dm
+++ b/code/modules/reagents/chemistry_properties/prop_negative.dm
@@ -74,8 +74,8 @@
 		return
 	var/mob/living/carbon/C = M
 	if(!C.wear_mask) // If not wearing a mask
+		C.apply_damage(potency, TOX)  // applies potency toxin damage
 		return
-	C.apply_damage(potency, TOX) // applies potency toxin damage
 
 /datum/chem_property/negative/corrosive
 	name = PROPERTY_CORROSIVE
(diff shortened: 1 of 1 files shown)


code/modules/reagents/chemistry_properties/prop_negative.dm
```diff
@@ -74,8 +74,8 @@
 		return
 	var/mob/living/carbon/C = M
 	if(!C.wear_mask) // If not wearing a mask
+		C.apply_damage(potency, TOX)  // applies potency toxin damage
 		return
```
Geevies, line 78:
```suggestion
	if(C.wear_mask) // Wearing a mask
		return
	C.apply_damage(potency, TOX)  // applies potency toxin damage
```

Geevies (changes requested):

code/modules/reagents/chemistry_properties/prop_negative.dm
Diegoflores31: Why?

Geevies (approved):

stanalbatross (approved):

PR #1312, opened 2022-10-23
Xenos Build Animations

Warfan1815:
## About The Pull Request

Adds weak & thick resin structure construction animations for:

-Transparent Membrane
-Standard Walls
-Doors
-Reflective Walls (these are not special but instead just use the standard walls in stead of the reflective wall for the animation, this is mainly because it is a rare thing to be constructed and shouldn't cause too much confusion.)

Also there are different framerate icon_states for all 3 current "speeds" of construction in the game for all of the constructions animations. I wanted this to be handled by code, not just repeating the same sprite 3 times with different delays, but DM has forced my hand.

The "purple splotches" remain, and the animation *should* be interruptible once the weeds are destroyed, as it uses the same code for destruction the purple splotches do.

For your viewing pleasure:

https://cdn.discordapp.com/attachments/746346213443633232/1033547599757705266/2022-10-23_02-04-21_Trim.mp4
https://cdn.discordapp.com/attachments/746346213443633232/1033553802630004777/2022-10-23_02-10-50_Trim.mp4
(or if you don't want to download them I posted them in the coder-channel)

## Why It's Good For The Game

It lends itself to giving satisfaction to building things as a builder, and stops walls from being fathomed out of thin air right infront of marines, Instead marines will be fully aware of the imminent threat of a wall being constructed and will be more aware of their situation. Also it helps somewhat with immersion.

## Changelog

:cl:
add: Adds build animations for some of the resin structures (thick & weak versions of: Walls, Doors, Transparent Membrane & Reflective Walls)
imageadd: Added the construction animations icons for some of the resin structures 
/:cl:


code/modules/mob/living/carbon/xenomorph/Powers.dm
```diff
@@ -99,6 +99,19 @@
...
+	if(RC.build_animation_effect)
+		warning = new RC.build_animation_effect(current_turf)
+
+		if(wait_time != 2 SECONDS)
+
+			if(wait_time == 1 SECONDS)
+				warning.icon_state = "[warning.icon_state]Fast"
+
+			if(wait_time == 4 SECONDS)
+				warning.icon_state = "[warning.icon_state]Slow"
```
stanalbatross, line 111: switch statement for wait_time instead of if chain

stanalbatross (changes requested):

code/modules/mob/living/carbon/xenomorph/Powers.dm
Warfan1815: Done so.

stanalbatross (approved):
code is fine

code/modules/cm_aliens/weeds.dm
```diff
@@ -566,4 +566,23 @@
...
+	icon_state = "ThickConstruct"
+
+/obj/effect/resin_construct/weak
+	icon_state = "WeakConstruct"
+
+/obj/effect/resin_construct/transparent/thick
+	icon_state = "ThickTransparentConstruct"
+
+/obj/effect/resin_construct/transparent/weak
+	icon_state = "WeakTransparentConstruct"
```
stanalbatross, line 586: resolve this once you fix the naming issue

stanalbatross (changes requested):

Agameofscones (approved):
Hyup, guess that'll do

PR #1323, opened 2022-10-24
Rebalancing for berserkers buffs. [NO GPB UPDATE]

usnpeepoo:
## About The Pull Request

Turns out that having a lower cooldown on apprehend with a duration buff made the cooldown lower than what it really was, this PR adjusts this. Also, I've gave it an extremely slight increase in speed.
Also, the eviscerate life steal was too powerful, so it was lowered.

## Why It's Good For The Game

Berserker should be good, but not better than base or hedgehog ravager, this PR adjusts the recent buffs that were merged.

## Changelog

:cl:
balance: Berserkers apprehend now has an actual cooldown, its speed increased very slightly.
balance: Berserkers eviscerates life steal gain and max has been lowered
/:cl:


Shad0vvs (approved):
code good

PR #1325, opened 2022-10-25
fixes building shit on doors

stanalbatross:
## About The Pull Request

xenos may no longer build constructions (other than sticky and fast resin) on tiles with doors on them, be they open or not

this is because the door blocks the construction

should not be a major issue as xenos can slash doors to break them

this solves issues with trap placement on doors that then close, etc

## Why It's Good For The Game

consistency fix

## Changelog

:cl:
fix: xenos can no longer build most structures on the same tile as open doors
/:cl:


harryob (approved):
looks like a rather neat and tidy solution to me

PR #1342, opened 2022-10-27
CAS Refactor Part 2 - Composable Weapon Effects

fira:
## About The Pull Request

*This requires extra testing*

Part of forever-WIP CAS backend refactors. The end goal is to provide ways to have more interesting interactions/weapons, with mechanics such as roof impact, interception, dynamic or moving targets, stateful weapons (charge up etc), and so on...

Part 1: Dynamic Gimbals and Bugfixing
Part 2: Composable Weapon Effects                                                     <---- We are here
Part 3: Weapon/Ammo handling refactor
Part 4: Firing safeties and weapon firing refactor
Part 5: Dynamic ammo usage rework
Part 6: Weapon Concept implements

## Why It's Good For The Game

Making the whole of the CAS firing sequence work by composition allows to easily make variants (firing modes, unusual weapons, etc) of our current process while still leveraging old code. It also makes it modular.

This also paves way for the following features, altough unusued yet: 
 * Signal locked weapon firing: use the signal target rather than ground target when possible. This may allow moving laser targets integration in the future.
 * Roof impacting: a live warhead will be blocked by roof at impact time, not just avoided by the firing block. This builds way for future roof-piercing or roof-damaging weaponry.
 * Generalized warning effects: allow us to have more effects than the green dot. 
 * Interception: Warhead launch includes handling for Terminal Misses (eg. Roof), or In-Flight Interception, allowing to add appropriate effects if needed

## Changelog

:cl:
refactor: Refactored most of CAS Weaponry effect code to pave way for future features
/:cl:


fira:
Oversight to think about: "roof miss" currently removes the fired warhead. It's "normally" not a big issue because you can't fire at roofs anyway currently (that's environemental safety in Part 4), but would be confusing for players if you fire next to roof, it deviates onto it, and no boom arrives.

Should probably disable it for standard ammunition / for now

PR #1344, opened 2022-10-27
adds new round alert to tgs, removes some legacy from the old box

### LINKED ISSUE #87: Auto reconnect does not work
## Reproduction:

Auto reconnection to the server upon restart, despite what the TG Chat prompt says, does not work - unless the user manually picks reconnect from either the client itself or via closing the window and using the BYOND Hub, they will not reconnect to the server on restart.

harryob:
## About The Pull Request

this removes the flag that prevents the server from restarting and forces a shutdown
this adds new round alert functionality via TGS instead, which requires two config strings being set:
1) new_round_alert_channel for the TGS channel tag it should send new round alerts too
2) new_round_alert_role_id for the role it should ping with the new round alerts

also closes #87 

## Why It's Good For The Game

new round alert broken, this fixes
also allows more logging to be sent to discord via send2adminchat

## Changelog

:cl:
server: new round alert lives again, via tgs
fix: autoreconnect now works
/:cl:

### DIFF
--- a/ColonialMarinesALPHA.dme
+++ b/ColonialMarinesALPHA.dme
@@ -112,6 +112,7 @@
 #include "code\__HELPERS\#maths.dm"
 #include "code\__HELPERS\_lists.dm"
 #include "code\__HELPERS\_time.dm"
+#include "code\__HELPERS\chat.dm"
 #include "code\__HELPERS\cmp.dm"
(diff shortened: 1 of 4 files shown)


fira (approved):
awww i liked the touch of having the operation name 

edit: i'm blind

fira (approved):

PR #1345, opened 2022-10-27
Fix TGS rustg Cross-Compile Dependencies Error

fira:
## About The Pull Request

Some setups seem to fail to link rustg to zlib because they're unable to find the library.

Why exactly using the newer multiarch packages instead of legacy lib32 one makes a difference is unclear, probably residing in pkg-config or rustg crosscompile quirks...

How the live server booted up just fine the first time despite this is also anyone's guess...


harryob (approved):
i can verify that somehow due to whatever rustg weirdness, this does fix the script

PR #1349, opened 2022-10-27
Fixes exploit that allowed infinite facehuggers from an eggmorpher

morrowwolf:
## About The Pull Request

Rechecks all checks post-confirmation to stop alert stacking exploits.

## Why It's Good For The Game

Bug bad.

## Changelog

:cl: Morrow
fix: Fixes exploit that allowed infinite facehuggers from an eggmorpher
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/special/egg_morpher.dm
+++ b/code/modules/cm_aliens/structures/special/egg_morpher.dm
@@ -227,7 +227,7 @@
 				marine_count++
 		playable_hugger_limit = round(marine_count / 5)
 
-	var/current_hugger_count = 0
+	current_hugger_count = 0
 	for(var/mob/mob as anything in GLOB.living_xeno_list)
 		if(isXenoFacehugger(mob))
 			current_hugger_count++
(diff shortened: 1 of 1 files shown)


code/modules/cm_aliens/structures/special/egg_morpher.dm
```diff
@@ -183,6 +183,9 @@
 	..()
 
 /obj/effect/alien/resin/special/eggmorph/attack_ghost(mob/dead/observer/user)
+	if(alert(user, "Are you sure you want to become a facehugger?", "Confirmation", "Yes", "No") == "No")
```
Geevies, line 186: imo, it'd be better to run the check before and after the alert, otherwise it'd be annoying

Geevies (changes requested):

code/modules/cm_aliens/structures/special/egg_morpher.dm
```diff
@@ -212,6 +212,29 @@
 	if(alert(user, "Are you sure you want to become a facehugger?", "Confirmation", "Yes", "No") == "No")
 		return
 
+	if(world.time - user.timeofdeath < 3 MINUTES)
```
Geevies, line 215: could you possibly isolate this into a proc so it isn't duplicated?

Geevies (changes requested):

PR #1357, opened 2022-10-28
Removes ceramic plates from new contractors

morrowwolf:
## About The Pull Request

Removes ceramic plates from new contractors. It's too much.

## Why It's Good For The Game

The distress signal responses should be slightly balanced and ceramic plates are just too much. We can look at potentially buffing them in other ways later but they're already very strong.

## Changelog

:cl: Morrow
balance: Removes ceramic plates from new contractors.
/:cl:


carlarctg:
why not just need the health of their ceramic plates?

stanalbatross (approved):

PR #1358, opened 2022-10-28
Fix for ordering ammo.

Segrain:
## About The Pull Request

I am not surprised that it recently suddenly ceased working. I am confused how has had it been working _at all_ until now.
All the ASRS supply datums are subtypes of actual orderable datums, but with buyability turned off, special category that cannot be picked on the console so they are never seen and a mildly complicated system that checks for that buyability and adds them to a special list, from which controller then regularly picks random crates to send via ASRS. The issue is, they keep the same name as their immediate parents, so `supply_packs[P.name] = P` in controller's setup results in ASRS subtype overriding the actual orderable datum's entry instead of being in addition to it - thus, all the crates that have ASRS subtypes cannot be manually ordered. For some mysterious reason, it only started being an observed issue very recently instead of having being a problem for years as it at first glance should.
Also, CM dev strikes again: `"Utility"` category of crates (which is where ASRS random metal/plasteel/sandbags crates were grouped) had been deleted _two years ago_. Picking a random crate does not need to randomly choose between munitions, utility and either when all it _has_ is munitions. Setting them to `"ASRS"` group and assembling the lists of orderable and random crates separately based on that group apparently solves everything, including the stray comms parts crate occasionally showing up from ASRS (presumably because it is not buyable since static comms, and that was the only thing that mattered for ASRS to pick it).

## Why It's Good For The Game

Is important fix.

## Changelog

:cl:
fix: Requisitions once again can properly order all ammunition boxes and mortar shells.
fix: ASRS should no longer randomly deliver crate of components for communications tower.
/:cl:

### DIFF
--- a/code/datums/ASRS.dm
+++ b/code/datums/ASRS.dm
@@ -11,119 +11,115 @@
 #define ASRS_VERY_LOW_WEIGHT	50
 #define ASRS_LOWEST_WEIGHT		100
 
-/datum/supply_packs/asrs
-	cost = ASRS_LOWEST_WEIGHT
-	buyable = 0
-
 /datum/supply_packs/gun/ammo_hpr/asrs
 	buyable = 0
-	group = "Munition"
+	group = "ASRS"
 	cost = ASRS_LOWEST_WEIGHT
 
 /datum/supply_packs/ammo_rounds_box_smg/asrs
 	buyable = 0
-	group = "Munition"
+	group = "ASRS"
 	cost = ASRS_MEDIUM_WEIGHT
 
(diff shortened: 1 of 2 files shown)


fira (approved):
Makes sense to me.

Won't this increase amount of ASRS ammo due to removal of Utility category?

Segrain:
> Won't this increase amount of ASRS ammo due to removal of Utility category?

If it does, I am calling that a fix too. `add_random_crate()` should not have 40% chance to do null because of some unused category. If somebody will want to rebalance the resulting totals, we can do so by changing that complicated formula that involves current xeno count and yields the number of crates to (attempt to) spawn.

fira (approved):

PR #1359, opened 2022-10-28
update dmapi to v6.0.6 (provides support for 515 even though nothing else does)

harryob:
## About The Pull Request

i am dependabot and i'm making an automated pr to bump the tgs version (i'm not but i should automate this)

## Why It's Good For The Game

tgs will continue to work when/if we update to 515


fira (approved):
GBP farming i say!!!!

Issue #1352, opened 2022-10-28
Roundstart CL spawning in wrong area and without proper access


khalifjordan:
## Description
I have played two rounds as roundstart CL and both times spawned in the general cryopod area and not in my office. I have also spawned with no access to CIC or Research and seem to have no ability to open stairwell shutters either.

## Difference between expected and actual behavior
No access to CIC, Research or ability to open stairwell shutters; these are all things CL's should have access too

## Steps to reproduce
Unsure how to reproduce other than setting CL roll to High?

## Issue bingo (not required to be all checked)

- [x] Issue could be reproduced at least once
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

khalifjordan:
I finally found someone cooperative enough to let me look at the ID in CIC while playing as CL - the ID has 0 access other than USCM IFF selected as on. Not sure if that helps.

hotimep:
feature

ThePiachu:
Started looking into it - for some reason Nightmare variant of CL is loading, hence the messed up spawn and access.

PR #1360, opened 2022-10-28
Makes Liaison spawn in the correct place

BeagleGaming1:
## About The Pull Request

Removed the “nightmare” section of the liaison's job.
It also removes its attached liaison nightmare equipment preset as it is unused without this.

hope I didn’t break something
This should fix: https://github.com/cmss13-devs/cmss13/issues/1352
## Changelog
:cl:
fix: Fixed CL incorrectly spawning at late join cryo room at round start
/:cl:

### DIFF
--- a/code/game/jobs/job/civilians/other/liaison.dm
+++ b/code/game/jobs/job/civilians/other/liaison.dm
@@ -18,11 +18,6 @@
 	SIGNAL_HANDLER
 	active_liaison = null
 
-/datum/job/civilian/liaison/nightmare
-	flags_startup_parameters = NO_FLAGS
-	gear_preset = /datum/equipment_preset/uscm_ship/liaison/nightmare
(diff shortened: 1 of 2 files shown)


fira:
I'm a bit confused, it's been like this for years, why does this cause this?

fira:
Looks like one of the job system "it accidentally worked" cases

Need to also remove the gear preset

fira (approved):

ThePiachu:
From what I could gather looking into this, it's due to one of those global variable maps mapping all the job / gear datums to some key that the /nightmare preset didn't rename. Not sure which though.

fira:
Yeah the job system can't make that distinction at all. It just sort of falls right for some reason using dark magic. There used to be the same kind of issues with Whiskey Outpost roles before we refactored it.

stanalbatross (approved):

PR #1369, opened 2022-10-29
Adds unlocking of ID locked items

PhantomEpicness:
revived from: https://github.com/cmss13-devs/cmss13/pull/1049
## About The Pull Request
ID locked items now unlock when swiped with their owner's ID card after a confirmation prompt.

## Why It's Good For The Game

Removes stupid situations where a spec doesnt want an item, but takes it out of the box, leading to the item being useless. This has notably screwed over many Sniper Spec spotters as their entire kit is useless because the spec touched it.

Yes I know command IDs override, but CO/XO have better things to do than to go all the way down to alpha prep to unlock a VP78 for a PFC.

 Yes I know people usually throw the box, but sometimes that doesn't happen, either by accident or indecision. Currently, it is unintuitive and dumb to be unable to do anything at all, even as the owner, to locked gear.

Now the owner (whoever it is locked to) can swipe their ID to remove the ID locking altogether. 

Edit: A concern was raised that someone could steal someone else's ID and steal their shit

Stealing IDs is already against the rules so this is a non-issue. Added a log_admin just in-case

## Changelog

:cl: TotalEpicness5
add: ID locked items (ex: items in spec box) can now be unlocked by their owner swiping their ID on it.
/:cl:


fira:
i'd argue this is more of a msg_admin_niche thing

code/game/objects/items.dm
```diff
@@ -263,6 +263,14 @@ cases. Override_icon_state should be a list.*/
...
 // I have cleaned it up a little, but it could probably use more.  -Sayu
 /obj/item/attackby(obj/item/W, mob/user)
+	if(istype(W,/obj/item/card/id))
+		if(((flags_item & MOB_LOCK_ON_EQUIP)) && (user == locked_to_mob))
+						if (alert(user, "Would you like to disable the ID lock for \the [src]? This cannot be undone!", "Confirm", "Yes", "No") == "Yes")
+				to_chat(user, SPAN_WARNING("You unlock \the [src]."))
+				src.flags_item &= ~MOB_LOCK_ON_EQUIP
+				locked_to_mob = null
+				log_admin("[user] has removed the ID lock on \the [src]. Owner of the item was [locked_to_mob] (CKEY: ([user.ckey]))")
+			return
```
harryob, line 273: indentation all wacky here, you might want to use defines instead of strings for the Yes and No 🤷

harryob (changes requested):

stanalbatross:
CI fail

carlarctg:
just remove id locking spec gear entirely there's no reason for it to exist

code/game/objects/items.dm
PhantomEpicness: weird how it messed up like that
fixed

ThePiachu:
I'd just remove the ID lock from the box. Quick glance on the code shows that the MOB_LOCK_ON_EQUIP flag also exists on armour, helmets, etc. that we might not want to unlock. Heck, my initial thought was "please don't unlock pill bottles..." which I don't think this PR does...

PR #1371, opened 2022-10-29
Nightmare Rewrite (Now with Almayer support)

fira:
## About The Pull Request

Long overdue rewrite of Nightmare to unscrew some code bits, simplify implementation, and widden possibilities!

Testmerged for a few days, still need reviewing

## Why It's Good For The Game

Less crutch and moving parts, and expanded possibilities via, finally, addition of ship maps nightmare.

## File format changes

* `pick` configuration node (picks some options among several) can now pick several with `amount`, and uses weights directly set on the child nodes
   Example:
     before: `{"type": "pick", "weight":[1,2,3], "choices": [A, B C]}`
     now: `{"type": "pick", "choices": [ {A, "weight": 1}, {B, "weight": 2}, {C, "weight": 3} ], "amount": 2}`

* Main file should now be called `nightmare.json` instead of `mapgen.json`, and the map JSON file uses `nightmare_path` to point to the directory

* All file paths in config are now relative, so much shorter. Map insert nodes use the main map directory, Nightmare includes use the nightmare path.

You can look at the included files as example.

### Scheduling
After rewriting this half a dozen times, I stripped out as many scheduling moving parts as possible. 

Now Nightmare tasks resolution will run in a single blocking proc, and SSticker will block up to 30 seconds waiting for it to be ready.

In the event this somehow fails, nightmare will stop launching tasks, and game will start anyway.

It's not perfect but far simpler than alternatives...

### Multiple contextes
NM now instanciates 3 contexts: 
 * One for operations done pertaining to ground map
 * One for ship map
 * A global round one

Map ones are straightforward, but Global one is to enable further usages of nightmare - eg. tie in to global states (player count, pred rounds, set a global variation for jobs, etc)

## Changelog

:cl:
refactor: Reimplemented Nightmare system, now with support for ship maps and round-wide events.
refactor: Predator rounds are now rolled by Nightmare. As a side-effect, they can now be toggled before round start.
/:cl:


carlarctg:
something I've been interested in for a while is adding an insert that requires th round to strictly not be a pred round, does this make it possible?

fira:
@carlarctg It would require also refactoring pred rounds enabling, then yes. Pred round could set a global variable in NM (or outright be managed by NM) then you put it as condition with a global variable on the insert

fira:
Done the above as a proof-of-concept for round-wide nightmare setup

Nightmare now handles rolling pred round, and by essence of nightmare being in-lobby prep, this also means you can toggle it before game start

fira:
Being Testmerged, doesn't seem to have any particular issues

code/controllers/subsystem/nightmare.dm
```diff
@@ -0,0 +1,138 @@
...
+/// Reads a JSON file, returns list of config nodes in the file
+/datum/controller/subsystem/nightmare/proc/parse_file(filename)
+	RETURN_TYPE(/list/datum/nmnode)
+	. = list()
+	var/data = file(filename)
+	if(!data)
+		log_debug("Nightmare: Failed to read config file: [filename]")
+		CRASH("Could not get requested nightmare config file!")
+	if(data) data = file2text(data)
+	if(data) data = json_decode(data)
```
harryob, line 112:
```suggestion
	if(data) 
		data = file2text(data)
	if(data) 
		data = json_decode(data)
```
something something styles.md doesn't like inline control statements

code/controllers/subsystem/nightmare.dm
```diff
@@ -0,0 +1,138 @@
...
+	RETURN_TYPE(/list/datum/nmnode)
+	if(!islist(parsed)) return
+	var/list/datum/nmnode/nodes = list()
+	if(!parsed["type"]) // This is a JSON array
+		for(var/list/spec as anything in parsed)
+			var/datum/nmnode/N = read_node(spec)
+			if(N) nodes += N
+	else // This is a JSON hash
+		var/datum/nmnode/N = read_node(parsed)
+		if(N) nodes += N
```
harryob, line 126:
```suggestion
			if(N) 
				nodes += N
	else // This is a JSON hash
		var/datum/nmnode/N = read_node(parsed)
		if(N) 
			nodes += N
```
same

code/modules/nightmare/nmtasks/mapscheduler.dm
```diff
@@ -0,0 +1,35 @@
...
+
+/datum/nmtask/scheduler/mapload/proc/register_tainted_bounds(datum/nmtask/task, list/bounds)
+	tainted_bounds.len++
+	tainted_bounds[tainted_bounds.len] = bounds
+
+/datum/nmtask/scheduler/mapload/proc/patch_lighting()
+	var/list/tainted = list()
+
+	for(var/list/bounds as anything in tainted_bounds)
+		var/list/TT = 	block(	locate(bounds[MAP_MINX], bounds[MAP_MINY], bounds[MAP_MINZ]),
```
harryob, line 25:
```suggestion
		var/list/TT = block(locate(bounds[MAP_MINX], bounds[MAP_MINY], bounds[MAP_MINZ]),
```
is this spacing important? i know block is a fucky little define, so lmk

harryob (changes requested):
there's quite a lot of single letter var names, too, but most of them are only referenced next to their definitions so i'm not particularly bothered

code/modules/nightmare/nmtasks/mapscheduler.dm
fira: oh no that's probably an accidental leftover from refactoring similar code in another version

harryob (approved):

Nanu308 (approved):
we  merge

PR #1377, opened 2022-10-30
zombie disease rework

aceroyuno:
## About The Pull Request
It reworks the zombie disease to not be an immediate death sentence upon contraction. Instead now, it progresses based on whether or not it has a reagent in the body that is administered by zombie claw attacks. Once humans start coughing up black bile, that is when the disease no longer needs the zombie reagent.

## Why It's Good For The Game
No more will people be systematically slaughtered like animals when the infection starts. No more will zombies just run around trying to slash as many people as they can before they get downed, now they have to actually focus targets, which makes the slowdown all the better since there is actually a point to it other than getting another chance for the disease to proc. 

## Changelog

:cl: aceroy
add: Added new zombie reagent that affects the progression of the zombie Virus
balance: Zombie virus now progress based on different criteria. Zombie slashes now apply a new reagent to the target (15u a slash), at 30 units of the reagent the disease is able to progress. If the virus reaches stage 3 it will progress no matter what. And if the target is dead, the virus will progress.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry_reagents/other.dm
+++ b/code/modules/reagents/chemistry_reagents/other.dm
@@ -653,7 +653,7 @@
 	color = "#604030" // rgb: 96, 64, 48
 	chemclass = CHEM_CLASS_UNCOMMON
 
-/datum/reagent/blackgoo
+/datum/reagent/blackgoo //This is the version that quickly causes a person to be infected
 	name = "Black goo"
 	id = "blackgoo"
 	description = "A strange dark liquid of unknown origin and effect."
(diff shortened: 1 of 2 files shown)


code/modules/reagents/chemistry_reagents/other.dm
```diff
@@ -666,14 +666,27 @@
...
+	name = "Weak Black goo"
+	id = "blackgooweak"
+	custom_metabolism = REAGENTS_METABOLISM
+
+/datum/reagent/blackgoo/weak/reaction_mob(mob/m, var/method, var/volume)
+	return
+
+
+
+
```
stanalbatross, line 689:
```suggestion
```

aceroyuno:
@Geevies There is a current merge conflict with your recent zombie rework and mine. I've changed the format of everything to fit with your change, however there is still some mechanical differences in how I implemented it versus yours. I would like your approval to merge mine since you're the one I am having conflict with.

PR #1380, opened 2022-10-30
Faction Comms Overhaul

realforest2001:
## About The Pull Request
UPP, PMCs and CLF all now have General, Engineering, Medical, CCT (JTAC) and Command comms.
UPP have Kommando comms also.
All three factions can now update comms towers properly and operate as actual factions rather than oversized ERTs.

All UPP/PMC & CLF channels share colours. PMC colour is the current CCT colour.
image

## Why It's Good For The Game
Makes HvH a lot more dynamic rather than other factions being entirely on two channels. Also allows HvH where PMCs and UPP/CLF can be involved without all sharing the CCT network.

## Changelog

:cl:
fix: Removed bloat/duplicate code in the contractors ERT while doing headset updates
add: Added Command, Engineering, Medical and CCT channels for PMCs, CLF and UPP.
add: Added headsets and encryption keys for the above, and a few other channels.
add: Added keys to talk on faction channels without needing to juggle keys to make :H work.
add: UPP channels mostly reflect the USCM channel allocations (command is #V compared to USCM :V) using V/M/N/J/U General/T Commando
add: CLF channels are A Med/B Engi/C Cmd/D CCT/G General
add: PMC channels are E Engi/F Med/P General/O CCT/Z Cmd /K Whiteout
/:cl:


stanalbatross (changes requested):
`.-channel` for radios being removed is a major detriment to communication as it's way faster than pressing : and many people (including myself) are used to using . instead

no merge with this change still in it

ThePiachu:
As little as I use `#`s in my radio, I agree with Stan - we shouldn't be deprecating the functionality.

Maybe a better solution would be to let all the factions use the same comm schema, but connect to their individual departments. So if you used `:m` on UPP radio it would send a message to UPP Medbay, CLF would send to CLF Medbay, etc. It would make the radio keys not interoperable, but that's fine I guess?

Of course this would require a bigger code rewrite.

In the meantime you could merge some of the code refactor (the 1 to TRUE) separately so the next iteration of this would be a bit simpler.

realforest2001:
> As little as I use `#`s in my radio, I agree with Stan - we shouldn't be deprecating the functionality.
> 
> Maybe a better solution would be to let all the factions use the same comm schema, but connect to their individual departments. So if you used `:m` on UPP radio it would send a message to UPP Medbay, CLF would send to CLF Medbay, etc. It would make the radio keys not interoperable, but that's fine I guess?
> 
> Of course this would require a bigger code rewrite.
> 
> In the meantime you could merge some of the code refactor (the 1 to TRUE) separately so the next iteration of this would be a bit simpler.

Prior to this PR, using # would not work anyway. It's been in the code for an extremely long time, but never actually worked. See carbon/human/say.dm

code/controllers/subsystem/communications.dm
```diff
@@ -69,39 +69,61 @@ Radiochat range: 1441 to 1489 (most devices refuse to be tune to other frequency
...
+var/const/CLF_ENGI_FREQ	= 1273
+var/const/CLF_MED_FREQ	= 1274
+var/const/CLF_CCT_FREQ	= 1275
+
+var/const/AI_FREQ 		= 1500
+var/const/CCT_FREQ		= 1350//TEMP FOR USE IN MAKING SURE THE OTHERS ARE IN THE RIGHT PLACES
+
+//USCM High Command (USCM 1470-1499)
+var/const/HC_FREQ		= 1471
+var/const/SOF_FREQ		= 1472
```
stanalbatross, line 106: out of order

code/game/machinery/telecomms/presets.dm
```diff
@@ -160,7 +160,7 @@
 	icon_state = "relay"
 	id = "UPP Relay"
 	hide = TRUE
-	freq_listening = list(RUS_FREQ, CCT_FREQ)
+	freq_listening = list(UPP_FREQ, UPP_CMD_FREQ, UPP_ENGI_FREQ, UPP_MED_FREQ, UPP_CCT_FREQ, UPP_KDO_FREQ)
```
stanalbatross, line 163: can use UPP_FREQS define here

code/game/machinery/telecomms/presets.dm
```diff
@@ -171,11 +171,11 @@
 	return ..()
 
 /obj/structure/machinery/telecomms/relay/preset/tower/faction/clf
-	freq_listening = list(CLF_FREQ, CCT_FREQ)
+	freq_listening = list(CLF_FREQ, CLF_CMD_FREQ, CLF_ENGI_FREQ, CLF_MED_FREQ, CLF_CCT_FREQ)
 	faction_shorthand = "CLF"
 
 /obj/structure/machinery/telecomms/relay/preset/tower/faction/pmc
-	freq_listening = list(PMC_FREQ, CCT_FREQ)
+	freq_listening = list(PMC_FREQ, PMC_CMD_FREQ, PMC_ENGI_FREQ, PMC_MED_FREQ, PMC_CCT_FREQ, WY_FREQ)
```
stanalbatross, line 178: same and same

code/game/machinery/telecomms/presets.dm
```diff
@@ -251,9 +251,11 @@ GLOBAL_LIST_EMPTY(all_static_telecomms_towers)
 				if(FACTION_SURVIVOR)
 					freq_listening |= COLONY_FREQ
 				if(FACTION_CLF)
-					freq_listening |= list(CLF_FREQ, CCT_FREQ)
+					freq_listening |= list(CLF_FREQ, CLF_CMD_FREQ, CLF_ENGI_FREQ, CLF_MED_FREQ, CLF_CCT_FREQ)
 				if(FACTION_UPP)
-					freq_listening |= list(RUS_FREQ, CCT_FREQ)
+					freq_listening |= list(UPP_FREQ, UPP_CMD_FREQ, UPP_ENGI_FREQ, UPP_MED_FREQ, UPP_CCT_FREQ, UPP_KDO_FREQ)
+				if(FACTION_WY,FACTION_PMC)
+					freq_listening |= list(PMC_FREQ, PMC_CMD_FREQ, PMC_ENGI_FREQ, PMC_MED_FREQ, PMC_CCT_FREQ, WY_FREQ)
```
stanalbatross, line 258: same same same

stanalbatross (changes requested):

stanalbatross:
after this is merged now that I have actually learned map() procs in js I can take another look at making the UI show the channel key

stanalbatross (approved):

PR #1389, opened 2022-10-31
Fixed #1352 - wrong CL gear preset

### LINKED ISSUE #1352: Roundstart CL spawning in wrong area and without proper access

ThePiachu:
## About The Pull Request

Fixed bug #1352 . CLs were getting wrong presets - a Nightmare variant which had wrong ID access and wrong starting text. Fixed it by removing the preset - it doesn't seem to be used anywhere.

I think the bug was caused by one of those global variables that holds all the job / gear datums. It was probably indexed based on some field that was not altered between /datum/equipment_preset/uscm_ship/liaison and /datum/equipment_preset/uscm_ship/liaison/nightmare . I tried figuring out what it was indexed by but it would probably take much more time than this is worth...

## Why It's Good For The Game

It lets CLs access the doors they need and spawn at their office, otherwise they have to ahelp or get crafty to get back into where they need to go.

## Changelog

:cl:
fix: CLs should start with proper access and start location now.
/:cl:

### DIFF
--- a/code/game/jobs/job/civilians/other/liaison.dm
+++ b/code/game/jobs/job/civilians/other/liaison.dm
@@ -18,11 +18,6 @@
 	SIGNAL_HANDLER
 	active_liaison = null
 
-/datum/job/civilian/liaison/nightmare
-	flags_startup_parameters = NO_FLAGS
-	gear_preset = /datum/equipment_preset/uscm_ship/liaison/nightmare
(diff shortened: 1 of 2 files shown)


harryob:
duplicate of #1360 😔

ThePiachu:
Whops! Didn't notice the PR.

PR #1405, opened 2022-11-02
Larva afk remover

Shad0vvs:
## About The Pull Request

Larva (that have a ckey) will have 120 seconds to evolve or move from the pool, if they are within the spawn pool by 2 tiles as a larva at any point after this they will instantly be placed back into the pool. 
No ckey larva instantly get placed in as usual.

## Why It's Good For The Game

bugfix + no more overhead dealing with afk larva.

## Changelog

:cl:
add: Larva within range of the pool after 120 seconds will be placed back in the pool to help remove afk larva.
fix: Fixed spawn pools not having a hive check on which larva go back in the pool.
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/special/spawn_pool.dm
+++ b/code/modules/cm_aliens/structures/special/spawn_pool.dm
@@ -10,6 +10,7 @@
 	var/surge_cooldown = 90 SECONDS
 	var/surge_incremental_reduction = 3 SECONDS
 	var/mob/melting_body
+	var/larva_leeway_time = 120 SECONDS //Amount of time larva get to evolve before going back in the pool.
 
 	luminosity = 3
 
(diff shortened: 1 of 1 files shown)


code/modules/cm_aliens/structures/special/spawn_pool.dm
```diff
@@ -107,11 +108,12 @@
 		return
 
 	for(var/mob/living/carbon/Xenomorph/Larva/L in range(2, src))
-		if(!L.ckey && L.poolable && !QDELETED(L))
-			visible_message(SPAN_XENODANGER("[L] quickly dives into the pool."))
-			linked_hive.stored_larva++
-			linked_hive.hive_ui.update_pooled_larva()
-			qdel(L)
+		if(L.hivenumber == linked_hive.hivenumber)
+			if((!L.ckey || (world.time >= (L.life_time_start + larva_leeway_time))) && L.poolable && !QDELETED(L))
```
stanalbatross, line 112: what actually changes here if ckey is kept

Shad0vvs: Thats original code for ghosted ones, we want ghosted to go away immediately instead of having a delay like the ones with ckeys i've just added.

Shad0vvs:
Might not need merge considering resolving TGUI issues instead.

code/modules/cm_aliens/structures/special/spawn_pool.dm
fira, line 112:
```suggestion
			if((!L.ckey || (world.time >= (L.life_time_start + larva_leeway_time))) && L.poolable))
```
if it's `gc_destroyed` it's not `in range()` to begin with

fira (approved):

PR #1412, opened 2022-11-03
Intel Tweaks - telephone, skills

ThePiachu:
## About The Pull Request

Added a telephone to the Intel room (feels like they should be able to communicate more readily with the dropships and boots on the ground). Gave the Intel skill to POs, SLs and RTOs. Gave IOs an optional telephone backpack in their gear loadout.

## Why It's Good For The Game

After playing a few rounds it feels like Intel isn't being collected as readily, especially from the combat zones. Part of it could be due to not very many people having access to the Intel skill required to see the Objectives. Out of the boots on the ground, only IOs and the Synth can do it. So figured I'd tweak things a little bit to get them in a better direction.

First of all, telephone in the IO room. It will help communicate with squads and what have you IF comms are down. I don't think it will see too much use, but it fits with the job really.

image

Secondly, a few more people got the Intel skill. POs mainly because they are the only Officers without it (hecking CMO has it, I think POs deserve it more than them). SLs got it because they are on the way to being SOs, so it makes a bit of sense. I don't expect them to be using it much though seeing how they are in the thick of things. RTOs got it because they are meant to be communicating with everyone, so they have a more natural progression into being IOs, AND they are more likely to have a bit of time to check the Objectives than the SLs (another reason for SLs getting the skill - RTOs having it).

Thirdly, gave IOs an optional radio telephone backpack in case they want to roll with that setup. Saw an IO ask for it the other round and honestly, it's not a bad idea as an option for them to keep in touch even when the comms are down and what have you.

So with these changes hopefully the squads will be more aware of the objectives for stuff like uploading data and what have you, while not giving every unga access to it. I was considering adding an Intel Skill Pamphlet, but that would probably be too much.

## Changelog

:cl:
add: Added a telephone to IO's room. Now you can call the RTOs on the field to get you your intel!
add: IOs can now pick a radio telephone as their starting gear if they prefer to focus on communication over having extra backpack space.
balance: Gave POs, SLs and RTOs basic Intel skill so they can see the Intel Objectives.
/:cl:


Agameofscones (approved):
Mapping is A-ok

ThePiachu:
EDIT:

Gave IOs an optional radio telephone backpack. Saw an IO rolling with it the other round and honestly it's not a bad idea, and works pretty well with this phone addition to their office.

Issue #1222, opened 2022-10-10
Impossible 'upload data' objectives


Mark9013100:
## Description

Over multiple rounds, despite getting all intel items (papers, manuals, folders), I've failed to receive passwords for two terminals, thus preventing completion of the 'upload data' objective.

## Difference between expected and actual behavior

For the passwords to be given from intel items.

## Steps to reproduce

Retrieve all intel items, read, count the number of intelligence terminals. This is best done on LV624.

## Issue bingo (not required to be all checked)

- [x] Issue could be reproduced at least once
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Segrain:
Known issue. It is somewhere on my list, along with getting passwords for terminals that do not actually exist.

ThePiachu:
I think I know what this might be caused by - Objectives only get generated for regular map. If we have any Terminal on a Nightmare Insert, it will get an Objective but the Objective won't have any clues associated with them (so it won't be connected to other things like Folders and so on).

We ought to generate those Clues whenever a new Objective spawns, which might be a bit tricky due to map spawn order not being predictable. I might see what I can do about it soon since I'm currently in the midst of making sure IO stuff works 100%...

PR #1413, opened 2022-11-03
Fixed View Objective null bug

### LINKED ISSUE #832: View Objectives interface sometimes BSODs when viewing the data tab

ThePiachu:
Should fix #832 and maybe #1222

## About The Pull Request

Fixed the null-related "BSOD" that appears on View Objectives screen:

image
(this is after reading all the paper scraps, progress reports, folders and manuals, we have all 3/3 Terminal Objectives, while LV usually spawns with 5 Terminals)

It was caused by us trying to pass null Terminal to TGUI. We were passing a null Terminal because:
- Nightmare Inserts were deleting terminals in some areas, such as LV crashed ship and science nexus
- Deleted terminals nulled out the Terminals variable in their objective
- The deleted terminals never deleted the objectives they were relating to

Now the terminals call for their associated objective to be deleted when they are being deleted, which untangles the objective from all other objectives.

Fixes #832

## Why It's Good For The Game

The BSOD was making it impossible for IOs to complete their work - they were unable to view other Terminal or Disk passwords. With this we shouldn't see those null errors and the game shouldn't break IOs' fun.

## Changelog

:cl:
fix: View Objectives Data "BSOD" problem should be fixed. It was caused by some Nightmare inserts deleting terminals.
/:cl:

### DIFF
--- a/code/modules/objectives/data_retrieval.dm
+++ b/code/modules/objectives/data_retrieval.dm
@@ -48,7 +48,7 @@
 	initial_area = get_area(terminal)
 
 /datum/cm_objective/retrieve_data/terminal/Destroy()
-	terminal.objective = null
+	terminal?.objective = null
 	terminal = null
 	return ..()
 
@@ -243,6 +243,7 @@
 
 /obj/structure/machinery/computer/objective/Destroy()
 	objective?.terminal = null
+	qdel(objective)
 	objective = null
 	return ..()
 


ThePiachu:
I guess I'll look at the other objectives and how they handle being deleted and give them the same treatment.

PR #1419, opened 2022-11-04
IO Oversight Fix

Misttyy:
## About The Pull Request

This fixes a couple small oversight. The IOs had a radio before, but it wasn't actually accessible ingame, IOs started with irregular hunger and their ; on radios went to public.

## Why It's Good For The Game

These are a couple small oversights I believe from when I.O. was readded. Should made the role a little easier to get into.

## Changelog

:cl: Mistfrag
fix: IOs now start with low nutrition, rather than half. IOs have their proper headset integrated into gameplay and their default ; now = intel.
spellcheck: IO headset description updated to include missing channels.
/:cl:

### DIFF
--- a/code/game/jobs/job/command/auxiliary/intel.dm
+++ b/code/game/jobs/job/command/auxiliary/intel.dm
@@ -22,6 +22,11 @@
 			total_positions_so_far = positions
 	return positions
 
+/datum/job/command/intel/generate_entry_conditions(mob/living/carbon/human/H)
+	..()
+	if(!Check_WO())
+		H.nutrition = rand(NUTRITION_VERYLOW, NUTRITION_LOW)
+
 AddTimelock(/datum/job/command/intel, list(
 	JOB_SQUAD_ROLES = 5 HOURS
 ))
--- a/code/game/machinery/vending/vendor_types/intelligence_officer.dm
+++ b/code/game/machinery/vending/vendor_types/intelligence_officer.dm
@@ -55,7 +55,7 @@ GLOBAL_LIST_INIT(cm_vending_clothing_intelligence_officer, list(
 		list("Armor", 0, /obj/item/clothing/suit/storage/marine/rto/intel, MARINE_CAN_BUY_ARMOR, VENDOR_ITEM_MANDATORY),
 		list("Expedition Pack", 0, /obj/item/storage/backpack/marine/satchel/intel, MARINE_CAN_BUY_BACKPACK, VENDOR_ITEM_MANDATORY),
 		list("Gloves", 0, /obj/item/clothing/gloves/marine/insulated, MARINE_CAN_BUY_GLOVES, VENDOR_ITEM_MANDATORY),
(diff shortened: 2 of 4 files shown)


ThePiachu:
Disagree with using Intel as a default channel. Medical doesn't default to medchannel, engineering doesn't default to engi channel, intel shouldn't default to intel channel. I play Intel a lot and I'd expect to be able to use basic comms as default.

Definitely need to clean up the comm radios and the IO encryption key, but:
- If we were to use it, I'd keep the name obj/item/device/encryptionkey/io , it's consistent with others
- Seeing that IOs use command radios, I'd just delete the encryption key and the radio

fira:
I also think this was intended - you don't see eg. Dropship Crew with JTAC as default, either

Misttyy:
> Disagree with using Intel as a default channel. Medical doesn't default to medchannel, engineering doesn't default to engi channel, intel shouldn't default to intel channel. I play Intel a lot and I'd expect to be able to use basic comms as default.
> 
> 
> 
> Definitely need to clean up the comm radios and the IO encryption key, but:
> 
> - If we were to use it, I'd keep the name obj/item/device/encryptionkey/io , it's consistent with others
> 
> - Seeing that IOs use command radios, I'd just delete the encryption key and the radio

I'm not sure I follow on the relevancy here. 

You can't use Almayer comms ground side, where you'll be for the majority of the round. 

I've flirted with the idea of making all other departments default to their department radio, ala how squad radios work, but that's a bit out of the scope of this PR.

code/game/jobs/job/command/auxiliary/intel.dm
```diff
@@ -22,6 +22,11 @@
 			total_positions_so_far = positions
 	return positions
 
+/datum/job/command/intel/generate_entry_conditions(mob/living/carbon/human/H)
+	..()
+	if(!Check_WO())
+		H.nutrition = rand(NUTRITION_VERYLOW, NUTRITION_LOW)
```
ThePiachu, line 28: Hmm, since IOs would be hungry on the level of cryo marines, where would you expect them to get food? As I understand MREs are meant to be used when you deploy and get hungry, not as your starting source of food. And their closest source of food is what, lower medbay or squad preps? That's a long way to hobble. Why do you think this will improve the IO gameplay?

Also, I think we handle the basic hungryness you spawn with via

```
/datum/equipment_preset/uscm/load_status(mob/living/carbon/human/H)
	H.nutrition = rand(NUTRITION_VERYLOW, NUTRITION_LOW)
```

ThePiachu: Alright, I've tested how much impact it has and it basically makes you take extra 10 seconds to go from cryo to briefing food vendors. So overall it might be 1-2 minute detour after you've geared up, which is fine enough.

Still make it a part of the gear preset by removing this:
```
/datum/equipment_preset/uscm/intel/load_status()
	return //No cryo munchies
```

ThePiachu (changes requested):
Overall, the cryo hunger is okay enough, I'm personally not a fan of it, but I won't block the PR on those grounds.

Giving IOs more channels is fine enough for me, but I still don't think they should default to IO channel rather than the Almayer channel.

PR #1421, opened 2022-11-04
Tp say

ThePiachu:
They should always be upper and lower case, so be it!

## About The Pull Request

Removed unused comm channels.

## Why It's Good For The Game

I was in the process of documenting the available Comm Channels I stumbled on a few that were never used in our code. So freed a few letters for us, a whole WHIRL.

I would also either delete the support for cyrylic equivalents, or actually support them. They are dangling in the code and I'm not sure anyone uses them, but didn't want to pull the plug just yet without consulting first.

## Changelog

:cl:
del: Removed unused comm channels, brought to you by the letters WHIRL.
/:cl:


harryob (changes requested):
these aren't unused - with the possible exception of the ears options. for example, whisper can be used to quickly whisper without the whisper verb, eg, say ":w This will be whispered." similarly, :h is used to speak on the current department selected radio, eg, if you're medical :h will output on the MedSci channel. :i is also used for talking on intercoms when you're adjacent to them

ThePiachu:
Ah, my mistake. I was searching for "RADIO_MODE_WHISPER" rather than "whisper". I guess there is some other kind of code I ought to clean up then to make sure those things are used properly (aka replace the instances of `if("whisper")` with `if(RADIO_MODE_WHISPER)`).

Onto refactoring!

ThePiachu:
Ah, sorry about that. I don't see the Draft option anywhere on the screen though, so how would I do that?

PR #1422, opened 2022-11-04
More objectives delete themselves properly, deleted redundant file

ThePiachu:
## About The Pull Request

Building on [#1413](https://github.com/cmss13-devs/cmss13/pull/1413) - made sure other objectives also delete themselves gracefully (only ones I haven't touched were the esoteric ones like power, comms, etc. and research chemicals). Also deleted a file that was storing duplicate code for item retrieval objectives (It was already unlinked in the DME).

With this and the 1413 PR, all obj based objective should gracefully delete themselves when their tied item is deleted, such as by Nightmare Inserts. Should address the rare bug related to that mentioned by fira here - https://github.com/cmss13-devs/cmss13/pull/1413#pullrequestreview-1167185506 .

## Why It's Good For The Game

Less broken objetives means better game for IOs, less potential room for null reference errors, and this also lays good groundwork for fixing #1222 - if we have objectives that delete themselves, we can then build new clues into old objectives and link newly created objectives into existing clues, etc.

## Changelog

:cl:
code: All remaining objectives (other than computer data objectives, addressed separately) should gracefully delete themselves when their host item is deleted.
del: Removed an old, unused, duplicate file relating to objectives.
/:cl:


fira (approved):
Looks good to me. Deleting them properly means `/datum/cm_objective/Destroy` gets to properly unlink them in the objective tree. So hopefully they stop appearing in random places now

fira:
Thinking back - you sure it's Nightmare causing these issues? I actually planned it out when adding it and made it so `/datum/controller/subsystem/objectives/proc/generate_objectives` is called only on round start, and not game init. By that point, all documents should have been deleted/replaced already

ThePiachu:
I'm pretty sure it's Nightmare inserts. From what I've seen, the only times the Computers had dangling Clues was when those computers were nowhere to be seen on LV - usually the medical dome and the crashed ship.

I guess I could make a separate issue about that?

ThePiachu:
I guess there COULD be an edge case where like, a computer is created with base map, then overwritten with Nightmare insert and put in deletion, THEN the objective controller runs and connects the deleted computer to the intel stuff. I'd hope that's not the case because that would mean the deletions aren't handled properly somewhere else (maybe objective controller?).

PR #1427, opened 2022-11-04
Experimental: Mouse clicks Subsystem

fira:
## About The Pull Request

---  EXPERIMENTAL ---
--- TESTMERGE ME ---

When Walter ported the Input subsystem, we saw a marked performance benefit. 

That's because (from my understanding of /tg/ discussions in that topic) despite actually being slower, it avoids the bulk of movement code to be scheduled on verbs time. Verbs (including movement) get executed at the end of the tick, and will NOT be properly accounted by the MC for scheduling.

I emit henceforth the theory that this also applies to click actions. From experience, debugging on live you see a lot of overtime in click-adjacent procs, mainly guns. 

This adds similar treatment to clicks (backing by a MC Subsystem), in hope of solving that issue.

## Why It's Good For The Game

This should in theory have the same effect as we observed with SSinput: a marginal performance loss, but game running more smooth because it needs not overtime on clickies anymore.

The effect should be most noticeable on low-mid pop, because high pop is simply overloaded.

## Changelog

:cl:
refactor: Refactored click code to use a Subsystem, hopefully smoothing things.
/:cl:

### DIFF
--- a/ColonialMarinesALPHA.dme
+++ b/ColonialMarinesALPHA.dme
@@ -197,6 +197,7 @@
 #include "code\controllers\subsystem\atoms.dm"
 #include "code\controllers\subsystem\cellauto.dm"
 #include "code\controllers\subsystem\chat.dm"
+#include "code\controllers\subsystem\clicks.dm"
 #include "code\controllers\subsystem\communications.dm"
 #include "code\controllers\subsystem\dcs.dm"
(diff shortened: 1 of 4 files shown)


harryob (approved):
looks good for TMing - will be interesting

fira:
other fish to fry for now

PR #1430, opened 2022-11-04
Dropship matrices

kiVts:
## About The Pull Request

Adds matrices for research to make. While technically anyone can print them, only research can make good use of it. 
Printed at auto lathe, does nothing by itself - it requires a vial of chemical with one of two new properties. Photosensitive, and crystallization. The first, when inserted into the matrix, gives night vision. The second gives a bigger view. the quality of the matrix is losely dependent on the higher level of the property(eg  if level is <=5 then it gives a bad, unpleasant dark green color filter, if >= 6 then it gives NVG color filter. and same applied to wider view. If chemical has nothing we need in it, it gives a normal camera. Refuses to accept not full/not pure vials. Vial cant be removed after you insert it. if both chemicals are present, it only picks one(should pick wide by itself), can be changed so that both upgrades are given, The properties are rare as MST

probably contains a lots and lots of shitcode because this is first time me doing glorified copy pasting instead of just copypasting(harpoon)

(thanks everyone in coder-channel for help)

The code works in game, but theres still some todos.
mainly two

- make it work on direct view too, the problem right now is that it works a bit differently than FM camera. and I couldn't be bothered to look into it
- currently the sprite I use is the implant case, while it looks okay-ish Im believe there could be a better one.( I dont think its used anywhere though)

## Why It's Good For The Game

It gives a way for research to affect other aspects of the game, by giving a nice, and good way to give POs their part of the pie when researchers do something.

## Changelog

:cl:
add: Adds camera matrices for dropships, in order to work it needs a vial with chemical which has a special property that gives a FM camera buffs, dependent on the level of the property.
add: Two new properties for research to help make Camera matrices.
/:cl:


fira:
This sounds real cool, you got pictures to show how this looks like?

I think it's kinda important this would be the case on cameras of both firing modes though

kiVts:
screenshots in a moment

kiVts:
nightvision
this is what you get everything above level 5
image
this is what you get everything under level 6
image

wider view
this is what you get everything under level 6
image
this is what you get everything above level 5
image

fira (approved):
Code looks good to me, and the idea too

kiVts:
Back to draft for today while I make it work for direct view and make each level bit more meaningful

kiVts:
Back to draft for today while I make it work for direct view and make each level bit more meaningful

kiVts:
alright everything is done, it now works with direct view, made apply_upgrades a separate proc, and had to shift things in direct view to make it work
now any level has its own impact, for example, if you have level 3 you would get this color 
image
if you have level 6 you would get this color
image
any level >7 and filter is gone, but noise never goes away. and it works for every level diffrently

same applies to wider view, for example level 1 gives you 6 tiles (7 is default) up to a maximum of 15 tiles on level 10( theres no limit however)

kiVts:
You get a little bit yellow tinted color but I think its okay and actually okay

PR #1431, opened 2022-11-05
Fixes Statpanel Loading (not really, but mostly prevent error)

fira:
## About The Pull Request

This half-fixes Statpanel loading issues.

1. It removes the arbitrary delays added by c0ders of the past that seem to do nothing at all in practice
2. It adds a dummy loop check in statpanel main script to retry initialization
3. This causes the panel to load, although still display 'Loading...'

How initialization works to begin with, you may ask ?
It's VERY SIMPLE, let's take as example a client reconnect, which exhibits our dreaded issue:

1. In /client/New, the statpanel window is opened
2. In /client/New, we attempt statpanel init by serializing it relevant data (THIS DOESN'T WORK)
3. JS side statpanel loads and notifies BYOND via winset that it has loaded (THIS DOESN'T WORK EITHER)
4. (30 SECONDS ELAPSE)
5. Statpanel failed to load, click here to reload the panel

This doesn't seem to actually be an assets problem. Having either side retry init fixes this. "Fixes" because the status tab remains on "Loading..." until you click something.

Frankly this whole thing needs to be put on-standard with the /tg/ one which uses their JS BYOND lib and probably (hopefully) doesn't have this issue. I very highly suspect thedonkified initially added these three-way loading fallbacks due to having problems aswell.

## Why It's Good For The Game

It's less pain for everyone involved, including me

## Changelog

:cl:
fix: Half fixed statpanel failing to load. It might still behave erratically.
/:cl:

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -357,10 +357,9 @@ GLOBAL_LIST_INIT(whitelisted_client_procs, list(
 		to_chat(src, FONT_SIZE_HUGE(SPAN_BOLDNOTICE("YOUR BYOND VERSION IS NOT WELL SUITED FOR THIS SERVER. Download latest BETA build or you may suffer random crashes or disconnects.")))
 
 	// Initialize tgui panel
-	src << browse(file('html/statbrowser.html'), "window=statbrowser")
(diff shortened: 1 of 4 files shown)


harryob (approved):
tested this on my tgs local, seems to restore statpanel loading. thank god

Shad0vvs (approved):

PR #1432, opened 2022-11-05
Small radio refactor

ThePiachu:
## About The Pull Request

I was in the process of documenting the available Comm Channels I stumbled on a few that were never used, and some places that could use a small refactor, so cleared things up.

Removed left / right radio key bindings, we don't support that (and the left ear overlaps with LANGUAGE_TSL - they both use `:l`). Removed Syndicate radio. Removed vestigial cyrylic keyboard support - it was way outdated and I don't think anyone was supporting it on our codebase anymore.

Finally did a small refactor so two places use DEFINES rather than strings for whispers and intercom.

## Why It's Good For The Game

Removing old code makes the codebase better. Less overlap between language / radio channel makes the code less ambiguous come execution time, and DEFINES are better than strings just in case we change the strings in the future.

## Changelog

:cl:
del: Removed old, unused radio bindings
del: Removed old radio bindings for cyrylic keyboards that were way outdated
code: Small refactor of code around whisper / intercom radio channels
/:cl:

### DIFF
--- a/code/__DEFINES/radio.dm
+++ b/code/__DEFINES/radio.dm
@@ -24,6 +24,7 @@
 #define RADIO_CHANNEL_CLF "CLF"
 #define RADIO_CHANNEL_COLONY "Colony"
 #define RADIO_CHANNEL_DEPARTMENT "department"
+#define RADIO_CHANNEL_HEADSET "headset"
 #define RADIO_CHANNEL_DUTCH_DOZEN "DD"
 #define RADIO_CHANNEL_ENGI "Engi"
 #define RADIO_CHANNEL_ERT "Response Team"
--- a/code/__DEFINES/radio.dm
+++ b/code/__DEFINES/radio.dm
@@ -15,9 +15,6 @@
 #define RADIO_VOLUME_IMPORTANT_STR "Important"
 #define RADIO_VOLUME_CRITICAL_STR "Critical"
 #define RADIO_MODE_WHISPER "whisper"
-//Left and right ears
-#define RADIO_EAR_LEFT "left ear"
-#define RADIO_EAR_RIGHT "right ear"
 
 //Radio channels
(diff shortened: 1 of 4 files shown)


fira:
What do you mean by "we don't support that" ? We do have support for both ear slots -- do they not work with these bindings ?

ThePiachu:
I think the other ear is for a walkman? I remember trying to put a radio in there and it didn't work?

ThePiachu:
My mistake, you CAN put the radio in the other ear. BUT you can't put in two radios in at the same time:

image

ThePiachu:
items.dm line 530

`			if(WEAR_L_EAR)
				if(H.wear_l_ear)
					return FALSE
				if(HAS_TRAIT(src, TRAIT_ITEM_EAR_EXCLUSIVE))
					if(H.wear_r_ear && HAS_TRAIT(H.wear_r_ear, TRAIT_ITEM_EAR_EXCLUSIVE))
						if(!disable_warning)
							to_chat(H, SPAN_WARNING("You can't wear [src] while you have [H.wear_r_ear] in your right ear!"))
						return FALSE
				if(!(flags_equip_slot & SLOT_EAR))
					return FALSE
				return TRUE
			if(WEAR_R_EAR)
				if(H.wear_r_ear)
					return FALSE
				if(HAS_TRAIT(src, TRAIT_ITEM_EAR_EXCLUSIVE))
					if(H.wear_l_ear && HAS_TRAIT(H.wear_l_ear, TRAIT_ITEM_EAR_EXCLUSIVE))
						if(!disable_warning)
							to_chat(H, SPAN_WARNING("You can't wear [src] while you have [H.wear_l_ear] in your left ear!"))
						return FALSE
				if(!(flags_equip_slot & SLOT_EAR))
					return FALSE
				return TRUE`

fira:
Interesting. Then there is no point, indeed

fira (approved):

PR #1435, opened 2022-11-05
Dropship Fueling Mini-Game

Geevies:
## About The Pull Request

When a dropship that has a fuel enhancer and/or cooling system lands, the flight control console will now blare out a compound to add to those equipments, which reduces the cooldown time before a pilot can take off again.

The compound can be found in a compound silo, which can be found in the construction area south of the hanger. There are four possible compounds to add to the equipment.

Upon adding the correct compound, the cooldown is reduced by 10 seconds, otherwise, it's increased by 10 seconds.

## Why It's Good For The Game

It's something neat for MTs to handle, and perhaps DCCs if they're so inclined.

## Changelog

:cl:
balance: Fuel enhancers and cooling systems now have an active-reload system which reduces shuttle cooldown time.
/:cl:

### DIFF
--- a/code/modules/cm_marines/compound_silo.dm
+++ b/code/modules/cm_marines/compound_silo.dm
@@ -1,0 +1,191 @@
+/obj/structure/compound_silo
+	name = "compound silo"
+	desc = "A large silo containing a compound used to speed up dropship engine cooldowns."
+	icon = 'icons/obj/objects.dmi'
+	icon_state = "weldtank_old"
+	density = TRUE
+	anchored = TRUE
+	health = 250
+
+	var/compound_type = COMPOUND_PHI
+
+	var/obj/item/nozzle/attached_to
+	var/atom/tether_holder
+	var/tether_range = 7
+
+/obj/structure/compound_silo/omega
+	compound_type = COMPOUND_OMEGA
+
+/obj/structure/compound_silo/tau
+	compound_type = COMPOUND_TAU
+
+/obj/structure/compound_silo/epsilon
+	compound_type = COMPOUND_EPSILON
+
+/obj/structure/compound_silo/Initialize()
+	. = ..()
+
+	name = "[compound_type] silo"
+	attached_to = new /obj/item/nozzle(src, compound_type)
+	RegisterSignal(attached_to, COMSIG_PARENT_PREQDELETED, .proc/override_delete)
+
+/obj/structure/compound_silo/proc/override_delete()
+	SIGNAL_HANDLER
+	recall_nozzle()
+	return COMPONENT_ABORT_QDEL
+
+/obj/structure/compound_silo/proc/recall_nozzle()
+	if(ismob(attached_to.loc))
+		var/mob/M = attached_to.loc
+		M.drop_held_item(attached_to)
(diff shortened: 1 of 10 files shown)


code/modules/cm_marines/compound_silo.dm
```diff
@@ -0,0 +1,193 @@
...
+	. = ..()
+
+	name = "[compound_type] silo"
+	attached_to = new /obj/item/nozzle(src, compound_type)
+	RegisterSignal(attached_to, COMSIG_PARENT_PREQDELETED, .proc/override_delete)
+
+/obj/structure/compound_silo/proc/override_delete()
+	SIGNAL_HANDLER
+	recall_nozzle()
+	return COMPONENT_ABORT_QDEL
```
fira, line 35: if you're gonna use the atrocity that `COMPONENT_ABORT_QDEL` is (is there really a good reason?), please at least check for `force` argument so it still works with things like nightmare or shuttles that use maploading

code/modules/cm_marines/compound_silo.dm
```diff
@@ -0,0 +1,193 @@
...
+
+/obj/item/nozzle/proc/remove_attached()
+	attached_to = null
+	reset_tether()
+
+/obj/item/nozzle/proc/reset_tether()
+	SIGNAL_HANDLER
+	if (tether_effect)
+		UnregisterSignal(tether_effect, COMSIG_PARENT_QDELETING)
+		if(!QDESTROYING(tether_effect))
```
fira, line 110: why use QDESTROYING here? if there's a qdel loop it might be better handled otherwise

code/modules/cm_marines/compound_silo.dm
```diff
@@ -0,0 +1,193 @@
...
+/obj/item/nozzle/proc/silo_move_handler(var/datum/source)
+	SIGNAL_HANDLER
+	zlevel_transfer = FALSE
+	if(zlevel_transfer_timer)
+		deltimer(zlevel_transfer_timer)
+	UnregisterSignal(attached_to, COMSIG_MOVABLE_MOVED)
+	reset_tether()
+
+/obj/item/nozzle/proc/try_doing_tether()
+	zlevel_transfer_timer = TIMER_ID_NULL
```
fira, line 190:
```suggestion
/obj/item/nozzle/proc/try_doing_tether()
	SIGNAL_HANDLER
	zlevel_transfer_timer = TIMER_ID_NULL
```

code/modules/mob/mob.dm
```diff
@@ -169,11 +169,11 @@
 // Use for objects performing visible actions
 // message is output to anyone who can see, e.g. "The [src] does something!"
 // blind_message (optional) is what blind people will hear e.g. "You hear something!"
-/atom/proc/visible_message(message, blind_message, max_distance, message_flags = CHAT_TYPE_OTHER)
+/atom/proc/visible_message(message, blind_message, max_distance, message_flags = CHAT_TYPE_OTHER, show_small_icon = FALSE)
```
fira, line 172: please update the signature of `/mob/visible_message` to match, it's already different enough as it stands

fira (changes requested):

fira:
I'd argue this is more Feature than Balance

PR #1436, opened 2022-11-05
Added access to the back door of Requisitions

ThePiachu:
## About The Pull Request

I was testing which areas SOs have access to and noticed they were lacking access to one of the back rooms to the Requistiopn, while they were perfectly capable of walking in the side doors. So for the sake of consistency, made that door match the access to the others since if you can walk in from the front to that room, why shouldn't you be able to walk in from the back?

image
(the door on the right here was missing the access of ACCESS_MARINE_LOGISTICS)

## Why It's Good For The Game

More consistent access to rooms is always good.

## Changelog

:cl:
fix: Added a missing access permit to the Requisition's back door
/:cl:

### DIFF
--- a/maps/map_files/USS_Almayer/USS_Almayer.dmm
+++ b/maps/map_files/USS_Almayer/USS_Almayer.dmm
@@ -30852,7 +30852,8 @@
 "cbo" = (
 /obj/structure/machinery/door/airlock/almayer/maint{
 	req_access = null;
-	req_one_access = list(2,21)
+	req_one_access = null;
+	req_one_access_txt = "21"
 	},
 /obj/structure/machinery/door/firedoor/border_only/almayer,
 /turf/open/floor/almayer{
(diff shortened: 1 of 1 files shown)


maps/map_files/USS_Almayer/USS_Almayer.dmm
```diff
@@ -30852,8 +30852,7 @@
 "cbo" = (
 /obj/structure/machinery/door/airlock/almayer/maint{
 	req_access = null;
-	req_one_access = null;
-	req_one_access_txt = "21"
+	req_one_access = list(2,21)
```
fira, line 30855: please use `req_one_access_txt` for this, list instances in map files are deprecated

fira (changes requested):

maps/map_files/USS_Almayer/USS_Almayer.dmm
ThePiachu: Alright, fixed!

Huffie56:
Since your fixing access to door being weird.There is a lack of access require for the backdoor of the kitchen not sure if it's intended I could check for more if you wish?

fira (approved):

PR #1438, opened 2022-11-05
Infiltrator Synthetic and new Admin Faction feature

Unknownity:
## About The Pull Request

This PR brings in a new synthetic species based on the soul franchise of Aliens of synthetics disguised as humans in media. The Infiltrator synthetic species appear human at a glance, having the same blood color, and regular naming scheme..this also brings a new trait to the mix, named as TRAIT_INFILTRATOR_SYNTH which provides an extra camouflage to the infiltrator by scrambling all medHUDs to remove their synthetic status (icon), an Infiltrator can still be exposed by getting scanned from an analyzer, (attempting) to get pilled/injected and if they sustain enough damage, their flavor text in examine will show their synthetic parts. Xenomorphs being xenos, are able to see through the disguise instantly. The infiltrator's skill is the equivalent of a Ship Synthetic with the only three exceptions being that it now has the Antag skill to use the chloroform cloth its provided with and it's Vehicle skill has been increased to CREWMAN so they can use the gunner seat and their Pilot skill has been increased to EXPERT so they can use the dropship's weapon systems, making them truly possible to infiltrate as ANY role they want to. The gear preset of the infiltrator has its default faction set to neutral.

Another feature to be introduced to work with the former is an admin one, which is being able to change someone's faction, the required permission to use said feature is anyone who can varedit since you can do this in VVs too but an average admin might not know how, the feature can be found in the Admin Player Panel in the Physical tab alongside Set Squad. NOTE: This only changes the mob's faction, not the mob's ID card faction. all weapon IFF related things (SG,Sentries) is tied to the ID card so you will have to VV the ID to do the same, however certain things like the shipside explosive restrictions is tied to the mob's faction variable, so with this feature you can change the mob's faction from USCM to X and they will now be able to use grenades and place C4 on the Almayer during code green. I tried finding a way to make it simpler for the ID faction change but gave up.

Also fixed CLF synths not being gen one visually, I made them in mind to only appear as gen one and at some point, they all turned to gen threes visually with the cosmetic update or something.

## Why It's Good For The Game

Increases Event soul by adding synthetics who can actually pretend to be humans without people with medHUDs instantly calling them out now and makes an Admin's life easier with faction changing of the mob (but not the ID card faction changing..for now that is)

## Changelog

:cl:Unknownity
add: Added Infiltrator Synthetics that can be used for events, these guys are almost indistinguishable to humans, better watch your back in the future..
admin: Added a simpler way for admins to change a mob's faction, it's found in the player panel alongside set squad.
fix: Fixed CLF Synths visually appearing as Gen 3 synthetics
/:cl:


code/datums/mob_hud.dm
```diff
@@ -348,6 +348,8 @@ var/list/datum/mob_hud/huds = list(
 	holder4.icon_state = "hudblank"
 
 	if(species && species.flags & IS_SYNTHETIC)
+		if(HAS_TRAIT(src, TRAIT_MASKED_BIO))
```
Geevies, line 351: this isn't generic enough, as this only applies to synths while being a generic looking trait. additionally, i don't think this makes "hudhealthy" show up, which automatically looks suspicious. you probably need to write some custom code for that

Geevies (changes requested):

PR #1461, opened 2022-11-07
Tweaks to Vehicle Ammo in Requisitions

ThePiachu:
## About The Pull Request

While I was working on the original version of this PR I have made a few tweaks to package descriptions, capitalisation and so on for Vehicle Ammo in Requisitions. Figured those changes were good on their own and they can be merged to the code proper even if the rest of that PR got rejected.

None of these changes are that groundbreaking, but they are a neat tweak.

## Why It's Good For The Game

Consistent capitalisation and naming of crates draws less attention to itself. Consistency with putting flammable ammo into boxes that label them as such helps with immersion. Removing the "Modules" part of the category name prevents players from being confused when they don't see any modules listed in the category.

## Changelog

:cl:
fix: Renamed the "Vehicle Modules and Ammo" category to "Vehicle Ammo" (since it doesn't contain actual Vehicle Modules) to avoid confusion.
fix: Vehicle flamethrower ammo now comes in crates designated for flamable contents
spellcheck: Fixed spelling on TAA-AP Minigun magazine crate
fix: Made some vehicle crates a bit more accurate as to their content description
/:cl:

### DIFF
--- a/code/datums/supply_packs/_supply_packs.dm
+++ b/code/datums/supply_packs/_supply_packs.dm
@@ -18,7 +18,7 @@
 var/list/all_supply_groups = list(
 	"Operations",
 	"Weapons",
-	"Vehicle Modules and Ammo",
+	"Vehicle Ammo",
 	"Attachments",
 	"Ammo",
 	"Weapons Specialist Ammo",
--- a/code/datums/supply_packs/hardpoint_modules.dm
+++ b/code/datums/supply_packs/hardpoint_modules.dm
@@ -14,7 +14,7 @@
 	cost = RO_PRICE_CHEAP
 	containertype = /obj/structure/closet/crate/ammo
 	containername = "LTB cannon ammo crate"
-	group = "Vehicle Modules and Ammo"
+	group = "Vehicle Ammo"
 
 /datum/supply_packs/ammo_ltaaap_minigun
 	name = "LTAA-AP Minigun magazines (x3)"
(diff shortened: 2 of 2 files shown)


fira:
I dunno.... There's also a Vehicle Fabricator which could serve as alternative and integrate into tech rewards

ThePiachu:
Yes, my initial idea was to integrate it into tech webs and give them points to let them fabricate things. But since it doesn't seem to be on the map right now and the ammo is already in requisitions, and I have zero experience with modifying techwebs, it's not a bad place to put it.

PLUS if the APC is on the field and needs field repairs, a VC won't go all the way back to Almayer to make those things, while even if SO / XO / CO / RO could have access, it's on the other end of the ship and I think nobody would care to make that walk.

JeserTheFox:
Right now, vehicles are not supposed to have parts replaceable.

ThePiachu:
Is there a plan on making them replaceable in the future? Would you like me to make it so you can't buy them at the moment but they would be in the code in case we enable them?

ThePiachu:
Alright, since we don't want Vehicle Modules in Req, the PR has been trimmed down a lot to avoid these. Now it's just some small naming and other tweaks that don't change the functionality at all.

Huffie56:
So i did add a new category for ASRS and you need to also change name in DME file ColonialMarinesALPHA.dme or it won't work i think

fira (approved):

PR #1464, opened 2022-11-07
Gave SLs other kits to start with

ThePiachu:
## About The Pull Request

I've seen a number of times the SLs just take their starting pyro kit and toss it out for someone in their squad to take, which isn't a great thing if a large chunk of your essential kit is not used. So I figured - why not give the SLs some variety of what they can pick up to define their kit.

So the kits are mostly the kits you can find in Requisitions, plus the mini-pyro kit they had, an L42A kit if they want to go for something with IFF, and a basic engineering supplies kit which has a full electronics pouch (3 batteries, 3 apc boards), a construction pouch with 50 metal, 30 plasteel and 15 barbed wire, and welding goggles, so they only need to grab a toolbelt and they can fully repair the comms by themselves and do some basic defences around it.

image

Hopefully these will appeal to most SLs without being too unbalanced. If you think the numbers in the engineering kit are off or some of the kits don't fit well with the role, let me know, I'm a bit out of touch with the current meta.

## Why It's Good For The Game

More gear options means the players will be able to have the experience they want without having to bother Requisitions.

## Changelog

:cl:
add: SLs can now pick from a variety of starting weaponry kits rather than be forced to get the pyro kit.
code: Small refactor of flags around which items you can vend from a vendor.
/:cl:


ThePiachu:
After discussing it on Discord, I will be taking this PR in a slightly different direction. I don't know how to draft this PR though...

ThePiachu:
Removed the Mk1 rifle (didn't know that was a gun that was supposed to be avoided, figured if it's left in the code as a default loadout it should be good). Added the job kits and the rifle kit instead.

morrowwolf (changes requested):
These are the old PFC kits. Go through the kits and actually craft them from the ground up to be specific to SL.

ThePiachu:
Based on the discussion on Discord people seemed to like the idea of SL having kits for their squad to buff them and fill in the roles. But yeah, if we want to avoid those entirely, I'll have another look at what kind of kits the SL could have for themselves... Probably at this point it would be gun kits, maybe materials kit if you don't have engis, and maybe like a radio kit with a backpack that fits the SL...

ThePiachu:
Alright, out with the kits for your squad, in with more regular kits you can get in Requisitions as well as a new kit for the more crafty SLs. Let me know if something needs balancing here...

morrowwolf (changes requested):
Right off the bat MK1 kit is going to be denied. Remove it please.

ThePiachu:
Removed the MK1 rifle from the SL vendor as requested.

stanalbatross (commented):
my main concern is that this streamlines the SL gear-up procedure too much, and may lead to dropship AFKing
however SLs having more time predrop is good as it allows for squad RP
will approve for now, we'll see how it plays out

PR #1465, opened 2022-11-07
Grammar fixes and stuff

Unknownity:
## About The Pull Request

Fixes grammar mistakes left behind from the recently merged PR 1430 and also makes things more readable and easily understandable for the average user.
Photosensetive, Photosensitive? I don't care if they're both right but the latter sounds much more better and reads better.
Also the code CRS for the crystallization property already belongs to Corrosive so I changed it to a code that isn't owned by any property yet

Also the biggest change of all, the changelog for the PR will now actually show up so people know this new feature even exists and made the CL a bit better.

## Why It's Good For The Game

Proper English good and lets people know this exists

## Changelog

:cl:Kivts
add: Adds custom camera matrices for dropships which can be found in autolathes, in order for one to work, it needs a vial with a chemical which has any of the two new properties which provide upgrades to the camera system, dependant on their property level.
add: Added two new chemical properties which are used for the new feature, Crystallization and Photosensitive, which per order, increase the camera view of the dropship cam system and grants night vision to it.
spellcheck: Fixes grammar mistakes of the aforementioned new additions in the code.
/:cl:

### DIFF
--- a/code/__DEFINES/chemistry.dm
+++ b/code/__DEFINES/chemistry.dm
@@ -153,7 +153,7 @@
 #define PROPERTY_FLUXING 			"fluxing"
 #define PROPERTY_NEUROCRYOGENIC		"neurocryogenic"
 #define PROPERTY_CRYSTALLIZATION	"crystallization"
-#define PROPERTY_PHOTOSENSETIVE		"photosensetive"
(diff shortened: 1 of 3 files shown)


Unknownity:
also erm..remove the feature tag since it technically already exists in the game

kiVts:
🙃

PR #1471, opened 2022-11-08
Added some small items to Requisitions, fixed Full Electronics Pouch

ThePiachu:
## About The Pull Request

During a CT round I noticed the Electronics Pouch was missing from Requisitions, so added it. Similarly, I had a few SGs ask for batteries, so I added a small amount of them into the vendor as well (way less then regular electronics vendor). I like organising things around Requisitions and I was surprised how few Clipboards there are in that space (I think the only one was in RO's office), so I added a few on the various tables around there to help all those requests organised. Finally, one time I took the Full Electronics Pouch only to realise it wasn't actually full made me disappointed, so I actually filled that pouch with two more sets of battery and APC board.

## Why It's Good For The Game

Requisition should have all the pouches available since that's the one place people expect to come and get some. Since Requisitions has Smartgunner backpacks that come with the batteries, it's not much of a stretch to have a few batteries as well for the Smartgunners since they are basically using them like ammo. And if you advertise a pouch as being full, it should be actually full.

## Changelog

:cl:
add: Added a missing Electronics Pouch to Requisitions vendor
add: Added a few High-Capacity Power Cells to Requisitions so Smartgunners can get some spares easier
add: Added a few clipboards around Requisitions to help keep all the papers organised
fix: Full Electronics Pouch is now actually full
/:cl:


Agameofscones (approved):
Mapping is 🆗

stanalbatross (approved):

PR #1474, opened 2022-11-08
ASRS Pizza

fira:
## About The Pull Request

After coming onto the 738474th RO that tried to commemorate some event by ordering pizza from ASRS, and seeing their disappointement as they searched through the menus, yet one more time, i figured maybe this ought to be a thing.

Have you never been a CT working dilligently through the whole operation, to be promitted a break by the RO who ends up in complete disarray as the promised pizza is nowhere to be found?

NO MORE! 

This adds *randomized* pizza in ASRS. Comes in a stack of 3 for 4000$. Each pizza has random ingredients. This should hopefully be reserved for special occasions.

## Why It's Good For The Game

RP Hooks, maybe "soul" as some would say. 

Hopefully niche enough to not undermine the Mess own niche

Debated removing the Reqs Margherita with a note saying it was put back in ASRS, but didn't want to touch too much things.

## Changelog

:cl:
add: You can now order randomized pizza at ASRS, for special occasions or reqsreation.
/:cl:

### DIFF
--- a/code/datums/supply_packs/misc.dm
+++ b/code/datums/supply_packs/misc.dm
@@ -355,4 +355,12 @@
 	containername = "\improper donuts crate"
 	group = "Supplies"
 
-
+/datum/supply_packs/pizzas
+	name = "pizza ready-to-eat (x3)"
+	contains = list(
+		/obj/item/pizzabox/mystery/stack
+	)
+	cost = RO_PRICE_NORMAL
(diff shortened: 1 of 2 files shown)


harryob (approved):
we already have crayons for marines to eat, why not mystery pizza?

Huffie56:
It may conflict with my pull request https://github.com/cmss13-devs/cmss13/pull/1479.
no idea what we do in that case....

PR #1490, opened 2022-11-10
Removes cooldown on Queen gib timer for higher plasma costs and time

Nanotrasen-Spy:
Removes cooldown on Queen gib timer. In return it takes 20 seconds to gib (from 8) and 400 plasma (from 200)

## About The Pull Request

Removes cooldown on Queen gib timer. In return it takes 20 seconds to gib (from 8) and 400 plasma (from 200)

## Why It's Good For The Game

It doesn't make sense that the biggest and baddest Queen can't wretch someone apart because 'they need to wait to feel like it again' beyond their control. All it does is make it used explicitly and only for meta targeting specs and COs rather than, really, using it on anyone they could so choose to. In return it now takes 20 seconds and 400 plasma. For reference, Grow ovipositor costs 500 plasma and takes 20 seconds. The queen is a special caste and her abilities should reflect it other than the obvious neuro and screech just like the old days. The longer time should alleviate gib spams since with the time it takes now and how prevalent the Queen is in battle, it'd require the Queen to be out of fire to avoid being marine DPS'd. Cool and interesting abilities like this are soul

## Changelog

:cl:
balance: Removed Queen's gut ability cooldown timer. However it now costs 400 plasma and 20 seconds now (from 200 plasma and 8 seconds)
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Abilities.dm
@@ -149,13 +149,12 @@
 	..()
 
 /datum/action/xeno_action/activable/gut
-	name = "Gut (200)"
+	name = "Gut (400)"
 	action_icon_state = "gut"
 	ability_name = "gut"
 	macro_path = /datum/action/xeno_action/verb/verb_gut
 	action_type = XENO_ACTION_CLICK
-	xeno_cooldown = 15 MINUTES
-	plasma_cost = 200
+	plasma_cost = 400
 	cooldown_message = "You feel your anger return. You are ready to gut again."
 
 /datum/action/xeno_action/activable/gut/use_ability(atom/target)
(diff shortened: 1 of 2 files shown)


Huffie56:
I agree that cool down should be remove but you need to think about balance:
how many people queen can gib now/min now?
One every 15min due to cooldown
how many people will be able to gib/min after?
Queen as 1000 plasma? she can just probably chain gib marines with only the 20s being the limit.
doubling the time to gib and removing the cooldown is an huge boost to queen in some situation but a nerf to quickly gib special marines in combat.

stanalbatross:
No I don't think we will let the queen spam gib every good player she sees 

Also this basically mandates combat queening since being able to gib is *very* powerful, and nigh confirms the gibbing of any spec or hell any good marine if they die slightly out of position


Gibbing is too punishing to be this common. Consider this denied.

PR #1491, opened 2022-11-10
Adds flamer kits to the rifleman vendor

joookschad:
## About The Pull Request

Adds flamer kits back to the rifleman vendor, for 30 points.
## Why It's Good For The Game
Flamers are currently very limited, and I feel like putting them in the vendor would make marines push better

## Changelog

:cl:
add: Flamer kits are now in the rifleman vendor, for 30 points

/:cl:


stanalbatross:
Putting them back in vendors = literally every single marine can get one

Currently the flamer is too powerful for this to be OK

You can increase the number of them in req but this is too much

stanalbatross (changes requested):
Do my comment, can't draft because on mobile

PR #1492, opened 2022-11-10
Adds intel pamphlets to all squad vendors

joookschad:
## About The Pull Request
Adds intel pamphlets to all squad vendors

## Why It's Good For The Game

This change will help with the collection of intel; IO's will have assistants from squads.

## Changelog

:cl:
add: All squad roles can now buy intel pamphlets to help out IO's!

/:cl:


ThePiachu:
> All this to say, if intel pamphlets were to be added to vending machines I would rather they be bundled with an intel encryption key (and probably a data scanner too) so that we don't remove the ability/necessity to coordinate in order to complete objectives.

A mini intel kit! I think we had that at one point...

morrowwolf:
I feel we're doing this backwards, give IOs the ability to buy intel pamphlets they can give to people. Make it decently expensive but not insane.

morrowwolf:
This should have staled out and I believe it is the wrong direction as well as you have negative GBP. Get some bugs fixed and speak to an architect.

PR #1497, opened 2022-11-10
Fixes Tripod Lamps and candelabras

kugamo:
## About The Pull Request

By request @Agameofscones this PR fixes Tripod lamps and candelabras not starting on. Also adds ability to break them as a xeno. Also added ability to break normal flashlights: bringing the existing functionality for guns and the rail-light to them.

dreamseeker_v17JWIgabv

dreamseeker_JpWW7FA5X0

## Why It's Good For The Game

These items have been in our maps for a long while now but have remained non-functional, having a game with working features is good. For turning them off: having the same mechanics for similar items is good game design, its snowflakey otherwise.

## Changelog

:cl:
fix: Fixed tripod lights and candelabras not starting as on, fixed not being able to break these items and flashlights as xeno.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/flashlight.dm
+++ b/code/game/objects/items/devices/flashlight.dm
@@ -16,6 +16,9 @@
 	var/raillight_compatible = TRUE //Can this be turned into a rail light ?
 	var/toggleable = TRUE
 
+	var/can_be_broken = TRUE //can xenos swipe at this to break it/turn it off?
+	var/breaking_sound = 'sound/handling/click_2.ogg' //sound used when this happens
+
 /obj/item/device/flashlight/Initialize()
 	. = ..()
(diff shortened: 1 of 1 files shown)


DylonBR:
Shouldn't this have a balance or feature tag?

Oshuka1997:
this is a balance change not just a fix change , allowing all xenos to slash this lights takes away from having to rely on drones or sentinels to melt them

kugamo:
These lights didnt work at all before this pr, and its been made by request of the dev architect. The only case in which an unapproved balance change may apply is for flashlights, only because now they can be broken which makes them like all the other marine light sources (guns and rail lights). If a dev wants to label this a balance change for that then ok, but im not really breaking any new ground here.

PR #1499, opened 2022-11-10
Re-adds warrior bonebreak punch

carlarctg:
## About The Pull Request

Title. Additionally, warrior punch will slow on hit if it fails to break bones. This punch slow will apply to synthetics and yautjas as well, alongside xenomorphs (who recieve a differing amount based on their size)

Added a new species variable that contains a limb's minimum damage required for it to break in a fracture() call. Does not actually break the limb at that threshold by itself. If null, cancels any bonebreaks.

## Why It's Good For The Game

I enjoy Warrior bonebreaks for various reasons.
- They make Warrior feel like a menace, something to be afraid of.
- They uniquely interact with the human medical system in a way that I believe should be encouraged. (Stuff like applying chemicals like boiler tail stab does, it's a great way to mesh both xeno and human and add depth to the systems)
- They act as good attrition for marines and cause them to fear corners and going alone even further.

If needed, I could simply change the new minimum bonebreak damage variable to, say, 20, which will mean you can't run up to a marine and break their bones then run away. You'd have to slash them first.

Plus woyers got a nerf with lifesteal that was entirely part of phasing out shields in general so maybe this can help

GA approval:

https://media.discordapp.net/attachments/833081107113181204/1040352725621284874/image.png

## Changelog

:cl:
add: readded woyar punch
/:cl:

### DIFF
--- a/code/game/objects/effects/aliens.dm
+++ b/code/game/objects/effects/aliens.dm
@@ -195,33 +195,6 @@
 
 	var/bonus_damage = 5
 
-/obj/effect/xenomorph/spray/weak/apply_spray(mob/living/carbon/M)
-	if(ishuman(M))
-		var/mob/living/carbon/human/H = M
-
-		var/damage = damage_amount
-
-		var/should_stun = FALSE
-		for (var/datum/effects/acid/A in H.effects_list)
-			should_stun = TRUE
-			damage += (-1*(A.duration - A.original_duration))*(A.damage_in_total_human/A.original_duration)
-			damage += bonus_damage
-
-			//qdel(A)
-			break
-
-		to_chat(H, SPAN_DANGER("Your feet scald and burn! Argh!"))
-		H.emote("pain")
-		if (should_stun && !H.lying)
-			H.KnockDown(stun_duration)
-		H.last_damage_data = cause_data
-		H.apply_armoured_damage(damage * 0.5, ARMOR_BIO, BURN, "l_foot", 50)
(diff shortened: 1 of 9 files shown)


LoJoFan:
> Additionally, warrior punch will slow on hit if it fails to break bones. This punch slow will apply to synthetics and yautjas as well, alongside xenomorphs (who recieve a differing amount based on their size)

Yautja already get a ton of CC slows that make it very annoying to fight more than two good xenos. Synthetics are already vulnerable enough to warriors.

> Why It's Good For The Game

I enjoy Warrior bonebreaks for various reasons.

    They make Warrior feel like a menace, something to be afraid of.
    They uniquely interact with the human medical system in a way that I believe should be encouraged. (Stuff like applying chemicals like boiler tail stab does, it's a great way to mesh both xeno and human and add depth to the systems)
    They act as good attrition for marines and cause them to fear corners and going alone even further.

If needed, I could simply change the new minimum bonebreak damage variable to, say, 20, which will mean you can't run up to a marine and break their bones then run away. You'd have to slash them first.

With the latest warrior rework, warrior is already a very formidable opponent. In fact, in my opinion, the cooldown reduction per slash already makes it too strong.

I have well over 200 hours as warrior and I can confidently tell you I break at least 2 bones per lunge target without switching limb, and can juggle usually two or three marines at a time. Switching limbs, you could probably get at least 4.

This is a bad idea.

stanalbatross:
Warrior is enough of a menace / threat already and Marines already fear corners due to them

Little reason for this aids mechanic to come back

PR #1501, opened 2022-11-11
fixes the new teleport commands to use in world rather than item list

harryob:
## About The Pull Request

this is why you run CI again before merging btw, future new-maints


morrowwolf:
Speedmerging because master is kill. Tests checked out.

PR #1504, opened 2022-11-11
reduces the ingredient cost and improves ingredient consistency for recipes

Huffie56:
## About The Pull Request

Since creating food isn't giving any buff to marine eating it i was thinking that reducing the ingredient require to one of each would improve.
I also improve coherence by replacing flour either by bun for burger or by dough for cake pie etc...

## Why It's Good For The Game
Recipe would be more coherent so easier to learn and teach in-game.
Less ingredients require mean that cooks will be more happy.

## Changelog

:cl:
add: spesslaw require two meatball instead of five.
add: replace flour with flatbread for pumpkiepie
add: Replace Flour for bun in roburger and roburger_unsafe
add: Replace Flour for dough in cakes(orange,lemon,lime,chocolate,brain), appletart and creamcheesebread.
balance: Reduce ingredient require to one of each(example banana bread that cost three dough will only cost one dough).
/:cl:


harryob:
make sure your PR titles are descriptive as to *what* they change - as this is the message that ends up being the final commit message when it is merged

(like i've edited this one to)

ThePiachu (approved):

PR #1506, opened 2022-11-11
Update Statpanel/TGUI from upstream

fira:
## What this does

 * updates TGUI/Statpanel
 * Fixes statpanel loading
 * Port tgui_input_text 
 * Probably breaks random TGUI interfaces (need indepth check)
 
## What this doesn't do

 * Port tgui alert/number/list, backend needs additional tweaks for that (here or another MR)
 * Port tgui_say : this can be implemented separately if warranted

## Why It's Good For The Game

Maintainability, new features, less bugs

## Changelog

:cl: Fira, harryob
refactor: ported new /tg/ tgui and statpanel. this should fix statpanel loading and hopefully not break everything else.
/:cl:


fira:
Needs an indepth check of as many interfaces as possible to ensure they aren't broken

harryob:
lgtm for testmerging, only way we’re gonna be able to gather enough usage on interfaces

fira:
TM bugs addressed, there's still 3 outstanding issues to investigate 
 * MC tab availability. we modified it on CM
 * there might be issues with the tickets tab now due to changes in how the system works at /tg/. unsure
 * see if we can realistically bring back the font size setting without breaking things too bad

harryob (approved):
would like this merged quick because i'm sure it'll cause tgui conflicts

PR #1510, opened 2022-11-11
Berserker re-tweaking (NO GPB UPDATE)

usnpeepoo:
## About The Pull Request

Turns out, berserker is too powerful. It's speed with apprehend stacked with frenzy pheros, and it's normal rage speed boost which made it almost identical to a runner but even more powerful. In order to fix this, I removed the ability for berserkers to gain any frenzy aura buff, i've made the rage lock out 8 seconds instead of 6 (too damn low), i've also reduced the speed buff greatly.

## Why It's Good For The Game

Berserker shouldn't be a runner, and it's lock out proved to be too low. This PR fixes it's issues and makes it balanced.

## Changelog

balance: Berserkers can no longer use frenzy auras
balance: Berserkers speed buff has been lowered, it's rage lockout 2 seconds longer.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/ravager/ravager_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/ravager/ravager_abilities.dm
@@ -68,7 +68,7 @@
 	xeno_cooldown = 15 SECONDS
 
 	// Config values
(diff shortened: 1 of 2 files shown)


carlarctg:
cringe!

PR #1195, opened 2022-10-07
Medical APC unique features update

JeserTheFox:
## About The Pull Request

1. Added MED APC inbuilt medevac stretcher module and added support for it to dropship's medevac. Same requirements for lift as for regular stretcher, but applies to vehicle, as well as requirements for vehicle to stand still and be at >50% hull health. Medevac stretchers disable beacon after use.
2. Reduced medevac cooldown from 60 to 20 seconds.
3. Made medevac and fulton cooldowns into defines in dropships.dm
4. Added MED APC versions of crew health monitor and recharger. Functionality is identical.
5. Added MED APC version of the viewport, which provides users with medical HUD when surverying vehicle exterior, but disables user's current eye piece. Automatically enables eyepiece after user stops utilizing the viewport. 
6. Moving while using a viewport cancels out use. Viewports now provide default for multitile vehicles 1 tile view range increase.
7. Added sprites for MED APC variants of viewports, crew monitor terminal, recharger, medevac stretcher as well as exterior medevac animation, which will be activated later once vehicle visual code rework is finished.
8. Mapped in new machinery into MED APC interior map.
9. Some fixes in vehicle_unbuckle action.
10. Added handling for multitiles in resist().
11. Disabled deconstructing medevac_stretcher with a wrench
12. Added ability for xenos to use viewports.

## Why It's Good For The Game

Following the plan of each APC specialization having unique mechanics and modules, here is the update for MED APC. In theory, should make medevac more often utilized (both with MED APC present and not), meaning more work for shipside docs. MedHUD viewports should be of great help to those marines who stick with MED APC medic and act as body grabber for medic to treat. Crew monitor should improve MED APC rescue and triage role. In summary, should make MED APC an actually viable alternative choice for currently dominating transport APC. 
Also some fixes, as usual.

## Changelog

:cl: Jeser
add: Added inbuilt medevac stretcher for MED APC. Same requirements for succesful lifting as for the regular medevac stretcher, but applies to vehicle, as well as requirements for vehicle to stand still and be at more than half hull integrity. Medevac stretchers disable beacon after succesful lift.
balance: Reduced medevac cooldown from 60 to 20 seconds.
add: Added MED APC version of armored vehicles viewports, the External Medi-Cameras Terminal. Provides users with medical HUD when surverying vehicle exterior, but disables user's current eyepiece. Automatically enables eyepiece after user stops utilizing the viewport. 
[...]

PR #1513, opened 2022-11-11
Medevac Buffs

ihatethisengine:
## About The Pull Request
Reduces cooldown of medevac from 60 seconds to 20 seconds. PO no longer needs to manually activate the winch, so medevac can be operated from the cockpit. What's more, you can operate medevac by interacting with the medevac system itself, and even if you don't have the skills of a pilot, you can use it if you have the skills of a doctor (which means nurse can run medevac). And finally, the medical stretcher is now automatically activated when deployed.

I know there is a PR by jeser that reduces cooldown, but it stuck in PR hell anyway and also I changed more stuff.

## Why It's Good For The Game
Since we want to force wounded marines to go shipside, we must provide them a more convenient way to reach the Almayer. Medevac was always underutilized because it required too much coordination and unnecessary actions (e.g. having to activate medical stretcher every time, keep in mind a huge portion of the medic playerbase still has no idea you need to do this). PO had to spend their limited fly-by time (which should normally be used on firemissions) to manually activating the winch, which was always annoying. And cooldown is ridiculous. You have at best three minutes of fly-by, so that means you could use medevac only twice (remember that you needed to run to the system every time) per fly-by, which is definitely not enough.

## Changelog

:cl: ihatethisengine
balance: reduced the medevac cooldown from 60 seconds to 20 seconds.
add: anyone with the skills of a doctor or a pilot can manage the medevac by interacting with the system itself.
qol: medical stretcher automatically activates when deployed.
/:cl:


fira:
Sixty to Twenty, that's a lot no?

But more seriously i don't understand the point of having the stretcher remote activated. Won't the pilot (or someone) need to be there to triage the patient anyway?

ihatethisengine:
> Sixty to Twenty, that's a lot no?
> 
> But more seriously i don't understand the point of having the stretcher remote activated. Won't the pilot (or someone) need to be there to triage the patient anyway?

Not gonna lie, I just took numbers from jeser PR. 

If you mean the winch, well, first, usually the evacced person is already stabilized. Secondly, even if triage is required, manually activating the winch still means unnecessary actions. Why do we need extra clicks in this case?

FGRSentinel:
> Sixty to Twenty, that's a lot no?
> 
> But more seriously i don't understand the point of having the stretcher remote activated. Won't the pilot (or someone) need to be there to triage the patient anyway?

To give an idea, reducing the timer to 20 seconds will increase the medevac "throughput" (number of Marines evac'd) to about 6-9 per flight, per winch, from 2-3. This number is based entirely on how quickly the winch is operated and if the dropship has a fuel enhancer or not (which adds about a minute to the flight time). On one hand, that's a pretty big increase overall. On the other, this combined with smarter rappel management would allow shorter wait times to return to the fight.

Lafrien:
> reducing the timer to 20 seconds will increase the medevac "throughput" (number of Marines evac'd) to about 6-9 per flight, per winch

This also means that every marine picked up by Medevac is every marine not on Planetside for a good bunch couple of minutes until patched up at very least.

So while it is a Marine Medevac buff, it is also an indirect Xeno buff.

fira (approved):

Geevies:
for anyone reviewing and or merging, take a look at https://github.com/cmss13-devs/cmss13/pull/1195 first

stanalbatross (changes requested):
`balance: manual activation of the winch is no longer required for operating the medevac.`

this is not entirely desirable as it encourages the PO to completely ignore the patient they have just medevaced, who may be dying from OD / organ damage / hugged or need to be placed in a bluebag

please drop this change

ihatethisengine:
Understandable, will remove the change as soon as I get home.

ihatethisengine:
I didn't forget, just been a little busy lately. Will finish this next week

ihatethisengine:
We are so back

fira (approved):

code/__DEFINES/dropships.dm
```diff
@@ -9,3 +9,5 @@
 #define DROPSHIP_MAX_AUTO_DELAY 60 SECONDS
 #define DROPSHIP_MIN_AUTO_DELAY 10 SECONDS
 #define DROPSHIP_AUTO_RETRY_COOLDOWN 20 SECONDS
+
```
Diegoflores31, line 12: What the extra line for ?

code/modules/cm_marines/dropship_equipment.dm
```diff
@@ -922,7 +921,7 @@
 		return
 	if(!ship_base) //not installed
 		return
-	if(!skillcheck(user, SKILL_PILOT, SKILL_PILOT_TRAINED))
+	if(!skillcheck(user, SKILL_PILOT, SKILL_PILOT_TRAINED) && !skillcheck(user, SKILL_MEDICAL, SKILL_MEDICAL_DOCTOR))
```
Diegoflores31, line 924: It should be OR not AND

ihatethisengine: No, it is "if NEITHER". If both not true, then send "you can't do that" and return.

Diegoflores31: have you tested that ? i think its should be II

Diegoflores31: but if it works like that just mark this as resolved . maybe i am just reading wrong

ihatethisengine: Again, it returns TRUE if user doesn't have neither pilot skills NOR doctor skills. So if they have pilot skills or doctor skills it returns FALSE. If I change it to || it will return TRUE if user doesn't have pilot skills OR doctor skills. This, doctor won't be able to operate it as they don't have pilot skills and vice versa. And my goal is opposite.

morrowwolf (approved):
Fine for balance.

Fira says code good.

Will change things around if we get another dropship likely.

PR #1516, opened 2022-11-12
Berserker re tweakening: The rebalanced [NO GPB UPDATE]

usnpeepoo:
## About The Pull Request

Berserker is too powerful with 3 stackable buffs. This PR fixes it. Also, the rage lockout duration was way too short so i've increased it by 1 second. I've also tuned apprehend to have a shorter duration. The reason for this is that I've introduced a pheromone cap system for each possible pheros, and gave berserker a maximum capable frenzy of 2.9 (or moderate pheromone). This is to curb strong/very strong frenzy making berserkers run like a fucking mach speed train, and without the band aid fix of removing frenzy from being used on berserker all together. Huge thanks to drathek, who showed me how to do this (and pretty much did the aura capping himself).

## Why It's Good For The Game

Berserker needed fine tuning again, it's fun now but at the cost of having some pieces of its kit overtuned. Namely, the 3 stackable speed buffs from frenzy pheros, apprehend, and its base slash increase which all together made berserker look like a runner that slowed on the first hit, dealt 45 damage, and quickly racked up a ton of armor. Also, the rage lockout being so low meant that it had too little downtime, especially mid combat. 

## Changelog

:cl: Usnpeepoo, Drathek
add: Pheromones can be capped on xenomorphs, and you can cap specific pheromones at specific values
balance: Berserkers are capped to Moderate pheromones with frenzy, other pheromones are not capped
balance: Berserkers apprehend is now a second shorter.
balance: Berserkers rage lockout has been increased by a second
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
@@ -197,6 +197,7 @@
 	var/regeneration_multiplier = 1
 	var/speed_modifier = 0
 	var/phero_modifier = 0
+	var/received_phero_caps = list()
 	var/acid_modifier = 0
 	var/weed_modifier = 0
 	var/evasion_modifier = 0
@@ -293,7 +294,6 @@
 	var/warding_aura = 0
 	var/recovery_aura = 0
 	var/ignore_aura = FALSE // ignore a specific pherom, input type
-
 
 	//////////////////////////////////////////////////////////////////
 	//
@@ -943,6 +943,15 @@
 		current_aura = null
 		to_chat(src, SPAN_XENOWARNING("You lose your pheromones."))
 
+	// Also recalculate received pheros now
+	for(var/capped_aura in received_phero_caps)
+		switch(capped_aura)
+			if("frenzy")
+				frenzy_new = min(frenzy_new, received_phero_caps[capped_aura])
+			if("warding")
+				warding_new = min(warding_new, received_phero_caps[capped_aura])
+			if("recovery")
+				recovery_new = min(recovery_new, received_phero_caps[capped_aura])
 
 /mob/living/carbon/Xenomorph/proc/recalculate_maturation()
 	evolution_threshold =  caste.evolution_threshold
--- a/code/modules/mob/living/carbon/xenomorph/abilities/ravager/ravager_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/ravager/ravager_abilities.dm
@@ -68,7 +68,7 @@
 	xeno_cooldown = 15 SECONDS
 
 	// Config values
-	var/speed_buff = 0.6
+	var/speed_buff = 0.7
 	var/buff_duration = 6 SECONDS
 
 
--- a/code/modules/mob/living/carbon/xenomorph/abilities/ravager/ravager_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/ravager/ravager_abilities.dm
@@ -68,7 +68,7 @@
 	xeno_cooldown = 16 SECONDS
 
 	// Config values
-	var/speed_buff = 0.8
+	var/speed_buff = 0.75
 	var/buff_duration = 7 SECONDS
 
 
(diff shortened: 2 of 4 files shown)


Pyrodes:
it was fine the way it was before the buff. high skill floor strain for real men

usnpeepoo:
> it was fine the way it was before the buff. optional high-risk high-reward strain for real men

This fixes the problem with frenzy making berserkers go at Mach speed, and fine tuned some of the stuff I buffed

code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
```diff
@@ -197,6 +197,7 @@
 	var/regeneration_multiplier = 1
 	var/speed_modifier = 0
 	var/phero_modifier = 0
+	var/received_phero_caps = list()
```
Geevies, line 200: instead of a list, i reckon you can just make a var for each phero type and put the cap there. then, when a xeno gets its pheremone added, you can do min(phere_cap, phere_received)

code/modules/mob/living/carbon/xenomorph/mutators/strains/ravager/berserker.dm
```diff
@@ -29,7 +29,7 @@
 	R.health_modifier -= XENO_HEALTH_MOD_MED
 	R.armor_modifier += XENO_ARMOR_MOD_VERYSMALL
 	R.speed_modifier += XENO_SPEED_FASTMOD_TIER_3
-
+	R.received_phero_caps["frenzy"] = 2.9 // Moderate
```
Geevies, line 32: ideally you use the define for moderate frenzy here, instead of plain numbers

Geevies (changes requested):

LoJoFan:
> balance: Berserkers apprehend is now a second shorter, but gives a stronger speed boost

6 seconds wasn't short to begin with. 5 seconds isn't short either. This looks like a second shadowbuff seen how the apprehend is already faster than a marine with light armor when off weeds (literally undodgeable and inescapable once the super slow hits).

Proposed changes to apprehend: 

- lasts four seconds
- has the current speed
- speed boost fades one second after hitting the apprehend slash IF you hit it. Otherwise lasts four seconds.

> balance: Berserkers are capped to Moderate pheromones with frenzy, other pheromones are not capped

What about frenzy fruits?

code/modules/mob/living/carbon/xenomorph/mutators/strains/ravager/berserker.dm
usnpeepoo: theres no moderate frenzy define, the numbers are quite literally what defines moderate frenzy (or any other frenzy for that matter)
2.0 to 2.9 is moderate tier

code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
usnpeepoo: yeah i doubt a lot of xenos would have limited pheros ima keep it as is

code/modules/mob/living/carbon/xenomorph/abilities/ravager/ravager_abilities.dm
```diff
@@ -65,11 +65,11 @@
 	action_type = XENO_ACTION_CLICK
 	ability_primacy = XENO_PRIMARY_ACTION_1
 	plasma_cost = 0
-	xeno_cooldown = 17 SECONDS
+	xeno_cooldown = 16 SECONDS
```
stanalbatross, line 68: keep the cooldown at 17 (i.e. increase by 1 second)

stanalbatross (changes requested):

code/modules/mob/living/carbon/xenomorph/abilities/ravager/ravager_abilities.dm
usnpeepoo: xir yes xir

stanalbatross (approved):

Skatuk:
How about removing the heal. Its already powerful with out it.

Issue #1517, opened 2022-11-12
Battlefield Executions broke (lol)


hotimep:
## Testmerges:

LV-522

## Description
BATTLEFIELD EXECUTIONS DO NOT WORK!!!! I CANNOT PERMAKILL MARINES!!!!

## Difference between expected and actual behavior
There's supposed to be a windup when you BE someone, (harm intent, aim at head, the windup happens).
What happens right now is that the game just acts if you're PBing someone

## Steps to reproduce
1. Load mateba, works with deagle too probably (awesome)
2. Check if high-impact ammo is loaded (it should be)
3. Aim for the head
4. Try and BE (it doesn't happen)

## Issue bingo (not required to be all checked)

- [x] Issue could be reproduced at least once
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #1518, opened 2022-11-12
Fixes battlefield executions

morrowwolf:
## About The Pull Request

We have just gone beyond the byond hard limit of 24 bitshifts which caused highimpact rounds to no longer function properly which broke mechanical battlefield executions. And of course this is CM so we had an unused flag that I removed and removed what was associated with it as a bandaid fix. At some point we'll need to do something about the ammo flags.

closes https://github.com/cmss13-devs/cmss13/issues/1517

## Why It's Good For The Game

BUG BAD UNGA

## Changelog

:cl: Morrow
fix: Fixed mechanical battlefield executions.
del: Deletes unused silenced ammo flag.
/:cl:

### DIFF
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -20,25 +20,24 @@
 #define AMMO_SNIPER				(1<<6)
 #define AMMO_ANTISTRUCT			(1<<7) // Primarily for railgun but can be implemented for other projectiles that are for antitank and antistructure (wall/machine)
 #define AMMO_SKIPS_ALIENS 		(1<<8)
-#define AMMO_IS_SILENCED 		(1<<9) //Unused right now.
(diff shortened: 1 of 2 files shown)


harryob (approved):

hotimep:
fixes #1517

fira:
That's why you shouldn't even use these bitfields for most behaviors and i turned down a couple MRs in the past about it x((((
Most of these behaviors can be refactored as signals/components or inheritance, you don't need a GLOBAL FLAG for a snowflake behavior like THIS ONE...

harryob:
> That's why you shouldn't even use these bitfields for most behaviors and i turned down a couple MRs in the past about it x((((
> 
> Most of these behaviors can be refactored as signals/components or inheritance, you don't need a GLOBAL FLAG for a snowflake behavior like THIS ONE...

so true so true!!!

PR #1524, opened 2022-11-12
Forced Punctuation

Geevies:
## About The Pull Request

Forces characters to speak with punctuation at the end of their sentences, lends to a higher roleplay atmosphere.

## Why It's Good For The Game

I'm not entirely sure why this was a preference in the first place, but this makes RP go hard.

## Changelog

:cl:
del: Removed the automatic punctuation preference, you always punctuate now.
/:cl:

### DIFF
--- a/code/__DEFINES/mode.dm
+++ b/code/__DEFINES/mode.dm
@@ -83,7 +83,7 @@
                                                    // MUTUALLY EXCLUSIVE TO TOGGLE_AUTO_EJECT_MAGAZINE
 #define TOGGLE_EJECT_MAGAZINE_TO_HAND		(1<<6) // This toggles whether manuallyejecting magazines from guns will cause you to unwield your gun
                                                    // and put the empty magazine in your hand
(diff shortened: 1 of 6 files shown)


stanalbatross (changes requested):
This adds absolutely nothing to roleplay and will just make everyone sound robotic when they speak

Why would we take away options from players?

ihatethisengine:
> This adds absolutely nothing to roleplay and will just make everyone sound robotic when they speak
> 
> Why would we take away options from players?

How does a player's illiteracy positively affect roleplay? We're already forcing a capital letter at the beginning of the message, what's the problem with forcing a dot at the end of the message?

SexmasterWasTaken:
> How does a player's illiteracy positively affect roleplay? We're already forcing a capital letter at the beginning of the message, what's the problem with forcing a dot at the end of the message?

There's also a preference to capitalise letters at the start of a message, it's just turned on by default. All you need to do is the same with this. Forcing it would be annoying.

PR #1526, opened 2022-11-13
Removes PL Pamphlet + PL Squad Skills

Misttyy:
## About The Pull Request

This removes the 45 point powerloader skill pamphlet from squad vendors and gives all squad roles powerloader skill basic, letting them carry floodlights, assist with setting up the DS or etc.

## Why It's Good For The Game

It's a neat little thing that popped up into my head. From what I remember, powerloader skills were restricted due to them being able to be used in combat(?). Powerloader skills are rather useless in combat right now, at least, and it doesn't seem necessary to lock them for squad roles. Req still retains a pamphlet for the odd scenario. Should be nice for the odd rp scenario, and there's enough powerloaders on map if anyone is a shithead with them.

https://youtu.be/YPMk-EEyOpE

## Changelog

:cl: Mistfrag
del: Removed powerloader skill locks for squad roles (Pfc-SL), removes the pamphlets from their vendors.
:cl:


Geevies (approved):
code's fine. i think the main concern here is a stupid PFC stealing a powerloader for shits and giggles, but i don't think that's gonna be an issue. if you're up for it, you can steal the powerloader improvements i made in the logistics truck PR to make this less of an issue

stanalbatross:
Marking as DNM for internal discussion as I don't think this is a good idea (powercreep, streamlining, and negates the importance of roles with power loader skills)

Nanu308:
Some good arguments were brought up against it which is why i will be revoking the approval and closing the PR. 

It has already been propgated down the line away from Engineering and Req which should be the primary department/s.

Tldr. too much streamlining and allowing PFC's to easily do req etc. jobs.

PR #1527, opened 2022-11-13
Ghost Hijack Timer

fira:
## About The Pull Request

Displays to ghosts in status pane when hijack will end.

image

By request of podrick

## Why It's Good For The Game

Transparency

## Changelog

:cl:
qol: Ghosts now have a timer displaying time left to hijack end in Statpanel.
fix: Evac timer in Status pane will now display as such.
/:cl:

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -944,10 +944,20 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 		if(client.admin_holder)
 			. += "Players Ready: [SSticker.totalPlayersReady]"
 
+	. += ""
+
+	if(SSticker.mode?.force_end_at)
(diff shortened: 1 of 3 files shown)


ihatethisengine:
> . += "Evacuation: [eta_status]"

Make it clearer for humans as well I guess

fira:
> > . += "Evacuation: [eta_status]"
> 
> Make it clearer for humans as well I guess

Yeah good point. The timer displaying on its own is kinda stupid

fira:
even fixed it for the legacy AIs we don't use!!! ain't that nice?

Geevies (approved):
if you're feeling really amped, you could add the total time before a hijacked ship lands as well. i've always wanted that for playing music during hijack

fira:
Oh, would rather do it separately cause it's a bit less straightforward, do remind me on occasion

PR #1529, opened 2022-11-13
Added unique animations for tail stab, sharp and blunt.

carlarctg:
## About The Pull Request

Title.

Defenders when tail slamming will have their direction turn to a random side to imply spinning their tail into the target's face.

Other castes will instead look backwards for to show their tail extending out and stabbing.

Defenders also now have a custom on-target slash animation for tail slam.

If the caste hasn't moved in 0.5s since the tail stab, their direction will revert back to the pre-stab direction. This looks super awesome.

## Why It's Good For The Game

Cool effects good.

## Changelog

:cl: carlarc, DrMacCool
add: Defenders when tail slamming will have their direction turn to a random side to imply spinning their tail into the target's face.
add: Other castes will instead look backwards for to show their tail extending out and stabbing.
add: Defenders also now have a custom on-target slash animation for tail slam.
/:cl:


Geevies (commented):
not so sure about the sharp tail stab flipping the xeno around, the tails are pretty long so i figure they'd keep looking at the target they're stabbing. also i don't think it's necessary to automatically reset the dir, since the player might be moving into the new direction (fleeing) after making the strike. i think the player can manually manage the dir

carlarctg:
> not so sure about the sharp tail stab flipping the xeno around, the tails are pretty long so i figure they'd keep looking at the target they're stabbing. also i don't think it's necessary to automatically reset the dir, since the player might be moving into the new direction (fleeing) after making the strike. i think the player can manually manage the dir

it only resets the dir if they're still in the same direction, the effect looks sick ill post a vibeo soon

carlarctg:
https://streamable.com/2c6pz9
https://streamable.com/7z56zh see it looks sick

Geevies (approved):
still not sure about the dir thing, since it'll turn you back around if you start moving away in the direction the tail stab moved you, but that's probably ok

kugamo (approved):
Sprite LGTM, approved.

PR #1537, opened 2022-11-13
Improved Hijack Atmosphere

Geevies:
## About The Pull Request

Improves hijack atmosphere by making the initial crash more violent and visceral with additional crashing sounds, and makes the on-ship fight more dire by making random pipes explode and cause fires on the ship.

## Why It's Good For The Game

Hijack is actually fairly... chill? You're just fighting xenos on the ship. The initial crash is tense, but after that it's just another map without any hazards. This adds some hazards to the ship, and makes it feel more like a critically damaged ship the marines NEED to evacuate from.

## Changelog

:cl:
add: Improved hijack atmosphere, during the initial crash, and subsequent battle.
/:cl:

### DIFF
--- a/code/game/area/almayer.dm
+++ b/code/game/area/almayer.dm
@@ -13,6 +13,8 @@
 	ambience_exterior = AMBIENCE_ALMAYER
 	ceiling_muffle = FALSE
 
+	var/hijack_explosive_immune = FALSE
+
 /area/shuttle/almayer/elevator_maintenance/upperdeck
 	name = "\improper Maintenance Elevator"
 	icon_state = "shuttle"
@@ -690,6 +692,7 @@
 	name = "\improper Lifeboat Docking Port"
 	icon_state = "selfdestruct"
 	fake_zlevel = 1 // upperdeck
+	hijack_explosive_immune = TRUE
 
 /area/space/almayer/lifeboat_dock
 	name = "\improper Lifeboat Docking Port"
@@ -703,6 +706,7 @@
 	icon_state = "shuttle2"
 	requires_power = 0
 	flags_area = AREA_NOTUNNEL
+	hijack_explosive_immune = TRUE
 
 //Placeholder.
 /area/almayer/evacuation/pod1
--- a/code/game/gamemodes/colonialmarines/colonialmarines.dm
+++ b/code/game/gamemodes/colonialmarines/colonialmarines.dm
@@ -12,6 +12,9 @@
 	var/research_allocation_interval = 10 MINUTES
 	var/next_research_allocation = 0
 
+	COOLDOWN_DECLARE(last_hijack_explosive_barrage)
+	var/time_between_hijack_explosive_barrages = 5 SECONDS
+
 ////////////////////////////////////////////////////////////////////////////////////////
 ////////////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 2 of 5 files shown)


LoJoFan:
damaged life support =/= random atmospherics explosions. fire would also not spread (on account of everything being made of metal)

i don't think either party would enjoy randomly exploding

TopHatPenguin:
I wonder if it would be possible to mess with the layered grilles that appear in maint etc and reuse those around the designated crashsite to display that parts of the ceiling have collaped/are collapsing. Additionally something to add on is just generally bits of debris falling into the hallway and blocking it due to the crash (you need to break through it). 

Both have some potential balance issues though. The former being that the layered ceiling grilles could potentialy obsfucate sprites and get used to hide under from the camera view. The latter being debris fall in front of you.

harryob:
> damaged life support =/= random atmospherics explosions. fire would also not spread (on account of everything being made of metal)
> 
> i don't think either party would enjoy randomly exploding

“randomly” - telegraphed with a five second warning. you’ll be fine

fira (approved):

JohnnyRum:
I imagine part of the code is dedicated to identifying where marines are holding and excluding those areas? Because these fires, even if small and telegraphed, have the potential to harm the marines greatly if they occur within the cade squares built to defend the lifeboats for example, which rarely are bigger than 5 by 5.

fira:
The lifeboats themselves and i believe dropships too are considered Space/Shuttles and distinct from Almayer so exempt from this logic. In theory fire won't appear there ,though it could appear next to and clip inside

fira (changes requested):
Retracting approval for now due to poor performance when tested in the process() chunk slowing down hijack a lot, we need to find a solution about it. Either pre-calc list of turfs and reuse it, or roll coordinates directly, for example, or whatever else lessens the load

Opinion only but the incoming/crash sounds are realllly loud

hotimep:
what if you made lights on the ship flash red

Herotitan772:
This is essentially a marine nerf which after hijack is I think we can all agree is not needed as marines are almost always fucked no matter what they do. It won’t effect xenos 95% of the time because they are on the offensive, they are just faster than marines and can move away from the explosion with ease, they also arnt as tightly packed as marines and xenos unlike marines arnt mega fucked by getting set on fire, best case scenario is a marine after getting set on fire is loosing 3/4 of their health and needing to see a medic and sit back for a minute or so, worst case you die, get organ damage, IB and 650 burns (which at hijack means you are essentially perma) whereas a xeno who has significantly more health than a marine can survive for a very long time being on fire (we have all seen woyers literally strolling trough fire to drag some poor marine to his very warm death).

So an example; marines are holding lifeboats there are a hundred tiles I can think of that a fire bomb going off would be catastrophic to Marines, either blocking retreat paths (they might not even know the “pipe” is “bursting”), destroying Cade lines, landing in the middle of a bunch of marines (which will certainly cause some people to die), and the worst possible place for it to land would be……… in the fucking supply pile where the ammo and explosives is, so now marines have a choice they can either not bring supplys to the boats, sort through the supplys they do have and remove everything…. **Combustible** or they can choose to possibly have their entire supply cache, all the support staff behind the cades, half the marines and a Cadeline or 2 just disappear in a very fun random detonation.

Maybe if the lifeboats area is made immune to this pipe explosion it could work but would still just be a marine nerf

Whilst I do agree hijack is pretty boring and a waste of potentially up to 45 minutes I don’t think this is the way to fix it

hotimep:
ok ask geeves if you want to make hijack 1000x% more soulful and atmospheric make the ship shake every so often during hijack

LoJoFan:
> The lifeboats themselves and i believe dropships too are considered Space/Shuttles and distinct from Almayer so exempt from this logic. In theory fire won't appear there ,though it could appear next to and clip inside

Knowing this, it would remove interactivity from hijack and make nothing feasible but to evacuate or hold lifeboats or evacuation pods. The rest of the ship becomes unusable, i.e. no more briefing, CIC, req holds

The next logical step is to hold hijack to a vote if it's going to be the same thing repeated over and over again.

I have some ideas about how to make hijack "more atmospheric" without people randomly dying to pipes but I would be told WYCI.

Huffie56:
The sounds even if a bit loud where are great addition to the atmosphere and being thrown in the air if not buckle add to it.
Maybe make a separate pull request for the sounds and the extra effect so we can enjoy them now ?
Also DS crashing is already extremely slow every time i crash as xeno we have to wait 10-30s before the crash effect as finish don't want it to take 1 more min.... 
Maybe you can gain performance to add the effect if you refactor DS crashing code?
thanks for the addition feel great to not have the same boring hijack.

Huffie56:
After multiple round i found the number of explosion around the ship to be way way to many. can you please tune it down like divided them by three.
thanks

PR #1541, opened 2022-11-13
Spitter Rework

carlarctg:
## About The Pull Request

Design doc:

https://hackmd.io/@waltuh/Sy0A1SnEo
Slightly inaccurate as my brainworms enticed me to change and add mini features.

Approved by Walter, ignorepproved by Gevonius and Satanbatros

https://cdn.discordapp.com/attachments/280051925154660363/1041475609165045770/image.png

Changes:
- Spit doesn't spatter by default, instead it's now a faster, weak 7-tile* projectile that deals 20 damage with a faster spit cooldown. Fully accurate at 6 tiles, inaccurate at 7 tiles but can sometimes hit.
- Frenzy replaced with 'charge spit'. Halved speed buff, added +5 armor buff, the next spit will deal 10 more damage and coat humans in acid but have only 5* tiles of range.
- Acid spray halves damage every time someone walks on it. However, its damage is spread over legs and feet, and if someone who is spattered with acid is hit by it, their acid spatter will be strengthened, dealing more damage, lasting longer, and needing two rolls to clear up. Also, the bonus damage didn't actually work, now it does.

* Projectile range code is SHIIIT and breaks on diagonals so the range variable is increased.

Also, queen acid spit spatters and has 1 second less extra cooldown because find and replace caused it and I didn't think it was a change worth removing. It's neat, maybe they'll actually use it now.

Added support for balloonchat colors. (Even TG doesn't have this, we're awesome now!)

Renamed vision_distance parameter to max_distance so it's similar to visible_message

## Why It's Good For The Game

As stated in the hackmd, spitter is very ineffective without using the oversight-exploit infinite acid spray spam, and its combo (acid spatter into spray) does not actually help, as the stun stops the human from getting further hit by the spray and the bonus damage does not actually apply thanks to shitcode. This PR makes it so that the combo is indeed more effective than making humans walk into the spray.

Again as stated, spitter suffers from a strange issue where it's actually not good at harassing from range despite that being its purpose, since it has a low range. By letting it be long ranged by default and choose to go short range, it adds a lot of depth and versatility and lets it actually harass marines.

As always they can just. Shoot it to make it go away. 

## Changelog

:cl:
refactor: Added support for balloonchat colors. (Even TG doesn't have this, we're awesome now!)
add: Spitter rework!
add: Spit is now full screen range but weaker.
add: Frenzy is renamed and causes spit to inflict spatter coating.
add: Acid spray's damage is halved every time it hits a human, but if it hits someone coated in acid it will enhance it, making it more dangerous and need two rolls to shake off.
/:cl:


fira:
City? Urban maps confirmed???

kugamo (approved):
Sprite changes are fine, approved.

code/datums/effects/weak_spray_stack.dm
```diff
@@ -0,0 +1,24 @@
...
+
+	. = ..()
+
+/datum/effects/weak_spray_stack/Destroy()
+	if(affected_atom)
+		LAZYREMOVE(affected_atom.effects_list, src)
+
+	if(ishuman(affected_atom))
+		var/mob/living/carbon/human/H = affected_atom
+		to_chat(H, SPAN_WARNING("You feel the last of the weak acid coating on your legs dissipate."))
```
harryob, line 22: u know we gotta do it to them what is a H?????

code/game/objects/effects/aliens.dm
```diff
@@ -193,49 +193,58 @@
...
 		var/damage = damage_amount
 
-		var/should_stun = FALSE
-		for (var/datum/effects/acid/A in H.effects_list)
-			should_stun = TRUE
-			damage += (-1*(A.duration - A.original_duration))*(A.damage_in_total_human/A.original_duration)
+		var/buffed_splash = FALSE
+		var/datum/effects/acid/A = locate() in H.effects_list
+		if(A && A.acid_enhanced == FALSE) // can't stack the bonus every splash. thatd be nuts!
+			buffed_splash = TRUE
```
harryob, line 207: you touched too much of this code (it appears entirely in green) to not have to change this var name

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1583,9 +1583,13 @@
 
 	sleep(sleep_amount)
 
-	visible_message(SPAN_DANGER("[src] has successfully removed the acid!"), \
-			SPAN_NOTICE("You get rid of the acid."), null, 5)
-	extinguish_acid()
+	if( extinguish_acid() )
+		visible_message(SPAN_DANGER("[src] has successfully removed the acid!"), \
+				SPAN_NOTICE("You get rid of the acid."), null, 5)
```
harryob, line 1588: i'm declaring war on null args
you WILL max_distance = 5

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1583,9 +1583,13 @@
...
+	else
+		visible_message(SPAN_DANGER("[src] has managed to get rid of some of the acid!"), \
+				SPAN_NOTICE("You manage to get rid of some of the acid... but it's still melting you!"), null, 5)
```
harryob, line 1591: and here

code/modules/mob/living/carbon/xenomorph/abilities/spitter/spitter_powers.dm
```diff
@@ -8,30 +8,46 @@
...
 	addtimer(CALLBACK(src, .proc/remove_effects), duration)
 
 	apply_cooldown()
 	..()
 	return
 
-/datum/action/xeno_action/onclick/spitter_frenzy/proc/remove_effects()
+/datum/action/xeno_action/onclick/charge_spit/proc/disable_spatter()
+	SIGNAL_HANDLER
+	var/mob/living/carbon/Xenomorph/X = owner
```
harryob, line 37: uh oh

code/modules/mob/living/carbon/xenomorph/castes/Spitter.dm
```diff
@@ -60,7 +60,7 @@
 
 /datum/behavior_delegate/spitter_base
 	name = "Base Spitter Behavior Delegate"
-
+/*
```
harryob, line 63: must we keep dead code in the codebase

harryob (changes requested):

code/modules/mob/living/carbon/human/human.dm
Diegoflores31: to be fair thats a 3 LETTER var 😸

code/modules/mob/living/carbon/xenomorph/castes/Spitter.dm
Diegoflores31: 3:)

code/game/objects/effects/aliens.dm
carlarctg: don

code/modules/mob/living/carbon/xenomorph/abilities/spitter/spitter_powers.dm
carlarctg: don

code/modules/mob/living/carbon/xenomorph/castes/Spitter.dm
carlarctg: don

code/modules/mob/living/carbon/human/human.dm
carlarctg: fixed

harryob (approved):
woahoahoaoh i'm going crazy i'm gonna do it i'm gonna

PR #657, opened 2022-07-11
anti CL powergame

stanalbatross:
## About The Pull Request

stops CLs from giving marines le op gear predrop 

## Why It's Good For The Game

turns out being able to give one guy pills that make him super fast are a bit unbalanced, especially doing so predrop

this should make it so that you have the chance to roleplay a bit more before spamming out your op stim pills

## Changelog

:cl:
add: added green pens, the CL gets a green pen in their briefcase
balance: upped the price of ultrazine in the CL briefcase to 25 points
balance: you can no longer buy ultrazine or neurostimulator implants from the CL briefcase before 12:25
code: made the code for update_icons on pens much nicer
add: added DCC spawn landmarks to the uss runtime
/:cl:

PR #1543, opened 2022-11-14
CL Breifcase Bugfix

PalJohnson:
## About The Pull Request

Fixes timer issue with briefcase accessibility. In short the previous PR #657 sought to limit items from being dispensed prior to marine drop to discourage power gaming. The timer is supposed to make it so you cant access certain items by 12:25 (20min in game). The fix applied a value of 25min which in theory should work but the server start time is 12:05. This means items cannot be retrieved until 12:30.

## Why It's Good For The Game

Fixes oversight caused by server time beginning 5min.

:cl:
fix: fixed timer on CL briefcase not dispensing items at 12:25 server time
/:cl:


fira:
This isn't the correct way to fix this. Server start might be anytime.

fira (changes requested):
To fix this properly use `SSticker.mode.round_time_lobby`

fira:
Would like a clarification on that by original author @stanalbatross

stanalbatross:
It's meant to be at 12:25, if the timer doesn't take preround time into account then this is fine

fira (approved):
That's fine then

PR #1561, opened 2022-11-15
internal mortar camera console fix v2 + VV option for toggling needing APC power for machinery

### LINKED ISSUE #1163: Mortar Camera not working!
## Description
The M402 Mortar's View Camera function is not working in any form or function. Clickin on the button results in no windows prompt, with or without flare shells fired.

When clicked, the View Camera function is meant to pull up a separate, security camera like window that will show any current flare camera shells on the ground, giving a view for the mortar operator in order to optimize shooting in areas that will require long term fire support. Currently, the button will do absolutely nothing, in both WO and standard map rotations and game modes.  Font also seems to be off, may not be apart of the bug.

## Steps to reproduce
1. Acquire a M402 Mortar within the current merge branch.
2. Set up previously mentioned mortar.
3. Click on the view camera function.
4. Fire a flare shell in desperation. 
5. Click on the view camera function once more to have nothing happen.
6. Reconnect.
7. Click on view camera to once again have nothing happen.

- [ x ] Issue could be reproduced at least once
- [ x ] Issue happened in a recent (less than 7 days ago) round
- [ x ] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

NoKohi:
## About The Pull Request

this changes and expands upon the "needs_power" variable to machinery, by default this is set to TRUE on machinery (nothing is changed here); if this is set to FALSE, then the machinery in question will work in an unpowered area as if though the power was on for it;

this adds a VV option "toggle needing power" for machinery, which toggles the needs_power var on/off ands call the power_change proc on it, so that it functions accordingly

this changes stat checks in the autolathe to calling inoperable and the base machinery attack_hand proc, this brings it line with other things in-game

## Why It's Good For The Game

this was originally supposed to fix #1163, but this changes the currently merged implementation of this variable (which fixed #1163), so that it functions in a less wonky way and avoids having any machinery with the variable set to 0 appearing to be unpowered visually and when examining, despite actually being functional

admins can now easily set up machines in unpowered areas for events, e.g. autodocs in caves, by using the VV option

I was informed something like this would be the preferred solution to fixing the internal mortar camera console

## Changelog

:cl:
admin: adds a VV option for "toggle needing power", this toggles a machine needing APC power
/:cl:


fira (approved):
i prefered the other solution but hey, it works

morrowwolf:
Try the thing we talked about in discord.

fira:
i merged ihate2's PR on account of it being a simpler fix to the mortar camera problem itself, but this still leaves two adjacent issues that aren't addressed, if you feel they're in scope of this PR as we disccused before:
 * power updates when changing via VV, to ensure icon updates
 * the autolathe that should use inoperable()

NoKohi:
also why was his fix merged and my initial fix wasn't
i feel cheated :(

fira (approved):

fira:
it was just simpler for the purpose of mortar only, so made more sense to merge immediately rather than waiting. then we can address the rest here

PR #1563, opened 2022-11-15
Mousetraps now hurt, a lot.

carlarctg:
## About The Pull Request

Mousetraps now hurt mobs depending on their size. Facehuggers will take enough damage to be squashed instantly, Larvae will just barely survive after a 1-second stun.
Humans will scream in pain and take a tiny amount of damage, so will Runners. Normal xenos are barely affected, and anything that can't be slugged is completely immune. Massive xenomorphs instantly crush the trap.

Increased larvae health by 5 because, let's be honest, it won't impact anything and lets me differentiate between huggers and larvae without unnecessarily nerfing the former.

## Why It's Good For The Game

It would be INCREDIBLY funny to set up mousetraps around the FOB or other transited areas to make facehuggers instagib. This is the entire reason for this change.

Additionally, it's just fun! It's a little of that old, worn-out SS13 charm that we so sorely lack and need. Imagine MTs setting up mouse traps around maintenance to catch unaware runners, or around medbay to not let larva escape. 

You may think the latter is a possible issue with the PR, but i SINCERELY doubt this will be done even one in a hundred rounds, so when it happens it will be just unexpected and silly enough to be worth it. If someone starts placing them every round, we can change it then, and not before.

## Changelog

:cl:
add: Mousetraps when stepped on now hurt ranging from a LOT to very little depending on the mob's size.
/:cl:


Huffie56:
Here is what i think:
Humans should be immune to some or all the damage if they wear shoes or better boots like for glass shard.
maybe just make them scream.
Also change name and sprite and make a different item for it to be more lore friendly.(mouse trap killing huger is a meme)
Pretty sure you can find someone to reskin mouse trap on discord.
Please consider to change this.
Thanks for your contribution.

LoJoFan:
the "5 hp larva buff" will make a huge difference in whether one bullet from a weapon will crit them or not

stanalbatross:
>when it happens it will be just unexpected and silly enough to be worth it

no, not really, it'll be annoying and frustrating for the player on the recieving end

>If someone starts placing them every round

which they will

>we can change it then, and not before

or I can deny this now and save us all the hassle

PR #1548, opened 2022-11-14
Xeno blood chemistry

ThePiachu:
## About The Pull Request

I was playing a bit of Research lately and noticed two problems. First, Marines don't retrieve enough xeno corpses because they don't see the benefit of doing so most of the time. Secondly, when Research actually gets xeno corpses they usually take a small sample to analyse for points and dumps the corpses without using up all the blood, which felt like a waste. So I figured I'd try giving Marines a more tangible reason to go fetch xeno corpses - stims and more explosives!

First, background info. Every Xeno has 560 units of blood in them which is usually split between their Acidic Blood and a Plasma (only exception being hugger, chestburster, Spitter and Trapper Boiler with only acidic bloods). When you have a pheromone in a xeno, you usually get 280 units of it. Here is [a spreadsheet covering what you can get from which caste](https://docs.google.com/spreadsheets/d/1xKf9r0YFWXPomOM69Ka9JqzXOL7FJYfK2UB_BBTm3jE/edit?usp=sharing). Exceptions are Healer Drone, Hivelord, Praetorian and Queen that have more diluted pheromones, with the last one having 112 units of 3 different ones. So our working number is 280u acid blood and one plasma.

IV Drips take 4 units of liquid out per tick, and you can take a bit more out manually with syringes. It will take a few minutes to process each corpse entirely, even with multiple IVs. You are taking out acidic blood and plasma at the same time, so basically acidic blood is a waste product you give to engineering for them to have some fun.

Xeno stims I based off of the SOF stims - they consume 1 unit per 1 minute, and usually come with a dosage of 5, OD of 15, critical of 25. If you take more than one stim you burn through them quicker. My first draft takes 10 units of plasma and combines it with two other chemicals to make 1 unit of stim. So one corpse gives you 28 minutes of a stim, which I'd expect to be almost 6 5 minute doses depending on how Researchers splice them. Of course, the number can be tweaked, but unless we get some faster way of exsanguinating xenos, the amount of stims you get from the blood might be okay. Will need to test to see how it performs, but it should keep a few researchers busy.

Stims are as follows:
- Purple Plasma Stim (from all drone strains) - heals all kinds of damage (on the level of basic med per type), speeds user up a bit, repairs mechanical limbs, creates some blood, stops a bit of pain. Generally, weaker than Unga Juice but still neat.
- Pheromone Plasma Stim (Hivelord, Prae, Healer Drone, Queen) - heals all kinds of damage even better, good painstopper, creates blood, stops bone breaks.
[...]

PR #1565, opened 2022-11-15
Xeno blood is now finite

ThePiachu:
## About The Pull Request

While working on #1548 I got a comment how Xeno corpses produce infinite blood. That's obviously something that needs to be fixed if we want to do anything with Xeno blood products in the future (whether it's stims, exotic materials, equipment upgrades and so on). So I made the xeno blood actually tick down when you draw it out, and just to be sure I added a handle_blood() proc to xeno's Life() to make them regen blood while they are alive. You never know what ungas might do with a syringe...

Also fixed an edge case where you couldn't get the last drop of blood out of a mob if you were drawing too much. If that's the case we just take what's left.

## Why It's Good For The Game

If we want xeno blood to be useful for future development, it can't be a limitless resource.

## Changelog

:cl:
fix: Xeno blood is now a finite resource.
add: Living xenos now regenerate blood just in case some unga has the idea to attack one with a syringe.
/:cl:

### DIFF
--- a/code/modules/mob/living/blood.dm
+++ b/code/modules/mob/living/blood.dm
@@ -136,7 +136,7 @@
 		return
 
 	if(blood_volume < amount)
-		return
+		amount = blood_volume
 
 	var/b_id = get_blood_id()
 	if(!b_id)
@@ -160,6 +160,9 @@
 	if(!O.reagents)
 		return
 
+	if(blood_volume < amount)
+		amount = blood_volume
+
 	var/b_id = get_blood_id()
 	if(!b_id)
 		return
@@ -171,6 +174,7 @@
 	for(var/plasma in plasmas)
 		O.reagents.add_reagent(plasma,amount / plasmas.len) //An even amount of each plasma and blood type
 
+	blood_volume = max(0, blood_volume - amount) // Removes blood if human
 	return 1
 
 
--- a/code/modules/mob/living/blood.dm
+++ b/code/modules/mob/living/blood.dm
@@ -138,7 +138,7 @@
 
 //Gets blood from mob to the container, preserving all data in it.
 /mob/living/carbon/proc/take_blood(obj/O, amount)
-	if(!O.reagents || amount <= 0)
+	if(!O.reagents || amount <= 0 || blood_volume <= 0)
 		return
 
 	if(blood_volume < amount)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/blood.dm
```diff
@@ -171,6 +180,7 @@
 	for(var/plasma in plasmas)
 		O.reagents.add_reagent(plasma,amount / plasmas.len) //An even amount of each plasma and blood type
 
+	blood_volume = max(0, blood_volume - amount) // Removes blood if human
```
stanalbatross, line 183:
```suggestion
	blood_volume = max(0, blood_volume - amount)
```
this comment is wrong and so is its use in /carbon

code/modules/mob/living/blood.dm
```diff
@@ -160,6 +166,9 @@
 	if(!O.reagents)
 		return
 
+	if(blood_volume < amount)
+		amount = blood_volume
```
stanalbatross, line 170: should probably sanity check for blood volume here so you don't go through the entire proc with `amount` being zero, same for /carbon

stanalbatross (changes requested):

code/modules/mob/living/blood.dm
```diff
@@ -132,11 +138,11 @@
 
 //Gets blood from mob to the container, preserving all data in it.
 /mob/living/carbon/proc/take_blood(obj/O, amount)
-	if(!O.reagents)
+	if(!O.reagents || amount <= 0)
```
stanalbatross, line 141:
```suggestion
	if(!O.reagents || amount <= 0 || blood_volume <= 0)
```

code/modules/mob/living/blood.dm
```diff
@@ -157,9 +163,12 @@
 	. = ..()
 
 /mob/living/carbon/Xenomorph/take_blood(obj/O, var/amount)
-	if(!O.reagents)
+	if(!O.reagents || amount <= 0)
```
stanalbatross, line 166:
```suggestion
	if(!O.reagents || amount <= 0 || blood_volume <= 0)
```

stanalbatross (changes requested):

fira (approved):

harryob (approved):

PR #1575, opened 2022-11-15
Fixes Defender Tail Sweep knockdown, fire/acid resist knockdown and vehicle collision knockdown

TheGamerdk:
## About The Pull Request

The stun refactor thingy forgot to include the force = TRUE for this (checked the others, might've missed some)

This all meant that
1. Defenders tail sweep no longer knocked down marines
2. Marines and Predators that resisted out of acid/fire didn't get knocked down
3. Xenos that resisted out of fire didn't get knocked down
4. People hit by vehicles didn't get knocked down
5. Some predator weapon didn't knockdown xenos correctly

(Feel free to fix this mess, the flags on the effect proc didn't make much sense)

## Why It's Good For The Game

Bugs are bad

## Changelog

:cl:
fix: Defenders can now knockdown with tailsweep once again!
fix: You're now once again knocked down when resisting fire or acid. In addition you're once again knocked down when hit by the APC
/:cl:


Segrain:
Is the flag really necessary? The problem is that arguments are in the wrong order. Does fixing that alone not suffice on its own?

TheGamerdk:
> Is the flag really necessary? The problem is that arguments are in the wrong order. Does fixing that alone not suffice on its own?

While not knowing this codebase I’m going out on a limb and assuming that the force argument isn’t just replaceable, it depends ojb if CANKNOCKDOWN is permanently set
See https://github.com/cmss13-devs/cmss13/blob/4c17f8d3d57b61735baa2f4de8e2bf18ba6a564d/code/modules/mob/mob_status_procs.dm#L132 for reference

TheGamerdk:
> > Is the flag really necessary? The problem is that arguments are in the wrong order. Does fixing that alone not suffice on its own?
> 
> While not knowing this codebase I’m going out on a limb and assuming that the force argument isn’t just replaceable, it depends ojb if CANKNOCKDOWN is permanently set See
> 
> https://github.com/cmss13-devs/cmss13/blob/4c17f8d3d57b61735baa2f4de8e2bf18ba6a564d/code/modules/mob/mob_status_procs.dm#L132
> 
> for reference

turns out CANKNOCKDOWN is always set, bar admin intervention.
Flag is unneeded, arguments just need to be unscrambled

PR #1580, opened 2022-11-16
Take/apply mob damage + armor vs armour sanity refractor

PhantomEpicness:
## About The Pull Request
Standardizes mob damage procs to apply_ rather than some of them being take_ and most being apply_
Standardizes using the American "armor" in apply_armored_damage and related procs rather than the British "armour"

## Why It's Good For The Game

It's honestly insane how we have zero standardization (or at least in practice) for our most commonly used proc categories such as damage. It should be standardized. No one should have to spend 10 minutes trying to figure out why they can't use apply_overall_damage  just to figure out that the proc is actually called "take_overall_damage" instead while seeing that pretty much every proc next to it uses "apply_"

Also, differences in nomenclature and slang are retarded. half of armor should be called "armour" while the other half is "armor". This isn't TWE bri ish royal marines, this is USA colonial marines. Thus, this should be standardized on most used procs.

This is step one of further standardization, mainly because I am malding right now. 

/end rant

## Changelog

:cl: TotalEpicness
refactor: Standardized mob damage proc naming.
/:cl:


PhantomEpicness:
merge conflict hell

PR #1593, opened 2022-11-17
New water overlay system

kugamo:
## About The Pull Request

This PR adds a new way of rendering humans in water, unique from the legacy way of doing it (having static overlays of the turf that the human would render under -- looked especially bad while moving). This method draws the overlay on the mob itself, which looks alot better while moving. The water overlays were broken in the ambient occlusion PR. 

- [ ] Make sure these changes doesn't obliterate the server
- [ ] Add logic for when laying down in water
- [ ] redo sprites to not look doo doo
- [ ] Either decide on no directional movement sprites, or create new system to handle them
- [ ] Add logic so this all works on xenos too

## Why It's Good For The Game

Visually appealing, adds to atmosphere and the enviroment of the game.

https://user-images.githubusercontent.com/16618648/202414163-76420021-177b-402d-9777-c2de9c9220cb.mp4

## Changelog

:cl:
refactor: Refactored water overlays so that they render on mob and not on turf.
/:cl:


code/datums/effects/water.dm
```diff
@@ -0,0 +1,27 @@
...
+
+/datum/effects/water/New(var/atom/A)
+	..(A)
+	if(ishuman(A))
+		var/mob/living/carbon/human/H = A
+		H.update_effects()
+
+/datum/effects/water/Destroy()
+	if(affected_atom)
+		LAZYREMOVE(affected_atom.effects_list, src)
```
fira, line 17: this is already done by parent

Huffie56:
I love the effect at waist level...
In my opinion you should start with the lowest water depth possible the ankle level or bellow it.
I think that's the one that will give you the less work to do? since it won't affect speed etc....
what's your opinion on that @kugamo  ?

fira (commented):
i dunno.. the use of `Exited` and `in effects_list` is kinda snowflake.. might be able to find a better solution with an element

kugamo:
> I love the effect at waist level...
> In my opinion you should start with the lowest water depth possible the ankle level or bellow it.
> I think that's the one that will give you the less work to do? since it won't affect speed etc....
> what's your opinion on that @kugamo ?

If you mean making it look like the character is only in ankle deep water, i tried keeping consistent with the water level in the legecy system. What the sprite looks like wont affect lag, just how its drawn on the character will.. so changing just the sprite shouldnt matter speed wise. But if I get this system to a feasible state i will polish with adding it for shorelines too, ankle deep…

stanalbatross:
You could try copy the footsteps element for this

code/datums/effects/water.dm
kugamo: parent nulls `affected_atom` so I cannot call it before calling `update_effects()` to remove the spirte, nor can I leave it up to the parent's `Destroy()` since it doesnt call `update_effects()` 😔

fira (changes requested):
Can't do like this, not architecturally sound and might be perf impacting -- if you attach a component to all humans, and only humans, what's the point of it being a component?

I'd suggest to use signals like COMSIG_MOVABLE_TURF_ENTERED, COMSIG_MOVABLE_MOVED or COMSIG_MOVABLE_PRE_MOVE that already exist

An example i could see is eg.
 * Entered of turf applies a component "handle water overlay" which hooks COMSIG_MOVABLE_MOVE
 * this handler does the stuff and deletes itself if the loc is not matching
 
it's kinda flaky there might be better solutions too...

monkeysfist101 (approved):
I dig it. Sprites approved.

PR #1597, opened 2022-11-17
Simple Walking Implementation

Geevies:
## About The Pull Request

Implements the UI button and keybinding to toggle movement intents, as well as toggling to walk and slowing down. Also supports xenos using walking icon_states, but only if they have walking states in the first place.

Ideally this acts as a foundation to implement more whacky movement intent based mechanics, like slipping on wet floors if you're running, or xenos gaining invulnerability to all threats, physical or magical.

## Why It's Good For The Game

It's neat, I guess. A relic of the old times that can act as a fun gameplay mechanic, especially if the tactics get expanded upon in the future. If not, extreme roleplayers can just have fun, walking around.

## Changelog

:cl:
add: Re-added the basics of walking.
/:cl:


stanalbatross:
For now the only thing I'm worried about is the fumble potential

We can unbind the key, but can you add a pref to hide the UI button?

In fact this has given me an idea to make prefs to hide all the UI buttons individually

Something to do when I get my PC fixed

code/__DEFINES/mobs.dm
```diff
@@ -150,6 +150,10 @@
 
 //=================================================
 
+/////////////////MOVE DEFINES//////////////////////
```
fira, line 153: triple /// for annotations etc

code/modules/mob/living/carbon/xenomorph/update_icons.dm
```diff
@@ -39,6 +39,15 @@
 			Q.queen_standing_icon = icon_xeno
 			Q.queen_ovipositor_icon = get_icon_from_source(CONFIG_GET(string/alien_queen_ovipositor))
 
+	var/mutation_caste_state = "[mutation_type] [caste.caste_type]"
+	if(!walking_state_cache[mutation_caste_state])
+		var/cache_walking_state = FALSE
+		for(var/state in icon_states(icon))
+			if(findtext(state, "Walking"))
```
fira, line 46: whew that part's pretty dubious

what's the point/problem at stake? can't we just define on castes which has walks and which not?
i think ideally we'd just want all of them to have placeholder icons

code/modules/mob/living/carbon/inventory.dm
```diff
@@ -7,7 +7,8 @@
 	update_inv_handcuffed()
 
 /mob/living/carbon/proc/legcuff_update()
-	recalculate_move_delay = TRUE
```
fira, line 10: why remov?

harryob:
conflicting with yourself - SAD!

ThePiachu:
I mean, I removed that code back in the day for a reason, mostly because people would toggle walk by accident by pressing "5" and not realise it. Nobody ever walked anywhere, except when they triggered it by accident or really niche roleplay stuff. Plus if we don't have any special mechanics for walking other than "you go slower", I don't really see a point of bringing it back as is.

Slipping was removed AFAIR because it similarly was only used for griefing. First it was AFAIR that you couldn't slip if you had your boots on, and now I don't think you can slip without some special stuff like an acid trap or maybe space lube...

code/modules/mob/living/carbon/xenomorph/update_icons.dm
Geevies: my main concern is for stuff like strains that may have running and not walking, using this automated check we don't have to worry about going through each one and setting the correct one, also future-proofs it

code/modules/mob/living/carbon/inventory.dm
Geevies: `set_movement_intent` handles it

code/modules/mob/living/carbon/xenomorph/update_icons.dm
fira: Ugh, in all honesty this looks more like cornercutting to me... There's no real reason we'd not just have them all defined like the rest -- and even then it could just be a flag on xenos modified by apply_mutator

if you don't mind carlarc's son posting this on discord, i guess we can merge it

fira (approved):

code/modules/mob/living/carbon/xenomorph/update_icons.dm
harryob: truly we will be memed on for this but that's okay

PR #1602, opened 2022-11-18
Fulton Objectives now delete themselves on a gibbed body

### LINKED ISSUE #1583: Retrieve objective null reference error
## Testmerges:

- #965

## Description

Recently while doing some light IO work I noticed this time the Retrieve tab had a null reference error:

image

Not to have a null error there.

## Steps to reproduce

1) Do intel
2) Periodically check Retrieve tab
3) ???
4) Null error

ThePiachu:
## About The Pull Request

Fulton objective did not clear itself when a given body was gibbed. Fixed that.

Fixes #1583

## Why It's Good For The Game

Less NREs, less IOs not knowing where their fultons went.

## Changelog

:cl:
fix: Gibbing a mob that had a fulton attached to them won't create a null reference error in View Objectives window any more.
refactor: Removed the old, unused Objective Controller code.
/:cl:


ThePiachu:
Actually, I'm going to investigate why gibbed mobs don't send the qdel signal (at least to the fulton) which is the root cause of this issue...

ThePiachu:
Okay, after a few hours of debugging, I finally figured it out. When we deleted an objective, it didn't delete itself from active objectives. The old version of the controller used to do that, but the new version did not. Also found the old, unused objective controller code is still in the files mucking things for people that want to figure out what's going on, so I removed it.

fira (approved):
good catch...

PR #1608, opened 2022-11-18
Projectile Subsystem and misc adjustments

fira:
## About The Pull Request

What this does:
 * Adds a projectile subsystem for scheduling their movement. The follow_flightpath sleeps are detrimental to perf, and instead this allows them to be updated 2x as often.
 * Projectile speed is +-15% and has a firing offset, to result in a proper bullet trail instead of a dotted line.
 * Gun projectile firing angle takes into account scatter and clicked pixel. This is *mostly* a visual change, but also affects direction bullets take past initially aimed point.
 * Decouples gun/projectile from sleeps. This has caused many practical issues in the past.
 * Removes simple_mobs firing and AMMO_SCANS_NEARBY. They're defunct/broken.
 * Removes projectile speed-up from AMMO_ROCKET - instead granting them full speed. This only really mattered for thermobaric launcher at long range.

## Why It's Good For The Game

Smoother, more visual feedback (perhaps to a fault, as it shows how goofy our system is)

Example of in-turf targeting, two trajectories based on clicking either X mark:
image

Video example of difference, yes, it is subtle:
https://streamable.com/ph18yj
https://streamable.com/ndq7f5

## Changelog

:cl:
add: Added +-15% variation to bullet speed and firing offsets to result in a better bullet trail.
add: Gun firing now takes into account click position. This can bend the bullet trajectory beyond the clicked point.
add: Projectiles now attempt to mimick a proper bullet stream by offsetting slightly. This is purely visual.
code: Removed rocket speed-up, and gave them full speed. Only really matters for offscreen admin-spawned thermobaric rockets.
fix: Fixed deletion issues with projectiles, and decoupled it from gun logic due to recurring issues.
del: Removed simple_mob-s shooting and AMMO_SCANS_NEARY. They're defunct/broken.
fix: Removed a number of micro sleeps in explosion code. This may reduce amount of shrapnel hits.
code: Tweaked Aimed Shot homing logic which should make it even more accurate in the offchance it ever matters.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -67,21 +67,25 @@
 	return FALSE
 
 /datum/ammo/proc/do_at_half_range(obj/item/projectile/P)
+	SHOULD_NOT_SLEEP(TRUE)
 	return
 
 /datum/ammo/proc/on_embed(var/mob/embedded_mob, var/obj/limb/target_organ)
 	return
 
 /datum/ammo/proc/do_at_max_range(obj/item/projectile/P)
+	SHOULD_NOT_SLEEP(TRUE)
 	return
 
 /datum/ammo/proc/on_shield_block(mob/M, obj/item/projectile/P) //Does it do something special when shield blocked? Ie. a flare or grenade that still blows up.
 	return
 
 /datum/ammo/proc/on_hit_turf(turf/T, obj/item/projectile/P) //Special effects when hitting dense turfs.
+	SHOULD_NOT_SLEEP(TRUE)
 	return
 
 /datum/ammo/proc/on_hit_mob(mob/M, obj/item/projectile/P, mob/user) //Special effects when hitting mobs.
+	SHOULD_NOT_SLEEP(TRUE)
 	return
 
 ///Special effects when pointblanking mobs. Ultimately called from /living/attackby(). Return TRUE to end the PB attempt.
@@ -89,6 +93,7 @@
 	return
 
 /datum/ammo/proc/on_hit_obj(obj/O, obj/item/projectile/P) //Special effects when hitting objects.
+	SHOULD_NOT_SLEEP(TRUE)
 	return
 
 /datum/ammo/proc/on_near_target(turf/T, obj/item/projectile/P) //Special effects when passing near something. Range of things that triggers it is controlled by other ammo flags.
@@ -3270,4 +3275,4 @@
 /datum/ammo/hugger_container/proc/spawn_hugger(var/turf/T)
 	var/obj/item/clothing/mask/facehugger/child = new(T)
 	child.hivenumber = hugger_hive
-	child.leap_at_nearest_target()
+	INVOKE_ASYNC(child, /obj/item/clothing/mask/facehugger.proc/leap_at_nearest_target)
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -2819,13 +2819,13 @@
 	if(!wasfrozen)
 		target.frozen = TRUE
 	//sleeps during this proc...
-	target.throw_atom(fired_proj.firer, get_dist(fired_proj.firer, target)-1, SPEED_VERY_FAST)
+	INVOKE_ASYNC(target, /atom/movable.proc/throw_atom, fired_proj.firer, get_dist(fired_proj.firer, target)-1, SPEED_VERY_FAST)
 	//inmediately after the sleeps, unfreezes so they can wiggle around.
 	if(!wasfrozen)
 		target.frozen = FALSE
 
 	qdel(tail_beam)
-	addtimer(CALLBACK(src, /datum/ammo/xeno/oppressor_tail/proc/remove_tail_overlay, target, tail_image), 0.5 SECONDS) //needed so it can actually be seen as it gets deleted too quickly otherwise.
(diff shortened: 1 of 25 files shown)


harryob (approved):
this certainly looks good for a TM - i'm sure there's edge cases that i can't predict rn that'll come up then

ManoPablo:
mate, something needs to be done to rockets, it feels awful to use the Sadar while this is being tmerged, please remove the variation on Sadar rockets it throws the aim off bad, and you simply can't lead properly because you don't have a good idea on how fast the rocket is going, if you want to keep that on rockets add a balance tag because it really affects aim on the sadar making it much much harder for even experienced players to aim

fira:
Rocket speedup was super painful and unreliable to deal with when it came to variable speed and travel carryover so i just removed it entirely, giving them full speed all the time instead...

This impacts two situations:
 - It makes the RPG hit more reliably and slightly faster beyond PB range (max 15-20%) which should fall back in line with people expectations from before
 - It makes admin-spawned thermobaric launcher faster without having to fire offscreen for it, but that's basically never used

code/modules/projectiles/projectile.dm
```diff
@@ -134,13 +146,12 @@
...
-	if(!original)
-		original = istype(original_override) ? original_override : target
-	src.is_shrapnel = is_shrapnel
+// Target, firer, shot from (i.e. the gun), projectile range, projectile speed, original target (who was aimed at, not where projectile is going towards)
+/obj/item/projectile/proc/fire_at(atom/target, atom/F, atom/S, range = 30, speed = 1, atom/original_override)
+	SHOULD_NOT_SLEEP(TRUE)
+	original = original || original_override || target
 	if(!loc)
-		if (!is_shrapnel)
+		if (projectile_flags & PROJECTILE_SHRAPNEL)
```
Segrain, line 154:
```suggestion
		if (!(projectile_flags & PROJECTILE_SHRAPNEL))
```

fira: OOPS... Good catch

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1496,7 +1501,7 @@
 	damage_falloff = 0
 
 /datum/ammo/bullet/sniper/on_hit_mob(mob/M,obj/item/projectile/P)
-	if(P.homing_target && M == P.homing_target)
+	if(M == P.original)
```
Segrain, line 1504:
```suggestion
	if((P.projectile_flags & PROJECTILE_HOMING) && (M == P.original))
```

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1518,7 +1523,7 @@
 	))
 
 /datum/ammo/bullet/sniper/incendiary/on_hit_mob(mob/M,obj/item/projectile/P)
-	if(P.homing_target && M == P.homing_target)
+	if(M == P.original)
```
Segrain, line 1526:
```suggestion
	if((P.projectile_flags & PROJECTILE_HOMING) && (M == P.original))
```

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1541,7 +1546,7 @@
 	penetration = 0
 
 /datum/ammo/bullet/sniper/flak/on_hit_mob(mob/M,obj/item/projectile/P)
-	if(P.homing_target && M == P.homing_target)
+	if(M == P.original)
```
Segrain, line 1549:
```suggestion
	if((P.projectile_flags & PROJECTILE_HOMING) && (M == P.original))
```

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1641,7 +1646,7 @@
 	shell_speed = AMMO_SPEED_TIER_6
 
 /datum/ammo/bullet/sniper/anti_materiel/on_hit_mob(mob/M,obj/item/projectile/P)
-	if(P.homing_target && M == P.homing_target)
+	if(M == P.original)
```
Segrain, line 1649:
```suggestion
	if((P.projectile_flags & PROJECTILE_HOMING) && (M == P.original))
```

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1669,7 +1674,7 @@
 	))
 
 /datum/ammo/bullet/sniper/elite/on_hit_mob(mob/M,obj/item/projectile/P)
-	if(P.homing_target && M == P.homing_target)
+	if(M == P.original)
```
Segrain, line 1677:
```suggestion
	if((P.projectile_flags & PROJECTILE_HOMING) && (M == P.original))
```

Segrain (changes requested):

fira:
The sleep safety net isn't good enough, causing erroring bullets to be stuck in midair. This is a problem because it appears we still have sleeps in `/mob/living/bullet_act` somehow (i suspect heavily `/mob/living/carbon/Xenomorph/apply_damage`)

code/modules/projectiles/ammo_datums.dm
Segrain, line 1649:
```suggestion
	if((P.projectile_flags & PROJECTILE_BULLSEYE) && M == P.original)
```
Missed one. Not that it is supposed to be used anyway.

harryob (approved):
code quality lgtm, still possible there's something being overlooked, but

morrowwolf (approved):
Code LGTM. Testmerges went well.

PR #1614, opened 2022-11-19
Survivor Variant Picking + Survivor Priority

Geevies:
## About The Pull Request

Split the synthvivor roles into distinct equipment presets, instead of a proc that equips them.

Players can now select a variant of survivor they'd prefer to spawn as from a list of:
- Civilian Survivor
- Security Survivor
- Scientist Survivor
- Medical Survivor
- Engineering Survivor
- Corporate Survivor

They will only spawn with a loadout related to that variant, unless they hit a special survivor spawn, in which case they will get the special spawn's equipment.

As for the special spawners, I've implemented a priority system to them, so rare spawners with equipment can be made to have a high priority, meaning it'll be selected before other spawners of lower priority are selected. I've applied this to the FORECON spawners, meaning the CO will have the highest priority, SL the next, and then the FORECON specialists, and then the generic FORECON grunts.

## Why It's Good For The Game

It makes it easier to make a survivor character to actually RP with if you can write their background (medical, security, etc), and it also makes special roles appear more often if the Nightmare spawn actually spawns.

## Changelog

:cl:
add: Added survivor variant picking, in the character creation menu, under special roles. This makes it so you only spawn as that survivor type, unless you hit a special spawn, such as CLF or FORECON, for example.
add: Added survivor spawner priority, so that the CLF and FORECON CO spawners are used before others.
/:cl:


ThePiachu:
Nice, hopefully this will make the nightmare insert survivors spawn a bit more readily!

code/datums/map_config.dm
```diff
@@ -67,9 +69,7 @@
 		/datum/equipment_preset/survivor/engineer
 	)
 
-	synth_survivor_types = list(
-		/datum/equipment_preset/synth/survivor
-	)
+	synth_survivor_types = typesof(/datum/equipment_preset/synth/survivor)
```
morrowwolf, line 72: You don't want to pick up the FORECON synthvivor (or other future map specific synthvivors) so you'll have to be explicit with your list.

morrowwolf (changes requested):

PR #1618, opened 2022-11-19
Queen Charge

Geevies:
## About The Pull Request

Re-adds Queen Charge using Charger backend. It's pretty expensive plasma-wise so they won't be replacing crushers.

## Why It's Good For The Game

https://youtu.be/1sE-YwK6_PI

## Changelog

:cl:
add: Re-added Queen charge using Charger code.
/:cl:


stanalbatross:
Please provide an actual "why it's good for the game"

novaepee:
>queen charge back in CM

image

CABSHEP:
Rename it to "Get out of Jail Free Card", because that is what it will always be used for and that was the reason it was removed. 
Now Queen will either just charge, screech and then retreat, or just charge to run away.

kugamo (changes requested):
Can we make sure the damage overlays work while the queen is charging, they were added in the mean time since this was in the game.

PR #1373, opened 2022-10-29
[IDB IGNORE] moves all the sprites into the main repo

harryob:
## About The Pull Request

no i didn't just go crazy and do this i asked before

i also included xenos_old and xenos_halloween for funsies

## Why It's Good For The Game

it was a bit inconvenient and it simplifies things for TMs, etc

## Changelog

:cl:

/:cl:

PR #1619, opened 2022-11-19
Strain icon format change

aceroyuno:
## About The Pull Request
Adds var/mutation_icon_state to the xenomorph base type. All sprite logic is handled by this var as opposed to the mutation_type var. Strains will no longer need an updated sprite sheet and can now use the base sprite with the original .dmi file.

Had to make a new one since the last one went stale.

## Why It's Good For The Game
Allows for creation of strains gameplay wise not to be affected by the strains icon logic. And it makes it simpler for checking code for spriters. If it has no sprite for it's strain, it defaults to using the normal icon state.

## Changelog

:cl: aceroy
refactor: Changed xenomorph icon logic to use new var mutation_icon_state and not mutation_type
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
@@ -122,6 +122,7 @@
 	var/acid_level = 0
 
 	// Mutator-related and other important vars
+	var/mutation_icon = null
 	var/mutation_type = null
 	var/datum/mutator_set/individual_mutators/mutators = new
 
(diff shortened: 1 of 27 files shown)


fira:
Think we should first roll sprites merge update then run it together... Lots of changes being made to xeno sprites currently

aceroyuno:
Alright, I'll mark this as draft. Think you can send me a link to the current stuff so I can take a gander.

aceroyuno:
Don't mind me. Just using github on the phone. Close and comment should not be right next to eachother

fira:
Basically waiting on #1373 - TestMerging it next round. 
This is gonna simplify xeno icon handling a lot, it shouldn't actually conflict i think but...

PR #1621, opened 2022-11-19
Added a Research Objectives Screen

ThePiachu:
## About The Pull Request

I've played a bit of Research and noticed how disorganised it feels at times. There really isn't an easy way of tracking which chems were done and which you still need to synthesize / analyze. So I figured I'd give Research a similar objectives screen to IOs:

image

At the moment it's a bit sparse, but ideally we would have more objectives for Research to do in the future so we could expand the screen as needed.

The screen doesn't really reveal any information Research wouldn't know, asides from the names of unidentified chems before they are located which isn't really all useful by itself.

Also cleaned the code around chems. Moved two global variables into the Chemical Data struct, added one around tracking unfinished chems, and fixed a piece of code that didn't properly check for chem flags (it had `&&` instead of `&` and would reject all generated chems).

## Why It's Good For The Game

It helps Research stay organised and know which chems they still need to work on.

## Changelog

:cl:
add: Added a View Research Objective verb. This will let Researchers be able to view which chems they still need to synthesize and analyze to keep more organized.
fix: Fixed the chem code that saves researchable chems. It used not to save custom chems, now it does.
code: Cleaned up the code around tracking chems and chem objectives a little. It now actually tracks which chems haven't been researched.
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/ResearchMemories.js
+++ b/tgui/packages/tgui/interfaces/ResearchMemories.js
@@ -23,9 +23,9 @@ export const ResearchMemories = (props, context) => {
   return (
     <Window width={650} height={700} theme={theme}>
       <Window.Content scrollable>
-        <Section title='Clearance: ${clearance}'>
+        <Section title={'Clearance: ' + clearance}>
         </Section>
-        <Section title='Research Credits: ${research_credits}'>
+        <Section title={'Research Credits: ' + research_credits}>
         </Section>
 
         <Objectives />
@@ -85,8 +85,7 @@ const Objectives = (props, context) => {
 
   return (
     <Section
-      title="Objectives"
-      }>
+      title="Objectives">
       <LabeledList>
         {data.objectives.map((page) => {
           return (
--- a/tgui/packages/tgui/interfaces/ResearchMemories.js
+++ b/tgui/packages/tgui/interfaces/ResearchMemories.js
@@ -70,27 +70,8 @@ const CluesAdvanced = (props, context) => {
               px="1rem"
               py=".5rem">
               <Flex.Item>
-                {!!clue.color && (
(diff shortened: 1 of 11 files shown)


tgui/packages/tgui/interfaces/ResearchMemories.js
```diff
@@ -0,0 +1,108 @@
+import { useBackend, useLocalState } from '../backend';
+import {
+  Button,
+  Flex,
+  Section,
+  Box,
+  Tabs,
+  Fragment,
```
harryob, line 8: don't need to import button and fragment if you're not using them

tgui/packages/tgui/interfaces/ResearchMemories.js
```diff
@@ -0,0 +1,108 @@
...
+
+  const { clearance, research_credits, theme, clue_categories } = data;
+
+  return (
+    <Window width={650} height={700} theme={theme}>
+      <Window.Content scrollable>
+        <Section title={'Clearance: ' + clearance}>
+        </Section>
+        <Section title={'Research Credits: ' + research_credits}>
+        </Section>
```
harryob, line 29: empty fragments autoclose, linter fail. you might wanna use something other than sections for this though

harryob (changes requested):

tgui/packages/tgui/interfaces/ResearchMemories.js
ThePiachu: What would you suggest? I'm really bad at JS...

ThePiachu: Changed it to Flex.Item

ThePiachu:
Okay, tweaked the screen a touch based on the feedback:
image
(moved the Research Credits into a smaller sub-section)

fira (approved):

PR #1623, opened 2022-11-20
Smartfridge tgui

mullenpaul:
## About The Pull Request

Migrates the smart fridge to TGUI and reworks internal logic.

Adds a standard component to use for displaying an electrical access panel for hacking. The goal is to indicate that there is an access panel and to try and make the UI more immersive.

## Why It's Good For The Game

Smartfridge with some items loaded
image

Smartfridge with panel open and wire status lights active. Solid green for good or flashing red for bad.
image

## Changelog

:cl:
fix: fixed bug in turing autodispensor so that pill bottles can be in local smart fridge
refactor: migrated smart fridge to tgui
refactor: added common electrical panel for tgui components
/:cl:


tgui/packages/tgui/interfaces/common/ElectricalPanel.tsx
```diff
@@ -0,0 +1,184 @@
...
+            wrap
+            className="WarningIcon"
+            direction="column">
+            <Flex.Item>
+              <Icon name="bolt" size={2} />
+            </Flex.Item>
+            <Flex.Item>
+              <span>
+                Electrical hazard authorised <br />
+                personnel only
```
harryob, line 65: `Electrical Hazard • Authorised Personnel Only` imo

tgui/packages/tgui/interfaces/SmartFridge.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+export const SmartFridge = (_, context) => {
+  const { data } = useBackend<SmartFridgeData>(context);
+  return (
+    <Window theme="weyland" width={400} height={600}>
+      <Window.Content className="SmartFridge" scrollable>
+        <Stack vertical>
+          {data.secure && (
+            <Stack.Item>
+              <NoticeBox>Smart fridge is secure</NoticeBox>
```
harryob, line 153:
```suggestion
              <NoticeBox>Smart Fridge is on Secured Mode</NoticeBox>
```
or something to that effect

harryob (changes requested):

harryob:
also - imo it needs to be a fluid element for the wires

tgui/packages/tgui/interfaces/common/ElectricalPanel.tsx
mullenpaul: This makes sense

tgui/packages/tgui/interfaces/SmartFridge.tsx
mullenpaul: sounds good, went with Smart Fridge is in secure mode

mullenpaul:
updated pictures to represent changes

code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -128,194 +126,253 @@
...
-			var/count = chemical_data.shared_item_quantity[K]
-			if (count > 0)
-				networked_items.Add(list(list("display_name" = html_encode(capitalize(K)), "vend" = i, "quantity" = count)))
+/obj/structure/machinery/smartfridge/ui_status(mob/user, datum/ui_state/state)
+	. = ..()
+	if(inoperable())
+		return UI_CLOSE
+
+/obj/structure/machinery/smartfridge/ui_static_data(mob/user)
+	. = ..(user)
```
stanalbatross, line 166: ? why user in the brackets here

also not sure if this needs a parent call at all

code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -128,194 +126,253 @@
...
 
 	var/list/wire_descriptions = get_wire_descriptions()
 	var/list/panel_wires = list()
 	for(var/wire = 1 to wire_descriptions.len)
 		panel_wires += list(list("desc" = wire_descriptions[wire], "cut" = isWireCut(wire)))
 
-	if (panel_wires)
-		data["wires"] = panel_wires
+	.["electrical"] = list(
+		"electrified" = seconds_electrified > 0,
```
stanalbatross, line 181: I am going to invoke the "you touch it you fix it" clause to say PLEASE refactor the seconds_electrified thing to use actual timers/cooldowns rather than /process

code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -128,194 +126,253 @@
...
+	src.add_fingerprint(user)
 
-	if (href_list["toggletransfer"])
-		if(is_secure_fridge)
-			if(locked == FRIDGE_LOCK_COMPLETE)
-				to_chat(usr, SPAN_DANGER("Access denied."))
+	switch(action)
+		if("vend")
+			if(!ispowered)
+				to_chat(user, SPAN_WARNING("[src] has no power."))
```
stanalbatross, line 272:
```suggestion
				to_chat(user, SPAN_WARNING("\The [src] has no power."))
```

code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -128,194 +126,253 @@
...
 				return FALSE
-			if(!allowed(usr) && locked == FRIDGE_LOCK_ID)
-				to_chat(usr, SPAN_DANGER("Access denied."))
+			if (!in_range(src, usr))
+				return FALSE
```
stanalbatross, line 275: should be handled by ui state, no?

code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -128,194 +126,253 @@
...
+				if (i <= 0)
+					return TRUE
+		if("transfer")
+			if(!ispowered)
+				to_chat(user, SPAN_WARNING("[src] has no power."))
 				return FALSE
-			if(!allowed(usr) && locked == FRIDGE_LOCK_ID)
-				to_chat(usr, SPAN_DANGER("Access denied."))
+			if (!in_range(src, usr))
+				return FALSE
```
stanalbatross, line 311: same for these checks

code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -128,194 +126,253 @@
...
-		var/list/networked_items = list()
-		for (var/i=1 to length(chemical_data.shared_item_quantity))
-			var/K = chemical_data.shared_item_quantity[i]
-			var/count = chemical_data.shared_item_quantity[K]
-			if (count > 0)
-				networked_items.Add(list(list("display_name" = html_encode(capitalize(K)), "vend" = i, "quantity" = count)))
+/obj/structure/machinery/smartfridge/ui_status(mob/user, datum/ui_state/state)
+	. = ..()
+	if(inoperable())
+		return UI_CLOSE
```
stanalbatross, line 163: n.b. should make this not close if it's the wires UI being accessed (check for panel open here)

stanalbatross (changes requested):

code/game/machinery/kitchen/smartfridge.dm
mullenpaul: I think this was me early-optimising. The idea was that the handling of hacking/electrical stuff would be handled upstream. The proc contains user as a param so I thought you would want to pass it in (like most languages) - I'm guessing that isn't needed?

In any case, upstream of this does nothing right now so not going to pre-optimise.

code/game/machinery/kitchen/smartfridge.dm
mullenpaul: Done using cooldowns

code/game/machinery/kitchen/smartfridge.dm
mullenpaul: Was just copied out of original, stripped.

code/game/machinery/kitchen/smartfridge.dm
mullenpaul: Good catch

code/game/machinery/kitchen/smartfridge.dm
mullenpaul: I took the inoperable check out of ui_status and switched the check to now allow vend & transfer - allowing wire mending for cut/fix/pulse

stanalbatross (approved):

PR #1633, opened 2022-11-20
Changes HG 37 shotguns to have dual internal tube magazines

NoKohi:
## About The Pull Request

this makes it so that the HG 37 shotguns have dual internal tube magazines - this means you can do stuff like load one magazine with slugs and one magazine with buckshot, then switch them around for whatever the situation calls for

currently you change the internal magazine being used via toggle burst fire mode

this changes the internal magazines for the HG 37-12 and 37-17 to have 4 and 5 shells capacity respectively, but they still have 8+1 and 10+1 capacity because they get two magazines now, so iunno

## Why It's Good For The Game

the description of the HG 37 shotguns would suggest that they function this way; monkeysfist wanted this and carlarc said he'd do it (then I did it because I thought this was a cool idea when I saw it, sorry carlarc)

also it makes the HG have a unique feature which is neat

## Changelog

:cl:
add: changed the HG 37 shotguns to have dual internal tube magazines, switch between them via toggle burst fire mode
code: cleaned up a few single letter var names
/:cl:


code/modules/projectiles/guns/shotguns.dm
```diff
@@ -1049,8 +1053,44 @@ can cause issues with ammo types getting mixed up during the burst.
...
+	QDEL_NULL(secondary_tube)
+	QDEL_NULL(active_tube)
+	. = ..()
+
+/obj/item/weapon/gun/shotgun/pump/cmb/proc/do_toggle_tube(var/mob/user)
+	if(!ishuman(user) || user.is_mob_incapacitated())
+		return FALSE
+	var/obj/item/weapon/gun/shotgun/pump/cmb/shotgun = user.get_active_hand()
+	if(!istype(shotgun))
+		to_chat(usr, "You must be holding [src] to change the internal tube magazine")
```
carlarctg, line 1074: usr

code/modules/projectiles/guns/shotguns.dm
```diff
@@ -1049,8 +1053,44 @@ can cause issues with ammo types getting mixed up during the burst.
...
+		return
+	if(!active_tube)
+		return
+	if(active_tube == primary_tube)
+		active_tube = secondary_tube
+	else
+		active_tube = primary_tube
+	to_chat(user, "You switch the internal tube magazine.")
+	playsound(src, 'sound/machines/click.ogg', 20, 1)
```
carlarctg, line 1083:
```suggestion
	playsound(src, 'sound/machines/click.ogg', 20, TRUE)
```

code/modules/projectiles/guns/shotguns.dm
stanalbatross, line 1074: \the

code/modules/projectiles/guns/shotguns.dm
```diff
@@ -1049,8 +1053,44 @@ can cause issues with ammo types getting mixed up during the burst.
...
+	shotgun.current_mag = active_tube
+
+	return TRUE
+
+/obj/item/weapon/gun/shotgun/pump/cmb/verb/toggle_tube()
```
carlarctg, line 1088: really don't think it should be called 'toggle', to toggle something is to turn it on or off. this is just swapping between identical tubes. Change to 'swap'?

code/modules/projectiles/guns/shotguns.dm
```diff
@@ -1049,8 +1053,44 @@ can cause issues with ammo types getting mixed up during the burst.
 
 /obj/item/weapon/gun/shotgun/pump/cmb/Initialize(mapload, spawn_empty)
 	. = ..()
+	primary_tube = current_mag
+	secondary_tube = new current_mag.type(src, spawn_empty? 1:0)
```
carlarctg, line 1057:
```suggestion
	secondary_tube = new current_mag.type(src, spawn_empty ? TRUE : FALSE)
```

code/modules/projectiles/guns/shotguns.dm
```diff
@@ -1049,8 +1053,44 @@ can cause issues with ammo types getting mixed up during the burst.
...
+	QDEL_NULL(primary_tube)
+	QDEL_NULL(secondary_tube)
+	QDEL_NULL(active_tube)
+	. = ..()
+
+/obj/item/weapon/gun/shotgun/pump/cmb/proc/do_toggle_tube(var/mob/user)
+	if(!ishuman(user) || user.is_mob_incapacitated())
+		return FALSE
+	var/obj/item/weapon/gun/shotgun/pump/cmb/shotgun = user.get_active_hand()
+	if(!istype(shotgun))
```
fira, line 1073: could just check shotgun != src instead i think

code/modules/projectiles/guns/shotguns.dm
```diff
@@ -1049,8 +1053,44 @@ can cause issues with ammo types getting mixed up during the burst.
...
+		return
+	if(!active_tube)
+		return
+	if(active_tube == primary_tube)
+		active_tube = secondary_tube
+	else
+		active_tube = primary_tube
+	to_chat(user, "You switch the internal tube magazine.")
+	playsound(src, 'sound/machines/click.ogg', 20, 1)
+	shotgun.current_mag = active_tube
```
carlarctg, line 1084: what's up with this ``shotgun.`` looks unnecessary

code/modules/projectiles/guns/shotguns.dm
```diff
@@ -1049,8 +1053,44 @@ can cause issues with ammo types getting mixed up during the burst.
...
+	. = ..()
+
+/obj/item/weapon/gun/shotgun/pump/cmb/proc/do_toggle_tube(var/mob/user)
+	if(!ishuman(user) || user.is_mob_incapacitated())
+		return FALSE
+	var/obj/item/weapon/gun/shotgun/pump/cmb/shotgun = user.get_active_hand()
+	if(!istype(shotgun))
+		to_chat(usr, "You must be holding [src] to change the internal tube magazine")
+		return
+	if(!active_tube)
```
fira, line 1076: when would that happen?

carlarctg (commented):
vitch

stanalbatross (changes requested):
Also
- having this on the CMB instead of a parent type like pump/dual_tube/ is somewhat annoying for future coders if they want to add another kind of dual tube shotgun

- you could make the tube toggle by toggling burst fire or something instead of having a snowflake verb and action button. As it stands this will be annoying to use because you'll either have to press the action button / weapon tab verb or make a macro (and those will eventually be removed), so I'd recommend making this bindable / on burst fire or something.

code/modules/projectiles/guns/shotguns.dm
NoKohi: what's wrong here :(

code/modules/projectiles/guns/shotguns.dm
NoKohi: when trying to do it off hand

NoKohi:
so far the most trouble I've had is trying to figure out a neat solution to the whole swapping hotkey thing

NoKohi:
the latest commit has a bug where if you pick up two guns, and then drop a dual tube shotgun, you'll lose the "toggle burst fire" verb in the weapons category (you can still toggle it via other means however this specific instance is gone until you re-equip the gun, which is bad however rare it may be) - my proposed solution to this problem would be to divide it up like unique action does it, to allow for interactions that don't override the verb in the tab (I'll do this later but I don't like this solution and think there has to be a better way)

I'm not satisfied with the burst fire toggle for changing the internal magazine and think that there has to be a better way of implementing this, one where even your below average player could figure it out through mechanics they're already familiar with - I don't think it's worth adding its own hotkey for, because it's too specific an interaction to be worth it (people were unironically proposing adding a second unique action)

monkeysfist proposed adding a fire delay + 150ms after you switch magazines, to prevent a one-two combination with slugs and buckshot; however I'm not sure if this is warranted, as you can already do this normally just by loading buckshot and then a slug last - a delay like that would make the shotgun less viable, imo, and would undermine the appeal of being able to switch the magazines to begin with

also monkeysfist wanted a variant which was pump/semi variable - this seems doable to me

code/_onclick/hud/screen_objects.dm
```diff
@@ -382,7 +382,7 @@
 
 		if("Toggle burst fire")
 			var/obj/item/weapon/gun/G = user.get_held_item()
-			if(istype(G)) G.toggle_burst()
+			if(istype(G)) G.use_toggle_burst()
```
harryob, line 385: this is such an ugly file generally, but i’d still prefer the control statement be on a different line

code/modules/projectiles/gun_helpers.dm
```diff
@@ -683,51 +674,62 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
...
-	to_chat(usr, SPAN_NOTICE("[icon2html(src, usr)] You [flags_gun_features & GUN_BURST_ON ? "<B>enable</b>" : "<B>disable</b>"] [src]'s burst fire mode."))
+	to_chat(user, SPAN_NOTICE("[icon2html(src, user)] You [flags_gun_features & GUN_BURST_ON ? "<B>enable</b>" : "<B>disable</b>"] [src]'s burst fire mode."))
+
+/obj/item/weapon/gun/verb/use_toggle_burst()
+	set category = "Weapons"
+	set name = "Toggle Burst Fire Mode"
+	set desc = "Toggle on or off your weapon burst mode, if it has one. Greatly reduces accuracy."
+	set src = usr.contents
+
+	var/obj/item/weapon/gun/G = get_active_firearm(usr)
```
harryob, line 728: G -> active_firearm

code/modules/projectiles/gun_helpers.dm
```diff
@@ -683,51 +674,62 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
...
+	if(!G) return
```
harryob, line 729: newline for the early return

code/modules/projectiles/guns/shotguns.dm
```diff
@@ -1025,11 +1025,53 @@ can cause issues with ammo types getting mixed up during the burst.
 	return ..()
 
 //-------------------------------------------------------
+
+/obj/item/weapon/gun/shotgun/pump/dual_tube
+	name = "\improper generic dual-tube pump shotgun"
```
harryob, line 1030: do we need the improper? given its lowercase anyway

harryob (changes requested):

code/modules/projectiles/guns/shotguns.dm
NoKohi: nope

code/modules/projectiles/guns/shotguns.dm
```diff
@@ -1025,11 +1029,53 @@ can cause issues with ammo types getting mixed up during the burst.
...
+		to_chat(user, SPAN_WARNING("You must be holding \the [src] in your active hand to switch the active internal magazine!")) // currently this warning can't show up, but this is incase you get an action button or similar for it instead of current implementation
+		return
+	if(!current_mag)
+		return
+	if(current_mag == primary_tube)
+		current_mag = secondary_tube
+	else
+		current_mag = primary_tube
+	to_chat(user, SPAN_NOTICE("[icon2html(src, user)] You switch \the [src]'s active magazine to the [(current_mag == primary_tube) ? "<b>first</b>" : "<b>second</b>"] magazine."))
+	playsound(src, 'sound/machines/click.ogg', 15, TRUE)
```
stanalbatross, line 1066: this sound is kinda weird and doesn't fit the action imo

stanalbatross (changes requested):
the rest of the code is fine

PR #1636, opened 2022-11-20
Marines can now shake eachother up from non-weaken stuns.

carlarctg:
## About The Pull Request

The shitcode of shaking up used to only check for lying or sleeping, which meant that you couldn't actually shake up people who got STUN only, like if they were at the border of a screech. It *should* still have reduced effects, but the check inmediately goes into an else for hugging that returns, meaning the stun reduction doesn't pass....

I died to this and I'm malding super hard about it because it's so dumb. speedmrege

## Why It's Good For The Game

The ability that shakes marines up from stuns should shake marines up from stuns

## Changelog

:cl:
fix: Marines can now shake eachother out of their stupor (They can shake stunned players that aren't lying down)
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_attackhand.dm
+++ b/code/modules/mob/living/carbon/human/human_attackhand.dm
@@ -208,6 +208,7 @@
 	if (w_uniform)
 		w_uniform.add_fingerprint(M)
 
+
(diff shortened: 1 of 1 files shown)


fira (approved):
Makes sense to me

PR #1642, opened 2022-11-20
More automobiles

Lengord:
Added direction sprites for old car sprites and one new, and made them actual vehicles to drive them.

## About The Pull Request
I still had sprites from Cakey request for vehicles sprites, so why not to try. Copy-pasted some code, seems work ok'ish, you can drive them atleast. Also added cramped interriors for them.
Issues: 
1) When hardpoint damaged badly, on treads it looks like one large black spot
2) After vehicle broken overlay updates only when you install new hardpoint or wrench it to 100%
3) Still moves after wheels hardpoint broken
4) Probably it unbalanced
5) Can't make spawner preset to choose different vehicle icons if it is available
(coding not my thing. 100% i messed up stuff, if you can do code, please!!! do)

## Why It's Good For The Game

If it's bad for game atleast can be used for existing props, or for events. 

## Changelog

:cl:

add: Added more vehicles to drive, and interrior maps
imageadd: added direction sprites for some vehicles

/:cl:


monkeysfist101 (approved):
Sprites LFTM

code/modules/vehicles/truck/civvan.dm
```diff
@@ -0,0 +1,57 @@
+/obj/vehicle/multitile/civvan
+	name = "Civilian van"
+	desc = "A rather cheep hunk of metal with some wheels, you know what to do. Entrance on the sides."
```
harryob, line 3: cheap

code/modules/vehicles/truck/civtruck.dm
```diff
@@ -0,0 +1,65 @@
+/obj/vehicle/multitile/civtruck
+	name = "Civilian truck"
+	desc = "A rather cheep hunk of metal with some wheels, you know what to do. Entrance on the sides."
```
harryob, line 3: cheap

code/modules/vehicles/truck/crane.dm
```diff
@@ -0,0 +1,61 @@
+/obj/vehicle/multitile/crane
+	name = "Crane"
+	desc = "A rather cheep hunk of metal with some treads, you know what to do. Entrance on the left side."
```
harryob, line 3: cheap

harryob (changes requested):

code/modules/vehicles/truck/civvan.dm
Diegoflores31: typos are soul . CHEEP CAR

ThePiachu (changes requested):
Okay, a few things:

1) Please rebase since I'm getting a black screen when trying to run this
2) I don't think the new vehicle files were included in the .dme so you can't test them as is
3) I think all the trucks start without wheels and you can't use the van wheels on them to repair them, so maybe it would be good to at least get the cars working by default, and maybe have broken down versions as a sub-type

Otherwise, I like the idea of having more civilian vehicles. Maybe one day we could have something to transport actual cargo and all! So even if we don't use them straight away, it would be nice to have them.

PR #1648, opened 2022-11-21
Added a super snowflake vendor for devs and admins to play with

ThePiachu:
## About The Pull Request

I was doing some work on picking a set of clothes for a fun ERT when I realised we could use a vendor for admins / devs to figure out what kind of items look good together and so on before they spawn a bunch of mobs with them. So I added a few Super Snowflake Vendors that have all the items of a given type in them. The UI is a bit slow due to the number of items, and when I tried putting all of the items together into one vendor it lagged out the game, so I split them up into separate types.

With some feedback from carlarc, I also made the Super Snowflake vendor be easily customizable - if you spawn the base variant it will ask you to provide an item path to stock. You can also easily configure it while mapping to stock a given set of subitems.

## Why It's Good For The Game

It makes it easier to match interesting outfit combinations for ERTs and what have you. This can let admins communicate more clearly what they might want in a gear preset, and for devs to easier test how things look together.

## Changelog

:cl:
add: Added a set of Super Snowflake Vendors for admins and devs to play with. If you spawn the base one you can stock it with any item and its subtypes as defined when you first spawn it / interact with it.
/:cl:

### DIFF
--- a/code/game/machinery/vending/vendor_types/dress.dm
+++ b/code/game/machinery/vending/vendor_types/dress.dm
@@ -155,3 +155,84 @@
 			IT.add_fingerprint(usr)
 			LAZYADD(id_card.vended_items, item_path)
 			return TRUE
+
+//A clothing vendor for admins and devs to test all the clothes in the game
+/obj/structure/machinery/cm_vending/clothing/super_snowflake
+	name = "\improper Super Snowflake Vendor"
+	desc = "WARNING: The quantity of clothes contained within can slow down reality."
+	icon_state = "snowflake"
+	use_points = FALSE
+	show_points = FALSE
+	use_snowflake_points = FALSE
+	vendor_theme = VENDOR_THEME_COMPANY
+	vend_delay = 10
+	var/list/items
+
+/obj/structure/machinery/cm_vending/clothing/super_snowflake/get_listed_products(mob/user)
+	return items
+
+/obj/structure/machinery/cm_vending/clothing/super_snowflake/uniform
+	name = "\improper Super Snowflake Vendor, Uniforms"
+
+/obj/structure/machinery/cm_vending/clothing/super_snowflake/uniform/get_listed_products(mob/user)
+	if(!items)
+		items = list()
+		items += list(list("UNIFORM", 0, null, null, null))
+		for(var/obj/item/I as anything in typesof(/obj/item/clothing/under))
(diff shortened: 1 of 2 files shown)


fira:
I like the idea already but lemme put forward one even more crazy:
What if instead this was a blank vendor, and the adding functionality was a tool/verb rather than thru usage? So you can use it multiple times to make a custom vendor.
Might even be extended to other vendors, though i don't think most would support this properly...

ThePiachu:
Okay, so I changed the vendor like this:

1) It takes a list of objects in now, so it would be easy to dynamically make a vendor on the map or something
2) Admins got a verb to add items to the vendor, so they can create an empty vendor and populate it with stuff they want to give to people

This should be enough to make a custom vendor on the fly and have it dispense the needed items.

Not sure how we're handling such admin verbs, should it be in this file or somewhere else...

ThePiachu:
So @fira - is this how you wanted the vendor to function, or should I use some kind of different setup for admin adding new things to it?

harryob (approved):

code/game/machinery/vending/vendor_types/dress.dm
```diff
@@ -160,3 +160,100 @@
...
+	//If we don't have an object type, we ask the user to supply it
+	if(!item_types)
+		var/obj/item/chosen = get_item_category_from_mob(user)
+		if(!chosen)
+			return
+		item_types = list(chosen)
+		to_world("item_types - [item_types.len]")
+
+	if(!items)
+		to_world("item_types - [item_types.len]")
```
harryob, line 187: debug?

harryob (changes requested):

code/game/machinery/vending/vendor_types/dress.dm
ThePiachu: debug...

harryob (approved):

harryob (changes requested):
seems like every item is out of stock - regardless of if you use a preset vendor or add to one yourself?

ThePiachu:
Ah, my mistake, had to make it "use points" while still keeping everything at 0 points. Should be good now.

harryob (dismissed):
apologies - missed this the first time around: but if a type is selected with subtypes, every subtype is also added, and it's not possible to just add the specific type suggested, as far as i can tell. is this intentional?

harryob (approved):

PR #1649, opened 2022-11-21
Commanding Officer Survivor Role

morrowwolf:
## About The Pull Request

The newest map LV-522 has an official commanding officer survivor. Making it an explicit role will allow for a consistency in setting and atmosphere on the map as well as a new angle for the CO whitelist to explore. In the future we may branch out to leadership roles on other maps in different settings.

## Why It's Good For The Game

An expansion of the CO whitelist into survivor gameplay will allow for consistent and interesting dynamics that would otherwise not be explored effectively. This is already something that can occur in game, this PR formalizes it.

## Changelog

:cl: Morrow
add: Added a new role, Commanding Officer Survivor, for anyone CO whitelisted.
/:cl:


code/game/objects/effects/landmarks/survivor_spawner.dm
```diff
@@ -68,17 +69,6 @@
...
-/obj/effect/landmark/survivor_spawner/lv522_forecon_major
-	equipment = /datum/equipment_preset/survivor/forecon/major
-	intro_text = list("<h2>You are the commanding officer of the USS Hanyut.</h2>",\
-	"<span class='notice'>You are VERY aware of the Xenomorph threat.</span>")
-	story_text = "This entire operation went shits up when your team investigated the atmospheric processor two weeks ago, from that point on you've been surviving by the skin of your teeth waiting your ship's automated distress beacon to be picked up... until now."
-
-/obj/effect/landmark/survivor_spawner/lv522_forecon_major/check_can_spawn(var/mob/living/carbon/human/survivor)
-	if(RoleAuthority.roles_whitelist[survivor.ckey] & WHITELIST_COMMANDER)
-		return ..()
-	return FALSE
```
fira, line 80: You could add this as entry message for the new job. 
But, why make it a separate role if we have this? What'd i miss?

morrowwolf: Alright so let me try to break down my thought process.

The intro text and story text is very close to the default map story_text and creating a system for a CO surv only map specific story text seems like overdoing it for a tiny change.

The spawners are not consistent. LV-522 has like 12 spawners for survivors of which one was the LV-522 FORECON Major. The chances of someone with CO whitelist getting the role was pretty low and many people did not want to play regular survivor but instead just the CO survivor. 

This role allows for rolling *just* for the CO survivor as well as adding the building blocks for future map additions for "leadership" survivor roles. Nothing is set in stone on that side of the fence (if it even happens).

harryob (approved):
code looks good to me for TM

morrowwolf (commented):
This has been testmerged for four days and the mapping is literally a one line removal. I am SELF APPROVING mapping.

PR #1652, opened 2022-11-21
corrects egg + stair planing

harryob:
## About The Pull Request

eggs need to be on the game plane so they can't be mischievously hidden under literally anything else (though funny)
stairs need to be on the floor plane so weeds + other resin items don't appear under them. it also makes them look less glitchy

## Why It's Good For The Game

eggs could be placed under anything, now they cannot
stairs are less goofy looking, and can't have weeds secreted under them

## Changelog

:cl:
fix: corrected the layering for eggs and stairs.
/:cl:

### DIFF
--- a/code/game/objects/structures/misc.dm
+++ b/code/game/objects/structures/misc.dm
@@ -143,6 +143,7 @@ obj/structure/xenoautopsy/tank/larva
 	unacidable = TRUE
 	health = null
 	layer = TURF_LAYER
+	plane = FLOOR_PLANE
 	density = 0
 	opacity = 0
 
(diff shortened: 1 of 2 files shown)


fira (approved):

ThePiachu:
I would think hiding eggs behind junk is a feature not a bug. But yeah, depending on how much down the layers they are it might be a bug.

harryob:
> I would think hiding eggs behind junk is a feature not a bug. But yeah, depending on how much down the layers they are it might be a bug.

there was specific layering in place for this already, but the *plane* means that they are in completely different layering groups. the plane being moved to being on the floor plane means it will appear under *every* object. being on the game plane means it'll appear under specifically layered objects

PR #1653, opened 2022-11-21
Fixes machinery access

fira:
## About The Pull Request

Handling of machinery access causes issues - we've had these before but they seemingly never were really fixed:
 1. Machinery Topic does not honour statpanel click relay. This means a statpanel interact will *also* count as an action.
 2. Overall logic for accessing machinery is busted, but this isn't apparent unless the above happens. This fixes it by binding it to advanced tool users, with an override for queen and predalien usage of dropship console.

## Why It's Good For The Game

Longstanding, very bad bugs.

## Changelog

:cl:
fix: Fixed a bug with machinery mishandling clicks in statpanel.
fix: Machinery now require fine manipulation to be used. As of current this means Humans only (including Yautjas/Synth/Monkeys, but minus Zombies).
/:cl:

### DIFF
--- a/code/game/machinery/machinery.dm
+++ b/code/game/machinery/machinery.dm
@@ -200,20 +200,23 @@ Class Procs:
 	return (stat & (BROKEN|additional_flags))
 
 /obj/structure/machinery/Topic(href, href_list)
-	..()
+	. = ..()
+	if(.)
+		return TRUE
 	if(inoperable())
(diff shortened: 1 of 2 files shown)


harryob (approved):
roonie take the wheel

fira:
This need an extra fix to also allow preadliens from operating the console after hijacking it

PR #1666, opened 2022-11-22
Gives SEA endurance 2

DexterDude:
## About The Pull Request

Gives the SEA endurance skill 2 to put the on par with Squad Leaders and the XO/CO's skillsets.

 Geeves said that they didn't see a problem with these changes. 

https://discord.com/channels/150315577943130112/1044522693405458454/1044529732525490176

## Why It's Good For The Game

The Senior Enlisted Advisor is essentially the middle ground between command and squad leaders realistically speaking, and they're also well trained and likely very physically fit given their status as a pseudo "drill instructor" on the Almayer. It makes sense that they would get endurance 2 when squad leaders do as well.

## Changelog

:cl:
balance: Changes the Senior Enlisted Advisor's endurance skill from 1 to 2
/:cl:


fira:
Well all UPP has Endurance 4 and all CLF has Endurance 5! Might as well give all USCM Endurance 3 minimum then.

No more seriously, what for?

DexterDude:
> Well all UPP has Endurance 4 and all CLF has Endurance 5! Might as well give all USCM Endurance 3 minimum then.
> 
> No more seriously, what for?

Those are ERT roles that should be stronger than the shipside marines simply because there are far fewer than them, and as far as I know the CLF armor is kinda bad so the endurance makes up for it. I'm not sure that's really relevant to the PR, but I'm not a dev.

It's inconsistency that should be fixed. I stated in the PR that the SL and XO have it and the SEA is essentially a middleground between those two roles and they both have the same skill. As an addition if I understand correctly it'll also make them able to do more pushups since I feel the SEA should be able to do more than 15. I don't see an issue in adding it, personally, it'll hardly affect balance at all, it was just something I noticed while reading the wiki.

DexterDude:
no clue why it closed it, wasn't intentional

morrowwolf (approved):
Code approved. 7 day maint balance approval.

PR #1670, opened 2022-11-22
Fixed resin fruit disappearing magically from bloodstream

carlarctg:
## About The Pull Request

Nutrition property will no longer make the chem disappear if the owner is full, it'll just not process.

## Why It's Good For The Game

Having 13u of fruit resin then suddenly 0u because of this weird proc is bad.

## Changelog

:cl:
fix: Fixed resin fruit disappearing magically from bloodstream.
/:cl:


fira:
I think someone made a similar change for regular food and this was deemed intentional

carlarctg:
well it sucks! fruit resin gets deleted from the holder when it has other beneficial properties

harryob (approved):

PR #1679, opened 2022-11-23
Refactor eyesight into trait as well as adds opiate receptor deficiency trait

morrowwolf:
## About The Pull Request

This PR refactors poor eyesight into the trait system as well as adding a new trait called "opiate receptor deficiency". Both are in a new "biology" section under traits. 

Opiate receptor deficiency traits multiplies the potency of the "painkilling" effect of drugs by 0.25 (1/4th of the effect). It should almost always be a detriment to the user so as to not have balance concerns. **TO BE CLEAR: THIS TRAIT WAS A COMMISSION**

People now have two trait points to spend as poor eyesight has been moved over to it and robotic limbs cost one point now. This should not overly effect how traits function as robotic limbs are already mutually exclusive from one another and there are enough options to warrant it now.

Robotic limb preference group name was also reduced in size as the warning is clearly in the description of the trait and the name was too large/clunky/out of place.

## Why It's Good For The Game

Refactoring code into common systems makes them easier to manage and creates a system to be built off of in the future.

## Changelog

:cl: Morrow
add: Adds new trait "opiate receptor deficiency" which reduces potency of painkilling effects.
add: Adds a trait point (now at two) for use in character creation.
refactor: Refactors the poor eyesight preference into the trait system.
spellcheck: Removes extraneous wording around robotic limb preferences.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry_properties/prop_positive.dm
+++ b/code/modules/reagents/chemistry_properties/prop_positive.dm
@@ -196,20 +196,21 @@
 /datum/chem_property/positive/painkilling/process(mob/living/M, var/potency = 1, delta_time)
 	if(!..())
 		return
-
-	M.pain.apply_pain_reduction(PAIN_REDUCTION_MULTIPLIER * potency)
+	var/effective_potency = (CHECK_BITFIELD(M.disabilities, OPIATE_RECEPTOR_DEFICIENCY) ? potency * 0.25 : potency)
+	M.pain.apply_pain_reduction(PAIN_REDUCTION_MULTIPLIER * effective_potency)
 
 /datum/chem_property/positive/painkilling/process_overdose(mob/living/M, var/potency = 1, delta_time)
 	if(!..())
 		return
-
-	M.pain.apply_pain_reduction(PAIN_REDUCTION_MULTIPLIER * potency)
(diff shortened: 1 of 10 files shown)


code/modules/reagents/chemistry_properties/prop_positive.dm
```diff
@@ -196,20 +196,21 @@
 /datum/chem_property/positive/painkilling/process(mob/living/M, var/potency = 1, delta_time)
 	if(!..())
 		return
-
-	M.pain.apply_pain_reduction(PAIN_REDUCTION_MULTIPLIER * potency)
+	var/effective_potency = (CHECK_BITFIELD(M.disabilities, OPIATE_RECEPTOR_DEFICIENCY) ? potency * 0.25 : potency)
```
carlarctg, line 199: probably OOS but this would probably be best implemented as a mob/living variable that controls painkiller potency rather than checking for a character quirk for future Usage

carlarctg (commented):
contributor approved 👍

fira (approved):
code good
i don't think there's a neccessity to expand on this via either variables or elements right now, this can be done when usecases arise

this might need a balance tag, unsure

harryob (approved):

PR #1692, opened 2022-11-24
fixes power load values

### LINKED ISSUE #1688: Power load values not initializing, updating and working properly.
## Description
Power load values do not initialize, update and work properly. This issue appears to have been caused by the TP overhaul done on June 12, 2020, although not entirely(?)

Expected behavior:

- Power load values should go to 0 W when the APC power for them is turned off, they should go to their idle power or active power draw when turned on.

Actual behavior: 

- Equip and environmental power load values do not update for the most part. This is because machines (including machinery.dm itself, and power.dm), for the most part, do not make use of use_power or update_use_power, which in turn means that their power draw never changes. This means that no calculations are ever done for them to begin with.
- Lights change their active power usage based off luminosity (a quick fix would have this be based off brightness as this value does not change like luminosity does) and as a result will go into the negative when the power for them is turned off (you can keep doing this and the value will keep decreasing into the negative).
- Lights do not initialize their power load properly (it should be active power use + idle power use, however it seems to only be idle power use, maybe something more wonky). This means that the lights power load value would still go into the negative as they would subtract active power + idle power from idle power use (however this would not keep decreasing forever).
- There are other issues in general with associated procs and variables being used incorrectly (e.g. update_use_power(4080) in shuttle_console.dm).

## Steps to reproduce
Go to an APC. Turn it on/off, and look at the values for the power channels.

NoKohi:
## About The Pull Request

added two calls for update_use_power in power_change in power.dm

changes active_power_usage for lights to be based off brightness instead of luminosity and places it in initialize, as active_power_usage realistically shouldn't be fluctuating

there were two initialize procs for computers, I combined them into one because why not, security consoles now use update_use_power

## Why It's Good For The Game

this fixes #1688 - now if you turn off the power in an area, the total load will go to 0 W, things should use the proper amount of power when initializing (i.e. you shouldn't get any mystery extra power draw), lights won't magically go into the negative and more

## Changelog

:cl:
fix: fixed power load values
/:cl:

### DIFF
--- a/code/game/machinery/computer/camera_console.dm
+++ b/code/game/machinery/computer/camera_console.dm
@@ -81,7 +81,7 @@
 			concurrent_users += user_ref
 		// Turn on the console
 		if(length(concurrent_users) == 1 && is_living)
-			use_power(active_power_usage)
+			update_use_power(2)
 		// Register map objects
 		user.client.register_map_obj(cam_screen)
 		user.client.register_map_obj(cam_background)
@@ -207,7 +207,8 @@
 		current = null
 		last_camera_turf = null
 		range_turfs = list()
-		use_power(0)
+		if(use_power)
+			update_use_power(1)
 		STOP_PROCESSING(SSfastobj, src)
 	user.unset_interaction()
 
--- a/code/game/machinery/computer/computer.dm
+++ b/code/game/machinery/computer/computer.dm
@@ -17,9 +17,6 @@
 	. = ..()
 	if(processing)
 		start_processing()
-
-/obj/structure/machinery/computer/Initialize()
-	. = ..()
 	power_change()
 
 /obj/structure/machinery/computer/initialize_pass_flags(var/datum/pass_flags_container/PF)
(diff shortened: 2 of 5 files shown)


NoKohi:
there's one issue where if somebody has the UI for security consoles open, and you switch off and then on the power, the power draw will reset to idle power use; the solution to this would be to have the update_use_power at the same thing that updates the UI to it being functional for people using it, I might do this if I get around to it

however the alternative to this is to have all the power load values not work at all, as opposed to having this one thing not update properly (also security camera consoles have their own issue that the UI really ought to close or actually render static properly or w/e when you're far away from it, because as it stands you can just watch the active camera from any distance)

harryob:
> (also security camera consoles have their own issue that the UI really ought to close or actually render static properly or w/e when you're far away from it, because as it stands you can just watch the active camera from any distance - I'd fix this but the _cabal_ would come after me)

you will change the UI_STATUS for the camera console to _strict... NOW!!! (not in this pr)

ThePiachu (approved):

PR #1697, opened 2022-11-25
Fixed missing changelog

ThePiachu:
Fixed missing changelog


harryob:
note to self: make an action checking the PR changelog is valid if extant

PR #1704, opened 2022-11-26
Explosion stun is reduced by bomb armor. Xenos are now Knocked down instead of KO'd when exploded.

carlarctg:
## About The Pull Request

As title says.

Knockdowns are multiplied by 1 - bomb armor. Grenadier armor will roughly halve stun times, normal armor reduces them by roughly 12% due to preexisting armor values.
Knockouts are the same, but the total time has been halvedd, so marines spend some time KO'd, some time stunned, instead of inmediately getting up the second they wake up from unconsciousness.
The white-flash effect is also affected by the bomb armor multiplier.

Xenomorphs will now be able to see their surroundings while being juggled by grenades.

## Why It's Good For The Game

Explosions are very annoying, stupidly so at this point. It is genuinely a bad idea to carry grenadier armor as _the grenadier_ because it means that you will spend 10 years sleeping from an explosion while a runner tackle-drags you to hive, even if it wasn't you that caused the boom. I don't think any other servers have explosions as annoying as ours, and CM is probably the server with the most explosions per minute.

> Knockdowns are multiplied by 1 - bomb armor. Grenadier armor will roughly halve stun times, normal armor reduces them by roughly 12% due to preexisting armor values.
> Knockouts are the same, but the total time has been halved, so marines spend some time KO'd, some time stunned, instead of inmediately getting up the second they wake up from unconsciousness.
> The white-flash effect is also affected by the bomb armor multiplier.

Bomb armor should not be worse than no bomb armor because it'll mean getting captured. This is **insane.** This PR will hopefully lessen the amount of armorless grenadiers.

Grenadier aside, explosions are so, _so_ punishing as a marine, I don't see the issue with slightly reducing the stun times as a byproduct of actually having them be affected by armor. You already spend precious moments completely vulnerable on the ground knocked away from any help by several tiles, limping if you manage to actually get up, all of these things usually happening through no fault of your own, by a SADAR that misclicked, a backpack alpha throwing OT nades into a wall, a CAS pilot firing a direct lase minirocket without looking. 

Additionally, this paves the way for future marine bomb armor buffs. I have plans, when this is merged, to create a rework for heavy armor that lets it soak FF damage (bullets, explosions).

> Xenomorphs will now be able to see their surroundings while being juggled by grenades.

Being juggled by grenades as a Xenomorph is already intensely annoying, but having your vision completely removed is very, very frustrating as they cannot analyze the battlefield to decide their next action, especially when thanks to our wonderful codebase and tech debt the KO screen persists for seconds after waking up. I think it makes a lot of sense for the perfect biomechanical weapon not to take a nap every time a grenade nicks it.

I do see some possible issues arising from this, though. Warriors inmediately lunging after an explosion, Lurkers pouncing, etc. None of these are anything that wouldn't have happened from a normal slug stun, so I don't really see the problem. If absolutely necessary, being dazed could weaken the abilities to curb this, but I don't think it is required.

This PR isn't necessarily meant to buff either side, but more so just... reduce the frustration and annoyance caused by explosions, for every player.

## Changelog

:cl:
add: Explosion knockout is now half that of explosion knockdowns.
add: Bomb armor now reduces the stun and flash timer on explosions for marines. For grenadier armor, this means it's roughly halved.
add: Xenomorphs are no longer slept by explosions.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -122,15 +122,15 @@
 	if(lying)
 		severity *= EXPLOSION_PRONE_MULTIPLIER
 
-	if(severity >= 30)
-		flash_eyes()
+
 
 	var/b_loss = 0
 	var/f_loss = 0
 
 	var/damage = severity
+	var/bomb_armor = getarmor(null, ARMOR_BOMB)
 
-	damage = armor_damage_reduction(GLOB.marine_explosive, damage, getarmor(null, ARMOR_BOMB))
+	damage = armor_damage_reduction(GLOB.marine_explosive, damage, bomb_armor)
 
 	last_damage_data = istype(cause_data) ? cause_data : create_cause_data(cause_data)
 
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -147,20 +147,28 @@
...
+	var/bomb_armor_mult = ((CLOTHING_ARMOR_HARDCORE - bomb_armor) * 0.01)
+
+	if(severity >= 30)
+		flash_eyes(flash_timer = 4 SECONDS * bomb_armor_mult)
+
+	// Stuns are multiplied by 1 reduced by their medium armor value. So a medium of 30 would mean a 30% reduction.
+	var/knockdown_value = severity * 0.1
+	var/knockdown_minus_armor = min(knockdown_value * bomb_armor_mult, 1 SECONDS)
+	var/obj/item/Item1 = get_active_hand()
+	var/obj/item/Item2 = get_inactive_hand()
```
stanalbatross, line 160: we do not use capital letters in our var names carlarc

carlarctg: i didnt make these vars they wer alredy heree!!!!!

ThePiachu (approved):

PR #1709, opened 2022-11-26
Fixes a bad skill index for ghost previews

### LINKED ISSUE #1654: Bad skill index for a ghost preview
## Description

So I started using VS Code and it caught some weird error that doesn't get logged but is still a bit annoying. Apparently, if you set your role preference to someone with a Surgery Skill as High, it will trigger this error:

image

Which I'm guessing means the code is trying to set the preview's skills and is figuring out how to give it an overlay for the surgery skill?

It doesn't seem to affect anything else though.

VS Code shouldn't hang itself whenever I set a doctor to High.

## Steps to reproduce

1) Load game with VS Code Debugger
2) Set role preference of CO / Synth / Doctor / etc. to High
3) Click "Done"
4) The Debugger will stop to display the error.

ThePiachu:
## About The Pull Request

Fixes #1654 , means less runtime errors for ghost previews and transforming a person into a xeno.

## Why It's Good For The Game

Less runtime errors good, plus makes debugging easier since this error comes up pretty often.

## Changelog

:cl:
fix: Fixed a runtime error around selecting a role with high surgery skill.
/:cl:

### DIFF
--- a/code/datums/skills.dm
+++ b/code/datums/skills.dm
@@ -171,11 +171,15 @@
 
 // Returns the skill DATUM for the given skill
 /datum/skills/proc/get_skill(var/skill)
+	if(!skills)
+		return null
 	return skills[skill]
 
 // Returns the skill level for the given skill
 /datum/skills/proc/get_skill_level(var/skill)
 	var/datum/skill/S = get_skill(skill)
+	if(!S)
+		return -1
 	if(QDELETED(S))
 		return -1
 	return S.get_skill_level()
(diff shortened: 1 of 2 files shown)


fira:
Why do they end up without skills? Shouldn't they always have em'?

ThePiachu:
I'm slowly piecing together why this triggers over at the original error's Issue...

ThePiachu:
Alright, figured it out - turns out we were qdelling old skills and doing a skillcheck on them, which is a bad. So added that fix in as well to assign it to null after a qdel and now it seems to work!

morrowwolf (approved):

stanalbatross (approved):

PR #1667, opened 2022-11-22
Intel skill, IO phone, SO access, pilot area vendor access

ThePiachu:
## About The Pull Request

Previous version of this got closed automatically ( #1412 ) so improved it a little bit with some other tweaks.

So in general:
- Gave basic Intel skills to POs (only Officers that don't have it, a hecking CMO has that skill so they should too), SLs and RTOs (just so someone in any given squad would be able to help IOs)
- Added a phone to IO room so they can phone squads about where to pick up intel from or how to upload data from a computer.
- Added a telephone backpack to IO gear loadouts. I saw an IO rolling with it and it was a pretty neat idea.
- Gave SOs access to Marine Prep and Requsitions so they can fill in for a CT at roundstart if there are none (happened to me twice in the last week playing an SO!)
- Finally, fixed access to the three vendors in the north hangar in the pilot area. Obviously they were meant to be accessed by the POs and DCCs, but their access was all over the place. So gave them the regular access they have (ACCESS_MARINE_ENGINEERING, ACCESS_CIVILIAN_ENGINEERING) as well as the access someone tried to map in but didn't do it properly for POs, DCCs and VCs - ACCESS_MARINE_PILOT, ACCESS_MARINE_CREWMAN

Previously the telephone mapping being added to IO area was approved by Agameofscones.

## Why It's Good For The Game

Letting some Marines have access to intel while not letting it be a general thing will hopefully lead to a bit more intel being picked up and processed closer to the frontlines. Emphasizing the telephone as a tool for IOs will hopefully help with the communication between IOs and the Marines. SO access to Requisitions will alleviate the headache of having no CTs at roundstart that sometimes happen. Fixed vendor access will mean pilots and vehicle crew will have ready access to the tools they need.

## Changelog

:cl:
add: Added a telephone to IO's room. Now you can call the RTOs on the field to get you your intel!
add: IOs can now pick a radio telephone as their starting gear if they prefer to focus on communication over having extra backpack space.
balance: Gave POs, SLs and RTOs basic Intel skill so they can see the Intel Objectives.
balance: Gave SOs access to Marine Prep and Requisitions (so they can fill in as CTs at the start of the round if there are none)
fix: Fixed the access to the three vendors in the pilot area of the hangar, they should now be accessible to POs and DCCs as was intended.
/:cl:

PR #1711, opened 2022-11-26
Gave SOs access to basic Medical areas

ThePiachu:
## About The Pull Request

Previous PR of this: #1667

Gave SOs access to basic Medical since they have the skill to utilise it and it's a popular option during deployment.

## Why It's Good For The Game

Giving SOs Medical access will let them put their medical skill to better use.

## Changelog

:cl:
balance: Gave SOs access to basic Medical areas.
/:cl:


stanalbatross (changes requested):
`Gave SOs access to Marine Prep and Requisitions`
This is not desirable as it may lead to SOs ignoring orders to gear up for deployment without CIC approval, the main reason they do not get access by default is so that they have to go through another person to ensure they aren't just doing it to get attachments etc for deployment. Also may lead to early round powergame as a spare CT in req rather than focus on main role / job. Please drop this part, the medical access is fine though

PR #1712, opened 2022-11-26
Gave the Intel skill for PO, SL, Radio

ThePiachu:
## About The Pull Request

Previous PRs of this: #1412 #1667

A few more people got the Intel skill. POs mainly because they are the only Officers without it (hecking CMO has it, I think POs deserve it more than them). SLs got it because they are on the way to being SOs, so it makes a bit of sense. I don't expect them to be using it much though seeing how they are in the thick of things. RTOs got it because they are meant to be communicating with everyone, so they have a more natural progression into being IOs, AND they are more likely to have a bit of time to check the Objectives than the SLs (another reason for SLs getting the skill - RTOs having it).

## Why It's Good For The Game

Letting some Marines have access to intel while not letting it be a general thing will hopefully lead to a bit more intel being picked up and processed closer to the frontlines.

## Changelog

:cl:
balance: Gave POs, SLs and RTOs basic Intel skill so they can see the Intel Objectives.
/:cl:


stanalbatross (changes requested):
please remove from RTO, they get too much free time / don't have enough responsibilities and them getting intel skill will likely lead to powergame / them potentially focussing on intel gathering rather than the main focus of their role

ThePiachu:
Removed the skill from RTOs

PR #1406, opened 2022-11-02
Tweaks to Pressurized Reagent Canister Pouch, Research autoinjector QoLs, etc.

ThePiachu:
## About The Pull Request

I took a stab at improving the quality of life around Pressurized Reagent Canister Pouches. I was disappointed the empty ones didn't start with an autoinjector, so I added that. That required some code refactor (partially due to the oxycodone variant being a bit snowflak-y for ERTs).

Then I had an idea it would be nice to support people using two of those, so I added a tint to their tanks to display what chemicals they use (it's a mix of the chemical and the blue-ish canister, so the colours aren't 100% accurate, but I didn't want to mess with the icons so I'll leave that for spriters to tweak as needed). The canisters themselves also get tinted based on their contents.

I noticed the canisters themselves were a little bit exploitable - you could fill them up pretty easily with regular beakers and the flag `CAN_BE_DISPENSED_INTO` that seems to have been made for them wasn't doing anything, so I fixed that. Now those canisters can only be filled by putting them directly into the chem dispenser or using the big tanks. With that in place, I gave the pouch a right-click option to remove the canister. This will let people carry a spare canister and swap them as needed if they really like those chems. To that end, the spare canisters are now also medi-lathable (plus a slight adjustment of the pouch cost to indicate it contains both the canister and the pouch).

I noticed the autoinjector options in the medilathe were a bit thin when it came to the one-jab autoinjectors and the really small injectors, so I added more options for Researchers to dose their patients (down to 1u injectors for those really low OD chemicals).

Since Research seems to like their janitor buckets, I added an option of printing them in regular lathes. They cost twice as much as mop buckets.

I noticed only the squad medics come with the empty Pressurized Reagent Canister Pouch, so added them to the medical staff.

Finally gave Synths some love and gave them most tools to be robust - all the variants of the Pressurized Reagent Canister Pouch everyone else gets, plus the three sizes of empty autoinjectors, and the MS-11 Smart Refill Tank only squad medics got. I'm sure they will be pleased and will be able to utilise those oversized injectors for something useful.

## Why It's Good For The Game

I think this is good for the game because it addresses a few of the frustrations I felt when playing with those tools. It makes them easier to use, avoids the small trap of getting the Empty Pressurized Reagent Canister Pouch without an autoinjector and now knowing what to do with it, and gives the Synths the toys they were missing to bring them in-line with the rest of the staff.

## Changelog

:cl:
qol: Empty Pressurized Reagent Canister Pouch now comes with an empty Autoinjector.
[...]

PR #1722, opened 2022-11-26
Fixes to Pressurized Canister behaviour and letting you remove the Canister from its pouch

ThePiachu:
## About The Pull Request

Originally part of #1406

While messing with the Pressurized Reagent Pouch I realised the Pressurized Canister inside of it behaved a bit strangely. After reading the code I think I got to the gist of what was supposed to happen, fixed things and them improved the QoL of dealing with them.

First, now you can remove the Pressurized Canister from the pouch with a right click. It was really strange that they popped out so easily from the Pouch into the Dispenser but otherwise you couldn't get them out.

Secondly, made sure the Pressurized Canister's flags were properly respected. It's a SEALED container, BUT you can put reagents into it via a Dispenser or the Reagent Tank. So I had to fix the code around that to make it happen. Otherwise, you could pour things from beakers into the Pressurized Canister once you got it out which didn't seem intentional.

Thirdly, removed the invalid transfer amount list from the Pressurized Canister.

Fourthly, a final QoL - empty Reagent Tanks start by accepting chems rather than dispensing them to remove the unnecesary step when dealing with them.

## Why It's Good For The Game

Improves the QoL around Pressurized Canisters and Reagent Tanks, also fixes a bug resulting in broken behaviour for the Pressurized Canister which is always good.

## Changelog

:cl:
qol: You can now remove the Pressurized Canister from the Pressurized Reagent Canister Pouches by right-clicking on it.
fix: The Pressurized Canister can now only be filled through the chem dispensers and reagent tanks as it was intended.
fix: The Pressurized Canister doesn't have an invalid list of possible transfer amounts anymore.
qol: Empty reagent tanks are set by default to accept chems rather than try dispensing nothing.
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -128,6 +128,9 @@ directive is properly returned.
 /atom/proc/is_open_container()
 	return flags_atom & OPENCONTAINER
 
+/atom/proc/is_open_container_or_can_be_dispensed_into()
+	return flags_atom & OPENCONTAINER || flags_atom & CAN_BE_DISPENSED_INTO
+
 /atom/proc/can_be_syringed()
 	return flags_atom & CAN_BE_SYRINGED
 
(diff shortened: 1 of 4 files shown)


fira (approved):
I'd write a paragraph about the addition of an /atom proc for this if this nonsense wasn't an atom flag to begin with (GAAHH)

PR #1723, opened 2022-11-26
Reagent Canister and Pouch overlays

ThePiachu:
## About The Pull Request

Originally part of #1406

I have tried going into the field with two separate Pressurized Reagent Canister Pouches (like Tricord and Revival), but what I found frustrating was that it was a bit hard to tell them apart at first glance. So I made the Pressurized Canister change its tint based on what's in it, and similarly let the Canister part of the Pouch change tint accordingly so now you can tell which chem is where more easily.

## Why It's Good For The Game

It makes it easier for doctors / medics to use the correct chems by just glancing at their pouches.

## Changelog

:cl:
qol: All Pressurized Reagent Canister Pouches now change their canister tint based on the colour of the chemicals stored therein (while still being mixed with the default blue-ish tint of the canister). Canisters themselves similarly change their tint based on contents.
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/glass.dm
+++ b/code/game/objects/items/reagent_containers/glass.dm
@@ -549,8 +549,10 @@
 	update_icon()
 
 /obj/item/reagent_container/glass/pressurized_canister/update_icon()
+	color = COLOR_WHITE
 	if(reagents)
 		color = mix_color_from_reagents(reagents.reagent_list)
+	..()
 
 /obj/item/reagent_container/glass/bucket
 	desc = "It's a bucket. Holds 120 units."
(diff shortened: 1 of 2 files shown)


code/game/objects/items/reagent_containers/glass.dm
```diff
@@ -539,6 +544,15 @@
 	to_chat(usr, SPAN_WARNING("[src] has no transfer control valve! Use a dispenser to fill it!"))
 	return
 
+/obj/item/reagent_container/glass/pressurized_canister/on_reagent_change()
+	update_icon()
+
+/obj/item/reagent_container/glass/pressurized_canister/update_icon()
+	color = COLOR_WHITE
+	if(reagents)
+		color = mix_color_from_reagents(reagents.reagent_list)
```
fira, line 553: Confused - do you change both the item and its contents separately?

ThePiachu: You mean updating the Pressurized Reagent Canister Pouch and the Pressurized Canister? Since you can smack the Pouch against a Reagent Tank to change the contents of the Canister, then take the Canister out through a chemical dispenser (or with the other PR, a right click) so you could see it having a wrong tint.

fira (approved):

PR #1724, opened 2022-11-26
Fixes the Ghillie suit  OBJ having four directions

theselfish:
## About The Pull Request

If you throw the Ghillie suit, it has a direction. Like the mob version.

## Why It's Good For The Game

It's weird, and shouldn't be like that.

## Changelog

:cl:
fix: Ghillie suit no longer has a direction when thrown.
/:cl:


Agameofscones (approved):
lgtm

PR #1755, opened 2022-11-29
M39 Rebirth

Diegoflores31:
## About The Pull Request

Changes some mechanics of how M39 works and puts it on pair with other primary weapons winout making it the best sidegun in the game.

**M39 Holster**
Reduced Maximun Weight class to 4 . which will now require you to fold your stock and not have any attachment like grips on your weapon in order to store it.

**Foldable and Solid stock**
Took away slowdown from foldable stock . reason ? currently M39 was SLOWER than L42 with stock( 0.35 and 0.45) . this should aleaviate the speed difference betwen the best chasing weapon and M39.

Folded M39 stock no longer makes your weapon worse than having it stockless. making it be in pair with the other folding stocks.

Solid M39 stock received some stats from Folded M39 stock to make it an "upgrade" in exchange of the inability to be stored in belts + extra Wield delay.

Folded M39 Stock wield delay changed from Fast to Very Fast to compensate the inability of using grips.

Foldable and Solid stock now provide -1 and -2 burst scatter reduction while active.

## Why It's Good For The Game

This PR aims to make M39 a Primary weapon that is actually worth using winout turning it into a laser beam of death like it used to be.

**Why the holdster nerf ?**
while it provides a cool mechanic for the weapon it also restricts the abuse of this weapon preventing 3 weapon builds and similars and rewards players that use a entire weapon slot for the use of the weapon.
(Will add an audio effect soon)

**slowdown and wield delay**
While this weapon was supposed to be a chasing weapon . it was somehow slower than a Marksman rifle thus making marines loose chases with the gun and eventually just discarting it as a valid option . this PR aims to fix that issue and instead transfer the slow to the Solid stock.

## Changelog

:cl:
balance: M39 Holster now requires the weapon to be folded and winout attachments that change its size.
balance: Adjusted Folding and Solid Stock stadistics to be in pair with other stocks.
balance: M39 is able to use vertical grips again.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -1502,22 +1502,23 @@ Defined in conflicts.dm of the #defines folder.
 	attach_icon = "smgstock_a"
 	pixel_shift_x = 42
 	pixel_shift_y = 11
-	wield_delay_mod = WIELD_DELAY_FAST
 	hud_offset_mod = 5
 
 /obj/item/attachable/stock/smg/New()
 	..()
 	accuracy_mod = HIT_ACCURACY_MULT_TIER_7
-	recoil_mod = -RECOIL_AMOUNT_TIER_3
-	scatter_mod = -SCATTER_AMOUNT_TIER_6
+	recoil_mod = -RECOIL_AMOUNT_TIER_4
+	scatter_mod = -SCATTER_AMOUNT_TIER_8
+	burst_scatter_mod = -2
 	delay_mod = 0
+	wield_delay_mod = WIELD_DELAY_FAST
 	movement_onehanded_acc_penalty_mod = -MOVEMENT_ACCURACY_PENALTY_MULT_TIER_5
 	aim_speed_mod = CONFIG_GET(number/slowdown_low)
 
 
 /obj/item/attachable/stock/smg/collapsible
 	name = "submachinegun folding stock"
-	desc = "A Kirchner brand K2 M39 folding stock, standard issue in the USCM. The stock, when extended, reduces recoil and improves accuracy, but at a reduction to handling and agility. Seemingly a bit more effective in a brawl. This stock can collapse in, removing almost all positive and negative effects, however it slightly increases spread due to weapon being off-balanced by the collapsed stock."
+	desc = "A Kirchner brand K2 M39 folding stock, standard issue in the USCM. The stock, when extended, reduces recoil and improves accuracy, but at a reduction to handling and size. Seemingly a bit more effective in a brawl. This stock can collapse in, removing all positive and negative effects."
 	slot = "stock"
 	melee_mod = 10
 	size_mod = 1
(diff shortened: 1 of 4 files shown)


code/modules/projectiles/magazines/smgs.dm
```diff
@@ -64,6 +64,7 @@
 	name = "\improper M39 HV extended magazine (10x20mm)"
 	desc = "A 10x20mm caseless HV extended submachinegun magazine. Powerful propellant allows the bullet increased travel speed and minor penetration capabilities, noticeably improving its efficacy at long ranges, although it still suffers significantly compared to a rifle bullet."
 	max_rounds = 72
+	w_class = SIZE_MEDIUM // big ammo big size DUHH
```
carlarctg, line 67: add this to other extended smg mags if they have a significantly bigger magsize too

Diegoflores31: already

Diegoflores31: someone planned ahead of me and all big mags were already MEDIUM size with defines on each one

stanalbatross (changes requested):
I will start off by saying the mag size change is a MASSIVE buff and completely trashes storage economy/balance for SMG ammo. The mags already hold 48 rounds. The SMG does not need a mag buff. Remove this.

Diegoflores31:
**update**
reverted smoll magazines as per GA request

code/modules/projectiles/gun_attachables.dm
```diff
@@ -1516,26 +1517,24 @@ Defined in conflicts.dm of the #defines folder.
...
+	wield_delay_mod = WIELD_DELAY_VERY_FAST
+	burst_scatter_mod = -1
 	delay_mod = 0
 	movement_onehanded_acc_penalty_mod = -MOVEMENT_ACCURACY_PENALTY_MULT_TIER_5
 	//it makes stuff much worse when one handed
 	accuracy_unwielded_mod = -HIT_ACCURACY_MULT_TIER_3
 	recoil_unwielded_mod = RECOIL_AMOUNT_TIER_4
 	scatter_unwielded_mod = SCATTER_AMOUNT_TIER_10
-	//but at the same time you are slowish when 2 handed
-	aim_speed_mod = CONFIG_GET(number/slowdown_low)
```
stanalbatross, line 1530: dont entirely remove this, you can reduce it but do not remove it wholly

stanalbatross (changes requested):

code/modules/projectiles/gun_attachables.dm
```diff
@@ -1484,22 +1484,23 @@ Defined in conflicts.dm of the #defines folder.
...
 	delay_mod = 0
+	wield_delay_mod = WIELD_DELAY_FAST
 	movement_onehanded_acc_penalty_mod = -MOVEMENT_ACCURACY_PENALTY_MULT_TIER_5
 	aim_speed_mod = CONFIG_GET(number/slowdown_low)
 
 
 /obj/item/attachable/stock/smg/collapsible
 	name = "submachinegun folding stock"
-	desc = "A Kirchner brand K2 M39 folding stock, standard issue in the USCM. The stock, when extended, reduces recoil and improves accuracy, but at a reduction to handling and agility. Seemingly a bit more effective in a brawl. This stock can collapse in, removing almost all positive and negative effects, however it slightly increases spread due to weapon being off-balanced by the collapsed stock."
+	desc = "A Kirchner brand K2 M39 folding stock, standard issue in the USCM. The stock, when extended, reduces recoil and improves accuracy, but at a reduction to handling and size. Seemingly a bit more effective in a brawl. This stock can collapse in, removing all positive and negative effects."
```
stanalbatross, line 1503: reduction to size? this makes no sense

stanalbatross (changes requested):

code/modules/projectiles/gun_attachables.dm
Diegoflores31: It was removed to match current L42 speed . From 0.45 to 0.35 same as L42

MistChristmas: Maybe he means `aim_speed_mod = 0`

Diegoflores31: Aim speed mod = 0 is the same than not having any since 0 is the default

code/modules/projectiles/gun_attachables.dm
Diegoflores31: Got it . Will modify it tmorrow

DesertFox2:
Maybe instead of adding a slowdown to a weapon, that is meant to be a chaser, perhaps fix the marksman rifle being a better chaser than the M39? (Which it isn't supposed to be.)

IMO that change just absolutely kills the M39 from its literal purpose, which is to fight at close range and be able to chase down faster Xenos.

Diegoflores31:
> Maybe instead of adding a slowdown to a weapon, that is meant to be a chaser, perhaps fix the marksman rifle being a better chaser than the M39? (Which it isn't supposed to be.)
> 
> IMO that change just absolutely kills the M39 from its literal purpose, which is to fight at close range and be able to chase down faster Xenos.

You are not reading correctly . I am making it faster not slower

harryob (approved):
the code will function, balance is another matter

PR #1650, opened 2022-11-21
Cargo ammo and ammo box mapping

ThePiachu:
## About The Pull Request

When I was playing cargo I spent a good half an hour in a round just mindlessly packing ammo magazines into boxes in squad prep. It didn't put a dent in the squad prep supply, and I barely got a handful of boxes. So I thought to myself that this is pretty much a waste of time for cargo and decided to code a better solution:

https://www.youtube.com/watch?v=cnXcEYAV8P4

So now select vendors (opt-in via `VEND_LOAD_AMMO_BOXES`) support ammo consolidation. They count the number of ammo magazines you have and from that they derive the number of magazine boxes you can vend. If you vend a magazine, it updates the number of boxes available, and if you vend a box, it updates the number of magazines available (as well as all derived boxes - see the 3 pack of grenades and 25 pack box).

The `item_to_box_mapping` tracks ammo boxes (minus loose ammo), grenade boxes, grenade packets and mine boxes.

Most notable affected vendors - Requisition ammo vendor, Requisition vendor that features grenades, Squad vendors that have ammo in them.

So now Requisition will be able to easily raid Squad Vendors to stock up their ammo drops and save countless hours of mind-numbing cargo work.

This code ALSO correctly works when you're re-stocking a vendor with either individual magazines or magazine boxes. Correct amounts are updated everywhere. So you *could* take a magazine box, put it in a vendor and thus let people vend 16 magazines out of it seemlessly. Really useful just incase you need to restock Requisitions with individual ammo or something...

Other notes:
- Boxes of magazines are put directly under the corresponding ammo so you can vend them in larger amounts easier. Useful for 3-packs of grenades
- We should add a Shotgun Shell Box Box so we can also handle those easily...
- Nailgun ammo box had to be converted from being /smg/ since that created an invalid ammo box that nobody used.
- Nulled out a magazine type for an intermediary box that later gets used for MREs and all that
- Widened the display for ammo mags to 3 digits since we might hit those numbers for some types of ammo now.

## Why It's Good For The Game

Gameplay around loading magazines into ammo boxes is not interesting, so cutting it down to minimum is for the best.

## Changelog

:cl:
add: Added an automated ammo box management system to various vendors stocking bulk ammo and grenades. It will automatically combine ammo magazine into boxes, and divide boxes into individual magazines (or grenades, MRE packets, etc.).
qol: Cargo will no longer need to pack individual vended ammo magazines into boxes thanks to the ammo box management system. Your chains have been broken!
[...]

PR #1759, opened 2022-11-29
Cargo ammo and ammo box mapping (re-up)

ThePiachu:
## About The Pull Request

Previous version of this PR ( #1650 ) claimed to have changed 384 files, which would be impossible to review. So re-uploading this PR with hopefully a sane amount of files changed...

---

When I was playing cargo I spent a good half an hour in a round just mindlessly packing ammo magazines into boxes in squad prep. It didn't put a dent in the squad prep supply, and I barely got a handful of boxes. So I thought to myself that this is pretty much a waste of time for cargo and decided to code a better solution:

https://www.youtube.com/watch?v=cnXcEYAV8P4

So now select vendors (opt-in via `VEND_LOAD_AMMO_BOXES`) support ammo consolidation. They count the number of ammo magazines you have and from that they derive the number of magazine boxes you can vend. If you vend a magazine, it updates the number of boxes available, and if you vend a box, it updates the number of magazines available (as well as all derived boxes - see the 3 pack of grenades and 25 pack box).

The `item_to_box_mapping` tracks ammo boxes (minus loose ammo), grenade boxes, grenade packets and mine boxes.

Most notable affected vendors - Requisition ammo vendor, Requisition vendor that features grenades, Squad vendors that have ammo in them.

So now Requisition will be able to easily raid Squad Vendors to stock up their ammo drops and save countless hours of mind-numbing cargo work.

This code ALSO correctly works when you're re-stocking a vendor with either individual magazines or magazine boxes. Correct amounts are updated everywhere. So you *could* take a magazine box, put it in a vendor and thus let people vend 16 magazines out of it seemlessly. Really useful just incase you need to restock Requisitions with individual ammo or something...

Other notes:
- Boxes of magazines are put directly under the corresponding ammo so you can vend them in larger amounts easier. Useful for 3-packs of grenades
- We should add a Shotgun Shell Box Box so we can also handle those easily...
- Nailgun ammo box had to be converted from being /smg/ since that created an invalid ammo box that nobody used.
- Nulled out a magazine type for an intermediary box that later gets used for MREs and all that

## Why It's Good For The Game

Gameplay around loading magazines into ammo boxes is not interesting, so cutting it down to minimum is for the best.

## Changelog

:cl:
add: Added an automated ammo box management system to various vendors stocking bulk ammo and grenades. It will automatically combine ammo magazine into boxes, and divide boxes into individual magazines (or grenades, MRE packets, etc.). The boxes will appear at the bottom of the vendor (yes, this also includes the regular grenade boxes that used to be higher).
qol: Cargo will no longer need to pack individual vended ammo magazines into boxes thanks to the ammo box management system. Your chains have been broken!
qol: Requisitions vendor now stocks 3-packs of grenades as well as individual HEDP grenades.
qol: Requisitions ammo vendor now can vend a lot more individual magazines (actual number of magazines remains unchanged, just the ammo boxes have been consolidated into magazines).
qol: Requisition vendors now vend to floor when they are not vending to the front desks. This will make filling crates of ammo boxes or rappels easier.
code: Minor changes to code around some ammo boxes to remove one phantom box and prevent intermediate box types from being indexed when they shouldn't be.
code: Refactored the code that checks whether items are in mint enough condition to re-stock.
/:cl:


ThePiachu:
Testmerge went 95% well, one exploit found

ThePiachu:
As some people requested, ammo / grenade / etc. boxes now have their own category on the bottom. This has a small downside of moving the grenade box that used to be higher down to the bottom, but well, consistency is consistency...

ThePiachu:
I think a final tweak for Req for now - if the req vendors don't vend to the front desk, they vend to the turf rather than to the vendor location. This means requisition can more easily load up crates with all the ammo boxes or rappels they have since they don't need to pick each one up.

ThePiachu:
Fixed the bug of APC ammo vendor not handling the ammo / ammo boxes correctly. Turns out I was missing a flag check. Also made the APC ammo vendor compatible with this PR so now you can restock the vendor quite easily with the boxes.

stanalbatross (changes requested):
As an architect I'm going to have to say that the automatic ammo box thing can't go in. It's way too much streamlining and trivialises making extremely large caches of ammo for Marines without any interaction from req or effort put in to actually fill the boxes. I've seen the fob be crammed full of ammo boxes during the TM because of this purely from PFCs using the automatic box system in their vendors. The system simply makes it too easy to spam out a ton of ammo boxes.

ThePiachu:
Hmm, what if we disabled this for the main squad vendors? So you can't easily get infinite ammo boxes, but requisition still benefits from being able to more easily vend individual grenades, grenade packets, rarer ammo boxes, etc. It would also let vendors like the APC restock more easily with ammo boxes.

ThePiachu:
Removed the box consolidation from cargo prep vendors.

fira (approved):

PR #1763, opened 2022-11-29
fixes some backend validation for the bioprinter

### LINKED ISSUE #1747: Nullspace Item Spawning: Bioprinter
## Description

Bioprinter allows for any path to be spawned. The player has no control over where this object spawns, it will always spawn in null space. Potentially game altering if they spawn something like a queen that can be player controlled and buff other xenos possibly. May also be possible to get things out of nullspace.

Spawn only items in a list.

harryob:
## About The Pull Request

closes #1747 

## Why It's Good For The Game

ehh

also i'm feeling dumb and there's def a more obvious way... maybe.

## Changelog

:cl:
fix: added some additional validation to bioprinter spawning
/:cl:

### DIFF
--- a/code/game/machinery/bioprinter.dm
+++ b/code/game/machinery/bioprinter.dm
@@ -120,6 +120,8 @@
 					valid_recipe = TRUE
 					break
 			if(!valid_recipe)
+				log_admin("[key_name(usr)] attempted to print an invalid recipe ([recipe]) on \the [src].")
+				message_staff("[key_name(usr)] attempted to print an invalid recipe ([recipe]) on \the [src].")
 				return FALSE
 			var/datum/bioprinter_recipe/recipe_datum = new recipe
 			stored_metal -= recipe_datum.metal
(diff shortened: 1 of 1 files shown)


CocaColaTastesGood:
I just realized how weird this code is. It stores an instance of each recipe, but then proceeds to create another recipe? This is a decent solution. It'd need a refactor to be better I believe IE: store the ref of each product in TGUI, when player presses craft button it uses that ref to make the product/get cost by using an in scope locate 'locate(params["product_ref"]) in products'

code/game/machinery/bioprinter.dm
```diff
@@ -114,6 +114,13 @@
 	switch(action)
 		if("print")
 			var/recipe = params["recipe_id"]
+			var/valid_recipe = FALSE
+			for(var/datum/bioprinter_recipe/product_recipes in products)
+				if(product_recipes.type == text2path(recipe))
+					valid_recipe = TRUE
+					break
+			if(!valid_recipe)
```
mullenpaul, line 122: Maybe right out an admin message saying an illegal recipe was requested. Probably excessive but it would help monitor whats going on.

CocaColaTastesGood:
The message sends unsanitized text to admins

PR #1765, opened 2022-11-29
Master Controller backport (also tweaks Database, Early Init Runtimes, Speech Throttling)

fira:
## About The Pull Request

This ports updates to the Master Controller to our codebase, last update having been done 2 years ago.

Due to large scope a number of other things are affected directly or indirectly:

 * Updated MC, SSgarbage, and SStimer - this notably brings MC init stages
 * Database QM/EM load in early init stage, allowing players to connect during game loading
 * Rewrote chat spam prevention, using a per-character budget. Old system prevented speaking during init.
 * SSearlyruntimes added, registering early-init errors so they can be seen in STUI

This also tweaks/fixes extra small things:
 * Ping system won't tell you you're disconnected during loading anymore (it early inits)
 * Object camera updates are not ran during early-init anymore
 * Shuffled and added some /tg/ defines we can use, including PROC_REF and CALL_RUSTG

## Why It's Good For The Game

Better handling of scheduling, fixes several issues and increases possibilities (especially via init stages), plus just generally easing porting of /tg/ code

## Changelog

:cl:
code: Updated the Master Controller from /tg/
code: Database subsystems now load early, letting players join during game initialization.
code: Rewrote speech spam filter, using a character budget over time. Old one wouldn't let you speak during init. Yes, you should still be able to use copypastas (within reason)
code: Changed error handling to work during static init, so early errors can be displayed in STUI instead of forgotten.  
/:cl:


harryob (approved):
something will break 🤯

code/_byond_version_compat.dm
```diff
@@ -0,0 +1,49 @@
+// This file contains defines allowing targeting byond versions newer than the supported
+
+//Update this whenever you need to take advantage of more recent byond features
+#define MIN_COMPILER_VERSION 514
+#define MIN_COMPILER_BUILD 1556
```
stanalbatross, line 5: Why add this when we have the build blacklist system, that also has min version stuff…

code/controllers/mc/admin.dm
```diff
@@ -53,6 +63,35 @@ INITIALIZE_IMMEDIATE(/obj/statclick)
...
+	var/list/controllers = list()
+	var/list/controller_choices = list()
+
+	for (var/datum/controller/controller in world)
+		if (istype(controller, /datum/controller/subsystem))
+			continue
+		controllers["[controller] (controller.type)"] = controller //we use an associated list to ensure clients can't hold references to controllers
+		controller_choices += "[controller] (controller.type)"
+
+	var/datum/controller/controller_string = input("Select controller to debug", "Debug Controller") as null|anything in controller_choices
```
stanalbatross, line 83: Tgui input list?

stanalbatross (changes requested):

code/_byond_version_compat.dm
fira: This is actually DM-side version for compiler, not client-side version

code/controllers/mc/admin.dm
fira: This is probably intended since you'd want to eg. debug TGUI controller due to issues - the other way is to use Statpanel Status entries but that is also controller backed and might also fail. Verbs are always available via command bar even if everything else fails

Can replace it if you think it's desirable to have the menu there

code/_byond_version_compat.dm
stanalbatross: aha

code/controllers/mc/admin.dm
stanalbatross: fair, can keep old list

harryob (approved):
fix the big conflicto and we should... be good to go.

code/modules/admin/verbs/adminhelp.dm
```diff
@@ -16,9 +16,9 @@ GLOBAL_DATUM_INIT(ahelp_tickets, /datum/admin_help_tickets, new)
 	/// The set of all resolved tickets
 	var/list/resolved_tickets = list()
 
-	var/obj/statclick/ticket_list/astatclick = new(null, null, AHELP_ACTIVE)
-	var/obj/statclick/ticket_list/cstatclick = new(null, null, AHELP_CLOSED)
-	var/obj/statclick/ticket_list/rstatclick = new(null, null, AHELP_RESOLVED)
+	var/obj/effect/statclick/ticket_list/astatclick = new(null, null, AHELP_ACTIVE)
+	var/obj/effect/statclick/ticket_list/cstatclick = new(null, null, AHELP_CLOSED)
+	var/obj/effect/statclick/ticket_list/rstatclick = new(null, null, AHELP_RESOLVED)
```
harryob, line 21: something here is fucking with the ability of admins to view total tickets

harryob (changes requested):

code/controllers/subsystem/atoms.dm
```diff
@@ -100,6 +100,12 @@ SUBSYSTEM_DEF(atoms)
 		qdeleted = TRUE
 	else if(!(A.flags_atom & INITIALIZED))
 		BadInitializeCalls[the_type] |= BAD_INIT_DIDNT_INIT
+	/*
+	else
+		SEND_SIGNAL(A,COMSIG_ATOM_AFTER_SUCCESSFUL_INITIALIZE)
+		if(created_atoms && from_template && ispath(the_type, /atom/movable))//we only want to populate the list with movables
+			created_atoms += A.get_all_contents()
+	*/
```
morrowwolf, line 108: Remove commented code please.

code/controllers/mc/subsystem.dm
```diff
@@ -241,12 +280,23 @@
...
+/*
+/datum/controller/subsystem/vv_edit_var(var_name, var_value)
+	switch (var_name)
+		if (NAMEOF(src, can_fire))
+			//this is so the subsystem doesn't rapid fire to make up missed ticks causing more lag
+			if (var_value)
+				update_nextfire(reset_time = TRUE)
+		if (NAMEOF(src, queued_priority)) //editing this breaks things.
+			return FALSE
+	. = ..()
```
morrowwolf, line 301: Remove commented code please.

code/controllers/mc/globals.dm
```diff
@@ -29,6 +29,13 @@ GLOBAL_REAL(GLOB, /datum/controller/global_vars)
 	msg = "Edit"
 	return msg
 
+/*
+/datum/controller/global_vars/vv_edit_var(var_name, var_value)
+	if(gvars_datum_protected_varlist[var_name])
+		return FALSE
+	return ..()
+*/
```
morrowwolf, line 37: Remove commented code pleaase

morrowwolf (changes requested):

PR #1779, opened 2022-11-30
Remove wrong max_w_class comment

ThePiachu:
## About The Pull Request

As per `/obj/item/storage`'s `max_w_class` - `//Max size of objects that this object can store (in effect only if can_hold isn't set)`. However, the last part is not the case, and some code uses that functionality - you can't store guns in specific gun holsters if they get too big. So I removed the misleading part of the comment.

## Why It's Good For The Game

Comments should reflect what's going on in the code not to mislead programmers.

## Changelog

:cl:
fix: Fixed a misleading comment around storage items' max_w_class
/:cl:

### DIFF
--- a/code/game/objects/items/storage/storage.dm
+++ b/code/game/objects/items/storage/storage.dm
@@ -15,7 +15,7 @@
 	var/list/click_border_start = new/list() 		//In slotless storage, stores areas where clicking will refer to the associated item
 	var/list/click_border_end = new/list()
 	var/list/hearing_items							//A list of items that use hearing for the purpose of performance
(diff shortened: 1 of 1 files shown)


Diegoflores31:
Wait . But what if we actually want the item to be restricted by weight too ?? We already have bypass weight limit option . But we dont have any exception for can_hold bypassing weight limit

ThePiachu:
Hmm... I guess that's a fair use case. In which case I should delete the comment that states otherwise.

ThePiachu:
So the PR is not about just removing that wrong comment.

fira (approved):
For future ref do make these dmdoc comments when you touch em', might as well

PR #1780, opened 2022-11-30
Adds a limit on the amount of channels you can be on at once. Limit is set to 2 for now

Watermelon914:
## About The Pull Request

Limits the amount of radio channels you can be on at once. This limit is currently at 2 for most roles and is subject to change depending on the results of the testmerge.
Auxiliary roles have the limit at 3.

## Why It's Good For The Game

Purpose of this change is to bring a new challenge to marine players by forcing them to switch channels if they want to communicate with someone. This forces players into a more realistic communication network as information needs to be relayed between different players to propagate messages to and from the various different radio channels.

Currently, the communication network for CM is very basic, and commanders can afford to tune into every single radio channel at once with little downside, as over-speaking doesn't exist with text chat. This is too basic and removes the difficulty of communication, so this is what this solution aims to solve.

Currently, the limit is set to 2, but this can be increased if necessary as 2 might be too little for some roles. Testmerges will help to discover which roles need more than 2 radio channels active at once.
## Changelog

:cl:
balance: Limits the amount of communication channels you can be on at once to 2 for most roles. Auxiliary roles such as CO/XO/IO/PO and such can access 3 channels at once.
balance: Removes sending one message to a bunch of radio channels at once.
balance: Removes command channel access from CTs
/:cl:

### DIFF
--- a/code/__DEFINES/dcs/signals.dm
+++ b/code/__DEFINES/dcs/signals.dm
@@ -503,3 +503,7 @@
 // Checks whether the xeno can build a thick structure regardless of hive weeds
 #define COMSIG_XENO_THICK_RESIN_BYPASS "xeno_thick_resin_bypass"
 	#define COMPONENT_THICK_BYPASS (1<<0)
+
+/// Called when a radio is getting mobs in range to hear a radio message
+#define COMSIG_RADIO_GET_MOB_IN_RANGE "radio_get_mob_in_range"
+	#define COMPONENT_SKIP_RADIO_CHECK (1<<0)
--- a/code/__HELPERS/game.dm
+++ b/code/__HELPERS/game.dm
@@ -146,28 +146,27 @@
 	return hear
 
 /proc/get_mobs_in_radio_ranges(var/list/obj/item/device/radio/radios)
-
-	set background = 1
-
 	. = list()
 	// Returns a list of mobs who can hear any of the radios given in @radios
 	var/list/speaker_coverage = list()
 	for(var/obj/item/device/radio/R in radios)
-		if(R)
-			//Cyborg checks. Receiving message uses a bit of cyborg's charge.
-			var/obj/item/device/radio/borg/BR = R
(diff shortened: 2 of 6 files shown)


TosterIMC:
I think Radio Telephone Operator and/or Squad leader should get three beacouse RTO and SL need that third channel maybe someone like Comanding Oficcer should have exclusive four chanels avaible but i think thats a good pr for the game.

Yellows91:
Its pretty good overall but my only complaint that MPs cannot access every SHIPSIDE com line and they realistically are forced to always have atleast MP comms online

ZAB909:
This is the worst idea I've ever heard. Have you ever played lowpop? You are basically required as CT to have command comms, SLs to have multiple squad channels or even all of them, cause the other SLs don't exist. Or there's no SO' so you have to listen to other comms yourself. 

What if I'm an engineer in Charlie, and want to hear alpha comms? I just won't be able to. Cause I'd have almayer, Charlie, engie comms. And that's it. 

I'm an IO, I have Intel, command, jtac, almayer. I can't get a squad comm key, my headset is full. Why would you implement that. 

I think people who enjoy this PR have never actually had to have all squad channels open at once and organise the marine groundside cause everyone else is dead or not doing their job. 

"Just use a command headset lol" Command headsets are valuable because people whom are good at the game maximise what they can do and properly organise the team. This will just make ppl get them more

ZAB909:
This doesn't even mention that there's times where people need to do shit in all squad comms. PO asking for help cause there's a xeno. PO Saying there are xenos on the right side of the drop ship when landing. SL or RTO saying in squad comms that there's danger close fire support coming. Most people being dead, and the survivors being random amounts per squad an hour plus into a round. 

We already have the difficulty of having too many comm lines open at once, it simply goes too fast for most people to manage. And I think that's an accurate enough disadvantage for the user deciding if they want that many comm channels. The required 2-4 max will just mean people will either put in and out comm keys when saying something important or just take a command radio.

PR #1783, opened 2022-12-01
Comm limit nanu

Nanu308:
## About The Pull Request
First of all, I'm fairly pissed at people coming with personal attacks on a change they do not agree with and then proceeding to attack Walter without taking a second to think for themselves rather than scream about it.

I'll be taking over this PR https://github.com/cmss13-devs/cmss13/pull/1780 and re-opening it. Concerns, issues etc. should be put in the Discord Feedback thread in a respectable format.

This limits the amount of radio channels you can have enabled at the same time depending on your role.

## Why It's Good For The Game
_My interest in this PR is primarily two things:_

> 
> 1. Establish a more proper CoC for sending information up and down the chain.
> 
> 2. Hinder communication "powergaming" where one person is on every channel. IC you are not required to know every single tidbit of information.
> 
> However, there are some proper issues that would need to be addressed if this was to make it through.
> 
> 1. Look at impact on lowpop rounds. And how command staff/lack of command staff will handle coordination between squads.
> 
> 2. How do you if implemented alleviate people just using one "common" channel such as the command one for everything, leading to more clutter in a single channel.
> 

## Changelog
:cl: Nanu
balance: Limits the amount of communication channels you can be on at once to 2 for most roles. Auxiliary roles such as CO/XO/IO/PO and such can access 3 channels at once.
balance: Removes sending one message to a bunch of radio channels at once.
balance: Removes command channel access from CTs
/:cl:

### DIFF
--- a/code/__DEFINES/dcs/signals.dm
+++ b/code/__DEFINES/dcs/signals.dm
@@ -506,3 +506,6 @@
 
 // Used for smothering fires upon weather event start/stop
 #define COMSIG_GLOB_WEATHER_CHANGE "!weather_event_changed"
+/// Called when a radio is getting mobs in range to hear a radio message
+#define COMSIG_RADIO_GET_MOB_IN_RANGE "radio_get_mob_in_range"
+	#define COMPONENT_SKIP_RADIO_CHECK (1<<0)
--- a/code/__HELPERS/game.dm
+++ b/code/__HELPERS/game.dm
@@ -146,28 +146,27 @@
 	return hear
 
 /proc/get_mobs_in_radio_ranges(var/list/obj/item/device/radio/radios)
-
-	set background = 1
-
 	. = list()
 	// Returns a list of mobs who can hear any of the radios given in @radios
 	var/list/speaker_coverage = list()
 	for(var/obj/item/device/radio/R in radios)
-		if(R)
-			//Cyborg checks. Receiving message uses a bit of cyborg's charge.
-			var/obj/item/device/radio/borg/BR = R
-			if(istype(BR) && BR.myborg)
-				var/mob/living/silicon/robot/borg = BR.myborg
-				var/datum/robot_component/CO = borg.get_component("radio")
-				if(!CO)
-					continue //No radio component (Shouldn't happen)
-				if(!borg.is_component_functioning("radio") || !borg.cell_use_power(CO.active_usage))
-					continue //No power.
+		if(SEND_SIGNAL(R, COMSIG_RADIO_GET_MOB_IN_RANGE, .) & COMPONENT_SKIP_RADIO_CHECK)
+			continue
 
-			var/turf/speaker = get_turf(R)
-			if(speaker)
-				for(var/turf/T in hear(R.canhear_range,speaker))
-					speaker_coverage[T] = T
+		//Cyborg checks. Receiving message uses a bit of cyborg's charge.
+		var/obj/item/device/radio/borg/BR = R
+		if(istype(BR) && BR.myborg)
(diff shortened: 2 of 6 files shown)


Trubbl3:
this is the first time i heard the term "communication powergaming", how is it any powergaming for command to command its troops, if it powergaming why the xenos only have one channel to communicate and we are okay with it?

Herotitan772:
TLDR: not on the discord so just wanted to share my 2 cents

This is criticism not a personal attack or off topic, please take it as such.

**So marines already as is have to have control of an active comms tower to be able to communicate over radio**
- Multicast is a large part of command being able to communicate with the squads and removing that is like removing the hivemind for xenos, its just unthinkable.
- restricting command roles to only 3 channels wont stop them messaging out, it will only be a problem when they are being sent messages as they wont know to have the channel turned on. also just for clarification there is a total of 16 comms channels (command, jtac, intel, WY, req, engineering, medical, alpha, bravo, charlie, delta, foxtrot, echo, alamyer, colony, MP) so the chance of cic staff with 8 or so channels available missing a message directed towards them is very high
- there is already a system in place to stop people from having every channel, you are limited to having a max of 3 keys in a headset, you cannot insert any more.
- this seems like a poor way to achieve what is essentially another comms nerf, i would argue that it would make more sense to restrict access to getting new channels, eg. remove easily accessible high level headsets.
- one of the main things that i though of when i saw this TM was the fact that this is a game and all i can see this TM doing is making the game less fun, it doesn't make xenos players enjoy the game more, and it will only frustrate marines. so if the reason we play this game is to have fun why are we trying to make it less fun?
- lastly i just wanted to say that restricting the amount of active channels isn't the right idea, (i don't think marines need **another** comms nerf in the first place) if you really wanted to nerf marines comms then you should look at making it harder to get access to the channels in the first place not how many of those channels they can have running at once

Googles-Hands:
#### I got a question on my mind

I'd like to ask if squad headsets have their inbuilt squad channel count to the channel limit, if not, great.
Though if it does count towards the limit, I'd like to provide some thoughts on why that is an issue, it keeps Squad Engies, Squad Medics and Squad Leaders locked to their starting channels, with SL's being affected most, since they have JTAC and Command by default, which would result in them having to switch between the channels, which I think would prove to be a very large problem.
Squad Engies and Medics would also be blocked from being mortar operators/JTAC spotters as they would need to tune out their engie/medic comms to facilitate that.

#### I'd also like to give my take on the two issues listed in your PR:

1. Even on lowpop rounds, if there is any form of command, they can assign aSL's, who all automatically get command access, through which can be coordinated. Departmental complaints (medics calling for more medics or engies calling for help) can be done without the need for command to hear and relay it, if other engies/medics hear it, they will hopefully respond, if there is no response, the engies/medics can tell their SL who can then bring the issue to the attention of command.
2. Remove the spare command headsets from the CIC armoury, otherwise I do not think this can be really helped. Though I also do not think it is a major issue, the command channel would be a central link through which the operation can be coordinated, if a person with legitimate access spams on it, command can just tell them to knock it off and keep it to their specialized channel (squad comms for SL's and SO's).

#### Final thoughts
The success of this PR hinges on one thing, SL's/aSL's being competent, since it elevates them from glorified arrows that point to the front of the battle. So then that also requires an actual thought process behind it so PVT Baldie McGreen isn't selected to coordinate their squad and instead PFC McPlatinumMedal gets put in a position of responsibility. It's a question of skill on the marines part to find and actually have suitable leaders, because without them, this PR will cripple marine coordination.

Crowbar764:
Teamwork and communication are the two most essential parts of the game and community. They should be encouraged, not restricted.

Please don't remove something just because you can.

PR #1786, opened 2022-12-01
boosts the number of operations stalebot can do

harryob:
tgstation/tgstation#66307

it hasn’t blown up anything over there, and our stalebot has gone wonky.

we get 1000 operations like this an hour, this only uses 300 (max) cron’d to happen once a day, so… should be good

(https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#requests-from-github-actions)

## About The Pull Request

Things will close

## Changelog

yet another harry devops change


ThePiachu (approved):

fira (approved):

Nanu308:
head maintainer gaming, pre 24h merge

PR #1790, opened 2022-12-01
Adds an extremely simple ingame netspeak notifier.

AnCopper:
## About The Pull Request

Sends a notification to the player when they engage in netspeak.

## Why It's Good For The Game

It's better than nothing.

https://streamable.com/jqh8q3

## Changelog

:cl:
server: Added an ingame netspeak notifier.
/:cl:

### DIFF
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -160,6 +160,11 @@ var/list/department_radio_keys = list(
 		for(var/obj/O as anything in listening_obj)
 			if(O) //It's possible that it could be deleted in the meantime.
 				O.hear_talk(src, message, verb, speaking, italics)
+				
+	var/static/regex/netspeak = regex("\\b(lol|brb|wtf|lmao|sus|kek|idk|omg)\\b", "i")
+	if(findtext(message, netspeak))
+		to_chat(src, "<span class='warning'>Netspeak is not permitted in character!</span>")
+		return
 
 	//used for STUI to stop logging of animal messages and radio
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/say.dm
```diff
@@ -160,6 +160,11 @@ var/list/department_radio_keys = list(
 		for(var/obj/O as anything in listening_obj)
 			if(O) //It's possible that it could be deleted in the meantime.
 				O.hear_talk(src, message, verb, speaking, italics)
+				
+	var/static/regex/netspeak = regex("\\b(lol|brb|wtf|lmao|sus|kek|idk|omg)\\b", "i")
```
stanalbatross, line 164: No, the lowercase letter i is not netspeak

stanalbatross, line 164: Also this should pull from a list in a config file probably

code/modules/mob/living/say.dm
```diff
@@ -160,6 +160,11 @@ var/list/department_radio_keys = list(
...
+	if(findtext(message, netspeak))
+		to_chat(src, "<span class='warning'>Netspeak is not permitted in character!</span>")
```
stanalbatross, line 166: <span> tags are not used in CM
We use macros such as SPAN_WARNING()

stanalbatross (changes requested):

harryob (changes requested):
also generally needs usage of the configuration datum to toggle on/off

PR #1791, opened 2022-12-01
Renamed M3-H armor to M3-EOD armor and specialized it against ballistic and explosive damage.

carlarctg:
## About The Pull Request

Added extra desc_lore from Misty's armor renaming PR.

Heavy armor protects a lot more from bullets and explosions, but doesn't protect any more than medium armor does against acid anymore.

## Why It's Good For The Game

> Heavy armor protects a lot more from bullets and explosions, but doesn't protect any more than medium armor does against acid anymore.

As per https://discord.com/channels/150315577943130112/458150341742166017/1047942193630957610

This effectively specializes 'heavy armor' in HvH to a degree. It will reduce damage from FF and explosions. a full buckshot PB will for example put you in softcrit but still allow you to walk around, while the same with medium armor would cause you to fall unconscious. You can survive 1 more burst of m41, mod88, and 1 more l42 shot before being softcritted. Effectively it just means you can fight at the frontline for longer. Amusingly this means it's stronger than legacy b18... but who cares.

I was also considering increasing internal damage resistance further, but I'm unsure.

## Changelog

:cl:
bal: M3-H armor renamed to M3-EOD armor.
bal: M3-EOD armor is now highly resistant to ballistics and explosives, and has lost its additional acid protection.
/:cl:

### DIFF
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -551,16 +551,17 @@ var/list/squad_colors_chat = list(rgb(230,125,125), rgb(255,230,80), rgb(255,150
 	flags_atom = NO_SNOW_TYPE|NO_NAME_OVERRIDE
 
 /obj/item/clothing/suit/storage/marine/heavy
-	name = "\improper M3-H pattern heavy armor"
-	desc = "A heavier version of the standard M3 pattern armor, cladded with additional plates. It sacrifices speed for more durability."
-	specialty = "\improper M3-H pattern heavy"
+	name = "\improper M3-EOD pattern armor"
+	desc = "A heavier version of the standard M3 pattern armor, the armor is primarily designed to withstand ballistic, explosive, and internal damage, with the drawback of increased bulk and thus reduced movement speed, alongside little additional protection from standard blunt force impacts and none from biological threats."
+	desc_lore = "This configuration of the iconic armor was developed during the Canton War in 2160 between the UPP and USCM - Designed in response to a need for higher protection for ComTechs assigned as EODs during the conflict, this is the pinnacle of protection for your average marine. The shoulders and kneepads have both been expanded upon heavily, covering up the arteries on each limb. A special spall liner was developed for this suit, with the same technology being used in the M70 Flak Jacket being developed at the same time."
+	specialty = "\improper M3-EOD pattern"
 	icon_state = "H1"
 	armor_variation = 6
 	armor_melee = CLOTHING_ARMOR_MEDIUMHIGH
-	armor_bullet = CLOTHING_ARMOR_HIGH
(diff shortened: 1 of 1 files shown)


code/modules/clothing/suits/marine_armor.dm
```diff
@@ -551,16 +551,17 @@ var/list/squad_colors_chat = list(rgb(230,125,125), rgb(255,230,80), rgb(255,150
 	flags_atom = NO_SNOW_TYPE|NO_NAME_OVERRIDE
 
 /obj/item/clothing/suit/storage/marine/heavy
-	name = "\improper M3-H pattern heavy armor"
-	desc = "A heavier version of the standard M3 pattern armor, cladded with additional plates. It sacrifices speed for more durability."
-	specialty = "\improper M3-H pattern heavy"
+	name = "\improper M3-EOD pattern armor"
```
ThePiachu, line 554: Also update all the other instances the armour's name is mentioned, like the vendors.

ThePiachu (changes requested):

ThePiachu (changes requested):
Also, it's `balance` not `bal` for the changelog.

CABSHEP:
It is nice that it is a change to objectively the worst armor (since previous SL armor is in PFC vendor now, before you could make an argument), but it is is also sad that this is relegating it to HvH only.

And even then it won't get much use anyway as speed is still better than tanking one more bullet, or burst. 

But there is no point in trying to argue to change the direction you took with it armor. So instead I'm thinking about going further with the general idea that would help both HvH and normal rounds to the point of making it actually viable.
What if it had special property that massively reduced damage from projectiles with 5 armor penetration, or less? Like on top of what it already does additional lets say ~50% of the final damage would also be cut, but only for those specific projectiles.
(no idea how that could be implemented, maybe additional armor for calculation for those bullets)

That would mean heavy armor would really shrug normal pulse rifle ammo and buckshot. For 30 bullet protection it has for 10x24 it would make 10/11 damage per bullet (30/33 per burst) and for buckshot PB 52. 
Bullets that have higher penetration would work normally, thus HvH wouldn't be unbalanced as only CLF and Freelancers would have any problem with damaging the marine.

Sacrifacing speed and armor slot (im comparsion to medium) for truly massive anti-FF help would make this armor unique, something to actually consider between light armor. 

If anything (without further adjusting of anything) it would be nice to add a helmet for the armor too, since bullets still sometimes hit head and explosions definatelly do. But be sure to make it be only wearable with heavy armor and not others, because it would be an objective buff that every marine would want.

ThePiachu (approved):
Code looks good

ThePiachu:
Alright, since this has gone un-GA-reviewed for a month, I'm merging it.

PR #1801, opened 2022-12-02
Replaced power use magic numbers with proper defines

ThePiachu:
## About The Pull Request

While looking at #1692 I noticed we are using magic numbers for `update_use_power`, so I changed them to nice defines so we don't have some wrong calls like that issue was fixing.

## Why It's Good For The Game

Less magic numbers makes code more understandable.

## Changelog

:cl:
code: Replaced magic numbers around power usage with proper defines.
/:cl:


NoKohi:
good change :)

fira (changes requested):
This is great but should extend to `use_power` which uses the same values

Check the proc itself, `/obj/structure/machinery/proc/calculate_current_power_usage`

And this that uses it directly (may need to use update_use_power instead):
`/obj/structure/machinery/telecomms/relay/preset/tower/mapcomms/attack_hand`
(technically also holopad and such but we should just remove these..)

ThePiachu:
Heck, now this PR turned from 18 to 87 files modified! :D

harryob (approved):

fira:
Oh wow I didn't realize there were that many type-set instances, i had only searched code references with langserver...

ThePiachu:
Fixed the merge conflict

fira (approved):

PR #1833, opened 2022-12-03
Changes SEA loadouts & gives them sidearm options

DexterDude:
# About the pull request
Changes the Senior Enlisted Advisor's loadout up, removing some of the stuff it starts with but putting those options inside the job's vendors if they weren't already there as well as adds a few other handguns for them to use. Specifically, the SEA is given these things in their vendors:

Clothing Vendor:
- Officer cap
- USCM cap
- USCM Beret
- Tan USCM Beret
- Pistol Holster Belt
- M44 Holster Belt

Weapons Vendor:
- M4A3 Custom with magazines (would have done AP because the custom gets marksman but was worried about powercreep)
- M44  Custom with marksman magazines
- 88Mod4 with magazines
- VP78 with magazines

Originally started as a PR just to add the gun options with the approval of the mentor overseer guys but I decided I might as well make the starting loadout more customizable 

# Testing status

I've tested the changes and they seem to work well. I attempted to make it so that the SEA would only get 5 magazines instead of using points but that was something I couldn't really understand, and from the looks of it requires the vendor would need to be an entirely different path and thus having the consequence of the SEA being able to vend 4 different handguns and the associated ammo. I highly doubt they'll get more use out of 9 MK1 magazines than they will out of 5 so power-gaming shouldn't be an issue.

When I play SEA it gets tedious and annoying to remove all the stuff that I spawn with on the start and toss it into a big ugly pile in the corner just so I can equip what I actually want. Although the DI hat and M44 Custom are iconic gear of the SEA I personally don't use either and other SEA players might not as well. You can still get them of course, but SEA is provided more options without having to toss everything into a corner.

As for why I added the VP78, I think it's cool to give them more options than the standard without giving them guns like the M1911 or M16. VP78 isnt exactly more powerful than other guns from what I've heard of it and experienced in testing, and that shouldn't be a real concern because the SEA is a 99% of the time shipside role (much like the XO who also gets it as an option). In the cases hijack happens it really won't make much of a difference compared to them using other guns. It also looks and sounds cool.

Not really needed here, pretty simple change. Tested it myself, geared up like I usually do as SEA without needing to throw several things out.

# Changelog

:cl:
add: Gave the SEA the option at their vendor to choose from the M4A3 Custom, M44 Custom, 88Mod and VP78 pistols with associated magazines
add: Added the USCM cap, USCM Officer Cap, USCM Beret, and the USCM Tan Beret to the SEA's vendor
del: Removed the large general pouch, drill instructor hat, and M44 custom belt from the SEA's loadout
/:cl:


monkeysfist101 (changes requested):
Ditch the two custom pistols and the VP78. If you want them, you can spend the snowflake points like everyone else. The 78 can be acquired from req. The rest is fine.

DexterDude:
> Ditch the two custom pistols and the VP78. If you want them, you can spend the snowflake points like everyone else. The 78 can be acquired from req. The rest is fine.

The SEA already spawns with the custom revolver in their belt, there is no reason at all for me to change that and not give them it through this PR. M4A3 custom is because their M44 is a custom revolver. 

I can understand the VP78 however, but why would I go to req to get it as a shipside role? I understand it's not used a lot but a big concern of SEAs from other people is them gearing up for stuff like they know shit is about to go down with gear that doesn't come from their vendor and taking it from people who might actually use it. That's why their spec skills were removed. This removes the issue of them wasting stuff in req.

Could you elaborate on why the VP78 is a big issue? The Executive Officer, another mostly shipside role also gets it and it isn't an issue. The VP78 isn't even explicitly more powerful than other weapons from what I tested.

code/game/machinery/vending/vendor_types/crew/sea.dm
```diff
@@ -4,14 +4,20 @@ GLOBAL_LIST_INIT(cm_vending_gear_sea, list(
 		list("PERSONAL PRIMARY (CHOOSE 1)", 0, null, null, null),
 		list("M41A MK1 Pulse Rifle", 0, /obj/item/weapon/gun/rifle/m41aMK1, MARINE_CAN_BUY_ESSENTIALS, VENDOR_ITEM_REGULAR),
 
-		list("SPARE AMMUNITION", 0, null, null, null),
-		list("M41A MK1 Magazine", 9, /obj/item/ammo_magazine/rifle/m41aMK1, null, VENDOR_ITEM_REGULAR),
-
-		list("GUN ATTACHMENTS (CHOOSE 1)", 0, null, null, null),
```
PalJohnson, line 10: Whats with the removal of the attachment? This was put there mainly so the SEA doesn't poach attachments from req or squad prep.

DexterDude: As I said in the mentor channel, the removal was because it used the attachment tag which is now used by the sidearms so that the SEA only purchases one. SEA doesn't really need an attachment on any of their guns, all of them are good without one, if you really need one in emergency you can go to a squad prep or req.

PR #1839, opened 2022-12-04
Carrier QoL improvements - grab multiple eggs, grab huggers from morpher, restock morpher with eggs

ThePiachu:
# About the pull request

After playing the Carrier for a bit I noticed its gameplay could be improved a bit in terms of QoL.

First, sometimes it's hard clicking all the eggs the queen has laid, especially when some other sprite like a bush or a potted plant get in the way. So I let them grab all the eggs from a given turf when they try storing any one egg.

Secondly, refilling from an egg morpher was a bit fiddly (click on morpher to make hugger appear, middle-click on the hugger to take it, try not to misclick or you will be trying to slap one hugger with another), so I let Carriers just top themselves up from the egg morpher by using their hugger ability.

Thirdly, it tends to get a bit hard to restock the egg morphers far away from the hive, especially later in the round, even though there are like 50 eggs sitting by the queen nobody is taking, so I let xenos use an egg on the morpher to restock it with the facehugger from the egg directly.

It makes the life of hive janitors a lot more pleasant, which should result in more eggs being used rather than lying there doing nothing.

<details>

</details>

# Changelog

:cl:
add: You can now refill an Egg Morpher directly with an egg by clicking on it with the egg.
qol: Carriers can now take facehuggers directly from the Egg Morpher with their facehugger ability.
qol: Carriers now attempt to pick up all the eggs on a turf if they use their egg ability on an egg lying on the ground.
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/special/egg_morpher.dm
+++ b/code/modules/cm_aliens/structures/special/egg_morpher.dm
@@ -118,6 +118,7 @@
 				SPAN_XENONOTICE("You place the child from an egg into \the [src]."))
 			user.temp_drop_inv_item(egg)
 		stored_huggers = min(huggers_to_grow_max, stored_huggers + 1)
+		playsound(src.loc, "sound/effects/alien_egg_move.ogg", 25)
 		qdel(egg)
 		return
 	return ..(I, user)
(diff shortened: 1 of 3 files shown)


ThePiachu:
Added an egg opening sound for when you refill a morpher with an egg.

harryob (approved):

PR #1841, opened 2022-12-04
Cinematic Crit

Geevies:
# About the pull request

You can now hear people talking around you if you're in knocked-out, but only if they're right next to you. You can now talk while knocked-out, but you can only whisper. You cannot use the radio when speaking this way. Your sound environment changes when you're knocked out, making things sound more spooky. Stuttering no longer destroys your words.

Ported the stuttering code from: https://github.com/Aurorastation/Aurora.3/pull/7344 (not the markup fixes though, out of scope)

It makes the knocked-out experience way less painful by letting you at least talk to people to some degree, but only if they're right up against you, making it much more cinematic and meaningful.

Also obliterates the stupid stuttering because H-H-H-H, while funny, EXTREMELY funny, does get old after a while.

images taken before the stuttering port, and i'm not launching two clients again

image
image

# Changelog

:cl:
add: You can now whisper while knocked out, but cannot use the radio this way.
add: You can now hear what people are saying while knocked out, but only if they're adjacent you.
add: Getting knocked-out now changes the sound environment to reflect your brain getting blasted.
add: Ported a change to stuttering which makes it less H-H-H-like.
/:cl:

### DIFF
--- a/code/__HELPERS/type2type.dm
+++ b/code/__HELPERS/type2type.dm
@@ -263,3 +263,18 @@
 		else //regex everything else (works for /proc too)
 			return lowertext(replacetext("[the_type]", "[type2parent(the_type)]/", ""))
 
+// Converts a string into a list by splitting the string at each delimiter found. (discarding the seperator)
+/proc/text2list(text, delimiter="\n")
+	var/delim_len = length(delimiter)
+	if (delim_len < 1)
+		return list(text)
+
+	. = list()
+	var/last_found = 1
+	var/found
+
+	do
+		found       = findtext(text, delimiter, last_found, 0)
+		.          += copytext(text, last_found, found)
+		last_found  = found + delim_len
+	while (found)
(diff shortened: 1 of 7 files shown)


ihatethisengine:
>says softly
I think you need to change it for knocked-out

Merrgear:
we have to manually He he helpa now. bad PR ban geeves for life

Googles-Hands:
Does the knocked out state include the normal sleep/chemically induced sleep state or just the crit state, because if it's both, thats a bit weird, you shouldn't be able to hear perfectly/whisper while asleep.

code/__HELPERS/type2type.dm
```diff
@@ -263,3 +263,18 @@
...
+
+	. = list()
+	var/last_found = 1
+	var/found
+
+	do
+		found       = findtext(text, delimiter, last_found, 0)
+		.          += copytext(text, last_found, found)
+		last_found  = found + delim_len
+	while (found)
```
harryob, line 280:
```suggestion
	while (found)
		found       = findtext(text, delimiter, last_found, 0)
		.          += copytext(text, last_found, found)
		last_found  = found + delim_len
```
goofy ass proc

harryob (changes requested):

stanalbatross (dismissed):

PR #1842, opened 2022-12-04
Stealthy Actions

Geevies:
# About the pull request

The little spinny bar above your head no longer appears if your alpha is below maximum. This affects: Scouts, lurkers, predators, and probably some other stuff I missed.

I always thought it was a bit silly that you could track things down not by seeing them, but by seeing the game UI floating above their heads. This should make it a little more engaging to track down the bad guy while giving them a chance to Metal Gear their way out of trouble.

I tried planting a C4 on something while cloaked, rapidly cloaked and uncloaked to confirm that the thing works, walked away midway through to make sure it disappeared correctly, etc. etc.

# Changelog

:cl:
balance: The little spinny thing above your head no longer appears when you're partially cloaked, like scouts, lurkers, or predators.
/:cl:

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1035,8 +1035,8 @@ var/global/image/action_purple_power_up
 		has_target = TRUE
 
 	// Only living mobs can perform timed actions.
-	var/mob/living/L = user
-	if(!istype(L))
+	var/mob/living/busy_user = user
+	if(!istype(busy_user))
 		return FALSE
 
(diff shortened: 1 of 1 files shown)


stanalbatross:
It's called an action wheel…

fira:
Doesn't the wheel inherit carrier's alpha?

stanalbatross:
> Doesn't the wheel inherit carrier's alpha?

this is a better solution than disabling it entirely, and is better from a balance perspective too

PR #1843, opened 2022-12-04
Chameleon Lurker

Geevies:
# About the pull request

Added a new Lurker strain, Chameleon, which loses its ability to pounce and cloak temporarily, but gains faster movement, faster walking speed, and the ability to go invisible while in stalk mode.

It brings back lurker in a way it used to be, which is neat I guess. Also makes use of the movement intent system I added recently, which is also neat, I guess.

I cloaked while in stalk mode to make sure it correctly automatically cloaks, I ran around to make sure the speed was right, and I checked my action bar to make sure the abilities were gone.

# Changelog

:cl:
balance: Added a Chameleon lurker, which can't pounce or cloak temporarily, but gains faster movement speed and cloaks in stalk mode.
/:cl:


harryob (changes requested):
otherwise code lgtm

Tsurupeta:
> It brings back lurker in a way it used to be

Old lurker could pounce on targets, so you are not bringing old one back, you just remove his pounce while giving him semi-permanent invis.

PR #1844, opened 2022-12-04
Stealth Typing

Geevies:
# About the pull request

Made the typing indicator only appear if you have full alpha. Also removed an unused proc.

Same as https://github.com/cmss13-devs/cmss13/pull/1842 but in this case the change is even lesser because you can actually disable typing indicators as a preference.

Believe it or not, but I tried typing while cloaked.

# Changelog

:cl:
balance: You no longer have a typing indicator while cloaked. The regular speech bubble when speaking still pops up, though.
/:cl:


fira (approved):

stanalbatross:
alternatively could just make it inherit alpha

PR #1845, opened 2022-12-04
Adds Burrower to Whiskey Outpost

Unknownity:
# About the pull request

The PR brings in a new threat to the Dust Raiders by including in the Burrower caste to xeno waves. The simple part, the burrower will serve as a breaching hazard that marines need to be wary of, which provides a purpose for certain roles (like Honor Guards) to keep an eye over the field. The burrower's purpose is to primarily try and burrow through marine defenses to attack unsupervised areas and destroy their defenses and equipment from behind and/or kill lone individuals. The burrower's health is reduced to 365 from 500 before wave 15 and by accounting for their slow running speed and lack of being able to heal from weeds, they can easily be killed by a group of marines, especially when alone. THEY ARE UNABLE TO CREATE TUNNELS (big caps so you know) to prevent some comical highly unbalanced plays such as the hive bypassing multiple layers of cades by simply tunneling to the abandoned shack and attacking people from behind. Mapping has been checked for and I concluded that it will be extremely hard for the burrower to burrow all the way to the bunker beaches while not being detected, past that point, it becomes easier for them, there are no areas that they can burrow to which is also unreachable by marines so they can always be intercepted. Burrowers only spawn at wave 7 (the time where marines get a grace period and cryo units) which gives max 3 burrowers during that wave, wave 14 (second last wave xenos get) and as a possible random wave during wave 15 (15 chooses one of the available random waves) the random wave is a special one that only burrowers spawn with a current max total being 20. Wave 14 and 15 burrowers serve as the final lategame assault against people inside the bunker, especially against those who have heavily reinforced themselves which would require a lot of xeno casualties to break through the defenses.

I have also added a new random wave for wave 15 which is a Defender (with Crusher) only equivalent like we currently have for Runner Madness (Runners and ravagers) and Spitter Madness (Spitters and Preatorians), Defender madness has only 30 max xenos (the default for wave 15 is 50) due to them being more tanky. This brings up a total of possible random waves from 2 to 4, so a Burrower only wave has a 1/4th chance to spawn.

Makes marines more cautious and encourages them to patrol when the time arrives and gives xenos a unique caste that does something else than bruteforcing against cades and also another lategame option against bunker defenses

I waited until wave 7 occured to check if the burrower becomes available, I kept spawning as one to check their limited quantity and I checked that their health did get decreased. I also checked that they are unable to dig tunnels during the gamemode and that they can still dig tunnels during other gamemodes (Extended, Distress etc.) and checked that they are unable to burrow to unreachable places.

# Changelog

:cl: Unknownity
balance: Added Burrowers to Whiskey Outpost.
add: Added a new Defender/Crusher only random wave for Wave 15 in Whiskey Outpost.
/:cl:


carlarctg:
yes  yes yesy yesy sysey

code/modules/mob/living/carbon/xenomorph/Abilities.dm
```diff
@@ -29,6 +29,10 @@
 		to_chat(X, SPAN_XENOWARNING("You can't do that from there."))
 		return
 
+	if(SSticker?.mode?.hardcore)
```
fira, line 32: hardcore should be set on the xenos themselves too, more straightforward

fira (commented):
I suggest tidying up the PR about text for next ones, it's hard to read X_x

code/modules/mob/living/carbon/xenomorph/Abilities.dm
Unknownity: Pretty sure setting hardcore checks on xenos for making tunnels is less straightforward with minimal differences and with more code while giving a hardcore check only on the ability takes the least amount of code.

if anything, hardcore requires to be refactored entirely but this is out of scope of the PR

ThePiachu (approved):

PR #1846, opened 2022-12-05
Added a rolling crate to Almayer

ThePiachu:
# About the pull request

I've noticed the trash cart crate that's on a few maps is rather slow, despite having wheels. So I lowered their drag delay from 3 to 1.5. Seeing how often IOs and CTs have to suffer by dragging all their crates around, I figured I'd add a few rolling crates onto Almayer (about 1 per department) with a drag delay of 1. This should let the various departments be able to move some loose items a bit more handily, but not so common as to replace regular crates.

It will make cargo and IO's runs between their department and the dropships a bit less of a slog, which is always nice.

<details>

</details>

# Changelog

:cl:
add: Added a rolling crate to a few departments on the Almayer. Since it has wheels its movement delay is much lower than regular crates. CTs and IOs should be able to use these to speed up their job.
add: Trash carts found on various planets also have their movement delay lowered, although they are not as fast as the rolling crate.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/crates.dm
+++ b/code/game/objects/structures/crates_lockers/crates.dm
@@ -225,7 +225,6 @@
 	icon_opened = "cooler-oj_open"
 	icon_closed = "cooler-oj_closed"
 
-
 /obj/structure/closet/crate/hydroponics
 	name = "hydroponics crate"
 	desc = "All you need to destroy those pesky weeds and pests."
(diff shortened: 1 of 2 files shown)


fira:
Despite mostly playing Reqs, I'm honestly not a fan of this change because it effectively deprecates powerloaders carrying. 
They are already barely faster than crate pushing, or marginally slower if chain pushing.
If there is a need for this, we could just make the powerloader faster instead

ThePiachu:
Powerloader is still used for dealing with the DS and moving the big big crates, and IOs can't use powerloaders last I checked, while I often see them dragging two crates of intel at a time at the speed of a snail.

That being said, I wouldn't mind buffing the powerloader speed a bit since yeah, you don't see them for moving crates all too often.

PR #1850, opened 2022-12-05
Ghost Langchat + Fixes

fira:
# About the pull request

image
Had a look, turns out it was much simpler than expected, so here we are. 

This displays abovehead chat for ghosts within view range when they use deadchat or emotes.

It can be disabled by one of: 
 - Toggling off Langchat entierly
 - Toggling off own ghost visibility will disable it for others while speaking
 - Toggling off Ghost Vision will disable text alongside the ghost visuals

Also fixes some bugs.

More tooling for ghost interaction and for observers to observe the round together.

Testing details:
 * Tested ghost to ghost langchat works closeby with 2 clients
 * Tested with one ghost outside regular range into zoom range
 * Tested offsets are correct and inherited when ghosting from Queen
 * Tested disabling self visibility disables outputting langchat locally
 * Tested disabling ghost vision disables outputting langchat locally
 * Tested langchat isn't leaked to living mobs (prevented by broadcast logic)
 * Tested emotes work similarly

# Changelog

:cl:
add: Observers now have overhead text! It can be toggled with the regular Langchat setting, or hiding your/others ghosts.
fix: Fixed an oversight which would prevent either own Toggle Dchat or server's from working (they had the same name).
fix: Fixed Overhead chat not aligning properly with mob icon size, eg. for large xenomorphs.
/:cl:

### DIFF
--- a/code/datums/langchat/langchat.dm
+++ b/code/datums/langchat/langchat.dm
@@ -66,6 +66,21 @@
 	if(appearance_flags & PIXEL_SCALE)
 		langchat_image.appearance_flags |= PIXEL_SCALE
 
+/mob/langchat_make_image(var/override_color = null)
+	var/new_image = FALSE
+	if(!langchat_image)
+		new_image = TRUE
+	. = ..()
+	// Recenter for icons more than 32 wide
+	if(new_image)
+		langchat_image.maptext_x += (icon_size - 32) / 2
+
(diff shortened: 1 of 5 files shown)


harryob (approved):

harryob:
funnier before the orbiting fix mind

PR #1857, opened 2022-12-06
Makes janitorial cart slashable

CrafticalDecoy:
# About the pull request
Makes the janitorial cart slashable! With no old commits bleeding in this time!

Marines can shoot above the cart, and it blocks xenos while being unslashable. No more free moveable defense object.

# Changelog

:cl:
fix: The janitorial cart can now be slashed.
/:cl:

### DIFF
--- a/code/game/objects/structures/janicart.dm
+++ b/code/game/objects/structures/janicart.dm
@@ -179,3 +179,15 @@
 		overlays += "cart_bucket"
 	if(signs)
 		overlays += "cart_sign[signs]"
+
+/obj/structure/janitorialcart/attack_alien(mob/living/carbon/Xenomorph/M)
+	if(M.a_intent == INTENT_HARM)
+		M.animation_attack_on(src)
+		playsound(src, 'sound/effects/metalhit.ogg', 25, 1)
+		M.visible_message(SPAN_DANGER("[M] slices [src] apart!"),
+		SPAN_DANGER("You slice [src] apart!"), null, 5, CHAT_TYPE_XENO_COMBAT)
+		qdel(src)
+		return XENO_ATTACK_ACTION
+	else
+		attack_hand(M)
+		return XENO_NONCOMBAT_ACTION


code/game/objects/structures/janicart.dm
```diff
@@ -179,3 +179,12 @@
...
 	if(signs)
 		overlays += "cart_sign[signs]"
+
+/obj/structure/janitorialcart/attack_alien(mob/living/carbon/Xenomorph/M)
+	M.animation_attack_on(src)
+	playsound(src, 'sound/effects/metalhit.ogg', 25, 1)
+	M.visible_message(SPAN_DANGER("[M] slices [src] apart!"),
+	SPAN_DANGER("You slice [src] apart!"), null, 5, CHAT_TYPE_XENO_COMBAT)
+	qdel(src)
+	return XENO_ATTACK_ACTION
```
harryob, line 189:
```suggestion
/obj/structure/janitorialcart/attack_alien(mob/living/carbon/Xenomorph/xeno_attacker)
	xeno_attacker.animation_attack_on(src)
	playsound(src, 'sound/effects/metalhit.ogg', 25, 1)
	xeno_attacker.visible_message(SPAN_DANGER("[xeno_attacker] slices \the [src] apart!"),
	SPAN_DANGER("You slice \the [src] apart!"), max_distance = 5, message_flags = CHAT_TYPE_XENO_COMBAT)
	qdel(src)
	return XENO_ATTACK_ACTION
```
u get 1 total rewrite on the house instead of me making u do this yourself

i know it's all copypasta'd but that code is BAD
first we use a more descriptive var name than M (ugly)
then we change all the references from M to our new var name
things need a \the before them to make sure they don't just show up as Crusher (XX-121) slices janitorial cart apart!
also we hate null arguments so i directly define which args we're using instead

harryob (changes requested):

code/game/objects/structures/janicart.dm
CrafticalDecoy: Noted for the future, thanks for the assist!

harryob (approved):

PR #1868, opened 2022-12-07
Carriers once again drop huggers when killed

ThePiachu:
# About the pull request

I'm not sure why the Carriers only drop eggs and not huggers when killed, but I missed that feature and Stan also mentioned it ought to be brought back, so here it is. It would be a nice feature so Carriers aren't just a liability to the Hive when they die...

It makes Carriers actually scary to take down if one is in close quarters with you which is a nice feature to have.

<details>

</details>

# Changelog

:cl:
add: Carriers once again drop huggers when killed.
fix: Fixed the Egg Morpher iterating from a wrong number for spawning huggers on destruction.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/Carrier.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Carrier.dm
@@ -124,6 +124,15 @@
 	if(.)
 		var/chance = 75
 
+		if (huggers_cur)
+			//Hugger explosion, like an egg morpher
+			var/obj/item/clothing/mask/facehugger/hugger
+			visible_message(SPAN_XENOWARNING("The chittering mass of tiny aliens is trying to escape [src]!"))
+			for(var/i in 0 to huggers_cur)
+				if(prob(chance))
+					hugger = new(loc, hivenumber)
+					step_away(hugger, src, 1)
+
 		while (eggs_cur > 0)
 			if(prob(chance))
 				new /obj/item/xeno_egg(loc, hivenumber)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/castes/Carrier.dm
```diff
@@ -124,6 +124,15 @@
 	if(.)
 		var/chance = 75
 
+		if (huggers_cur)
+			//Hugger explosion, like an egg morpher
+			var/obj/item/clothing/mask/facehugger/hugger
+			visible_message(SPAN_XENOWARNING("The chittering mass of tiny aliens is trying to escape [src]!"))
```
carlarctg, line 130:
```suggestion
			balloon_alert_to_viewers(SPAN_XENOWARNING("The chittering mass of tiny aliens is trying to escape [src]!"))
```

carlarctg: seems like a perfect time to use em

ThePiachu: I dunno, I don't think we ever use overhead messages like these for narration:
image

It's an interesting function otherwise!

code/modules/mob/living/carbon/xenomorph/castes/Carrier.dm
```diff
@@ -124,6 +124,15 @@
...
+			for(var/i in 0 to huggers_cur)
```
fira, line 131: Shouldn't that start from 1 ?

harryob (approved):

PR #1877, opened 2022-12-07
Flares visual rework

fira:
# About the pull request

image

This:
 * Makes flares land randomly rotated after they spin in throwing
 * Reworks sprites to separate the flame burn animation and so it's always facing north side
 * Converts flame burn to a greyscale animation using multiply+additive to achieve similar looks in any color

So at the end of the day, flares of arbitrary colors land randomly with their flame facing up.

Some explanations and caveats:
 * For the "flame always up" part, the simpler way of doing this was to fit the flares into half the spritesheet, to have their lighting point in center of it. This means it rotates like the hands of a clock, and the flame can always be put in the middle. Otherwise, this requires snowflake in-code calculations to position it right.
 * As a result, alignment is approximative (one pixel off) depending on rotation.
 * To achieve this, 2 lighting pixels had to be cropped from Signal Flare's back.
 * To achieve this, the flame animation had to be reduced from about 19-21 pixels long to roughly 18 to fit into frame. I'm not entirely happy about this.
 * You can't straight up greyscale the flame, because it's not "equally" as "red" as it is "white" (consider RGB for the 3 main pixels of old flare burn: (239,71,63), (255,119,101), (255,172,161)). To get around this, we apply it twice: once as traditional greyscale tint for the flame gradient, and a second time as additive mask with the broader flame color. This is probably wrong as I have no idea what i'm doing, but it seems to work overall....
 * It might make sense to have a bigger gap in the gray scale as a result. I've never done this, i dunno!

It's just cool visually and helps anyone wanting to use colored flares in the future.

See above

# Changelog

:cl:
add: Flares will now land in different positions as they spin from throwing.
add: Flares flame is now separate and will always face up.
add: Converted flares flames to grayscale, allowing flares of any color to be made.
/:cl:


Wintermote:
>You can't straight up greyscale the flame, because it's not "equally" as "red" as it is "white" (consider RGB for the 3 main pixels of old flare burn: (239,71,63), (255,119,101), (255,172,161)). To get around this, we apply it twice: once as traditional greyscale tint for the flame gradient, and a second time as additive mask with the broader flame color. This is probably wrong as I have no idea what i'm doing, but it seems to work overall....

https://www.youtube.com/watch?v=gJ2HOj22gDo
It sounds like you may have run into the issue that was brought up in this video:
desaturation gives an inaccurate Value to the desaturated colors.

The easiest solution is to add a new layer, make it white, and change the blending mode of that new layer to Color.

This should give a more accurate greyscale that preserve the original Values better, for when you re-apply color to the image later.

fira:
> The easiest solution is to add a new layer, make it white, and change the blending mode of that new layer to Color.

I'll have to look into this properly when i get a moment. That sounds like what i tried to do but as i said i likely done it wrong. I suspect increasing greyscale differences in the sprite would also improve contrast on the flame.

morrowwolf (approved):
Code LGTM

FransFeiffer (approved):
Sprites LGTM

PR #1879, opened 2022-12-07
Cross-ciphering fix

RenaRenaRe:
# About the pull request
Fixes the cross-ciphering property so it actually works instead of producing a never ending stream of runtime errors and a broken larva that never develops. I'm deliberately being vague about what it does because I think it's supposed to be a secret, the wiki tells you to "find out in game!", but you can just read the code since it isn't complicated.
This is my first time using github so sorry if I've fucked something up.

Things should function the way they're supposed to.
Cross-ciphering is an extremely hard to get property that I'm not sure if anyone has ever actually made in game, now on the incredibly rare occasion that somebody actually makes it it should work correctly.

<details>

</details>

# Changelog

:cl:
fix: Cross-ciphering now works correctly
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry_properties/prop_special.dm
+++ b/code/modules/reagents/chemistry_properties/prop_special.dm
@@ -141,15 +141,15 @@
 	if(amount < 10)
 		return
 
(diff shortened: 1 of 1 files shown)


fira (approved):
Sshhhh.

PR #1880, opened 2022-12-07
Revamp light floor tiles

moltoretardo:
# About the pull request
When I initially fixed the light tiles, I found many unused sprite of different colors, decided to add them and change a bit how it works.
Here are the different options:
lights

- Multitools now change the color by clicking on the tile
- Click on them to turn on/off
- Tiles can now be broken by xenos and environment (tested with grenades)
- To repair the floor tile, put a normal light bulb in it 
lightbulb
- Use a crowbar to remove the tile, if it's broken it'll simply destroy it
- Build one by using wire on glass and then one metal (unchanged)
- Their current light emission is the same as a flashlight, may be tweaked eventually to be less
- Currently the broken sprite and the off sprite is the same, should get a broken sprite eventually...
- Most of the old code wasn't used at all or didn't work, rewrote most of it

I think it's a good addition in the toolbelts of Maintenance Technicians, they can decorate rooms and use them for signalization instead of drawing with crayons on the floor. There shouldn't be any balance issues since it's basically a costly flashlight, light bulbs are, as far as I know, only obtainable from the autolathe making it not very reliable. Auto-light replacer doesn't work purposely so it doesn't get weaponized, their purpose is solely to embellish rooms.

It's hardly something that is used, and if it gets abused I'll simply nerf them down to ashes (can't have nice things) 
arroww

I've tested everything I could think of, everything should work.
<details>

Nade breaking some of the tiles
beforenade

nade destroy

</details>

# Changelog

:cl: Molto
add: Readded the different light floor colors, use multitool to cycle through them
fix: Fixed light tiles being indestructible unless using a crowbar
code: Moved some part of the code that were only used by light tiles in floors.dm into light.dm directly 
refactor: Made one letter variables more lettered
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/tiles/light.dm
+++ b/code/game/objects/items/stacks/tiles/light.dm
@@ -8,23 +8,10 @@
 	attack_verb = list("bashed", "battered", "bludgeoned", "thrashed", "smashed")
 	stack_id = "light floor tile"
 	turf_type = /turf/open/floor/light
-	var/on = 1
-	var/state //0 = fine, 1 = flickering, 2 = breaking, 3 = broken
 
-/obj/item/stack/tile/light/New(var/loc, var/amount=null)
+/obj/item/stack/tile/light/attackby(obj/item/item_in_hand as obj, mob/user as mob)
 	..()
-	if(prob(5))
-		state = 3 //broken
-	else if(prob(5))
-		state = 2 //breaking
-	else if(prob(10))
-		state = 1 //flickering occasionally
-	else
-		state = 0 //fine
-
-/obj/item/stack/tile/light/attackby(var/obj/item/O as obj, var/mob/user as mob)
-	..()
-	if(istype(O,/obj/item/tool/crowbar))
+	if(istype(item_in_hand,/obj/item/tool/crowbar))
 		new/obj/item/stack/sheet/metal(user.loc)
 		amount--
 		new/obj/item/stack/light_w(user.loc)
--- a/code/game/objects/items/stacks/tiles/light.dm
+++ b/code/game/objects/items/stacks/tiles/light.dm
@@ -11,7 +11,7 @@
 
 /obj/item/stack/tile/light/attackby(obj/item/item_in_hand as obj, mob/user as mob)
 	..()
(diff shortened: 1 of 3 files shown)


code/game/objects/items/stacks/tiles/light.dm
```diff
@@ -8,23 +8,10 @@
...
-		state = 2 //breaking
-	else if(prob(10))
-		state = 1 //flickering occasionally
-	else
-		state = 0 //fine
-
-/obj/item/stack/tile/light/attackby(var/obj/item/O as obj, var/mob/user as mob)
-	..()
-	if(istype(O,/obj/item/tool/crowbar))
+	if(istype(item_in_hand,/obj/item/tool/crowbar))
```
stanalbatross, line 14: Check for tool trait don't istype

code/game/turfs/floor.dm
```diff
@@ -96,6 +96,7 @@
 	else if(is_light_floor())
 		icon_state = "light_broken"
 		broken = 1
+		SetLuminosity(0)
```
stanalbatross, line 99: Does this not have defines

code/game/turfs/light.dm
```diff
@@ -10,33 +10,81 @@
...
+
+/turf/open/floor/light/attack_hand(mob/user as mob) //turning the light on and off
+	if(!broken && on)
+		on = FALSE
+		update_icon()
+		to_chat(user, SPAN_NOTICE("You turn the light off."))
+	else if(!broken && !on)
+		on = TRUE
+		update_icon()
+		to_chat(user, SPAN_NOTICE("You turn the light on."))
```
stanalbatross, line 75: This can be shorter, use on = !on and use a [on ? "On" : "off"] thing in the tochat

moltoretardo: That's so awesome

code/game/turfs/floor.dm
moltoretardo: I haven't found any, I can create them if needed though

harryob: that'd be OOS for this PR, luminosity is a pretty big thing

harryob (approved):

PR #1891, opened 2022-12-08
Fix for accuracy falloff.

Segrain:
# About the pull request

Followup to #1541, in the sense that all the complaints about spitters being unable to land any hits at intended 7 tile range finally made me clear this single line off my todolist. Discovered more weirdness in need of fixing in the process, but that is a story for another day.

image

# Changelog

:cl:
fix: Projectiles no longer experience an unintended massive drop in accuracy immediately past their maximum accurate range.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -551,7 +551,7 @@
 		if(distance_travelled <= ammo.accurate_range_min) // If bullet stays within max accurate range + random variance
 			effective_accuracy -= (ammo.accurate_range_min - distance_travelled) * 10 // Snipers have accuracy falloff at closer range before point blank
 	else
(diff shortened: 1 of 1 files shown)


fira:
I always been seeing this as intended, but i guess it doesn't make a ton of sense...

Segrain:
Spitters, as seen above, drop from no range penalty at 6 tiles to -70% at 7 tiles. That is hardly a reasonably intended behaviour. And as anybody who has ever used a x4 scope at full range can confirm, there is a mysterious point past which you can see, but cannot actually hit anything, because rifle's accurate range is... what, 12 tiles? So past it, you suddenly get _-130%_ accuracy at first tile past and even worse by the edge of the view. That is beyond any reason.

fira (approved):

ThePiachu (approved):

PR #1893, opened 2022-12-09
Fire penetration in chem grenades

RenaRenaRe:
# About the pull request
Makes it so chemicals with fire penetration (currently only available through research) can be used in chemical fires such as OT nades to make the fire have penetration. Previously fire penetration would only work if used in a custom flamer tank.
It requires at least 10u of the fire penetration chem, I chose that number since it's consistent with other OT mechanics such as 10u phoron making incen shrapnel. If people think that only requiring 10u is too powerful I could change it, but personally I don't think that's an issue at all.
The chemical with fire penetration also has to have another fire property on it (such as oxidizing or fueling) for it to work, which matches what the wiki says and makes sense since the chemical should have to actually be flammable for this to work.

Bonus feature: Adds ravagers to the OT demolition simulator, so you can actually see it working yourself.

I see no reason why fire penetration should only work on flamers, allowing it to work in chem grenades is more intuitive and increases the options of what you can do with it.

<details>

https://user-images.githubusercontent.com/118639975/206601472-b8d1d976-e0eb-4632-a5df-c219328f5a35.mp4

</details>

# Changelog

:cl:
add: Fire penetration now works with OT grenades and other chemical fires, not just as flamer fuel.
add: Ravagers are now a dummy option in the demolition simulator.
/:cl:


fira:
But why require a minimum of 10u at all?

RenaRenaRe:
> But why require a minimum of 10u at all?

Do you think it should just be a single unit turns a whole fire grenade into fire penetrating? It could work that way I guess but that feels kinda weird. But honestly I don't mind either way so I'm open to feedback on what everyone else thinks about that

code/modules/reagents/Chemistry-Holder.dm
```diff
@@ -624,6 +625,8 @@
 			intensity += R.intensitymod * R.volume
 			duration += R.durationmod * R.volume
 			radius += R.radiusmod * R.volume
+			if(R.fire_penetrating && R.volume >= 10)
```
stanalbatross, line 628: use define for 10

stanalbatross (changes requested):

code/modules/reagents/Chemistry-Holder.dm
RenaRenaRe: Would it be good to also add defines for some of the old code like for phoron requiring 10u and phosphorus having a radius of volume/10?

stanalbatross: yeah

RenaRenaRe:
So I replaced a bunch of magic constants with defines, hopefully the chemistry define file was the right place to put them (these are chemical explosions/fires after all).
The part with FLAMESHAPE_IRREGULAR I'm like 99% sure is unused code but I didn't want to remove it just in case, so I just added a define for that too

harryob (approved):

code/__DEFINES/chemistry.dm
```diff
@@ -264,3 +264,15 @@
...
+	Chemical explosions/fires
+*/
+#define CHEM_FIRE_PENETRATION_THRESHOLD	10 //The minimum amount of a fire penetrating chemical required to turn a fire into fire penetrating
+#define CHEM_FIRE_STAR_THRESHOLD		30 //An intensity greater than this will cause a fire to be star shape
+#define CHEM_FIRE_IRREGULAR_THRESHOLD	15
+#define CHEM_FIRE_PHOSPHORUS_PER_RADIUS	10 //Amount of phosphorus that equals 1 radius of white phosphorus smoke
+#define EXPLOSION_PHORON_THRESHOLD		10 //The minimum amount of phoron required to turn shrapnel into incendiary shrapnel
+#define EXPLOSION_MIN_FALLOFF			25
+#define EXPLOSION_BASE_SHARDS			4
+#define INCENDIARY_SHARDS_MAX_REDUCTION	4 //The maximum amount of shards is divided by this number if the shards are incendiary
```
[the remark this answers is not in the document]
RenaRenaRe: I am not using that and didn't realize it mattered.
Does this just mean there should always be an empty line at the end of a file?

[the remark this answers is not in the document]
harryob: > I am not using that and didn't realize it mattered.
> Does this just mean there should always be an empty line at the end of a file?

it's part of our code standards that there's a new line at the end of the file - the EditorConfig plugin on VSC enforces this automatically

PR #1894, opened 2022-12-09
Slug Nerf

usnpeepoo:
# About the pull request

This PR nerfs slugs by removing the 2 second super slow. It also increases the damage by 10. This is to reduce the insane potential slugs have with their slow, and to even out slugs a little more.

Slugs are very powerful, and the fact that it has a 7 second slow (2 of which are superslow seconds) is insane. It knocks down as well, which leaves time to quickly equip another weapon and is simply too much. This PR would fix that issue, however it will not leave slugs in a terrible state as it still has the slow, and marines could possibly still kill the slugged xeno within the time frame (without the need for a 7 second slow).

i made a private server, tested the damage and slow on both runner/lurker/warrior/other t1s. It feels like a reasonable amount.

</details>

# Changelog
:cl:
balance: Slugs now longer superslow.
balance: Slugs have their damage increased by 10.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -135,8 +135,7 @@
 		var/mob/living/carbon/Xenomorph/target = L
 		to_chat(target, SPAN_XENODANGER("You are shaken and slowed by the sudden impact!"))
 		target.apply_effect(0.5, WEAKEN)
-		target.apply_effect(2, SUPERSLOW)
-		target.apply_effect(5, SLOW)
+		target.apply_effect(4.5, SLOW)
 	else
 		if(!isYautja(L)) //Not predators.
 			L.apply_effect(1, SUPERSLOW)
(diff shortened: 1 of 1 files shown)


fira:
That's not how status effects work.
 - 5 does not mean 5 seconds but 5 ticks of Human Life controller which should result in random between 8 and 10 seconds depending on ticking phase
 - The durations do not stack, Slow and SuperSlow reduce separately with SuperSlow overriding, so it is really 5 and not 7

usnpeepoo:
> 

so would i reduce or increase it then

Diegoflores31:
werent direct number changes not allowed ?

fira:
> so would i reduce or increase it then

I dunno depends on what you want to do. But I'm just putting forward the fact that status effects balancing is pretty awkward and frankly a waste of time. That being said the SuperSlow is real random as a result (2 to 4 seconds) so it might make sense to remove it...

stanalbatross:
kinda unjustified nurf... slugs are not that strong... gonna close

PR #1921, opened 2022-12-11
Layering issues fix

kugamo:
# About the pull request

There are alot of layering issues rn, this PR takes RADICAL sets to resolve them.

Possibly insanely lag inducing, possibly the answer to all layering issue for all time.

It remains to be seen.

Layering issues are evil and deserve to be smited, this PR is an enactment of god's righteous punishment.

# **BEFORE:**

https://user-images.githubusercontent.com/16618648/206926848-730f98cc-00dc-4309-b87f-6a4c094f66fa.mp4

https://user-images.githubusercontent.com/16618648/206927199-2d6f1ebb-cd6d-4f4e-aea2-f04e2c84e803.mp4

https://user-images.githubusercontent.com/16618648/206927200-c2a08678-2f55-4d0b-b348-cff727f40d9f.mp4

# **AFTER:**

https://user-images.githubusercontent.com/16618648/206926841-6d557f96-51d4-435a-b647-d27a7fd74a7d.mp4

https://user-images.githubusercontent.com/16618648/206926857-ae326b4a-7d01-4ee7-a1c2-a3fe3939c6ac.mp4

https://user-images.githubusercontent.com/16618648/206926859-5ebcd551-cab2-4518-aa92-12054b20b3d5.mp4

# Changelog

:cl:
fix: Fixed a bunch of layering issues.
/:cl:

### DIFF
--- a/code/__DEFINES/layers.dm
+++ b/code/__DEFINES/layers.dm
@@ -68,6 +68,8 @@
 
 #define LOWER_ITEM_LAYER 2.99 //for items that should be at the bottom of the pile of items
 
+// ------------------ BEGIN FLUX ZONE -------------------------> flux zone layers are managed by their y position and are set dynamically live (3.0 --> 5.0)
+
 //#define OBJ_LAYER 3
 
 #define ABOVE_SPECIAL_RESIN_STRUCTURE_LAYER 3.01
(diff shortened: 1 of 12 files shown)


fira (commented):
Uhhhhhh i'm not sure this is a good idea...

fira:
Okay there's a bunch of issues both practical and conceptual, i'm just gonna put them up as comment since it's rather broad:
 * This will break any mob layer dynamic changes because every movement will reset to initial() 
 * For example this will interfere with layering for lying down, and xeno hiding
 * This should use Moved() not Move() because you don't want to update this on failed movement
 * This won't work on anything forceMove()'d
 * You could use setDir() instead but this will still have most of the above caveats
 * Using y/256, while clever, is pretty arbitrary and uncertain, maps can be bigger than that
 * All that math should be using defines such as OBJ_LAYER
 * That +0.2 is a layer define of its own
 * Shouldn't rely on numerical values of directions as in `dir < 3` 
 * an override on something like /mob/Move won't destroy performance, but is still somewhat imapcting

fira:
There was a project that done Three-Quarter perspective for SS13 i think, could be worth looking into how they handled this problem

kugamo:
Theres no stricture on only using initial(layer), the function works fine when fed redefined layers like in the case lying down. I just didnt account for  that scenario, i did it for table flipping as you can see

PR #1954, opened 2022-12-15
Fixes blood and footprints blocking weed clearing.

carlarctg:
# About the pull request

Fixes blood and footprints blocking weed clearing.

Hitting a turf with any item will first check if the item has weeds on it, and attack them if so. You probably shouldn't be able to pry up floor tiles while the floortile is covered in alien resin..

Nah.

<details>

</details>

# Changelog

:cl:
fix: Fixes blood and footprints blocking weed clearing.
fix: Hitting a turf with any item will first check if the item has weeds on it, and attack them if so. You probably shouldn't be able to pry up floor tiles while the floortile is covered in alien resin..
/:cl:

### DIFF
--- a/code/game/objects/effects/decals/cleanable/blood/blood.dm
+++ b/code/game/objects/effects/decals/cleanable/blood/blood.dm
@@ -5,6 +5,7 @@
 	density = 0
 	anchored = 1
 	layer = ABOVE_WEED_LAYER
+	mouse_opacity = 0
 	icon = 'icons/effects/blood.dmi'
 	icon_state = "mfloor1"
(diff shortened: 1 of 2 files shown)


code/game/objects/effects/decals/cleanable/blood/blood.dm
```diff
@@ -5,7 +5,6 @@
 	density = 0
 	anchored = 1
 	layer = ABOVE_WEED_LAYER
-	mouse_opacity = 0
```
morrowwolf, line 8: Maybe I'm missing something but why change mouse opacity?

carlarctg: easerasedf

stanalbatross (approved):

PR #1969, opened 2022-12-17
removes extra nanoui code that resulted in double pop up menus with the evac pods

### LINKED ISSUE #1673: New escape pod tgui doesn't replace old ui
## Description
The new escape pod control tgui made here: #1176 doesn't actually replace the old UI, it just layers over it, though the old UI isn't what it was anymore either. I've made a screenshot

pod issue
(I know I probably should have made a recording and moved the new ui, but trust me, at least for me it happens every time)

It should just get rid of the old UI box entirely, no?

## Steps to reproduce
Go into an escape pod and click on the pod control panel

harryob:
there is no justifiable reason that the shuttlepods are so spaghetti code. i must rip out all of this

:cl:
fix: you don't get two uis for the escape pods anymore
/:cl:

closes #1673

### DIFF
--- a/code/game/machinery/embedded_controller/simple_docking_controller.dm
+++ b/code/game/machinery/embedded_controller/simple_docking_controller.dm
@@ -9,44 +9,6 @@
 	docking_program = new/datum/computer/file/embedded_program/docking/simple(src)
 	program = docking_program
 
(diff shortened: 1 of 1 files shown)


Huffie56:
this is linked to issue https://github.com/cmss13-devs/cmss13/issues/1673

harryob:
> this is linked to issue #1673

linked ty

stanalbatross (approved):

PR #1996, opened 2022-12-19
Carrier buffs and xeno ability toggles

ThePiachu:
# About the pull request

After playing a few rounds as various Carriers, it really felt they could use a small buff to work way better with how basically they are expected to build Hive Clusters now.

So the changes are:
- Vanilla and Shaman Carriers now have 500 plasma instead of 400 (would've given them 450 but the defines are in increments of 100). Eggsac still sits at 600.
- Base carrier can carry 8 eggs (up from 7), and Eggsac can carry 12 (up from 7). The latter especially suffered from not being able to carry more than its base variant.
- Shaman can now plat weeds (since we do have support for 5 xeno actions everything still fits). It was strange they could make special structures but not base weeds as a Drone evo strain...
- Eggsac's egg generation is now a toggle instead of an active ability. It uses plasma per tick and generates an egg every 30 seconds. This makes it less fiddly and lets you plant eggs more actively. Still the same cost overall.
- To support that, I added a new xeno ability type - active toggle. For now it's used by Hivelord for Resin Walker and Eggsac for generating the egg

Hopefully the active egg generation will actually make Eggsac a strong frontline xeno, since currently it's not that popular, not to mention it's still a liability for the Hive if they die...

Carriers are usually very plasma starved, especially when they are expected to build Hive Clusters for their Eggs, so mitigated that a bit. Eggsac carrying more eggs means it is compensated for losing huggers, and being able to generate eggs on an ongoing basis makes that ability less fiddly. The ability toggle makes it so the Resin Walker actually deactivates properly when you run out of plasma and cleans up the code a bit. Shaman being able to plant weeds means they can properly support the frontline without needing other Drones for the base weeds.

<details>

</details>

# Changelog

:cl:
balance: Base and Shaman Carriers now have 500 base plasma, up from 400. Eggsac remains at 600 plasma.
balance: Base and Shaman Carriers can now store up to 8 eggs, up from 7. Eggsac now can store a whopping 12 eggs, up from 7. Be careful not to die while carrying those and give Marines greeno material!
balance: Shaman can now plant weeds. This bumped all their other abilities one slot over shortcut-wise.
qol: Eggsac's egg generation ability is now a toggle that uses plasma per tick to generate eggs at an ongoing rate until full or out of plasma.
code: Hivelord's Resin Walker now uses the same toggle structure.
fix: Hivelord's Resin Walker now deactivates properly when you run out of plasma.
fix: Fixed a runtime around Shaman's ability initialisation.
/:cl:


harryob:
2 weeks, no ga required. has been extensively TMd

PR #1592, opened 2022-11-17
Resprites cats and kitten sprite [corgis and crabs no longer replaced]

sleepynecrons:
## About The Pull Request

Replaces sprites for both cats, the kitten, [EDIT: Doesn't replace corgis/crabs]

## Why It's Good For The Game

Legacy cat sprites bad

animalResprites

## Changelog

:cl: Sleepynecrons
imageadd: New icons for cats and the kitten. Added onmob icons for Jones when holding in hands.
imagedel: Removed old icons for cats and kitten
code: Added a holder for the kitten so object sprites and descriptions are consistent when picking one up.
/:cl:

PR #2020, opened 2022-12-19
Revert "Resprites cats and kitten sprite [corgis and crabs no longer …

silencer-pl:
# About the pull request

As it says on the tin, this reverts the new cat sprites added in #1592 

Why? I'm glad you asked!

Here is a incomplete list of the issues:

- The front and back views both depict the cat with its paws pointing inwards. This is absolute not how cats ever walk. The only time a cat keeps their paws like that when walking is when they are about three months old and are taking their first steps. Outward pointing paws is how cats maintain their balance among other this. Anyone who has ever seen a cat walk irl is guaranteed to immediately notice it, it plainly looks outright wrong, like something is wrong with the cat. 

image 
image

- Cat whiskers are usually more or less exactly as wide as the cats body assuming the cat has a healthy weight. They are also typically straight, not curved like they are on the sprite. This is because among other things, cats use their whiskers as a way of being able to tell if they can squeeze into a tight spot somewhere. On the sprite everything about the whiskers is wrong - they protrude from the wrong spot, they look like a moustache than actual whiskers and they are pure white, while irl cat whiskers typically match the colors of the cats fur.

image

- Cat whiskers protrude from above the cat's upper lip and below the cat's nose. On the sprite, the cats whiskers either look like they're coming from its chin, or outright look like they're attached to the back of the cats head. This, combined with the inward facing legs, makes the cat look like a cartoon character, something that stands in stark contrast to the more realistic, grounded art style present on CM (or even the old sprite). The resting sprite has a similar issue, where the whiskers look more like a very runty lion's mane and not actual whiskers.

image

- The Original Jones was a calico, which is a unique 3 color breed that has some very unique characteristics. This led to a lot of IC jokes about Jones being actually female (90+% of Calicos are female, all Calico males are sterile and are nearly guaranteed to have health issues) , this bit of IC lore was changed for no reason whatsoever.

Even if the above would be fixed, and I don't think they can due to pixel real estate, the cats build still looks extremely awkrward and curvy. In short, the cat sprites seems to be more suited towards the somewhat more... Cartoony look of a lot of current ss13 forks. The old sprite not only is way closer to how cats look and act IRL, it also seems to match CMs aesthetic way more.

There are further anatomy issues (the seems a bit too high and may be somewhat too big, there is an accented mane below the head which is not something plain shorthairs typically have. 

This information has been sourced by looking at any of the three cats and one kitten currently within my line of sight

The old cat sprite:

- Looks way more anatomically correct and as such does not seem like something ooc

- Matches the more realistic style of CM way better than the new one does

- Was a calico, a fairly unique but instantly recognizable cat among the more cat oriented among us versus trying to look somewhat like an American Shorthair, but failing at that.

- Did not seem to need any changes or upgrades in the first place.

</details>

# Changelog

:cl:silencer_pl
imageadd: Reverts all cat sprites to the ones previously used. Jones is a calico again.
/:cl:


code/modules/mob/holder.dm
```diff
@@ -64,31 +64,17 @@
 	desc = "It's a small maintenance robot."
 	icon_state = "drone"
 
-/obj/item/holder/Runtime
```
realforest2001, line 67: These are things that can be kept in, with tweaks to make them use appropriate sprites.

code/modules/mob/living/simple_animal/friendly/cat.dm
```diff
@@ -111,7 +111,6 @@
 	icon_state = "cat"
 	icon_living = "cat"
 	icon_dead = "cat_dead"
-	holder_type = /obj/item/holder/blackcat
```
realforest2001, line 114: See above

code/modules/mob/holder.dm
silencer-pl: Readded

code/modules/mob/living/simple_animal/friendly/cat.dm
silencer-pl: As above, readded

silencer-pl:
Sigh, I switched where one of the holders are by accident so the non dmi files still show changes, to prevent too much commits though, I'll leave it as is. Effectively all is being changed back now are the dmis.

code/modules/mob/holder.dm
```diff
@@ -63,7 +63,7 @@
 	name = "maintenance drone"
 	desc = "It's a small maintenance robot."
 	icon_state = "drone"
-
+	
 /obj/item/holder/Runtime
```
realforest2001, line 67: The only other thing I'd say on this is you could change the pathing so they all inherit and don't have to be 100% written out again, but that's not strictly necessary. Just removes having to specify icon_state for each one over and over.
Would probably look something like this
`/obj/item/holder/cat
/obj/item/holder/cat/kitten
/obj/item/holder/cat/jones
/obj/item/holder/cat/black
/obj/item/holder/cat/black/runtime`

silencer-pl: Haven't forgotten, but I'll look into this in a few. It can probably be simplified yeah.

silencer-pl:
An icon conflict most likely means this MR will have to be reopened as a new one anyway, which is what I'll do. This is totally not because I should have started this on my fork instead of just making a one click revert request considering that we're keeping some of the code. See you in a few :P

PR #2063, opened 2022-12-22
Adding the Amphibious Landing event map

ThePiachu:
# About the pull request

Recently someone ran an event on it by hand-loading it into the server. Figured it might be better if we actually have it in the code so it can be refined. At least for now it can be on a PR that could be loaded for the event if it's not good enough to merge for the time being.

Event maps are fun.

<details>

</details>

# Changelog

:cl: Monkeyfist, WarFan1815, ThePiachu
add: Added the Amphibious Landing event map to the codebase.
/:cl:


morrowwolf:
Consider this a second do not merge tag. I don't think we should add event maps to the main repo and I'm not sure Monkeyfist wants his map to be shared.

ThePiachu:
As discussed, the map is being uploaded to the Map Depo instead - https://github.com/cmss13-devs/cmss13-map-depot/pull/1 .

PR #2103, opened 2022-12-25
515 compatibility

harryob:
merry christmas CM you survive another major byond update

:cl:
code: 515 compatibility is here
/:cl:

this:
- uses libcall define instead of call for external calls
- uses the new PROC_REF, TYPE_PROC_REF and GLOBAL_PROC_REF macros everywhere
- fixes tgui
- renames anything that uses /final/ as this is a keyword now (sorry japanese and chinese, sorry stan)
- updates standards


stanalbatross (approved):
looks fine I hope it got TMed first!

PR #2106, opened 2022-12-25
Fix for contractors presets' flags.

Segrain:
# About the pull request

Followup to #1380. With null for flags presets do not get added to the global list of presets and thus cannot be accessed later.
#2105 beat me to it, but root typepaths that are not actually full presets probably do not need to be given any flags, so that they do not appear in the list themselves to the confusion of somebody trying to use Create Humans / Select Equipment / etc.

Is fix.

# Changelog

:cl:
fix: Contractors ERT once again gets its proper equipment.
/:cl:


harryob:
i just like inheritance too much (lazy)

harryob (approved):

PR #2133, opened 2022-12-28
(grep pt1) adds a CI test for the PROC_REF define, and re-enables CI tests that we pass

harryob:
we don't need any more .procs sneaking in when i'm not looking

this also re-enables a bunch of the other grep tests. eventually they'll all be back, when we're compliant


harryob:
CleanShot 2022-12-28 at 11 54 10@2x
like this

tools/ci/check_grep.sh
```diff
@@ -4,95 +4,132 @@ set -euo pipefail
...
+#if grep -P '^/*var/' code/**/*.dm; then
+#	echo
+#	echo "${RED}ERROR: Unmanaged global var use detected in code, please use the helpers.${NC}"
+#	st=1
+#fi;
+#echo "Checking for whitespace issues"
+#if grep -P '(^ {2})|(^ [^ * ])|(^	+)' code/**/*.dm; then
+#	echo
+#	echo "${RED}ERROR: space indentation detected.${NC}"
+#	st=1
```
stanalbatross, line 70: honestly we could uncomment this, it should all be gone now

tools/ci/check_grep.sh
```diff
@@ -4,95 +4,132 @@ set -euo pipefail
...
+#fi;
+#if grep -Pzo '/obj/structure/machinery/power/apc[/\w]*?\{\n[^}]*?pixel_[xy] = -?[013-9]\d*?[^\d]*?\s*?\},?\n' maps/**/*.dmm ||
+#	grep -Pzo '/obj/structure/machinery/power/apc[/\w]*?\{\n[^}]*?pixel_[xy] = -?\d+?[0-46-9][^\d]*?\s*?\},?\n' maps/**/*.dmm ||
+#	grep -Pzo '/obj/structure/machinery/power/apc[/\w]*?\{\n[^}]*?pixel_[xy] = -?\d{3,1000}[^\d]*?\s*?\},?\n' maps/**/*.dmm ;	then
+#	echo "${RED}ERROR: found an APC with a manually set pixel_x or pixel_y that is not +-25.${NC}"
+#	st=1
+#fi;
+#if grep -P '^/area/.+[\{]' maps/**/*.dmm;	then
+#	echo "${RED}ERROR: Vareditted /area path use detected in maps, please replace with proper paths.${NC}"
+#	st=1
```
stanalbatross, line 54: this too

tools/ci/check_grep.sh
harryob: handled in #2141

tools/ci/check_grep.sh
harryob: handled in #2142

PR #2138, opened 2022-12-28
(grep pt2) moves all global variables to be managed by the GLOB controller

harryob:
i’ll also enable the grep tests for this. consistency and the ci wants it

:cl:
refactor: all global variables are now managed by the correct controller
/:cl:


harryob:
conflicts for no reason given globals are in world.vars already

PR #2142, opened 2022-12-28
(grep part 1.6) removes a bunch of varedited areas, incorrectly adjusted apcs + uppercase in json

harryob:
these are all grep tests to make us nice and compliant


Agameofscones (approved):
nice mapping regex, checks out OK

Issue #1315, opened 2022-10-23
Splints


RedIBat:
## Description

You can apply splint to yourself, while it is in your first aid kit pouch or dropped on the ground.

## Difference between expected and actual behavior

Applying splint process should stop immediately if you put it in your pouch, drop it on the ground.

## Steps to reproduce

Pick up splint, click on your sprite to apply it. Drop it or press E to put it in your pouch. Splints still being applied, if your char does not move.

## Issue bingo (not required to be all checked)

- [ ] Issue could be reproduced at least once
- [ ] Issue happened in a recent (less than 7 days ago) round
- [ ] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Segrain:
One of many, many manifestations of a well known thing. I will get to fixing it, one day...

PR #2160, opened 2022-12-30
Fixes defib dropping, splint dropping/storing exploit/oversight.

somepersondunno:
# About the pull request

Fixes Issue #1315, as well as a common thing I see of medics starting the defib timer, then dropping their defib in order to do other things (pill, suture the patient, etc).

This fixes some nobrainer issues. If this causes any issues, please let me know, as I am very new to this whole thing.

<summary>I tested this on a private server and did everything imagineable to interrupt the actions (splinting and defibbing). I splinted myself, moving, dropping and storing the item, and each time the action was interrupted. I did the same to someone else. I also killed then defibbed a mob, and every time I tried something similar the action was interrupted, so this is working as intended.</summary>

# Changelog

:cl:
fix: fixed issues/exploits surrounding defibrillators and splints
/:cl:

### DIFF
--- a/code/game/objects/items/devices/defibrillator.dm
+++ b/code/game/objects/items/devices/defibrillator.dm
@@ -167,7 +167,7 @@
 	playsound(get_turf(src),'sound/items/defib_charge.ogg', 25, 0) //Do NOT vary this tune, it needs to be precisely 7 seconds
 
 	//Taking square root not to make defibs too fast...
-	if(!do_after(user, 7 SECONDS * user.get_skill_duration_multiplier(SKILL_MEDICAL), INTERRUPT_NO_NEEDHAND|BEHAVIOR_IMMOBILE, BUSY_ICON_FRIENDLY, H, INTERRUPT_MOVED, BUSY_ICON_MEDICAL))
+	if(!do_after(user, 7 SECONDS * user.get_skill_duration_multiplier(SKILL_MEDICAL), INTERRUPT_ALL|BEHAVIOR_IMMOBILE, BUSY_ICON_FRIENDLY, H, BUSY_ICON_MEDICAL))
 		user.visible_message(SPAN_WARNING("[user] stops setting up the paddles on [H]'s chest."), \
 		SPAN_WARNING("You stop setting up the paddles on [H]'s chest."))
 		return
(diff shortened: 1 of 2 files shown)


carlarctg:
does this mean swapping hands will break the action? because that would suck ass

somepersondunno:
> does this mean swapping hands will break the action? because that would suck ass

this is an unavoidable issue without programming in another interrupt, but forcing players to wait on tasks like attaching a gun, splinting or defibbing is just part of the game.

morrowwolf (changes requested):
Dropping defib and having it continue is intentional. Please put it back.

somepersondunno:
Fixed.

code/game/objects/items/devices/defibrillator.dm
```diff
@@ -167,7 +167,7 @@
 	playsound(get_turf(src),'sound/items/defib_charge.ogg', 25, 0) //Do NOT vary this tune, it needs to be precisely 7 seconds
 
 	//Taking square root not to make defibs too fast...
-	if(!do_after(user, 7 SECONDS * user.get_skill_duration_multiplier(SKILL_MEDICAL), INTERRUPT_NO_NEEDHAND|BEHAVIOR_IMMOBILE, BUSY_ICON_FRIENDLY, H, INTERRUPT_MOVED, BUSY_ICON_MEDICAL))
+	if(do_after(user, time_to_take * user.get_skill_duration_multiplier(SKILL_MEDICAL), INTERRUPT_NO_NEEDHAND, BUSY_ICON_FRIENDLY, target, INTERRUPT_MOVED, BUSY_ICON_MEDICAL))
```
harryob, line 170: you need to define time_to_take and redefine H to target in definition

harryob (changes requested):

Segrain:
> Dropping defib and having it continue is intentional. Please put it back.

No, it is absolutely _not_.

Segrain:
> > does this mean swapping hands will break the action? because that would suck ass
> 
> this is an unavoidable issue without programming in another interrupt, but forcing players to wait on tasks like attaching a gun, splinting or defibbing is just part of the game. edit: virtually every action that works similarly (on the same type of timer) requires you to wait.

No. Just no. You are trying to remove the entire mechanic that exists for a reason and purpose. Justifying it as fixing a bug is, as the expression goes, throwing out the baby with the water. I am going to save you the trouble of fixing everything else other reviewers have pointed out and simply close this on design principles.
(And yes, I will get to actually writing a proper fix. Soon. It is a sizeable work to go over every usage case of `do_after`'s interrupts and ensure that only those that need be affected are affected.)

PR #2162, opened 2022-12-30
reimplements multi-highlight

### LINKED ISSUE #122: Cannot make more than one highlighted words section
## Description

Goon chat's word highlighter is currently broken as it only allows one highlighted colour type to be shown.

It should allow for multiple words to be highlighted with a multitude of different colours.
Currently it only allows one.

## Steps to reproduce

Open up Goon chat's settings in game and see that only one text box is allowed to be used rather than multiple.
image

There should be an option to add more than one text box but currently as shown above, there is no option.

stanalbatross:
# About the pull request

reimplements old chat multi-highlight

fixes #122

image

easier highlighting and more customisable

<details>

</details>

# Changelog

:cl: Stanalbatross, TheDonkified
ui: re-added the old multiple highlight system, originally made by thedonkified. You can add up to 10 individual highlights, customise their colours, and toggle if they highlight the whole message or not.
ui: you can now use regex for text highlighting. The syntax is /[regex expression]/, eg /dog[0-9]/. There are no quantifiers, but you can use capture groups. The regex must read a minimum of two characters.
/:cl:

### DIFF
--- a/tgui/packages/tgui-panel/settings/SettingsPanel.js
+++ b/tgui/packages/tgui-panel/settings/SettingsPanel.js
@@ -163,7 +163,7 @@ const TextHighlightSettings = (props, context) => {
   const highlightSettings = useSelector(context, selectHighlightSettings);
   const dispatch = useDispatch(context);
   return (
-    <Section fill>
+    <Section fill scrollable>
       <Section level={2} p={0}>
         <Flex direction="column">
           {highlightSettings.map((id, i) => (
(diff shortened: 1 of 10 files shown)


harryob (approved):
we really should try to get this upstream because it’ll be overwritten next time we update tgui

tgui/packages/tgui-panel/settings/SettingsPanel.js
```diff
@@ -157,62 +152,41 @@ export const SettingsGeneral = (props, context) => {
...
+      </Button>
+    </Section>
+  );
+};
+
+const TextHighlightSettings = (props, context) => {
+  const highlightSettings = useSelector(context, selectHighlightSettings);
+  const dispatch = useDispatch(context);
+  return (
+    <Section fill>
```
harryob, line 166:
```suggestion
    <Section fill scrollable>
```
otherwise you can add too many sections and it goes off the bottom of the window

harryob (changes requested):

PR #2169, opened 2022-12-31
Fixes xeno QDELing "currently transported" marines after missing hijack

### LINKED ISSUE #2148: Getting deleted instead of turned Forsaken on hijack takeoff deletes any mob you swallowed as well
### Testmerges

#1795, #1767, #2045

### Description of the bug

As the description says, in the rather exotic case of:

- A xeno having swallowed someone and transporting them
- The Queen taking off and triggering the hijack mechanic
- The xeno not passing a Forsaken check, meaning they get deleted instead

It the xeno happens to be transporting a marine, the Marine will be deleted as well even if they were alive and not capped. This just happened during a round and was ahelped and after some debubbing, present staff nailed it down to the above mechanic.

I would expect that a xeno getting deleted would just dump out its stomach contents, whatever they may be, the same they do when they get put into crit or whatever.

1. Swallow someone
2. Get deleted by the hijack/Forsaken mechanic while having someone in your stomach

### Issue Bingo

moltoretardo:
# About the pull request
Made a condition checking the content of the xeno and forcing it to throw up if not empty

Fixes #2148
I'm a free GBP enjoyer (if I did it right)

<details>

</details>

# Changelog

:cl: Molto
fix: fixed marine qdel on hijack
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
@@ -671,10 +671,6 @@
 		tracked_marker.xenos_tracking -= src
 		tracked_marker = null
 
-	if(stomach_contents.len)
-		devour_timer = 0
-		handle_stomach_contents()
-
 	if(mind)
 		mind.name = name //Grabs the name when the xeno is getting deleted, to reference through hive status later.
 	if(IS_XENO_LEADER(src)) //Strip them from the Xeno leader list, if they are indexed in here
--- a/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
@@ -671,6 +671,10 @@
 		tracked_marker.xenos_tracking -= src
 		tracked_marker = null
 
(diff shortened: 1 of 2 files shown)


harryob (changes requested):
should be done in mob/living/carbon/Xenomorph/Destroy()

moltoretardo:
> should be done in mob/living/carbon/Xenomorph/Destroy()

I globalized it like asked, but since it now applies to every instance of xeno getting QDEL'd it looks a bit silly when one jumps into the ASRS pit while holding something, since it'll jump, throw up, get deleted, marine get deleted.

harryob:
> > should be done in mob/living/carbon/Xenomorph/Destroy()
> 
> 
> 
> I globalized it like asked, but since it now applies to every instance of xeno getting QDEL'd it looks a bit silly when one jumps into the ASRS pit while holding something, since it'll jump, throw up, get deleted, marine get deleted.

i always forget about that infernal pit. feel free to revert - that's goofy. i'll uhh refactor it later

moltoretardo:
Reverted as asked

harryob (approved):

Issue #2182, opened 2023-01-01
Van takes ages to enter as a human


chowdermcarthor:
### Testmerges

#2045, #2156, #1767

### Description of the bug

Trying to enter or exit a van as a human takes far longer than normal, around 10-20 seconds. Xenos are not affected and can enter/exit the van in around 2-3 seconds.

### What's the difference with what should have happened?

You should be able to enter the van in seconds.

### How do we reproduce this bug?

1. Be a human.
2. Find a van.
3. Try to enter it and wait an absurdly long time to do so.

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #2186, opened 2023-01-01
stack qol and vehicle entry speed fix

stanalbatross:
# About the pull request

removes the annoying "merge sheets" popup that happened when you clicked a stack in your offhand

now you can split off stacks with altclick

also fixed carlarc vehicle entry speed bug

Fixes https://github.com/cmss13-devs/cmss13/issues/2182

# Changelog

:cl:
fix: vehicle entry speeds will no longer be extraneously multiplied
qol: you can now alt-click stacks to split off items from them. As a result of this, attacking a stack in your offhand with one of the same type will no longer ask you how many you want to merge into it.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -296,6 +296,23 @@ Also change the icon to reflect the amount of sheets, if possible.*/
 		if(!oldsrc)
 			break
 
+/obj/item/stack/clicked(mob/user, list/mods)
+	if(mods["alt"])
+		var/desired = tgui_input_number(user, "How much would you like to split off from this stack?", "How much?", 1, amount-1, 1)
+		if(!desired)
+			return
+		var/obj/item/stack/newstack = new src.type(user, desired)
+		transfer_fingerprints_to(newstack)
+		user.put_in_hands(newstack)
+		src.add_fingerprint(user)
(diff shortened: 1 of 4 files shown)


moltoretardo:
Fixes #2182

(But will never fix my broken heart from this blatant steal of my GBPs)

fira (approved):

code/game/objects/items/stacks/stack.dm
```diff
@@ -296,6 +296,23 @@ Also change the icon to reflect the amount of sheets, if possible.*/
 		if(!oldsrc)
 			break
 
+/obj/item/stack/clicked(mob/user, list/mods)
+	if(mods["alt"])
+		var/desired = tgui_input_number(user, "How much would you like to split off from this stack?", "How much?", 1, amount-1, 1)
+		if(!desired)
+			return
+		var/obj/item/stack/newstack = new src.type(user, desired)
```
Segrain, line 304: Feels like this should once again check that desired is indeed less than amount (what if you altclick it several times in a row and only then input numbers, for example?). Something like `if(!use(desired)){return}` before making new stack instead of `use(desired)` after.

Segrain (changes requested):

harryob (approved):

PR #2200, opened 2023-01-03
Fix for splitting stacks.

Segrain:
# About the pull request

Followup to #2186.

Without this return, `click()` proceeds as normal after creating the split stack, which results in `attackby()` on remains of old stack by new stack, which results in consuming old stack into new stack entirely, which results in no splitting at all.
Also a sanity check to prevent certain black sorcery tricks.

# Changelog

:cl:
fix: Splitting stacks actually works now.
fix: Aliens and telekinetics no longer can abuse splitting stacks.
/:cl:


harryob (approved):
😬

PR #2204, opened 2023-01-04
TGUI autolathe

stanalbatross:
# About the pull request

image

converted the autolathe to TGUI and somewhat cleaned up its code

# Changelog

:cl:
ui: autolathes now use tgui
code: cleaned up autolathe code, made stuff use defines, proc-ified some stuff, made bools use TRUEFALSE
code: made a few common vars use TRUEFALSE
/:cl:


tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,272 @@
...
+      <CurrentlyMaking />
+    </Section>
+  );
+};
+
+const CurrentlyMaking = (props, context) => {
+  const { act, data } = useBackend(context);
+  const { currently_making } = data;
+
+  return currently_making ? (
```
mullenpaul, line 69: Why use a ternary operator? Make the assumption that currently_making is not undefined and use conditional rendering to render this component

tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,272 @@
...
+    <Fragment>
+      <Box height="5px" />
+      <Button
+        fluid
+        textAlign="center"
+        content={
+          'Currently making: ' +
```
mullenpaul, line 76: I'm not a huge fan of having decision making etc inside the jsx like this - if the conditional rendering above is used then this can be refactored out into a const and it'll be easier to read. You could also use string interpolation to simplify this

tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,272 @@
...
+    </Section>
+  );
+};
+
+const CurrentlyMaking = (props, context) => {
+  const { act, data } = useBackend(context);
+  const { currently_making } = data;
+
+  return currently_making ? (
+    <Fragment>
```
mullenpaul, line 70: Fragment can use <> </> as shorthand

tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,272 @@
...
+    queued,
+    printables,
+    selectable_categories,
+    queuemax,
+  } = data;
+
+  return (
+    <Section title="Queue">
+      <Flex direction="column">
+        {queued.map((item) => {
```
mullenpaul, line 101: If your code segment immediately returns a component then you don't need this - point it directly to the component.

`=> (<Flex...`

tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,272 @@
...
+          return (
+            <Flex.Item key={item}>
```
mullenpaul, line 103: You can't use a complex object as a key - it should be a simple number/string etc. When trying to determine if the component needs to be re-rendered, it does an equality check on the key, if this is a complex object then it checks if it is the same object (simple equality check) and always returns false.

tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,272 @@
...
+  } = data;
+
+  const [currentSearch, setSearch] = useLocalState(
+    context,
+    'current_search',
+    ''
+  );
+
+  const categories = [];
+  for (let i = 0; i < printables.length; i++) {
```
mullenpaul, line 160: You don't really want to use for loops unless absolutely necessary, this should be done using a filter and map

`printables.filter(x => categories.includes(x.recipe_category)).map(x => x.recipe_category)`

tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,272 @@
...
+  const [currentCategory, setCategory] = useLocalState(
+    context,
+    'current_category',
+    'All'
+  );
+
+  const filteredPrintables = printables.filter(
+    (val) =>
+      (val.recipe_category === currentCategory || currentCategory === 'All') &&
+      val.name.toLowerCase().match(currentSearch)
```
mullenpaul, line 176: I believe the currently accepted way to do a case insensitive search (besides regex) is localeCompare https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare - it also has good non-English language support

tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,272 @@
...
+              ))}
+            </Tabs>
+          </Stack.Item>
+        )}
+        <Stack.Item>
+          <Input
+            fluid
+            value={currentSearch}
+            placeholder="Search for an item"
+            onInput={(e, value) => setSearch(value.toLowerCase())}
```
mullenpaul, line 201: If you use localeCompare above then you don't need to make this lower case

tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,272 @@
...
+          />
+        </Stack.Item>
+        <Stack.Item grow>
+          <Flex direction="column">
+            {filteredPrintables.map((val, index) => {
```
mullenpaul, line 206: Same as above regarding opening a code block then immediately returning

tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,272 @@
...
+                              val.materials[Object.keys(materials)[1]]
+                              : val.materials[Object.keys(materials)[0]]
+                                ? val.materials[Object.keys(materials)[0]]
+                                : val.materials[Object.keys(materials)[1]]) +
+                            ') '}
+                        </Button>
+                      </Flex.Item>
+
+                      {(val.has_multipliers && (
+                        <Fragment>
```
mullenpaul, line 237: <Fragment> can become <> </>

tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,272 @@
...
+                              ', ' +
+                              val.materials[Object.keys(materials)[1]]
+                              : val.materials[Object.keys(materials)[0]]
+                                ? val.materials[Object.keys(materials)[0]]
+                                : val.materials[Object.keys(materials)[1]]) +
+                            ') '}
+                        </Button>
+                      </Flex.Item>
+
+                      {(val.has_multipliers && (
```
mullenpaul, line 236: This first bracket ( is not necessary

tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,272 @@
...
+                                      })
+                                    }
+                                  />
+                                </Flex.Item>
+                              ))}
+                            </Flex>
+                          </Flex.Item>
+                        </Fragment>
+                      )) ||
+                        null}
```
mullenpaul, line 258: This seems like it isn't required

tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,272 @@
...
+                    </Flex>
+                  </Box>
+                  <Box height="2.5px" />
+                </Flex.Item>
+              );
+            })}
+          </Flex>
+          <Box height="10px" />
+          <ElectricalPanel />
```
mullenpaul, line 267: This seems out of place in this component - I feel that it should be in the root component

mullenpaul (changes requested):
I really don't like the need to iterate through Object.keys of variables being passed in. It makes the code a lot more difficult and verbose. Unfortunately, the only way to get around this is to rewrite how the variables populate the ui_data, which is pretty painful

mullenpaul:
I haven't had a chance to boot this up and look at the UI but I've had a look through the code itself. Can do a UI review tomorrow

tgui/packages/tgui/interfaces/Autolathe.js
stanalbatross: eslint or prettier forces it in for some reason

tgui/packages/tgui/interfaces/Autolathe.js
stanalbatross: if I don't use the || null it puts this weird plaintext zero after the button for ones with no multipliers

tgui/packages/tgui/interfaces/Autolathe.js
stanalbatross: if it's not in here it ends up layering behind the list

tgui/packages/tgui/interfaces/Autolathe.js
stanalbatross: couldn't seem to get this working properly

tgui/packages/tgui/interfaces/Autolathe.js
mullenpaul: It's not the end of the world, I would say either the parent component neds to have its placement managed to refactor there or leave as is.

mullenpaul:
UI looks fine - would be good to add spacers between the multiple purchase options

mullenpaul (approved):

PR #2209, opened 2023-01-04
gbp tweak

stanalbatross:
# About the pull request

tweaked around some GBP values

"Feature" = -5 -> "Feature" = -6: feature and balance have the same cost, and I wanted feature to be more expensive because 1. it's more review work 2. so people will properly tag their PRs

"Refactor" = 10 -> "Refactor" = 6: 10 for a refactor is a lot when people fairly liberally use the tag. Down to 6 so it mirrors feature

"Mapping" = -3: this is definitely some review work, plus what aint broke don't fix. Therefore this is neutralised by a fix

"UI" = 3: tgui is awesome

# Changelog

no player facing changes


Nanu308 (approved):
lgtm

PR #2210, opened 2023-01-04
modern say/me/ooc/looc input, comms hotkey (it has a preference)

harryob:
image

- ***it has a preference it has a preference it has a preference it has a preference it has a preference it has a preference it has a preference***

image image image

ok now that's cleared up

image
it has modes for say, me, comms, ooc and looc

image

when you do :v it'll look like this 
image
when you do #b it'll look like this
image

***it has a preference it has a preference it has a preference it has a preference it has a preference it has a preference it has a preference***

you can drag around the top/left/bottom/right lines to move the input where you want it

typing indicators are also different
image image image

***it has a preference it has a preference it has a preference it has a preference it has a preference it has a preference it has a preference***

though there is the option to interrupt people typing when they take damage - given that this ***HAS A PREFERENCE*** people would just disable it out of annoyance, so i decided not to enable this

🆑
ui: the say input is now modernised - this has a preference and can be disabled in the preferences tab / settings on character setup
/🆑


stanalbatross (changes requested):
image
medsci colour seems to break it...

Crowbar764:
Does this have a preference?

harryob:
> Does this have a preference?

unclear at this time

Chainsawicus:
When you have typing indicators off and select to go back to the old style, it permanently enables typing indicators in the new style that you cannot turn off.

harryob:
> When you have typing indicators off and select to go back to the old style, it permanently enables typing indicators in the new style that you cannot turn off.

fixed

harryob:
fixed bubbles hanging around after death / unconsciousness (i had to tie into life.dm in an ugly way, if it's lag inducing to add that access and proc call, i can always do a refactor to set_stat like tg)

Agameofscones (approved):
I like your sprites you funny coder man. Sprites approved

stanalbatross (approved):

PR #2212, opened 2023-01-04
asset log antispam

stanalbatross:
# About the pull request

prevents asset log from taking up 50 quadribillion mbs of log space

the tgui log i commented out is irrelevant, already handled by the opening log

# Changelog

:cl:
admin: asset sending and tgui making is no longer logged to save space 
fix: fixes the spawn pool splash overlay sticking around for ages
/:cl:


Nanu308:
This should probably go into it's own logfile that we can clean out more often honestly, hear with @fira

harryob:
this is a config option anyway, and one we have disabled. it still spams the in game asset log, mind, but who cares about that list

PR #2214, opened 2023-01-04
added matchword and matchcase back into chat highlight

stanalbatross:
# About the pull request

well, that was surprisingly easy.

image
image

# Changelog

:cl:
ui: added the match word and match case options back to highlight settings, match word is overidden if regex is used
/:cl:


harryob (approved):
lgtm

PR #2226, opened 2023-01-05
pickup keybind

harryob:
you now remember your last two dropped items (dropped in any manner), and can automatically pick it up with F (you can rebind this as you please)

:cl:
qol: instead of using macros for pickup, you can now use the built in keybind - defaulting to F, but can be changed. it remembers your last two dropped items
/:cl:


morrowwolf (changes requested):
I'd like to see adjacent tiles as well picking up if possible.

morrowwolf (changes requested):
Move the stuff around like we talked on discord about

morrowwolf (approved):

PR #2241, opened 2023-01-07
adds a dme validator

harryob:
makes sure everything in the game files is either included in the dme or deleted


harryob:
draft because although we should do this there's a bunch of things that i don't wanna just delete from the code yet (ie, the emote keybinds (we need to refactor emotes))

stanalbatross (changes requested):
image
#include these instead of removing, they're useful

harryob:
> image
> #include these instead of removing, they're useful

wait all of these work why weren't they included in the first place

stanalbatross (approved):

Issue #2227, opened 2023-01-05
Forced Movment over Weeds/Sticky resin incurs Move Delay


MaximusRexCM:
### Testmerges

Yes, but the bug has been present for 7 years.

### Description of the bug

This is an issue that has been in the game for 7+ years. If a human mob is forcibly moved over weeds/sticky resin, it will stack movement delay equal to the number of weed/sticky tiles transported over. This applies after being pushed, dragged, or warrior flung. After the forced movement, you get one tile of normal movement, then are frozen for a variable amount of time depending on how far you were force-moved. 

### What's the difference with what should have happened?

When forcibly transported over weeds/sticky resin, you should NOT be frozen after taking one step. 

### How do we reproduce this bug?

1. Human mob stands near weeds
2. Human mob is flung/dragged/pushed over weeds
3. Human mob takes one normal speed step, then is frozen. 
...

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Nanu308:
I've confirmed this to be the case as well where movement delay can stack up when pushed by another mob across something that inflicts movement delay.

Anthem2000:
The dragging on weeds and warrior fling should be covered in the pr (warrior fix should apply to all tiles, not just weeds).

PR #2242, opened 2023-01-07
Fix for slowdown stacked tile movement when flung or dragged onto tile.

Anthem2000:
# About the pull request
Fix for a few issues outlined in #2227, basically, for combat scenarios where there are weeds the slowdown tiles should no longer be stacked if a mob is dragged onto the weeds by another mob, as well as if the warrior flings a mob over some stacked tiles. The issue with the harm intent push has yet to be solved though. This is my first attempt at a bug fix and I learned dm 2 days ago so please don't bully me.

It would be ideal if slowdown tiles didn't stack.

Did some manual testing and it seemed to work, going to do more in-depth testing in the following few days (I have a strong feeling there might be some performance issues with my implementation but we will see). 

# Changelog

:cl:
fix: fix for the stacked slowdown tiles
/:cl:

### DIFF
--- a/code/modules/cm_aliens/XenoStructures.dm
+++ b/code/modules/cm_aliens/XenoStructures.dm
@@ -158,12 +158,12 @@
 /obj/effect/alien/resin/sticky/Crossed(atom/movable/AM)
 	. = ..()
 	var/mob/living/carbon/human/H = AM
-	if(istype(H) && !H.lying && !H.ally_of_hivenumber(hivenumber))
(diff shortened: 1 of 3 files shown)


carlarctg (commented):
I didn't even know this bug existed, but I'm not sure about this fix. Pretty sure you can just check to see if the target is being thrown and don't add any slowdown in that scenario.

`if(AM.throwing)
    return`

PR #2249, opened 2023-01-07
Brain Slugs

realforest2001:
# About the pull request
Ports cortical borers from Paradise SS13 (as of 14 months ago)

For event use only.

<details>

</details>

# Changelog

:cl:
add: Ports Paradise SS13 Cortical Borers.
add: Dead borers can be picked up and put into a grinder to extract Cortical Enzyme which is used to make an anti-borer sedative. Anti-parasite drugs will also sedate a borer. An overdose on Anti-parasite drugs will force a borer out of the host, this will not work with the modified enzyme.
add: If a borer's host dies, they will be forced to release control. Once the host goes permadead, they'll be forcibly ejected from the corpse.
/:cl:


code/modules/borer/borer.dm
```diff
@@ -0,0 +1,301 @@
...
+
+	. = ..()
+
+
+/mob/living/simple_animal/borer/Life(delta_time)
+	..()
+
+	if(host)
+		if(!stat && host.stat != DEAD)
+			if((host.reagents.has_reagent("sugar") && !host.reagents.has_reagent("benzyme")) || host.reagents.has_reagent("bcure"))
```
carlarctg, line 169: probably should make these a chem property instead

code/modules/borer/borer.dm
```diff
@@ -0,0 +1,301 @@
...
+				if(!docile)
+					if(controlling)
+						to_chat(host, SPAN_XENODANGER("You feel the soporific flow of sugar in your host's blood, lulling you into docility."))
```
carlarctg, line 172: probs should also change this to 'a chemical in the blood makes you fall asleep' or whatnot, so it doesnt specify sugar

code/modules/borer/borer_chemicals.dm
```diff
@@ -0,0 +1,103 @@
+/datum/borer_chem
```
carlarctg, line 1: make sure they have CM-based descs and names - 'anti toxin' stands out

code/modules/borer/borer.dm
realforest2001: They're meant to know what the prevention is, though I may look at changing it into a trait like you suggested.

code/modules/borer/borer_chemicals.dm
realforest2001: Fiddly mainly because the chem name in the menu is the ID it uses, it can be changed but would require some reformatting on what it's checking. Would be nice though, so definitely on the list to do.

ColonialMarinesALPHA.dme
```diff
@@ -136,6 +136,7 @@
 #include "code\__HELPERS\shell.dm"
 #include "code\__HELPERS\text.dm"
 #include "code\__HELPERS\type2type.dm"
+#include "code\__HELPERS\unique_ids.dm"
```
harryob, line 139:
```suggestion
```

harryob (changes requested):

Agameofscones (approved):
ability icon states are good.

morrowwolf:
Just for TMs so closing.

PR #2252, opened 2023-01-08
Changes the MST to have rubber bullets at roundstart.

NewyearnewmeUwu:
# About the pull request

This changes the mess technician to have rubber bullets in their pistol belt rig as well as a rubber mag loaded in their M4A3.

Mess Technician has had a reputation for being a somewhat griefy role, and it's kind of unreasonable that the cook would start with lethal ammo when similar roles do not. They can still acquire lethal mags from various places if need be, but this should help reduce the number of mess tech incidents.

<details>

</details>

# Changelog

:cl:
add: Added a new 'training' belt containing only rubber M4A3 mags.
balance: Mess Technicians now spawn with rubber mags in their pistol belt.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -987,6 +987,11 @@
 	for(var/i = 1 to storage_slots - 1)
 		new /obj/item/ammo_magazine/pistol(src)
 
+/obj/item/storage/belt/gun/m4a3/training/fill_preset_inventory()
+	handle_item_insertion(new /obj/item/weapon/gun/pistol/m4a3/training())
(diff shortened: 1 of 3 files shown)


morrowwolf:
This will lead to so many people shooting the hell out of the MT thinking he is firing with lethals.

NewyearnewmeUwu:
> This will lead to so many people shooting the hell out of the MT thinking he is firing with lethals.

This is a fair concern, but my hope is that it will lead to less unjustified usage. The CL also has a pistol with rubber bullets, and I do not see the CL get in as many shooting battles as MSTs.

PR #2259, opened 2023-01-08
orbit update

harryob:
fun update of the orbit menu from tg

thanks jlsnow!

port of tgstation/tgstation#70547

CleanShot 2023-01-08 at 19 57 37@2x

:cl:
ui: the orbit menu gives you a little extra info
/:cl:


tgui/packages/tgui/interfaces/Orbit/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+import { capitalizeFirst } from 'common/string';
+import { useBackend, useLocalState } from 'tgui/backend';
+import { Button, Collapsible, Icon, Input, LabeledList, Section, Stack } from 'tgui/components';
+import { Window } from 'tgui/layouts';
+import { getDisplayColor, getDisplayName, isJobOrNameMatch } from './helpers';
+import type { Observable, OrbitData } from './types';
+
+export const Orbit = (props, context) => {
+  return (
+    <Window title="Orbit" width={400} height={550}>
```
stanalbatross, line 12: preserve old orbit menu's width and height (I changed it because we get so many players it was hard to use when small)

harryob, line 12:
```suggestion
    <Window title="Orbit" width={500} height={700}>
```

stanalbatross (approved):

PR #2260, opened 2023-01-08
Acid splash damage nerf

stanalbatross:
# About the pull request

nerfed acid blood splash damage back to 12

The change to acid blood, coupled with the introduction of other mechanics that make xenos stronger up close like tailstab, has led to a large reduction in the power of PBing - also leading to buck becoming underused and considered the weaker of the pair compared to slugs. 

PBing being harder encourages marines to sit in cades and discourages pushing, which contributes to stagnant rounds that nobody enjoys.

Therefore this PR is good for the game as it makes the comparatively weak buckshot more viable and encourages some more aggression in combat. Plus, now that xenos have tailstab, they can actively defend against PBers, rather than relying on passive acid blood damage to do it.

@ malders er marines learned how 2 dodge 

<details>

</details>

# Changelog

:cl:
balance: reduced acid blood splash damage to 12 from 25
code: cleaned up a bit of code to avoid an unneeded proc call in the code that handled acid blood splashes
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
@@ -237,7 +237,7 @@
 	var/list/tackle_counter
 	var/evolving = FALSE // Whether the xeno is in the process of evolving
 	/// The damage dealt by a xeno whenever they take damage near someone
-	var/acid_blood_damage = 25
+	var/acid_blood_damage = 12
 	var/nocrit = FALSE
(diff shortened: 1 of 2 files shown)


morrowwolf:
IMO make it 15. 25 is quite a bit right now. Also we may want to relook at synth resistances if this is going back down.

morrowwolf (approved):
Code good.

7+ day maintainer balance approval.

I think this may be a bit of an overnerf but we can take a look again in the future if it turns out to be so.

PR #2263, opened 2023-01-09
Goonify chemical retrieval

morrowwolf:
# About the pull request

Changes chemical retrieval to use goons and a new wey-yu research consultant. 

Goons have decent non-lethals and don't have huge guns that make people want to go to war instead of being normal human beings and not starting a lethal incident between the USCM and Wey-Yu. Going "loud" should be scary and having a dedicated Wey-Yu researcher come to make sure the chemicals were actually handled correctly makes sense in my brain.

<details>
I did not test this.
</details>

# Changelog

:cl: Morrow
balance: Changes chemical retrieval to use goons and a wey-yu research consultant rather than PMCs.
add: Wey-yu goon presets now have zip ties.
add: Added wey-yu research consultant equipment.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry_properties/prop_special.dm
+++ b/code/modules/reagents/chemistry_properties/prop_special.dm
@@ -96,7 +96,7 @@
 		H.contract_disease(new /datum/disease/xeno_transformation(0),1) //This is the real reason PMCs are being sent to retrieve it.
 
 /datum/chem_property/special/DNA_Disintegrating/trigger()
-	SSticker.mode.get_specific_call("Weyland-Yutani PMC (Chemical Investigation Squad)", TRUE, FALSE, holder.name)
+	SSticker.mode.get_specific_call("Weyland-Yutani Goon (Chemical Investigation Squad)", TRUE, FALSE, holder.name)
 	chemical_data.update_credits(10)
 	message_staff("The research department has discovered DNA_Disintegrating in [holder.name] adding 10 bonus tech points.")
 	var/datum/techtree/tree = GET_TREE(TREE_MARINE)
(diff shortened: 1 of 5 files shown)


code/modules/reagents/chemistry_properties/prop_special.dm
```diff
@@ -96,7 +96,7 @@
 		H.contract_disease(new /datum/disease/xeno_transformation(0),1) //This is the real reason PMCs are being sent to retrieve it.
 
 /datum/chem_property/special/DNA_Disintegrating/trigger()
-	SSticker.mode.get_specific_call("Weyland-Yutani PMC (Chemical Investigation Squad)", TRUE, FALSE, holder.name)
+	SSticker.mode.get_specific_call("Weyland-Yutani Goon (Chemical Investigation Squad)", TRUE, FALSE, holder.name)
```
harryob, line 99: i can’t believe this uses strings my god

harryob (approved):
code good + we should do this

code/modules/reagents/chemistry_properties/prop_special.dm
morrowwolf: Out of scope <joker emoji>

Segrain:
Have thoughts, will comment later. Remind me if I forget again.

stanalbatross (changes requested):
The PMCs are armed because the chemical turns people into xenos, and they have no idea who's ingested it (that's why they have so many flamers). This might make more sense if the chemical was something relatively harmless, but this one is genuinely dangerous,  and I don't think what is in effect just a bunch of goons would be sent to retrieve a chemical that turns people into the universe's most potent bioweapon. 

If you want to encourage chem ERTs to use nonlethals I suggest making their mod88s use rubber bullets, but going as far as to turn the entire ERT into budget MPs is too much when dealing with a potential threat of this calibre.

PR #2272, opened 2023-01-10
Added a destructible terrain support for admin event panel

ThePiachu:
# About the pull request

I've had fun a few times when the admins ran the destructible terrain event. Figured I'd make it so running the event would be really simple for anyone so we could have fun like this more often!

To use it simply go to the Event Panel and find it under Events. It converts all the ground-level walls (except the map boundaries) destructible.

It makes the admin job of setting up the event really simple.

<details>

</details>

# Changelog

:cl:
admin: Admins can now easily run the destructible terrain event using the Event Panel.
/:cl:

### DIFF
--- a/code/modules/admin/topic/topic_events.dm
+++ b/code/modules/admin/topic/topic_events.dm
@@ -41,7 +41,7 @@
 			if(tgui_alert(usr, "Are you sure you want to toggle all ground-level terrain destructible?", "Confirmation", list("Yes", "No"), 20 SECONDS) != "Yes")
 				return
 			toggle_destructible_terrain()
-			message_staff("[key_name_admin(usr)] toggled destrictible terrain.")
+			message_staff("[key_name_admin(usr)] toggled destructible terrain.")
 		if("blackout")
 			if(alert(usr, "Are you sure you want to do this?", "Confirmation", "Yes", "No") != "Yes")
 				return
--- a/code/modules/admin/topic/topic_events.dm
+++ b/code/modules/admin/topic/topic_events.dm
@@ -38,10 +38,10 @@
 				communications_blackout(1)
 			message_staff("[key_name_admin(usr)] triggered a communications blackout.")
 		if("destructible_terrain")
-			if(alert(usr, "Are you sure you want to make all ground-level terrain destructible?", "Confirmation", "Yes", "No") != "Yes")
+			if(tgui_alert(usr, "Are you sure you want to toggle all ground-level terrain destructible?", "Confirmation", list("Yes", "No"), 20 SECONDS) != "Yes")
 				return
-			enable_destructible_terrain()
-			message_staff("[key_name_admin(usr)] enabled destrictible terrain.")
+			toggle_destructible_terrain()
+			message_staff("[key_name_admin(usr)] toggled destrictible terrain.")
 		if("blackout")
 			if(alert(usr, "Are you sure you want to do this?", "Confirmation", "Yes", "No") != "Yes")
 				return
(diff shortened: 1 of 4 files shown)


code/modules/events/destructible_terrain.dm
```diff
@@ -0,0 +1,11 @@
+//Enables destructible terrain on the ground level
+/proc/enable_destructible_terrain()
+	var/list/turfs = get_all_of_type(/turf/closed/wall, TRUE)
+	for(var/turf/closed/wall/turf_for_editing in turfs)
+		if(turf_for_editing.z != 3)
```
stanalbatross, line 5: is_ground_level() is a more robust check for this, but could be a lot more intensive

code/modules/events/destructible_terrain.dm
```diff
@@ -0,0 +1,11 @@
...
+			//We only want to edit ground turfs!
+			continue
+		if(istype(turf_for_editing, /turf/closed/shuttle))
```
stanalbatross, line 8: an istype in this will be VERY intensive

stanalbatross (changes requested):

stanalbatross (changes requested):
also consider adding some way to revert this

code/modules/admin/topic/topic_events.dm
```diff
@@ -37,6 +37,11 @@
 			else
 				communications_blackout(1)
 			message_staff("[key_name_admin(usr)] triggered a communications blackout.")
+		if("destructible_terrain")
+			if(alert(usr, "Are you sure you want to make all ground-level terrain destructible?", "Confirmation", "Yes", "No") != "Yes")
```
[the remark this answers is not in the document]
ThePiachu: Fixed

code/modules/events/destructible_terrain.dm
ThePiachu: Fixed

code/modules/events/destructible_terrain.dm
ThePiachu: Would you rather me check the name of the turf or something else?

I moved the check to only check turfs that are `hull` so it should be a little bit less costly...

ThePiachu:
> also consider adding some way to revert this

Added a way to revert this by checking `initial` if something was a `hull` and then flipping its current state.

code/modules/admin/topic/topic_events.dm
```diff
@@ -37,6 +37,11 @@
 			else
 				communications_blackout(1)
 			message_staff("[key_name_admin(usr)] triggered a communications blackout.")
+		if("destructible_terrain")
+			if(tgui_alert(usr, "Are you sure you want to toggle all ground-level terrain destructible?", "Confirmation", list("Yes", "No"), 20 SECONDS) != "Yes")
+				return
+			toggle_destructible_terrain()
+			message_staff("[key_name_admin(usr)] toggled destrictible terrain.")
```
morrowwolf, line 44: Small typo

morrowwolf (changes requested):
Other than that one thing looks good to me.

code/modules/events/destructible_terrain.dm
```diff
@@ -0,0 +1,13 @@
+//Toggles the destructible terrain on the ground level
+/proc/toggle_destructible_terrain()
+	var/list/turfs = get_all_of_type(/turf/closed/wall, TRUE)
+	for(var/turf/closed/wall/turf_for_editing in turfs)
+		if(!is_ground_level(turf_for_editing.z))
+			//We only want to edit ground turfs!
+			continue
+		//If the initial wall was indestructible, we toggle indestructability
+		if(initial(turf_for_editing.hull))
+			if(istype(turf_for_editing, /turf/closed/shuttle))
```
morrowwolf, line 10: /turf/closed/shuttle won't be getting picked up here as it's not a /turf/closed/wall

morrowwolf (changes requested):
I was wrong one more thing

morrowwolf (approved):

PR #2275, opened 2023-01-10
Moves CO tablet locker back to the COs toilet

senorsapo:
im tired of XOs stealing it

🆑
tweak: remade the weapon status nanoui into tgui
/🆑


senorsapo:
im open to adding it to the CO vendor instead

stanalbatross:
Insufficient justification for the PR (fill out the template properly + consider asking the XO for it back)

Plus this unnecessarily handicaps XOs when there's no CO on by forcing them to hack into the COs room to get the tablet just like before

Closed

PR #2282, opened 2023-01-11
Phone Networks (5G?)

realforest2001:
# About the pull request
Adds Phone Networks, by default this won't have any impact whatsoever, but it allows for mappers to use phones on colonies without giving free access to call the almayer.

Adds a little bit of realism to the phones, they're a great communicator, but I think it odd that a colony phone would be able to call the almayer. It also opens the opportunity for phones that can call eachother, and another network, but the other network can't call them.

Tested locally, works as intended. Phones don't show up in the lists unless they share receiving network with the transmitting network.
<details>

</details>

# Changelog

:cl:
add: Adds networks to phones. Phone A can only call other phones who receive one of Phone A's transmit networks.
add: RTO packs set their category on initialize via a variable rather than hardcode, allowing for customisation in maps.
/:cl:

### DIFF
--- a/code/__DEFINES/equipment.dm
+++ b/code/__DEFINES/equipment.dm
@@ -546,3 +546,9 @@ var/global/list/uniform_categories = list(
 #define RADIO_FILTER_TYPE_INTERCOM 1
 #define RADIO_FILTER_TYPE_INTERCOM_AND_BOUNCER 2
 #define RADIO_FILTER_TYPE_ANTAG_RADIOS 3
+//=================================================
+
+//=================================================
+#define PHONE_RTO "RTO"
+#define PHONE_MARINE "Marine"
+#define PHONE_UPP_SOLDIER "Soldier"
--- a/code/modules/cm_phone/phone.dm
+++ b/code/modules/cm_phone/phone.dm
@@ -78,21 +78,21 @@ GLOBAL_LIST_EMPTY_TYPED(transmitters, /obj/structure/transmitter)
 /obj/structure/transmitter/proc/get_transmitters()
 	var/list/phone_list = list()
 
-	for(var/t in GLOB.transmitters)
-		var/obj/structure/transmitter/T = t
-		if(TRANSMITTER_UNAVAILABLE(T) || !T.callable) // Phone not available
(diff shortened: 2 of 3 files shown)


stanalbatross (changes requested):

code/__DEFINES/equipment.dm
```diff
@@ -546,3 +546,9 @@ var/global/list/uniform_categories = list(
 #define RADIO_FILTER_TYPE_INTERCOM 1
 #define RADIO_FILTER_TYPE_INTERCOM_AND_BOUNCER 2
 #define RADIO_FILTER_TYPE_ANTAG_RADIOS 3
+//=================================================
+
+//=================================================
+#define PHONE_RTO "RTO"
+#define PHONE_MARINE "Marine"
+#define PHONE_UPP_SOLDIER "Soldier"
```
stanalbatross, line 554: should be "UPP" or "UPP Soldier" surely?

code/modules/cm_phone/phone.dm
```diff
@@ -72,13 +75,15 @@ GLOBAL_LIST_EMPTY_TYPED(transmitters, /obj/structure/transmitter)
...
-/proc/get_transmitters()
+/obj/structure/transmitter/proc/get_transmitters()
 	var/list/phone_list = list()
 
 	for(var/t in GLOB.transmitters)
 		var/obj/structure/transmitter/T = t
 		if(TRANSMITTER_UNAVAILABLE(T) || !T.callable) // Phone not available
 			continue
+		if(!(T.network_receive in networks_transmit))
+			continue
```
stanalbatross, line 86: why does this use t and T both as different variables dear god the unreadability

stanalbatross (changes requested):

code/__DEFINES/equipment.dm
realforest2001: Unless it's edited by an admin, it would only show up on UPP phones. So same logic would be to prefix all other categories with USCM

code/modules/cm_phone/phone.dm
realforest2001: Glad I can say it wasn't me at least!

PR #2295, opened 2023-01-12
(Experimental) Squad Based Briefing Area

spartanbobby:
# About the pull request
This PR changes the open hall briefing room into 5 smaller briefing areas with one central command area with the idea being SLs will meet privately with the command staff before going to their squad's briefing area to give the briefing to their own men

The h(c)ope is that SLs may be able to build up more of a report with their men while the smaller briefing areas could possibly lead to more RP since the chat wouldn't be flooded as much my last cope is that it could lead to more inter-squad rivalry but it's a long shot

you'll notice it's alot of hope and cope I'd prefer if this was tested for two or three days to gauge if people would even bother with it and a possible vote by the community to see where their preference lies 

https://i.imgur.com/TCBXA8j.jpeg

# Changelog

:cl: SpartanBobby
mapadd: added maint shutters linked to the buttons in the MP briefing area
maptweak: Changes Almayer briefing room to be segregated by squad
maptweak: Fixes a missing piece of environmental storytelling that was removed with the recent maintbar PR
/:cl:

### DIFF
--- a/maps/map_files/USS_Almayer/USS_Almayer.dmm
+++ b/maps/map_files/USS_Almayer/USS_Almayer.dmm
@@ -30905,7 +30905,6 @@
 "caO" = (
 /turf/open/floor/almayer{
 	dir = 4;
-	icon_regular_floor = null;
 	icon_state = "cargo_arrow"
 	},
 /area/almayer/living/briefing)
@@ -37969,7 +37968,6 @@
 	},
 /turf/open/floor/almayer{
 	dir = 4;
-	icon_regular_floor = null;
 	icon_state = "cargo_arrow"
 	},
 /area/almayer/living/briefing)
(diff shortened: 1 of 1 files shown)


realforest2001 (changes requested):
image
Might want to rename these doors.

image
The shutters here can be bypassed via kitchen, I'd suggest moving them forward to line up with the hydroponics door.

image
Can't find anything linked to this button or it's opposite in the other checkpoint, assuming it was intended to be removed but overlooked.

image
Overall I would put these shutters on a separate button inside this room and put some on those doors, though you could make the echo specific shutters in this case start open, but that's just a suggestion.

realforest2001 (approved):

morrowwolf (changes requested):
First, this is really cool.

Sadly, this doesn't seem to be playing out the way we'd hope.

If this is going to go through we'll have to move things around in such a way that old briefing ability is still possible via opening shutters or somesuch.

PR #2306, opened 2023-01-13
playtime tgui re-up

stanalbatross:
# About the pull request

image

regression fixed in this, opened as the original pr staled out

fk nanoui

# Changelog

:cl: mullenpaul, stanalbatross
ui: converted the playtime menu to tgui
/:cl:


tgui/packages/tgui/interfaces/Playtime.tsx
```diff
@@ -0,0 +1,107 @@
...
+      </TableCell>
+    </>
+  );
+};
+
+const PlaytimeTable = (props: { data: PlaytimeRecord[] }, context) => {
+  return (
+    <Table>
+      {props.data
+        .slice(1)
```
harryob, line 49:
```suggestion
        .slice(props.data.length > 1 ? 1 : 0)
```

tgui/packages/tgui/styles/interfaces/Playtime.scss
```diff
@@ -0,0 +1,35 @@
...
+    padding: 0;
+    height: 34px;
+  }
+  table {
+    border: 1px solid white;
+    border-radius: 4px;
+  }
+
+  tr {
+    border-bottom: 1px solid white;
```
harryob, line 14:
```suggestion
  .Table__cell {
    border-bottom: 1px solid white !important;
  }
```
to override the theme

tgui/packages/tgui/styles/interfaces/Playtime.scss
```diff
@@ -0,0 +1,35 @@
...
+  }
```
harryob, line 15:
```suggestion
  .Table__cell {
    border-bottom: 1px solid white !important;
  }
```
the issue with using the hive_status theme is that hivestatus.scss contains .Table__cell setting the bottom border to 1px, which makes the bottom of the xeno ranks look pretty goofy. this overrides the theme for this interface. kinda hacky but what can you do

tgui/packages/tgui/interfaces/Playtime.tsx
harryob, line 49:
```suggestion
        .slice(props.data.length > 1 ? 1 : 0)
```
this allows for the observer data to show up

harryob (changes requested):

tgui/packages/tgui/styles/interfaces/Playtime.scss
```diff
@@ -0,0 +1,34 @@
...
+.PlaytimeInterface {
+  td {
+    padding: 0;
+    height: 34px;
+  }
+  table {
+    border: 1px solid white;
+    border-radius: 4px;
+  }
+  .Table__cell {
```
harryob, line 12:
```
    padding: 0px !important;
```
this is to stop the theme from overriding the cell padding to be massive and making xenos have a massive table in comparison to humans/observer. i forgot this 

(also we should just use a different theme, or move all the hivestatus thing to not be a theme but a stylesheet for that interface... but whatever)

harryob (changes requested):

PR #2307, opened 2023-01-13
removes haunted house event map from main repo

stanalbatross:
# About the pull request

removes the map and all refs to it 

it was a one-off event map that was used like twice, we didn't even use it this year's halloween, this is exactly what the map depot exists for 

as time goes on it will cause more and more issues (failing unit tests, failing ci, etc etc) and be generally more of a pain in the arse for everyone so I'm removing it before that all happens

event maps are not to be held in the main repo, they go in the map depot

plus this saves CI time + the map fails CI to start off with

# Changelog

:cl:
del: removed the unmaintained haunted house event map
/:cl:


harryob (approved):
gtfo

PR #2312, opened 2023-01-14
GAU consistency buff 

Tisx1:
# About the pull request

Currently the GAU is still in a somewhat sorry state compared to the other weapons CAS has, this is due to it's lack of consistency. In the video below, you will see just how pitiful the GAU can be esp when used in single fire mode. The reason behind this is that a burst from the GAU is unable to completely cover the AoE. (40 bullets to cover at 64 tile area (8x8)) The goal of this PR is to buff the GAU's firerate and ammo count to make it so that standing still in a GAU AoE is just as suicidal as standing in any other direct CAS AoE. 

I also modified the impact sound to account for more impacts.

Makes the GAU's AoE much more consistent and deadly to be in for long periods of time, which in turn should make the GAU worth it's point cost compared to other CAS weapons, and also make it worth the point cost to use during transport CAS. 

<details>

Before: https://cdn.discordapp.com/attachments/458150341742166017/1052649816179626035/GAU.mp4

After: https://youtu.be/uajQn6L5gu4

</details>

# Changelog

:cl: Tisx
balance: Buffed GAU Fire rate and ammo count to be more consistent. 
soundadd: Slightly modified GAU impact sound to compensate for increased impacts from GAU buff 
/:cl:


cuberound:
the area is 9x9

fira:
This is literally irrelevant to 90% of situations because people will clear the GAU hit area to begin with no matter how long it lasts.
This doesn't improve consistency as a result other than for overkilling whatever gets locked in it.
12 seconds of continuous fire especially if you have a FM of potentially up to 24 firing instances is unwarranted.

Hard pass for me.

PR #2319, opened 2023-01-15
Re-adds roundstart Tank, adjusts it to be an IFV rather than MBT.

LynxSolstice:
# About the pull request

This PR re-adds the tank in a heavily nerfed configuration, this is to bring it more in-line to beeing less an MBT and more a light tank, specifically an IFV, which, while an Infantry Fighting Vehicle, can be called a light tank. This PR removes the LTB cannon from being accessible, and moves many of the more high-tier weapons and armor into the spare parts section, as well as increasing the spare parts point prices considerably.

When Geeves TM'd the tank previously with access to only the autocannon, I personally thought it was interesting, and I thought it added a little bit of variety and better offensive (ie. Support by Fire) roles to the Vehicle Crewmen. Since Geeves is no longer developing for CMSS13, this idea of the tank being a more, IFV-like vehicle would have remained dead. My objective with this PR was to provide more variety to the VC role, and allow a new generation of tankers to learn and enjoy the vehicle as was possible in the older days of CM (which I myself never experienced.)

<details>
<summary>I had a few challenges with this code but as it stands now, the vehicle spawner is hardcoded with the selected_vehicle variable being connected to the APC as the way it checks to change the variable to the tank is using if(istype(V, /obj/effect/vehicle_spawner/tank)) which, if the inverse we done, I would need to make 3 separate checks for each APC variant as they are not sub-typed. The code works properly, when the tank is spawned the variable flips to be selected_vehicle = tank, which allows the tank parts to be seen, selected, and spawned.</summary>

https://i.imgur.com/JPSeuO4.png
https://i.imgur.com/EiXvEHv.png

</details>

# Changelog

:cl:LynxSolstice
add: Re-added the tank as a spawnable vehicle at roundstart.
del: Deleted the tank cannon from being spawnable.
balance: Moved many of the tank items to spare parts, increased the price of these objects in order to make them harder to access, will require a lot of points.
/:cl:

### DIFF
--- a/code/game/machinery/vending/vendor_types/crew/vehicle_crew.dm
+++ b/code/game/machinery/vending/vendor_types/crew/vehicle_crew.dm
@@ -50,8 +50,7 @@
 	SIGNAL_HANDLER
 	UnregisterSignal(SSdcs, COMSIG_GLOB_VEHICLE_ORDERED)
 
-	if(istype(V, /obj/vehicle/multitile/tank))
-		selected_vehicle = "TANK"
+	selected_vehicle = "TANK"
 
 	if(istype(V, /obj/vehicle/multitile/apc))
 		available_categories &= ~(VEHICLE_ARMOR_AVAILABLE|VEHICLE_INTEGRAL_AVAILABLE) //APC lacks these, so we need to remove these flags to be able to access spare parts section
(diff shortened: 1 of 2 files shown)


LynxSolstice:
https://i.imgur.com/EiXvEHv.png https://i.imgur.com/JPSeuO4.png

stanalbatross:
tank is not coming back to normal gameplay

PR #2320, opened 2023-01-15
Vehicles are removed. (No longer spawnable unless via admin variable editing for events)

LynxSolstice:
# About the pull request

Makes vehicles un-spawnable by normal means.

Vehicles have no place in the infantry centric gameplay loop, and were only added as an april fool's joke years ago. They have no hard counter, not in HvH and definitely not in HvX. they're a mobile 3x3 inaccessible area, that by design, blocks both Xenos and Marines from moving through those tiles. Many maps were restructured to allow vehicles to function within them and yet those maps still do not function properly with vehicles because CM and it's maps were designed around purely infantry vs Xenos/infantry gameplay. Years of tweaking and reworking CM has still not solved these fundamental issues in the design of the vehicles. they cannot as they currently stand, function properly during normal gameplay. Marines can go horizontal, vehicles cannot. CM would be better off if maps did not have to cater to under-utilized vehicles are simply harm the experience of both Xenos and Marines.

<details>
<summary> It works, the console acts as if a vehicle were already spawned and does not let you spawn a, as it sees it, second vehicle.
You can no longer set preferences for the VC role, nor late join as one. </summary>

</details>

# Changelog

:cl:LynxSolstice
del: Removed ability to access vehicles during normal gameplay. Removed the Vehicle Crewman role.
/:cl:


morrowwolf (changes requested):
The reasons presented for why this is good for the game are intensely inadequate for such a  removal.

Diegoflores31:
Undocumented Changes : Removed VC slots.

LynxSolstice:
> Undocumented Changes : Removed VC slots.

How would I document it properly? I thought the commit shows up?

Diegoflores31:
> > Undocumented Changes : Removed VC slots.
> 
> How would I document it properly? I thought the commit shows up?

Add it to the changelog

LynxSolstice:
oh woops, I didn't know that button on the change requested would remove the other people

morrowwolf (commented):
I think that without a dedicated maintainer to continue to work through issues with the APC it's time to put it to rest. It definitely stands out as a setting piece and is "movie-like" but it just doesn't fit well in the gameplay loop. Finding some way to make it not function as a 3x3 mobile wall is the most important concern for me if we leave it in.

harryob (approved):
yes we should do this i want to move on

stanalbatross:
I am content to approve this because vehicles remain stuck in the mud as hindrances to marines and massive 3 tile combat blockers on the front line (when they go there)

compensation for the loss of med / command apc's utilities can come in like a portable sensor tower or something if needed further down the line

PR #2335, opened 2023-01-17
Makes omnisentries act on battery life as well as makes them rotate

NewyearnewmeUwu:
# About the pull request
 
This PR nerfs omnisentries down to 500 rounds per sentry and makes them rotate. Each time they rotate, they have a delay, so that faster castes can outmaneuver it. Thanks to @carlarctg for da help with rotating.

It also adds a timed life for the sentries so that they are emphasized as _temporary_ field assets. Each shot reduces the battery life (10 minutes) by 3 seconds as of writing, and you can save battery by turning the sentry off. As they say, use it or lose it.

Omnisentries are a pain to fight for most castes because of the fact that they're usually hidden in 1x1 wired cade boxes. They also have an _absurd_ 1500 ammo pool, which makes no sense. Why are airdroppable sentries carrying 3x more than the default ones comtechs use? Plus, it looks really cool, and emphasizes their role as positional harassment vs instant death laser pillboxes that you need a T3 to deal with. The limited lifespan also incentivizes smart placement of them versus simply using them as the previously-stated instant death laser pillboxes, while providing a means of counterplay to xenos- each shot takes 5 seconds off the battery life.

<details>

</details>

# Changelog

:cl:
balance: omnisentries now rotate, have 500 rounds total, and have a 5 minute lifespan. Each shot reduces that lifespan by 5 seconds.
/:cl:

### DIFF
--- a/code/modules/defenses/sentry.dm
+++ b/code/modules/defenses/sentry.dm
@@ -21,7 +21,8 @@
 
 	var/omni_directional = FALSE
 	var/sentry_range = SENTRY_RANGE
-
+	var/has_lifespan = FALSE //Used for omnisentry lifespan.
+	var/lifespan = 0 //See above.
 	var/damage_mult = 1
 	var/accuracy_mult = 1
 	handheld_type = /obj/item/defenses/handheld/sentry
@@ -60,6 +61,10 @@
 
 	if(!target && targets.len)
 		target = pick(targets)
+	if(has_lifespan == TRUE)
+		src.lifespan = src.lifespan-1
+		if(src.lifespan <= 0)
+			handle_empty()
 
 	get_target(target)
 	return TRUE
@@ -107,6 +112,8 @@
 
 /obj/structure/machinery/defenses/sentry/get_examine_text(mob/user)
 	. = ..()
+	if(has_lifespan == TRUE)
+		. += SPAN_NOTICE("[src] has [lifespan] second\s remaining on its battery.")
 	if(ammo)
 		. += SPAN_NOTICE("[src] has [ammo.current_rounds]/[ammo.max_rounds] round\s loaded.")
 	else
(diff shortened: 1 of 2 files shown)


cuberound:
if you want it to be used in certain way, what about instead of making it worse overall in every role. honestly, omnisentry SUCKS as position harassment, when single T3 can just walk to them and destroy them, when they are not protected by cades, eaven when there are marines around. it is not gona be used as frontline harassment with 4 tile range and now with rotate super long cooldown that makes anyone able to outmanuver them.

code/modules/defenses/sentry.dm
```diff
@@ -107,6 +112,8 @@
 
 /obj/structure/machinery/defenses/sentry/get_examine_text(mob/user)
 	. = ..()
+	if(has_lifespan == TRUE)
```
mullenpaul, line 115: Same as above. This examine text can also be put into the subclasses get_examine_text

code/modules/defenses/sentry.dm
```diff
@@ -217,15 +224,19 @@
...
+		playsound(loc, 'sound/mecha/mechturn.ogg', 50, 1)
+		last_fired = world.time + 0.5 SECONDS
+		return
 	actual_fire(A)
 
 	if(targets.len)
 		addtimer(CALLBACK(src, PROC_REF(get_target)), fire_delay)
 
 /obj/structure/machinery/defenses/sentry/proc/actual_fire(var/atom/A)
+	if (has_lifespan == TRUE)
```
mullenpaul, line 238: Same as above. I would also put this into the subclass

mullenpaul (changes requested):
I've reviewed the code as presented but I'm not convinced that a 'battery life' is a good way of balancing this. If a battery charge is the path forward then this code should be reworked to use a battery, rather than a timer with an IC justification of an internal power store. It seems pretty gamey to put a battery life on this sentry when literally no other sentry does.

PR #2338, opened 2023-01-17
(DNM) One VC test proposal TM PR

PhantomEpicness:
# About the pull request
A test to see if the gunner is truly the problem with APC

Gunner is simply a holdout from the tank eras that has zero purpose on APC with crap weapons. It has only accomplished forcing drivers to play 3x3 frontline blocker and prevented players learning the role as they now fight with someone else. In addition, gunner is extremely boring in APC due to the weapon’s intended and fair design to be awful to prevent front lining. No such skill exists, nor fun, nor anything to do but repair and shoot. Which means that gunners are dying to get any action in their AFK role, thereby promoting soft-griefing.

This is an experiment to FORCE APC into a support only role and should not be merged

transport APC FPWs are untouched since they actually are widely accepted as garbage

<details>
compiles

</details>

# Changelog

:cl: totalepicness5
balance: VC slots have been reduced to one, APC weapons have been made unavalible.
/:cl:

### DIFF
--- a/code/game/jobs/job/command/auxiliary/tankcrew.dm
+++ b/code/game/jobs/job/command/auxiliary/tankcrew.dm
@@ -1,7 +1,7 @@
 /datum/job/command/tank_crew
 	title = JOB_CREWMAN
-	total_positions = 2
-	spawn_positions = 2
+	total_positions = 1
+	spawn_positions = 1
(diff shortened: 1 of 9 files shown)


morrowwolf:
Not needed. VCs are being removed.

PR #2344, opened 2023-01-18
Make the certificate actually spawn in on non-highpop rounds

JackieEstegado:
# About the pull request

Currently the power-loader certificate is only available in Req on high-pop (and also for 45 points in marine vendors, so people almost never take it). This means that it's currently useless for anything but marines doing battlemecha shenanigans with the power-loaders, since its not accessible exactly when it would be needed for a good reason (equipping the dropship when there are 0 POs or MTs on lowpop).

Will fulfill the role which I assume the certificate was supposed to fill, that is give an option for marines to spare themselves from 6+ minute round-trip transport on lowpop. While it may technically affect the balance a little, its just in terms of reducing an annoying thing skeleton crews have to deal with.

<details>

Shouldn't need any testing.

</details>

# Changelog

:cl:
balance: Requisitions will now always start with at least one powerloader certificate.
/:cl:

### DIFF
--- a/code/game/machinery/vending/vendor_types/requisitions.dm
+++ b/code/game/machinery/vending/vendor_types/requisitions.dm
@@ -133,7 +133,7 @@
 		list("Sentry Gun Network Laptop", 4, /obj/item/device/sentry_computer, VENDOR_ITEM_REGULAR),
 		list("JTAC Pamphlet", round(scale * 1), /obj/item/pamphlet/skill/jtac, VENDOR_ITEM_REGULAR),
 		list("Engineering Pamphlet", round(scale * 1), /obj/item/pamphlet/skill/engineer, VENDOR_ITEM_REGULAR),
-		list("Powerloader Certification", round(scale * 0.1), /obj/item/pamphlet/skill/powerloader, VENDOR_ITEM_REGULAR),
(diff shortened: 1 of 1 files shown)


code/game/machinery/vending/vendor_types/requisitions.dm
```diff
@@ -133,7 +133,7 @@
 		list("Sentry Gun Network Laptop", 4, /obj/item/device/sentry_computer, VENDOR_ITEM_REGULAR),
 		list("JTAC Pamphlet", round(scale * 1), /obj/item/pamphlet/skill/jtac, VENDOR_ITEM_REGULAR),
 		list("Engineering Pamphlet", round(scale * 1), /obj/item/pamphlet/skill/engineer, VENDOR_ITEM_REGULAR),
-		list("Powerloader Certification", round(scale * 0.1), /obj/item/pamphlet/skill/powerloader, VENDOR_ITEM_REGULAR),
+		list("Powerloader Certification", round(scale * 1), /obj/item/pamphlet/skill/powerloader, VENDOR_ITEM_REGULAR),
```
morrowwolf, line 136: Do (round(scale * 0.1) + 1) so as to guarantee at least 1 but not totally upset the amount we generally want in a higher pop round.

morrowwolf (changes requested):

PR #2348, opened 2023-01-19
Adds a biological trait in the form of photosensitivity.

NewyearnewmeUwu:
# About the pull request

This is my biggest PR to date and adds a fun, completely useless trait: Photosensitivity! Are you an edgy vampire wannabe or did you have too much to drink the night before you went back into hypersleep? Who knows, just avoid flashes!

Most 'tinted' glasses (sunglasses, etc) provide protection from the light. I also redid some of the code in the disabilities define code to make it less ugly by converting it to bitshifting.

Not enough people know about the fact that we _have_ biological traits, and one of the best ways to deal with that is to add to it. This is a fun, flavorful addition that adds literally nothing to your character positively. You are also stunned for 5 seconds longer than normal if you're flashed, but this is mitigated by just not getting flashed or wearing flash protection.

<details>

</details>

# Changelog

:cl:
add: Added the photosensitivity trait.
code: Reformatted the biological trait defines to use bitshifting.
/:cl:


code/modules/clothing/glasses/glasses.dm
```diff
@@ -8,6 +8,7 @@
 	/// The amount of nightvision these glasses have. This should be a number between 0 and 1.
 	var/lighting_alpha = LIGHTING_PLANE_ALPHA_VISIBLE
 	var/invisa_view = FALSE
+	var/eye_tint = FALSE //Used for photosensitivity.
```
carlarctg, line 11: variable name does not accurately represent what it's used in: eye_tint makes me think it tints the eyes like thermals.

```suggestion
	var/sensitivity_protection = FALSE //Used for photosensitivity.
```

If anything it feels like this variable is completely unnecessary and you should just give EYE_PROTECTION_FLAVOR to the glasses affected. If you don't want to do that, make a new tier in-between flavor called EYE_PROTECTION_WEAK_FLAVOR with a value of 0.5 or something.

code/modules/mob/living/carbon/human/life/handle_regular_hud_updates.dm
```diff
@@ -64,6 +64,8 @@
 			overlay_fullscreen("eye_blurry", /atom/movable/screen/fullscreen/impaired, 5)
 		else if((disabilities & NEARSIGHTED) && !HAS_TRAIT(src, TRAIT_NEARSIGHTED_EQUIPMENT))
 			overlay_fullscreen("eye_blurry", /atom/movable/screen/fullscreen/impaired, 2)
+		else if((disabilities & PHOTOSENSITIVITY) && !HAS_TRAIT(src, TRAIT_PHOTOSENSITIVITY_EQUIPMENT))
```
carlarctg, line 67: As stated above, this seems highly unnecessary and a misuse of traits when there's a roughly good enough alternative that is already in use.

PR #2420, opened 2023-01-19
vendors (not those) tgui

harryob:
dreamseeker_28VJgvlusU

:cl:
ui: vendors (for goods, not equipment) now use tgui
/:cl:


tgui/packages/tgui/interfaces/Vending.tsx
```diff
@@ -0,0 +1,320 @@
...
+  const { disabled, free, product, redPrice } = props;
+  const customPrice = access ? 'FREE' : '$' + product.price;
+  let standardPrice = product.price ? '$' + product.price : 'FREE';
+  if (free) {
+    standardPrice = 'FREE';
+  } else if (product.category === 'Premium') {
+    standardPrice = 'COIN';
+  }
+  return (
+    <Button
```
mullenpaul, line 277: Maybe include a vend icon to maintain consistency with the other vendors?

tgui/packages/tgui/interfaces/Vending.tsx
```diff
@@ -0,0 +1,320 @@
...
+        act('vend', {
+          'ref': product.ref,
+        })
+      }>
+      {standardPrice}
+    </Button>
+  );
+};
+
+const CATEGORY_COLORS = {
```
mullenpaul, line 290: This should be managed via scss not in the jsx file

tgui/packages/tgui/interfaces/common/ElectricalPanel.tsx
```diff
@@ -90,41 +90,44 @@ const ElectricalPanelClosed = (props: BoxProps, context) => {
 const WireControl = (props: { wire: WireSpec; index: number }, context) => {
   const { data, act } = useBackend<ElectricalData>(context);
   const target = props.index + 1;
+  let boxColor = 'green';
```
mullenpaul, line 93: You shouldn't embed styles in the component code

mullenpaul (changes requested):
Just gone over the code, not tried it in game yet, can do that later today. The style information should really be separated out from the component code, it makes theming a lot easier to work with. If the electrical panel is going to be updated then submit a screenshot of what it now looks like

tgui/packages/tgui/interfaces/common/ElectricalPanel.tsx
harryob: shouldn't've been changed in the first place!

PR #2423, opened 2023-01-19
perc_health

realforest2001:
# About the pull request
Makes health scan use a new var for percentage rather than a hard set calculation, stops mobs showing up as more than 100% healthy.

I need the maintainers to tell me if this will lag the server to death, but I don't think it will.

No more 150% healthy synths/xenos!

<details>

</details>

# Changelog

:cl:
code: Added a new var updated on life proc to centralise calculations for health percentage.
/:cl:

### DIFF
--- a/code/modules/mob/living/living_health_procs.dm
+++ b/code/modules/mob/living/living_health_procs.dm
@@ -1,6 +1,20 @@
 
 //procs directly related to mob health
+/mob/living/Life(delta_time)
+	var/new_perc_health
+	if(isXeno(src))
+		var/mob/living/carbon/Xenomorph/Xeno = src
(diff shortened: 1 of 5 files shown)


code/modules/mob/living/living_health_procs.dm
```diff
@@ -1,6 +1,20 @@
 
 //procs directly related to mob health
-
+/mob/living/Life(delta_time)
```
Zonespace27, line 3: couldn't this be done with signals instead of hooking *into the Life() of every single `living` mob*?

stanalbatross (changes requested):
this is only used in the scanner tgui so just make it do the calculation in the UI_data generating proc rather than *on /Life for every /living mob* (and the medhud calculations can also just take place there)

PR #2426, opened 2023-01-20
Fixed Doctor's Delight exploit.

carlarctg:
# About the pull request

Added a 90u OD limit to Doctor's Delight. Overdosing will cause 1 TOX damage per cycle.

Doctor's delight is a neat little chemical that's little-known and some chefs or researchers have taken to mass-producing due to the fact that it's basically tricordazine+. I do not think it should have its healing properties removed nor should it have a classically tiny overdose limit. The former because that is a heavy-handed punishment to emergent player gameplay that doesn't impact the round (And the drink is made from Tricordazine anyways so it makes sense), and the latter to reward players for producing the chemical, without causing an infinite-chem exploit like right now.

<details>

</details>

# Changelog

:cl:
fix: Added a 90u OD limit to Doctor's Delight. Overdosing will cause 1 TOX damage per cycle.
/:cl:


Segrain:
I am going to supersede this. We apparently once again cannot have nice things, but if you want something done right...

PR #2438, opened 2023-01-20
On_mob_life and drdelight.

Segrain:
# About the pull request

Another fine remnant of legacy code is put to rest. Dialing the numbers back and forth to keep everything in line was annoying and timeconsuming, but I can revisit all other discrepancies unearthed in the process at some later date. The scope, at least, has been fulfilled. I am mostly satisfied with how it ended balancing, but nothing is ever safe from abuse - and if I will have to do this again, so help me.

We cannot have nice things. Time and again and again, it is always the same. Somebody finds a neat piece of something older than our server, and instead of letting it be and enjoying it in peace, abuses it so blatantly that everybody else has to suffer.
It always depresses me to waste a day of life writing a code solution for a singular player that cannot behave, yet here we are, again.

# Changelog

:cl:
refactor: Drinks are brought to be more in line with reagent system overall. Let me know if something breaks.
fix: Exploit involving legacy healing drink going infinite is put to an end.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry_reagents/drink.dm
+++ b/code/modules/reagents/chemistry_reagents/drink.dm
@@ -9,28 +9,29 @@
 	id = "drink"
 	description = "Uh, some kind of drink."
 	reagent_state = LIQUID
-	nutriment_factor = 1 * REAGENTS_METABOLISM
+	nutriment_factor = 0.5 * REAGENTS_METABOLISM
+	// Drinks should be used up faster than other reagents.
+	custom_metabolism = FOOD_METABOLISM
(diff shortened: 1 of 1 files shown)


Diegoflores31:
Shouldn't it be the same as tricord ? 0.5 of each is like 0.25 of heal oer tick plus it wears out twice as fast AND has 30 OD?  Also has relaxing for some reason so it slows you down. 
In summary you go to the kitchen take a glass of 50units and proceed to die.

morrowwolf (approved):
Code LGTM

realforest2001:
Please close and un-testmerge #2425 upon merging this.

PR #2446, opened 2023-01-22
Migrate alamo, normandy and escape pods to TGShuttle framework

mullenpaul:
# About the pull request

Ports the alamo and normandy to tgshuttles.

Tweaked the dropship hijack crash to cause more explosions and fire.

Bring the dropships into 2023

<details>

</details>

# Changelog

:cl:
refactor: alamo and normandy now using tgshuttles
refactor: escape pods now using tgshuttles
refactor: when a pilot flies the dropship optimisation will automatically apply
refactor: temporarily forced autopilot delay to 30s
maptweak: all maps now use tgshuttles for landing zones
maptweak: all escape pods now tgshuttles
admin: moved admin shuttle to shuttle manipulator
/:cl:


mullenpaul:
Still to do

- enforce game rules to not allow dropships to leave before timer.
- enforce game rules to not allow queen to pull dropships down before dropship lands

spookydonut (approved):

mullenpaul:
Reported bug: CO can't select landing zone

mullenpaul:
alamo launch time needs to be increased

mullenpaul:
marines seemed to open the back door of the alamo in flight and decided to go for a walk

mullenpaul:
the dropship landing audio que plays after landing

mullenpaul:
Radios not being extended to the reservation layer, no radios available in flight

mullenpaul:
Last big issue needing resolved is around door locks on launch. It seems to shove people out of the door into space if they are in the door tile itself - not to the take off zone

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -0,0 +1,359 @@
...
+	hijack.fire()
+	if(almayer_orbital_cannon)
+		almayer_orbital_cannon.is_disabled = TRUE
+		addtimer(CALLBACK(almayer_orbital_cannon, .obj/structure/orbital_cannon/proc/enable), 10 MINUTES, TIMER_UNIQUE)
+
+	marine_announcement("Unscheduled dropship departure detected from operational area. Hijack likely. Shutting down autopilot.", "Dropship Alert", 'sound/AI/hijack.ogg')
+
+	var/mob/living/carbon/xenomorph/xeno = user
+	xeno_message(SPAN_XENOANNOUNCE("The Queen has commanded the metal bird to depart for the metal hive in the sky! Rejoice!"), 3, xeno.hivenumber)
+	xeno_message(SPAN_XENOANNOUNCE("The hive swells with power! You will now steadily gain pooled larva over time."), 2, xeno.hivenumber)
```
Segrain, line 223:
```suggestion
	xeno_message(SPAN_XENOANNOUNCE("The hive swells with power! You will now steadily gain pooled larva over time."), 2, xeno.hivenumber)
	xeno.hive.abandon_on_hijack()
```

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -0,0 +1,359 @@
...
+		to_chat(xeno, SPAN_WARNING("There are too many marines near the hive."))
+		return
+
+	var/obj/docking_port/mobile/shuttle = SSshuttle.getShuttle(shuttleId)
+	if(linked_lz)
+		playsound(loc, 'sound/machines/terminal_success.ogg', KEYBOARD_SOUND_VOLUME, 1)
+		if(shuttle.mode == SHUTTLE_IDLE && !is_ground_level(shuttle.z))
+			SSshuttle.moveShuttle(shuttleId, linked_lz, TRUE)
+		else
+			to_chat(xeno, "The screen reads T-[shuttle.timeLeft(10)].")
```
Segrain, line 156: This needs:
- An announcement of some sort when the dropship is successfully called down;
- A differentiation between "could not call down because it is not ready yet, wait [timeLeft] seconds" and "it is already in flight, have patience".

Segrain (changes requested):

code/modules/shuttle/computers/dropship_computer.dm
mullenpaul: Sounds good. Will work on this.

code/modules/shuttle/helpers.dm
```diff
@@ -0,0 +1,134 @@
...
+				INVOKE_ASYNC(src, PROC_REF(force_lock_open_door), door)
+				is_locked = FALSE
+
+/datum/door_controller/single/proc/lockdown_door_launch(obj/structure/machinery/door/airlock/air)
+	var/list/door_turfs = get_turf(air)
+	if(istype(air, /obj/structure/machinery/door/airlock/multi_tile))
+		var/obj/structure/machinery/door/airlock/multi_tile/multi_door = air
+		door_turfs = multi_door.get_filler_turfs()
+	for(var/turf/door_turf in door_turfs)
+		for(var/mob/blocking_mob in door_turf.loc) // Bump all mobs outta the way for outside airlocks of shuttles
```
Segrain, line 112:
```suggestion
		for(var/mob/blocking_mob in door_turf) // Bump all mobs outta the way for outside airlocks of shuttles
```
`Door_turf` is a turf. Its `.loc` is the entire area, so everybody gets kicked out.

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -0,0 +1,377 @@
...
+		hijack(xeno)
+		return
+
+/obj/structure/machinery/computer/shuttle/dropship/flight/proc/hijack(mob/user)
+
+	// select crash location
+	var/result = tgui_input_list(user, "Where to 'land'?", "Dropship Hijack", almayer_ship_sections)
+	if(!result)
+		return
+	var/obj/docking_port/mobile/marine_dropship/dropship = SSshuttle.getShuttle(shuttleId)
```
Segrain, line 221:
```suggestion
	var/obj/docking_port/mobile/marine_dropship/dropship = SSshuttle.getShuttle(shuttleId)
	if(dropship.is_hijacked)
		return
	var/result = tgui_input_list(user, "Where to 'land'?", "Dropship Hijack", almayer_ship_sections)
	if(!result)
		return
	if(dropship.is_hijacked)
		return
```

mullenpaul:
slapping DMN until the outstanding issues are resolved

Nanu308:
Mapping lgtm

mullenpaul:
It looks like when the marines spam click the doors - they can bypass the force lock and keep it open

Mark9013100:
> It looks like when the marines spam click the doors - they can bypass the force lock and keep it open

That's not new, has been a bug for awhile now.

Also,

image

We had the Alamo going to LZ two and Normandy going to LZ one in a recent round.

blackdragonTOW:
It seems that the consoles that control the dropships are also not picking up states as they should.  Hangar consoles can only move DS's between LZs but cant bring them back for example.

Also it seems that the consoles can be destroyed (we just had on OB blow them up)

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -0,0 +1,404 @@
+/obj/structure/machinery/computer/shuttle/dropship/flight
+	name = "dropship navigation computer"
+	desc = "flight computer for dropship"
+	icon = 'icons/obj/structures/machinery/shuttle-parts.dmi'
+	icon_state = "console"
+	req_access = list(ACCESS_MARINE_LEADER, ACCESS_MARINE_DROPSHIP, ACCESS_WY_CORPORATE_DS)
+	unacidable = TRUE
```
Segrain, line 7:
```suggestion
	unacidable = TRUE
	exproof = TRUE
```
Somebody put OB into FOB and destroyed the ground console. Not sure whether all of the type should be `exproof` (or even completely `indestructible`), or only some, or the entire `computer/shuttle` instead, but _some_ protection _somewhere_ is definitely in order.

Segrain (changes requested):

blackdragonTOW:
Small addendum: it seems there's a lot of confusion on how to move the DS from any given LZ to the ship via the CiC consoles, and Hangar consoles.  According to the officers who have tried you can only move a given DS between the two LZs.  

Also it seems we do not have an autopilot mode anymore either.

mullenpaul:
The remote control consoles issue has been resolved. The different remote control terminals were loading the compatible landing zones during their init, this resulted in inconsistent landing zone selections. The planetside ones only referenced the planetside lzs, the hangar ones referenced planetside and lz2. The dropships are spawned after map spawn and therefore, always saw all landing zones. This is now dynamically calculated during ui_static_data

mullenpaul:
https://gyazo.com/6e7c68d151ae64cd2bab87ed8a344f70 when a dropship launches, if it has its sentry turrets deployed, the turret is left behind

PR #2449, opened 2023-01-22
the xm88 cursor now applies when it should, and has a preference

harryob:
moaning does work people, just look. i also added a preference if anyone wanted to added more tgmc-ification cursors in the future for the boomers to opt out of

:cl:
fix: the xm-88 mouse cursor was appearing when it shouldn't (without the scope), and you can now opt out of seeing custom mouse cursors entirely
/:cl:


code/__DEFINES/dcs/signals/atom/signals_item.dm
```diff
@@ -21,3 +21,6 @@
 #define COMSIG_SNACK_EATEN "snack_eaten"
 
 #define COMSIG_ITEM_PICKUP "item_pickup"
+
+#define COMSIG_ITEM_ZOOM "item_zoom"
```
Zonespace27, line 25: please do me a favor and code doc these defines with the args they take in

code/modules/client/preferences.dm
```diff
@@ -225,6 +225,8 @@ var/const/MAX_SAVE_SLOTS = 10
 
 	var/current_menu = MENU_MARINE
 
+	var/custom_cursors = TRUE
```
Zonespace27, line 228: codedoc

Zonespace27 (changes requested):

code/modules/client/preferences.dm
```diff
@@ -1735,6 +1738,19 @@ var/const/MAX_SAVE_SLOTS = 10
...
+					if(!result)
+						return
+					if(result == "Yes")
+						custom_cursors = TRUE
+						to_chat(src, SPAN_NOTICE("You're now using custom cursors."))
+					else
+						custom_cursors = FALSE
+						to_chat(src, SPAN_NOTICE("You're no longer using custom cursors."))
+					save_preferences()
+
```
stanalbatross, line 1752: for maintainability reasons shouldn't this use a proc, since it's used in multiple places?

stanalbatross (changes requested):

PR #2472, opened 2023-01-25
Shotgun slug nerf

morrowwolf:
# About the pull request

This PR lowers the shotgun slug's superslow effect (from 2 to 1) and slow effect (from 5 to 3).

This PR also moves some code around to hopefully make targeted knockback adjustments easier (should probably be a signal iunno).

Slug slow times seem a bit much currently and allow for marines to feel a bit too safe 1v1ing and chasing lower tier xenos even on weeds.

A moderate nerf to the effect lengths should bring them more in line. A harder nerf may need to be looked at in the future if this doesn't solve the issue. An important note is to make sure not to overnerf slugs to the point it is no longer used. Heavy M44 rounds are a fairly direct competitor.

Lastly, an offset for the lost slow times may need to be considered in some form. Damage/pen seems like the obvious answer if a bit off tune for the ammo intent and may start to infringe on flechette's niche.

<details>
I spawned in a runner and shot it with a shotgun slug. The effect occurred.

</details>

# Changelog

:cl: Morrow
balance: Lowered the shotgun slug's superslow effect (from 2 to 1) and slow effect (from 5 to 3).
code: Moved around ammo knockback code to make specific ammo knockback effect changes easier.
/:cl:


code/modules/projectiles/ammo_datums.dm
```diff
@@ -133,39 +125,16 @@
...
-/datum/ammo/proc/heavy_knockback(mob/living/L, obj/item/projectile/P, var/max_range = 6) //crazier version of knockback
-	if(!L || L == P.firer)
-		return
-	if(P.distance_travelled > max_range || L.lying)
-		return
-	if(L.mob_size >= MOB_SIZE_BIG)
-		return
-
-	shake_camera(L, 3, 4)
+/datum/ammo/proc/knockback_effects(mob/living/L, obj/item/projectile/P)
```
Zonespace27, line 128: one letter vars bad

also a `///` doc comment would help

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1419,8 +1416,21 @@
...
+		to_chat(target, SPAN_XENODANGER("You are shaken and slowed by the sudden impact!"))
+		target.apply_effect(0.5, WEAKEN)
+		target.apply_effect(2, SUPERSLOW)
+		target.apply_effect(5, SLOW)
+	else
+		if(!isYautja(L)) //Not predators.
+			L.apply_effect(1, SUPERSLOW)
+			L.apply_effect(2, SLOW)
+			to_chat(L, SPAN_HIGHDANGER("The impact knocks you off-balance!"))
+		L.apply_stamina_damage(P.ammo.damage, P.def_zone, ARMOR_BULLET)
```
Zonespace27, line 1433:
```suggestion
		L.apply_stamina_damage(P.ammo.damage, P.def_zone, ARMOR_BULLET)
		
```

CABSHEP:
There should be added something to make slug not useless against T3's and drone branch T2's (besides burrower) if slug is to be nerfed.

Otherwise I see slug nerfs as pointless. They aren't even that good at 1vs1 anything, because you need another marine with a normal gun to actually finish off a xeno. 
To actually 1vs1 xeno, marine has to hit slugs without a miss one after another. If he misses even once, xeno will escape. That requires either skill, or brainless xeno who refuses to dodge and just runs in a straight line. Not to mention luck, where bullet scatter will fuck even the best aim.
On top of that slugs even push xenos away from you, helping xeno escape. 

For a runner (good luck hitting runner with a slug even once, he can use his pounce to escape) it takes 4 slugs. 7 slugs for lurker (good luck with that, pounce to escape offscreen), 8 slugs to kill a drone, sentinel, burrower, spitter and warrior. To kill a defender you literally have to perfectly hit 10 slugs, as much as you can fit in shotgun. 

That is all. Rest of the xenos are basically unaffected by slugs. Might as well buff flamethrowers because ravagers are too strong. The same logic.

Better solution would be to just lower slug damage against drones, sentinels, burrowers, spitters and warriors so it requires 10, or 11 slugs to actually kill them, thus 1vs1 potential is neutered without affectind utility. 
Nerf DMG, buff AP. Slug balanced.

stanalbatross:
I agree with this PR and think that slugs are a bit overtuned right now

extra damage / ap isn't needed and will just make them into a shotgun fired sniper bullet which isn't what we want

approved for now, further adjustments can be made as neccessary

harryob (approved):
code Should be fine 😄

PR #2100, opened 2022-12-25
Sentry gun UI

mullenpaul:
# About the pull request

This PR takes the idea of the sentry laptop from the aliens film and introduces the ability to configure defensive structures. The bulk of the work here is to introduce a framework to allow other contributors to build weird and wonderful ways to interact with turrets in new ways.

The sentry laptop is a control point which can be used by trained engineers to link sentry guns and other defensive structures for monitoring and configuration.

The laptop as well as several radio keys are available in req.

## effects on sentry guns

- sentry guns can have their IFF status configured, allowing for USCM, USCM+WY and all humanoid (everything non xeno)
- sentry guns can act like cameras (not connected to the camera network) and have a 'gunsight' camera, allowing an operator to see what is in their kill zone

## effects on multitool

The multitool is used to link the sentry laptop to a sentry gun. The multitool gets a reference to the laptop and when used on the sentry gun it links the two. A sentry gun can only be linked to one laptop. When a sentry gun is linked to a laptop it is considered 'encrypted' and can't be dismantled until unlinked - this encourages linked sentry guns to be used in a static role.

## effects on radio

This PR introduces a new radio channel which has its own encryption key. When the sentry gun is engaged it will announce to the radio channel (CD 60 seconds).

## setup instructions

- spawn sentry laptop
- place on table
- use multitool on laptop to load encryption keys
- deploy sentry gun
- use multitool on sentry gun
- access laptop

### On startup
image

### Configuration page
image

### Status page
image

### Status page - defensive structure
image

### Overview Page
image

### Sentry 'camera gunsight'
image

Gunsight adjusts to the hit area of the sentry gun, DMR has a bigger camera, flamer smaller.

### Viewing camera that doesn't exist
image

# Explain why it's good for the game

Allows for more interaction with sentry guns from the engineer. Builds a framework to allow for interesting interactions with the sentry without tgui changes.

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl: mullenpaul, monkeysfist101
add: added sentry laptop to game
add: new radio channel for sentry gun updates, only listenable on radio headsets
balance: allowed for the configuration of IFF groups of sentry guns
refactor: multitools now can store encryption keys
refactor: added UI to manage sentry guns
imageadd: added sentry laptop sprite
fix: should have removed phantom icons being rendered in mortar camera and security camera
/:cl:

PR #2492, opened 2023-01-27
Sentry Code Refactor

BKryth:
# About the pull request

Moves some of the code for the sentry laptop from #2100 to the item object.

Reusability, allows other item objects to have a special property of being slowly placed onto a table and having a variable dictate its new state. 

<details>

</details>

# Changelog

:cl:
refactor: moved the code for sentry laptop's table placement to a parent object so it can be used by other objects.
/:cl:


BKryth:
Mainly did this so that I can implement the same table placement effect onto a future PR I'm making

code/modules/defenses/sentry_computer.dm
```diff
@@ -99,36 +97,20 @@
...
-	else
-		to_chat(usr, SPAN_WARNING("You fail to setup the laptop"))
-
 /**
  * Called to reset the state of the laptop to closed and inactive.
  */
-/obj/item/device/sentry_computer/proc/teardown()
-	setup = FALSE
+/obj/item/device/sentry_computer/teardown()
+	..()
```
harryob, line 107:
```suggestion
	. = ..()
```

harryob (changes requested):
preserving the return type is a good habit (even if there isn't one)

code/game/objects/structures/tables_racks.dm
```diff
@@ -305,9 +305,8 @@
 			to_chat(user, SPAN_WARNING("You slice at the table, but only claw it up a little."))
 		return
 
-	if (istype(W, /obj/item/device/sentry_computer))
-		var/obj/item/device/sentry_computer/computer = W
-		computer.setup(src)
+	if (W.has_special_table_placement)
+		W.set_to_table(src)
```
stanalbatross, line 309: do this check before any istypes etc to save on proc overhead, easier/more generic checks should come first in big procs

stanalbatross (changes requested):
just this one thing the rest is fine

code/game/objects/structures/tables_racks.dm
BKryth: done, moved to start, just below if(!W)

PR #2508, opened 2023-01-28
updates check_grep to do list formatting and balloon alerts

harryob:
makes check_grep linter output prettier + adds list formatting and balloon alert errors to the linter


stanalbatross (commented):
if only this had been atomised so I could only look at alerts

harryob:
> if only this had been atomised so I could only look at alerts (if this doesnt add new alerts im dumb)

it doesn’t it just makes 1 alert lowercase for the grep

stanalbatross (approved):
looks good

PR #2510, opened 2023-01-28
Unique 7-Stage larva HUD update, xeno leader star sprite update

Ultimate-Fluff:
# About the pull request
Stages 0 and 1, and 4 and 5 of a larval infection shared the same icon states in their HUD and their palettes were not universal, so I created unique sprites for each larval stage. 

Before: 
larva2

After:
image

In addition, I smoothed out the bursting stage's animation so the larva is more vigorously chewing its way out of the chest.

Before: 
image

After:
image

Finally, the Xeno Leader star received a small update so it actually looks like a star. Left is before, right is after. It was also moved one pixel to the right so it would not overlap on the Prime icon state.
image

Some HUD sprites have been shifted a few pixels here and there so they properly align and/or do not overlay/underlay each other.

Before:
image

After:
image

These are just a few quality of life changes. Plus, the star actually looks like a star, now, so, that's a good thing, now!

# Changelog

:cl: Puckaboo2
fix: Stage 1 and stage 4 of a larval infection will now properly be distinguished from other stages in a Xenomorph's HUD.
tweak: All larva stages in the Xenomorph HUDs now each share a unique sprite and have a uniform color palette. 
tweak: The bursting larva stage has been modified to make the larva appear more vigorous to escape its fleshy prison.
tweak: The Xenomorph Leader star has been polished to add shading.
fix: Some of the xeno HUD icon states have been shifted around in a manner which ensure none of them overlay or underlay any other HUD icon states.
/:cl:


kugamo:
You added an additional frame of animation to the queen eye by mistake it seems. Please fix.

kugamo (changes requested):
See above.

Ultimate-Fluff:
> See above.

Fixed.

kugamo (changes requested):
Atomize your PRs, you signifigantly changed the nature of the PR days after opening it by adding additional HUD changes. Which complicates the reviewing process.

In terms of the original jib of this PR, just the embryo HUD changes, everything but the animation is fine. Which will either need more work or removal for approval, but first..
A T O M I Z E.

PR #2516, opened 2023-01-29
changes the text bar + statusbar

harryob:
more screen real estate for fragging

statusbar visible (hovering over a non-turf)
dreamseeker_M3bflkrize

statusbar invisible
dreamseeker_5jT64pEFth

:cl:
ui: changes the main interface a bit
/:cl:

don't worry if everyone hates it i won't bother

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -324,6 +324,9 @@
 	apply_clickcatcher()
 	mob.reload_fullscreens()
 
+	if(prefs.auto_fit_viewport)
+		INVOKE_ASYNC(src, .verb/fit_viewport)
+
 /client/proc/create_clickcatcher()
 	if(!void)
 		void = new()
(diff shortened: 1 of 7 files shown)


brjoaoed:
Here is some feedback, the little bit more of screen useful for some people but, at least for me, it was a huge nuisance, i'm used to play using SS13 standard non-hotkey mode (the arrows key) and due to it, i can easily walk and write at the same time, multitasking communications while also not being an easy target standing still to write, however, with this update, it moves my attention to the right part of the screen, taking out my focus on combat to make sure i haven't made any spelling or grammar errors i keep looking away, my recommendation if this update were to be applied is having a toggle option within the chat preferences for those who like the vanilla one.

harryob:
> However, the thing I find immensely frustrating with this PR is the statusbar. It goes from invisible to constantly popping up and flickering, which draws the eye away from combat for something that isn't really important. This was never an issue before, as there was always SOMETHING in that corner, but going from nothing to something and having it constantly flicker just barely at the edge of your vision makes it so incredibly distracting that it's bad enough for me to be against the merging of this PR.

that was the point i was mostly curious about, it's very easy to make that bar static, or have a preference for the behaviour

stanalbatross (approved):
all looks fine

brjoaoed:
Is it possible to add in a button to have the option to disable it on the preferences,**please** ? Using this chat style is too much problematic due to my style of gameplay with the arrow keys + using the chat while walking, for sure i'm not the only one who uses SS13 standard arrow keys to keep the rest of the keyboard free to use the chat while they walk, and having it moved to the right, away from the bottom part of the screen makes this update a living nightmare.

PR #2456, opened 2023-01-23
Revives Geeves' PR for Uncloaked Action GUI elements.

LynxSolstice:
# About the pull request

This PR is a direct revival of https://github.com/cmss13-devs/cmss13/pull/1842 , it fixes the issue https://github.com/cmss13-devs/cmss13/issues/1665 .

# Explain why it's good for the game

Quoted from #1842, "[...] it was a bit silly that you could track things down not by seeing them, but by seeing the game UI floating above their heads. This should make it a little more engaging to track down the bad guy while giving them a chance to Metal Gear their way out of trouble."

# Testing Photographs and Procedure

<details>
<summary>It runs and cloaked players no longer have interaction circles above their heads. </summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl: LynxSolstice, Geeves
balance: interaction circles that appear of above your head while cloaked are now harder to see (does not include circles above interacted with objects (ie. nests, C4s, breaching charges, xenos eating people, whatever predators have). That is to say, interaction circles inherit the alpha of cloaked users.
/:cl:

PR #2541, opened 2023-02-01
Fix for do_after() and alpha check.

Segrain:
# About the pull request

Followup to #2456. In as much as reverting something and redoing it from scratch can be called a followup.
I was going to go on entire _rant_ about how when you copypaste somebody's code to retry making it, you should watch carefully whence and whither you are copypasting (that `break` belonged in its own `if` and not in the one inserted into the middle of it), how you should test that your version actually works, how `TRUE` and `FALSE` are different things and one thing is not another (this one should be obvious, but...). Then I switched to going to go on a rant about some extreme edge cases and how terrible Byond is at handling something about icons and alphas. Then I went to actually read documentation and realised that we have been trying to reinvent a bicycle and that it was not even going to liftoff anyway.
When all you had to do was revert the (relevant part of) [this](https://github.com/cmss13-devs/cmss13/commit/d523bb75462cc9367cc55ee8ba3f07e16c9d3acf).

Everything still broke from #2456, though. This has made a lot of people very angry and been widely regarded as a bad move.

...and I still reserve the right to say this: _actually test that your code actually works_.

# Changelog

:cl:
fix: Timed actions are once again actually correctly timed.
/:cl:

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1054,22 +1054,15 @@ var/global/image/action_purple_power_up
 		target_is_mob = TRUE
 
 	var/image/busy_icon
-	var/has_busy_icon = FALSE
 	if(show_busy_icon)
 		busy_icon = get_busy_icon(show_busy_icon)
 		if(busy_icon)
-			busy_icon.appearance_flags = RESET_ALPHA|KEEP_APART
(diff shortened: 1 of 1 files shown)


harryob (approved):
sigh

morrowwolf (approved):

PR #2543, opened 2023-02-02
escape menu

harryob:
adds some of the buttons from the verb lists into an escape menu for easy access

thanks mothblocks!

CleanShot 2023-02-02 at 17 18 07@2x

CleanShot 2023-02-02 at 17 17 41@2x

:cl:
ui: adds an escape menu, with some (not many) buttons on it
/:cl:


stanalbatross (approved):
this all looks fine, I'll give you the time to add a bit more to it

also ui sprites dont need approval

stanalbatross (approved):
what could POSSIBLY go wrong now

PR #2553, opened 2023-02-04
The ASRS will no longer steal your limbs

### LINKED ISSUE #2552: Black market ASRS pit removing limbs that it shouldn't remove
### Testmerges

#2482

### Description of the bug

The ASRS pit seems to be able to remove body parts that it probably shouldn't be able to remove. Like a mobs groin in the case that I saw, but that means it can probably remove someone's torso also.

While I get that the hacked ASRS pit is supposed to do a bunch of damage to people that get caught in it, it probably shouldn't be doing so much damage that its nuking people's groin off. I heard someone saying that it might be because the pit is just doing too much damage to the selected limb so that is probably a good place to start looking?

1. Hack the ASRS console to enable the Black Market
2. Place a couple of CTs on the pad and send it down
3. If the CTs get unlucky one of them will come back up with some limbs missing, and a groin can be one of those missing limbs.

### Issue Bingo

BeagleGaming1:
# About the pull request

Removes the 4% chance for the ASRS to remove limbs, as it can still delimb from the damage dealt.

The delimb chance could remove any limb, and as a result could remove the groin, torso, or head. While the head and torso would instantly perma-kill you, losing your groin causes you to lose your legs. As you are not meant to lose these limbs there is no way to replace them.

fixes #2552 
# Changelog

:cl:
fix: The ASRS can no longer steal your groin.
/:cl:

### DIFF
--- a/code/game/supplyshuttle.dm
+++ b/code/game/supplyshuttle.dm
@@ -560,10 +560,6 @@ var/datum/controller/supply/supply_controller = new()
 	shake_camera(mauled_human, 20, 1)
 	mauled_human.apply_armoured_damage(60, ARMOR_MELEE, BRUTE, rand_zone())
 
-	if(prob(4))
-		var/obj/limb/dropped_limb = pick(mauled_human.limbs)
-		dropped_limb.droplimb(FALSE, FALSE, "machinery")
-
 //Buyin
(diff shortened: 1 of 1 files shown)


carlarctg:
1984

harryob (approved):

PR #2573, opened 2023-02-05
Fixes the morgue having no exit stun.

### LINKED ISSUE #1735: Exiting a morgue has no delay timer.
## Description

People have used morgue to ambush xenos with dual shotguns and what have you. It feels exploitative, especially in comparison to possibly bursting out of a locker and that having a delay. Apparently it can be so bad are asking for a rule clarification if it's an exploit. So best patch that out.

There should be some delay to getting up from the morgue when you're alive, or maybe you should be put prone when you enter it so you'd have some kind of delay before you act.

## Steps to reproduce

1) Get into a morgue with a shotgun in hand
2) Emerge
3) Fire at xeno with no delay

NewyearnewmeUwu:
# About the pull request
This fixes  #1735 by adding code to the relaymove of morgues (the function that's called when someone inside something moves) to allow for an exit stun. I tested it and it works.

People were apparently exploiting the fact that it had no delay whatsoever, and this fixes that. Exploits bad grrr.

<details>

</details>

# Changelog

:cl:
fix: fixed exploits surrounding morgues having no exit delay
/:cl:

### DIFF
--- a/code/game/objects/structures/morgue.dm
+++ b/code/game/objects/structures/morgue.dm
@@ -114,7 +114,7 @@
 /obj/structure/morgue/relaymove(mob/user)
 	if(user.is_mob_incapacitated())
 		return
-	if(!(exit_stun == 0))
+	if(exit_stun)
 		user.stunned = max(user.stunned, exit_stun) //Action delay when going out of a closet (or morgue in this case)
 		user.update_canmove() //Force the delay to go in action immediately
 		if(!user.lying)
--- a/code/game/objects/structures/morgue.dm
+++ b/code/game/objects/structures/morgue.dm
@@ -114,9 +114,8 @@
 /obj/structure/morgue/relaymove(mob/user)
 	if(user.is_mob_incapacitated())
 		return
-	else
-		if(!(exit_stun == 0))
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/morgue.dm
```diff
@@ -111,9 +112,16 @@
 		. = ..()
 
 /obj/structure/morgue/relaymove(mob/user)
-	if(user.is_mob_incapacitated(TRUE))
+	if(user.is_mob_incapacitated())
 		return
-	toggle_morgue(user)
+	else
```
harryob, line 117: this else is redundant as you're returning if the if statement is true

harryob (changes requested):

code/game/objects/structures/morgue.dm
```diff
@@ -111,9 +112,15 @@
 		. = ..()
 
 /obj/structure/morgue/relaymove(mob/user)
-	if(user.is_mob_incapacitated(TRUE))
+	if(user.is_mob_incapacitated())
 		return
-	toggle_morgue(user)
+	if(!(exit_stun == 0))
```
harryob, line 117:
```suggestion
	if(exit_stun)
```

harryob (changes requested):

PR #2574, opened 2023-02-05
Pounces now properly toggle the hiding ability icon.

### LINKED ISSUE #1784: Pouncing as runner doesn't toggle hide icon state
### Testmerges

N/A

### Description of the bug

Hiding and then pouncing as a runner does not remove the outline from the hide ability.

When pouncing while hidden, the outline around the hide ability should be removed to accurately reflect that you are no longer hidden.

1. Be runner.
2. Hide.
3. Pounce.
4. Notice Hide still has the "activated" outline around it.

### Issue Bingo

NewyearnewmeUwu:
# About the pull request
This fixes #1784. It's a bit of a messy solution but it works. I transferred most of the hide use_ability code to the new update_button_icon() code I made for the ability and now pouncing properly toggles whether you're hiding or not. I also cleaned up the code in the hide and pounce code.

Bugs bad, this was kind of a tricky fix since you can't _really_ change another button from within a different action but I made it work.

<details>

</details>

# Changelog

:cl:
fix: pouncing now properly toggles the hiding ability button
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
@@ -362,83 +362,82 @@
 		for(var/mob/living/carbon/xenomorph/L in hive.xeno_leader_list)
 			L.handle_xeno_leader_pheromones()
 
-/datum/action/xeno_action/activable/pounce/use_ability(atom/A)
-	var/mob/living/carbon/xenomorph/X = owner
+/datum/action/xeno_action/activable/pounce/use_ability(atom/target)
+	var/mob/living/carbon/xenomorph/xeno = owner
 
 	if(!action_cooldown_check())
 		return
 
-	if(!A)
+	if(!target)
 		return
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
```diff
@@ -362,83 +362,82 @@
...
+	if(xeno.legcuffed)
+		to_chat(xeno, SPAN_XENODANGER("You can't [ability_name] with that thing on your leg!"))
 		return
 
 	if(!check_and_use_plasma_owner())
 		return
 
-	if(X.layer == XENO_HIDING_LAYER) //Xeno is currently hiding, unhide him
-		X.layer = MOB_LAYER
-		X.update_wounds()
```
harryob, line 393: what happened here btw

NewyearnewmeUwu: Basically, that adjusts its layer and updates it, but I moved that function to the update_button_icon proc I made to make it more self-contained.

PR #2596, opened 2023-02-08
removes tag vars from maps, other grep additions

harryob:
there's a check grep for it so you know we had to do it

they were literally useless

no cl no player facing changes hopefully god bless

also:
- adds a grep for to_chat sanity
- adds a grep for timer flag sanity
- changes EOL grep to pcre


harryob:
solidified (deatomized) into update paths PR

PR #2597, opened 2023-02-08
ports maplinters

harryob:
modern testing suite that isn't reliant on greps (and actually parses .dmm files) is much improved


morrowwolf (approved):
Mapping looks good, code is mostly just ported tools and TM has gone well.

PR #2599, opened 2023-02-08
adds updatepaths utility, uses it to remove all tag icon states + make a subtype

harryob:
there's a check grep for it so you know we had to do it

they were literally useless

this uses the new (to cm) updatepaths utility which parses and changes mapfiles dynamically, which is MUCH BETTER than regexing maps! yay!

no cl no player facing changes hopefully god bless


Nanu308:
Solve conflicts, and was this done in Regex? If so, check for map corruption

Nanu308:
Mapping seems fine

morrowwolf (approved):

PR #2608, opened 2023-02-09
adds a keybinding for showing held item

harryob:
qol also this one was egregiously hidden in the object menu with other hits like pull and pick up

:cl:
qol: you can now keybind showing held item
/:cl:

### DIFF
--- a/code/datums/keybinding/human.dm
+++ b/code/datums/keybinding/human.dm
@@ -192,7 +192,6 @@
 		shown_item.showoff(human_user)
 	return TRUE
 
-
 #undef QUICK_EQUIP_PRIMARY
 #undef QUICK_EQUIP_SECONDARY
 #undef QUICK_EQUIP_TERTIARY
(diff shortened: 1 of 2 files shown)


code/datums/keybinding/human.dm
```diff
@@ -174,6 +174,24 @@
...
 	return TRUE
 
+/datum/keybinding/human/show_held_item
+	hotkey_keys = list()
+	classic_keys = list()
+	name = "show_held_item"
+	full_name = "Show Held Item"
+	keybind_signal = COMSIG_KB_HUMAN_SHOW_HELD_ITEM
+
+/datum/keybinding/human/pick_up/down(client/user)
```
fira, line 184: Harrryyyyyyyyyyy... You left the wrong path when copypasting

fira (changes requested):

code/datums/keybinding/human.dm
harryob: i still think it should compile error

fira: well multiple overrides is valid DM and we got a few

fira (approved):

stanalbatross (approved):

PR #2616, opened 2023-02-10
Unique 7-stage larva HUD update

Ultimate-Fluff:
# About the pull request
Stages 0 and 1, and 4 and 5 of a larval infection shared the same icon states in their HUD and their palettes were not universal, so I created unique sprites for each larval stage. 

Before: 
larva2

After:
image

Finally, the Xeno Leader star received a small update so it actually looks like a star. Left is before, right is after. It was also moved one pixel to the right so it would not overlap on the Prime icon state.
image

These are just a few quality of life changes. Plus, the star actually looks like a star, now, so, that's a good thing, now!

# Changelog

:cl:
fix: Stage 1 and stage 4 of a larval infection will now properly be distinguished from other stages in a Xenomorph's HUD.
tweak: All larva stages in the Xenomorph HUDs now each share a unique sprite and have a uniform color palette.
tweak: The Xenomorph Leader star has been reshaped into a proper star with added shading and moved a pixel to the right to not overlay any other HUD icon states.
/:cl:


kugamo (approved):
Sprites Approved.

harryob (approved):
no code

PR #2621, opened 2023-02-10
Corrected overlaying HUD icon states

Ultimate-Fluff:
# About the pull request
Some HUD icon states for xenos overlay over each other because they are too close in proximity to each other. The most notable examples are the pheromone borders and the HUD star overlaying the Prime xenomorph icon state. I correctly spaced them out so none of the pixels of any HUD overlay each other.

Before
image

After
image

If it was not obvious that I have OCD, now it is.

</details>

# Changelog

:cl: Puckaboo2
fix: Xenomorph HUD sprites are now properly spaced out so borders no longer overlay one another.
/:cl:


Agameofscones (changes requested):
these should be overlays that are combined into one image, but c'est la vie.

Animation is very jerky, could use one or two smear frames. Please edit it.

Will approval when that is done

Ultimate-Fluff:
> these should be overlays that are combined into one image, but c'est la vie.
> 
> Animation is very jerky, could use one or two smear frames. Please edit it.
> 
> Will approval when that is done

I reverted the larva wiggle sprite back to its original state. I shouldn't have messed with it in the first place.

harryob (approved):
nocode

PR #2642, opened 2023-02-13
Debug TM - Polio Stage 1 - Identifying offending mobs

fira:
# About the pull request

Allows us to debug what mob triggers polio.

This is important because it might hint at cause. It's also likely it's a dead/dying mob.


harryob (approved):

harryob:
the great menace has gone for now but i'm sure it'll return in the future. until then this will remain, in wait

PR #2652, opened 2023-02-14
Say\Hearing performance 515 edition

Git-Nivrak:
# About the pull request
Improves the logic for mobs and items hearing what other people are saying to remove needless lag that we all know and hate!
Thing this breaks (on purpose):

~~Hearing in cryopods (needless)~~
~~Radios and tape recorders only work in line of sight (actual line of sight no byond one) unless someone is standing on the same tile \ carrying the radio (worthy sacrifice for performance)~~
assemblies can no longer be voice activated (pretty sure they're removed anyways)
~~items that use hearing no longer work in storages (When was that even useful anyways?)~~
Items that are in deep storage (Storage inside a storage or deeper) won't hear stuff (I doubt they ever did actually)
~~removed option for ghosts to hear the entire map (The fact that this is a default preference and even exists bothers me)~~

less lag = more fun?

<details>

</details>

# Changelog

:cl:
performance: Improved the logic of talking \ hearing to be more performant.
/:cl:


code/game/objects/items/devices/radio/radio.dm
```diff
@@ -75,7 +76,7 @@
 	for (var/ch_name in channels)
 		secure_radio_connections[ch_name] = SSradio.add_object(src, radiochannels[ch_name], RADIO_CHAT)
 
-	flags_atom |= USES_HEARING
+	LAZYADD(GLOB.special_hearing_objects, src)
```
fira, line 79: radio adds to special but removes from normal list

code/game/objects/items/devices/taperecorder.dm
```diff
@@ -227,6 +229,7 @@
 	time_warned = FALSE
 	update_icon()
 	update_sound()
+	LAZYREMOVE(GLOB.hearing_objects, src)
```
fira, line 232: same problem here, adds to normal removes from special

code/modules/clothing/suits/storage.dm
```diff
@@ -8,7 +8,9 @@
 	pockets.storage_slots = storage_slots
 	pockets.max_w_class = SIZE_SMALL //fit only small items
 	pockets.max_storage_space = 4
-	flags_atom |= USES_HEARING
+
+/obj/item/clothing/suit/storage/Destroy()
+	..()
```
fira, line 13: this does nothing

code/modules/mob/living/say.dm
```diff
@@ -97,31 +97,13 @@ var/list/department_radio_keys = list(
...
-						listening_obj |= O
-
-			for(var/mob/M as anything in GLOB.player_list)
-				if((M.stat == DEAD || isobserver(M)) && M.client && M.client.prefs && (M.client.prefs.toggles_chat & CHAT_GHOSTEARS))
-					listening |= M
-					continue
-				if(M.loc && (M.locs[1] in hearturfs))
-					listening |= M
+
+			for(var/mob/M in hearers(message_range, T))
```
fira, line 104: you can use `as anything` here, hearers() are always mobs

code/modules/mob/living/say.dm
```diff
@@ -135,10 +117,14 @@ var/list/department_radio_keys = list(
 
 		addtimer(CALLBACK(src, PROC_REF(remove_speech_bubble), speech_bubble), 3 SECONDS)
 
-		for(var/obj/O as anything in listening_obj)
-			if(O) //It's possible that it could be deleted in the meantime.
+		for(var/obj/O as anything in GLOB.hearing_objects)
+			if(O && (O in hearturfs)) //It's possible that it could be deleted in the meantime.
```
fira, line 121: not sure if intended - this won't pick up storage depth, eg an object in a bag

fira (changes requested):
for future ref, you don't need to use LAZY procs here since the lists are init'd, it's no big deal

Destroy() implementations should passthrough hint, this can be used in debugging, so you should use `. = ..()`

Herotitan772:
Not sure why you have an issue with ghosts being able to hear everything, i only use that. its good to find interesting flashpoints and to figure out the big picture

Nanu308:
This is good to be merged. Tested extensively.

MercadierY2K:
Same concerns as Segrain, with emphasis on the third point. The ability to quickly jump to and from conversations is hugely helpful for staff work, especially for spotting grief in real time.

Even aside from staff work, the ability to see and jump to interesting conversations is basically the only way to know what's going on in the round as a spectator. Whatever performance benefit this change gives - which likely won't be noticeable to players anyway - feels totally offset by how clunky it makes observing.

morrowwolf (changes requested):
I've seen myself hearing vendors and actions through walls as well as administrative issues with not being able to hear things required. Secondarily, it should be easy to record with tape recorders and that should include while equipped in some form. Unless you can prove some serious improvements to performance based from this PR then these issues need to be addressed before merging.

Watermelon914:
What's your source for `hearers` being more performant than the other built-in byond functions?

Git-Nivrak:
> What's your source for `hearers` being more performant than the other built-in byond functions?

When you do for (var/mob in hearers()) it will only calculate whatever function it has for mobs in range while if you do with what we had previously  is get all the atoms in range after setting luminosity to 6 so you can hear in the dark and then you'd loop over every atom which is a huge waste of time

Viewers() (might) work together with setting the luminosity as it only returns mobs but that seems like a worse way to do it

The point is hearers is not what improves the performance but rather the fact that you don't loop over every atom

Watermelon914:
> > What's your source for `hearers` being more performant than the other built-in byond functions?
> 
> When you do for (var/mob in hearers()) it will only calculate whatever function it has for mobs in range while if you do with what we had previously is get all the atoms in range after setting luminosity to 6 so you can hear in the dark and then you'd loop over every atom which is a huge waste of time
> 
> Viewers() (might) work together with setting the luminosity as it only returns mobs but that seems like a worse way to do it
> 
> The point is hearers is not what improves the performance but rather the fact that you don't loop over every atom

Can you change the description to not include 
`P.S It's called 515 edition because you can now do for(var/mob/M in hearers(range, T)) and hearers will only calculate the mobs rather than everything with its retarded n^3 "raytracing" algorithm`
Hearers is still doing a view() check. The expensive part of view() is doing it in the first place, not iterating through the results.

Git-Nivrak:
Yea sure if it bothers you that much

Git-Nivrak:
But note that it not only makes iterating faster it doesn't do the view check for things other than mobs
See https://www.byond.com/forum/post/2736777

Watermelon914:
> But note that it not only makes iterating faster it doesn't do the view check for things other than mobs See https://www.byond.com/forum/post/2736777

that's a micro-optimization rather than some significant performance boost.

code/modules/mob/living/say.dm
```diff
@@ -106,45 +106,42 @@ var/list/department_radio_keys = list(
...
+		for(var/obj/O as anything in GLOB.hearing_objects)
+			if(O && ((O in hearturfs) || (O.locs[1] in hearturfs))) //It's possible that it could be deleted in the meantime.
 				O.hear_talk(src, message, verb, speaking, italics)
 
+
+		for(var/obj/O as anything in GLOB.special_hearing_objects)
+			if(!O)
+				LAZYREMOVE(GLOB.special_hearing_objects, O)
+				continue
+			if(isInSight(O, T) || (O in hearturfs) || (isInSight(O.locs[1], T)))
```
Drulikar, line 143: Byond uses short-circuiting for boolean (see https://www.byond.com/docs/ref/#/operator/|| ) so it would likely be better to first test if O is in our list before performing a line of sight check.

code/modules/mob/living/say.dm
```diff
@@ -106,45 +106,42 @@ var/list/department_radio_keys = list(
...
-
-			for(var/mob/M as anything in GLOB.player_list)
-				if((M.stat == DEAD || isobserver(M)) && M.client && M.client.prefs && (M.client.prefs.toggles_chat & CHAT_GHOSTEARS))
-					listening |= M
-					continue
-				if(M.loc && (M.locs[1] in hearturfs))
-					listening |= M
+			for(var/mob/M as anything in hearers(message_range, T))
+				listening += M
+				hearturfs += M.locs[1]
```
Drulikar, line 111: Since we're primarily interested in looking up values, not iterating on every entry we enter into hearturfs, it would be better to construct this as an associative list. See https://hackmd.io/@tgstation/ry6DxGvDL#Associative-vs-Normal-Lists and https://www.byond.com/docs/ref/#/list/associations

code/game/machinery/cryopod.dm
```diff
@@ -182,7 +182,6 @@ GLOBAL_LIST_INIT(frozen_items, list(SQUAD_MARINE_1 = list(), SQUAD_MARINE_2 = li
 /obj/structure/machinery/cryopod/Initialize()
 	. = ..()
 	announce = new /obj/item/device/radio/intercom(src)
-	flags_atom |= USES_HEARING
 
 /obj/structure/machinery/cryopod/Destroy()
 	QDEL_NULL(occupant)
```
Drulikar, line 187: If there is an occupant, and this is destroyed, we will cause our reference in special_hearing_objects to go stale.

Also despawn_occupant below will leave this in the list incorrectly.

code/game/machinery/medical_pod/sleeper.dm
```diff
@@ -31,6 +30,7 @@
 
 
 /obj/structure/machinery/sleep_console/Destroy()
+	LAZYREMOVE(GLOB.hearing_objects, src)
```
Drulikar, line 33: We never add a sleep_console to the list, so unnecessary to remove it.

code/game/machinery/rechargestation.dm
```diff
@@ -87,6 +87,7 @@
 	return 1
 
 /obj/structure/machinery/recharge_station/stop_processing()
+	LAZYREMOVE(GLOB.hearing_objects, src)
```
Drulikar, line 90: Removing here means we will only remove it from the hearing objects once this charger fully recharges and still has no occupant. Should someone re-enter prior to this, you'll LAZYADD again without the LAZYREMOVE occurring.

code/game/objects/items/devices/radio/beacon.dm
```diff
@@ -8,7 +8,7 @@
 /obj/item/device/radio/beacon/Initialize()
 	. = ..()
 	GLOB.radio_beacon_list += src
-	flags_atom &= ~USES_HEARING // This subtype apparenlty does not want to hear stuff
+	LAZYREMOVE(GLOB.hearing_objects, src)
```
Drulikar, line 11: Unless Initialize can occur again for the same object, it shouldn't ever already be in the list to need to be removed yet.

code/modules/clothing/under/ties.dm
```diff
@@ -20,9 +20,9 @@
 /obj/item/clothing/accessory/Initialize()
 	. = ..()
 	inv_overlay = image("icon" = 'icons/obj/items/clothing/ties_overlay.dmi', "icon_state" = "[item_state? "[item_state]" : "[icon_state]"]")
-	flags_atom |= USES_HEARING
 
 /obj/item/clothing/accessory/Destroy()
+	LAZYREMOVE(GLOB.hearing_objects, src)
```
Drulikar, line 25: We never added this to the list, so don't need to remove it.

Drulikar (changes requested):
What is the purpose of the all the LAZYADD and LAZYREMOVE calls? Look at the define for these: They don't ensure a duplicate isn't added (LAZYOR would), instead they just ensure the list is non-null when adding and the list is nulled when emptied. If you're performance minded, you should just add when you know for sure it won't result in a duplicate, and just always have the list available (stop using the lazy defines to check if its null to create it and stop checking if its empty to null it). If not having duplicates is important, and you can't assure this is the case when adding, then use |= over += but see https://hackmd.io/@tgstation/ry6DxGvDL#Performance-of--vs--on-lists

PR #2657, opened 2023-02-15
Fixes a polio cause

Zonespace27:
This PR fixes monkeys being able to cause the human life SS to sleep (thus, polio) by making their `handle_npc` proc (which is called on `Life`) asynchronous. During several days of TM, I haven't heard of any new reports of polio

🆑 
fix: Fixed monkeys using objects causing polio
/ 🆑


Watermelon914 (changes requested):
this won't fix the bug, there are still other places that can cause infinite sleep. Also input/alert calls will instantly return if the target is clientless.
It does still probably fix a sleeping issue but there's no way to determine if this bug is fixed by changing things around because the codebase is too vast that it's infeasible to determine if one or multiple changes will fix infinite sleeping.

Also, your changelog and PR should be more descriptive. Describe what this is exactly doing and how it is performing the fix.

Zonespace27:
@Watermelon914 

I haven't heard of any new reports of polio, and regardless, this is a solution to a cause of polio
Anyway, PR updated

harryob (approved):
sure seems to have assuaged the outbreak

PR #2661, opened 2023-02-15
Makes the leaky pipe effect into a proper object.

NewyearnewmeUwu:
# About the pull request
This takes the 'leaky pipe' effect as seen here and makes it a proper object.
image
It also changes the colony-side (so far limited to Chance's Claim, since no other maps use this effect) to say 'colony' rather than 'Almayer' with a proper subtype.

Previously, it was defined using /obj/structure/prop/invuln and then manually assigning all the things like name, icon, etc. This is a super efficient way to do things and I may or may not have ideas about implementing these onto the Almayer as fixable messes that MTs can handle. Also, it's very movielike and makes the Almayer feel more alive.

<details>

</details>

# Changelog

:cl:
code: leaky pipes are now under a proper type. please use this in the future, mappers.
maptweak: all of the Chance's Claim leaky pipes now no longer reference the Almayer.
/:cl:

### DIFF
--- a/code/game/objects/structures/props.dm
+++ b/code/game/objects/structures/props.dm
@@ -892,7 +892,7 @@
 
 /obj/structure/prop/invuln/leaky_pipe/Initialize(mapload)
 	. = ..()
-	if(is_mainship_level(src.z))
+	if(is_mainship_level(z))
 		desc += "The Almayer has sprung a leak!"
 	else
 		desc += "The colony has sprung a leak!"
--- a/code/game/objects/structures/props.dm
+++ b/code/game/objects/structures/props.dm
@@ -893,7 +893,7 @@
 /obj/structure/prop/invuln/leaky_pipe/Initialize(mapload)
 	. = ..()
 	if(is_mainship_level(z))
-		desc += "The [MAIN_SHIP_NAME] has sprung a leak!"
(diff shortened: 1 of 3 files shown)


code/game/objects/structures/props.dm
```diff
@@ -883,6 +883,16 @@
...
 	density = FALSE
 
+/obj/structure/prop/invuln/leaky_pipe
+	name = "pipe water"
+	icon = 'icons/obj/structures/props/watercloset.dmi'
+	icon_state = "water"
+	desc = "The Almayer has sprung a leak!"
+	density = FALSE
+
+/obj/structure/prop/invuln/leaky_pipe/colony
```
harryob, line 893: you can be smart and make it get this description on initialize

this is pseudocode
```dm
/Initialize(mapload)
 . = ..()
 if(ismainshiplevel(z))
  desc = "[MAINSHIPNAME] has sprung a leak!"
 return
 if(isgroundlevel(z))
  desc = "[SSmapping.configs[GROUND_MAP].map_name] has sprung a leak!"
```

harryob (changes requested):

NewyearnewmeUwu:
Closing this so I can expand the scope of changes- overhead pipes themsleves also lack a proper type, so I'll need to do those as well.

PR #2668, opened 2023-02-16
Leaky Pipe 2: Revengeance

NewyearnewmeUwu:
# About the pull request
This is a continuation of #2661 that has gone far too far. I have given pipes- which were previously **individually defined** their own type which is used on maps. I also gave broken pipes and pipe water- leaking water from pipes- their own objects. I also added a layer specifically for pipes.

Look at how much fucking code this gets rid of. This removes _**over 3 times the amount of code it adds**_. Mappers. Why? Why would you do this? Why has nobody done this before? Even Chance's Claim was subject to this kind of inefficient, wasteful code.

Why? Everything works. It should've been done like this since the start.

There were over _**105**_ of these things on the Almayer alone. We had almost 200 of these sinful objects- individually defined objects. Please never do this again, future mappers adding new things.

<details>
I am losing my mind.

https://www.youtube.com/watch?v=f8mL0_4GeV0

</details>

# Changelog

:cl:
refactor: massively reworked overhead pipe code on Chance's Claim, New Varadero and the USS Almayer.
maptweak: overhead pipes on each map share a consistent, efficient type
/:cl:


NewyearnewmeUwu:
P.S. Not sure if it belongs in code or refactor, and also if there's any errors please let me know reviewers.

NewyearnewmeUwu:
My next PR is eliminating all the obj/structure{ lattices and adding a checkgrep.

Nanu308:
Just for reference, removing all instancing won't happen and neither will that be approved as we're just going to end up with code bloat on the other hand.

However, certain instancing such as this is indeed better as a coded object for mappers to place.

Mapping by itself lgtm here.

NewyearnewmeUwu:
> Just for reference, removing all instancing won't happen and neither will that be approved as we're just going to end up with code bloat on the other hand.
> 
> However, certain instancing such as this is indeed better as a coded object for mappers to place.
> 
> Mapping by itself lgtm here.

I'm not removing all instancing, I'm removing and adding in proper objects for the things that have been instanced using /obj/structure{ like the pipes here.

NewyearnewmeUwu:
harry you broke the merge conflict check.. i think

harryob (changes requested):

PR #2682, opened 2023-02-18
Fixes invincible base crusher

PhantomEpicness:
# About the pull request
Fixes an oversight that allowed base crusher to have half it's intended shield cooldown

Never intended on the first place and led to crusher being busted as fuck as it is currently.

This was never intended and was a mess up on my part. It fell through from the painfully long development that Charger had as months went by between testing sessions and TMs, along with my inexperience with larger projects and bad note taking at the time.

Maintainers are also supposed to filter stuff like this but after like a billion code reviews Charger had, I can see how it got through on their end as well. 

Nevertheless this dies here. 

funny contrib moment

it runs

<details>

</details>

# Changelog

:cl: Totalepicness
balance: Fixes base crusher having half it's intended cooldown for the shield ability
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/crusher/crusher_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/crusher/crusher_abilities.dm
@@ -68,7 +68,7 @@
 	action_type = XENO_ACTION_CLICK
 	ability_primacy = XENO_PRIMARY_ACTION_3
 	plasma_cost = 50
-	xeno_cooldown = 12 SECONDS
+	xeno_cooldown = 26 SECONDS
 	var/shield_amount = 200
 
 /datum/action/xeno_action/activable/fling/charger


Nanu308:
This is not a "bugfix" or if so stretching that definition quite hard.

This is not a code issue maintainers are meant to filter out but rather a direct GA issue as it's a pure number change and maintainers that are not GA's would most likely consider the define to be correct as it's for a cooldown.

Crusher nerfs were already on the table, but we'll see if this is enough.

Will be merged in 24h more than likely unless a Maintainer has a objection, but it will see a TM during that timeframe anyways after i am done cooking food ;D

PhantomEpicness:
> This is not a "bugfix" or if so stretching that definition quite hard.
> 
> This is not a code issue maintainers are meant to filter out but rather a direct GA issue as it's a pure number change and maintainers that are not GA's would most likely consider the define to be correct as it's for a cooldown.
> 
> Crusher nerfs were already on the table, but we'll see if this is enough.
> 
> Will be merged in 24h more than likely unless a Maintainer has a objection, but it will see a TM during that timeframe anyways after i am done cooking food ;D

changed to balance tags

PR #2683, opened 2023-02-18
Pounces now properly toggle the hiding icon. PART 2: REVENGEANCE

### LINKED ISSUE #1784: Pouncing as runner doesn't toggle hide icon state

NewyearnewmeUwu:
# About the pull request
Followup to #2574. This time I've implemented the unhide function into a element rather than awkwardly placing it in the general-powers.dm file.

This fixes #1784 and is way more elegant than my previous solution.

<details>
<summary>help harry is making me learn how to be a good coder</summary>

</details>

# Changelog

:cl:
fix: pouncing while hidden will now properly toggle the hiding icon.
/:cl:

### DIFF
--- a/code/datums/elements/xeno_pounce_unhide.dm
+++ b/code/datums/elements/xeno_pounce_unhide.dm
@@ -16,6 +16,6 @@
 /datum/element/xeno_pounce_unhide/proc/pounce_toggle(mob/living/carbon/xenomorph/xeno)
 	SIGNAL_HANDLER
 	for(var/datum/action/a in xeno.actions)
-		if a.name == "Hide"
+		if (a.name == "Hide")
 			a.update_button_icon(xeno)
 
--- a/code/datums/elements/xeno_pounce_unhide.dm
+++ b/code/datums/elements/xeno_pounce_unhide.dm
@@ -5,11 +5,11 @@
 	. = ..()
 	if(!isxeno(target))
 		return ELEMENT_INCOMPATIBLE
-	RegisterSignal(target, COMSIG_XENO_POUNCE_HIDE, PROC_REF(pounce_toggle))
+	RegisterSignal(target, COMSIG_XENO_POUNCE_UNHIDE, PROC_REF(pounce_toggle))
 
 /datum/element/xeno_pounce_unhide/Detach(datum/source, force)
 	UnregisterSignal(source, list(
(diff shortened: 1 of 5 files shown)


code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
```diff
@@ -389,9 +389,10 @@
 		return
 
 	if(X.layer == XENO_HIDING_LAYER) //Xeno is currently hiding, unhide him
+		SEND_SIGNAL(src, COMSIG_XENO_POUNCE_UNHIDE)
 		X.layer = MOB_LAYER
 		X.update_wounds()
-
+		to_chat(X, SPAN_NOTICE("You have stopped hiding."))
```
harryob, line 395: what if you, bear with me here, used a balloon alert (i like them)

```suggestion
		to_chat(X, SPAN_NOTICE("You have stopped hiding."))
		X.balloon_alert("not hiding")
```

code/datums/elements/xeno_pounce_unhide.dm
```diff
@@ -0,0 +1,21 @@
...
+	UnregisterSignal(source, list(
+		COMSIG_XENO_POUNCE_UNHIDE
+	))
+	return ..()
+
+/datum/element/xeno_pounce_unhide/proc/pounce_toggle(mob/living/carbon/xenomorph/xeno)
+	SIGNAL_HANDLER
+	for(var/datum/action/a in xeno.actions)
+		if (a.name == "Hide")
+			a.update_button_icon(xeno)
```
harryob, line 20:
```suggestion
	for(var/datum/action/current_action in xeno.actions)
		if(current_action.name == "Hide")
			current_action.update_button_icon(xeno)
```

harryob: also your update_button_icon proc doesn't need xeno passing into it, right?

harryob (changes requested):

PR #2557, opened 2023-02-04
Adds Economy Inflation random event

carlarctg:
# About the pull request

Adds a new (and currently the only) random event, 'Economy Inflation'. The USS Almayer recieves a broad-band signal sent to all nearby UA-affiliated colonies and vessels that requests a specific kind of vendors stationed on them have their prices increased, due to some randomly-selected incident backstory. There will be a random delay before the change comes into effect, so make sure you buy all you need beforehand! The announcement will only be visible shipside to avoid communications clutter.

Added 'product type' variable to non-CM vendors, such as food, beverages, souto, recreation, used for the above. 

Added a global list for non-CM vending machines.

# Explain why it's good for the game

Random events are a great way to spice up the round by introducing new, non-player-created variables that don't heavily impact the round and allow players to utilize them to create new RP or gameplay avenues. Such as a marine desperately buying out every Souto vendor before the prices increase by 400%, complaining on communications about the 'dang CLF taking over all those colonies' or 'the gosh darn economy these days', and so forth. 

This PR also introduces an actual random event to use as template for single celled contributor organisms to be able to copypaste it and make a new event that way.

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl:
add: Adds a new (and currently the only) random event, 'Economy Inflation'. The USS Almayer recieves a broad-band signal sent to all nearby UA-affiliated colonies and vessels that requests a specific kind of vendors stationed on them have their prices increased, due to some randomly-selected incident backstory. There will be a random delay before the change comes into effect, so make sure you buy all you need beforehand! The announcement will only be visible shipside to avoid communications clutter.
code: Added 'product type' variable to non-CM vendors, such as food, beverages, souto, recreation, used for the above. 
code: Added a global list for non-CM vending machines.
/:cl:

PR #2689, opened 2023-02-19
Revert "Adds Economy Inflation random event"

blackdragonTOW:
Reverts cmss13-devs/cmss13#2557

RP value aside, this event is triggering way too often in its current state.  The frequency of the event should tuned down significantly.   Given that the automated event system gives us an array of events, and this is the only event, maybe we should have other events to go alongside this instead of it being THE event.  

tl;dr Inflation happens way too often to be "random"


harryob:
increase the weight of the “nothing happens” event

PR #2692, opened 2023-02-20
Removes knockback from certain ammo types.

DecemberIsHere:
# About the pull request

This PR removes the push back effect from certain ammo types, such as heavy revolver, buckshot, M4RA impact rounds, et cetera. This does not affect slugs however. Note, that the push back effect only works within 2 tiles or on point blank, so this PR mainly affects ammo types such as the heavy revolver and buckshot.
Tested on a dev server with these ammo types.

The push back/knockback effect from stunning ammo types have been in the game for a while, however despite their visually appealing effect it is not good for the game. Not only does it provide a safe net for xenos that are punished by a PB with a shotgun, it also doesn't take into account in game factors such as being on weeds, slowed, thus punishing the player for utilizing a debuffing effect (the actual stuns they used) and can prevent follow up hits from other marines. The damage received by wall slams was also extremely low, to the point where its unnoticeable.

I believe this should be test merged, to see the effect of "knockback" ammo types not pushing xenos away, and decide if it should truly stay.

# Changelog

:cl:
balance: Removed the push back effect from certain ammo types
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -113,11 +113,14 @@
 
 	shake_camera(living_mob, 3, 4)
 	knockback_effects(living_mob, fired_projectile)
-	slam_back(living_mob, fired_projectile)
 
 /datum/ammo/proc/slam_back(mob/living/living_mob, obj/item/projectile/fired_projectile)
 	//Either knockback or slam them into an obstacle.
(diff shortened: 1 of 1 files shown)


Zonespace27:
Fwiw, you don't need to make a new PR to change a changelog label, you can just hit the "edit" button up here
image

morrowwolf:
The push back gives good visceral feedback for when something gets stunned and I'm happy with the current state of balance with it existing.

I appreciate the time you put into this but this change won't be the way we go for now.

PR #2694, opened 2023-02-20
Increases rounds needed to vote for WO, and brings back its minimum pop limit.

DecemberIsHere:
# About the pull request

This PR increases the minimum amount of rounds needed for WO to appear in gamemode voting by 20, and reverts the change that removed its 130 pop limit.

WO appears far too much, I highly doubt it was intended to appear at least once a day, sometimes more so if the rounds end too fast.  Also, the MR that changed WO into a voteable gamemode mentioned it to be a gamemode in which it would once every 5 days, this is clearly not the case. 
image

This will increase the rarity of WO, and make it more appealing to play (and more fun, as there will be more players on each side with the pop limit).

# Changelog

:cl:
add: Increased the amount of rounds needed to vote for WO by 20, and brought back the 130 pop limit for the map
/:cl:

### DIFF
--- a/code/game/gamemodes/colonialmarines/whiskey_outpost.dm
+++ b/code/game/gamemodes/colonialmarines/whiskey_outpost.dm
@@ -76,7 +76,7 @@
 	hardcore = TRUE
 
 	votable = TRUE
-	vote_cycle = 25 // approx. once every 5 days, if it wins the vote
+	vote_cycle = 45 // Every 45 rounds
 
 	taskbar_icon = 'icons/taskbar/gml_wo.png'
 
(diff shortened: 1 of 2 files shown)


Zonespace27:
> image This shouldn't happen at all.

`Whiskey` is part of the [NATO Phonetic Alphabet](https://en.wikipedia.org/wiki/NATO_phonetic_alphabet), of which part of the name of an operation pulls from. The indication that it's 3 near-back-to-back whiskey outpost rounds is decieving.

Zonespace27:
This hasn't been touched, doesn't fix it right, and I don't see the need for this, personally.

PR #2698, opened 2023-02-21
Fixes crates being anchored when they land.

NewyearnewmeUwu:
# About the pull request
I did a typo in the commit's title but this should fix the issue we have wherein sometimes crates are not movable when they land. They get anchored when they launch.

Bugs bad gr.. I didn't actually test this but it should work.

<details>

</details>

# Changelog

:cl:
fix: crates are now always unanchored when they land
/:cl:

### DIFF
--- a/code/modules/cm_tech/droppod/supply.dm
+++ b/code/modules/cm_tech/droppod/supply.dm
@@ -7,6 +7,8 @@
 /obj/structure/droppod/supply/open()
 	. = ..()
 	for(var/atom/movable/content as anything in contents)
+		if(istype(content, /obj/structure/closet/crate))
+			content.anchored = FALSE
 		content.forceMove(loc)
 	qdel(src)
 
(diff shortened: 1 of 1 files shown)


code/modules/cm_tech/droppod/supply.dm
```diff
@@ -7,6 +7,8 @@
 /obj/structure/droppod/supply/open()
 	. = ..()
 	for(var/atom/movable/content as anything in contents)
+		if(istype(content, /obj/structure/closet/crate))
```
Zonespace27, line 10: this feels like a bandaid; how does it get anchored in the first place?

Zonespace27: (discussed on discord, the crate is anchored to prevent it being shoved pre-transport, so doing this seems alright)

Zonespace27 (approved):

PR #2711, opened 2023-02-23
hotkey dlc [probably not broken]

harryob:
deleting admin keybinds will get me in trouble with Big Admin (sorry)

changes the sanitizing too
some other stuff

they seem quite stable

:cl:
fix: your hotkeys actually save
fix: unbinding hotkeys works properly
fix: no ghost "Unbound" when you rebind the key to a new key
fix: "Unbound" appears when you unbind a key in a janky way
ui: the keybind menu now searches every category for hotkeys, not just the category you're in
ui: the keybind menu is less janky to use, and doesn't require you to click to select your input
/:cl:


morrowwolf (approved):
Has been TM'd without significant issue. Code looks fine.

PR #2713, opened 2023-02-23
Ghosts will no longer be called Unknown.

NewyearnewmeUwu:
# About the pull request
If you ghost while labeled Unknown for various reasons, ie. being skinned by a pred or _being_ a pred with their mask on, you will no longer be called Unknown in the afterlife. This allows for easier recognition of ghosts and makes conversing in dchat easier.

This seems like an obvious thing to fix.

<details>

</details>

# Changelog

:cl:
fix: you will no longer become Unknown when you ghost
/:cl:

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -107,6 +107,8 @@
 
 	if(!name) //To prevent nameless ghosts
 		name = capitalize(pick(first_names_male)) + " " + capitalize(pick(last_names))
+	if(name == "Unknown")
+		name = body.real_name
 	change_real_name(src, name)
 
 	minimap = new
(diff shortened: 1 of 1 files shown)


NewyearnewmeUwu:
Fun fact: if a pred skins you after you die with a normal name, and you re-enter your corpse, you become unknown!

code/modules/mob/dead/observer/observer.dm
```diff
@@ -107,6 +107,8 @@
 
 	if(!name) //To prevent nameless ghosts
 		name = capitalize(pick(first_names_male)) + " " + capitalize(pick(last_names))
+	if(name == "Unknown")
+		name = body.real_name
```
stanalbatross, line 111: recipe for trolling -> set name to "unknown" -> observe -> you have no body and this forces a runtime

add sanity check please

stanalbatross (changes requested):

code/modules/mob/dead/observer/observer.dm
NewyearnewmeUwu: you can't even name yourself Unknown for some reason, but I've done as you asked

harryob (approved):

PR #2721, opened 2023-02-25
Lv 624 junglification

LynxSolstice:
# About the pull request

I thought LV624 wasn't jungle-y enough for being a jungle planet, so I was talking to harryob and he told me I can just, map it myself, so I did.

Adds more vines and shrubs and trees and greenery, makes the place feel less like a desert as soon as you cross the river.

# Testing stuff

It runs and initializes and loads, it's a map edit.

# Changelog

:cl:LynxSolstice
maptweak: Makes LV624 more jungle-y and less desert-like. North of the river changed to reflect this. All grass turfs (turf/open/gm/grass) replaced with their weedable variant (turf/open/gm/grass/weedable). The colony grass turfs were replaced with a visually distinct WO variant, that is still unweedable.
/:cl:


Zonespace27:
Adding balance tag due to the jungle area being weedable

Agameofscones (changes requested):
You should use a different grass on the opposite bank, this will help differentiate what is weedable and what isn't.

You need to double check the inserts to make sure they blend properly where pass through turfs aren't appropriate.

image
changes like these, where new walls (that are indestructible) are introduced should be noted and considered more carefully. These western walls choke up the pre-drop combat lane. Please remove them, the whole area deals with enough resin spam as is.

image
tall grass clusters behind doorways should be avoided.

image
flora object spam like this looks repetitive and reduces overall performance. Try to avoid this much ground clutter. You should remove the majority of this.

Overall, I think this is OK, however I do believe the sandy turf paths helped in navigation in and around the south of river region, and added extra visual variety. I think you should re-add some in some capacity, right now, this makes the map harder to navigate for seasoned and new players, by reducing visual variety and landmarking, which is a big negative.

Agameofscones:
image
Don't do this. All these turfs are on the same bleed layer and will look identical in game as they are in editor. Which is to say, really bad because there is zero blending on the border.
image
Don't do this sort of object spam. Its bad for visibility, changes the area's whole balance, and on top of that vines are fucking awful to deal with on either side and SERIOUSLY favor xenos. You've also got double vines on some tiles. Don't do that, one vine per tile.
image
Don't put vines on top of dense turfs (walls), it looks awful, perspective clash.
image
Can you explain your reasoning for changing all grass to this type? Its an old legacy type that we shouldn't be using anymore honestly. You should instead be using the parent type of the weedable grass `turf/open/gm/grass/weedable` (which is what it was beforehand, this is a pointless refactor as they're functionally identical types, and the original doesn't have WO code baggage)
Regarding your earlier question, dirt grass borders can be weedable, or not, depends on the type you use. Check the object tree.
image
As before, no vine spam. Please don't use vines.
image
These are indestructible walls and change the gameplay of this area. Is this a cosmetic PR or one intending to balance? Right now it chokes up the main combat area in LV with excess amounts of objects that will catch clicks, block player vision, and generally be a huge nuisance.
image
Do not put tall grass in the middle of combat lanes where it did not exist previously.
image
Noted earlier, but these tiles are visually **identical**. This is _really_ bad for communicating to the xeno player (and to an extent the marine player) what can be, and cannot be weeded. A new player who does not have LV veterancy cannot accrue this knowledge without trial and error, this will lead to frustration.

I am, after further review, not in favor of this PR. It does not improve the map in its present state. The changes to sightlines are a big negative, mostly because of vine spam. The increase in object density is also a pretty large negative for map performance and click catching. Due to combat delay cooldown from melee the increase in click-catch objects significantly debuffs marines in this area.

T-fort and beach is a primary combat area, and the reduction in visible legibility is not a good thing.

While the intentions of this PR are well meaning, it winds up just being a pretty silent xeno buff, which is very much not needed on LV.

PR #2741, opened 2023-02-27
escape menu fix

harryob:
the menu no longer claims you're on the almayer on WO
the menu no longer hates LV522

:cl:
ui: the escape menu was made a bit more readable on certain maps and gamemodes
/:cl:


morrowwolf (approved):
Code looks good

PR #2747, opened 2023-02-28
M4RA handling sound stolen from l42A

CapCamIII:
# About the pull request

Gives l42 handling sounds(loading, unloading, chambering) to M4RA

Changes L42A mention to M4RA in HPR desc.

ITS NOT THE FIRING SOUND

I meant to do this but I forgot to and I logged onto CM and heard the shitty base sound effects and went like 😭 😭 😭 so I'm fixing it with the L42 sounds because those sound really good and it has l42 stats mostly 😄 

also fixes an error in hpr desc which is good because fixing things is good

<details>

I didn't test it but it compiled.

</details>

# Changelog

:cl:
add: M4RA now uses L42A handling effects for loading/unloading and chambering
spellcheck: HPR now says M4RA instead of L42A when going over the other 10x24 marine weapons
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -1052,7 +1052,7 @@
 
 /obj/item/weapon/gun/rifle/lmg
 	name = "\improper M41AE2 heavy pulse rifle"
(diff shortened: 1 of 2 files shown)


morrowwolf (approved):
Much much better +1, code LGTM

PR #2753, opened 2023-03-02
kills the old maploader, refactors usage in interiors and removes deprecated admin verb

harryob:
the old maploader was kinda slow and this was messy snowflake code

new subsystem +++ nice awesome

:cl:
refactor: the interior manager is gone, long live ssinterior
/:cl:


fira (changes requested):
Should remove `code/modules/maps` entirely while you're at it and free us from the bad maploader

fira (approved):
perfection

morrowwolf (approved):
Seconding code good and TM has gone without any issues.

PR #2754, opened 2023-03-02
notifications when castes are available to evo into

harryob:
it was kinda annoying

also removes options you can't evolve into from the list

:cl:
qol: you now receive notifications when it is possible to evolve into certain castes, based on roundstart time locks
qol: unavailable castes are no longer shown in the evolve list
/:cl:


Ultimate-Fluff:
Yos yos, thank you, Harry!

morrowwolf (approved):

PR #2764, opened 2023-03-03
you can now see emotes while deaf

NewyearnewmeUwu:
# About the pull request
There's a bug where if you're deaf you can't see any emotes, including your own. This fixes it.

Bugs bad. Also a ton of medics give thumbs up in case people are deaf and this allows them to do that without the bug interfering.

<details>

</details>

# Changelog

:cl:
fix: you can now see emotes while deaf
/:cl:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -190,11 +190,12 @@
 // Show a message to all mobs in earshot of this atom
 // Use for objects performing only audible actions
 // message is output to anyone who can see, e.g. "The [src] does something!"
-/atom/proc/audible_message(message, max_distance, message_flags = CHAT_TYPE_OTHER)
+// deaf_message (optional) is what deaf people will see e.g. "[X] shouts something silently."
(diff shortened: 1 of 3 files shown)


code/modules/mob/mob.dm
```diff
@@ -194,7 +194,7 @@
 	var/hear_dist = 7
 	if(max_distance) hear_dist = max_distance
 	for(var/mob/M as anything in hearers(hear_dist, src.loc))
-		M.show_message(message, 2, message_flags = message_flags)
+		M.show_message(message, (isdeaf(M) ? 1 : 2), message_flags = message_flags)
```
harryob, line 197: use the defines for showmessage

harryob: also this isn't really the right approach, if you're deaf you should miss all audible messages. don't we want to change what's calling it

harryob (changes requested):

harryob (changes requested):

PR #2774, opened 2023-03-05
predator flavour text fix

harryob:
predator flavour text was being encoded twice, leading to apostrophe weirdness

:cl:
fix: whitelisted predators can now use apostrophes in their flavour texts
/:cl:


realforest2001 (approved):
Works as advertised.

morrowwolf (approved):

PR #2777, opened 2023-03-05
New *warcry sound variations

XSlayer300:
# About the pull request
Adds 23 male warcry and 12 female warcry variations for additional variety, content all taken from Battlefield V (as some of our *warcry are used from that game, especially the female variations)

# Changelog

:cl:
soundadd: Added 23 male *warcry variations.
soundadd: Added 12 female *warcry variations.
/:cl:


Agameofscones (approved):
one of these days all these ripped sounds will come back to bite us in the ass.

Then again we've had stuff like this since 2016 so 🤷

Herotitan772:
I feel like this TM is a bad addition. The main problem i have with it is that the new soundbytes are of varying quality and really ruin my immersion, the other main issue i have is that its also kinda..... strange to hear some british guy say "woo hoo lets go" instead of the normal generic warcrys.

all in all i just think its best to keep to the same audio quality for all the soundbytes and preferably the same nationality

XSlayer300:
> I feel like this TM is a bad addition. The main problem i have with it is that the new soundbytes are of varying quality and really ruin my immersion, the other main issue i have is that its also kinda..... strange to hear some british guy say "woo hoo lets go" instead of the normal generic warcrys.
> 
> all in all i just think its best to keep to the same audio quality for all the soundbytes and preferably the same nationality

Problem is that all of the added voicelines are of the same nationality, these ARE the American voicelines used in that game. However, they are the following:

- American soldiers from the German war story.
- American Elites from the multiplayer.
- American soldiers from the multiplayer segment of the game (which is the main one), Female variants included as Elites nor singleplayer AI do not have female characters

Now the thing is that, the voicelines added prior to this PR are somewhat British, the female voices especially since they came from an Elite whose nationality is British. Not to mention the fact that everyone in CM-SS13 has a different nationality too, but we'll ignore it since we don't have a dedicated voice option.

In addition, these are directly ripped from the game files without any modification. Now I don't know how did the original person made it got those files (probably recording it on some YT website, idk), but that is what I've tried with.

PR #2783, opened 2023-03-06
Nerfs to the Scout Custom M4RA.

DecemberIsHere:
# About the pull request

This PR reverts undocumented changes to the M4RA custom scout rifle, which was recently merged. Namely, the instant wield, the fire rate delay decrease by 2, a damage multiplier of 0.30, and a accuracy multiplier as well.

Undocumented, balance changes in a PR that says "with less balance changes", which severely buffed the scout rifle to be an extremely quick and safe to say OP weapon. Not only does it have FULL AP (meaning no castes aside from max rage berserker can have a semblance of reduced damage), 66 damage on just the base ammo (which is given in abundance), it also have absolutely 0 fall off (which i did not change, as I believe scouts shouldn't have falloff but in the case of current scout rifle it is far too powerful with the other buffed stats). It absolutely shreds almost every caste, and is bad balancing. This PR reverts those changes.

image

This isn't documenting the insane buffs to the scout rifle, and I'd go so far as to say the author of the PR understood these changes would make the rifle extremely powerful. Also, the issue with "omega slime" with ammo switching on scout could be solved without the need for a full stat increase on damage, accuracy, fire rate. 

# Changelog
:cl:
balance: Nerfs to the Scout Custom M4RA.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/guns/specialist.dm
@@ -570,7 +570,6 @@
 
 	flags_gun_features = GUN_AUTO_EJECTOR|GUN_SPECIALIST|GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER
 	map_specific_decoration = TRUE
-	wield_delay = WIELD_DELAY_NORMAL
 	aim_slowdown = SLOWDOWN_ADS_QUICK
 	flags_item = TWOHANDED|NO_CRYO_STORE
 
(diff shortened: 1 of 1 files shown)


Unknownity:
They were documented "M4RA Custom(scout gun), gets L42 stats as well, with the exception of having less of a damage mult, meaning when not using the spec ammo, it is out-preformed by the standard m4RA."
balance: Scout M4RA now has L42/M4RA standard stats with the exception of lower damage.
balance: Scout M4RA now can fit magnetic harness, laser sight, however it can no longer fit recoil compensator

CapCamIII:
Still not undocumented changes, you going to address that?

morrowwolf (approved):
Using the tools at your disposal (custom ammo) to maximize your effectiveness is not sliming. If that specific combo is too powerful then that's a different discussion but using equipment creatively should not be looked down on.

Otherwise, yeah this was a bit much. Possibly more nerfs in the future if the damage is still too wacky.

PR #2760, opened 2023-03-03
Altitude Control: Console, Skills, TGUI, Shuttle, OB, And Hijack changes.

Moonshanks:
# About the pull request
This pull request adds a feature via the /machinery/computer/attitude_control_console.

The USS Almayer can now be "moved" between three different levels of orbit/proximity to the AO:

High,
Optimal,
Low

Each level comes with changes to the duration of transport timers, time to transport during hijack, and OB cooldown timer, equal to the GLOB.ship_alt value.

High makes OB cooldowns take 200 seconds longer, dropship transport times take 50% longer, and the dropship hijack transport time take 50% longer

Optimal, the default which will stay in place if no one touches the console leaves everything with their default times and cooldowns,

Low, OB cooldowns take 200 shorter, dropship transport times are 50% shorter, the dropship hijack transport time is 50% shorter.

While in Low orbit the Almayer will periodically shake (50% chance to shake every 30 seconds), the Almayer's thrusters will also start building up heat as they battle with making minor adjustments due to the dangerous proximity to the AO. They will gain 10% of their max heat each 30 seconds, once they reach 100% the ship will be forced into the highest orbit to cool off, and cool off slower than normal (5% every 30 seconds) until its orbit is changed.

While in Optimal or High orbit the Almayer thrusters will cool off by 10% per 30 seconds.

The Almayer's orbit may only be changed by those with a "navigations" skill of 1. (only the CO and Synth -- EDIT: XO now has the skill too -- currently but I may add a dedicated RP role for this mechanic later down the line). The orbital level may only be changed every 90 seconds and when it changes the ship will shake violently causing every mob on the z level to fall over.

This PR does not place the Altitude Control Console on the map, so currently, these features don't do anything within a normal round unless staff spawns in the console, however I will be uploading a mapping PR changing the Astronavigations deck if this PR is accepted.

Planned for the future but not yet approved connected to this PR is the "Navigations Officer" the highest auxiliary support personnel with skills the same as an SO, sans a 1 in Navigations, and a 1 or 2 in piloting. The idea for this future role would be set out in the PR, but it would represent a mainly fluff officer role that was unable to deploy under normal SoP.

# Explain why it's good for the game

This PR is a non-intrusive way to add more nuanced gameplay mechanics to COs, especially on the quieter rounds when they aren't swamped by OW duties.

[...]

PR #2814, opened 2023-03-10
Navigations officer

Moonshanks:
# About the pull request

Adds the role Navigations Officer, as set out in: https://github.com/cmss13-devs/cmss13/pull/2760

The Navigations Officer is a 2nd Lieutenant with the skill "navigations" so that it can use the navigations console. They fit in above the Pilot Officer within the auxiliary support roles. They have skills similar to a PO with some better command skills, and worse medical skills to discourage them from deploying.

The idea of the role is that it's a command officer which can fill the duty of either a PO or an SO. But under SOP should be unable to deploy outside of the dropship unless modified by the CO.

This RP is in place to foster a RP role which is held to the higher RP standard of officers.

This PR has been tested in full and works as intended. There is no vendor (yet) included in this PR

It will allow for another RP slot that has a unique and specific mechanic attached to it. It will also be the first (2nd if you count war reporter) RP role to be held to the same RP standards as an officer.

<details>

</details>

# Changelog

:cl: add: The Navigations Officer role, and their custom flavour text uniform. 
/:cl:


Moonshanks:
Closing PR in favour of a different implementation of a similar idea.

PR #2821, opened 2023-03-12
Restores ALL missing icon states in obj/item/weapon/gun path

Ultimate-Fluff:
# About the pull request
This PR restores EVERY. SINGLE. missing icon state or fixes all missing icon state errors in the ```/obj/item/weapon/gun``` path. No need for reasoning. Ammo boxes will be next~

# Changelog

:cl: Puckaboo2
code: Made some icon state text strings empty so they would not cause missing icon state errors.
del: Deleted some lines of code that caused missing icon state errors.
imageadd: All weapons that can be camouflaged now have a basic USCM sprite to make the linter happy.
imageadd: The Energy Pistol now has an icon state.
/:cl:


nauticall (approved):
sprites look good! approved

code/modules/cm_preds/yaut_weapons.dm
```diff
@@ -892,12 +893,11 @@
...
 	return TRUE
 
-
 /obj/item/weapon/gun/energy/yautja/plasma_caster
 	name = "plasma caster"
 	desc = "A powerful, shoulder-mounted energy weapon."
-	icon_state = "plasma"
-	item_state = "plasma_wear"
+	var/dummy_icon_state = "plasma"
+	var/dummy_item_state = "plasma_wear"
```
Watermelon914, line 900: call this base_xxxx_state instead, dummy is a bad name for what it's used for

Watermelon914 (changes requested):

code/modules/cm_preds/yaut_weapons.dm
Ultimate-Fluff: No problem; thank you.

PR #2824, opened 2023-03-12
Vehicle Elevator

realforest2001:
# About the pull request
Makes the elevator start raised. Fewer meme suicides.

Suicide bad.

<details>

</details>

# Changelog

:cl:
add: Vehicle Elevator now starts raised.
/:cl:

### DIFF
--- a/code/modules/shuttle/vehicle_elevator.dm
+++ b/code/modules/shuttle/vehicle_elevator.dm
@@ -49,8 +49,9 @@
 	height = 5
 	dwidth = 2
 	dheight = 2
+	roundstart_template = /datum/map_template/shuttle/vehicle
 
 /obj/docking_port/stationary/vehicle_elevator/adminlevel
 	name = "Adminlevel Vehicle Elevator Dock"
 	id = "adminlevel vehicle"
-	roundstart_template = /datum/map_template/shuttle/vehicle
+	roundstart_template = null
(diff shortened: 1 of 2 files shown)


harryob (approved):

mullenpaul (approved):
I would prefer if the Almayer stationary dock and the admin level stationary docks are both subclasses. It would require a map change and isn't strictly necessary but it would be a cleaner solution.

mullenpaul (approved):
Code wise, perfect. This is the standard that I believe docking ports should be done.

code/modules/shuttle/vehicle_elevator.dm
```diff
@@ -43,14 +43,18 @@
...
-	id = "almayer vehicle"
+	name = "Root Vehicle Elevator Dock"
+	id = "root vehicle"
 	width = 5
 	height = 5
 	dwidth = 2
 	dheight = 2
 
+/obj/docking_port/stationary/vehicle_elevator/almayer
+	name = "Alamyer Vehicle Elevator Dock"
```
harryob, line 54: alamyer

harryob (changes requested):

code/modules/shuttle/vehicle_elevator.dm
realforest2001, line 54:
```suggestion
	name = "Almayer Vehicle Elevator Dock"
```

PR #2831, opened 2023-03-13
Negative Traits are free

wshuwshuwshi:
# About the pull request

Changes the cost of negative biology traits to 0.

Soul. If you're a masochist you can load yourself up on all the bad traits. When more bad traits get added you can make life hell for yourself. Could be fun.
I know it's a strawman but I'm gonna say it again. Having bad eyesight doesn't stop you from learning a new language. I mean maybe if you're blind it might stop you from learning how to read and write it but it wont stop you from speaking it.

</details>

# Changelog

:cl:
balance: go ham on negative traits. They don't give you more points, they just don't take a point now
/:cl:


morrowwolf (commented):
We're not looking for John Marine with 17 negative traits. These should be rarer and not all in one place.

PR #2854, opened 2023-03-16
ports TGUI and TGshuttle for Requisitions

insertnamehere123:
# About the pull request

This ports over the TGshuttle and TGUI stuff for cargo and it's related stuff

Modernizes the code and UI somewhat

This is very much a work in progress so a lot of the old stuff is still in because I need to read through them to get a sense of what does what so I can get the new stuff to fit in without things exploding

<details>

</details>

# Changelog

:cl:
add: adds TGshuttle and TGUI to cargo
del: deletes the old UI and might replace some older code
ui: adds TGUI to cargo
code: changes cargo code and anything it depends on and vice versa
refactor: Basically reworks some cargo stuff
maptweak: this deals in turfs so maps might be affected

/:cl:


code/game/supplyshuttle.dm
```diff
@@ -46,6 +46,192 @@ var/datum/controller/supply/supply_controller = new()
 
 //SUPPLY PACKS MOVED TO /code/defines/obj/supplypacks.dm
 
+/obj/docking_port/stationary/supply
```
mullenpaul, line 49: I would recommend reworking this. Have /obj/docking_port/stationary/supply as your base class - with your width/height etc in place and have your in game, placed stationary docks as subclasses. You could have /Almayer and /offsite - for example - as subclasses.

You want to do this because the sizes make your base class really simple. I don't know how exactly but you could have the offsite just being in infinite transit, so that you don't actually need to place anything. I made a similar request in https://github.com/cmss13-devs/cmss13/pull/2824

code/game/supplyshuttle.dm
```diff
@@ -46,6 +46,192 @@ var/datum/controller/supply/supply_controller = new()
...
+/obj/docking_port/mobile/supply/proc/clean_railing(datum/source)
+	SIGNAL_HANDLER
+	railings -= source
+
+/obj/docking_port/mobile/supply/canMove()
+	if(is_station_level(z))
+		return check_blacklist(shuttle_areas)
+	return ..()
+
+/obj/docking_port/mobile/supply/proc/check_blacklist(areaInstances)
```
mullenpaul, line 132: I would generally recommend trying to split this logic out. Conceptually I have tried to keep the shuttles code explicitly based around interacting with map stuff, other map objects as part of your transition. This logic could be pulled out into a datum just to improve cohesion

code/game/supplyshuttle.dm
```diff
@@ -46,6 +46,192 @@ var/datum/controller/supply/supply_controller = new()
...
+					var/mob/living/carbon/human/human_to_sell = a
+					if(human_to_sell.stat == DEAD && can_sell_human_body(human_to_sell, faction))
+						continue
+				if(is_type_in_typecache(a, GLOB.blacklisted_cargo_types))
+					return FALSE
+	return TRUE
+
+/obj/docking_port/mobile/supply/request(obj/docking_port/stationary/S)
+	if(mode != SHUTTLE_IDLE)
+		return 2
```
mullenpaul, line 154: I would really prefer if the defines were used rather than magic numbers

code/game/supplyshuttle.dm
```diff
@@ -46,6 +46,192 @@ var/datum/controller/supply/supply_controller = new()
...
+						continue
+				if(ishuman(a))
+					var/mob/living/carbon/human/human_to_sell = a
+					if(human_to_sell.stat == DEAD && can_sell_human_body(human_to_sell, faction))
+						continue
+				if(is_type_in_typecache(a, GLOB.blacklisted_cargo_types))
+					return FALSE
+	return TRUE
+
+/obj/docking_port/mobile/supply/request(obj/docking_port/stationary/S)
```
mullenpaul, line 152: I'm not sure why you would modify the request call like this

code/game/supplyshuttle.dm
```diff
@@ -46,6 +46,192 @@ var/datum/controller/supply/supply_controller = new()
...
+	if(mode != SHUTTLE_IDLE)
+		return 2
+	return ..()
+
+/obj/docking_port/mobile/supply/proc/buy(mob/user)
```
mullenpaul, line 157: Again, I would recommend splitting out the logic of what the supply shuttle does into somewhere else, either a datum or even its own controller if need be. The shuttle doesn't need to be so tightly coupled to the supply system

code/game/supplyshuttle.dm
```diff
@@ -688,6 +874,272 @@ var/datum/controller/supply/supply_controller = new()
...
+		ui.open()
+
+/datum/supply_ui/ui_static_data(mob/user)
+	. = list()
+	.["categories"] = GLOB.all_supply_groups
+	.["supplypacks"] = SSpoints.supply_packs_ui
+	.["supplypackscontents"] = SSpoints.supply_packs_contents
+	.["elevator_size"] = supply_shuttle?.return_number_of_turfs()
+
+/datum/supply_ui/ui_data(mob/living/user)
```
mullenpaul, line 931: This function does a lot of work, would it be possible to refactor out the different parts to make it easier to read?

code/game/supplyshuttle.dm
```diff
@@ -46,6 +46,192 @@ var/datum/controller/supply/supply_controller = new()
```
mullenpaul, line 46: all docking port classes should be moved to where the rest of the shuttles code is - for discoverability purposes, all shuttles should be in the same part of the code base

code/game/machinery/doors/railing.dm
```diff
@@ -15,6 +15,8 @@
 	closed_layer = WINDOW_LAYER
 	density = TRUE
 
+	var/obj/docking_port/mobile/supply/linked_pad
```
mullenpaul, line 18: It's unclear why you need a bidirectional reference

code/game/supplyshuttle.dm
```diff
@@ -688,6 +874,272 @@ var/datum/controller/supply/supply_controller = new()
 		</p></div>"
 
 	name = "[name] - [ordername]"
+
+/datum/supply_ui
```
mullenpaul, line 878: I imagine that the next steps here are to refactor this work into some sort of control computer. It should probably be in its own file

Nanu308:
Undraft the PR and resolve conversation/comments made when you've dealt with them @insertnamehere123

Nanu308:
Maps might be affected? 

You gotta check this out, and not add maptweak into CL unless you're sure cuz it puts the mapping tag down

PR #2878, opened 2023-03-18
Add UPP warcries

victorjosephespinoza:
# About the pull request

Replaces the normal warcry for the UPP faction to use russian voices instead.

The warcries are mostly stuff like  `za rodinu` and `uraa`, so yeah, pretty much just typical soviet warcries.

I haven't focused on adding dozens of voicelines due to the fact that this is a minor faction whose appearance is only in events and/or ERT's. However, I can try to get some more, if requested.

Lately, I have noticed an increase of HvH events (in which, I have participated). I found that it is quite uninmersive how every UPP soldier is literally yelling in english at the same time as marines are also yelling the same voicelines. So yeah. I kind of found it just weird and since then I've been thinking of adding something like this to the server.

Tested it myself, works. I can upload a video if it is really needed, however.

# Changelog

:cl:H20Begod
add:  sound/voice/upp_warcry/* (Sound files, such as `warcry_male_1` , for the UPP)
code: changed `code/modules/mob/living/carbon/human/emote.dm`, in order to add conditionals that will check a player's faction. Right now, it's a simple conditional, however, the code is there to be changed to an switch should somebody else come and add more faction-based voicelines.
/:cl:


harryob (approved):
code fine

harryob:
for the purpose of the audience at home

https://user-images.githubusercontent.com/55142896/226740352-a6d02fb9-9b7d-44c5-804b-fd3e55c429f4.mp4

PR #2888, opened 2023-03-19
Fix for shuttle flight duration.

Segrain:
# About the pull request

Followup to #2760. Let us not throw away all calculations by multiplying the starting value instead of the one actually calculated.
Also, the person in charge of running all the technical systems on the ship should be able to actually _run the technical systems on the ship_.

Is fix.

# Changelog

:cl:
fix: Shuttles' transit time is once again properly calculated.
/:cl:


harryob (approved):
eesh

PR #2903, opened 2023-03-20
Gives SO service jacket on spawn

PhantomEpicness:
# About the pull request
Gives SO service jacket on spawn. Part of a minor CIC update.

They are officers yet don't have the officer jacket. Also useful should they wake up during hijack or red alert so they aren't totally naked.

runs, doesn't crash and burn the server
Screenshot_43

<details>

</details>

# Changelog

:cl: TotalEpicness
add: Gives Staff Officers service jacket on spawn.
/:cl:


morrowwolf (changes requested):
So I like dress options and I really like the look of the service jacket.

But here's my issue. Take a side by side look at the SO and XO. Without the HUD icon it is difficult to tell them apart with this.

So two options:
Make the XO and SO spawns more different.
Just add the service jacket to the SO vendor so it's an option rather than the vision we have for them on spawn.

Kitsunemitsu:
Maybe try a dress jacket instead?

PR #2909, opened 2023-03-21
Reverts Tail Jab and speed changes on Vampire

carlarctg:
# About the pull request

Vampire's Tail Jab now needs you to directly click the target like a tail stab, like it used to be.

> Vampire's Tail Jab now needs you to directly click the target like a tail stab, like it used to be.

I regret making this change, it ended up just making it far too easy to just hit and run from a safe distance and be annoying without any effort in the hands of the Vampire. Not only that, but a lot of people disliked it so since in the end nobody liked this change and I think it actively worsened Vampire and its place in the game I've decided to revert this. Not to mention it has to use the dumb LRP telegraphs, which tail stab doesn't.

> Increased Vampire speed back to its default.

Lowering the speed buff by a tier made Vampire go from pretty fast to ridiculously slow. I've had people compare it to Ravager in terms of slowness, and while it's not accurate, it's not too far off. It makes it very difficult to actually be, well a flanking caste, even if it's a sideliner instead of a backliner, as it makes it very very hard to run from marines with your natural slowness off-weeds. I don't know why the fuck reducing this by one tier made vampire super slow but that's shitcode for you. With it being faster Vampires can be more confident in their harassment tactics instead of needing to hide behind walls like a corner-lunging Warrior because their speed made them easily kiteable.

<details>

</details>

# Changelog

:cl:
balance: Vampire's Tail Jab now needs you to directly click the target like a tail stab, like it used to be.
/:cl:


morrowwolf (changes requested):
Tail Jab change fine.

Slowness is unlikely to get reverted unless rush can only be used if clicking on a human mob like warrior grab.

harryob (approved):

Issue #2923, opened 2023-03-22
Megaphone chat isn't logged for ghosts


Doubleumc:
### Testmerges

#2761, #2723, #2846, #2703, #2913, #2921

### Description of the bug

As a ghost, when viewing someone use the megaphone the text correctly appears above their head but it is not logged in the chat window.
![image](https://user-images.githubusercontent.com/14267245/226805795-2705fb5f-7afa-46f2-a6f5-d58b8860d5f4.png)

### What's the difference with what should have happened?

Chat should appear above someone's head along with being logged to the chat window.

### How do we reproduce this bug?

1.Become a ghost.
2.Observe subject with a megaphone.
3.Have the subject broadcast with the megaphone.

### Issue Bingo

- [ ] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #2930, opened 2023-03-22
FIX: Megaphone chat is now sent to observers

victorjosephespinoza:
closes https://github.com/cmss13-devs/cmss13/issues/2923

Makes megaphone messages appear in dchat too.

fixing bugs and oversights is always good

tested by myself

# Changelog

:cl:H20Begod
fix: Ghosts can now see megaphone messages
code: When a message is sent (using the megaphone), a new iteration over  GLOB.observer_list has been added to send the a notification to observers.
/:cl:


ItsVyzo:
changes _solves_ to _fixes_ so it autocloses the issue once merged

morrowwolf (changes requested):
We don't want every observer to hear what is broadcasted via the megaphone just the ones in view.

victorjosephespinoza:
> We don't want every observer to hear what is broadcasted via the megaphone just the ones in view.

What about now with the changes made in 0f9a935af3ed41ea0d798a3ea76b3f02e0861809? 
All mobs in view are iterated, but if the mob is not an human or an observer, it gets excluded to the other message.

morrowwolf (changes requested):

code/game/objects/items/devices/megaphone.dm
```diff
@@ -32,14 +32,13 @@
 	log_admin("[key_name(user)] used a megaphone to say: >[message]<")
 
 	if((src.loc == user && !user.is_mob_incapacitated()))
-		var/list/mob/living/carbon/human/human_viewers = viewers(user) // slow but we need it
-		for(var/mob/living/carbon/human/listener in human_viewers)
-			if(isyautja(listener)) //NOPE
+		var/list/mob/human_viewers = viewers(user) // slow but we need it
```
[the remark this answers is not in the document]
victorjosephespinoza: True, forgot of that. commit on the way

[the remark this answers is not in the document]
victorjosephespinoza: 996de84812f596c863e26b92fe7dd88a9a96b5b7

code/game/objects/items/devices/megaphone.dm
```diff
@@ -32,14 +32,14 @@
 	log_admin("[key_name(user)] used a megaphone to say: >[message]<")
 
 	if((src.loc == user && !user.is_mob_incapacitated()))
-		var/list/mob/living/carbon/human/human_viewers = viewers(user) // slow but we need it
-		for(var/mob/living/carbon/human/listener in human_viewers)
-			if(isyautja(listener)) //NOPE
+		// get mobs in the range of the user
+		var/list/mob/listeners = viewers(user) // slow but we need it
+		for(var/mob/listener in listeners)
+			if(!ishuman(listener) && !isobserver(listener))
```
harryob, line 38:
```suggestion
			if(!ishuman_strict(listener) && !isobserver(listener))
```
otherwise yautja (which are just a species of human) won't get caught

harryob (changes requested):

code/game/objects/items/devices/megaphone.dm
victorjosephespinoza: wasn't aware of that. Will test it out now thanks

victorjosephespinoza: 00f8d12c99f8419cdc83f9063366f62f9faeab35

also fixed long text being visible by filtering out mobs in a new list, and used humansynth_strict instead of is_human like you recommended, because I noticed synths couldn't hear the broadcast either

harryob: > 

good catch

harryob (approved):

PR #2936, opened 2023-03-23
Flayed ghosts can no longer have offsets

NewyearnewmeUwu:
# About the pull request
This makes sure that ghosts (usually ones who have been flayed) cannot become a ghost with offset on. Otherwise you have a pixel_x of about 6 and a pixel_y of about 9 if you're flayed and ghost.

Bugs bad and it looks ugly.

<details>

</details>

# Changelog

:cl:
fix: Ghosts can no longer become offset.
/:cl:

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -113,6 +113,10 @@
 			name = body.real_name
 	change_real_name(src, name)
 
+	if(pixel_x != 0 || pixel_y != 0) //To prevent weirdly offset ghosts.
+		pixel_x = 0
+		pixel_y = 0
+
 	minimap = new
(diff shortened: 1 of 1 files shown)


code/modules/mob/dead/observer/observer.dm
```diff
@@ -113,6 +113,10 @@
 			name = body.real_name
 	change_real_name(src, name)
 
+	if(pixel_x != 0 || pixel_y != 0) //To prevent weirdly offset ghosts.
+		pixel_x = 0
+		pixel_y = 0
```
harryob, line 118: you might as well just set the pixel_x and _y to 0 regardless then

harryob (changes requested):

PR #2938, opened 2023-03-23
Vampire Lurker's rush now properly reduces cooldown upon a hit.

### LINKED ISSUE #2836: Vampire Rush not properly reducing CD if hit
### Testmerges

#2521, #2776, #2756, #2752, #2753, #2777

### Description of the bug

When properly hit, Vamp rush CD is never reduced. While I'm way too lazy to figure out the fix myself, I assume that the object is not properly given when nothing is passed into the apply_cooldown_override function, when "owner" is required(though owner might be defaulted, not entirely sure). Additionally, it's not even properly set to the right number of 30, as per the original PR's statement of "Reduced to three seconds if landed", whereas the current code(should) set it to 4 seconds. 

Rush cooldown should be set to three seconds, but is defaulted to 6 seconds instead.

1. Rush normally and miss a target and record time until off CD
2. Rush a target and land it and record time until off CD
3. Both are the exact same time.

### Issue Bingo

NewyearnewmeUwu:
# About the pull request
Fixes #2836. This adds a check to make sure that apply_cooldown doesn't early return if an ability has a lowered/increased cooldown upon a hit/miss respectively. It also lowers the on-hit CD for Vampire's rush to 3, rather than the 4 it is in the code right now.

Bugs bad.

<details>

</details>

# Changelog

:cl:
fix: Vampire Lurker's rush now properly reduces cooldown upon a hit.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/lurker/lurker_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/lurker/lurker_powers.dm
@@ -109,7 +109,7 @@
 	target.apply_armoured_damage(get_xeno_damage_slash(target, xeno.caste.melee_damage_upper), ARMOR_MELEE, BRUTE, "chest")
 	playsound(get_turf(target), 'sound/weapons/alien_claw_flesh3.ogg', 30, TRUE)
 	shake_camera(target, 2, 1)
-	apply_cooldown(0.5, TRUE)
+	apply_cooldown(0.5)
 
 /datum/action/xeno_action/activable/flurry/use_ability(atom/targeted_atom) //flurry ability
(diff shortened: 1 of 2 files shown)


morrowwolf (changes requested):
This has turned into more than you probably signed up for but if we are going to allow the cooldown reduction then rush can only be used via spriteclicking as it's currently already pretty wild how fast vamp lurkers move on top of an escape.

Drulikar:
Just so its not lost, regarding requiring the ability to be a sprite click:
> A noteable example of this is tail stab
> As in /datum/action/xeno_action/activable/tail_stab/use_ability(atom/targetted_atom) where the atom targeted needs to be the human; not if the atom is a tile should it figure out the human in the tile to hit.

PR #2311, opened 2023-01-14
PMC Survivor weapon lock fix

Unknownity:
# About the pull request

Fixed PMC Survivors being locked out of using PMC weapons.
Alright faction code is a mess, I tried 4-5 different ways so the weapon lock also checks for faction_group but it never worked, the only solution left was switching their main faction of survivor to PMC. they should still have survivor in the faction_group list. Unless someone has a better idea on how to make the code check for faction_group too (because I couldn't make it work with 4 different ways I did it) then this is the only way.
Fixes #2310

# Explain why it's good for the game
it's a bug fix.

# Testing Photographs and Procedure
spawned as surv pmc, gave myself NSG and checked if I could shoot, i could shoot

# Changelog

:cl: Unknownity
fix: PMC Survivors can now use PMC weapons.
/:cl:

PR #2943, opened 2023-03-24
Fixes PMC survivors spawning with ID locked NSG-23

LynxSolstice:
# About the pull request

Title. (Replaces it with the stripped, unlocked variant.)

Oversight, PMC survivors can't fire their own rifles. This bugfixes that.

Changed obj/item/weapon/gun/rifle/nsg23/stripped to obj/item/weapon/gun/rifle/nsg23/no_lock/stripped

# Changelog

:cl:LynxSolstice
fix: Fixes PMC survivors spawning with ID locked NSG-23s.
/:cl:

### DIFF
--- a/code/modules/gear_presets/_select_equipment.dm
+++ b/code/modules/gear_presets/_select_equipment.dm
@@ -828,7 +828,7 @@ var/list/rebel_rifles = list(
 			H.equip_to_slot_or_del(new /obj/item/weapon/gun/shotgun/pump/dual_tube/cmb(H), WEAR_L_HAND)
 			H.equip_to_slot_or_del(new /obj/item/ammo_magazine/handful/shotgun/buckshot(H), WEAR_IN_BACK)
 		if(3)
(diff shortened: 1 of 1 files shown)


Zonespace27 (approved):

Unknownity:
> > Wrong way to tackle this, refer to my PR #2311 Someone removed the faction group from the PMCs.. add FACTION_WY and FACTION_PMC to them
> > your way will allow non-PMCs to use these weapons.
> 
> This was decision was on purpose. They would not have the same exact IFF as factions coming in if they were on the colony and not all of the people using that proc fit nicely into PMC/WY. Just use the no lock version or switch out the gun.

PMC and WY do not have IFF against marines, currently, they would not even have IFF against non-surv PMCs (that's bad). If the intention was to remove the factions so they lose IFF capabilities against marines then that was unnecessary as they already didn't have IFF for them, if the intention was to remove IFF capabilities against their own faction and W-Y, in the most friendly way I can say, that's stupid.

PR #2710, opened 2023-02-23
id locks can now be removed

harryob:
captain+ (so XO or CO) can override the lock, as can the person it is locked to, by swiping their id

:cl:
qol: id locks can now be removed icly
/:cl:

PR #2955, opened 2023-03-26
Fix for attack chain, again.

Segrain:
# About the pull request

Followup to #2710 and probably an alternative to #2952. One day we will have some `#define`s like `ATTACKBY_NO_AFTERATTACK` instead of trying to remember whether this or that or yonder thing should return true or false...

Is fix.

# Changelog

:cl:
fix: Items doing things by clicking on floors can do those things again.
/:cl:


harryob:
i've incorporated this into #2952 - totally forgot to also revert the flipping in the root proc there, sorry

harryob (approved):

PR #2956, opened 2023-03-26
Introduces a new way to rename items, and cassette pouches now take regulation tapes.

NewyearnewmeUwu:
# About the pull request
This implements a way of renaming items using the rename component. In this case, it checks for a trait and if it has it the rename component will name an item the initial item name with the new name in parentheses. For example, if you rename a regulation tape "Interview 1", it's gonna show up as "regulation tape (Interview 1)".

Tapes now also store the name for each side of the component, so if you take the previously mentioned label of Interview 1 and then flip it, the other side will be unlabeled until you alter it.

Also, regulation tapes now fit in casette pouches for use by the reporter.

This is a better way to handle tapes for tape recorders, makes things more immersive for Combat Correspondents and is more convenient.

<details>

</details>

# Changelog

:cl:
add: Cassette pouches now fit regulation tapes.
add: Flipping a previously labeled tape will store the label for that side and vice versa.
add: Tapes now have a unique renaming system.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/taperecorder.dm
+++ b/code/game/objects/items/devices/taperecorder.dm
@@ -370,6 +370,9 @@
 	var/radial_icon_file = 'icons/mob/radial_tape.dmi'
 	var/list/cassette_colours = list("blue", "gray", "green", "orange", "pink_stripe", "purple", "rainbow", "red_black", "red_stripe", "camo", "rising_sun", "orange", "blue", "ocean", "aesthetic")
 	var/list/cassette_map_themes = list("solaris", "ice", "lz", "dam", "worstmap")
+	inherent_traits = list(TRAIT_ITEM_RENAME_SPECIAL) //used to make the rename component work specially.
+	var/flipped_name //used to store the tape's name for one side
+	var/unflipped_name //see above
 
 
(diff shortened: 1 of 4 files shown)


code/game/objects/items/devices/taperecorder.dm
```diff
@@ -370,6 +370,9 @@
 	var/radial_icon_file = 'icons/mob/radial_tape.dmi'
 	var/list/cassette_colours = list("blue", "gray", "green", "orange", "pink_stripe", "purple", "rainbow", "red_black", "red_stripe", "camo", "rising_sun", "orange", "blue", "ocean", "aesthetic")
 	var/list/cassette_map_themes = list("solaris", "ice", "lz", "dam", "worstmap")
+	inherent_traits = list(TRAIT_ITEM_RENAME_SPECIAL) //used to make the rename component work specially.
+	var/flipped_name //used to store the tape's name for one side
+	var/unflipped_name //see above
```
harryob, line 375: code doc with /// before the var

harryob (changes requested):

harryob (approved):

PR #2973, opened 2023-03-29
Fixes oversight with IOs not getting Fireman Carry.

LynxSolstice:
# About the pull request

Adds Fireman Carry level 1 to the IO role. (The only officer role to not have it I believe.)

Fixes an oversight with the IO skills when switching the fireman carry skills from policing to an independent skill.

Compiles and runs. IOs have the skill.

# Changelog

:cl: LynxSolstice
fix: Fixes an oversight with the IO role not receiving the Fireman Carry skill when it was split off from policing, making it consistent with all other officer roles + the SL role.
/:cl:


morrowwolf (changes requested):
When we spoke on discord I said, "leave it for now". I'm not sure what got lost in translation.

PR #2984, opened 2023-03-31
Fixes preds not having their bones broken, rebalances it as well.

DecemberIsHere:
# About the pull request

Preds have had their bones unbreakable since march of 2021, due to a change in limbs.dm code regarding bonebreak invulnerability chances during tech webs. This PR fixes the longstanding bug, but also rebalances it due to the major concerns of extremely easy bone breaks which were a common thing in the past.

Bug, similar to preds hard crit bug. This does not necessarily gimp pred fights versus humans, and its likely that a pred would be slowed due to pain/blood loss within the damage necessary to accrue a bone break. Preds also have bones, and although they're a stronger species are not immune to bone breaks. Medicomp currently fixes bone breaks, and thwei allows for autoheal within a set amount of time. Overall, nothing will really change aside from preds being able to break their bones again.

Tested on a private server with multiple ammo types/common marine loadouts along with xenos. Works as intended.

# Changelog

:cl: Usnpeepoo
fix: Preds are able to break their bones again
balance: Pred bones have their minimum damage to break requirements balanced.
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -1064,7 +1064,7 @@ treat_grafted var tells it to apply to grafted but unsalved wounds, for burn kit
 
 	//stops division by zero
 	if(owner.chem_effect_flags & CHEM_EFFECT_RESIST_FRACTURE)
-		bonebreak_probability = 100
+		bonebreak_probability = 0
 
 	//If you have this special flag you are exempt from the endurance bone break check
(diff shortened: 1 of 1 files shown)


harryob (approved):
code fine needs bälance

morrowwolf:
TM'd not stale.

morrowwolf (approved):
Seems to do fine. Complaints minimal. Council happy. All systems nominal.

PR #2986, opened 2023-03-31
view refactor

harryob:
datumises view because datums are cool and stupid number strings are not cool


blackcrystall:
Make animate for view, use proc animate(
and blur idea - make it focus on the pixel where you point the mouse, and all other objects are blurred

PR #2999, opened 2023-04-01
"Bottomless Magazine" variable

BeagleGaming1:
# About the pull request

Adds a var for a bottomless magazine to guns, sentry's, and vehicle hardpoints.
As long as the magazine has at least 1 bullet, it'll keep firing that type of bullet.

More funny vars to grief with

# Changelog

:cl:
admin: Adds a variable for a "bottomless mag" for guns
/:cl:

### DIFF
--- a/code/modules/defenses/defenses.dm
+++ b/code/modules/defenses/defenses.dm
@@ -22,6 +22,7 @@
 	var/locked = FALSE
 	var/composite_icon = TRUE
 	var/display_additional_stats = FALSE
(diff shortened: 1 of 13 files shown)


Segrain:
Do we _really_ need an entire variable on every gun, checked every single time a gun is fired, on the off chance once in a blue moon somebody will want to make a fancy event about it? Varediting `current_rounds` to arbitrarily high number emulates an infinite magazine probably well enough for any practical purpose.

harryob (changes requested):
segrain’s correct, this can be a trait instead

PR #3049, opened 2023-04-11
Prosthetic limbs can now be de-calibrated where normal bones would break.

NewyearnewmeUwu:
# About the pull request
If a robotic limb takes enough damage to cause a fracture, it will now 'de-calibrate' and require calibration again.

Synths are immune to this. I also made it so that replacing a synth's limb makes it into a 'synthskin' limb, rendering the new limb immune to breaking like this.

Adds more complexity to the game by adding a downside to robotic limbs (that being that they can break). This is also good practice for doctors.

<details>

</details>

# Changelog

:cl:
balance: Prosthetics can now be decalibrated by damage where normal limbs would break.
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -280,7 +280,7 @@
 	if(!is_ff && take_damage_organ_damage(brute, sharp))
 		brute /= 2
 
-	if(CONFIG_GET(flag/bones_can_break) && !(status & (LIMB_ROBOT|LIMB_SYNTHSKIN)))
+	if(CONFIG_GET(flag/bones_can_break) && !(status & (LIMB_SYNTHSKIN)))
 		take_damage_bone_break(brute)
 
 	if(status & LIMB_BROKEN && prob(40) && brute > 10)
@@ -1055,7 +1055,7 @@ treat_grafted var tells it to apply to grafted but unsalved wounds, for burn kit
 			return FALSE
 
 /obj/limb/proc/fracture(bonebreak_probability)
-	if(status & (LIMB_BROKEN|LIMB_DESTROYED|LIMB_ROBOT|LIMB_SYNTHSKIN))
+	if(status & (LIMB_BROKEN|LIMB_DESTROYED|LIMB_UNCALIBRATED_PROSTHETIC|LIMB_SYNTHSKIN))
 		if (knitting_time != -1)
 			knitting_time = -1
 			to_chat(owner, SPAN_WARNING("You feel your [display_name] stop knitting together as it absorbs damage!"))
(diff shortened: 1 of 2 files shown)


code/modules/organs/limbs.dm
```diff
@@ -1083,17 +1083,30 @@ treat_grafted var tells it to apply to grafted but unsalved wounds, for burn kit
...
-			SPAN_WARNING("You hear a loud cracking sound coming from [owner]!"),
-			SPAN_HIGHDANGER("Something feels like it shattered in your [display_name]!"),
-			SPAN_HIGHDANGER("You hear a sickening crack!"))
-		playsound(owner, "bone_break", 45, TRUE)
+		if(status & (LIMB_ROBOT))
+			owner.visible_message(\
+				SPAN_WARNING("You see sparks coming from [owner]'s [display_name]!"),
+				SPAN_HIGHDANGER("Something feels like it broke in your [display_name] as it spits out sparks!"),
+				SPAN_HIGHDANGER("You hear electrical sparking!"))
+			var/datum/effect_system/spark_spread/spark_system = new /datum/effect_system/spark_spread()
```
harryob, line 1091:
```suggestion
			var/datum/effect_system/spark_spread/spark_system = new()
```

code/modules/organs/limbs.dm
```diff
@@ -1083,17 +1083,30 @@ treat_grafted var tells it to apply to grafted but unsalved wounds, for burn kit
...
+				SPAN_HIGHDANGER("Something feels like it shattered in your [display_name]!"),
+				SPAN_HIGHDANGER("You hear a sickening crack!"))
+			playsound(owner, "bone_break", 45, TRUE)
 		start_processing()
-
-		status |= LIMB_BROKEN
-		owner.pain.apply_pain(PAIN_BONE_BREAK)
-		broken_description = pick("broken","fracture","hairline fracture")
-		perma_injury = min_broken_damage
+		if(status & (LIMB_ROBOT))
```
harryob, line 1103:
```suggestion
		if(status & LIMB_ROBOT)
```

harryob (changes requested):

Diegoflores31:
dont robotic limbs already "break" after receiving 10 damage ? also this could potentially break Repairing Trait .

NewyearnewmeUwu:
It still repairs, but it won't recalibrate it.

morrowwolf (changes requested):
Having a non-functional arm should also disable the hand. Same deal with leg/feet.

A new surgery at level 1 (so corpsman can do it) will need to be made for just calibrating a limb as they currently cannot do it.

PR #3069, opened 2023-04-14
overhauls logging

harryob:
round id for logs and stuff

:cl:
server: GLOB.round_id is now available for an incrementing id of rounds
/:cl:


blackcrystall:
make entity load priority high, and map value change when mapping loading ground map + .sync() use on round, for safety proposes

morrowwolf (approved):

PR #3043, opened 2023-04-10
Fixes the contents showing under crates/barrels when you disassemble them

Hopekz:
I hate this bug so much I don't know why it wasn't fixed until now it makes maintenance looting so much more of a hastle than it needs to be (you need to right click on everything you disassemble or alt click the turf to see the stack? Who thought this was a good idea?)

The code takes the material of the crate and reorganizes the stack using a new proc to ensure it is on the bottom of the list meaning that the material will show up on the bottom.

dreamseeker_sm4yWDBgEP

:cl: Hopek
fix: Disassembled crates/barrels will now correctly show their contents above the deconstruction material meaning that you don't have to right click or alt click the turf to see what you looted.
/:cl:

PR #3075, opened 2023-04-15
Fixes the contents showing under crates/barrels when you disassemble them V2

Hopekz:
Fixed edition of #3043

### Original PR description:
I hate this bug so much I don't know why it wasn't fixed until now it makes maintenance/planetside looting so much more of a hastle than it needs to be (you need to right click on everything you disassemble or alt click the turf to see the stack? Who thought this was a good idea?)

The code takes the material of the crate and reorganizes the stack using a new proc to ensure it is on the bottom of the list meaning that the contents will show up on the top.

image

:cl: Hopek
fix: Disassembled crates/barrels will now correctly show their contents above the deconstruction material meaning that you don't have to right click or alt click the turf to see what you looted. Doesn't make the contents break reality this time.
fix: Fixed duplicate crate/barrel disassembly material spawning because it was being created in two places. If people want 2 material per crate/barrel they'll have to intentionally code it in this time vs it being a bug.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/largecrate.dm
+++ b/code/game/objects/structures/crates_lockers/largecrate.dm
@@ -18,16 +18,35 @@
 	return
 
 /obj/structure/largecrate/proc/unpack()
-	for(var/atom/movable/A in contents)
-		A.forceMove(loc)
+	var/turf/current_turf = get_turf(src) // Get the turf the crate is on
+
 	playsound(src, unpacking_sound, 35)
+
+	var/obj/item/stack/sheet/material_sheet // Variable to store the reference of the crate material
+	if(parts_type) // Create the crate material and store its reference
+		material_sheet = new parts_type(current_turf, 2)
+
+	// Move the objects back to the turf, above the crate material
+	while(contents.len)
+		var/atom/movable/current_atom = contents[1]
+		current_atom.forceMove(current_turf)
+
 	deconstruct(TRUE)
 
+	// Move the crate material to the bottom of the turf's contents
+	if(material_sheet)
+		move_to_bottom(material_sheet, current_turf)
+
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/crates_lockers/largecrate.dm
```diff
@@ -18,16 +18,36 @@
...
+
 	playsound(src, unpacking_sound, 35)
+
+	/// Store the reference of the crate material
+	var/obj/item/stack/sheet/material_sheet
+	if(parts_type) // Create the crate material and store its reference
+		material_sheet = new parts_type(current_turf, 2)
+
+	// Move the objects back to the turf, above the crate material
+	while(contents.len)
```
harryob, line 31: why this over for(var/atom/movable/moving_atom in contents)?

Hopekz: > why this over for(var/atom/movable/moving_atom in contents)?

Just what I decided to go with while I was typing it up.
In my mind there might not be contents to process or there might be a lot of contents so I just put that in a while.

harryob: > > why this over for(var/atom/movable/moving_atom in contents)?
> 
> Just what I decided to go with while I was typing it up. In my mind there might not be contents to process or there might be a lot of contents so I just put that in a while.
> 
> Do I need to change it?

iterating over a list with a for loop is slightly better optimized, and is preferred in our style guide. while loops can be finicky in dm and cause the whole mc to grind to a stop. if there's a lot of contents, or no contents, a for loop will handle it just as well

Hopekz: > for(var/atom/movable/moving_atom in contents)

No worries; updated.

Issue #1884, opened 2022-12-08
When Spotter Designator broke while zoomed in, vision remains zoomed. Forever.


rudew:
### Testmerges

#1760 #1841 #955 #1707

### Description of the bug

I got CAS'ed, which broke my Spotters Designator, which was in my left hand at the time. After which, regardless of looking through new binoculars, blinding myself, or ghosting and being put back into my body, my vision remained zoomed. Whenever I was ghosted, however, the vision was not zoomed, meaning it was probably attached to the body itself.

### What's the difference with what should have happened?

My binoculars breaking should not force me into being forever zoomed.

### How do we reproduce this bug?

1. Hold binoculars, maybe specifically the spotters variety.
2. Have them break while you are holding them, maybe specifically in the left hand, maybe specifically from CAS, but likely any explosion.
3. Voila, I guess?

### Issue Bingo

- [ ] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

ItsVyzo:
Probably fixed in #3079

ItsVyzo:
Not fixed

ItsVyzo:
Duplicate #266

PR #3079, opened 2023-04-15
Misc bino fixes

### LINKED ISSUE #372: Laser Designator Issues

BeagleGaming1:
# About the pull request

Binoculars now properly disable on death and being dropped

fixes #372 
(fixes potentially #1884, am unsure)

# Changelog

:cl:
fix: Removes some ways of using binos when you should not be able to
/:cl:

### DIFF
--- a/code/game/objects/items/devices/binoculars.dm
+++ b/code/game/objects/items/devices/binoculars.dm
@@ -106,7 +106,7 @@
 		return
 	if(mods["ctrl"])
 		if(user.stat != CONSCIOUS)
-			to_chat(user,SPAN_WARNING("You cannot use [src] while incapacitated."))
+			to_chat(user, SPAN_WARNING("You cannot use [src] while incapacitated."))
 			return FALSE
 		if(SEND_SIGNAL(user, COMSIG_BINOCULAR_HANDLE_CLICK, src))
 			return FALSE


code/game/objects/items/devices/binoculars.dm
```diff
@@ -97,27 +101,30 @@
 		return TRUE
 	return ..()
 
-/obj/item/device/binoculars/range/handle_click(mob/living/carbon/human/user, atom/A, list/mods)
+/obj/item/device/binoculars/range/handle_click(mob/living/carbon/human/user, atom/targeted_atom, list/mods)
 	if(!istype(user))
 		return
 	if(mods["ctrl"])
+		if(user.stat != CONSCIOUS)
+			to_chat(user,SPAN_WARNING("You cannot use [src] while incapacitated."))
```
harryob, line 109:
```suggestion
			to_chat(user, SPAN_WARNING("You cannot use [src] while incapacitated."))
```

harryob (changes requested):

PR #3086, opened 2023-04-16
Reflective walls no longer care about IFF

Hopekz:
Why:
IFF isn't in the bullet but the gun sighting system.
The wall redirects the bullet back; therefore the bullet would no longer be IFF sighted.

:cl: Hopek
balance: The resin reflective walls no longer recognize IFF bullets and send them back as if they were regular bullets because an alien wall shouldn't know the difference because it is just a wall.
/:cl:

### DIFF
--- a/code/game/turfs/walls/wall_types.dm
+++ b/code/game/turfs/walls/wall_types.dm
@@ -1096,7 +1096,7 @@
 			P.damage *= brute_multiplier
 		return ..()
 	if(P.ammo.flags_ammo_behavior & AMMO_NO_DEFLECT)
-		// Bullet gets absorbed if it has IFF or can't be reflected.
+		// Bullet gets absorbed if it can't be reflected.
 		return
 
 	var/obj/item/projectile/new_proj = new(src, construction_data ? construction_data : create_cause_data(initial(name)))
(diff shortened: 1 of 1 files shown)


blackcrystall:
🏅 
a good one, but better to use same projectile for return fire, clear it permutated and redirect somewhere back, with removed iff group (i don't think if on hpop 4 smartgunner with automatic fire abuse shooting like 10 bullets per every smart per second don't kill server, mathematic, 40 bullets * 2... server cries in the corner, garbage controller su in other room corner)

Nanu308:
The smartgunners are normally not the one in the front, those that are punished for the SG's mistakes are regular marines.

This is why it was done as it is currently and why i do not see a reason to change it, unless @morrowwolf disagrees i will most likely close this.

Hopekz:
> The smartgunners are normally not the one in the front, those that are punished for the SG's mistakes are regular marines.
> 
> This is why it was done as it is currently and why i do not see a reason to change it, unless @morrowwolf disagrees i will most likely close this.

I actually built this PR because at the moment there isn't really a counter play to the "murderball" strategy of around 3-5 SG's together (depending if the CO joins them) but on testing realized that this isn't too effective in terms of countering the smart gunners themselves but you are right their team would most likely take most of the fire.

I think that part is intentional though to descentivize shooting at these walls but at the same time I can see your point about it potentially being frustrating for the marine team to deal with projectiles.

In my vision of how the game is played a smart gunner doesn't blindly shoot at the hive when they notice their bullets are being reflected back at them which I think might be what these walls are designed to do.

No worries though if we see it differently because everyone has their own vision of what this project should be.

PR #3089, opened 2023-04-17
Resin Whisperers can now interact with resin doors with Coerce Resin

Hopekz:
Resin Whisperers can now remotely interact with resin doors.
Just press the middle mouse button on them with the "Coerce Resin" mode.

dreamseeker_eDGE5ycfNT

:cl: Hopek
add: Resin Whisperers can now use Coerce Resin to open and close resin doors.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/mutators/strains/hivelord/resin_whisperer.dm
+++ b/code/modules/mob/living/carbon/xenomorph/mutators/strains/hivelord/resin_whisperer.dm
@@ -77,6 +77,16 @@
 	if(!target_turf)
 		return
 
+	/// Check if the target is a resin door and open or close it
+	if(istype(target_atom, /obj/structure/mineral_door/resin))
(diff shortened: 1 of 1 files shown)


harryob (approved):
very neat

victorjosephespinoza:
cool addition, however, one minor thing.

You should check the type of the hive. 

Otherwise other hives will be able to spam clicks to get the following message:
image

Hopekz:
> cool addition, however, one minor thing.
> 
> You should check the type of the hive.
> 
> Otherwise other hives will be able to spam clicks to get the following message: image



> cool addition, however, one minor thing.
> 
> You should check the type of the hive.

TryToSwitchState(owner) handles this so they can only open resin doors on hives they are a part of.

> Otherwise other hives will be able to spam clicks to get the following message: image

The message is sent directly to the user that is opening the doors so in my opinion if they want to spam their own game they should be free to do so.
If I get extra time I am going to implement this but I find this a non-issue.

PR #3090, opened 2023-04-17
Fix a start now error and add the ability of queuing the start of the game

Hopekz:
This PR does two things.

Fixes this error when trying to start early
dreamseeker_lIUnkd0lFZ

And adds queuing meaning that if an admin wants to start a game early during loading; it will now tell them that the game will launch as soon as it is available then waits for the game to be ready before starting.

Before this PR it just tells you that the game isn't ready then you have to wait for it to load and launch the "start now" command again.

Does not bypass the "are you sure?" check because it has been moved to the front.

Honestly made this PR because I hate waiting for the start I just want to do it once when I see the game window then step away for like a minute instead of having to wait for it.

:cl: Hopek
add: Adds the support for queuing the round start meaning that if an admin pressed "start now" it will actually wait until the game is loaded then immediately start the game as expected versus telling you to try later.
fix: fixed the "start now" verb displaying that the game has already started when it is loading because it didn't understand how to read the game state properly.
/:cl:

### DIFF
--- a/code/modules/admin/tabs/round_tab.dm
+++ b/code/modules/admin/tabs/round_tab.dm
@@ -165,12 +165,13 @@
 	set name = "Start Round"
 	set desc = "Start the round RIGHT NOW"
 	set category = "Server.Round"
-
-	if (!SSticker)
-		alert("Unable to start the game as it is not set up.")
-		return
 	if (alert("Are you sure you want to start the round early?",,"Yes","No") != "Yes")
 		return
+	if (SSticker.current_state == GAME_STATE_STARTUP)
+		alert("Game is setting up and will launch as soon as it is ready.")
+		message_admins(SPAN_BLUE("[usr.key] has started the process to start the game when loading is finished."))
+		while (SSticker.current_state == GAME_STATE_STARTUP)
+			sleep(50) // it patiently waits for the game to be ready here before moving on.
 	if (SSticker.current_state == GAME_STATE_PREGAME)
 		SSticker.request_start()
 		message_admins(SPAN_BLUE("[usr.key] has started the game."))
--- a/code/modules/admin/tabs/round_tab.dm
+++ b/code/modules/admin/tabs/round_tab.dm
@@ -169,7 +169,7 @@
 		return
 	if (SSticker.current_state == GAME_STATE_STARTUP)
 		alert("Game is setting up and will launch as soon as it is ready.")
-		message_admins(SPAN_BLUE("[usr.key] has started the process to start the game when loading is finished."))
+		message_admins(SPAN_ADMINNOTICE("[usr.key] has started the process to start the game when loading is finished."))
 		while (SSticker.current_state == GAME_STATE_STARTUP)
 			sleep(50) // it patiently waits for the game to be ready here before moving on.
 	if (SSticker.current_state == GAME_STATE_PREGAME)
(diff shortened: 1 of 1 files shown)


blackcrystall:
STOP MAKING THAT SHIT LOOPS
please...
why? you still can use UNTIL (better loop), you can make that like var, and when initialization finishing game checks for that value and start round if that value true

Hopekz:
> STOP MAKING THAT SHIT LOOPS please... why? you still can use UNTIL (better loop)

Your "better" solution is the same shit except packaged up in a different flavor-text.
It doesn't actually matter if something loops "while" or "until" something.
There are appropriate times for both but in this case it doesn't matter.

In this case the "while" is appropriate here because the GAME_STATE_STARTUP always goes to GAME_STATE_PREGAME. It is expected consistent behavior because if it doesn't behave this exact way the round can't start anyway with or without the while loop.

tl;dr
While waits for GAME_STATE_PREGAME and Until waits for GAME_STATE_PREGAME; no difference.

code/modules/admin/tabs/round_tab.dm
```diff
@@ -165,12 +165,13 @@
...
 	set desc = "Start the round RIGHT NOW"
 	set category = "Server.Round"
-
-	if (!SSticker)
-		alert("Unable to start the game as it is not set up.")
-		return
 	if (alert("Are you sure you want to start the round early?",,"Yes","No") != "Yes")
 		return
+	if (SSticker.current_state == GAME_STATE_STARTUP)
+		alert("Game is setting up and will launch as soon as it is ready.")
```
harryob, line 171: given alert sleeps until it's pressed, do we actually want this, given you can't cancel *out* of an alert anyway?

code/modules/admin/tabs/round_tab.dm
```diff
@@ -165,12 +165,13 @@
...
+		message_admins(SPAN_BLUE("[usr.key] has started the process to start the game when loading is finished."))
```
harryob, line 172:
```suggestion
		message_admins(SPAN_ADMINNOTICE("[usr.key] has started the process to start the game when loading is finished."))
```
for the sake of tgchat's filtering system

code/modules/admin/tabs/round_tab.dm
```diff
@@ -165,12 +165,13 @@
...
+		while (SSticker.current_state == GAME_STATE_STARTUP)
```
harryob, line 173: alternatively, you could register a signal to COMSIG_GLOB_MODE_PREGAME_LOBBY which fires when the ticker moves to GAME_STATE_PREGAME to then SSticker.request_start

but you don't really gotta

harryob (approved):
this is cool

harryob (changes requested):
those top two are actual comments, hit the wrong button

code/modules/admin/tabs/round_tab.dm
Hopekz: Modified

code/modules/admin/tabs/round_tab.dm
Hopekz: I am not going to do this unless I have to because I am now committed to make the "DON'T USE WHILE" people mad

harryob (approved):

PR #3091, opened 2023-04-17
Fixes PMC (and others) backup SMG from being put back in the holster once drawn

Hopekz:
With the gun drawn the SMG becomes too large to put back in the holster so you can take it out but not have a place to put it back.

This bug occurred to me when I spawned in as the sniper PMC and when I drew my SMG I had no way to put it back except until after I took the stock off or collapsed it.

:cl: Hopek
fix: The m39 elite variant sub machine gun can now be put back in your holster if you spawn with it as a PMC (and whoever else uses them)
/:cl:


Hopekz:
> Why not give it the `/obj/item/attachable/stock/smg/collapsible` instead?

Yeah sure why I not I just thought it looked more like an SMG without it

harryob (approved):

PR #3094, opened 2023-04-18
Resin whisperer rework

Hopekz:
Continuation of #3089

This rework is mainly focused on moving the Resin Whisperer to a more focused role of pushing the front-lines with his team.
This needs a test merge to see if people like it.

Here is a summary of the major changes:
1. The Resin Whisperer can no longer use the Coerce Resin ability when not standing on weeds.
2. The Resin Whisperer can only place resin nodes on existing weeds to slowly "push" weeds outward versus just coating a large unwedded area on their own. **(I promise this is leading somewhere and isn't just nerfs bear with me please)**
3. The Resin Whisperer whisperer can no longer plant weeds normally; the ability has been removed **(please just a little while longer)**
4. The plant weeds ability for the Resin Whisperer has been moved to a construction in the Resin structure tab. This means that it must now be built using Coerce Resin. This means that the Resin whisperer can now remotely build resin nodes. This requires them to channel the construction as per normal and costs substantially more than it did previously.
5. There is a warning displayed to the marines when the Resin Whisperer is going through the channeling process to build a Resin node on the exact tile the resin node will appear.
6. The resin whisperer can "refresh" a resin node meaning that they can build another resin node on top of an existing one to restart the spreading process if the node stopped growing and there is spaces to fill (this is how you get past doors). Refreshing a node resets it back to low health to grow over time.

**Keep in mind that these resin notes start off weak as per the usual mechanic and gain HP over time to the usual amount we are used to so a marine can literally walk up to them to stab them in a single hit to remove them.** We normally don't see this in gameplay because there is usually an alien right on the nodes being placed so a marine doesn't get a chance to attack the node quickly after it being placed.

What will possibly be changed depending on how the test merge plays out:
- Currently the resin node cost for the Whisperer is 2x the cost of a thick membrane. I can see this changing to 3x if they're able to place too many of them.
- Longer casting time for the channeling of the resin nodes
- Axing the whole rework entirely because the community will be outraged and send bitcoin assassins to my home

dreamseeker_ugCZ52AeUN

"But how do I spread past a door when resin nodes stop spreading over time?"
Just refresh a node like in this example below. Keep in mind that the node starts will low HP again meaning it'll be 1 hit away from death

dreamseeker_jOi3IveIu7

Another GIF because why not:
dreamseeker_qLSICX4u7Y

# Changelog

:cl: Hopek
add: Resin Whisperer has been reworked. Major changes include being able to place resin nodes remotely on weeds, not being able to use coerce resin when standing outside of weeds and losing the ability to regularly plant weeds outside of current existing weeds. This changes the Resin whisperer to a new "slow crawl" frontline advance gameplay where he assists his team in pushing the front lines forward as he is able to both advance weeds and build structures on them.
/:cl:


code/modules/mob/living/carbon/xenomorph/castes/Hivelord.dm
```diff
@@ -115,3 +115,10 @@
 	toggle_resin_walker()
 	to_chat(bound_xeno, SPAN_WARNING("You feel dizzy as the world slows down."))
 	bound_xeno.recalculate_move_delay = TRUE
+
+/// This check mainly exists because of the new resin node ability for resin whisperer.
+/mob/living/carbon/xenomorph/hivelord/proc/on_weeds()
+    var/turf/T = get_turf(src)
+    if(locate(/obj/effect/alien/weeds) in T)
+        return TRUE
+    return FALSE
```
harryob, line 124: this is using space indentation which linters hate

harryob (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/Hivelord.dm
Hopekz: Thank you!

harryob (approved):
code is fine

code/modules/mob/living/carbon/xenomorph/mutators/strains/hivelord/resin_whisperer.dm
```diff
@@ -67,6 +68,11 @@
 	action_type = XENO_ACTION_CLICK
 
 /datum/action/xeno_action/activable/secrete_resin/remote/use_ability(atom/target_atom, mods)
+	var/mob/living/carbon/xenomorph/hivelord/hivelord_mob = owner
+	if(!hivelord_mob.on_weeds())
```
Segrain, line 72: image
This ability is also being used by the queen, so trying to treat her as a hivelord breaks her ability to build from ovipositor completely.

Segrain (changes requested):

code/modules/mob/living/carbon/xenomorph/mutators/strains/hivelord/resin_whisperer.dm
```diff
@@ -67,6 +68,11 @@
 	action_type = XENO_ACTION_CLICK
 
 /datum/action/xeno_action/activable/secrete_resin/remote/use_ability(atom/target_atom, mods)
+	var/mob/living/carbon/xenomorph/hivelord/hivelord_mob = owner
```
harryob, line 71: this needs to check if the mob is a hivelord first

harryob (changes requested):

code/modules/mob/living/carbon/xenomorph/mutators/strains/hivelord/resin_whisperer.dm
Hopekz: > This ability is also being used by the queen, so trying to treat her as a hivelord breaks her ability to build from ovipositor completely.

My bad I didn't realize the queen used the same ability.
I went ahead and added a check for it.

code/modules/mob/living/carbon/xenomorph/mutators/strains/hivelord/resin_whisperer.dm
```diff
@@ -67,6 +68,13 @@
 	action_type = XENO_ACTION_CLICK
 
 /datum/action/xeno_action/activable/secrete_resin/remote/use_ability(atom/target_atom, mods)
+	var/mob/living/carbon/xenomorph/xeno_owner = owner
+	if(xeno_owner.mutation_type == HIVELORD_RESIN_WHISPERER)
+		var/mob/living/carbon/xenomorph/hivelord/hivelord_mob = owner
+		if(hivelord_mob.mutation_type == HIVELORD_RESIN_WHISPERER &&  !hivelord_mob.on_weeds()) // There is a chance that queen can't place down buildings in ovi build view so we place the rein whisperer check here.
```
Segrain, line 74:
```suggestion
		if(!hivelord_mob.on_weeds()) // There is a chance that queen can't place down buildings in ovi build view so we place the rein whisperer check here.
```

morrowwolf (approved):
Delay on building needs to match the time it takes to build. Building more than one structure at a time is problematic.

PR #3096, opened 2023-04-19
Maintenance Jack Tweaks

BeagleGaming1:
# About the pull request

- Adds prying flooring
- Gives more variables
- Breaks everything
- Adds wrench surgery

Sprites by sleepynecrons

These are changes I would have included in the original PR if I had received the feedback about it, 

# Changelog

:cl:
balance: Maintenance Jack can pry flooring and do wrench surgery
/:cl:


Nanu308:
So it would be a all-in-one tool and also a useful weapon?

Sorry but no, that is extreme powercreep and even if its a synth tool too strong.

Waseemq1235 (commented):
imo just make it an all-in-one tool but keep it the same size and make it deal the damage of a wrench. seems fair enough to me.

dont know if nanu agrees though D:

Katskan:
Going to weigh in because I think a "balance" call is being made here without applicable experience, and our other forums for feedback are gone. 

There is currently a very entrenched meta in synthetic tools whereby the baton is king, and mesons a close second. The baton is our focus here. There are a scarce few options for replacing the baton, the only really viable choice being a machete or B5 hammer. If you keep in mind that the original maintenance jack damage is the same as a machete, which is significantly weaker than the baton or the B5 hammer, and that it has greater attack delay and lower speed, it is the fourth strongest melee option for a shipside synthetic to carry. If you want to make it have lower damage, it will not change the meta picks at all. If you leave it as it was originally proposed on par with a machete, it has real potential to de-throne the baton by giving a viable pick that trades huge damage for more utility and soul. 

Also wrench damage is straight up a bad idea, synthetics punch with more damage than they hit things with tools for. At minimum boot knife damage would be appropriate.

code/game/machinery/doors/airlock.dm
```diff
@@ -551,6 +551,9 @@ GLOBAL_LIST_INIT(airlock_wire_descriptions, list(
 	update_icon()
 
 /obj/structure/machinery/door/airlock/attackby(obj/item/C, mob/user)
+	if(istype(C, /obj/item/weapon/maintenance_jack)) //Its got custom code but some parts of the attackby break it
```
Zonespace27, line 554: This is what signals exist for

code/game/objects/items/tools/maintenance_tools.dm
```diff
@@ -535,72 +546,120 @@
...
+		if(!airlock.locked)
+			to_chat(user, SPAN_NOTICE("You cannot disable bolts on a door that is already unbolted."))
 			return
-		if(!resin_door.density || user.action_busy || user.a_intent == INTENT_HARM)
+
+		if(requires_skills_unbolt)
+			if(!skillcheck(user, SKILL_ENGINEER, SKILL_ENGINEER_MASTER)) //Engi 3 is much faster
+				user.visible_message(SPAN_DANGER("[user] begins to search for [airlock]'s bolts!"),\
+				SPAN_NOTICE("You search for [airlock]'s bolts."))
+				if(!do_after(user, unskilled_unbolt_time SECONDS, INTERRUPT_ALL, BUSY_ICON_HOSTILE, src, INTERRUPT_ALL)) //Otherwise it takes an extra 15 seconds
```
Zonespace27, line 606: Move the second defines to the variables

Zonespace27 (changes requested):
Not a full review

Nanu308:
Lower dmg to little below bootknife and keep the size so it requires a spot

FransFeiffer (changes requested):
I know it's nitpicky, but the handle outline and the bit at the top of the jack use (1,1,1,), please change it to something like the below picture. Otherwise the sprites are fire.
image

FransFeiffer (approved):
ok we're good to go 👍

code/game/objects/items/tools/maintenance_tools.dm
```diff
@@ -535,72 +546,120 @@
...
+			if(flooring.hull_floor) //no interaction for hulls
+				return
+			if(flooring.weeds)
+				return attackby(src, user)
+
+			to_chat(user, SPAN_WARNING("You forcefully pry off [flooring], destroying it in the process."))
+			playsound(src, 'sound/items/Crowbar.ogg', 25, 1)
+			flooring.make_plating()
+			return
+	. = ..()
```
harryob, line 662:
```suggestion
	return ..()
```

code/game/machinery/doors/airlock.dm
```diff
@@ -551,6 +551,9 @@ GLOBAL_LIST_INIT(airlock_wire_descriptions, list(
 	update_icon()
 
 /obj/structure/machinery/door/airlock/attackby(obj/item/C, mob/user)
+	if(istype(C, /obj/item/maintenance_jack)) //Its got custom code but some parts of the attackby break it
+		return
+
```
harryob, line 556: pseudocode but you can put this in airlock/attackby

```dm
if(SEND_SIGNAL(C, COMSIG_ITEM_ATTACK_AIRLOCK, src) & COMSIG_ITEM_CANCEL_AIRLOCK_ATTACK)
 return
 ```
 
then you can...

 ```dm
/obj/item/maintenance_jack/Initialize()
RegisterSignal(src, COMSIG_ITEM_ATTACK_AIRLOCK, PROC_REF(handle_airlock_attack))

/obj/item/maintenance_jack/proc/handle_airlock_attack(source, obj/structure/machinery/door/airlock/attacked_door)
    // you can do all your airlock specific code here
    return COMSIG_ITEM_CANCEL_AIRLOCK_ATTACK
```

harryob (changes requested):

Nanu308:
No mapping file being touched in latest commit?

BeagleGaming1:
> No mapping file being touched in latest commit?

Nope

PR #3101, opened 2023-04-19
no usr in ui act

harryob:
this is flaky as ui_act is not guaranteed to be called by a verb, as we could queue it up later. using ui.user ensures we actually get a reference

no player facing changes

### DIFF
--- a/code/controllers/subsystem/shuttles.dm
+++ b/code/controllers/subsystem/shuttles.dm
@@ -562,11 +562,11 @@ SUBSYSTEM_DEF(shuttle)
 
 	return data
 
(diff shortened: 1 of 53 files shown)


Drulikar:
Patch of the previous changes if you prefer; but you will want to merge with master and resolve conflicts first: 
[patch.diff.txt](https://github.com/cmss13-devs/cmss13/files/11427435/patch.diff.txt)

Watermelon914:
This PR has been inactive for some time, feel free to re-open (as a new PR so that it appears at the top of the tracker) when you decide to work on this again.

PR #3102, opened 2023-04-19
queen locking the ds doors should disable the dropship being ob'd

harryob:
looks like this bit of code got fuddled around during tgshuttling the dropships

:cl:
fix: obs can no longer be fired after the dropship doors are overridden by the queen
/:cl:


Diegoflores31:
> looks like this bit of code got fuddled around during tgshuttling the dropships
> 
> 🆑 fix: obs can no longer be fired after the dropship doors are overridden by the queen /🆑

shouldnt that be when the queen Locks the DS ? because i think queen can override the doors from outside

morrowwolf (approved):

PR #3115, opened 2023-04-20
Fix huggers unable to hug unconscious people

TeDGamer:
# About the pull request

Facehuggers can now pounce hug unconscious people.

This will need to be test merged as I am editing the pounced_mob, which will likely affect runners and lurkers. 

It doesn't seem to have any side effects on runner when I tried this but never hurts to test it. I think it freezes lurkers now when they pounce unconscious people as if they pounced a conscious person. That could be looked into.

It's a weird bug that if someone is sleeping, they can't be pounce hugged so you need to use the longer attach method. They're asleep for crying out loud! That's a coup de grâce waiting to happen.

Here's a good way to replicate the bug with the current non-fixed code:
1. Spawn a human
2. Toggle sleeping the human ( context menu -> toggle sleeping )
3. Spawn as playable hugger
4. Hug the unconscious human

This is a perfectly healthy sleeping human. Huggers should be able to pounce them.

All it does is cause them to leap at the human and not hug them. 

# Changelog

:cl:
fix: Pouncing unconscious people will no longer stop the pounce effects ( such as hugging )
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/XenoProcs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/XenoProcs.dm
@@ -273,7 +273,7 @@
 		return
 
 	var/mob/living/carbon/M = L
-	if(M.stat || M.mob_size >= MOB_SIZE_BIG || can_not_harm(L) || M == src)
+	if(M.stat == DEAD || M.mob_size >= MOB_SIZE_BIG || can_not_harm(L) || M == src)
 		throwing = FALSE
 		return
 


realforest2001:
The issue I see with this is it will allow lurkers and runners the ability to pounce people who are stunned, and meaning they will never be able to stand again.

TeDGamer:
> The issue I see with this is it will allow lurkers and runners the ability to pounce people who are stunned, and meaning they will never be able to stand again.

I believe stunned is different from unconscious so that should not affect them.

M.stat only has CONSCIOUS, UNCONSCIOUS and DEAD. Stunned people should be CONSCIOUS but they have the weaken or stunned effect which I don't think affects stat.

morrowwolf:
I believe tedgamer is correct. We can TM it just in case of any weirdness but this looks fine on a quick look.

morrowwolf (approved):

PR #3119, opened 2023-04-21
Medical vendors limited supplies

TeDGamer:
# About the pull request

An attempt to remove infinite medical supply restocking.

Morrow told me not to revive Stan's attempt with the supply chain working on the ship but not groundside, but to implement it like this. 

Will need to be tested and values adjusted ( such as vendor prices) to see what fits the gameplay loop.

Medical attrition and preventing infinite medical supplies

# Changelog

:cl:
balance: All medical vendors can no longer restock or refill.
add: Medical vendors are now purchasable by requisitions.
/:cl:

### DIFF
--- a/code/game/machinery/vending/cm_vending.dm
+++ b/code/game/machinery/vending/cm_vending.dm
@@ -854,6 +854,10 @@ GLOBAL_LIST_EMPTY(vending_products)
 		stock(I, user)
 
 /obj/structure/machinery/cm_vending/sorted/proc/stock(obj/item/item_to_stock, mob/user)
+	if (istype(src, /obj/structure/machinery/cm_vending/sorted/medical))
+		to_chat(user, SPAN_WARNING("[src] cannot be restocked!"))
+		return
+
 	var/list/R
 	var/list/stock_listed_products = get_listed_products(user)
 	for(R in (stock_listed_products))
--- a/code/game/machinery/vending/cm_vending.dm
+++ b/code/game/machinery/vending/cm_vending.dm
@@ -854,10 +854,6 @@ GLOBAL_LIST_EMPTY(vending_products)
 		stock(I, user)
 
 /obj/structure/machinery/cm_vending/sorted/proc/stock(obj/item/item_to_stock, mob/user)
-	if (istype(src, /obj/structure/machinery/cm_vending/sorted/medical))
-		to_chat(user, SPAN_WARNING("[src] cannot be restocked!"))
-		return
-
 	var/list/R
 	var/list/stock_listed_products = get_listed_products(user)
 	for(R in (stock_listed_products))
--- a/code/game/machinery/vending/vendor_types/medical.dm
+++ b/code/game/machinery/vending/vendor_types/medical.dm
@@ -177,10 +177,6 @@
 		list("Syringe", round(scale * 7), /obj/item/reagent_container/syringe, VENDOR_ITEM_REGULAR)
 	)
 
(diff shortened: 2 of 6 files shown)


morrowwolf:
Will TM this Monday. Remind me if I forget.

code/game/machinery/vending/cm_vending.dm
```diff
@@ -854,6 +854,10 @@ GLOBAL_LIST_EMPTY(vending_products)
 		stock(I, user)
 
 /obj/structure/machinery/cm_vending/sorted/proc/stock(obj/item/item_to_stock, mob/user)
+	if (istype(src, /obj/structure/machinery/cm_vending/sorted/medical))
+		to_chat(user, SPAN_WARNING("[src] cannot be restocked!"))
+		return
+
```
realforest2001, line 860: `
/obj/structure/machinery/cm_vending/sorted/medical/stock(obj/item/item_to_stock, mob/user)
		to_chat(user, SPAN_WARNING("[src] cannot be restocked!"))
		return
`
Don't typecheck on the root object, just put the /stock(obj/item/item_to_stock, mob/user) proc onto /obj/structure/machinery/cm_vending/sorted/medical like above, next to wherever /obj/structure/machinery/cm_vending/sorted/medical is defined.

code/game/machinery/vending/vendor_types/medical.dm
```diff
@@ -17,40 +17,8 @@
...
-		)
-	var/list/stack_refill = list(
-		/obj/item/stack/medical/advanced/ointment,
-		/obj/item/stack/medical/advanced/bruise_pack,
-		/obj/item/stack/medical/ointment,
-		/obj/item/stack/medical/bruise_pack,
-		/obj/item/stack/medical/splint
-		)
+	var/list/chem_refill = null
+	var/list/stack_refill = null
```
realforest2001, line 21: If these vars aren't being used anywhere else should be fine to remove them entirely.

realforest2001 (changes requested):

code/game/machinery/vending/cm_vending.dm
TeDGamer: Yeah, I was having issues overriding the proc because I didn't realize I needed to remove proc in the child definition. Should be fixed.

code/game/machinery/vending/vendor_types/medical.dm
TeDGamer: I want to see how it plays out first and if we decide after the test merge we want to keep it this way, I'll remove these variables and all references from then on.

morrowwolf (commented):
Alright so this is being TM'd next round.

Just to head off concerns:
I am aware people use chem refills for the chem line and making chems. Increasing the totals available is under discussion or just keeping the base refill. For now, it'll just be a buffer for the early game rush.

The req prices are not final. They will likely be tweaked again.

It will take time for req and meta to adjust to changes. This will be tested and watched for that adjustment. Any significant balance offsets due to this will be addressed in time.

Design feedback goes here: https://forum.cm-ss13.com/w/pr-feedback/steps/step_1

Any design feedback posted on github is going to be deleted 👍

morrowwolf:
We're going to go a different direction with this.

I'd like each vendor to have a backend "real" amount of "chem juice" and various supplies that are stackable so we can stop the infinite resupply with one vendor. "Reloads" of medvendors will be available at req and just waiting on sprites for that.

This should stop the infinite supply while also maintaining ease of use. Chem dispensers will likely have a large chemical "reserve" to allow for it to be usable still.

PR #3126, opened 2023-04-22
Sniper / Spotter Cooldowns are handled in check_can_use

TeDGamer:
# About the pull request

Aimed shot and spotting should cooldown after they're actually used.

For the sniper, the can_use_check resets the cooldown based on the world time after aiming if the target moves into cover. By that logic, the sniper's cooldown should apply AFTER the shot takes effect or just as the shot is about to take effect.

The same logic applies to spotters.

The cooldown is based on the world time after aiming. The first cooldown add remains so we don't have a double aim similar to defender double headbutt issues. The cooldown is calculated at the world time after aiming so the first one doesn't increase the delay.

Refactored it so that can check use handles the cooldown proccing. So if you failed to use it ( like when the xenos move behind cover), it does its normal half cooldown. But if you can use it, procs the cooldown start. 

Cooldown should always be just as your about to take your action. Aiming time should not be factored in or else sniper basically continuously chain aim shots without factoring in the aim time and distance in the cooldown. 

How it's currently done:

Start aimed shot
Check if you can use
Start cooldown
Aiming time
Check if you can still use it and target not behind cover ( if they are, cooldown is half aimed shot delay time )
Fire!

This makes it so cooldown would actually be the aimed shot delay time - aiming time.  If aiming was as long as cooldown, the sniper could continuously keep sniping one after another with aimed shot.

Instead I made it so:
Start aimed shot
Check if you can use
Add cooldown to know that you are currently using the ability
Aiming time
Check if you can still use it and target not behind cover ( if they are, cooldown is half aimed shot delay time )
Start cooldown ( current world time plus the delay )
Fire!

If losing focus resets cooldown, then using your focus to fire should start the cooldown. can_check_use for sniper rifle handles the cooldown now.

# Changelog

:cl:
fix: Sniper's Aimed Shot and Spotter's Target cooldown is applied consistently in can_check_use.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/binoculars.dm
+++ b/code/game/objects/items/devices/binoculars.dm
@@ -438,7 +438,6 @@
 	if(!check_can_use(target))
 		return
 
-	COOLDOWN_START(designator, spotting_cooldown, designator.spotting_cooldown_delay)
 	human.face_atom(target)
 
 	///Add a decisecond to the default 1.5 seconds for each two tiles to hit.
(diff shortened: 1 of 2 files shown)


harryob (approved):
code looks good

PR #3128, opened 2023-04-22
Huggers can now scuttle resin doors

TeDGamer:
# About the pull request

Facehuggers can scuttle through resin doors now

Scuttle code should be used. Same size as larva and exact same scuttle code. 

# Changelog

:cl:
fix: Huggers can now scuttle under resin doors
refactor: Removed duplicated scuttle code
/:cl:

### DIFF
--- a/code/modules/cm_aliens/XenoStructures.dm
+++ b/code/modules/cm_aliens/XenoStructures.dm
@@ -374,8 +374,9 @@
 		var/mob/living/carbon/xenomorph/X = user
 		if (X.hivenumber == hivenumber)
 			if(X.scuttle(src))
-				return
-		return ..()
+				return // don't toggle door for small xenos
+			return ..() // toggles door for big xenos
+		return // enemy hive doesn't toggle door
 	if(iscarbon(user))
 		var/mob/living/carbon/C = user
 		if (C.ally_of_hivenumber(hivenumber))
--- a/code/modules/cm_aliens/XenoStructures.dm
+++ b/code/modules/cm_aliens/XenoStructures.dm
@@ -370,10 +370,15 @@
 		return attack_hand(user)
 
 /obj/structure/mineral_door/resin/TryToSwitchState(atom/user)
-	if(islarva(user) || isfacehugger(user))
+	if(islarva(user))
 		var/mob/living/carbon/xenomorph/larva/L = user
 		if (L.hivenumber == hivenumber)
 			L.scuttle(src)
+		return
+	if(isfacehugger(user))
+		var/mob/living/carbon/xenomorph/facehugger/F = user
+		if (F.hivenumber == hivenumber)
+			F.scuttle(src)
 		return
 	if(iscarbon(user))
 		var/mob/living/carbon/C = user
--- a/code/modules/cm_aliens/XenoStructures.dm
+++ b/code/modules/cm_aliens/XenoStructures.dm
@@ -370,7 +370,7 @@
 		return attack_hand(user)
 
 /obj/structure/mineral_door/resin/TryToSwitchState(atom/user)
-	if(islarva(user))
+	if(islarva(user) || isfacehugger(user))
 		var/mob/living/carbon/xenomorph/larva/L = user
 		if (L.hivenumber == hivenumber)
 			L.scuttle(src)
--- a/code/modules/cm_aliens/XenoStructures.dm
+++ b/code/modules/cm_aliens/XenoStructures.dm
@@ -371,12 +371,12 @@
 
 /obj/structure/mineral_door/resin/TryToSwitchState(atom/user)
 	if(isxeno(user))
-		var/mob/living/carbon/xenomorph/X = user
-		if (X.hivenumber == hivenumber)
-			if(X.scuttle(src))
-				return // don't toggle door for small xenos
-			return ..() // toggles door for big xenos
-		return // enemy hive doesn't toggle door
+		var/mob/living/carbon/xenomorph/xeno_user = user
+		if (xeno_user.hivenumber != hivenumber && !xeno_user.ally_of_hivenumber(hivenumber))
+			return
+		if(xeno_user.scuttle(src))
+			return
+		return ..()
 	if(iscarbon(user))
 		var/mob/living/carbon/C = user
 		if (C.ally_of_hivenumber(hivenumber))
--- a/code/modules/cm_aliens/XenoStructures.dm
+++ b/code/modules/cm_aliens/XenoStructures.dm
@@ -373,8 +373,9 @@
 	if(isxeno(user))
 		var/mob/living/carbon/xenomorph/X = user
 		if (X.hivenumber == hivenumber)
-			X.scuttle(src)
-		return
+			if(X.scuttle(src))
+				return
+		return ..()
 	if(iscarbon(user))
 		var/mob/living/carbon/C = user
 		if (C.ally_of_hivenumber(hivenumber))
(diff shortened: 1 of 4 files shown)


code/modules/cm_aliens/XenoStructures.dm
```diff
@@ -375,6 +375,11 @@
...
 			L.scuttle(src)
 		return
+	if(isfacehugger(user))
+		var/mob/living/carbon/xenomorph/facehugger/F = user
+		if (F.hivenumber == hivenumber)
+			F.scuttle(src)
+		return
 	if(iscarbon(user))
 		var/mob/living/carbon/C = user
 		if (C.ally_of_hivenumber(hivenumber))
```
realforest2001, line 385: I'd suggest changing this proc in general so that rather than typechecking larva or facehuggers, it does a mob_size check. 
Either way, if you don't want to do that, you need to change `islarva(user)` to `islarva(user) || isfacehugger(user)` and change `var/mob/living/carbon/xenomorph/larva/L = user` to ``var/mob/living/carbon/xenomorph/crawler` to account for it no longer being a larva exclusively.
No need to write the same code out twice.

realforest2001 (changes requested):

code/modules/cm_aliens/XenoStructures.dm
TeDGamer: Thoughts on doing it like this:
```
mob/living/carbon/xenomorph/proc/scuttle
      return
```

```
  if(isxeno(user))
     var/mob/living/carbon/xenomorph/X = user
     if (user.mob_size == MOB_SIZE_SMALL && X.hivenumber == hivenumber)
           X.scuttle(src)
    return
```

There's no problem with checking mob_size but the typecasting is what I'm worried about. My first iteration was exactly just adding "|| isfacehugger(user)"

But then it typecasts it to a larva. Seems to work similarly since their scuttle code is the exact same so we could always do it like that. An issue would only occur if huggers and larva scuttle code ever differed.

TeDGamer: I refactored it. Removed the duplicate, placed it on the parent level and handled it where it only affects xenos with the proper mob size. All xenos can scuttle but it does nothing other than return false for larger xenos. The boolean value is to ensure if they could properly scuttle

code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
```diff
@@ -1080,3 +1080,22 @@
...
+				to_chat(src, SPAN_WARNING("\The [atom] prevents you from squeezing under \the [current_structure]!"))
+				return FALSE
+		// Is it an airlock?
+		if(istype(current_structure, /obj/structure/machinery/door/airlock))
+			var/obj/structure/machinery/door/airlock/current_airlock = current_structure
+			if(current_airlock.locked || current_airlock.welded) //Can't pass through airlocks that have been bolted down or welded
+				to_chat(src, SPAN_WARNING("\The [current_airlock] is locked down tight. You can't squeeze underneath!"))
+				return FALSE
+		visible_message(SPAN_WARNING("\The [src] scuttles underneath \the [current_structure]!"), \
+		SPAN_WARNING("You squeeze and scuttle underneath \the [current_structure]."), null, 5)
```
harryob, line 1098:
```suggestion
		SPAN_WARNING("You squeeze and scuttle underneath [current_structure]."), max_distance = 5)
```

code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
```diff
@@ -1080,3 +1080,22 @@
...
+			if(atom != current_structure && atom.density && atom.BlockedPassDirs(src, move_dir))
+				to_chat(src, SPAN_WARNING("\The [atom] prevents you from squeezing under \the [current_structure]!"))
+				return FALSE
+		// Is it an airlock?
+		if(istype(current_structure, /obj/structure/machinery/door/airlock))
+			var/obj/structure/machinery/door/airlock/current_airlock = current_structure
+			if(current_airlock.locked || current_airlock.welded) //Can't pass through airlocks that have been bolted down or welded
+				to_chat(src, SPAN_WARNING("\The [current_airlock] is locked down tight. You can't squeeze underneath!"))
+				return FALSE
+		visible_message(SPAN_WARNING("\The [src] scuttles underneath \the [current_structure]!"), \
```
harryob, line 1097:
```suggestion
				to_chat(src, SPAN_WARNING("[current_airlock] is locked down tight. You can't squeeze underneath!"))
				return FALSE
		visible_message(SPAN_WARNING("[src] scuttles underneath [current_structure]!"), \
```

code/modules/cm_aliens/XenoStructures.dm
```diff
@@ -370,11 +370,13 @@
...
 /obj/structure/mineral_door/resin/TryToSwitchState(atom/user)
-	if(islarva(user))
-		var/mob/living/carbon/xenomorph/larva/L = user
-		if (L.hivenumber == hivenumber)
-			L.scuttle(src)
-		return
+	if(isxeno(user))
+		var/mob/living/carbon/xenomorph/X = user
+		if (X.hivenumber == hivenumber)
+			if(X.scuttle(src))
```
harryob, line 376:
```suggestion
		var/mob/living/carbon/xenomorph/xeno_user = user
		if (xeno_user.hivenumber == hivenumber)
			if(xeno_user.scuttle(src))
```

code/modules/cm_aliens/XenoStructures.dm
```diff
@@ -370,11 +370,13 @@
...
+				return // don't toggle door for small xenos
+			return ..() // toggles door for big xenos
+		return // enemy hive doesn't toggle door
```
harryob, line 379: could use early returns here to make this more understandable

harryob (changes requested):

code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
```diff
@@ -1080,3 +1080,22 @@
 
 /mob/living/carbon/xenomorph/handle_blood_splatter(splatter_dir, duration)
 	new /obj/effect/temp_visual/dir_setting/bloodsplatter/xenosplatter(loc, splatter_dir, duration)
+
+/mob/living/carbon/xenomorph/proc/scuttle(obj/structure/current_structure)
+	if (mob_size == MOB_SIZE_SMALL) // huggers, larva
+		var/move_dir = get_dir(src, loc)
+		for(var/atom/movable/atom in get_turf(current_structure))
+			if(atom != current_structure && atom.density && atom.BlockedPassDirs(src, move_dir))
+				to_chat(src, SPAN_WARNING("\The [atom] prevents you from squeezing under \the [current_structure]!"))
```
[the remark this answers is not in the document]
TeDGamer: Removed

code/modules/cm_aliens/XenoStructures.dm
TeDGamer: Fixed var name

code/modules/cm_aliens/XenoStructures.dm
TeDGamer: Early returned for more understandablility and fixed for allies of the hive

code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
TeDGamer: Removed /the. Used named parameter

harryob (approved):

PR #3130, opened 2023-04-22
Fix predaliens being unable to attack enemy xenomorphs

Emanthealmighty:
# About the pull request

Predaliens are currently unable to damage enemy xenomorphs by slashing them, as they will instead "lunge" at them, which is what happens when the attack comes out dealing 0 damage. This is due to the predalien's custom "melee_attack_modify_damage" function from their behavior delegate, which for some reason prevented them from hitting targets bigger than humans.

The check to see if the character is a yautja (in order to deal more damage to them) was also preventing enemy xenos from being targeted for some reason. I've simplified this part of the code which in turn also fixed the issue entirely while still functioning as intended.

Now predaliens can actually be a menace to both humans and xenos.

# Changelog

:cl:
fix: Fixed predaliens being unable to attack enemy xenomorphs.
/:cl:


code/modules/mob/living/carbon/xenomorph/castes/Predalien.dm
```diff
@@ -107,10 +107,7 @@ Your health meter will not regenerate normally, so kill and die for the hive!</s
 	kills = min(kills + 1, max_kills)
 
 /datum/behavior_delegate/predalien_base/melee_attack_modify_damage(original_damage, mob/living/carbon/A)
-	if(!iscarbonsizehuman(A))
-		return
-	var/mob/living/carbon/human/H = A
-	if(isspeciesyautja(H))
+	if(isyautja(A))
```
realforest2001, line 110: No need to change this as they do essentially the same thing and are unrelated to the carbon size. isspecies is also more efficient check as it's not typechecking and instead leaves room for special circumstances.

realforest2001 (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/Predalien.dm
Emanthealmighty: As stated in the PR's description, there was a need to change this as just removing the carbon size check was for some reason not sufficient to make the predalien able to slash other xenomorphs. I've rewritten it to use "isspeciesyautja" again by making the original code only trigger if we're sure the target is a human.

code/modules/mob/living/carbon/xenomorph/castes/Predalien.dm
```diff
@@ -107,11 +107,10 @@ Your health meter will not regenerate normally, so kill and die for the hive!</s
...
 /datum/behavior_delegate/predalien_base/melee_attack_modify_damage(original_damage, mob/living/carbon/A)
-	if(!iscarbonsizehuman(A))
-		return
-	var/mob/living/carbon/human/H = A
-	if(isspeciesyautja(H))
-		original_damage *= 1.5
+	if(ishuman(A))
+		var/mob/living/carbon/human/H = A
+		if(isspeciesyautja(H))
+			original_damage *= 1.5
```
harryob, line 113:
```suggestion
		var/mob/living/carbon/human/attacked_human = A
		if(isspeciesyautja(attacked_human))
			original_damage *= 1.5
```

harryob (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/Predalien.dm
```diff
@@ -107,11 +107,10 @@ Your health meter will not regenerate normally, so kill and die for the hive!</s
 	kills = min(kills + 1, max_kills)
 
 /datum/behavior_delegate/predalien_base/melee_attack_modify_damage(original_damage, mob/living/carbon/A)
-	if(!iscarbonsizehuman(A))
-		return
-	var/mob/living/carbon/human/H = A
-	if(isspeciesyautja(H))
-		original_damage *= 1.5
+	if(ishuman(A))
```
Diegoflores31, line 110: one letter var 💀

harryob (approved):

PR #3133, opened 2023-04-23
converts portable vendor to tgui

harryob:
more tgui

:cl:
ui: the portable vendor for CLs is now tgui
/:cl:


Drulikar (changes requested):
The vendor seems functional for a CL. However, the portable_vector for a syndicate is wonky: 
image
Apply 
[patch.diff.txt](https://github.com/cmss13-devs/cmss13/files/11411841/patch.diff.txt)
to update tools.dm
image

BeagleGaming1 (approved):

PR #3137, opened 2023-04-23
Allow all xenos to clear hatched eggs regardless of their hivenumber

Emanthealmighty:
# About the pull request

Xenomorph eggs can be destroyed by aliens belonging to other hives, but they can't actually be removed once destroyed. The same goes for hatched eggs. This PR allows all xenos to clear hatched eggs regardless of their hivenumber.

By allowing xenos to clear enemy hatched eggs, we also allow them to place their own constructions/eggs over where the cleared eggs used to be, which currently cannot be done at all, making eggs the only xenomorph structure that acts this way to my knowledge for no good reason (for example, corrupted xenos inside a containment chamber almost always have at least one or two inoperable tiles because they're occupied by the eggs of the normal hive which can't be removed).

# Changelog

:cl:
add: Xenomorphs can now clear eggs from other hives
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/egg.dm
+++ b/code/modules/cm_aliens/structures/egg.dm
@@ -40,6 +40,14 @@
 		. += "Ctrl + Click egg to retrieve child into your empty hand if you can carry it."
 
 /obj/effect/alien/egg/attack_alien(mob/living/carbon/xenomorph/M)
+	if(status == EGG_BURST || status == EGG_DESTROYED)
+		M.animation_attack_on(src)
+		M.visible_message(SPAN_XENONOTICE("\The [M] clears the hatched egg."), \
+		SPAN_XENONOTICE("You clear the hatched egg."))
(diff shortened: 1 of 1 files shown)


code/modules/cm_aliens/structures/egg.dm
```diff
@@ -40,6 +40,14 @@
 		. += "Ctrl + Click egg to retrieve child into your empty hand if you can carry it."
 
 /obj/effect/alien/egg/attack_alien(mob/living/carbon/xenomorph/M)
+	if(status == EGG_BURST || status == EGG_DESTROYED)
+		M.animation_attack_on(src)
+		M.visible_message(SPAN_XENONOTICE("\The [M] clears the hatched egg."), \
```
harryob, line 45:
```suggestion
		M.visible_message(SPAN_XENONOTICE("[M] clears the hatched egg."), \
```

harryob (changes requested):

harryob (approved):

PR #3138, opened 2023-04-23
Injector availability nerf

LynxSolstice:
# About the pull request
Nerfs the availability of 45u EZ autoinjectors for the general Marine population, adds two new single-use, 15u refillable injectors.

Reduces the availability of chems for Marines that aren't Hospital Corpsmen, thus reducing the overall meta phenomenon of non-medical Marines who use the excessive abundance of kelotane and bicaridine EZ injectors to self medicate a majority of their damage on their own. (They'll have to put in more effort to get the 45u EZ injectors, even if it's just, talking to a nurse/doctor/corpsman to vend it for them.) This change should help the general movement towards a more scarce medical system that seems to be happening imo.

image
[testing video](https://youtu.be/P4GpNUwnzfM)

# Changelog

:cl:LynxSolstice
add: Adds new, single use 15u kelotane and bicaridine autoinjectors.
balance: Removed the 3x15u autoinjectors from the firstaid pouches and adds the new autoinjectors to the "weaker" med vendors.
spellcheck: lower-cases the I in inaprovaline.
code: Instead of tricordrazine being the base injector for /obj/item/reagent_container/hypospray/autoinjector/skillless/, makes a new default and adds a specific subtype for tricordrazine (/obj/item/reagent_container/hypospray/autoinjector/skillless/tricordrazine)
imageadd: Added 2 new sprites for the bicaridine and kelotane single use 15u injectors.
/:cl:


Flerpin:
don't know why this needs to be added - this is just going to make it much more frustrating on marine side as it's going to slow down gameplay (not unless that's the intent of all these changes)

i would carry two sets of injectors on me - 2x bica, 2x kelo (and maybe some others but not counting those)
Now that would be 12 whole injections of 15u injections (6injections bica and 6 injections kelo), which is a fair amount - during prolongued engagements at chokes, it's not wholly uncommon to be very low on meds and have to go to a medic or make a trek back to resupply
With this change - We're taking the 12 injections and now I would only have FOUR injections. 2 only for bica and 2 only for kelo. 
with this change, feel like there's going to be two things that are gonna happen
1. - regular PFC marine gameplay is gonna slow down (maybe not a whole lot, but probably noticeable) as marines won't be able to heal regular damage up as often
2. gonna be a whole lot more pressure on medics to fill in the gaps and there's probably going to be an unintended effect of medics being pigeonholed into carrying in as much medical gear as possible whereas they had a little bit of liberty with their loadouts thanks to PFCs being able to handle their minor injuries without much trouble

Diegoflores31:
I personally think EZ autoinjectors in general just need to be gone apart from the tricord one .

morrowwolf (commented):
I appreciate the intention of this PR but this is not how I'd like to approach this issue.

PR #3142, opened 2023-04-24
New multitool sprites

morrowwolf:
# About the pull request

Updates the multitool sprite with an animated more lore friendly version.

I may have to go through and have it animate for every action it's used for if we really want it to look decent but hopefully there's a better way I just can't think of right now. Also, getting this up now so spriters can get the review out of the way early.

EDIT: current thought is to make the sprites we have now into the colony version (as they are supposed to represent Isolation sprites) and this one in this PR would be the USCM one with a feel from the movie device. Gonna pass it by the lore team for thoughts.

Sprites that are more consistent with the world we want to create are good.

<details>

</details>

# Changelog

:cl: Morrow, Quickload
imageadd: New multitool sprites
/:cl:


Drulikar (commented):
Code works, but it is likely not animating when you would want it to. afterattack is only called if the atom clicked on is not adjacent to us and returns false for click_adjacent. For example hacking any cm_vending will not trigger the animation on the multitool. Instead wherever TRAIT_TOOL_MULTITOOL is tested in a attackby proc, we want to animate the mutlitool.

Drulikar:
This might not be fully exhaustive, but this should address most places where the multitool would likely animate.
[patch.diff.txt](https://github.com/cmss13-devs/cmss13/files/11411402/patch.diff.txt)

nauticall (changes requested):
image
only suggestion would be to make the green bits in between the lights here into like a grayish black, like the below

besides that, it looks good 👌 

image

carlarctg:
incredibly funny that even this pr by the head walance moisturizer has no acreenshots

Drulikar:
> incredibly funny that even this pr by the head walance moisturizer has no acreenshots

https://github.com/cmss13-devs/cmss13/runs/12971745871

morrowwolf:
> incredibly funny that even this pr by the head walance moisturizer has no acreenshots

Skill issue honest to god

PR #3144, opened 2023-04-24
better commit descriptions

harryob:
they're currently polluted with the template's guide comments, the stuff in the \<!--->s, we may as well get rid of this so when a pr is merged, the commit description is neater

(no player facing changes)


Zonespace27 (approved):
Looks solid

PR #3151, opened 2023-04-25
Adds a Spec to the Foxtrot Cryo Squad.

LynxSolstice:
# About the pull request
Adds a Squad Weapons Specialist to the Foxtrot ERT bought through DEFCON.

Foxtrot as it currently stands is in a poor spot, hence the meme "Foxtrot gets called, it's 3 guys, 2 of which cryo immediately", hopefully the inclusion of a Foxtrot Spec will increase the number of people signing up for it, as it is a pretty big incentive.

Compiles and runs, ERT can trigger fine, still.

# Changelog

:cl:LynxSolstice
balance: Adds a Squad Weapons Specialist to the Cryorines.
/:cl:

### DIFF
--- a/code/datums/emergency_calls/cryo_marines.dm
+++ b/code/datums/emergency_calls/cryo_marines.dm
@@ -37,6 +37,11 @@
 		arm_equipment(H, /datum/equipment_preset/uscm/leader/cryo, TRUE, TRUE)
 		to_chat(H, SPAN_ROLE_HEADER("You are a Squad Leader in the USCM"))
 		to_chat(H, SPAN_ROLE_BODY("You are here to assist in the defence of the [SSmapping.configs[GROUND_MAP].map_name]. Listen to the chain of command."))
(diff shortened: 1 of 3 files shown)


morrowwolf (commented):
We might have to discuss how sets are going to work with foxtrot specs and I'm not 100% sold on this yet but let's give it a try.

Mister-moon1:
If all spec occupations are filled then new spawns cannot pick from any options, i would advise that you allow them to spawn with /obj/item/spec_kit as this is able to go around the issue

LynxSolstice:
> If all spec occupations are filled then new spawns cannot pick from any options, i would advise that you allow them to spawn with /obj/item/spec_kit as this is able to go around the issue

Thanks for the suggested fix, it should solve the issue.

morrowwolf (approved):
GA approval.

Finally makes cryo marines worth the point cost.

PR #3152, opened 2023-04-25
Changes M2C deploy time to be the same as other turrets

Hopekz:
This is an observation I had during a round where I watched an m2c gunner use his turret instead of a gun since it had a quick deploy time.
T m2c deploy time is currently 1 second vs the 4 it usually takes for all other turrets.
This looks like an oversight to be able to use it like a normal gun with that small deployment delay

:cl: Hopek
balance: rebalanced something
/:cl:

### DIFF
--- a/code/modules/cm_marines/smartgun_mount.dm
+++ b/code/modules/cm_marines/smartgun_mount.dm
@@ -136,7 +136,7 @@
 		return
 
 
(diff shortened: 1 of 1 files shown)


morrowwolf:
M2C is meant to be a rapid deployed assault weapon.

PR #3165, opened 2023-04-26
Adds the IB check status function to the examine menu.

LynxSolstice:
# About the pull request

Adds the Internal Bleeding check to the examine menu, increasing ease of use.

Having to go through the right click menu, can at times be cumbersome, now, this allows medical personnel to see both internal and external bleeds through the examine menu.

Compiles, runs, works properly. Riflemen cannot see the IB examine text as they do not have the necessary skills for the check. 
image

# Changelog

:cl:LynxSolstice
qol: Trained personnel can now see the internal bleeding check-status verb text in the shiftclick/rightclick examine gui.
/:cl:


morrowwolf (commented):
I personally like the intentional nature of specifically checking for IB or other special factors via check status while next to a casualty. As such I will be closing this.

PR #3167, opened 2023-04-26
Runner push exploit fix

LynxSolstice:
# About the pull request
Runners can no longer push other mobs, which was being used to exploit the movement speed debuff of grabbing mobs, (one xeno drags another while a runner pushes, moving all 3 at a high velocity,) and being used to push humans farther than the runner could drag in that same time frame, as well as not needing to disarm the human first in order to do so. (Runners pushing people 10+ tiles in a few seconds while they cannot react was a common issue.)

Runners being able to use their movement speed to push other mobs such as a runner grabbing a queen or a human to get around the grab speed debuff was frankly, widespread and well known, this PR eliminates that exploit in-order to lessen the amount of cheesing runners can do with the pushing mechanic (which, with this PR, is none, as they cannot push anymore).

[Testing video](https://youtu.be/1v7NYxwTmow)

# Changelog

:cl:LynxSolstice
fix: Runners can no longer push humans or other xenos which was being unfortunately exploited to moved grabbed items without delay or to push humans very fast without grabbing them.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -352,7 +352,7 @@
 
 	if(isxeno(L) && !islarva(L))
 		var/mob/living/carbon/xenomorph/X = L
-		if(X.mob_size >= MOB_SIZE_BIG || (ishuman(src) && !isyautja(src))) // Small xenos can be pushed by other xenos or preds
+		if(X.mob_size >= MOB_SIZE_XENO_BIG || (ishuman(src) && !isyautja(src))) // Small xenos can be pushed by other xenos or preds
 			now_pushing = FALSE
 			return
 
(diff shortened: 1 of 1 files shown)


RenaRenaRe:
Based on the issues you've described, would it not be better to just increase the movement delay when a runner pushes somebody? Maybe even add a special case for pushing someone who's grabbing someone else if that's such a big issue. Not allowing runners to push anybody ever seems unnecessary.

LynxSolstice:
> Based on the issues you've described, would it not be better to just increase the movement delay when a runner pushes somebody? Maybe even add a special case for pushing someone who's grabbing someone else if that's such a big issue. Not allowing runners to push anybody ever seems unnecessary.

Please keep the design feedback contained to either the Discord (You can ping me in the CM discord with @LynxSolstice#8534) or on the Forum, (@LynxSolstice)

morrowwolf (changes requested):
This is not an exploit. This is regular game mechanics being used in an annoying way.

I don't believe this is the way to do this either.

Remake this with additional movement delay when pushing not the same faction, with a balance tag, and a write up on why this is important for game balance.

PR #3175, opened 2023-04-27
minimap perf improvements + almayer renabled

harryob:
various performance improvements to the minimap courtesy of tiviplus. also re-enabled it on the almayer because ??? why not

🆑 
qol: ghosts can now see the minimap on the almayer
/:cl:


morrowwolf (approved):
Code looks fine but I'm gonna TM it because I'm paranoid

PR #3182, opened 2023-04-29
Readds skill multipliers to several procs

nauticall:
# About the pull request

Several skill checks have been readded for procedures that had skill bonuses applied to them (such as strip/equip speed on other people and yankling out objects. People skilled in CQC, for example, now have their CQC skill bonus applied to the when stripping people, something useful for MPs and extremely useful for Synthetics on the field.

The field where the respective skill normally would've gone for these checks was left blank previously; I simply put the skills back in.

The largest part of this PR is that it finally re-adds the strip speed bonus to MPs/Synths. Other tasks, such as putting backpacks on xenos (for some reason) and removing embedded objects (not shrapnel) should be quicker for skilled medical roles as well. Unskilled roles are still unaffected; this isn't a buff across the board, after all.

# Changelog

:cl:
fix: Readded several missing skill bonuses to several procs, like stripping and removing embedded objects, that previously were left blank. Some actions will be much faster for characters that are proficient in the required skills; particularly synthetics, medical staff/corpsmen, and MPs.
/:cl:

### DIFF
--- a/code/_onclick/human.dm
+++ b/code/_onclick/human.dm
@@ -108,7 +108,7 @@
 				return
 			user.visible_message(SPAN_NOTICE("\The [user] starts unstrapping \the [back_item] from [xeno]"), \
 			SPAN_NOTICE("You start unstrapping \the [back_item] from [xeno]."), null, 5, CHAT_TYPE_FLUFF_ACTION)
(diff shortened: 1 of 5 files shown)


harryob (approved):
huh

Nanu308:
In part balance I'll say @morrowwolf can make a judgement call on

PR #3208, opened 2023-05-02
Update carrier throws to be projectiles

TeDGamer:
# About the pull request

Huggers being dropped caused issues in the past which led to dropped huggers needing time to become active ( 5 to 15 seconds ).

The issue with this is it gutted carrier throws. Carrier throws are basically useless since how throws work is that it drops the item before throwing the atom. This always resulted in the hugger going idle for 5 to 15 seconds.

What this PR does to keep dropped huggers from becoming a thing is make it so that the carrier actually shoots a hugger projectile. The projectile acts similar to a resin trap, if it hits, it weakens like the trap and then "sleeps / spawns" the same amount of time as the resin trap before hugging. This will require tweaking to see what fits but now carriers are no longer useless when throwing huggers.

Carriers having the abilty to throw huggers but them not doing anything for 5 to 15 seconds / dying before doing anything is a bit of a bug so I wanted to fix that to make them a bit of a force to reckon with without bringing back hugger combat / dropping huggers being an issue.

The code is very similar to resin trap triggering except instead of sleeping, we are spawning the hugger task.

trap.dm
```
if(RESIN_TRAP_HUGGER)
			trap_type_name = "hugger"
			var/obj/item/clothing/mask/facehugger/FH = new (loc)
			FH.hivenumber = hivenumber
			set_state()
			visible_message(SPAN_WARNING("[FH] gets out of [src]!"))
			sleep(15)
			if(FH.stat == CONSCIOUS && FH.loc) //Make sure we're conscious and not idle or dead.
				FH.leap_at_nearest_target()
```

https://www.youtube.com/watch?v=S_FLfet3dwE

# Changelog

:cl:
fix: Carrier huggers are no longer dropped and thrown leading to idle huggers.
balance: Carriers have a windup before they throw huggers now
balance: Carrier throws huggers as projectiles and on hit, is treated similarly to resin traps.
balance: If a hugger hits nothing at the max range of the carrier, it pauses longer than if it would hit a mob before the hugger ai takes over.
/:cl:


code/modules/mob/living/carbon/xenomorph/castes/Carrier.dm
```diff
@@ -249,13 +249,19 @@
...
+		return
+
 	if(!threw_a_hugger)
 		threw_a_hugger = TRUE
 		for(var/X in actions)
 			var/datum/action/A = X
 			A.update_button_icon()
-		drop_inv_item_on_ground(F)
-		F.throw_atom(T, 4, caste.throwspeed)
+		var/obj/item/projectile/huggerProjectile = new /obj/item/projectile(src.loc, create_cause_data(initial(caste_type), src))
```
Drulikar, line 260:
```suggestion
		var/obj/item/projectile/huggerProjectile = new /obj/item/projectile(loc, create_cause_data(initial(caste_type), src))
```

code/modules/mob/living/carbon/xenomorph/castes/Carrier.dm
```diff
@@ -249,13 +249,19 @@
 		to_chat(src, SPAN_WARNING("You need a facehugger in your hand to throw one!"))
 		return
 
+	if(!do_after(src, 0.5 SECONDS, INTERRUPT_NO_NEEDHAND, BUSY_ICON_HOSTILE, numticks = 1))
+		return
+
 	if(!threw_a_hugger)
 		threw_a_hugger = TRUE
 		for(var/X in actions)
 			var/datum/action/A = X
```
Drulikar, line 258: Fix single var usage in this proc. See https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#dont-use-abbreviations

code/modules/projectiles/ammo_datums.dm
```diff
@@ -2886,6 +2886,30 @@
...
 
+/datum/ammo/xeno/facehugger
+	name = "thrown hugger"
+	icon_state = "facehugger_thrown"
+	flags_ammo_behavior = AMMO_XENO|AMMO_SKIPS_ALIENS|AMMO_STOPPED_BY_COVER|AMMO_IGNORE_ARMOR
+
+	max_range = 4
+	shell_speed = AMMO_SPEED_TIER_1
+
+/datum/ammo/xeno/facehugger/on_hit_mob(mob/target, obj/item/projectile/P)
```
Drulikar, line 2897: Fix single var usage in this proc. See: https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#dont-use-abbreviations

Drulikar (changes requested):
Sorry I initially thought I was only going to leave one message.

Balance wise I'm not sure about this: I feel like it maybe should have a more obvious/longer windup. If they aren't already, the projectile huggers should be able to be shot down mid-air and fail if they have a helmet/mask (but remove it). Also, the count of huggers that you currently have seems really high if this is viable. If you were skilled with this you could single-handedly take 16 marines out of the battle before needing to restock (and its easy to make a forward eggmorpher).

morrowwolf:
I have a very hard time seeing this lead to good gameplay. Sorry man.

PR #3221, opened 2023-05-03
added functionality for faxing photos, bundles.

FreetoZ:
# About the pull request

Added functionality for faxing photos and paper bundles, and also refactored some code in this domain.

Admins can now be harassed with multiple pages of paper and photos. 

# Changelog

:cl:
add: Added functionality for faxing photos and paper bundles
fix: Fixed examine text for fax machines
fix: Fixed examine text for paper
/:cl:


FreetoZ:
I think I fixed my blunder, should be good now.

FreetoZ:
linter be lintering, all good now.

code/modules/client/client_procs.dm
Drulikar, line 1: All changes to this file can be discarded if we remove the IMG link.

code/game/machinery/fax_machine.dm
```diff
@@ -164,73 +175,59 @@ var/list/alldepartments = list()
...
-			if(istype(I, /obj/item/paper))
-				usr.drop_inv_item_to_loc(I, src)
-				tofax = I
-				to_chat(usr, SPAN_NOTICE("You put \the [tofax] into \the [src]."))
+			// Repeating code? This is not ideal. Why not put this functionality inside of a proc?
+			var/obj/item/I = ui.user.get_active_hand()
+			if(istype(I, /obj/item/paper) || istype(I, /obj/item/paper_bundle || istype(I, /obj/item/photo)))
+				ui.user.drop_inv_item_to_loc(I, src)
+				original_fax = I
+				to_chat(ui.user, SPAN_NOTICE("You put \the [original_fax.name] into \the [src]."))
```
Drulikar, line 209: Make the fax machine jam if more than 5 items are inserted in a bundle (different to_chat, maybe a sfx, disable send_fax (just need to check the length of the bundle if its a bundle).

code/game/machinery/fax_machine.dm
```diff
@@ -164,73 +175,59 @@ var/list/alldepartments = list()
...
-
-				COOLDOWN_START(src, send_cooldown, fax_cooldown)
-				SendFax(tofax.info, tofax.name, usr, target_department, network, src)
-				to_chat(usr, "Message transmitted successfully.")
-				. = TRUE
+			if(!original_fax)
+				to_chat(ui.user, SPAN_NOTICE("No paper loaded"))
+				return
+
+			copy_fax_paper()
```
Drulikar, line 182: We probably only need to copy if the copy isn't set. Just be sure to clear the copy when original_fax is removed.

Drulikar (changes requested):
Images do not currently work on first transmission (atleast when sent to high command): Neither via IMG link nor via the page view: 
image

Faxing this same bundle to the brig resulted in the images to appear in page, but still not via IMG button. Also each time the bundle is re-transmitted, the photos are duplicated in the fax datum (this is the 2nd transmission of a fax with two images): 
image

If we're including photos in the faxes, its probably unnecessary to have an IMG button anyways

Drulikar:
I don't think view/sending photos need to be admin only either. If its abused moderators can handle it. If it is expensive to view a fax, we can optimize/revert.

FreetoZ:
still working on the duplication issue.

code/game/machinery/fax_machine.dm
```diff
@@ -255,90 +266,82 @@ var/list/alldepartments = list()
...
-	msg_admin += "Receiving '[sentname]' via encrypted connection ... <a href='?FaxView=\ref[faxcontents]'>view message</a>"
+	fax_paper_copy = new()
+	fax_paper_copy.name = original_fax.name
 
-	var/msg_ghost = SPAN_NOTICE("<b><font color='#1b748c'>COLONIAL MARSHAL BUREAU FAX: </font></b>")
-	msg_ghost += "Receiving '[sentname]' via encrypted connection ... <a href='?FaxView=\ref[faxcontents]'>view message</a>"
+	if(istype(original_fax, /obj/item/photo))
+		var/obj/item/photo/faxed_photo = original_fax
+		photo_list += list("tmp_photo.png" = faxed_photo.img )
+		fax_paper_copy.info  += "<img src='tmp_photo.png' width='100%'/>"
```
Drulikar, line 285:
```suggestion
		fax_paper_copy.info  += "<img src='tmp_photo.png' width='192'/>"
```
100% just looks way too stretched and requires too much height when there's a fax that's wide (it also shouldn't be stretching on a printed piece of paper). We'll double it.

image

code/game/machinery/fax_machine.dm
```diff
@@ -255,90 +266,82 @@ var/list/alldepartments = list()
...
+		fax_paper_copy.info += "<hr>"
+		if(istype(papers[content], /obj/item/paper))
+			var/obj/item/paper/faxed_paper = papers[content] 
+			fax_paper_copy.info += faxed_paper.info
+		else // type photo
+			var/obj/item/photo/faxed_photo = papers[content]
+			if(!isicon(faxed_photo.img))
+				return
+			photo_list += list("tmp_photo[content].png" = (faxed_photo.img))
+			fax_paper_copy.info  += "<img src='tmp_photo[content].png' width='100%'/>"
```
Drulikar, line 301:
```suggestion
			fax_paper_copy.info  += "<img src='tmp_photo[content].png' width='192'/>"
```
100% just looks way too stretched and requires too much height when there's a fax that's wide (it also shouldn't be stretching on a printed piece of paper). We'll double it.

Drulikar (changes requested):
I think all technical issues are solved. Here is a polish pass to fix messages, behavior, and anything else that stuck out to me.

There are a fair few places where the messages and examines are a bit wonky: My suggestions should address them. I also tweaked how jamming is handled because I think its more interactive and interesting to require unjamming the machine.

By the way, you don't need to manually add suggestions I provide, just batch them, try them out, and edit as needed:
image

The examine text for bundles including images is also bit weird. Please fix that with this patch: 
[patch.diff.txt](https://github.com/cmss13-devs/cmss13/files/11413852/patch.diff.txt) See https://git-scm.com/docs/git-apply for information on how to apply a patch (e.g. `git apply PATH_TO_PATCH_FILE`)
image

code/modules/paperwork/paper.dm
Drulikar, line 1: get_examine_text is only called if you examine the paper, not use the paper. But we still need to browse_rsc any images for anyone that views this paper. See how paper_bundle.dm is handling its images.
image

1. Create a bundle of photos
2. Fax those photos to a department
3. Use (or examine) the paper that prints out without viewing the fax in chat.

Drulikar (approved):

PR #3223, opened 2023-05-03
CMB Black Market Investigation

BeagleGaming1:
# About the pull request

Ordering black market goods can either lower or increase a number. At 100, the CMB gets called to investigate and disable the black market.

People can purchase unlimited goods such as weapons or CLF crates. This causes req to have to at least pay attention to what they are buying, and to buy stuff other than clothing or weapons from other factions.

# Changelog

:cl:
add: Adds a CMB ERT to investigate the Black Market
add: Adds a [cmb] logo for papers
/:cl:


Diegoflores31:
Heat amount should lower itself over time OR you should be able to pay up a "Tax" to lower the heat 
because if not it just invalidates completely the fact that the crates cost money

sleepynecrons (changes requested):
the yellowish area needs cleaning up it's currently a complete mess. should be 4-6 colors at most
9WCW19t

make the blue when it's turned on a bit brighter
bGNyKWP
something like this where the bars in the front contrast more with the background

sleepynecrons (approved):
sprites look alright now approved 👍

morrowwolf:
Let's get this TM'd.

morrowwolf:
Spoke on discord but:

The arrival and dispatch messages seems to have not gone off and the objectives seems to have not been displayed.
Heat needs to be displayed easily to admins somewhere

Potentially out of scope:
CMB need to be set to their own faction
USCM can be in their faction_group though

That's all for now. Fix up at least the top two and we'll TM again.

code/datums/emergency_calls/inspection.dm
```diff
@@ -263,3 +263,45 @@
...
+/datum/emergency_call/inspection_cmb/black_market/New()
+	..()
+	dispatch_message = "Third Fleet High Command to [MAIN_SHIP_NAME], we have received inconsistent supply manifests and irregularities on the ASRS system aboard your ship, and have requested that a CMB Investigation Team to board and clear you of any wrongdoing."
+	arrival_message = "Incoming Transmission: [MAIN_SHIP_NAME], this is Anchorpoint Station with the Colonial Marshal Bureau. Be advised, we are dispatching an team of Marshals to board with you by request of GSO-91. Submitting authorized docking clearances, over."
+	objectives = "Investigate the inconsistencies aboard the [MAIN_SHIP_NAME]'s ASRS. In the case of illegal activity, collect evidence, and submit a report to the CMB Command at Anchorpoint Station. Work with the [MAIN_SHIP_NAME]'s military police force."
+
+	will_spawn_icc_liaison = TRUE
+	will_spawn_cmb_observer = FALSE
+
+/datum/emergency_call/inspection_cmb/black_market/create_member(datum/mind/M, turf/override_spawn_loc)
```
morrowwolf, line 284: one letter var "M"

code/game/objects/items/circuitboards/computer.dm
```diff
@@ -178,11 +185,35 @@
...
+
+		if(black_market_lock)
+			to_chat(user, SPAN_NOTICE("[src] has already been reset."))
+			return
+
+		playsound(tool, 'sound/machines/lockenable.ogg', 25)
+		user.visible_message(SPAN_NOTICE("[user] attaches [tool] to [src]."),\
+		SPAN_NOTICE("You begin to fix any tampering to [src]."))
+		tool.icon_state = "[tool.icon_state]_on"
+
```
morrowwolf, line 208: if(user.action_busy)
to_chat(user, "I'm already doing something moron")
return

Paraphrased of course

code/game/supplyshuttle.dm
```diff
@@ -147,6 +148,11 @@ var/datum/controller/supply/supply_controller = new()
 
 	//If any computers are able to order contraband, it's enabled. Otherwise, it's disabled!
 
+/obj/structure/machinery/computer/supplycomp/proc/lock_black_market(market_locked = FALSE)
+	for(var/obj/structure/machinery/computer/supplycomp/computer as anything in supply_controller.bound_supply_computer_list)
```
morrowwolf, line 152: I don't believe we need the "as anything" here

code/game/supplyshuttle.dm
```diff
@@ -147,6 +148,11 @@ var/datum/controller/supply/supply_controller = new()
...
+		if(market_locked)
+			black_market_lockout = TRUE
```
morrowwolf, line 154: Why are we looping but not referencing the computer

morrowwolf (changes requested):

code/game/supplyshuttle.dm
```diff
@@ -147,6 +148,10 @@ var/datum/controller/supply/supply_controller = new()
 
 	//If any computers are able to order contraband, it's enabled. Otherwise, it's disabled!
 
+/obj/structure/machinery/computer/supplycomp/proc/lock_black_market(market_locked = FALSE)
+	if(market_locked)
+		black_market_lockout = TRUE
```
morrowwolf, line 153: I'm pretty sure this proc is supposed to black market lockout all the currently connected computers. The issue with it before was not the for loop but that the for loop wasn't actually interacting with the bound computers it is meant to be locking but instead just resetting the one computer's lockout value.

morrowwolf: Oh and document the proc please (if I'm correct)

morrowwolf (changes requested):

PR #3232, opened 2023-05-04
Removed the "missile noise" from sound/weapons/dropship_sonic_boom

LynxSolstice:
# About the pull request

Removes the missile launching noise from the sonic boom FX that plays when a dropship is performing an FM.

Missile noise even when not using missiles because the sound wasn't properly cleaned up at any point: bad.
Just a sonic boom sound effect which keeps the sound profile consistent throughout all the FMs: good.

Changed the sound file, checks shouldn't have any issues. Compiles and runs without any errors.

# Changelog

:cl:LynxSolstice
soundadd: Cleaned up the sonic boom sound effect to remove the "missile launching" sound, which caused an inconsistency when the dropship wasn't using missiles.
/:cl:


fira (approved):

morrowwolf (approved):

morrowwolf (changes requested):
Should be normalized to -8db.

PR #1244, opened 2022-10-13
Updates VSC tooling with PrettierX and the DMI Editor

harryob:
## About The Pull Request

updating from tgui upstream, we can use prettierx for standardised formatting tgstation/tgstation#67935
the DMI viewer is cool too tgstation/tgstation#66834

this:
adds prettierx
adds eslint-config-prettier to automatically disable conflicting ESlint rules
adds prettier, ZipFS (for yarn plug'n'play), DMI viewer as recommended extensions
automatically uses prettier when file saved in VSC
sets default DMI viewer as the new DMI viewer extension

testing wise i deleted my local repo and VScode + settings, i reinstalled VSC fresh, cloned my fork and it worked. tgui still worked, too

i'd appreciate a couple of people checking out this PR branch and seeing if it fucks it up for them, mind

## Why It's Good For The Game

easier to sprite
nicer to read code
simple as

## Changelog

shouldn't be any player facing changes

PR #3234, opened 2023-05-04
ci should create errors when there's warnings on the codebase

harryob:
i don't want warnings

nothing player facing :3

### DIFF
--- a/.github/workflows/ci_suite.yml
+++ b/.github/workflows/ci_suite.yml
@@ -109,3 +109,25 @@ jobs:
     uses: ./.github/workflows/run_unit_tests.yml
     with:
       map: ${{ matrix.map }}
+
+  test_windows:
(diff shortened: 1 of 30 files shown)


fira:
why all the tgui changes? are these all whitespace or is that intended?

harryob:
> why all the tgui changes? are these all whitespace or is that intended?

when updating build.js to tg's current one, it also brought with it the check for if a file is formatted with prettier (which i missed from #1244 👀) so since 1244 prs have been merged without being prettier-ified. they're all just formatting standardization though

Drulikar (approved):
I wouldn't be surprised if for a while merge queues fail for anything that hasn't re-run tests, but otherwise I see nothing that these changes break. I can still build, test, and I see no noticeable differences in some of the terminals that were prettified.

PR #3235, opened 2023-05-04
Rework of the IO Computer Lab Area.

I3LiP:
# About the pull request

Reworks the IO Computer Lab Area on the Almayer and changes some door access permissions. Lobby area has had the disk decrypters removed and extra space made for crates and items dumped by other roles. The main Intel sorting area has been reworked to bring the disk decrypting into the more secure area with the rest of the Intel. The secure storage area has been altered to make it more suitable for storing high value items and less of a dumping ground. A small overwatch like area has also been added so that shipside IOs have a greater ability to support groundside IOs.

Makes the IO area more interesting and feel more like the offices of an actual intelligence agency. Adds some roleplaying items to encourage and enable more shipside RP. Should streamline the Intel sorting process so that IOs can work in a more appropriate space with better organisation. The overwatch area should allow shipside IOs to support groundside IOs better with tactical intelligence and encourage them to co-ordinate better. Gives shipside IOs something interesting and valuable to do while waiting for delivery of more Intel crates and disks to decrypt.

- Tested doors
- Tested consoles
- Tested fire alarms
- Checked APCs
- Checked lighting
- Checked cameras
- Checked for unexpected items

image

image
<details>

</details>

# Changelog
:cl:
mapadd: Reworked the IO area and added an overwatch type area for shipside IOs.
/:cl:


morrowwolf:
Brief lookover seems fine. Gonna TM it just in case I'm missing something integral to IOs.

morrowwolf (changes requested):
Space tile underneath table in between front room and back room.

East glass door on the same tile as above has extra accesses.

Hopekz:
Occasional IO main here dropping feedback:

This rework removes as important part of IO OCD of sorting the finished items in their own stack so you can do stuff like grab the processed folders and processed disks during evacuation for Roleplay purposes (Because IO's in RP make sure that stuff gets delivered to command at the end of evac).

Also was hoping that a rework removed that glass door aspect of it because no-one uses that ever. It is locked behind a door with access so no-one can get to that "line" to even wait there in the first place for whatever they need from IO.

Also why is the phone all the way in the back most of IO work happens in either the first room with the waiting line or in the second room so I just thought it was placed very "out of the way".
In RP I've had multiple scenario's where crew members ask to use our phone to make a call so now it would be weird to lead them all the way through to the sensitive information to do so but that part is more of a nuance.

On the topic of the line: Why does it exist? IO's use all the floor space in that room to sort and now there's a barrier there for some reason. IO's don't get a single person in that line let alone 2 to actually put a turnstile in there.

Thank you!

WolfOrion:
Platinum IO here to say that while the drastic cut to shelf space is definitely a hit to the sorting part of the job, I personally like the more aesthetic look of the rooms. If only we could put folders and papers in the filing cabinets but that might be beyond the scope of this PR. As well, the map table in the little command area shows the Almayer, not the area of operations - is this intentional?

IOs also lack training to use the overwatch console, but that's also beyond the scope of these changes I'm sure. It's just begging for us to form echo squad as an intel team and have a dedicated staff officer. Not sure what else can be done with it though.

I3LiP:
I've made some slight tweaks to add more racking and create a little more space in the main room to maintain the aesthetic of the new layout while giving a little more functionality. In terms of the lobby area it seems strange to do the sorting out there so hopefully this helps to bring it into the main area.

The phone was moved to the 'command' area so that it can be used to better co-ordinate between ground and shipside IOs.

The Tacmap showing the Almayer was a bug.

In terms of Echo squad and actual Overwatch that's indeed outside the scope of this PR (and my abilities). I might look at the filing thing in another branch however.

Thanks for the feedback.

New layout:
image

morrowwolf (approved):

PR #3237, opened 2023-05-04
Alamo and Normandy now have a Reinforced Metal Ceiling

Hopekz:
Stops this kind of stuff from happening and weird tactics like CAS/OB'ing the dropship because people know it can't be destroyed:

https://user-images.githubusercontent.com/24533979/236336813-8ecd8ba1-ba8e-4834-9acc-52814f466aec.mp4

The whole thing is indestructible might as well make the ceiling indestructible as well.

image
image

:cl: Hopek
balance: Alamo and Normandy now have a Reinforced Metal Ceiling meaning that CAS and OB's can't penetrate it.
/:cl:

### DIFF
--- a/code/game/area/Sulaco.dm
+++ b/code/game/area/Sulaco.dm
@@ -23,30 +23,36 @@
 	name = "\improper Dropship Alamo"
 	icon_state = "shuttlered"
 	base_muffle = MUFFLE_HIGH
+	ceiling = CEILING_REINFORCED_METAL
 
 /area/shuttle/drop1/LV624
 	name = "\improper Dropship Alamo"
 	ambience_exterior = AMBIENCE_LV624
 	icon_state = "shuttle"
+	ceiling = CEILING_REINFORCED_METAL
 
 /area/shuttle/drop1/prison
 	name = "\improper Dropship Alamo"
 	ambience_exterior = AMBIENCE_PRISON
 	icon_state = "shuttle"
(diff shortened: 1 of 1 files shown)


harryob (approved):

TeDGamer:
This doesn't completely fix the problem. Offsets can still cause it to hit areas that are reinforced like this.

Hopekz:
> This doesn't completely fix the problem. Offsets can still cause it to hit areas that are reinforced like this. 

If this is true then out of scope for this PR.
If you can exploit offsets to drop on to areas where you normally wouldn't be able to penetrate then that would be fixed by making air support drop code respecting ceilings after offset.

Regardless I think this sends a clear message that this isn't intended to be breached so now getting around this with an exploit would probably be ban-able

Diegoflores31:
woulndt that make you unable to drop OBs on the LZ ? -1

harryob:
> woulndt that make you unable to drop OBs on the LZ ? -1

no, the area that the dropships land is in the landing zone until the shuttle arrives. been that way since the shuttle rework

morrowwolf:
Let's just do this the hive core way as was discussed.

If the queen locks or calls down the dropship the core has its cooldown set to zero when it gets destroyed next. This puts the onus on the players to secure their new area and stops weird area roofing.

morrowwolf:
Alright, I have reconsidered again. Let's just do both.

PR #3240, opened 2023-05-05
Patches custom rank prefixes for Admin mod perms.

Tsunamico:
# About the pull request

As the tin says.

<details>

</details>

# Changelog

:cl:
fix: Patched custom prefixes for admin events.
/:cl:

### DIFF
--- a/code/datums/paygrades/helper.dm
+++ b/code/datums/paygrades/helper.dm
@@ -5,17 +5,20 @@
 
 	if(size)//Builds the prefix, if one should exist.
 		var/NP = ""
-		if(P.fprefix)//Factional (pre)prefix
-			NP = "[P.fprefix] "
-		if(P.prefix)//Actual rank prefix
-			NP =  "[NP][P.prefix] "
-			if(P.prefix == "C")//Override for Civilian ranks that don't have occupational prefixes.
-				if(gender && gender == FEMALE)
-					NP = "Ms. "
-				else if(gender && gender == MALE)
-					NP = "Mr. "
-				else
-					NP = "Mx. " //inclusivity win!
+		if(P != null)//extra check to avoid runtimes
+			if(P.fprefix)//Factional (pre)prefix
+				NP = "[P.fprefix] "
+			if(P.prefix)//Actual rank prefix
+				NP =  "[NP][P.prefix] "
+				if(P.prefix == "C")//Override for Civilian ranks that don't have occupational prefixes.
+					if(gender && gender == FEMALE)
+						NP = "Ms. "
+					else if(gender && gender == MALE)
+						NP = "Mr. "
+					else
+						NP = "Mx. " //inclusivity win!
+		else
+			NP = "[paygrade] "
 		return NP
 	else
 		return P.name


Tsunamico:
Sincerely, be careful when merging this. Running two rounds with this TMed too see if shit breaks ain't a bad idea. Cuz this fix is CURSED.

code/datums/paygrades/helper.dm
```diff
@@ -5,17 +5,20 @@
...
-		if(P.prefix)//Actual rank prefix
-			NP =  "[NP][P.prefix] "
-			if(P.prefix == "C")//Override for Civilian ranks that don't have occupational prefixes.
-				if(gender && gender == FEMALE)
-					NP = "Ms. "
-				else if(gender && gender == MALE)
-					NP = "Mr. "
-				else
-					NP = "Mx. " //inclusivity win!
+		if(P != null)//extra check to avoid runtimes
```
realforest2001, line 8:
```suggestion
		if(!P)//extra check to avoid runtimes
			return FALSE
```

code/datums/paygrades/helper.dm
```diff
@@ -5,17 +5,20 @@
...
+				NP =  "[NP][P.prefix] "
+				if(P.prefix == "C")//Override for Civilian ranks that don't have occupational prefixes.
+					if(gender && gender == FEMALE)
+						NP = "Ms. "
+					else if(gender && gender == MALE)
+						NP = "Mr. "
+					else
+						NP = "Mx. " //inclusivity win!
+		else
+			NP = "[paygrade] "
```
realforest2001, line 21:
```suggestion
		if(P.fprefix)//Factional (pre)prefix
			NP = "[P.fprefix] "
		if(P.prefix)//Actual rank prefix
			NP =  "[NP][P.prefix] "
			if(P.prefix == "C")//Override for Civilian ranks that don't have occupational prefixes.
				if(gender && gender == FEMALE)
					NP = "Ms. "
				else if(gender && gender == MALE)
					NP = "Mr. "
				else
					NP = "Mx. " //inclusivity win!
	else
		NP = "[paygrade] "
```

code/datums/paygrades/helper.dm
```diff
@@ -5,17 +5,20 @@
 
 	if(size)//Builds the prefix, if one should exist.
```
realforest2001, line 6: This is technically where the if(P) thing should go, rather than doing another check later

realforest2001 (changes requested):

code/datums/paygrades/helper.dm
Tsunamico: Leaving as is currently, messing with THAT is dangerous.

Tsunamico:
Should be alright now, It's a fine enough workaround!

Drulikar (approved):
Weird issues for me testing this until merged with master; but after the merge it seems to be working and I can create custom paygrades without runtimes. I could not find an admin way of setting up a squad, but I tried using the squad assignment computer (which dumbly initializes the squad while the marine has no ID so they get a N/A rank until the squad gets updated again) and could make sure the other case ran correctly too.

PR #3244, opened 2023-05-06
Medics now get to buy the EZ Autoinjectors if they want

Hopekz:
The only way to get these injectors at the moment is going through the auto injector pouches in preparations so I've just been watching medics go through like 12+ pouches on their own for weeks just to get that single injector out of it. 
Just let them buy the injector so they don't have to ruin 12+ autoinjector medical pouches for everyone else.

Was not 100% sure what the price should be so I set it at the higher amount of 2 because that's what the premium injectors go for currently.

:cl: Hopek
add: Medics can now just purchase the EZ Autoinjectors so they don't have to go through 12+ injector pouches to take them out specifically.
/:cl:


morrowwolf (approved):
They're still gonna go through the pouches but this is fine I suppose.

PR #2641, opened 2023-02-13
/vending tgui remake 

stanalbatross:
# About the pull request

remake of PR #2420 

yes this still has the `let`s etc in but it's more important to just get the functional UI in first, and if Harry couldn't make it work w/o the `let`s I don't think I'll be able to 

# Explain why it's good for the game

it's tgui

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl:
ui: converted money-based vendors to tgui
/:cl:

PR #3247, opened 2023-05-06
Add Cash & Access Support to Vending Machines & Fix ATM Security Levels

### LINKED ISSUE #1825: Prisoners can't use brig's vendors
### Testmerges

-

### Description of the bug

The brig common area has a number of vendors, but one problem is that those vendors require an ID to be able to pay for them, which the prisoners don't have. One solution would be to give them an ATM to withdraw money vie their account number and PIN (that nobody would remember, so probably would need to find it somehow via a verb), or set those vendors to be free.

Things in the brig common area should be usable by prisoners.

1. Get put in brig as per protocol without ID
2. Try to get some smokes - you can't

### Issue Bingo

Drulikar:
# About the pull request

This PR fixes #1825 more or less by fixing the ATM security levels 0 and 1 that were supposed to optionally require an ID (they didn't require inserting the ID into the machine but still required an ID on their person). Guards will still need escort prisoners up to the ATM, but as long as the prisoner remembers their account number (and optionally pin for security level 1 - it is always available in their Notes) and didn't upgrade their account security to 2 requiring their ID, they can withdraw funds as cash or into a charge card (recommended ;) ).

Vending machines can also now accept cash or charge cards and their servos have been oiled up to have a nice ker-chunk sound when vending. They will still prefer to withdraw funds from your account if able. Hacking their IDSCAN wire will still make the cost free.

SFX is from https://freesound.org/people/Emma7073/sounds/250194/ (Creative Commons) with some reverb and trimming.

Also apparently vendors were not checking access after #2641 but they will again.

What else are you going to use those funds for? Ultimately this should encourage more interactions between MPs and prisoners, and possibly elsewhere. Who's responsible for keeping the change in those machines anyways?

Also no more busting into SEA for rubber bullets.

<details>

https://youtu.be/Uq7W7ceyrQY

</details>

# Changelog
:cl: Drathek
add: Added cash support to vending machines
fix: Fixed ATM account security levels (now a ID card is actually optional like it says!)
fix: Fixed ATM not resetting failed attempts on successful login
fix: Fixed vendors not checking access requirements
soundadd: Added a ker-chunk sound for vending
spellcheck: Improperified all /obj/structure/machinery/vending names
code: Cleanup vending.dm
/:cl:


code/game/machinery/vending/vending.dm
```diff
@@ -555,25 +548,67 @@ GLOBAL_LIST_EMPTY_TYPED(total_vending_machines, /obj/structure/machinery/vending
...
 	record.amount--
+	playsound(src, "sound/machines/vending.ogg", 40, TRUE)
 	if(user.Adjacent(src) && user.put_in_hands(vended_item))
 		to_chat(user, SPAN_NOTICE("You take \the [record.product_name] out of the slot."))
 	else
 		to_chat(user, SPAN_WARNING("\The [record.product_name] falls onto the floor!"))
+	if(ripped_off)
+		flick(icon_deny,src)
+		sleep(1 SECONDS)
+		speak("Insufficent change. Please contact a supervisor.")
```
harryob, line 611:
```suggestion
		addtimer(VARSET_CALLBACK(src, vend_ready, TRUE), 1 SECONDS)
		addtimer(CALLBACK(src, PROC_REF(speak), "Insufficient change. Please contact a supervisor."), 1 SECONDS)
		return
```

harryob (changes requested):

code/game/machinery/vending/vending.dm
```diff
@@ -803,29 +877,29 @@ GLOBAL_LIST_EMPTY_TYPED(total_vending_machines, /obj/structure/machinery/vending
 				S.remove_from_storage(item_to_stock, user.loc)
 
 			qdel(item_to_stock)
-			user.visible_message(SPAN_NOTICE("[user] stocks [src] with \a [R.product_name]."),
-			SPAN_NOTICE("You stock [src] with \a [R.product_name]."))
-			R.amount++
+			user.visible_message(SPAN_NOTICE("[user] stocks the [src] with \a [product.product_name]."),
+			SPAN_NOTICE("You stock the [src] with \a [product.product_name]."))
```
harryob, line 881: the the should be being added by dm based on the properness of the datum - is this being weird - or is it just an easier solution than ensuring all the names of the vending types are \improper?

Drulikar: I just figure a vending machine should never be proper; even though many people like the capitalize its name and not override it with \improper so it incorrectly will be considered proper (there is only one uncapitalized vendor name, and one other capitalized vendor that is \improper). But I can restore the `\the` usage and either \improper or uncapitalize all vendor names.

morrowwolf (approved):

PR #3249, opened 2023-05-06
Fix being able to go above 100% blood volume without limit.

Huffie56:
# About the pull request

being at 100% mean you have 560 cl of blood but the real limit code wise is 600 or 107%
so to  help with that i made it so that when your above normal blood reduce by 0.1cl
and when your above limit it remove 1cl
that way you can't go above the limit fixed by the code.

fixes https://github.com/cmss13-devs/cmss13/issues/364

autocloses on merge

It prevent people to go above the blood limit in the code.

<details>

image

</details>

every-time you go above 600 it set you back to 600 as it's the limit of the blood volume in the code.
# Changelog
:cl:

fix: Fix being able to have infinite blood. soft limit above 100% reduce volume by 0.1 and hard limit at 107% prevent you to go above.

/:cl:

### DIFF
--- a/code/modules/mob/living/blood.dm
+++ b/code/modules/mob/living/blood.dm
@@ -17,7 +17,7 @@
 		if(blood_volume > max_blood)
 			blood_volume -= 0.1 // The reverse in case we've gotten too much blood in our body
 		if(blood_volume > limit_blood)
-			blood_volume -= 1 // The reverse in case we've got oveer the limit
+			blood_volume = limit_blood // The reverse in case we've got oveer the limit
 
 		var/b_volume = blood_volume
 
--- a/code/modules/mob/living/blood.dm
+++ b/code/modules/mob/living/blood.dm
@@ -14,8 +14,10 @@
 		//Blood regeneration if there is some space
 		if(blood_volume < max_blood)
 			blood_volume += 0.1 // regenerate blood VERY slowly
-		if(blood_volume > max_blood)
+		else if(blood_volume > max_blood)
 			blood_volume -= 0.1 // The reverse in case we've gotten too much blood in our body
+		else if(blood_volume > limit_blood)
+			blood_volume -= 1 // The reverse in case we've got oveer the limit
 
 		var/b_volume = blood_volume
 
--- a/code/modules/mob/living/blood.dm
+++ b/code/modules/mob/living/blood.dm
@@ -14,9 +14,9 @@
 		//Blood regeneration if there is some space
 		if(blood_volume < max_blood)
 			blood_volume += 0.1 // regenerate blood VERY slowly
-		else if(blood_volume > max_blood)
+		if(blood_volume > max_blood)
 			blood_volume -= 0.1 // The reverse in case we've gotten too much blood in our body
-		else if(blood_volume > limit_blood)
+		if(blood_volume > limit_blood)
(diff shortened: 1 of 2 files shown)


morrowwolf:
I think there's a better way to do this. Looks like Drathek is looking into it.

Drulikar (changes requested):
I like the changes here to assert we don't go over our maximum and slowly restore to normal, however the root of this issue is in that food has PROPERTY_HEMOGENIC and /datum/chem_property/positive/hemogenic/process somereason is increasing blood_volume 100 past the limit.

Please try this patch (since I had to alter files outside where you have made changes): 
[patch.diff.txt](https://github.com/cmss13-devs/cmss13/files/11416508/patch.diff.txt)
and see if you like the effect.

See https://git-scm.com/docs/git-apply for more information on how to apply a patch but basically its just `git apply PATH_TO_THE_PATCH_FILE`

Drulikar:
If you like the changes, be sure to update the changelog to reflect what we've changed. I recommend probably test merging this though since it has the potential to affect a lot of things since the bad effect of too much blood is much more possible.

We also might want to make it so predators don't lose blood slowly over time; but I don't think its a very significant change that they can't deal with.

Huffie56:
> I like the changes here to assert we don't go over our maximum and slowly restore to normal, however the root of this issue is in that food has PROPERTY_HEMOGENIC and /datum/chem_property/positive/hemogenic/process somereason is increasing blood_volume 100 past the limit.
> 
> Please try this patch (since I had to alter files outside where you have made changes): [patch.diff.txt](https://github.com/cmss13-devs/cmss13/files/11416508/patch.diff.txt) and see if you like the effect.
> 
> See https://git-scm.com/docs/git-apply for more information on how to apply a patch but basically its just `git apply PATH_TO_THE_PATCH_FILE`

this feel like magic i have never used git not even sure i set it up right...
i think i am just going change the code manually i think....

Huffie56:
i don't know why this variable is there but i assume it's some sort of check for carbon live form??
so i rather not touch it if that's ok?
image

Huffie56:
image
the mechanic that is suppose to give you oxy damage is broken  it seem.
but i rather fix it in a seperate PR if it's fine with you @Drulikar ?

morrowwolf:
I've shoved this into #3273 so I don't have to deal with conflicts. Will give credit.

Drulikar:
Just to clarify:

> this feel like magic i have never used git not even sure i set it up right... i think i am just going change the code manually i think....

You made this PR by pushing commits to a branch so you have git set up. Just for future reference, If you're using VSC, in the terminal window (Terminal > New Terminal if you don't have one open at the bottom already) that is where you can execute commands like `git status` or `git apply patch.diff.txt`

E.g. 
image

> i don't know why this variable is there but i assume it's some sort of check for carbon live form?? it's safe to assume that if it's there it's for a reason so i rather not touch it if that's ok? image

So blood_volume is a variable that exists on /mob/living - it does not originate from /mob/living/carbon so it was unnecessary for the code to cast the /mob/living to a /mob/living/carbon since the variable they were accessing exists already on the type it was given; thats why I removed the typecast (technically the iscarbon check is also unnecessary - but there can be a legitimate reason for this code to not run on non-carbon mobs).

PR #3254, opened 2023-05-07
Marine jackets

irRegularGuy646:
# About the pull request

adds a (unarmoured) jacket as a loadout option for 2 points. Tested.

an option for marines who genuinely prefer to not use armour.

# Changelog

:cl:
add: Added Cheap jacket as loadout option
/:cl:


morrowwolf:
You have stumbled into the classic argument: should regular marines have access to jackets that promote non-usage of armor.

Current administration says *drum roll*: no.

Use your armor. Roles that don't have mandatory deployment may have it added to their vendors.

irRegularGuy646:
why not?

PR #3260, opened 2023-05-07
fixes fultons using reservations

### LINKED ISSUE #3253: Fultons Broken?
### Testmerges

#3088, #3185, #3192, #3225, #3235

### Description of the bug

Attempted to use my fultons, and while they attached to the xenomorphs it did not then just send them into the air and the dropship pilot could not retrieve them.
image
It also left "ghost fultons" or something after we dragged the corpses away.
image

After attaching the fultons to them it should've sent them into the air for the dropship to retrieve.

1. Attach the fulton to a dead xenomorph or crate.
2. Nothing will happen probably.

### Issue Bingo

harryob:
apparently something was a soft dependency on low orbit and that was fultons, whoops

:cl:
fix: fixes fultons
/:cl:

closes #3253

### DIFF
--- a/code/game/objects/items/fulton.dm
+++ b/code/game/objects/items/fulton.dm
@@ -22,6 +22,7 @@ var/global/list/deployed_fultons = list()
 	var/atom/movable/attached_atom = null
 	var/turf/original_location = null
 	var/attachable_atoms = list(/obj/structure/closet/crate)
+	var/datum/turf_reservation
 
 /obj/item/stack/fulton/New(loc, amount, atom_to_attach)
 	..()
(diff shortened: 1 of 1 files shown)


code/game/objects/items/fulton.dm
```diff
@@ -157,8 +161,8 @@ var/global/list/deployed_fultons = list()
 
 /obj/item/stack/fulton/proc/return_fulton(turf/return_turf)
 
-	// Fulton is not in space, it must have been collected.
-	if(!istype(get_area(attached_atom), /area/space/highalt))
+	// Fulton is not in reservation, it must have been collected.
+	if(!(get_turf(src) in reservation.reserved_turfs))
```
fira, line 165: Correct me if I'm wrong but I think you'd still want to delete the reservation then, no ?

fira: Ideally probably on pickup proper instead, in case something happens to the fulton object in the meantime

harryob: oddly, /recall_fulton is the proc used when picking it up by a dropship

fira (approved):
right, i completely forgot about that, it is recalled to the dropship... my bad

Drulikar (approved):
I can confirm these changes work locally (tried leaving a client in a dead runner): 
image

PR #3261, opened 2023-05-07
Pred Cloak Drain Change

Ordosian:
# About the pull request

After discussing with the Yautja Council, I have changed how the cloak affects the energy drain.

Initial cloaking costs 100 now instead of 50.

The cloak no longer drains charge. Instead, you gain 15 charge per tic. (Default gain is 30 uncloaked)

Less preds standing around in the open. More spooky ghosts.

It was tested on Private Server. Works as intended.

# Changelog
:cl:
balance: Yautja initial cloak costs 100 charge instead of 50. Cloak no longer drains but charges by 15.
/:cl:

### DIFF
--- a/code/modules/cm_preds/yaut_bracers.dm
+++ b/code/modules/cm_preds/yaut_bracers.dm
@@ -283,7 +283,7 @@
 	var/mob/living/carbon/human/human = loc
 
 	if(cloaked)
-		charge = max(charge - 10, 0)
+		charge = min(charge + 15, charge_max)
 		if(charge <= 0)
 			decloak(loc)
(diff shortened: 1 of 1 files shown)


morrowwolf:
In discussion.

morrowwolf (changes requested):
Alright here's what I'd like:

Cloak takes zero power (as in does not effect power charge at all).
Cloak is disabled for 30 seconds when encountering water (river/sprayed/etc).
Cloak is disabled for 60 seconds when hit by EMP.

Changes to bracer power coming soonTM likely.

PR #3266, opened 2023-05-08
New incoming direct fire CAS sounds (kinda) 

Tisx1:
# About the pull request
So, uh just adding in some new more beefer more realistic sound for incoming direct fire from CAS, thing is, uh, the sound file has been in for like 6 months, (#1484)  I just uh, forgot to put it to use in the code, because I am clearly a genius who is just wants GBP. whatever the case, it's in use now. Also put it in the tripping balls in a boiler cloud stuff. 

Good sound atmosphere makes it sound like a rocket is actually falling from orbit at insane speeds. 
Heres the sound: 

https://user-images.githubusercontent.com/35430138/236721788-0cb727ba-b98e-45f1-bd63-50d7adc5709d.mp4

# Changelog
:cl: Tisx

soundadd: New sound for incoming direct fire from CAS. 

/:cl:


fira:
Just one thing, won't this be too long vs the actual weaponry ? I thought it lasted like ~1.5 seconds, so it would explode before the actual sound explosion

Drulikar (changes requested):
This has weird timing and just seems sort of odd for non-rocket weapons. I'd suggest making rocket subtypes override the warning_sound value so it can be rocket specific. Also consider adding this to rocket firemissions?

https://youtu.be/tdj1X_kyFmc

PR #3011, opened 2023-04-03
Goonify chemical retrieval (round two)

morrowwolf:
# About the pull request

This PR aims to do the following:
Replace the current PMC chem team with a corporate security chem team led by a researcher.
The researcher can call a PMC team for assistance so we have escalation if required. (admin approved)

This should allow for escalation of force for what should be a fairly cooperative situation. Weyland Yutani and the USCM are close friends in the Neroid sector. There is no reason this should blow up as much as it does but there should be some tension.

As always, this is a work in progress. Nothing is 100% final.

# Explain why it's good for the game

Read above.

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>
I briefly tested some features of this.

</details>

# Changelog

:cl: Morrow
add: Chemical retrieval is now primarily corporate goons with a backup team of PMCs if required.
/:cl:

PR #3268, opened 2023-05-08
Fix for ChemERT objectives.

Segrain:
# About the pull request

Followup to #3011.
This was brought to my attention after the chemical goons somehow did not receive their proper objective and the resulting raid escalated a lot. Once again, my gratitude to the player who sent me the screenshots.
Although I am honestly baffled by how this works. When tested locally, present setup works just fine. On live server, see next section. Is it a stable/beta Byond difference at work, maybe?
Still, this seems like an obvious thing to fix. Also, slightly modified the text to emphasise that there is (most likely) a _singular_ chemical that is the objective.

image
image

# Changelog
:cl:
fix: Chemical Investigation Goons should now properly be shown their actual objective.
/:cl:

### DIFF
--- a/code/datums/emergency_calls/goons.dm
+++ b/code/datums/emergency_calls/goons.dm
@@ -78,7 +78,7 @@
 
 	print_backstory(mob)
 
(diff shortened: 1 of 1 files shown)


harryob (approved):
yeah, different behaviour on 515 due to the proc_ref macros compiling differently dependent on version

fira (approved):
funny leftover... we should have fixed all of these months ago :|

PR #3281, opened 2023-05-09
Boots can now hold a scalpel

Hopekz:
I was playing medic and I thought "why can't I just replace my boot knife with a scalpel" and now I'd be able to. I'm sure other doctors would do so as well because it is one of their staple tools.

:cl: Hopek
add: Boots can now hold a scalpel
/:cl:


morrowwolf (changes requested):
No injectors, scalpel is fine

morrowwolf (approved):

fira (approved):

PR #3284, opened 2023-05-10
Throw Delay

morrowwolf:
# About the pull request

This adds a 0.4 second throw delay that takes effect after throwing twice,

I'm probably overengineering this at this point iunno.

People throwing a knife 20 times in less than a second is not intentional gameplay.

<details>

</details>

# Changelog
:cl: Morrow
balance: Adds a throw delay of 0.4 seconds which takes effect after throwing twice. Two hands, two throws.
/:cl:

### DIFF
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -140,6 +140,9 @@
 #define WIELD_DELAY_VERY_SLOW 10
 #define WIELD_DELAY_HORRIBLE 12
 
+///This is how long you must wait after throwing something to throw again
+#define THROW_DELAY 0.25 SECONDS
+
 //Explosion level thresholds. Upper bounds
 #define EXPLOSION_THRESHOLD_VLOW 50
 #define EXPLOSION_THRESHOLD_LOW 100
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -141,7 +141,7 @@
 #define WIELD_DELAY_HORRIBLE 12
 
 ///This is how long you must wait after throwing something to throw again
-#define THROW_DELAY 0.25 SECONDS
+#define THROW_DELAY (0.4 SECONDS)
 
 //Explosion level thresholds. Upper bounds
 #define EXPLOSION_THRESHOLD_VLOW 50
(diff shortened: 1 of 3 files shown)


Drulikar (approved):
Tested and works. I can no longer instantly kill a runner this way (though its still awfully fast).

Though in testing this, I managed to get the msg_admin_attack message to runtime if you throw a weapon causing the xeno to gib. This should address that (living_defense.dm in `/mob/living/hitby(atom/movable/AM)`):
```
			if(!istype(src,/mob/living/simple_animal/mouse))
				if(src.loc) // If we got gibbed by this attack, we have no loc
					msg_admin_attack("[key_name(src)] was hit by a [O], thrown by [key_name(M)] in [get_area(src)] ([src.loc.x],[src.loc.y],[src.loc.z]).", src.loc.x, src.loc.y, src.loc.z)
				else
					msg_admin_attack("[key_name(src)] was hit by a [O], thrown by [key_name(M)] in [get_area(M)] ([M.loc.x],[M.loc.y],[M.loc.z]).", M.loc.x, M.loc.y, M.loc.z)
```

code/__DEFINES/conflict.dm
```diff
@@ -140,6 +140,9 @@
 #define WIELD_DELAY_VERY_SLOW 10
 #define WIELD_DELAY_HORRIBLE 12
 
+///This is how long you must wait after throwing something to throw again
+#define THROW_DELAY 0.25 SECONDS
```
Zonespace27, line 144:
```suggestion
#define THROW_DELAY (0.25 SECONDS)
```
This is due to PEMDAS fuckery, and how this is actually `0.25 * 10`

code/modules/mob/mob_defines.dm
```diff
@@ -225,6 +225,12 @@
 	var/next_delay_delay = 10 // how much time we wait for next calc of move delay
 	var/move_delay
 
+	///Holds the time when a mob can throw an item next, only applies after two throws, reference /mob/proc/do_click()
+	var/throw_delay = 0
```
Zonespace27, line 229: we have the COOLDOWN_ macros

Zonespace27 (changes requested):

Drulikar:
Disregard my comment about thrown weapon runtimes, it will instead be addressed in #3301

harryob (approved):

PR #3293, opened 2023-05-11
ID Console Fixes

Drulikar:
# About the pull request

This PR fixes a couple runtimes with the ID Console trying to manipulate a non-existent target card. The linked account field is now hidden when there is no target card, is now a number field ranging from 111111 to 999999 (to match automatic generated accounts), and only submits the change after the field loses focus.

I have also mitigated the possibility of players getting the same account number, though this ID console makes no indication an account is unique.

Less runtimes, less number assignment spam, and a slightly more user friendly interface.

<details>

image

</details>

# Changelog
:cl: Drathek
fix: Fixed runtimes with the ID Console
fix: Account numbers are now far less likely to match another person's account
ui: Changed linked account field in the ID Console to a restricted input field
/:cl:

### DIFF
--- a/code/modules/cm_marines/marines_consoles.dm
+++ b/code/modules/cm_marines/marines_consoles.dm
@@ -85,8 +85,9 @@
 		if("PRG_print")
 			if(!printing)
 				if(params["mode"])
-					if(!authenticated)
+					if(!authenticated || !target_id_card)
 						return
+
 					printing = TRUE
 					playsound(src.loc, 'sound/machines/fax.ogg', 15, 1)
 					sleep(40)
@@ -164,7 +165,7 @@
 						return TRUE
 			return FALSE
 		if("PRG_terminate")
-			if(!authenticated)
+			if(!authenticated || !target_id_card)
 				return
 
 			target_id_card.assignment = "Terminated"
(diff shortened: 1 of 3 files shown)


fira:
What happens in case of collision if we fail to find an unused account name?

Drulikar:
> What happens in case of collision if we fail to find an unused account name?

So old code was just random number once.

New code is the same, but inside a loop that can iterate up to 100 times: if we can't find an account with the new number, we break the loop.

Still possible rng for all those attempts it still ends the loop with a duplicate - but this just means the marines with the same account withdraw money from the same account. I mean ultimately in the first place it can be a kind of funny thing to happen - but it felt a little too aggregious that it made zero attempt for uniqueness and it would be very difficult for a player to realize other than going the ATM to print out the transaction record.

Drulikar:
So if we still after 100 random attempts to find a unique number and end up with the same account number: It will create a new account (with a random pin) and just add that to the account list (so now theres two accounts with the same number in all_money_accounts). However, when accessing that account, the first account will be accessed (it just iterates the list in order until account number matches). So all funds will be removed from the original, pin will be used from the original (if security level is greater than 0).

fira (approved):

PR #3299, opened 2023-05-12
Scarring - Xeno Attrition

BeagleGaming1:
# About the pull request

Once a xeno gets damaged, scarring will slowly tick down the xenos health to its max health. It can be healed through hive-tier weeds, such as hive cores or clusters

This forces Xenos to consider their current situation and whether they should risk continuing or returning to a safe position momentarily to heal the scarring. By requiring hive-tier weeds, builder castes will have to be more thoughtful about construction, placement, and the safety of cores and clusters to help other Xenos. Due to the low ratio, while it would not immediately stop Xenos from fighting, they would need to consider how long they have been fighting. The scarring will also allow marines to know that they have contributed, as even if they did not kill the Xeno, they have still caused damage to it, which will take longer to heal. Xenos having attrition will cause them to have to strategize more, and instead of rushing out at once, coordinate on which Xenos will cover while the other Xenos need to heal the scarring. In situations such as the backlines, where there is less likely to be weeds, Xenos will need to get help from other castes to provide weeds or tunnels to allow them to heal from the scarring, and to add a cost to picking off groups of marines alone.

# Changelog

:cl:
balance: Adds scarring - an attrition mechanic for xenomorphs that slowly ticks down their max hp if damaged that can be healed by hive cores and hive clusters
/:cl:


Hopekz:
> > imo if anything there needs to be more feedback on this being a thing for the aliens. Like give them something they can look at to be like "yeah I'm pretty scarred time to go back" vs having to memorize your max health and know the mechanic is there before-hand. Maybe even an alert in chat that says that something changed.
> 
> Scarring has an icon on both your health HUD and health icon, as well as showing you both your max health and your max health after scarring

Good.
My bad I didn't get that from my first glance at the code

morrowwolf:
Heyo, we spoke in DMs about this. This isn't what we're looking for. I do hope you're not too discouraged.

PR #3303, opened 2023-05-12
Yautja armor rebalance

morrowwolf:
# About the pull request

https://docs.google.com/spreadsheets/d/1ywbo-AhXE3bALFb4y0VtaiXWKIlCgbhTIi7mZlvRVsk/edit?usp=sharing

Will update CL before merge as this is going to change a bit during testing likely. In the meantime, the sheets page linked will be updated.

The extreme variance between people who are aware of how predators function versus people who are unaware of how predators function when it comes to fighting is not enjoyable for the majority of the playerbase. Knowledge gap should be important, it should not be the difference between doing good damage and quite literally zero damage. Hopefully by the end of this we will have a predator that requires care and consistency to play rather than a walking tank for new people to plink at while avoiding all the good players.

<details>

</details>

# Changelog

:cl: Morrow
balance: Yautja armor rebalance
/:cl:

### DIFF
--- a/code/datums/pain/pain_yautja.dm
+++ b/code/datums/pain/pain_yautja.dm
@@ -1,5 +1,5 @@
 /datum/pain/yautja
-	max_pain = 200
+	max_pain = 220
 
 	threshold_mild = null
 	threshold_discomforting = null
--- a/code/datums/pain/pain_yautja.dm
+++ b/code/datums/pain/pain_yautja.dm
@@ -1,5 +1,5 @@
 /datum/pain/yautja
-	max_pain = 220
+	max_pain = 240
 
 	threshold_mild = null
 	threshold_discomforting = null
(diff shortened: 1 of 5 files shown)


Delams-The-SM:
While the idea behind PR is quite sound I do worry slightly about increased health from 150 to 200, I guess we will see how it all works out during test merges but still ...

Flerpin:
One thing to note is if the goal is to increase the melee damage of Predators if marines start blasting away preds - just be very careful of that.

Clan sword can already easily two tap crack a limb/hand/foot and two tap crack a chest with a good chance of RNG. Not to mention that combi-stick still remains the meta weapon for a lot of preds as it just straight up negates damage with rng - not unless the goal of the pred is to be a very high damaging glass cannon that can die to some very well placed series of shots.

harryob (approved):

PR #3309, opened 2023-05-13
"Hardcore" Trait

BeagleGaming1:
# About the pull request

If a rifleman has the "Hardcore" trait, they will instantly perma upon death.

It allows marines to practice and show confidence in their skills, by knowing that if they die they cannot get revived. Many marines already do this by DNRing, the only difference is that they could be revived soulless.
It is limited to riflemen only so a limited role that would actually matter if it went perma dead quickly won't lose slots.

# Changelog

:cl:
add: Adds the "Hardcore" trait for Riflemen
/:cl:

### DIFF
--- a/code/modules/character_traits/biology_traits.dm
+++ b/code/modules/character_traits/biology_traits.dm
@@ -113,3 +113,21 @@
 	..()
 
 //datum/character_trait/biology/bad_leg/unapply_trait(mob/living/carbon/human/target) // IMPOSSIBLE
+
+/datum/character_trait/biology/hardcore
+	trait_name = "Hardcore"
+	trait_desc = "One life. One chance. (Rifleman Only)"
+	applyable = TRUE
+	cost = 1
+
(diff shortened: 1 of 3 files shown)


code/modules/character_traits/biology_traits.dm
```diff
@@ -113,3 +113,21 @@
...
+
+/datum/character_trait/biology/hardcore
+	trait_name = "Hardcore"
+	trait_desc = "One life. One chance. (Rifleman Only)"
+	applyable = TRUE
+	cost = 1
+
+/datum/character_trait/biology/hardcore/apply_trait(mob/living/carbon/human/target, datum/equipment_preset/preset)
+	if(target.job != JOB_SQUAD_MARINE)
+		to_chat(target, SPAN_WARNING("Only riflemen can have the Hardcore trait."))
```
fira, line 125: Can't traits also be added via VV? That could be an useful thing to allow this way

harryob (approved):

PR #3316, opened 2023-05-13
Being stunned via slugs or buckshot will now decloak lurkers

usnpeepoo:
# About the pull request

When stunned by ammo such as slugs, heavy revolver, or buckshot, lurkers will be uncloaked.

Previously, lurkers used to have their cloak be temporarily removed with their icon for being stunned and cloaked not having existed. This meant, when a lurker was slugged, buckshotted, etc they would be visible. This PR brings that back but in the scope of lurkers actually losing their invisibility instead of it being an icon bug. This will help with lurkers being stunned repeatedly and ignoring any punishment by continuing to speed away with their cloak. It also is inconsistent with all other forms of cloak (scout, predator) in which their cloak will deactivate if spotted/punished for being spotted.  Hopefully it'll also improve slug usage versus lurkers and make that ammo type a little more viable when having to fend off invisible lurkers.

# Changelog

:cl: Usnpeepoo, Drathek
balance: Lurkers now decloak when stunned by ammo such as slugs or buckshot
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/Lurker.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Lurker.dm
@@ -137,6 +137,17 @@
 		if(LIA && istype(LIA))
 			LIA.end_cooldown()
 
+/mob/living/carbon/xenomorph/lurker/update_canmove()
+    . = ..()
+    if(!canmove && stealth)
+        var/datum/behavior_delegate/lurker_base/behavior = behavior_delegate
+        behavior.force_invisibility_off() // Being stunned will force invisibility off
+
+/datum/behavior_delegate/lurker_base/proc/force_invisibility_off()
+	var/datum/action/xeno_action/onclick/lurker_invisibility/LIA = get_xeno_action_by_type(bound_xeno, /datum/action/xeno_action/onclick/lurker_invisibility)
+	if (LIA && istype(LIA))
+		LIA.invisibility_off()
+
 /datum/behavior_delegate/lurker_base/append_to_stat()
 	. = list()
 	var/invis_message = (invis_start_time == -1) ? "N/A" : "[(invis_duration-(world.time - invis_start_time))/10] seconds."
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/Lurker.dm
```diff
@@ -137,6 +137,17 @@
 		if(LIA && istype(LIA))
 			LIA.end_cooldown()
 
+/mob/living/carbon/xenomorph/lurker/update_canmove()
```
fira, line 140: Uh that's a no for me. 

If you really wanted to do this, should be using `COMSIG_MOB_POST_UPDATE_CANMOVE` - and arguably, this is still backwards and incorrect versus the description of your PR which is to disable on stuns. Don't tie this in to movement.

fira (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/Lurker.dm
fira: For example this would also prevent them from keeping cloak while resting. It also applies to explosions and such. Any sort of immobilization which is not a hard CC will also do it. I strongly suggest not tying this in to movement, because it's not at all what you advertised in the PR.

Diegoflores31:
i think this change will rather make Stun combat almost mandatory for marines . if you want decloaks i think a better solution would be a Rng chance of decloak per bullet similar to how pred cloak works.

code/modules/mob/living/carbon/xenomorph/castes/Lurker.dm
Drulikar: Looking again, a more precise proc to intercept would be /mob/living/proc/handle_stunned() e.g.
`/mob/living/carbon/xenomorph/lurker/handle_stunned()` but yes either registering to COMSIG_MOB_POST_UPDATE_CANMOVE or COMSIG_LIVING_SET_EFFECT are also options (though the former signal IMO is worse than what is already being done here because of signal overhead).

fira:
To clarify all I'm asking to remove the DNM tag is that at the very least this be made to either do what it says, or say what it does. Past that it's a matter for balance people, even though I think having this tied to movement is an overall very bad thing

usnpeepoo:
> To clarify all I'm asking to remove the DNM tag is that at the very least this be made to either do what it says, or say what it does. Past that it's a matter for balance people, even though I think having this tied to movement is an overall very bad thing

Working on it

usnpeepoo:
will figure this out later

PR #3318, opened 2023-05-13
Fix huggers dying while attached to a host before they can impregnate it

Emanthealmighty:
# About the pull request

Fixes huggers occasionaly dying while attached to a host before they can impregnate it by preventing them from dying at all while attached to the host until they have impregnated it.

Gets rid of those not-so-rare occurrences where hugging a human would result in the hugger not actually infecting it, requiring an additional hugger to finish the job.

# Changelog

:cl:
fix: Fixed huggers occasionally dying before they can impregnate a host.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
@@ -77,6 +77,9 @@
 	addtimer(CALLBACK(src, PROC_REF(check_turf)), 0.2 SECONDS)
 	if(stat == CONSCIOUS && loc) //Make sure we're conscious and not idle or dead.
 		go_idle()
+	if(attached == TRUE)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
```diff
@@ -77,6 +77,9 @@
 	addtimer(CALLBACK(src, PROC_REF(check_turf)), 0.2 SECONDS)
 	if(stat == CONSCIOUS && loc) //Make sure we're conscious and not idle or dead.
 		go_idle()
+	if(attached == TRUE)
```
harryob, line 80:
```suggestion
	if(attached)
```

harryob (changes requested):

PR #3319, opened 2023-05-14
adds more observer action buttons

harryob:
this allows for non-verb usage of view crew manifest, hive status and joining the hunt, along with keybinds for all of them too

:cl:
qol: observer hud buttons for view crew manifest, hive status and joining the hunt (if you're a pred), along with keybinds for joining the hunt
fix: you can now hide the buttons at the bottom of the ghost hud with f12
/:cl:


Drulikar (changes requested):
Everything currently seems functional, though your PR notes that all of these actions should have bindable hotkeys and they do not.

Crew status is not bindable.
Hive status is bindable, but only works as a xeno.

Ideally join as xeno button should be after crew and hive status buttons so it will be adjacent to all join buttons.

Join as pred/xeno/ert are bindable and working.
Observer buttons for hunt, hive status, and joining pred/xeno/ert are working.

Also I recommend merging with master - this is based on such an old version that it is untestable somereason - DB error when editing character and various initialization will fail.

Drulikar (approved):

sleepynecrons (approved):
sproites

PR #3328, opened 2023-05-15
Event toggle for Permadeath

BeagleGaming1:
# About the pull request

Adds a toggle that allows staff to give everyone (except synths) the hardcore trait

Admin event funny

# Changelog

:cl:
admin: Adds an event toggle for staff to toggle "Hardcore", causing all humans to be unrevivable on death
/:cl:


code/modules/admin/tabs/admin_tab.dm
```diff
@@ -796,3 +796,21 @@
...
+
+/client/proc/toggle_hardcore_perma()
+	set name = "Toggle Hardcore"
+	set category = "Admin.Flags"
+
+	if(!admin_holder || !check_rights(R_MOD, FALSE))
+		return
+
+	if(!SSticker.mode)
+		to_chat(usr, SPAN_WARNING("A mode hasn't been selected yet!"))
```
Drulikar, line 808:
```suggestion
		to_chat(usr, SPAN_WARNING("A mode hasn't been selected yet!"))
		return
```

code/modules/mob/living/carbon/human/death.dm
```diff
@@ -43,8 +43,9 @@
 		return
 	GLOB.alive_human_list -= src
 	if(!gibbed)
-		if(HAS_TRAIT(src, TRAIT_HARDCORE))
-			status_flags |= PERMANENTLY_DEAD
+		if(HAS_TRAIT(src, TRAIT_HARDCORE) || MODE_HAS_TOGGLEABLE_FLAG(MODE_HARDCORE_PERMA))
+			if(!(species.flags & IS_SYNTHETIC)) // Synths wont perma
```
Drulikar, line 47: This will make it so synths cannot perma in TRAIT_HARDCORE as well. Is that intended or an oversight?

code/modules/admin/tabs/admin_tab.dm
```diff
@@ -796,3 +796,21 @@
...
+
+	if(!MODE_HAS_TOGGLEABLE_FLAG(MODE_HARDCORE_PERMA))
+		if(alert("Are you sure you want to toggle Hardcore mode on? This will cause all humans to instantly go perma on death.",,"Yes","Cancel") != "Yes")
```
Drulikar, line 811:
```suggestion
		if(!MODE_HAS_TOGGLEABLE_FLAG(MODE_HARDCORE_PERMA) && tgui_alert(usr, "Are you sure you want to toggle Hardcore mode on? This will cause all humans to instantly go perma on death.", "Confirmation", list("Yes", "Cancel")) != "Yes")
```
Alternatively change the string like message_admins below does if you want to confirm disable too.

Drulikar (changes requested):

morrowwolf:
Alright concerns handled.

nauticall:
Would recommend you state exactly _what_ the "hardcore" trait does in both this PR and the changelog so that nobody gets confused by what it means.

Drulikar (changes requested):
Please add `status_flags &= ~PERMANENTLY_DEAD` somewhere in the admin rejuvenation process e.g. living_health_procs.dm: `/mob/living/proc/heal_all_damage()` or `/mob/living/proc/rejuvenate()` since there's nothing that unsets this status other than temporarily during synth surgery (as in say admins had it on accidentally, someone dies, they get ahealed, then they die and perma again because its still set on them).

Otherwise seems fine.

Drulikar (approved):

PR #3333, opened 2023-05-16
ensures people always have at least a set of keybinds

harryob:
privates keep getting immobilized because they don't have any keybinds and not realising they can reset them in the hotkeys menu - i do not need to tell anyone that this is not ideal new player experience

:cl:
fix: new players will now no longer have no hotkeys, and be unable to move or talk, because that sucks
/:cl:


morrowwolf:
I'm TM'ing this for now.

Issue #1674, opened 2022-11-22
Hidden verbs are visible


ThePiachu:
## Description

[According to Fira](https://github.com/cmss13-devs/cmss13/pull/1646#pullrequestreview-1190256937) various Xeno Verbs like Hedgehog's `fire spikes`, `shed spikes` and all that shold be hidden, but they don't appear to be:

![image](https://user-images.githubusercontent.com/964559/203437390-da8bf292-219a-449b-b0d4-0be5277dfa11.png)

So I'm guessing there is some issue with the `Hidden` tag?

## Difference between expected and actual behavior

Apparently we aren't supposed to see various Xeno verbs and similarly hidden tags, but we do.

## Steps to reproduce

1) Play a Hedgehog Ravager
2) Check Alien verb category
3) See verbs that shouldn't be there

## Issue bingo (not required to be all checked)

- [x] Issue could be reproduced at least once
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

fira:
The difference between visibles and not is whether `category` is set. If category is unset, they won't show up in statpanel.

The bug part is that all of these should be `hidden` - there is actually a check for it in `init_verbs` and `add_verbs` but for some reason the check to `hidden` returns false and they are added to statpanel anyway.

Huffie56:
this seem to be for every strain older verb aren't cleared....

ItsVyzo:
This is still an issue

PR #3336, opened 2023-05-16
makes the join pred ghost button only appear during pred rounds

harryob:
i forgot this from the original pr after refactoring it into /Login()

:cl:
qol: whitelisted predators now only see the option to join pred rounds as a ghost in the top left when it is a pred round
/:cl:

### DIFF
--- a/code/__DEFINES/dcs/signals/signals_global.dm
+++ b/code/__DEFINES/dcs/signals/signals_global.dm
@@ -47,3 +47,6 @@
 
 // Used for smothering fires upon weather event start/stop
 #define COMSIG_GLOB_WEATHER_CHANGE "!weather_event_changed"
+
+/// From /datum/admins/proc/force_predator_round()
+#define COMSIG_GLOB_PREDATOR_ROUND_TOGGLED "!predator_round_toglged"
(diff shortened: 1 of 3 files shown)


morrowwolf (approved):

Huffie56:
sorry to bother you but i remember seeing this issue and since it's related in case you can had a fix for it?
don't know if it's still active...
https://github.com/cmss13-devs/cmss13/issues/1674
thanks

harryob:
> sorry to bother you but i remember seeing an issue and since it's related in case you can had a fix for it?
> don't know if it's still active...
> #1674
> thanks

not related

Drulikar:
> sorry to bother you but i remember seeing an issue and since it's related in case you can had a fix for it? don't know if it's still active... #1674 thanks

I don't think thats related in any way to this PR. However I would guess the fix for that would be to update the logic in verb_helpers.dm `/proc/add_verb(client/target, verb_or_list_to_add)` to check if the verb is hidden. `/client/proc/init_verbs()` checks for hidden, but `/proc/add_verb(client/target, verb_or_list_to_add)` does not. But may have unintended consequences that also need to be rectified.

Drulikar (approved):

PR #3339, opened 2023-05-16
the midway

harryob:
i am primarily opening this pr so i make sure to actually do it and not put it off forever and ever and ever

this will be a third ds to replace the whole in our hearts left by the vc bay. currently, it requires beacons deployed on the ground which the ~~tadpole~~ midway can land around

a whole bunch of the credit goes to various contributors on tgmc, primarily bravemole, Eskjjlj but also a bunch of others

yes this pr touches a ton of code it probably doesn't need to - i'll clean it up i promise


Watermelon914:
This PR has been inactive for some time, feel free to re-open (as a new PR so that it appears at the top of the tracker) when you decide to work on this again.

PR #3360, opened 2023-05-18
Zombies now have their own language and don't understand english

Hopekz:
Language processing happens in the mind; not the body. So just because you have the ability to speak English because you have a human body doesn't mean that you have the skill to do so.

Gameplay reason:
All zombies have headsets and they immediately know where survivors are going even before they get there. 
"I'm going to CIC" = a group of zombies already waiting there. 
Would be more fun an engaging if they "found" them and then swarmed them versus already waiting there for them.

I could have went a step further and went for the announcements too but I didn't feel like it with this PR

:cl: Hopek
balance: Zombies now have their own language and don't understand English
/:cl:

### DIFF
--- a/code/datums/diseases/black_goo.dm
+++ b/code/datums/diseases/black_goo.dm
@@ -106,6 +106,7 @@
 	if(human && human.loc)
 		if(human.stat == DEAD)
 			human.revive(TRUE)
+			human.remove_language(LANGUAGE_ENGLISH) // You lose the ability to understand english. Language processing is handled in the mind not the body.
 			var/datum/species/zombie/zombie_species = GLOB.all_species[SPECIES_ZOMBIE]
 			zombie_species.handle_alert_ghost(human)
 		playsound(human.loc, 'sound/hallucinations/wail.ogg', 25, 1)
--- a/code/datums/diseases/black_goo.dm
+++ b/code/datums/diseases/black_goo.dm
@@ -213,7 +213,7 @@
 	reagents.add_reagent("antiZed", 30)
 
 /datum/language/zombie
-	name = LANGUAGE_ZOMBIE
+	name = "Zombie"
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/human/species/zombie.dm
```diff
@@ -24,6 +24,7 @@
 	stun_reduction = 10
 	knock_out_reduction = 5
 	has_organ = list()
+	set_languages(list(LANGUAGE_ZOMBIE))
```
Zonespace27, line 27: set this on init

Hopekz: > set this on init

This is how all the other species do it I think

Zonespace27: image

it's not

Hopekz: Implemented; my bad I was tired when I made this

code/datums/diseases/black_goo.dm
```diff
@@ -213,16 +214,16 @@
...
-	ask_verb = "groans"
-	exclaim_verb = "groans"
-
-	key = "4"
+	desc = "A growling, guttural method of communication, only Zombies seem to be capable of producing these sounds."
+	speech_verb = "growls"
+	ask_verb = "grumbles"
+	exclaim_verb = "snarls"
+	color = "monkey"
+	key = "h"
```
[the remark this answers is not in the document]
Hopekz: > you overrode the language and key with the ones from the monkey language

I replaced the key with h because you can't get monkey and zombie at the same time since being a zombie would wipe out your monkey if you ever were to convert (assuming monkeys even can) and the monkey text looked cooler imo otherwise that's really the only reason

harryob (approved):

Issue #3371, opened 2023-05-19
bug LV light of plants/torches contiue to emit light after being destroyed/melted.


Huffie56:
### Testmerges

not relevant

### Description of the bug

When you destroy/melt plant's and torches that emit light the light stay after their destruction.

### What's the difference with what should have happened?

if the source of the light is destroy the light  should stop being produce.

### How do we reproduce this bug?

1.melt a luminous plant
2.wait for is destruction
3.notice how the area is still producing light.
...

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Huffie56:
old bug coming from githlab https://gitlab.com/cmdevs/colonial-warfare/-/issues/763

Huffie56:
i am working on a fix. those plant's are just badly made i am going to change a bunch of them code wise...

PR #3381, opened 2023-05-20
Fix bug that didn't remove luminosity after the object that emit them was destroy.

Huffie56:
# About the pull request

fixes https://github.com/cmss13-devs/cmss13/issues/3371

added a check if the entity is destroy that reset luminosity to 0.
adding to the code their own entity instead of doing change in dmm file for :
the cart and the bio luminescent tree also did it for glow goo that give acid damage.
give the proper entity name instead of doing change in dmm files for the light stick 

remove a bug.
less useless code in dmm files...

i tested on local server it work very fine.
<details>

</details>

# Changelog
:cl:

fix: remove the luminosity when torch and plant that produce the light are destroy.

/:cl:

### DIFF
--- a/code/game/objects/effects/decals/cleanable/misc.dm
+++ b/code/game/objects/effects/decals/cleanable/misc.dm
@@ -31,6 +31,16 @@
 	icon_state = "dirt"
 	mouse_opacity = MOUSE_OPACITY_TRANSPARENT
 
+/obj/effect/decal/cleanable/dirt/greenglow
+	name = "glowing goo"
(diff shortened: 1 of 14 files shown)


Huffie56:
going to create way more specific item instead of having them instance in strong DMM this will create issues if it's not the case already

Huffie56:
saw a lot of stuff changed in DMM that shouldn't will maybe make a PR for it later...

harryob (approved):
great pr

PR #3386, opened 2023-05-21
Xeno directional attack no longer friendly fires their own resin walls and doors

Hopekz:
dreamseeker_M0U60T6yV6

:cl: Hopek
fix: Xeno Directional attack no longer friendly fires their own walls and doors. You need to actually click on the sprite if you mean to destroy it
/:cl:

### DIFF
--- a/code/_onclick/xeno.dm
+++ b/code/_onclick/xeno.dm
@@ -2,7 +2,7 @@
 	Xenomorph
 */
 
-/mob/living/carbon/xenomorph/UnarmedAttack(atom/target, proximity, click_parameters, tile_attack = FALSE)
+/mob/living/carbon/xenomorph/UnarmedAttack(atom/target, proximity, click_parameters, tile_attack = FALSE, ignores_resin = FALSE)
 	if(lying || burrow) //No attacks while laying down
 		return FALSE
 	var/mob/alt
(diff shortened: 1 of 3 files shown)


Segrain:
> their own walls and doors

What if it is another hive's wall?

Hopekz:
> > their own walls and doors
> 
> What if it is another hive's wall?

They can click on the sprite like in the example gif

code/_onclick/xeno.dm
```diff
@@ -28,6 +28,12 @@
 			break
 		if (target == T && alt)
 			target = alt
+		if (T && ignores_resin) // Will not target resin walls and doors if this is set to true. This is normally only set to true through a directional attack.
+			if(istype(T, /obj/structure/mineral_door/resin))
+				return FALSE
+			if(istype(T, /turf/closed/wall/resin))
+				return FALSE
```
harryob, line 35:
```suggestion
		if (T && ignores_resin) // Will not target resin walls and doors if this is set to true. This is normally only set to true through a directional attack.
			if(istype(T, /obj/structure/mineral_door/resin))
				var/obj/structure/mineral_door/resin/attacked_door = T
				if(hivenumber == attacked_door.hivenumber)
					return FALSE
			if(istype(T, /turf/closed/wall/resin))
				var/turf/closed/wall/resin/attacked_wall = T
				if(hivenumber == attacked_wall.hivenumber)
					return FALSE
```

harryob (changes requested):

PR #3394, opened 2023-05-22
adds motion detector for PMC faction

VashVandez:
Adds a modified motion detector to detect signals not affiliated with FACTION_PMC. 

# About the pull request

Gives the PMCs a faction-aligned motion detector. 

The other factions like the UPP, Dutch's Dozen, etc., have a motion detector to call their own; let us have one, too! Will allow PMCs to potentially have motion detectors of their own without having to worry about their own squadmates making them go off. 

# Changelog

modified motion_detector.dm


Foxtrot13224:
Might want to fix the changelog

DOC-surge:
yes, I like this idea some test on a private server with items and the MD will react to PMCs even if they are holding one which is funny but annoying.  so please make this real

Diegoflores31:
fix your changelog

harryob:
your changelog needs to include any player facing changes, but there are no player facing changes here, so you're fine actually

PR #3401, opened 2023-05-23
Shotgun and buckshot tweaks

usnpeepoo:
# About the pull request

This PR increases buckshots damage from 65, to 125 and gives it 5 more AP. Shotgun has also been tweaked to have a longer fire delay and a reduction from 9, to 4 rounds (5 when pumped and another shell is loaded).

Shotguns are in a weird place ever since the double shotgun removal. Buckshot had been its main purpose, and was basically balanced around a person using two shotguns. I believe this PR would put shotguns into a specific role, where constant usage isn't encouraged, but it would remain as a powerful side choice that can either be used as slugs or through powerful point blanks. By removing the excessive amounts of shells (which isn't lore accurate), lowering the fire delay, and making buckshot hit harder, I believe shotguns would fit their role better instead of being in a limbo state where it isn't really useful aside being normally used against 3-4 castes. 
Basically, this lowers the PB requirement for most kills on xenos by 1~. IE a warrior will die in 2 PBs instead of 3.

# Changelog

:cl:
balance: Shotguns have a lowered fire delay and maximum shells
balance: Buckshots damage and AP have been increased
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1197,10 +1197,10 @@
 	accuracy_var_high = PROJECTILE_VARIANCE_TIER_5
 	accurate_range = 4
 	max_range = 4
-	damage = 65
+	damage = 125
 	damage_var_low = PROJECTILE_VARIANCE_TIER_8
 	damage_var_high = PROJECTILE_VARIANCE_TIER_8
-	penetration = ARMOR_PENETRATION_TIER_1
+	penetration = ARMOR_PENETRATION_TIER_2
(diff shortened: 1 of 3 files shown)


MusselMain:
will the ammo capacity change only affect the m37A2, or will it also affect the MK221, xm38, custom built shotgun, HG37-12, and HG37-17. because if it did, it could be fairly problematic for certain weapons, like the semiauto shotguns

morrowwolf (commented):
I appreciate the work and thought you put into this but I have no interest in this change at this time.

PR #3404, opened 2023-05-23
Removes Baton slugs from regular accessibility.

LynxSolstice:
# About the pull request

Makes the riot control launcher able to launch HIRR Slugs and changes it to use the M79 sounds rather than the legacy ones. Adds packets to the riot control cabinets and removes them from the regular vendors.

Moves the baton slugs to a policing role, rather than a meme ammo type for the M79. 

Runs, compiles, the baton slugs can be fired out of the grenade launcher.

# Changelog
:cl:LynxSolstice
add: HIRR slugs moved to the MP armory, grenade launcher allowed to fire them now. M81 grenade launcher now uses the M79 handling and firing sounds.
balance: Removes the HIRR slugs from all vendors.
/:cl:


Segrain:
"Meme ammo type"? Back when M79 was being designed, slugs were [a major part of] _its selling point_.

MusselMain:
'let's remove the one ammo type that is reusable when low on ammo, and is the only ammo type that allows M79 users to protect themselves at close range, and give it to the MP's who already have countless non lethals that perform better including the riot MK221, which is more effective, and specifically built for this purpose, effectively ruining the few players who use M79 only. that's a great idea, that definitely won't remove yet another stun option from marines'

morrowwolf:
I don't think this is the right way.

I will say I'm not a fan of the M79 as it currently functions.

PR #3411, opened 2023-05-24
Broken bones can now cause IB

BeagleGaming1:
# About the pull request

Unsplinted broken bones can cause IB in a similar way to organ damage

It only makes sense that loose bones in your limbs will cut you up.

# Changelog

:cl:
balance: Unsplinted broken bones that are not hands or feet can now cause IB
balance: bloodbag capacity raised to 300u
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life/handle_organs.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_organs.dm
@@ -25,6 +25,9 @@
 					var/damage = rand(3,5)
 					I.take_damage(damage)
 					pain.apply_pain(damage * PAIN_ORGAN_DAMAGE_MULTIPLIER)
+				if(E.is_broken() && prob(15))
+					E.add_bleeding(E, TRUE, rand(3,5))
+					custom_pain("You feel broken bones cutting at you in your [E.display_name]!", 1)
 
 	if(!lying && !buckled && prob(2))
 		var/left_leg_crippled = FALSE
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/life/handle_organs.dm
```diff
@@ -25,6 +25,11 @@
 					var/damage = rand(3,5)
 					I.take_damage(damage)
 					pain.apply_pain(damage * PAIN_ORGAN_DAMAGE_MULTIPLIER)
+				if(E.is_broken() && prob(15))
```
morrowwolf, line 28: Change prob to like 3

morrowwolf (changes requested):

morrowwolf (approved):
code and balance approved

MistChristmas:
If this is here to stay might we add IBs visible to the normal health scanner. So that we can see when someone has 5 IBs and where. Cause playing the guessing game of which frac(s) have the IB(s) isn't exactly a fun time.

BeagleGaming1:
> If this is here to stay might we add IBs visible to the normal health scanner. So that we can see when someone has 5 IBs and where. Cause playing the guessing game of which frac(s) have the IB(s) isn't exactly a fun time.

Right click, "Check Status", it shows you exactly which limb has the IB

morrowwolf (changes requested):
Let's do non-hands/non-feet only and drop prob to 2.

PR #3418, opened 2023-05-25
Second attempt to fix timers runtimes for overhead text system

### LINKED ISSUE #2042: addtimer called with a callback assigned to a qdeleted object
### Testmerges

???

### Description of the bug

During the UPP event we got a lot of these events:

```
runtime error: addtimer called with a callback assigned to a qdeleted object. In the future such timers will not be supported and may refuse to run or run with a 0 wait
proc name: stack trace (/proc/stack_trace)
  source file: unsorted.dm,1817
  usr: (src)
  src: null
  call stack:
stack trace("addtimer called with a callbac...")
 addtimer(/datum/callback (/datum/callback), 5, 0, null, "code/modules/maptext_alerts/sc...", 105)
 (/atom/movable/screen/text/screen_text/command_order): after play(EOD-501 (/client))
/datum/callback (/datum/callback): InvokeAsync()
Timer (/datum/controller/subsystem/timer): fire(0)
Timer (/datum/controller/subsystem/timer): ignite(0)
Master (/datum/controller/master): RunQueue()
Master (/datum/controller/master): Loop(2)
Master (/datum/controller/master): StartProcessing(0)
```

Shouldn't runtime

???

### Issue Bingo

Drulikar:
# About the pull request

This PR refactors how on screen alerts are handled somewhat, and adds gc_deleted to clients when they get deleted/disconnected. Now overhead messages should properly handle queuing of messages, and hopefully quit sending messages repeatedly.

Should hopefully fix #2042 

# Changelog
:cl: Drathek
fix: Fix runtime with a timer to fade out on screen orders again?
/:cl:


fira:
But clients don't use the traditional softGC system. They're scheduled for immediate hard deletion when they hit qdel...

Drulikar:
> But clients don't use the traditional softGC system. They're scheduled for immediate hard deletion when they hit qdel...

I don't know for sure, but I'd guess a client that either disconnects or gets gibbed in between the time they received a message and before it resolves (fades out and deletes). The previous fix for this merely tested !player; but the timer is tested QDELETED on that same variable.

Drulikar:
This code needs a refactor actually

fira (approved):

PR #3419, opened 2023-05-25
increases the synthetic piloting skill

Katskan:
# About the pull request

Fixes the oversight when we changed to TGUI pilot interface which made it so pilot level 2 is required for flybys and optimized flight. Confirmed with maintaining management that it was an oversight. 

Lets synthetics perform duties like medevac and fulton without it being jank, by allowing flybys.

</details>

# Changelog

:cl:
balance: synthetic pilot skill increased from 1 to 2
/:cl:


harryob (changes requested):
fix pr title, say what the pr is accomplishing as succinctly as possible (the pr title is used as the commit to the repo)

harryob (approved):

PR #3425, opened 2023-05-26
Fix Rare Emergency Locker Hard Delete

### LINKED ISSUE #3423: Flaky test create_and_destroy: /obj/structure/closet/emcloset/legacy hard deleted 1 times out of a total del count of 1
Flaky tests were detected in [this test run](https://github.com/cmss13-devs/cmss13/actions/runs/5085857408/attempts/1). This means that there was a failure that was cleared when the tests were simply restarted.
Failures:
```
create_and_destroy: /obj/structure/closet/emcloset/legacy hard deleted 1 times out of a total del count of 1 at code/modules/unit_tests/create_and_destroy.dm:124
```

Drulikar:
# About the pull request

This PR fixes a hard delete caught by the flaky test checker where the emergency locker would rarely pick its delete case, QDEL itself in init (though it didn't report this as the error, instead it was unhappy there were items with it in its loc). The INITIALIZE_HINT_QDEL seems to fix this. Also, the test is now forced to pick "both" since we don't want randomness in unit_tests, but forcing the "delete" case did reliably reproduce the fail case.

I also added code to null out /atom/movable/screen.master in case that ever would have been an issue.

Fixes #3423 

# Changelog
:cl: Drathek
fix: Fixed the emergency locker sometimes hard deleting
/:cl:


code/_onclick/hud/map_popups.dm
```diff
@@ -62,6 +62,11 @@
 	layer = GAME_PLANE
 	plane = GAME_PLANE
 
+/atom/movable/screen/Destroy()
```
Zonespace27, line 65: move this to be above `/atom/movable/screen/map_view`

code/_onclick/hud/map_popups.dm
```diff
@@ -62,6 +62,11 @@
...
+	master = null
+	//QDEL_NULL(hud) // Not currently ever used; not sure if it would just need to be nulled or qdel
```
Zonespace27, line 67:
```suggestion
	hud = null
```
The HUD owns this object, not the other way around, so we null instead of qdeling

Zonespace27 (changes requested):

Zonespace27 (approved):
lgtm

fira (approved):

PR #3428, opened 2023-05-26
Vending Refactors - Vending Categories and Multi Vending

fira:
# About the pull request

This does two things:
 * Refactors vending categories to use associative lists instead of bitflags. The big takeaway is that this lets us have arbitrary categories and not be limited to the 24 bitflags anymore. It also eliminates horrible left/right pouches logic, you just get two.
 * Adds the possibility to vend multiple items with a single button/category hit. This allows for bundling items, or just reducing the amount of buttons to mash when equipping gear.

Some big thanks to Sol for motivating and helping design this, and Drathek for figuring out icon issues.

People can not only put as many vending buttons as they want (categories/bitflags refactors), they can also put as little of them as they want (multi-vend for regroupping items under one button)

# Changelog
I think I accidentally axed CE's right to have two different sets of standard and combat pouches, but let's pretend that didn't happen.

No actual player facing changes.... yet.


code/game/machinery/vending/cm_vending.dm
```diff
@@ -1063,10 +1067,12 @@ GLOBAL_LIST_INIT(cm_vending_gear_corresponding_types_list, list(
 		var/priority = myprod[priority_index]
 
 		var/obj/item/I = item_ref
```
Drulikar, line 1069: Why not just have
`var/item_ref = myprod[3]`
be 
`var/obj/item/item_ref = myprod[3]`
since you've already proven that it cast to var/obj/item/I can pass the islist test. That'll also mean you don't have a single character var too.

Huffie56:
so i saw in your code that it handle giving the headset corresponding to the proper squad...
would it then be ok to remove the different subtype of for the surplus uniform vendor in there they hold squad specific headset?
they aren't even used on any maps...
https://github.com/fira/cmss13/blob/multivend/code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm

image

code/game/machinery/vending/cm_vending.dm
```diff
@@ -1164,82 +1168,76 @@ GLOBAL_LIST_INIT(cm_vending_gear_corresponding_types_list, list(
...
+		if(user.client?.prefs && (user.client?.prefs?.toggle_prefs & TOGGLE_VEND_ITEM_TO_HAND))
+			if(vendor.Adjacent(user))
+				user.put_in_any_hand_if_possible(new_item, disable_warning = TRUE)
+
 /proc/handle_vend(obj/structure/machinery/cm_vending/vendor, list/listed_products, mob/living/carbon/human/vending_human)
 	if(vendor.vend_flags & VEND_USE_VENDOR_FLAGS)
 		return TRUE
-	var/can_buy_flags = listed_products[4]
-	if(!(vending_human.marine_buy_flags & can_buy_flags))
+	var/buying_category = listed_products[4]
```
Drulikar, line 1234: This would probably involve changing a lot of files, but would be neat if instead of 4 it could be a define that says what index we're accessing containts e.g. listed_products[LISTED_PRODUCT_CATEGORY], etc. for the index usage.

code/game/machinery/vending/cm_vending.dm
fira: well that was existing code so i rolled with it but that's very goofy indeed, will change

code/game/machinery/vending/cm_vending.dm
fira: There's a problem with this in that the lists have different formats based on vendor type
I think this is out of scope for this PR in any case, would need to edit a lot more stuff and pray getting them all since there's no easy way to search for this

fira:
> so i saw in your code that it handle giving the headset corresponding to the proper squad... would it then be ok to remove the different subtype of for the surplus uniform vendor in there they hold squad specific headset? they aren't even used on any maps... https://github.com/fira/cmss13/blob/multivend/code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
> 
> image
> 
> i was hesitant to remove them in a previous PR because i didn't know how headset where handle...

i didn't touch the headsets assignment - i'm not sure how we can make this simpler honestly, this is just used for the per-squad common vendors

fira:
> So I know you haven't set up the one button in the vendor yet, but I will note that currently you can't spend any points: 

Yeah i'm a doofus, fixed that!

I didn't set up "all-in-one"  buttons because right now the level of granularity required by some vendors makes the categorys hard to actually regroup in a consistent fashion... It could be done in a more loose way vendor per vendor, or with multi-categories linked to the multi-items vended, but any kind of such rework right now requires an ASTRONOMICAL amount of vendor files editing (i gave up at ~15% of them)

code/__DEFINES/vendors.dm
```diff
@@ -1,30 +1,26 @@
...
-#define MARINE_CAN_BUY_COMBAT_SHOES (1<<17)
-#define MARINE_CAN_BUY_COMBAT_HELMET (1<<18)
-#define MARINE_CAN_BUY_COMBAT_ARMOR (1<<19)
-#define MARINE_CAN_BUY_COMBAT_R_POUCH (1<<20)
-#define MARINE_CAN_BUY_COMBAT_L_POUCH (1<<21)
-
-#define MARINE_CAN_BUY_KIT (1<<22)
-
-#define MARINE_CAN_BUY_ALL ((1<<23) - 1)
+#define MARINE_CAN_BUY_ALL list(MARINE_CAN_BUY_UNIFORM = 1, MARINE_CAN_BUY_SHOES = 1, MARINE_CAN_BUY_HELMET = 1, MARINE_CAN_BUY_ARMOR = 1, MARINE_CAN_BUY_GLOVES = 1, MARINE_CAN_BUY_EAR = 1, MARINE_CAN_BUY_BACKPACK = 1, MARINE_CAN_BUY_POUCH = 2, MARINE_CAN_BUY_BELT = 1, MARINE_CAN_BUY_GLASSES = 1, MARINE_CAN_BUY_MASK = 1, MARINE_CAN_BUY_ESSENTIALS = 1, MARINE_CAN_BUY_SECONDARY = 1, MARINE_CAN_BUY_ATTACHMENT = 1, MARINE_CAN_BUY_MRE = 1, MARINE_CAN_BUY_ACCESSORY = 1, MARINE_CAN_BUY_COMBAT_SHOES = 1, MARINE_CAN_BUY_COMBAT_HELMET = 1, MARINE_CAN_BUY_COMBAT_ARMOR = 1, MARINE_CAN_BUY_KIT = 1, MARINE_CAN_BUY_DRESS = 99)
```
mullenpaul, line 23: Formatting of the list would be good

code/game/machinery/vending/cm_vending.dm
```diff
@@ -1139,8 +1142,8 @@ GLOBAL_LIST_INIT(cm_vending_gear_corresponding_types_list, list(
 		var/list/myprod = ui_listed_products[i] //we take one list from listed_products
 		var/prod_available = FALSE
 		var/p_cost = myprod[2]
-		var/avail_flag = myprod[4]
-		if(points >= p_cost && (!avail_flag || buy_flags & avail_flag))
+		var/category = myprod[4]
+		if(points >= p_cost && (!category || ((category in H.marine_buyable_categories) && (H.marine_buyable_categories[category]))))
```
mullenpaul, line 1146: Single var H for human

mullenpaul (commented):
Code looks good - not had a chance to test it out

code/__DEFINES/vendors.dm
fira: Can try to do it but it's just gonna be super long vertically instead of horizontally.. And i try to avoid messing with whitespaces in defines :\

PR #3433, opened 2023-05-27
less medic vendors on WO

harryob:
man what the hell

:cl:
fix: there's less random vendors on WO
/:cl:


Drulikar (approved):
Though `/turf/open/jungle/impenetrable` could have been used instead of overriding turf/open/jungle

PR #3441, opened 2023-05-28
SL gets Signal Flare

ghostsheet:
# About the pull request

Add 1 pack of signal flares to SL essential Kit

Turns out SL calling in CAS is good

I've also considered adding the Signal Flare pack as a purchasable point based item for SL however
It seems that the Signal Flare pack has intentional scarcity as no one can buy it even RTO/Fireteam Leads,

# Changelog

:cl: ghostsheet
add: Added a M89-S Signal Flare pack to SL essential kit.
/:cl:


morrowwolf (changes requested):
Fine with the signal flares.

Let's leave the M4RA kit, we both may not use it but I do believe some take it for the barrel/AP mags.

ghostsheet:
Done

morrowwolf (approved):

PR #3442, opened 2023-05-28
Bugfix: Dropship attachment point layers

ghostsheet:
# About the pull request

Dropship attachment points for (medevac etc.)
No longer goes over weed and resin structure

attachment points

# Changelog

:cl: ghostsheet
fix: Dropship attachment points now goes under weed and resin.
/:cl:

### DIFF
--- a/code/modules/dropships/attach_points/templates.dm
+++ b/code/modules/dropships/attach_points/templates.dm
@@ -83,6 +83,7 @@
 /obj/effect/attach_point/crew_weapon
 	name = "crew compartment attach point"
 	base_category = DROPSHIP_CREW_WEAPON
+	plane = FLOOR_PLANE
 
 /obj/effect/attach_point/crew_weapon/dropship1
 	ship_tag = DROPSHIP_ALAMO


fira (changes requested):
I don't think weeds should cover them, we usually keep the gameplay functional elements visible

Should avoid touching the planes for layering only too, I'm sure this is a problem that can be solved at layering level

Watermelon914:
> I don't think weeds should cover them, we usually keep the gameplay functional elements visible
> 
> Should avoid touching the planes for layering only too, I'm sure this is a problem that can be solved at layering level

I don't actually think it should be an issue. Gameplay functional elements being layered over weeds is fine, but if they're visually jarring, then we need to make them not so.

In this case, weeds covering dropship attachment points doesn't really seem like a big deal, especially since the goal is for them to be part of the floor. Weeds should cover everything that's supposed to be on the floor plane/layer, and if they don't, then they need to be resprited so that they don't look like they're on the floor plane/layer

PR #3447, opened 2023-05-28
tacmap accounts for chestburst + asl

### LINKED ISSUE #3439: Tacmap: aSL's do not change a players' tacmap icon
### Testmerges

#3088, #3185, #3298, #3424, #3430, #3271, #3392

### Description of the bug

When using the tactical map on the CIC overwatch consoles, the icons of marines will not update to reflect their current squad status as acting Squad Leaders. 

For example, if you set a rifleman as the acting squad leader, the map will continue to display him as a rifleman, and the actual squad leader will continue to appear as the SL on the tacmap - which can get very confusing.

When switching aSL to someone else in a squad, their tacmap icon should change to reflect their current status as SL, and the previous SL's tacmap icon should revert to indicate either that they are no longer SL or that they are their original job icon.

1. Have an SL for one of the squads planetside
2. Go to an overwatch console for that squad and open the tactical map
3. Observe that the SL appears on the tactical map
4. Set a second marine as the acting squad leader
5. Observe that the SL icon did not swap to the now acting SL and they retain their original job icon still

### Issue Bingo

### LINKED ISSUE #3440: Tacmap: Chestburst marines appear with revive icons instead of flatline icons
### Testmerges

#3088, #3185, #3298, #3424, #3430, #3271, #3392

### Description of the bug

If a marine chestbursts while nested, the tacmap icon shows they are revivable instead of permadead

If a marine chestbursts while nested, their tacmap icon should show that they are permanently deceased and not that they are revivable

1. Infect a marine with a chestburster and place him in a wall nest
2. Go to the overwatch console and open the tactical map
3. Wait for the marine to chestburst
4. After the marine has chestburst his tacmap icon will indicate he is revivable with a green heart, instead of deceased with a red flatline

### Issue Bingo

harryob:
closes #3439
closes #3440

:cl:
fix: the tacmap doesn't think burst people are still kicking
fix: the tacmap respects asls
/:cl:

### DIFF
--- a/code/game/objects/items/devices/radio/headset.dm
+++ b/code/game/objects/items/devices/radio/headset.dm
@@ -354,7 +354,7 @@
 		set_undefibbable_on_minimap(z_level, marker_flags)
 		return
 
-	if(wearer.stat == DEAD)
+	if(wearer.stat == DEAD || wearer.chestburst == 2)
 		set_dead_on_minimap(z_level, marker_flags)
 		return
 
--- a/code/modules/gear_presets/_select_equipment.dm
+++ b/code/modules/gear_presets/_select_equipment.dm
@@ -266,18 +266,22 @@
 	else
 		background.color = MINIMAP_ICON_BACKGROUND_CIVILIAN
 
-	if(islist(minimap_icon))
-		for(var/icons in minimap_icon)
+	var/minimap_icon_to_use = minimap_icon ? minimap_icon : "unknown"
+
+	if(user?.assigned_squad?.squad_leader == user)
+		minimap_icon_to_use == "leader"
+
+	if(islist(minimap_icon_to_use))
+		for(var/icons in minimap_icon_to_use)
 			var/iconstate = icons ? icons : "unknown"
 			var/mutable_appearance/icon = image('icons/ui_icons/map_blips.dmi', icon_state = iconstate)
 			icon.appearance_flags = RESET_COLOR
 
-			if(minimap_icon[icons])
-				icon.color = minimap_icon[icons]
+			if(minimap_icon_to_use[icons])
+				icon.color = minimap_icon_to_use[icons]
 			background.overlays += icon
 	else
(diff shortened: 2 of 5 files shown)


code/game/objects/items/devices/radio/headset.dm
```diff
@@ -354,7 +354,7 @@
 		set_undefibbable_on_minimap(z_level, marker_flags)
 		return
 
-	if(wearer.stat == DEAD)
+	if(wearer.stat == DEAD || wearer.chestburst == 2)
```
BeagleGaming1, line 357: Why not have bursting show them as perma dead, instead of the revivable one?

BeagleGaming1 (changes requested):

code/game/objects/items/devices/radio/headset.dm
Watermelon914: also, wearer.chestburst == 2 isn't really clear, can you make the `2` into a define?

code/modules/gear_presets/_select_equipment.dm
```diff
@@ -266,18 +266,22 @@
...
 
-			if(minimap_icon[icons])
-				icon.color = minimap_icon[icons]
+			if(minimap_icon_to_use[icons])
+				icon.color = minimap_icon_to_use[icons]
 			background.overlays += icon
 	else
-		var/iconstate = minimap_icon ? minimap_icon : "unknown"
-		var/mutable_appearance/icon = image('icons/ui_icons/map_blips.dmi', icon_state = iconstate)
+		var/mutable_appearance/icon = image('icons/ui_icons/map_blips.dmi', icon_state = minimap_icon_to_use)
```
Watermelon914, line 284: make a mutable appearance here, not an image.
Images are more so meant for adding to a client's `images` variable. If your goal is to only use it as an overlay, it's better practice to use a mutable appearance.

PR #3118, opened 2023-04-21
Marine Major announcement delay

morrowwolf:
# About the pull request

This PR makes it so when a marine major occurs (no xenos left) it will hold for X minutes before it does the "credits" and song and dance.

The way I did this is probably cursed.

Time open for discussion. Currently at three minutes.

# Explain why it's good for the game

This should allow for a bit more RP time at the end of the round and potentially some time for medals, congratulations, etc.

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl: Morrow
add: Added delay to end of round announcement when it is a Marine Major.
/:cl:

PR #3448, opened 2023-05-28
Marine Major, Ares Announcement

ghostsheet:
# About the pull request

Adds an Ares Announcement when a marine major happens.

Adding onto Morrow's Cursed PR, blame Morrow for how he did it in #3118, not me

Currently marines don't know when they've won, therefore adding the 3 extra minutes is pointless as they spend that time hunting for a supposed last xeno that doesn't even exist.

With this Announcement, Marines will know they don't need to hunt the last xeno down anymore and can RP a conclusion and
Even award medals before the final Marine Major declaration happens.

ARES conclusion 2

# Changelog

:cl: ghostsheet
add: Added ARES conclusion announcement for marine major.
/:cl:

### DIFF
--- a/code/game/gamemodes/colonialmarines/colonialmarines.dm
+++ b/code/game/gamemodes/colonialmarines/colonialmarines.dm
@@ -153,6 +153,11 @@
 		var/rendered_announce_text = replacetext(SSmapping.configs[GROUND_MAP].announce_text, "###SHIPNAME###", MAIN_SHIP_NAME)
 		marine_announcement(rendered_announce_text, "[MAIN_SHIP_NAME]")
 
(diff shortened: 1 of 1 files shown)


Drulikar (changes requested):
The way it was done was intentional to not give more time to EORG, but still an opportunity for the commander to issue awards if they choose to. If this announcement should occur at all, it should be only on the command channel.

morrowwolf (commented):
What Drulikar said.

People were using the extra time to just start EORG early.

If we make it command channel only it may relieve some of that.

If that doesn't do it I'm likely just going to revert the delay which is a bummer

ghostsheet:
ARES conclusion 2
I guess this works

Drulikar (approved):

PR #3449, opened 2023-05-28
Canceling for the hive now resets its slowdown

### LINKED ISSUE #3446: Acid Runner "For The Hive" slowdown stays, even after cancelling the ability
### Testmerges

N/A

### Description of the bug

If you activate "For The Hive" (the bomb ability) you will get a slowdown. The problem arises, because if you use the ability again to cancel it, the slowdown is not removed.

Acid Runner "for the hive" slowdown should not be permanent

1. Be Acid Rouny // Also, have 250 acid for ability
2. Activate for the hive
3. Use ability again to cancel
4. Permanent slowdown!
...
//
Currently, a bandaid fix is to deevo and reevo to runner, then choose acider again. You lose all your stored acid by doing this though, and it costs some of your stored evo.

### Issue Bingo

Drulikar:
# About the pull request

This PR clears the slowdown effect applied when for the hive is canceled.

Though we probably want it to be punishing canceling a for the hive, it still costs acid so it can feel bugged if you remain slowed for 26-40 seconds depending on when you cancel the ability. Closes #3446 

# Changelog
:cl: Drathek
balance: Acid Runner's For the Hive ability now clears its slowdown effect when canceled
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/runner/runner_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/runner/runner_powers.dm
@@ -194,16 +194,21 @@
 	X.say(";FOR THE HIVE!!!")
 
 /datum/action/xeno_action/activable/acider_for_the_hive/proc/cancel_ability()
-	var/mob/living/carbon/xenomorph/X = owner
+	var/mob/living/carbon/xenomorph/xeno = owner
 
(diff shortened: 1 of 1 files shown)


morrowwolf (approved):
Balance approved :D

harryob (approved):

PR #3451, opened 2023-05-28
Fix xenos not ending overwatch when target dies

### LINKED ISSUE #3445: Xeno death + Xeno eye bug // Camera stays on corpse after death of xeno
### Testmerges

N/A

### Description of the bug

If you xeno eye a xeno before they die, your camera position stays on their corpse, even after they die.

Xeno eye should stop automatically once the xeno you are watching dies.

1. Xeno eye a xeno before they die
2. Xeno dies
3. You stay in xeno eye, despite them being dead // Your xeno eye (camera) is on their corpse.
...
//
While this is rarely abused, it should probably be fixed. Probably a main factor of #3444 because upon gibbing there is no corpse left for your camera to view.

### Issue Bingo

Drulikar:
# About the pull request

This PR fixes a logic error in code where non-queen xenos didn't properly test if their overwatched xeno was dead preventing the camera from resetting when they die.

Fixes #3445 

# Changelog
:cl: Drathek
fix: Fixed non-queen xeno overwatch persisting on a dead xeno
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/XenoOverwatch.dm
+++ b/code/modules/mob/living/carbon/xenomorph/XenoOverwatch.dm
@@ -121,7 +121,7 @@
 // Called from xeno Life()
 // Makes sure that Xeno overwatch is reset when the overwatched Xeno dies.
 /mob/living/carbon/xenomorph/proc/handle_overwatch()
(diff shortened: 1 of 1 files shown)


fira (approved):
hahaha

Issue #3444, opened 2023-05-28
Disconnections from server due to xeno eyeing a xeno before they gib


gitbirb:
### Testmerges

N/A

### Description of the bug

If you xeno eye a xeno, and they gib (like in an acid runner explosion), you will disconnect from the server for a bit.

### What's the difference with what should have happened?

Getting disconnected can mean life or death. Ideally, there would be no disconnect here.

### How do we reproduce this bug?

1. Be xeno (not queen)
2. Eye (watch xenomorph) an acid runner while they are exploding 
3. When the acid runner explodes, your client will disconnect from server
...
//
This is pretty consistently reported from what i know, for a lot of people it WILL disconnect you if you eye an acid runner while they are gibbing.
Not sure if it's exclusively acid rouny gibbing though. 

### Issue Bingo

- [ ] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Drulikar:
I am not able to reproduce this locally on a server using two clients on the same computer using the current latest commit on master [c697607]. https://youtu.be/g1jfwD53wI8

There is either a test merge that is relevant, or something else required in your reproduction steps.

gitbirb:
whoops i didn't mean to write "client will crash" its client will disconnect from server. My mistake

gitbirb:
> I am not able to reproduce this locally on a server using two clients on the same computer using the current latest commit on master [[c697607](https://github.com/cmss13-devs/cmss13/commit/c697607c00e5ef1e6414723cf640ad8c82ee164b)]. https://youtu.be/g1jfwD53wI8
> 
> There is either a test merge that is relevant, or something else required in your reproduction steps.

It might be something that only happens to normal xenos, not sure how queen eye works with it or if its coded differently.

Drulikar:
A non-queen xeno watching a xeno get gibbed xeno behaves the same as it does for queen. However, a non-queen xeno watching a xeno get killed will not cancel the overwatch. I can fix that (it is doing `observed_xeno == DEAD` instead of `observed_xeno.stat == DEAD`; queen's override of this proc uses .stat), but the former issue is more concerning.

If you think just the black screen is a disconnection, that's just you overwatching the xeno teleported to nullspace - and in my opinion it side the side effect of the connection being severed briefly

Segrain:
[...]

PR #3453, opened 2023-05-28
Fix Signal Flares Admin Log/JMP

fira:
# About the pull request

Fix issues when jumping to signal flares from the admin niche log. It'd still be 0,0,0 if the user still has it in hand. Worse, this appears to crash client when jumping to it -- but that's going to be a different fix because it probably affects more..

# Changelog
:cl:
fix: Fixed incorrect coordinates in Admin Niche log for signal flares.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/flashlight.dm
+++ b/code/game/objects/items/devices/flashlight.dm
@@ -524,8 +524,9 @@
 		anchored = TRUE
 		if(activate_message)
 			visible_message(SPAN_DANGER("[src]'s flame reaches full strength. It's fully active now."), null, 5)
-		msg_admin_niche("Flare target [src] has been activated by [key_name(user, 1)] at ([x], [y], [z]). (<A HREF='?_src_=admin_holder;[HrefToken(forceGlobal = TRUE)];adminplayerobservecoodjump=1;X=[x];Y=[y];Z=[z]'>JMP LOC</a>)")
-		log_game("Flare target [src] has been activated by [key_name(user, 1)] at ([x], [y], [z]).")
+		var/turf/target_turf = get_turf(src)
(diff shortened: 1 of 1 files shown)


BroseidonMOO (changes requested):
Failed test merge during round 17046, still causes client crash.

fira:
Well this fixes the JMP first and foremost. I think the client crash is a wider thing as evidenced by #3444. I've been trying to reproduce said crash but no dice...

code/game/objects/items/devices/flashlight.dm
```diff
@@ -524,8 +524,9 @@
 		anchored = TRUE
 		if(activate_message)
 			visible_message(SPAN_DANGER("[src]'s flame reaches full strength. It's fully active now."), null, 5)
-		msg_admin_niche("Flare target [src] has been activated by [key_name(user, 1)] at ([x], [y], [z]). (<A HREF='?_src_=admin_holder;[HrefToken(forceGlobal = TRUE)];adminplayerobservecoodjump=1;X=[x];Y=[y];Z=[z]'>JMP LOC</a>)")
-		log_game("Flare target [src] has been activated by [key_name(user, 1)] at ([x], [y], [z]).")
+		var/turf/target_turf = get_turf(src)
+		msg_admin_niche("Flare target [src] has been activated by [key_name(user, 1)] at ([target_turf.x], [target_turf.y], [target_turf.z]). (<A HREF='?_src_=admin_holder;[HrefToken(forceGlobal = TRUE)];adminplayerobservecoodjump=1;X=[target_turf.x];Y=[target_turf.y];Z=[target_turf.z]'>JMP LOC</a>)")
```
Watermelon914, line 528: we really need to change these into a define, but oh well

PR #3468, opened 2023-05-29
Fix for crate exploit.

Segrain:
# About the pull request

Other half of the issue might need some discussion first, but this will do for a start.

Exploits bad.
You know who you are.

# Changelog
:cl:
fix: Removed an exploit involving closets/crates.
/:cl:


fira (approved):
Aww it was an interesting quirk

PR #3471, opened 2023-05-29
Info tags for non-dogtags and info tags dropped on gib

BeagleGaming1:
# About the pull request

Adds information tags to all IDs, and has them drop when the mob is gibbed.
Also makes information tags indestructible.

Allows you to gather more information tags from fallen marines to return to the memorial

# Changelog

:cl:
add: IDs other than dog tags now have information tags
add: Humans drop information tags when gibbed
/:cl:


code/game/objects/structures/props.dm
```diff
@@ -1042,26 +1042,26 @@
...
 	return ..()
 
 /obj/structure/prop/wooden_cross/attackby(obj/item/W, mob/living/user)
-	if(istype(W, /obj/item/dogtag))
-		var/obj/item/dogtag/dog = W
+	if(istype(W, /obj/item/information_tag))
+		var/obj/item/information_tag/tag = W
 		if(!tagged)
 			tagged = TRUE
 			user.visible_message(SPAN_NOTICE("[user] drapes the [W] around the [src]."))
```
Drulikar, line 1057: This doesn't use \the when it should, but regardless it should just be draping a single information tag. Currently it is this:
image

code/game/objects/structures/props.dm
```diff
@@ -1042,26 +1042,26 @@
...
-			dogtag_name = popleft(dog.fallen_names)
-			dogtag_assign = popleft(dog.fallen_assgns)
-			dogtag_blood = popleft(dog.fallen_blood_types)
+			dogtag_name = popleft(tag.fallen_names)
+			dogtag_assign = popleft(tag.fallen_assgns)
+			dogtag_blood = popleft(tag.fallen_blood_types)
```
Drulikar, line 1060: When this causes the id tags to return back to one tag, it doesn't revert the pluralization:
image

Drulikar (changes requested):
It would also be nice if you could use a set of dog tags (you'd only normally acquire them if you have the skill or were given them), it checks if its weak reference owner is valid and alive, and if not separate the ID in the same way you do for a gibbed or unrevivable marine. Currently if you acquire a full ID you can't use it the same way as identification tag.
image

Also it would kinda be better if a collection of identification tags just had a single sprite that showed multiple when it was combined rather than inheriting the previous ID style. Its okay as a dog tag, but the new chip variants are kinda odd - especially when multiple form that singular chip.

PR #3477, opened 2023-05-30
Overwatch Signals for QDEL: Less client disconnection?

### LINKED ISSUE #3444: Disconnections from server due to xeno eyeing a xeno before they gib

Drulikar:
# About the pull request

This PR is sort of a stab in the dark to fix #3444 by making marine and xeno overwatching register a signal for qdeletion to end the overwatch faster. The actual source of this issue is apparently that the multithreaded map tick beta has an issue with crashing clients, so the changes made in this PR don't need to be permanent if the effect isn't as desirable when the beta is fixed.

Locally testing has not been able to reproduce the problem, but I can confirm I do not get the momentary blank screen when overwatching something that gets deleted.

Hopefully no more (or less) client disconnections when overwatching something that gets deleted.

<details>

qdel

</details>

# Changelog
:cl: Drathek
fix: Clients should hopefully avoid disconnection when observing someone that is deleted.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/XenoOverwatch.dm
+++ b/code/modules/mob/living/carbon/xenomorph/XenoOverwatch.dm
@@ -123,7 +123,6 @@
 
 /mob/living/carbon/xenomorph/proc/overwatch_handle_deletion(mob/living/carbon/xenomorph/deleted, forced)
 	SIGNAL_HANDLER
-	UnregisterSignal(deleted, COMSIG_PARENT_QDELETING)
 	overwatch(deleted, TRUE)
 
 // Called from xeno Life()
--- a/code/modules/mob/living/carbon/xenomorph/XenoOverwatch.dm
+++ b/code/modules/mob/living/carbon/xenomorph/XenoOverwatch.dm
@@ -71,6 +71,7 @@
 
 		SEND_SIGNAL(src, COMSIG_XENO_STOP_OVERWATCH, oldXeno)
 		UnregisterSignal(src, COMSIG_MOB_MOVE_OR_LOOK)
+		UnregisterSignal(oldXeno, COMSIG_PARENT_QDELETING)
 
 		if(oldXeno)
 			to_chat(src, SPAN_XENOWARNING("You stop watching [oldXeno]."))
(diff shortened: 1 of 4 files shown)


Wintermote:
I haven't noticed disconnection when overwatching using the overwatch console in CIC, and that should be the same process as the ghost overwatch, shouldn't it?

code/modules/mob/living/carbon/xenomorph/XenoOverwatch.dm
```diff
@@ -109,15 +110,22 @@
...
 		observed_xeno.hud_set_queen_overwatch()
 		SEND_SIGNAL(src, COMSIG_XENO_OVERWATCH_XENO, observed_xeno)
 		RegisterSignal(src, COMSIG_MOB_MOVE_OR_LOOK, PROC_REF(overwatch_handle_mob_move_or_look))
+		RegisterSignal(observed_xeno, COMSIG_PARENT_QDELETING, PROC_REF(overwatch_handle_deletion))
 
 	src.reset_view()
 
+/mob/living/carbon/xenomorph/proc/overwatch_handle_deletion(mob/living/carbon/xenomorph/deleted, forced)
+	SIGNAL_HANDLER
+	UnregisterSignal(deleted, COMSIG_PARENT_QDELETING)
```
Watermelon914, line 126: this isn't necessary

code/modules/mob/mob.dm
```diff
@@ -356,6 +356,10 @@
 		SEND_SIGNAL(client, COMSIG_CLIENT_RESET_VIEW, A)
 	return
 
+/mob/proc/reset_observer_view_on_deletion(atom/deleted, force)
+	SIGNAL_HANDLER
+	UnregisterSignal(deleted, COMSIG_PARENT_QDELETING)
```
Watermelon914, line 361: this isn't necessary, all signals get unregistered on deletion anyways

Watermelon914 (changes requested):

harryob (approved):
lets see

Drulikar:
#3525 might be a more comprehensive fix unless that's yet another different 515 flaw.

PR #3480, opened 2023-05-30
InfluxDB Job Occupation Graphing

fira:
# About the pull request

Adds InfluxDB support to job statistics, as an alternative to Discord Webhooks.

This doesn't sent event markers such as Round Start, First Drop, etc, because they're not metrics. Logging / ElasitcSearch would be better suited to that.

No code reuse here, the Subsystem does its own count due to much higher frequency of measurements (10-60 seconds VS 10 minutes interval)

Allows graphing of job occupations per round, including breakdown by team (humans vs xenos), and indexed by round_id. This allows for aggregations that will let us easily see which jobs are occupied over longer periods of time, contrary to the discord webhooks which are hard to reuse.

image

# Changelog
:cl:
add: Added support of logging of job occupations via InfluxDB.
/:cl:

### DIFF
--- a/code/__DEFINES/subsystems.dm
+++ b/code/__DEFINES/subsystems.dm
@@ -161,6 +161,7 @@
 #define SS_INIT_OBJECTIVES -31
 #define SS_INIT_LOBBYART   -33
 #define SS_INIT_MINIMAP    -34
+#define SS_INIT_INFLUXSTATS -50
 #define SS_INIT_STATPANELS -98
 #define SS_INIT_CHAT    -100 //Should be last to ensure chat remains smooth during init.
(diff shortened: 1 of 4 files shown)


fira:
This appears to have stopped working after 3 rounds during testing for an unknown reason...

fira:
Additionally the data format should probably be changed so jobs are individual tag of measurements rather than field, and the round_time removed

fira:
Works as intended now, though could use extra tags eg. the hive for xenos, and more metrics such as total involved players, tech points, and so on

fira:
image

fira:
Reworking this into a more comprehensive solution

PR #3482, opened 2023-05-30
When you ghost as a zombie your body will become a free mob.

Huffie56:
# About the pull request
When you ghost as a zombie your body will become a free mob.
i personalize part of the code for zombie warn you that you won't be able to use it if you ghost.

No more zombie wasted we recycle now.
<details>

</details>

# Changelog
:cl:
fix: When you ghost as an alive zombie you body will become a free mob and you won't be able to reenter your body.
/:cl:

### DIFF
--- a/code/datums/diseases/black_goo.dm
+++ b/code/datums/diseases/black_goo.dm
@@ -116,9 +116,6 @@
 		stage = 5
 		human.faction = FACTION_ZOMBIE
 		zombie_transforming = FALSE
-	//if zombie as no client when turning get him into the free mob list.in the work
-	//if(!human.client)
-		//human.free_for_ghosts()
 
 /obj/item/weapon/zombie_claws
 	gender = PLURAL
--- a/code/datums/diseases/black_goo.dm
+++ b/code/datums/diseases/black_goo.dm
@@ -116,10 +116,9 @@
 		stage = 5
 		human.faction = FACTION_ZOMBIE
 		zombie_transforming = FALSE
-	//if zombie is SSD or souless
+	//if zombie as no client when turning get him into the free mob list.
 	if(!human.client)
 		human.free_for_ghosts()
-
 
 /obj/item/weapon/zombie_claws
 	gender = PLURAL
--- a/code/datums/diseases/black_goo.dm
+++ b/code/datums/diseases/black_goo.dm
@@ -116,9 +116,9 @@
 		stage = 5
 		human.faction = FACTION_ZOMBIE
 		zombie_transforming = FALSE
-	//if zombie as no client when turning get him into the free mob list.
-	if(!human.client)
-		human.free_for_ghosts()
+	//if zombie as no client when turning get him into the free mob list.in the work
+	//if(!human.client)
+		//human.free_for_ghosts()
 
 /obj/item/weapon/zombie_claws
 	gender = PLURAL
--- a/code/datums/diseases/black_goo.dm
+++ b/code/datums/diseases/black_goo.dm
@@ -116,6 +116,9 @@
 		stage = 5
 		human.faction = FACTION_ZOMBIE
 		zombie_transforming = FALSE
+	//if zombie is SSD or souless
+	if(!human.client)
+		human.free_for_ghosts()
 
 
 /obj/item/weapon/zombie_claws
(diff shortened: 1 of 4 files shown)


code/modules/mob/dead/observer/observer.dm
```diff
@@ -365,6 +366,10 @@ Works together with spawning an observer, noted above.
 
 	ghost.set_huds_from_prefs()
 
+	//if zombie turn your body back to the free mob pool
+	if(iszombie(src))
+		src.free_for_ghosts()
```
harryob, line 371:
```suggestion
		free_for_ghosts()
```

harryob (changes requested):

Huffie56:
i need to make a test to check if someone is decap or as is head destroyed but i didn't found anything that match this so if you can help tag me. thanks.

code/modules/mob/dead/observer/observer.dm
```diff
@@ -365,6 +365,10 @@ Works together with spawning an observer, noted above.
 
 	ghost.set_huds_from_prefs()
 
+	//if zombie turn your body back to the free mob pool
+	if(iszombie(src))
+		free_for_ghosts()
```
Huffie56, line 370: i need to add a check so that zombie without their head don't become free mob...
i tried multiple stuff nothing work....

code/datums/diseases/black_goo.dm
```diff
@@ -116,7 +116,9 @@
 		stage = 5
 		human.faction = FACTION_ZOMBIE
 		zombie_transforming = FALSE
-
+	//if zombie as no client when turning get him into the free mob list.
+	if(!human.client)
+		human.free_for_ghosts()
```
Huffie56, line 122: need to make sure this doesn't create a conflict and between old owner of the corpse being able to return back in while the new zombie is on join free mob list...

Huffie56:
todo list : 
1-add a TGUI pop up for when a corpse is being turn into a zombie....
2-add a TGUI pop when a dead zombie get revive...

Huffie56:
i tried to add other checks for revive or when you turn into zombie to get more free mob zombie but those haven't been reliable...

Huffie56:
image
it only work for living zombie right now as i didn't find the check ... oh i have an idea...

BeagleGaming1:
How's this a "Code Improvement"?

Huffie56:
> How's this a "Code Improvement"?

code is my way to go when i don't know what to pick.

code/modules/mob/dead/observer/observer.dm
```diff
@@ -365,6 +365,9 @@ Works together with spawning an observer, noted above.
 
 	ghost.set_huds_from_prefs()
 
+	if(iszombie(src)&&stat == !DEAD)
```
Drulikar, line 368: Even when not inhabited by a client, a zombie can still be reviving (generally this should only not be occurring if we're decapitated); so you might want to either also check if we're reviving, or have the revival perform a check if we have a client and then also free_for_ghosts there if there is no client.

Drulikar (changes requested):

code/modules/mob/dead/observer/observer.dm
```diff
@@ -378,15 +381,12 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 	do_ghost()
 
 /mob/living/proc/do_ghost()
-	if(stat == DEAD)
-		if(mind && mind.player_entity)
-			mind.player_entity.update_panel_data(round_statistics)
-		ghostize(TRUE)
-	else
+	//if your a zombie and alive you will give up your right to your body.
+	if(iszombie(src)&&stat == !DEAD)
```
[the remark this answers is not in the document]
Drulikar: See `/datum/species/zombie/handle_death(mob/living/carbon/human/zombie, gibbed)` in zombie.dm

The conditioons for whether they get a revive timer is: they must be a zombie, they must not be gibbed, they must have a head that isn't qdeleted, and they must have a head that is not LIMB_DESTROYED.

Alternatively, you can check if the zombie species has a weakref to this zombie in its to_revive; but the former checks would likely be preferable to handle situations where the client is disconnecting as their mob is undoing the revival process.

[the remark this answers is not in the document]
Huffie56: i will try.

PR #3497, opened 2023-06-01
Bloodpack Self Application Allowance and Balance for Medically Trained

SpypigDev:
# About the pull request

Allows medically trained players to self-apply blood bags with some notable downsides for balance. Medics will now be able to attach blood packs to themselves with a 2x multiplier on the application time, as well as applying a slowness modifier while the bag is active. Attaching blood bags to yourself now applies 1 brute damage to add to the realism of awkwardly sticking a blood pack in your own arm, and you likely messing it up and it stinging.

It doesnt exactly make sense that medics would outright be unable to self-apply blood packs while in the field, and this change now allows them to do a shoddy job of it, but not outright denying them. Blood pack self-application will slow medics in the field (not just for the application time, but for the entire time it is applied) by a significant degree compared to their marine comrades. Not only does applying a blood pack to yourself take roughly 6 seconds (while can easily be interrupted), your movement speed debuff takes away the nimble and speedy aspect of gameplay that medics need to prioritise to survive, making you an easy picking should the tides of battle suddenly change. 

If in such a pinch, medics will be forced to either spend time disconnecting their blood bag (wasting critical seconds), or drop the bag and run (which would make them scream and harm them). I have frequently come across the occasion where I am low on blood, but particular factors do not allow me to find another medic or reach the FOB in my state (for example, low-pop HM), and I believe this balance would be a fair fix for this.

This PR is mainly an attempt to balance field blood packs for medics but also introduces unique downsides and situations to keep the new feature fair. 
# Changelog
:cl:
add: Flavour text for self-applying blood packs, similar to self-surgery
add: Variable 'self_attachment_modifier' in blood pack code
balance: Adds a small amount of damage and a considerable slowness to the player for self-applying blood pack
balance: Double the time it takes to connect the blood pack to yourself
balance: Increases the blood pack output modifier from 30 to 50
/:cl:


morrowwolf (commented):
I appreciate the work you've put into this and hope to see more of what you make.

At the same time, bypassing teamwork/interaction with others is not something I'm going to approve.

PR #3510, opened 2023-06-02
fixes largecrates being invulnerable to projectiles

### LINKED ISSUE #3345: Barrels and Crates can be used as barricades by xenos.
### Testmerges

#3088: 'Fixes offset issue for multithreading test' by LynxSolstice at commit 95d6c01a9c #3303: 'Yautja armor rebalance' by morrowwolf at commit 9a9576ff58 #3298: 'Project ARES - Resurrection' by realforest2001 at commit 1c7c6da316 #3271: 'Pred QOL updates' by realforest2001 at commit d8c2999636 #3333: 'ensures people always have at least a set of keybinds' by harryob at commit 3abe4f8f62 #3334: 'Maps for squaddies' by morrowwolf at commit 46528f7ce4 #3337: 'prevents people spacewalking' by harryob at commit 937cecf2de

### Description of the bug

(this is an exploit)
Barrels, Wooden Crates, and Supply Crates (the kind you crowbar open) can all be used as barricades by xenomorphs. 

The problem arises from how these items cannot be broken by guns. This allows a xenomorph to hide behind the object in order to avoid damage from gunfire. It can be even further exploited, since acid spray is not blocked by these objects, meaning you can fire acid spray at a target (or cadeline) without any return fire possible.

While this is fun and can create a lot of emergent gameplay, it is (probably) an exploit. Barrels and Crates should be breakable using guns. 

1. As xenomorph, hide behind a barrel or crate
2. gun shoots towards you
3. Bullet blocked by crate
4. magdump the crate to try to destroy it (the crate eats ALL those bullets)
5. Crate still standing. Xenomorph has an advantage.
...

### Issue Bingo

realkhad:
# About the pull request

makes it so largecrates take damage and break, no more soaking up projectiles
fixes #3345 

fix good

<details>

</details>

# Changelog
:cl:Khadd
fix: largecrates now take damage and break if hit enough times by a projectile
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/largecrate.dm
+++ b/code/game/objects/structures/crates_lockers/largecrate.dm
@@ -3,7 +3,6 @@
 	desc = "A hefty wooden crate."
 	icon = 'icons/obj/structures/crates.dmi'
 	icon_state = "densecrate"
-	health = 100
 	density = TRUE
 	anchored = FALSE
 	var/parts_type = /obj/item/stack/sheet/wood
@@ -78,7 +77,7 @@
 		unpack()
 
 /obj/structure/largecrate/bullet_act(obj/item/projectile/P)
-	take_damage(P.damage*0.5)
+	take_damage(P.calculate_damage(P.damage))
 	return TRUE
 
 /obj/structure/largecrate/mule
--- a/code/game/objects/structures/crates_lockers/largecrate.dm
+++ b/code/game/objects/structures/crates_lockers/largecrate.dm
@@ -3,6 +3,7 @@
 	desc = "A hefty wooden crate."
 	icon = 'icons/obj/structures/crates.dmi'
 	icon_state = "densecrate"
+	health = 100
 	density = TRUE
 	anchored = FALSE
 	var/parts_type = /obj/item/stack/sheet/wood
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/crates_lockers/largecrate.dm
```diff
@@ -3,6 +3,7 @@
 	desc = "A hefty wooden crate."
 	icon = 'icons/obj/structures/crates.dmi'
 	icon_state = "densecrate"
+	health = 100
```
fira, line 6: redundant

code/game/objects/structures/crates_lockers/largecrate.dm
```diff
@@ -71,6 +72,15 @@
 	if(power >= EXPLOSION_THRESHOLD_VLOW)
 		unpack()
 
+/obj/structure/largecrate/proc/take_damage(damage)
+	health -= damage
+	if(health <= 0)
+		unpack()
+
+/obj/structure/largecrate/bullet_act(obj/item/projectile/P)
+	take_damage(P.damage*0.5)
```
fira, line 81: You should use `/obj/item/projectile/calculate_damage()` rather than `P.damage` directly so it accounts for falloff/buildup

fira (changes requested):

code/game/objects/structures/crates_lockers/largecrate.dm
realkhad: changed in 23f031622b398a028d7c647be2216e30a361b959

fira (approved):

PR #3514, opened 2023-06-03
Fix Force Dropship Admin Verb

Drulikar:
# About the pull request

This PR fixes the force dropship admin verb that could be used to hijack a dropship. Now it will not runtime, nor require the admin to be adjacent to the console.

~For a non-crash, I would have it just open the shuttle manipulator, but that is a verb on the admin datum instead of the client~ - this verb also doesn't perform any admin permissions checking currently... (it probably should).

Admin verbs should work, otherwise they should be eliminated.

<details>

hijack

</details>

# Changelog
:cl: Drathek
admin: Fixed the force dropship verb (now known as Force Hijack) requiring the admin to be adjacent to the console and runtiming.
/:cl:

### DIFF
--- a/code/modules/admin/tabs/event_tab.dm
+++ b/code/modules/admin/tabs/event_tab.dm
@@ -444,7 +444,7 @@
 		return
 
 	var/obj/structure/machinery/computer/shuttle/dropship/flight/computer = dropship.getControlConsole()
-	computer.hijack(usr, force=TRUE)
+	computer.hijack(usr, force = TRUE)
 
 /client/proc/cmd_admin_create_centcom_report()
 	set name = "Report: Faction"
--- a/code/modules/admin/tabs/event_tab.dm
+++ b/code/modules/admin/tabs/event_tab.dm
@@ -431,11 +431,13 @@
 	var/list/shuttles = list(DROPSHIP_ALAMO, DROPSHIP_NORMANDY)
 	var/tag = tgui_input_list(usr, "Which dropship should be force launched?", "Select a dropship:", shuttles)
 	if(!tag) return
-	var/crash = 0
-	switch(tgui_input_list(usr, "Would you like to force a crash?", "Force crash", list("Yes", "No")))
-		if("Yes") crash = 1
-		if("No") crash = 0
-		else return
+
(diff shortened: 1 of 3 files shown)


code/modules/admin/tabs/event_tab.dm
```diff
@@ -431,11 +431,13 @@
...
-		if("Yes") crash = 1
-		if("No") crash = 0
-		else return
+
+	var/crash = tgui_input_list(usr, "Would you like to force a crash?", "Force crash", list("Yes", "No"))
+	if(!crash) return
+
+	if(crash == "No")
+		to_chat(usr, SPAN_WARNING("Use the shuttle manipulator to move a shuttle normally"))
+		return
```
realforest2001, line 440: Can you rename the proc to Force Hijack to fit with this please

realforest2001 (changes requested):

code/modules/admin/tabs/event_tab.dm
```diff
@@ -423,32 +423,27 @@
...
+		log_admin("Error: Attempted to force a dropship hijack but the shuttle datum was null. Code: MSD_FSV_DIN")
 		return
 
-	if(crash)
-		var/obj/structure/machinery/computer/shuttle/dropship/flight/computer = dropship.getControlConsole()
-		computer.hijack(usr)
-	else
-		to_chat(usr, SPAN_WARNING("Use the shuttle manipulator to normally move a shuttle"))
+	var/confirm = tgui_alert(usr, "Are you sure you want to hijack [dropship]?", "Force hijack", list("Yes", "No")) == "Yes"
+	if(!confirm) return
```
harryob, line 443:
```suggestion
	if(!confirm) 
		return
```

code/modules/admin/tabs/event_tab.dm
```diff
@@ -423,32 +423,27 @@
...
+
+	var/obj/structure/machinery/computer/shuttle/dropship/flight/computer = dropship.getControlConsole()
+	computer.hijack(usr, force=TRUE)
```
harryob, line 446:
```suggestion
	computer.hijack(usr, force = TRUE)
```

harryob (changes requested):

harryob (approved):

PR #3524, opened 2023-06-03
Changes rifleman sandbags from 20 points to 15

Hopekz:
In all my rifleman loadouts I seem to always have 15 points left over vs 20 so this change would allow people like me to actually take some sandbags with them if they buy other things. The price just seems too high at the moment so I'd like to see this value lowered.

For example If you just buy the pulse rifle kit and nothing else you have exactly 15 points left over.

image
image

:cl: Hopek
add: Rifleman sandbags now cost 15 points instead of 20 so that more riflemen can fit them in their loadout
/:cl:


ihatethisengine:
For the record I am a mere contributor and that was just an advice

morrowwolf:
Play comtech if you want more building supplies.

PR #3525, opened 2023-06-03
Port /tg/ PR #75835 - Linux 515 fcopy() crash fix

fira:
Details here: https://github.com/tgstation/tgstation/pull/75835

### DIFF
--- a/code/_byond_version_compat.dm
+++ b/code/_byond_version_compat.dm
@@ -47,3 +47,15 @@
 /// Call by name proc reference, checks if the proc is existing global proc
 #define GLOBAL_PROC_REF(X) (/proc/##X)
 #endif
+
+#if (DM_VERSION == 515)
+/// fcopy will crash on 515 linux if given a non-existant file, instead of returning 0 like on 514 linux or 515 windows
+/// var case matches documentation for fcopy.
+/world/proc/__fcopy(Src, Dst)
+	if (!fexists(Src))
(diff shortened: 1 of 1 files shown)


Drulikar:
Glad I added that 515 unit test; though also failing on TG too. Likely not worth merging until they sort out why or can confirm for sure that all those errors somehow are false positives.

Drulikar:
I can confirm they are not false positives: 
image
You just don't get icons.

fira:
Whathefuk...

Drulikar:
Looks like Walter's edit should be viable.

fira:
Indeed. I don't know why I thought fexists() would work with RSC bundled resources...

Drulikar:
This still seems kinda unstable. So character panel never ended up loading in the human model, and each time it was opened it would incrementally add in more clothes sprites. Everything else seemed fine, expect that it then proceeded to crash or disconnect (ending the debug run). These were the errors I experienced related to the character panel: 
[dd.log](https://github.com/cmss13-devs/cmss13/files/11645893/dd.log)

Its not obvious yet what caused a crash/disconnection though. I was using 515.1606. Will try again using 515.1608.

fira:
I'm not sure these are related, it all reads out as a null mob. Intriguing though...
In any case this is meant primarily for TM to the Linux server so we can find out what happens there

PR #3526, opened 2023-06-03
Remove old armor crit tip

fira:
Removes an outdated and misleading tip. Armor Crits were removed like over 6 months ago.


harryob (approved):
6 months ago? 🧓

PR #3527, opened 2023-06-03
Burrower burrow fixes

JackieEstegado:
# About the pull request
1. Burrowers couldn't cancel their burrowing if they were clicking a tile that was non-valid to burrow to.
-Fixed by putting the 'if' for catching the request to cancel the burrow, ahead of everything else, so it wouldn't stopped by a validity check, since that would be irrelevant.

2. If the burrow was cancelled in the _literal_ last second the burrow would complete regardless, due the 1 second timer to teleport to the tile already having been started, with nothing to stop it. This can cause hillarities like you see in the video where the burrower first unburrows on the original tile, then proceeds to teleport to the target tile.
-Fixed by quitting the method that gets called by the 1 sec timer if we already cancelled.

Because stuff like this makes burrower players pull their hairs out

https://github.com/cmss13-devs/cmss13/assets/15560820/69ff6381-35fc-4123-91fa-856c51cec31e

Jawohl.

# Changelog
:cl: Jackie_Estegado
fix: The Burrower's burrow ability will no longer be not cancelled just because you clicked on a tile that you wouldn't have been able to burrow to.
fix: Burrowing will no longer complete even though you cancelled it.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/burrower/burrower_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/burrower/burrower_powers.dm
@@ -88,6 +88,13 @@
 		to_chat(src, SPAN_NOTICE("You must be burrowed to do this."))
 		return
 
+	if(tunnel)
+		tunnel = FALSE
+		to_chat(src, SPAN_NOTICE("You stop tunneling."))
+		used_tunnel = TRUE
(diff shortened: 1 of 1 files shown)


Drulikar (approved):
Changes appear to be working in local testing, and I was able to reproduce the original issue on master.

PR #3534, opened 2023-06-04
HPR is cast into the void

morrowwolf:
# About the pull request

This PR removes the HPR from all vendors other than requisitions.

Ammo economy and scarcity of special weapons is integral to the CM marine experience.

The availability and increasing universality of the HPR goes completely counter to that.

Honest to god, this is a marine buff.

<details>

</details>

# Changelog

:cl: Morrow
balance: HPR is now only found in requisitions.
/:cl:

### DIFF
--- a/code/game/machinery/vending/vendor_types/requisitions.dm
+++ b/code/game/machinery/vending/vendor_types/requisitions.dm
@@ -35,7 +35,7 @@
 		list("SU-6 Smart Pistol", round(scale * 3), /obj/item/storage/box/guncase/smartpistol, VENDOR_ITEM_REGULAR),
 		list("MOU-53 Shotgun", round(scale * 2), /obj/item/storage/box/guncase/mou53, VENDOR_ITEM_REGULAR),
 		list("XM88 Heavy Rifle", round(scale * 3), /obj/item/storage/box/guncase/xm88, VENDOR_ITEM_REGULAR),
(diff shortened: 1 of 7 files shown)


SarahJohnson489:
Players already have to pack multiple trips of extra ammo beforehand if they care about not running out early in the op without having to rely on how req feels that round and how many others that already emptied their bags dogpiling said ammo from req. The corpse looting's gotten out of hand from assmad players trying to be the first to get at a perma's ammo. It was the same when you bought the HPR which used much more ammo. It's too easy to run through hpr mags leaving you dry if you don't pack extra ammo for yourself or do the jerk move and steal more than your share of everyone else's ammo. The HPR's ammo situation is not only in the same 'pack lots of extra' CM marine experience as everyone else but it's worse due to higher ammo consumption and having to empty m41/m4ra mags in the field instead of being able to just hoover up replacement ammo effortlessly from most permas like some other primaries.

Stock count is ridiculously low for striking it from everyone else. Has the same med/lowpop problem as the other restricted weapons of getting bent as a result of said low stock just cause more people felt like running it that round than others. That's pretty high up on the anti-CM marine experience.

It's a nerf. It's clearly a nerf there's no ambiguity whatsoever. Not just in faction effectiveness but needlessly restricting the variety and accessibility of alternate playstyles behind even more anti-CM marine experience be there for roundstart and rush req naked or get bent vendor stock counts. That's an indefensible bad experience for the player in the server playing the game that wants to run a build for their game or two after work only to get shut out for no good reason. Hundreds of living breathing actual people are affected by this stuff.

harryob (approved):

PR #3535, opened 2023-06-04
Remove defunct agents objects tracker

fira:
# About the pull request

The Tracker only tracked 3 types of common objects to begin with that have no purpose other than through objectives. It's also admin spawned right now. The code implementation is also super rigid and that makes it neither fancy nor reusable. This makes more non-reasons to have it than i care to count.

# Changelog
:cl:
del: Removed agents' tracker.
/:cl:


fira:
bah whatever i guess, don't wanna remove them from maps...

PR #3298, opened 2023-05-11
Project ARES - Resurrection

realforest2001:
# About the pull request
Overhauls ARES and makes it actually feel like a part of the ship and not a square box. 

Adds individual processors for different ARES systems that interact with oneanother to produce the same effects as we normally have.
Bioscans require the bioscan processor to be functional, however will not broadcast if the interface processor is not functioning.
The interface processor handles ARES communications (excluding Apollo) and is chiefly responsible for logging things to the ARES Console. This console can be used to review logs of all ARES' messages/announcements, Bioscans, the Apollo Link logs etc.

# Explain why it's good for the game
Adds a little immersion and use for ARES.
# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
add: Added an admin button for an ARES bioscan. Slightly refactored how Bioscans trigger for marines, relying on an ARES processor.
add: Added individual proccessors for ARES systems. These are WIP and will eventually have damage and repair interactions.
add: Added an ARES interaction console in the AI Core room, which holds logs for most ARES functions.
add: Added the ability for ARES console to call ERT or Distress.
add: Added motion triggers in ARES core and shipside comms that send alerts over Apollo.
add: Added a 1to1 conversation feature between ARES and users of the interface console.
add: Added preset open versions of blended poddoors.
add: Added a console for directing Working Joes. This is largely WIP for future PR(s).
add: Added subtypes of air pipes that don't explode on hijack, used these in ARES core.
maptweak: Remodelled ARES Core onto a fake-z, and added the new processors.
fix: door_control buttons now respect being indestructable when processing explosions.
maptweak: Fixed the M39s overflowing in brig armory due to use of landmarks.
/:cl:

PR #3536, opened 2023-06-05
Debug jumptocoord issues

fira:
# About the pull request

Adds debug + one small change to try to catch disconnects on admin JMP. The change should prevent cascading nullspace JMP links, at the very least.


fira:
see #3298

PR #3537, opened 2023-06-05
Xeno ability standardization

Drulikar:
# About the pull request

This PR is a followup to #3516 that prompted #3533. Xenos are able to point again, and marines can open bags again with MMB. Now all xeno abilities should return true when they succeed so that the input can be "consumed" and properly track usage stats.

In the process of testing vanguard, I also added the template_active state to its dash while it is waiting for timeout or second use for its followup attack. Also, it seemed silly that drones and hivelords could waste plasma transferring to themselves.

NOTE: Just like originally intended for the first PR, this is significant change and could have unintended effects. It needs testing to identify if there are any other undesirable effects that I need to address.

Fixes issues with the first PR and standardizes xeno use_ability procs by making sure they call parent so they should now all properly record usage statistics and return true on success.

# Changelog
:cl: Drathek
fix: Fix xenos being unable to point and marines unable to open bags with MMB
fix: Fix ghosts interacting with ghost join abilities and teleporting to UI icons.
fix: Disabled xenos transferring plasma to self (it just wasted plasma)
ui: Vanguard dash now uses a different icon state while waiting for second use (or timeout)
/:cl:


morrowwolf (approved):
Code good

Seems to be working as intended on testmerge.

We'll handle any edge cases that pop up.

PR #3541, opened 2023-06-05
Nerfs Lobby Music

fira:
# About the pull request

Reduces default volume of Lobby Music and makes it stop on joining the game.

Unfortunately have to reduce volume because the volume preferences are not saved.

Shamelessly untested

Less intrusive, and as with my 2 previous PRs on the topic, should discourage new people from just closing the window right away.

# Changelog
:cl:
fix: Lobby Music is a bit quieter and should now stop properly when you spawn in.
/:cl:

### DIFF
--- a/code/modules/client/client_defines.dm
+++ b/code/modules/client/client_defines.dm
@@ -42,7 +42,7 @@
 	var/played = 0
 	var/midi_silenced = 0
 	var/datum/soundOutput/soundOutput
-	var/list/volume_preferences = list(1, 0.5, 1, 0.7)//Game, music, admin midis, lobby music
+	var/list/volume_preferences = list(1, 0.5, 1, 0.6)//Game, music, admin midis, lobby music
 
 		////////////
 		//SECURITY//


code/modules/client/client_defines.dm
```diff
@@ -42,7 +42,7 @@
 	var/played = 0
 	var/midi_silenced = 0
 	var/datum/soundOutput/soundOutput
-	var/list/volume_preferences = list(1, 0.5, 1, 0.85)//Game, music, admin midis, lobby music
+	var/list/volume_preferences = list(1, 0.5, 1, 0.7)//Game, music, admin midis, lobby music
```
morrowwolf, line 45: I am a fellow lobby music enthusiast but uuuh... it's *really* loud. Can we go down to like... half?

morrowwolf (changes requested):

code/modules/client/client_defines.dm
fira: It's not a linear scale AFAIK so half is a lot quieter, but we could further decrease it at 0.6 or 0.5

Keep in mind also not all tracks have the same volume, we could try reducing the audio volume of the outlier files directly

fira:
```suggestion
	var/list/volume_preferences = list(1, 0.5, 1, 0.6)//Game, music, admin midis, lobby music
```

PR #3549, opened 2023-06-06
Fix for queen eye runtime.

Segrain:
# About the pull request

Followup to #3477.

image

# Changelog
:cl:
fix: Queen Eye no longer repeatedly throws runtime errors.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/XenoOverwatch.dm
+++ b/code/modules/mob/living/carbon/xenomorph/XenoOverwatch.dm
@@ -71,10 +71,10 @@
 
 		SEND_SIGNAL(src, COMSIG_XENO_STOP_OVERWATCH, oldXeno)
 		UnregisterSignal(src, COMSIG_MOB_MOVE_OR_LOOK)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/XenoOverwatch.dm
```diff
@@ -71,10 +71,10 @@
 
 		SEND_SIGNAL(src, COMSIG_XENO_STOP_OVERWATCH, oldXeno)
 		UnregisterSignal(src, COMSIG_MOB_MOVE_OR_LOOK)
-		UnregisterSignal(oldXeno, COMSIG_PARENT_QDELETING)
 
 		if(oldXeno)
 			to_chat(src, SPAN_XENOWARNING("You stop watching [oldXeno]."))
+			UnregisterSignal(oldXeno, COMSIG_PARENT_QDELETING)
```
Drulikar, line 77: Honestly should just also stick it under the !QDELETED check too since there already is one being performed. QDEL should already unregister it, so there's the possibility of it double unregistering currently.

PR #3556, opened 2023-06-07
Reduces SG Full Auto Scatter

fira:
# About the pull request

It's been bugging me for a long time, but when you fire for a good dozen seconds with the standard issue smartguns, the bullets start scattering. So, so far you'll say, good Fira, that's soulful! 

However, we have no ACTUAL recoil or similar mechanic. So letting go of the LMB for just even 20 miliseconds is enough to reset scatter to start of firing. **It's just a noobtrap with zero real gameplay elements.**

This reduces the max scatter so that bullets don't just start (after EIGHTY shots!) spraying a (roughly) 48° angle cone, but instead 12° which mostly stays on the same actual turfs. At this value the targeting impact is vastly minimized, but the projectile visuals retain significant scattering. 

I don't think this ACTUALLY qualifies as a "balance" change due to how irrelevant the "mechanic" was, but i'll slap it on.

Less of a noobtrap and pointless purely mechanical micromanagement so people can focus on playing the game.

I'd rather we get a recoil mechanic to make this meaningful but it's bit of a bigger problem...

# Changelog
:cl:
qol: Reduced USCM SG max scattering on Full Auto fire so you don't have to periodically let go of the fire button to keep it from firing way wide. 
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/smartgun.dm
+++ b/code/modules/projectiles/guns/smartgun.dm
@@ -88,7 +88,7 @@
 	burst_delay = FIRE_DELAY_TIER_9
 	fa_delay = FIRE_DELAY_TIER_SG
 	fa_scatter_peak = FULL_AUTO_SCATTER_PEAK_TIER_8
-	fa_max_scatter = SCATTER_AMOUNT_TIER_3
+	fa_max_scatter = SCATTER_AMOUNT_NONE
 	if(accuracy_improvement)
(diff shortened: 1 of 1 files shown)


morrowwolf:
Holy shit Fira doing a balance PR I thought I'd never see the day.

You make a good argument: approved balance-wise.

ihatethisengine:
You are the best! Thank you! Was a major issue for me when I played SG. Hopefully now I won't have to spam click.

PR #3563, opened 2023-06-08
Fix broken APC on 522 blocking movement forever

Steelpoint:
# About the pull request

Previously the broken APC outside of the central processor would permanently block all mob movement into it once the APC is destroyed, creating an odd situation where people can't enter a regular tile.

This PR changes it so the APC itself is blocking movement, and that it being destroyed will allow people to move into the tile. 

Logical consistency, makes people stop asking why you can't move into a open tile. 

# Changelog

:cl:
fix: Broken APC on 522 no longer blocks movement when destroyed
/:cl:


Drulikar:
The changes done in this PR are the same as the changes in the above mentioned PR except that they also modified the turfs underneath the APC. So for now, will close this. Thank you for submitting this though.

PR #3576, opened 2023-06-09
You can no longer use MMB to attack (probably)

ihatethisengine:
# About the pull request

https://github.com/cmss13-devs/cmss13/pull/3516 this PR changed behavior of middle mouse button clicking, which caused xenos to attack adjustment tiles when attempting to use abilities on cooldown. This PR won't let xenos to attack using MMB, I doubt anyone really used MMB for this. I don't know if it will break anything else tho.

It was unexpected behavior. If I am trying to use my ability, I don't want to attack.

<details>
I cannot test it for now.

</details>

# Changelog
:cl: ihatethisengine
fix: MMB cannot be used to attack again
/:cl:


Drulikar (approved):
Forced a merge from master because this PR was too out of date to even contain the PRs its trying to address.

Mentioned on discord, but probably the only possible change that might need to be done is to make this early return exclusive to xenos; but we can determine that with test merges.

However as is, I cannot currently find any issues with local testing for a xeno or marine.

PR #3577, opened 2023-06-09
Revive Fixes for Boiler: IFF Gas and Cause Data

Drulikar:
# About the pull request

This PR revives #3222 since it was basically fully complete except for a few lingering cause data issues and didn't yet address boiler gibbing gas. See linked PR or changelog for more details. See Testing Photographs and Procedure for some screenshots.

The core change in this PR is that obj/effect/particle_effect/smoke/xeno_burn/Initialize now looks at the faction of the xeno that caused the smoke in the cause_data (resolve_mob would also work except for boiler gibbing smoke) to determine what faction the new smoke should be.

Boilers are basically unusable currently if they aren't the normal hive. The IFF fixes to the gas allow this boiler to not hurt their own hive with their abilities.

<details>

Some cause data examples:
image

IFF'd acid gas:
image
image

Boiler gibbing gas is now IFF'd too:
image
image
</details>

# Changelog
:cl: TeDGamer Drathek
fix: Boilers from other hives generate proper colored IFFed gas
fix: Fixes some death causes with boiler acid gas (for bombard, shroud, and gibbing), neurotoxin gas/stabs, acid shotgun, and xeno spits
balance: Neurotoxin stops processing only in dead mobs
/:cl:


code/modules/mob/living/carbon/xenomorph/abilities/boiler/boiler_powers.dm
```diff
@@ -347,7 +346,7 @@
 		else if(stabbing_xeno.ammo == GLOB.ammo_list[/datum/ammo/xeno/boiler_gas])
 			var/datum/effects/neurotoxin/neuro_effect = locate() in carbon_target.effects_list
 			if(!neuro_effect)
-				neuro_effect = new /datum/effects/neurotoxin(carbon_target)
+				neuro_effect = new /datum/effects/neurotoxin(carbon_target, owner)
```
harryob, line 349:
```suggestion
				neuro_effect = new(carbon_target, owner)
```

code/modules/mob/living/carbon/xenomorph/abilities/boiler/boiler_powers.dm
```diff
@@ -283,27 +284,25 @@
...
 		return
 
-	X.visible_message(SPAN_XENOWARNING("The [X] fires a blast of acid at [A]!"), SPAN_XENOWARNING("You fire a blast of acid at [A]!"))
-
-	var/turf/target = locate(A.x, A.y, A.z)
-	var/obj/item/projectile/P = new /obj/item/projectile(X.loc, create_cause_data(initial(X.caste_type), X))
+	xeno.visible_message(SPAN_XENOWARNING("The [xeno] fires a blast of acid at [target]!"), SPAN_XENOWARNING("You fire a blast of acid at [target]!"))
 
+	var/turf/target_turf = locate(target.x, target.y, target.z)
+	var/obj/item/projectile/proj = new /obj/item/projectile(xeno.loc, create_cause_data("acid shotgun", xeno))
```
harryob, line 301:
```suggestion
	var/obj/item/projectile/proj = new(xeno.loc, create_cause_data("acid shotgun", xeno))
```

code/datums/effects/neurotoxin.dm
```diff
@@ -21,9 +21,8 @@
 	/// Stamina damage per tick. Major balance number.
 	var/stam_dam = 7
 
-/datum/effects/neurotoxin/New(atom/thing)
-	..(thing)
-	cause_data = create_cause_data("neurotoxic gas")
+/datum/effects/neurotoxin/New(atom/thing, mob/from = null)
+	..(thing, from, effect_name)
```
harryob, line 25: not entirely sure why this is preferable to

```suggestion
	. = ..()
	
	cause_data = create_cause_data(effect_name)
```

generally messing with the ..() signature can result in some particularly weird issues - but if this is fixing an issue i'm overlooking lmk

code/modules/mob/living/carbon/xenomorph/abilities/boiler/boiler_powers.dm
```diff
@@ -166,9 +167,9 @@
 	if(!actually_moving)
 		return
 
-	var/obj/effect/particle_effect/smoke/S = new /obj/effect/particle_effect/smoke/xeno_burn(get_turf(mover), 1, create_cause_data(initial(mover.caste_type), mover))
-	S.time_to_live = 3
-	S.spread_speed = 1000000
+	var/obj/effect/particle_effect/smoke/smoke_effect = new /obj/effect/particle_effect/smoke/xeno_burn(get_turf(mover), 1, create_cause_data("dumped acid gas", mover))
```
harryob, line 170:
```suggestion
	var/obj/effect/particle_effect/smoke/xeno_burn/smoke_effect = new(get_turf(mover), 1, create_cause_data("dumped acid gas", mover))
```

harryob (changes requested):

code/datums/effects/neurotoxin.dm
Drulikar: The parent constructor takes four arguments, this constructor is taking two: but in the case of this we want the third argument to always be effect_name and we don't care about the fourth argument.

Also the cause_data should contain the mob from.

Ultimately changing this will mean we waste the construction of cause_data in the parent because we need to redo it if we don't provide those three arguments to the parent constructor.

Drulikar: Currently reverted this change since I think its better to reduce the amount of times cause_data is constructed.

PR #3580, opened 2023-06-09
fixes ridged armor not appearing properly: or, how i learned to regret my own actions

NewyearnewmeUwu:
# About the pull request
Oops! Even more armor prefs bugs

image
<details>

</details>

# Changelog
:cl:
fix: fixed a bug with armor prefs
/:cl:

### DIFF
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -184,7 +184,7 @@ var/list/squad_colors_chat = list(rgb(230,125,125), rgb(255,230,80), rgb(255,150
 		new_look = GLOB.armor_style_list[user.client.prefs.preferred_armor]
 
 	icon_state = replacetext(icon_state,"1","[new_look]")
-	update_icon(user ? user : null)
+	update_icon(user)
 
 /obj/item/clothing/suit/storage/marine/pickup(mob/user)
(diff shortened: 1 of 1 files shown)


code/modules/clothing/suits/marine_armor.dm
```diff
@@ -184,7 +184,7 @@ var/list/squad_colors_chat = list(rgb(230,125,125), rgb(255,230,80), rgb(255,150
 		new_look = GLOB.armor_style_list[user.client.prefs.preferred_armor]
 
 	icon_state = replacetext(icon_state,"1","[new_look]")
-	update_icon()
+	update_icon(user ? user : null)
```
Drulikar, line 187:
```suggestion
	update_icon(user)
```
Unless theres some other nuance to this, such as 0 being a possibility and it must be null, its unnecessary to perform this check if all you're going to return is null.

NewyearnewmeUwu: It doesn't update the onmob for the armor for ridged armor for some godforsaken reason unless I do this.

Drulikar (approved):
If this was an example of the problem (and the light getting turns on fixes it), then yes this PR makes it so the sprite is correct from the start and does not get altered by turning on your armor light.

light

PR #3584, opened 2023-06-10
Explosive barricade upgrade buff

ihatethisengine:
# About the pull request

Explosive barricade upgrade is renamed to composite upgrade, explosion damage taken is now modified by 0.5 (was 0.75), added 0.5 anti-projectile modifier (brute barricade upgrade basically has 0.75). Latest only applies to brute projectiles, so acid damage stays the same. Also added 0.5 anti-fire modifier (acid barricade upgrade basically has 0.75 as well).

The explosive upgrade is almost never used because it is useless. This is a very weak insurance in case of a very bad situation, like an explosion inside the cades. Now it will be more like a dedicated anti-FF upgrade, so smartgunners or flamethrowers, for example, won't be grinding down cades so fast. Xenos won't be affected much because the only brute projectile they have is bone shards and those are not used much against cades.

<details>

</details>

# Changelog
:cl: ihatethisengine
balance: explosive barricade upgrade provides better protection against explosions (25% to 50%)
balance: explosive barricade upgrade provides strong protection against brute-based projectiles (50%)
balance: explosive barricade upgrade provides strong protection against fire (50%)
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade/barricade.dm
+++ b/code/game/objects/structures/barricade/barricade.dm
@@ -96,7 +96,7 @@
 				overlays += image('icons/obj/structures/barricades.dmi', icon_state = "+burn_upgrade_[damage_state]")
 			if(BARRICADE_UPGRADE_BRUTE)
 				overlays += image('icons/obj/structures/barricades.dmi', icon_state = "+brute_upgrade_[damage_state]")
-			if(BARRICADE_UPGRADE_EXPLOSIVE_AND_PROJECTILES)
+			if(BARRICADE_UPGRADE_EXPLOSIVE_AND_PROJECTILE)
 				overlays += image('icons/obj/structures/barricades.dmi', icon_state = "+explosive_upgrade_[damage_state]")
 
 	if(is_wired)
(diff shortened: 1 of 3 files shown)


ihatethisengine (commented):
oops what did I do

fira (approved):

ihatethisengine:
Added 50% protection against fire as well. Flame destroys cades very quickly, so this is another niche anti-FF use for the upgrade.

ihatethisengine:
If numbers are too busted, I can change them upon request.

code/game/objects/structures/barricade/barricade.dm
```diff
@@ -252,7 +254,7 @@
 	if(P.ammo.damage_type == BURN)
```
Drulikar, line 254: Please also fix the single letter vars here. See https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#variables
You can use VSC to do this easily by selecting the function, search, enable dropdown for replace, enable word, enable case, and then just for P and replace with `bullet`or something.

Drulikar (changes requested):
Everything else about this code wise looks fine.

code/game/objects/structures/barricade/barricade.dm
ihatethisengine: Done...

ihatethisengine: Or not

Drulikar (approved):

PR #3592, opened 2023-06-11
Larva surge is limited by marines/xenos ratio

ihatethisengine:
# About the pull request

Xenos after hijack now get larva based on marines/xenos ratio. Instead of infinite larva, larva surge will try to increase the initial amount of xenos on hijack to 50% of marines forces over time (with a minimum of 5 larvas, if xenos already have good numbers).

Initially, if I remember correctly, larva surge was brought into the game to discourage marines from early meta-evacuations, which is fair. But consequently, it really hurt the hijack sequence. Even if marines evac fair and square, larva surge still comes in action and makes situation for marines even worse, utterly discouraging everything but either boomrushing the Alamo or holding lifeboats to evac. 

This resulted in hijacks being very repetitive and boring. More than that, larva surge is extremely busted on lowpop due to the fact you can get around 20 xenos from nothing, making lowpop hijack even less interesting. So with this change marines will still get punished for evaccing with good numbers, but won't be penalized as much for honest evacuations. 

So hopefully, we will see more variety of hijacks and more interesting stories!

P.S. if you have a better formula, let me know.

<details>
My friend @Diegoflores31 tested this for me, thanks!

</details>

# Changelog
:cl: ihatethisengine
balance: larva surge is limited by marines/xenos ratio
fix: xenos no longer get free larva from abandoned facehuggers during hijack
/:cl:


Huffie56:
i don't know if you know but the ratio at round start is 1 xeno for 3 marines. you probably want to aim for atleast that as after hijack marines are holding strong points... your choice your PR...
also please edit your explain why it's good chapter : right now it's just a big bloc of text very hard to read.

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -285,6 +285,7 @@
 	var/larva_gestation_multiplier = 1
 	var/bonus_larva_spawn_chance = 1
 	var/hijack_burrowed_surge = FALSE //at hijack, start spawning lots of burrowed
+	var/hijack_burrowed_left = 0 //how many burrowed is going to spawn during larva surge
```
fira, line 288:
```suggestion
    /// how many burrowed is going to spawn during larva surge
	var/hijack_burrowed_left = 0 
```

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -907,7 +908,21 @@
...
 			embryo.hivenumber = XENO_HIVE_FORSAKEN
 		potential_host.update_med_icon()
+	var/shipside_humans_count = 0
+	var/xenos_count = 0
+	for(var/mob/living/carbon/human/current_human as anything in GLOB.alive_human_list)
+		if(isspecieshuman(current_human) || isspeciessynth(current_human))
+			var/atom/where = current_human
+			if (where.z == 0 && current_human.loc)
+				where = current_human.loc
+			if(is_mainship_level(where.z))
```
fira, line 918: what? just use get_turf

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -907,7 +908,21 @@
...
+				shipside_humans_count++
+	for(var/mob/living/carbon/xenomorph as anything in totalXenos)
```
fira, line 920: typing not valid

code/modules/cm_aliens/structures/special/pylon_core.dm
```diff
@@ -148,12 +148,17 @@
 			if(players_with_xeno_pref && players_with_xeno_pref.len && can_spawn_larva())
 				spawn_burrowed_larva(pick(players_with_xeno_pref))
 
-		if(linked_hive.hijack_burrowed_surge && (last_surge_time + surge_cooldown) < world.time)
+		if(linked_hive.hijack_burrowed_surge && linked_hive.hijack_burrowed_left > 0 && (last_surge_time + surge_cooldown) < world.time)
```
fira, line 151: Technically this is redundant if you disable the surge anyway

fira (changes requested):

fira:
Also i second 1:5 doesn't sound like a good ratio, but i'm no balanss person

ihatethisengine:
I mean, before there was no larva surge at all.  This has never been a big issue. If you hijack, usually it means you have already won and have enough xenos to deal with the rest of marines. So 1 per 5 over hive count sounds pretty fair for me.

For example, we have 30 xenos against 90 marines. This is pretty winnable ratio on its own. 60/5 = 12. After twelve minutes we would have 42 xenos (if no one dies of course). If we have 20 xenos against 100 marines (probably evac was not really justified), larva surge would grant 80/5=16 xenos. If anything, this is still too many, not too few. On low pop we can get situation like 10 xenos vs 30 marines. 20/5=4. A huge number for lowpop, but at least not as overwhelming as infinite amount.

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
ihatethisengine: I think I stole it from bioscan

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
ihatethisengine: Do you want me to rename totalXenos list?

Huffie56:
Redoing my argument.
1-You also need to consider that those new xeno probably won't be able to evolve to T2 since queen will be off ovi.
2-each time a marine will die it will reduce the maximum of xeno hive can get.
3- let's say their is 50 marines left shispide when you hijack.
with 1:5 ratio that would mean you will gain only up to 10 xeno to assault their position.
in my opinion this is to much of a nerf and would be close to removing the larva surge...

ihatethisengine:
2. Marines dying won't reduce the number, the number is being generated only once, when queen hijacks.
3. This is not quite how it works, I might need to rephrase it. But anyway, you get xenos **for free**. And 10 xenos is more than enough and anyway hijack rarely takes more than 10 minutes. More than that, okay, we got xenos POV, they need more xenos to assault marines' position. But what about marines? Are they just bound to fail anyway? Or maybe they should have a chance to hold out?

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
ihatethisengine: Hopefully that will work

Huffie56:
> 2. Marines dying won't reduce the number, the number is being generated only once, when queen hijacks.
> 
>     3. This is not quite how it works, I might need to rephrase it. But anyway, you get xenos **for free**. And 10 xenos is more than enough and anyway hijack rarely takes more than 10 minutes. More than that, okay, we got xenos POV, they need more xenos to assault marines' position. But what about marines? Are they just bound to fail anyway? Or maybe they should have a chance to hold out?

i didn't understood what you where doing at first it seem so i just took a look at the math in your code.
hijack_burrowed_left = max(n_ceil((ship side_humans_count - xenos_count) / 5), 1)

1-wouldn't it be more clear if you calculate the ratio you desire first?
desired number of xeno depending on ratio = ship-side_humans_count/5
2-then you calculate if the xeno have enough numbers.
if(xenos_count<desired xeno number
3-then you fill the larva surge pool as intended to respect the ratio.
surge pool max number = desired number of xeno - xeno_count

i don't know if it's clear.... just ignore if it's not.

ihatethisengine:
Hmm, interesting suggestion, but I am not quite sure what number or ratio should be desirable. 1/5 is not enough for sure, 1/3 sounds better, but then hives with better ratio won't get larva at all. Which is not bad on its own (they already have numbers), but is probably too much of a nerf. My idea would at least give some larva even if hive is strong enough, without giving too many.

Huffie56:
> Hmm, interesting suggestion, but I am not quite sure what number or ratio should be desirable. 1/5 is not enough for sure, 1/3 sounds better, but then hives with better ratio won't get larva at all. Which is not bad on its own (they already have numbers), but is probably too much of a nerf. My idea would at least give some larva even if hive is strong enough, without giving too many.

you could calculate the number you give them differently if they have more than the ratio you aim for?
maybe a fix number?

ihatethisengine:
Fixed number won't work because we have lowpop and highpop. Although 5 seems to be fine. I'll think about this. And let's see what balance dev has to say.

Wintermote:
This PR has been a long time coming, thank goodness.

The original hijack system made any progress at all during a shipside fight pointless unless you could kill xenos at stupid rates of attrition, like 2 xeno per minute if you wanted to do better than break even.

My biggest issue is how long it took for this PR to happen. Did someone have to die for us to get this fix? The amount of absurd arguments I've seen to protect the xeno autowin condition on hijack has left a very sour taste in my mouth. Anyway, tying the total hijack surge to marine force disposition is a more elegant way of providing xenos an advantage on hijack without turning it into straight up (pointless) seal clubbing on marines.

However, I think we need to take into consideration that hijack larva while theoretically infinite, was not infinite in practice - with 1 larva per minute (or was it 2minutes?) that's only +10 or +20 possible xenos before the round automatically concludes by timer (or whatever length the timer is, I'm pretty sure it's 20 minutes). Which really puts into perspective how little xenos needed hijack surge and how big an impact even a handful of xenos have.

With that in consideration I actually think your numbers have the potential to make things even worse for hijack.

Additionally, about 15% of the shipside crew are not marines and cannot fight effectively. Doctors, nurses, cargo techs, maintenance techs, and so forth not only don't have proper gear for fighting, they also lack skills for it. 90 shipside marines typically means 75 shipside marines (including MP's, Staff Officers, Intel Officers, etc. with combat training) and 15ish non combatants.

In my opinion xenos should only get a hijack surge of larva if there's a large disparity in xeno numbers vs marine numbers. Balance wise, 1 xeno is worth about 3 marines and to give xenos an advantage but not an autowin I'd try a 1:2.5 ratio. Which means in your cited example of 30 xenos vs 90 marines, I think the xenos should get (90/2.5) - 30 = 6 hijack larva; not 12. Adjust from there as needed but I wouldn't go under a 1:2 ratio, personally. 

An open air question is whether the hijack surge should be all at once or staggered; there's only 20 minutes so I think all at once may be better, but staggered gives marines incentive to push to and eliminate the hive core to deny reinforcements, and I think staggered would work better with lower xeno:marine ratios (e.g. 1:2) as it provides some counterplay and requires xeno strategizing about protection of the hive core.

And speaking on the hive core, though better in its own PR - I think we should prevent it being built in the dropship particularly because it's getting placed inside the cockpit where most marines can't even open the door to get to it, but also because I think it makes the xenos plan out what they're doing and where they land, so it's not simply "land closest to marines" anymore but now also where are they going to nest and how will it be protected, that comes in as something to think about - which in my view makes it more engaging and varied for both sides.

ihatethisengine:
Good points, changed formula to `hijack_burrowed_left = max(n_ceil(shipside_humans_count * 0.4) - xenos_count, 5)` (* 0.4 and / 2.5 is the same)

morrowwolf (changes requested):
Make it go to 50% of shipside population. Thanks.

ihatethisengine:
And done!

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -907,7 +909,19 @@
...
 			embryo.hivenumber = XENO_HIVE_FORSAKEN
 		potential_host.update_med_icon()
+	var/shipside_humans_count = 0
+	var/xenos_count = 0
+	for(var/mob/living/carbon/human/current_human as anything in GLOB.alive_human_list)
+		if(isspecieshuman(current_human) || isspeciessynth(current_human))
+			var/turf/turf = get_turf(current_human)
+			if(is_mainship_level(turf?.z))
+				shipside_humans_count++
+	for(var/mob/living/carbon/xenomorph/xenomorph as anything in totalXenos)
```
Drulikar, line 919: Get the count of xenos when the list is iterated on line 888; also only count xenos that are hijacking (that loop is already figuring out who isn't - e.g. just `continue` after `stored_larva++` and then tabbed one left add your tally `xeno_count++`).

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -907,7 +909,19 @@
...
+		if(isfacehugger(xenomorph))
+			continue
+		xenos_count++
 	hijack_burrowed_surge = TRUE
+	hijack_burrowed_left = max(n_ceil(shipside_humans_count * 0.5) - xenos_count, 5)
```
Drulikar, line 924: To compensate for the difference between a combat marine and a non-combat marine consider weighing the roles like is done with `latejoin_tally` via `RoleAuthority.calculate_role_weight`

code/modules/cm_aliens/structures/special/pylon_core.dm
```diff
@@ -151,9 +151,14 @@
 		if(linked_hive.hijack_burrowed_surge && (last_surge_time + surge_cooldown) < world.time)
 			last_surge_time = world.time
 			linked_hive.stored_larva++
+			linked_hive.hijack_burrowed_left--
 			announce_dchat("The hive has gained another burrowed larva! Use the Join As Xeno verb to take it.", src)
 			if(surge_cooldown > 30 SECONDS) //mostly for sanity purposes
 				surge_cooldown = surge_cooldown - surge_incremental_reduction //ramps up over time
+		if(linked_hive.hijack_burrowed_surge && linked_hive.hijack_burrowed_left < 1)
```
Drulikar, line 158: There isn't much reason why this can't just be done underneath the if block on line 151 `if(linked_hive.hijack_burrowed_surge && (last_surge_time + surge_cooldown) < world.time)` as just `if(linked_hive.hijack_burrowed_left < 1)` tabbed over one to this right where it is now.

code/modules/cm_aliens/structures/special/pylon_core.dm
```diff
@@ -151,9 +151,14 @@
...
+			linked_hive.hijack_burrowed_surge = FALSE
+			xeno_message(SPAN_XENOANNOUNCE("Hive surge is fading away. You will no longer gain burrowed larva over time."), 3, linked_hive.hivenumber)
```
Drulikar, line 160: The wording of this could be improved especially since its not currently called this in character at least as far as I can tell. E.g. "The hive's power wanes. You will no longer gain pooled larva over time."

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -907,7 +909,19 @@
...
+	for(var/mob/living/carbon/human/current_human as anything in GLOB.alive_human_list)
+		if(isspecieshuman(current_human) || isspeciessynth(current_human))
+			var/turf/turf = get_turf(current_human)
+			if(is_mainship_level(turf?.z))
+				shipside_humans_count++
+	for(var/mob/living/carbon/xenomorph/xenomorph as anything in totalXenos)
+		if(isfacehugger(xenomorph))
+			continue
+		xenos_count++
 	hijack_burrowed_surge = TRUE
```
Drulikar, line 923: This should only be set TRUE if we will allow even a single larva to spawn. Consider not only revising the message `xeno_message(SPAN_XENOANNOUNCE("The hive swells with power! You will now steadily gain pooled larva over time."), 2, hivenumber)` but also put it here depending on if we will be getting a surge rather than just first tick hitting the surge end. If we won't get a surge just have a unique message for that or just no message at all.

Drulikar (changes requested):
This is not something that can be tested well locally and must be test merged once ready to get a feel for what needs tweaking (e.g. surge_cooldown, surge_inccremental_reduction, and hijack_burrowed_left).

code/modules/cm_aliens/structures/special/pylon_core.dm
ihatethisengine: Done

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
ihatethisengine: Done, I think? Should work this way

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
Drulikar: Ok, though also still consider editing that message since it still suggests the surge is indefinite. But if nothing is immediately obvious on how to better word it, its fine.

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
ihatethisengine: Done and I think I fixed fugger issue, although probably it can be coded better. Please take a look.

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
ihatethisengine: Only thing that comes to my mind is that I can swap lines 277 and 278 in dropship_computer.dm and tell the exact number of larvae the hive will get.

Drulikar: I don't think I would want it to spit out the exact count. So its fine - ultimately just wanted you to be aware how it was worded elsewhere.

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
Drulikar: Be sure to add this to your changelog as a fix

Drulikar (approved):
Changes look fine in code - I'm sure Morrow will be on top of how it feels in practice.

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
ihatethisengine: Oh wait forgot another check

ihatethisengine: Now it should work

Drulikar: Yeah that makes sense 👍

PR #3594, opened 2023-06-11
Limits view range of large crate destroy message.

### LINKED ISSUE #3387: bug when a xeno destroy a crate every xeno get the message in their log.
### Testmerges

not relevant

### Description of the bug

bug when a xeno destroy a crate every xeno get the message in their log.

it should be restricted to the xeno present.

1.another xeno break a crate
2.see the message in your log
3.
...

### Issue Bingo

TheGamerdk:
# About the pull request

Oversight in message flags, fixes #3387

Bugs aren't so good

</details>

# Changelog

:cl:
fix: Xenos are no longer notified each time a sister destroys a large crate.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/largecrate.dm
+++ b/code/game/objects/structures/crates_lockers/largecrate.dm
@@ -64,7 +64,7 @@
 	M.animation_attack_on(src)
 	unpack()
 	M.visible_message(SPAN_DANGER("[M] smashes [src] apart!"), \
-					  SPAN_DANGER("You smash [src] apart!"), 5, CHAT_TYPE_XENO_COMBAT)
+					  SPAN_DANGER("You smash [src] apart!"), 5, message_flags = CHAT_TYPE_XENO_COMBAT)
 	return XENO_ATTACK_ACTION
 
 /obj/structure/largecrate/ex_act(power)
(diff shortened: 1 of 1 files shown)


harryob (approved):
evidently this was a hive critical message everyone Needed to know

code/game/objects/structures/crates_lockers/largecrate.dm
```diff
@@ -64,7 +64,7 @@
 	M.animation_attack_on(src)
 	unpack()
 	M.visible_message(SPAN_DANGER("[M] smashes [src] apart!"), \
-					  SPAN_DANGER("You smash [src] apart!"), 5, CHAT_TYPE_XENO_COMBAT)
+					  SPAN_DANGER("You smash [src] apart!"), 5, message_flags = CHAT_TYPE_XENO_COMBAT)
```
fira, line 67: Uh wait, there's more, this uses the /mob variant of visible_message() which has a different signature

`/mob/visible_message(message, self_message, blind_message, max_distance, message_flags = CHAT_TYPE_OTHER)`

you need to add the blind_message or a null in there to match, it's all off by one, not juste the message_flags

fira (changes requested):

code/game/objects/structures/crates_lockers/largecrate.dm
TheGamerdk, line 67:
```suggestion
					  SPAN_DANGER("You smash [src] apart!"), null, 5, CHAT_TYPE_XENO_COMBAT)
```

PR #3597, opened 2023-06-11
You can now use the Ghost verb when nested and still return to your body when freed

### LINKED ISSUE #1375: Freeing From Hive Doesn't Set You Back Into Body
## Description
Simple bug. Destroying the resin nest instead of doing the proper anim for it doesn't let someone who's ghosted back into their body.

It should let you back into your body after the nest is broken. Instead, if you break it with your knife instead of using a welder or etc to cut someone loose it keeps them ghosted.

## Steps to reproduce
- Get Capped
- Ghost
- Get freed via someone breaking the nest

### LINKED ISSUE #1552: Ghosting as infected is counter-intuitive
When you are infected and nested, you are given the opportunity to ghost. 

This uses a specific combination of `can_reenter_corpse` to avoid metagaming: the relation between ghost and body is preserved, but it is flagged as non-reenterable until freed. Otherwise a captured human could ghost/reenter at will to give out information on radio, for example.

This logic is handled by the nest. The logic for `Ghost` verb is separate and does not allow this. This means if a player says no to the popup then tries to ghost on their own later, they lose the ghost-body relationship for good.

Ahelps in past year have shown a recurrent tendency for this to happen.

Observed problems can likely be explained by:
 * The player doesn't get the nest ghosting alert() or closes it by accident
 * The player does not decide to ghost when getting alert, and assume they can later
 * The player did ghost via the popup, and tries to reenter but doesn't understand why they're unable

## Proposed resolution

Ideally, ghosting in nests should be refactored to also take into account self ghosting.

Additionally, Reenter-Corpse could provide a more specific message about it.

TheGamerdk:
# About the pull request

This is just confusing. Fixes #1552

Less confusion good

</details>

# Changelog

:cl:
fix: You'll now be allowed to reenter your nested body when freed, even if you used the ghost verb.
/:cl:

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -388,7 +388,7 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 			options = list("Aghost") + options
 		var/text_prompt = "Are you -sure- you want to ghost?\n(You are alive. If you ghost, you won't be able to return to your body. You can't change your mind so choose wisely!)"
 		var/is_nested = (buckled && istype(buckled, /obj/structure/bed/nest))
-		var/nest
+		var/obj/structure/bed/nest/nest
 		if(is_nested)
 			text_prompt += "\nSince you're nested, you will be given a chance to reenter your body upon being freed."
 			nest = buckled
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -402,10 +402,10 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 		if(location) //to avoid runtime when a mob ends up in nullspace
 			msg_admin_niche("[key_name_admin(usr)] has ghosted. [ADMIN_JMP(location)]")
 		log_game("[key_name_admin(usr)] has ghosted.")
-		var/mob/dead/observer/ghost = ghostize(is_nested) //FALSE parameter is so we can never re-enter our body, "Charlie, you can never come baaaack~" :3
+		var/mob/dead/observer/ghost = ghostize((is_nested && nest && !QDELETED(nest))) //FALSE parameter is so we can never re-enter our body, "Charlie, you can never come baaaack~" :3
 		if(ghost && !is_admin_level(z))
 			ghost.timeofdeath = world.time
-		if(is_nested && nest)
+		if(is_nested && nest && !QDELETED(nest))
 			ghost.can_reenter_corpse = FALSE
 
 /mob/dead/observer/Move(atom/newloc, direct)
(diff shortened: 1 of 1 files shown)


code/modules/mob/dead/observer/observer.dm
```diff
@@ -386,7 +386,13 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 		var/list/options = list("Ghost", "Stay in body")
 		if(check_rights(R_MOD))
 			options = list("Aghost") + options
-		var/response = tgui_alert(src, "Are you -sure- you want to ghost?\n(You are alive. If you ghost, you won't be able to return to your body. You can't change your mind so choose wisely!)", "Are you sure you want to ghost?", options)
+		var/text_prompt = "Are you -sure- you want to ghost?\n(You are alive. If you ghost, you won't be able to return to your body. You can't change your mind so choose wisely!)"
+		var/is_nested = (buckled && istype(buckled, /obj/structure/bed/nest))
+		var/nest
+		if(is_nested)
+			text_prompt += "\nSince you're nested, you will be given a chance to reenter you body upon being freed."
```
harryob, line 393:
```suggestion
			text_prompt += "\nSince you're nested, you will be given a chance to reenter your body upon being freed."
```

harryob (changes requested):

Huffie56:
just want to point out this issue that is related in case you didn't saw it....
unsure if your PR fix it?
https://github.com/cmss13-devs/cmss13/issues/1552

fira (changes requested):
This is the same way the nests themselves do it, and as noted in the linked issue it appears to sometimes fail for some reason - players in ahelps report they click yes on the nest prompt and can't reenter (i made extra sure, but could never reproduce myself)

This is still a good step forward in being less confusing to people

You should check nest also after the alert to avoid race conditions though

TheGamerdk:
> This is the same way the nests themselves do it, and as noted in the linked issue it appears to sometimes fail for some reason - players in ahelps report they click yes on the nest prompt and can't reenter (i made extra sure, but could never reproduce myself)
> 
> This is still a good step forward in being less confusing to people
> 
> You should check nest also after the alert to avoid race conditions though

Cursory glance at code shows it should be fine unless your ghost is recreated (which is possible, might be more wise to use the mind, that shouldn't get recreated iirc)

Added a race check condition to the ghostize() call

code/modules/mob/dead/observer/observer.dm
```diff
@@ -386,7 +386,13 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 		var/list/options = list("Ghost", "Stay in body")
 		if(check_rights(R_MOD))
 			options = list("Aghost") + options
-		var/response = tgui_alert(src, "Are you -sure- you want to ghost?\n(You are alive. If you ghost, you won't be able to return to your body. You can't change your mind so choose wisely!)", "Are you sure you want to ghost?", options)
+		var/text_prompt = "Are you -sure- you want to ghost?\n(You are alive. If you ghost, you won't be able to return to your body. You can't change your mind so choose wisely!)"
+		var/is_nested = (buckled && istype(buckled, /obj/structure/bed/nest))
+		var/nest
```
fira, line 391: need typing here to avoid errors

TheGamerdk, line 391:
```suggestion
		var/obj/structure/bed/nest/nest
```

harryob (approved):

PR #3599, opened 2023-06-11
Sentries now work correctly on dropships

### LINKED ISSUE #436: [BUG] Broken DS sentries - Electric Boogaloo 2
## Description
DS modular sentries and engineer's hand-held devices do not trigger on xenos after / or during DS transit.

Subj. Some modules and hand-held devices should work like under normal conditions, but for some reason those hand-held devices and modules cease to function at some point.

## Steps to reproduce
1) Install sentry as a module or place it as a hand-held device on Alamo or Normandy. 
2) Activate it. 
3) Complete cycling from surface to ship or from ship to surface.
4) Upon landing or during transit sentries would stop registering xenos as a targets, until you manually re-deploy them with multi-tool (in case of engineer's devices).

It's an old issue about which I reported 1 year ago, but didn't got any response, so it still present - https://gitlab.com/cmdevs/colonial-warfare/-/issues/537.

### LINKED ISSUE #3554: new anti-DS-sentry thing griefs the M56D, and it's not in a way that should happen
### Testmerges

#3470, #3540, #3537, #3298

### Description of the bug

whatever update removed the ability to place sentries in dropships eats your M56D mount making it unusable

one, i feel like comtechs are already getting enough shit, with the whole "not being able to do their one and only purpose" when it comes to sentries not working (cades are already gone, now its all engineer things full stop)

two, once the comtech places the M56D in the dropship, the mount ✨ CANNOT ✨ be removed and the only option is to unmount the gun and craft a new mount (which is stupid)

so either bring M56D use on dropships back or make them impossible to place entirely so people don't do it smh

1. enter a dropship
2. place an M56D
3. congratulations, your mount has been eaten, i hope you have 5 plasteel and 10 metal that you're not using to build cades 👍 (sarcasm, if you need to place a 56D in a dropship there's no way you have mats to spare)

### Issue Bingo

TheGamerdk:
# About the pull request

The sentries work by creating bounds and using those to check if anyone is in range. Since X and Y coordinates change when the dropships move, this would break until you redeployed the sentry.

Partially fixes #3554 and fixes #436 (Can't reproduce M56D not working)

Bugs bad.

# Changelog

:cl:
fix: Sentries now work correctly on dropships.
/:cl:

### DIFF
--- a/code/modules/defenses/planted_flag.dm
+++ b/code/modules/defenses/planted_flag.dm
@@ -83,6 +83,7 @@
 		apply_buff_to_player(H)
 
 /obj/structure/machinery/defenses/planted_flag/proc/turf_changed()
+	SIGNAL_HANDLER
 	if(range_bounds)
 		QDEL_NULL(range_bounds)
 
--- a/code/modules/defenses/planted_flag.dm
+++ b/code/modules/defenses/planted_flag.dm
@@ -27,6 +27,8 @@
 /obj/structure/machinery/defenses/planted_flag/Initialize()
 	. = ..()
 
+	RegisterSignal(src, COMSIG_ATOM_TURF_CHANGE, PROC_REF(turf_changed))
+
 	if(turned_on)
 		apply_area_effect()
 		start_processing()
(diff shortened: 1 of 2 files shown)


Huffie56:
sentries and M56D aren't suppose to be assemble or use on the DS.
except if they are modular(the one you need to attach via a work-loader)
image

TheGamerdk:
> sentries and M56D aren't suppose to be assemble or use on the DS. except if they are modular(the one you need to attach via a work-loader) image

That's a different issue, this is just fixing that sentries stop working when their position is changed mid-deployment.

fira (changes requested):
use SIGNAL_HANDLER annotations for signal enabled procs

morrowwolf:
This doesn't seem to allow sentries on/in shuttles/vehicles but instead fixes sentries breaking down after shuttle movement. Not a balance issue.

TheGamerdk:
> use SIGNAL_HANDLER annotations for signal enabled procs

Done

harryob (approved):

PR #3600, opened 2023-06-11
Cluster OB now respects if a hive core is created while it is firing.

TheGamerdk:
# About the pull request

It checks for OB protection on the turf now. It'll remove the turf from consideration and try a different one. If no valid turfs remain the OB stops.

Seems fair enough, very niche case though.

# Changelog

:cl:
balance: Cluster OBs will now no longer hit turfs that have gotten OB protection after the initial OB was fired.
/:cl:

### DIFF
--- a/code/modules/cm_marines/orbital_cannon.dm
+++ b/code/modules/cm_marines/orbital_cannon.dm
@@ -484,9 +484,6 @@ var/list/ob_type_fuel_requirements
 	var/list/turf_list = list()
 
 	for(var/turf/T in range(range_num, target))
-		if(protected_by_pylon(TURF_PROTECTION_OB, T))
-			continue
-
 		turf_list += T
 
 	for(var/i = 1 to total_amount)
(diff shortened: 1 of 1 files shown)


fira (approved):
I'd think the core should just eat the shell because otherwise this could deflect shells into a concentrated barrage, but well, alright

TheGamerdk:
> I'd think the core should just eat the shell because otherwise this could deflect shells into a concentrated barrage, but well, alright

We can do that too, but then you'd lose out on the part of the OB hitting the non-protected part. As you say, concentrated.

harryob (approved):

TheGamerdk:
> I agree with Fira. This is gonna take some moving around.
> 
> Add all the turfs in the range, then do the regular picking and fire_in_a_hole() calls.
> 
> In fire_in_a_hole() do the check for if the turf is protected by a pylon and then eat the explosion instead of having it all bottled up.
> 
> Maintaining density for this stuff is important for balance, surprised it was like this for so long.

misread this request, oops

PR #3603, opened 2023-06-12
Huggers no longer lose health over time on weeds

Hopekz:
From what I can tell the damage is meant to effect huggers outside of weeds so it is more of a QoL for huggers to make the role more bearable to play on the weeds.

If it is meant to effect huggers on weeds too then I'd like to challenge that idea because I find that part more of a chore vs a fun gameplay aspect.

:cl: Hopek
balance: Huggers no longer lose health over time on weeds.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
@@ -66,7 +66,7 @@
 		PF.flags_can_pass_all = PASS_ALL^PASS_OVER_THROW_ITEM
 
 /mob/living/carbon/xenomorph/facehugger/Life(delta_time)
-	if(stat != DEAD && !lying)
+	if(stat != DEAD && !lying && !locate(/obj/effect/alien/weeds) in get_turf(src))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
```diff
@@ -66,7 +66,7 @@
 		PF.flags_can_pass_all = PASS_ALL^PASS_OVER_THROW_ITEM
 
 /mob/living/carbon/xenomorph/facehugger/Life(delta_time)
-	if(stat != DEAD && !lying)
+	if(stat != DEAD && !lying && !locate(/obj/effect/alien/weeds) in get_turf(src))
```
Zonespace27, line 69:
```suggestion
	if(stat != DEAD && !lying && !(locate(/obj/effect/alien/weeds) in get_turf(src)))
```

Zonespace27 (dismissed):

morrowwolf (approved):
Fine with this from balance/design. As usual if unforeseen issues crop up it'll be axed.

harryob (approved):

PR #3611, opened 2023-06-12
Various Job/Slots balancing fixes

fira:
# About the pull request

Fixes various aspects of job scaling and spawning. This can have some minor balance effects.

MINIMALLY TESTED due to requiring a proper amount of players to go anywhere. TM this and expect it to explode.

See changelog below.

Less unknowns and side effects makes it easier to change game balance.

# Changelog
:cl:
fix: Fixes Alternate Preference job spawn options rarely counting double for roundstart burrowed larva count
fix: Fixes Cryoing not taking role weighting into account for larva removal, such as shipside scaling
fix: Fixes Job positions not being re-computed during job assignment, barring some xeno/pred slots at roundstart
fix: Roundstart larvas calculation now piggyback on larva latejoin tally, allowing fractional larvas
fix: Extraneous Xeno jobs during slot calculation can now result in a latejoin larva penalty to match
fix: Fixes Pooled larvas awkwardly disappearing each time someone latejoined if no Hivecore, now they just stop being counted
/:cl:


fira:
We discussed on discord with sg2002 and i was wrong about most of the effects of the problems this fixed, they are far more subtle than previously said due to bad understanding on my part.

However they technically still remain issues, so this attempts to address them. Comments and procs have been changed to try and make the intent clearer and more accurate to the actual problems.

PR #3621, opened 2023-06-13
Adds doors to the courtyard to perma cells.

Johannes2262:
# About the pull request
This PR replaces the windows between the courtyard and the permanent cells with doors which can be opened when a perma prisoner gets a trip to the courtyard authorized. If needed, I can add new windows next to the door.

According to ML, perma prisoners may be given access to the courtyard as reward for not causing trouble. However, currently the MPs need to cuff the prisoner, move them through the whole brig and put them into the courtyard through the warden's office, same when the prisoner gets moved back to the cell. With this PR, MPs can just open the cell door and let the prisoner enter the courtyard.

</details>

# Changelog

:cl: Johannes2262
maptweak: Replaced the windows between perma cells and the courtyard with doors.
/:cl:

### DIFF
--- a/maps/map_files/USS_Almayer/USS_Almayer.dmm
+++ b/maps/map_files/USS_Almayer/USS_Almayer.dmm
@@ -56156,7 +56156,6 @@
 	},
 /area/almayer/shipboard/brig/lobby)
 "pFg" = (
-/obj/structure/window/framed/almayer/hull/hijack_bustable,
 /obj/structure/machinery/door/poddoor/almayer/locked{
 	dir = 8;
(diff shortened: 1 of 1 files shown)


morrowwolf (commented):
Having a perma prisoner in the yard should be an active and conscious choice by MPs. Making it too easy takes away from the seriousness of permanent confinement.

Johannes2262:
So sitting in a room together with maybe two other people for two hours with only a botany minigame and maybe uno is not serious so the prisoner needs to sit in some 3x3 room alone with literally nothing to do for two hours?

PR #3627, opened 2023-06-14
Fix grab delay for xenos

Drulikar:
# About the pull request

This PR changes it so the check for `if(next_move >= world.time)` in `/mob/living/carbon/xenomorph/click` does not return true to "consume" the input preventing the clicked proc (as true it would short circuit - but in this case we didn't perform an action in click).

I do not think this change warrants a need to be test merged, but feel free to do so.

This should address the last remaining unintended change that players have discovered since the introduction of #3516 

<details>

https://github.com/cmss13-devs/cmss13/assets/76988376/9980947a-7d58-48e4-815d-b8c1b953d84c

</details>

# Changelog
:cl: Drathek
fix: Fix grab delay after a tackle for xenos
/:cl:

### DIFF
--- a/code/_onclick/xeno.dm
+++ b/code/_onclick/xeno.dm
@@ -127,7 +127,7 @@ so that it doesn't double up on the delays) so that it applies the delay immedia
 			return TRUE
 
 	if(next_move >= world.time)
-		return TRUE
+		return FALSE
 
 	return ..()
 


morrowwolf:
Just in case, let's TM it.

harryob (approved):

PR #3630, opened 2023-06-14
Forsaken being spawned now swaps all groundside resin to forsaken hive.

NewyearnewmeUwu:
# About the pull request
Forsaken are being slowed down by normal xeno hive weeds and killed by the normal hive's groundside acid traps. This fixes that. It also deletes normal hive groundside acid pillars as a TEMPORARY WORKAROUND because they will delete forsaken weeds even when swapped to forsaken hive.

This was requested by Morrow, and it works.
<details>

</details>

# Changelog
:cl:
add: Forsaken being spawned now swaps all groundside resin to forsaken hive.
/:cl:

### DIFF
--- a/code/__DEFINES/dcs/signals/signals_global.dm
+++ b/code/__DEFINES/dcs/signals/signals_global.dm
@@ -50,3 +50,6 @@
 
 /// From /datum/admins/proc/force_predator_round()
 #define COMSIG_GLOB_PREDATOR_ROUND_TOGGLED "!predator_round_toglged"
+
+// Used to convert all groundside weeds, etc to Forsaken hive.
+#define COMSIG_GLOB_GROUNDSIDE_FORSAKEN_HANDLING "!groundside_forsaken_handling"
--- a/code/__DEFINES/dcs/signals/signals_global.dm
+++ b/code/__DEFINES/dcs/signals/signals_global.dm
@@ -51,5 +51,5 @@
 /// From /datum/admins/proc/force_predator_round()
 #define COMSIG_GLOB_PREDATOR_ROUND_TOGGLED "!predator_round_toglged"
 
-// Used to convert all groundside weeds, etc to Forsaken hive.
+/// From /datum/game_mode/colonialmarines/proc/check_ground_humans()
 #define COMSIG_GLOB_GROUNDSIDE_FORSAKEN_HANDLING "!groundside_forsaken_handling"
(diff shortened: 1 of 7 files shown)


harryob (changes requested):

code/__DEFINES/dcs/signals/signals_global.dm
```diff
@@ -50,3 +50,6 @@
 
 /// From /datum/admins/proc/force_predator_round()
 #define COMSIG_GLOB_PREDATOR_ROUND_TOGGLED "!predator_round_toglged"
+
+// Used to convert all groundside weeds, etc to Forsaken hive.
```
fira, line 54:
```suggestion
/// From /datum/game_mode/colonialmarines/proc/check_ground_humans()
```

fira (approved):

ihatethisengine:
I am not sure if it should change all the weeds. Yesterday we had a really fun round, when I was playing forsaken and two xenos from regular sky came back on a pod. It was a tense fight, partly because they still had their walls, sticky and weeds.

code/game/gamemodes/colonialmarines/colonialmarines.dm
```diff
@@ -273,6 +273,7 @@
 
 	if(groundside_humans > (groundside_xenos * GROUNDSIDE_XENO_MULTIPLIER))
 		SSticker.mode.get_specific_call("Xenomorphs Groundside (Forsaken)", FALSE, FALSE)
+		SEND_GLOBAL_SIGNAL(COMSIG_GLOB_GROUNDSIDE_FORSAKEN_HANDLING)
```
morrowwolf, line 276: This really shouldn't need to be sent more than once on hijack itself. There's likely a better place to send (best guess, wherever it resets the hive core).

morrowwolf (changes requested):

code/game/gamemodes/colonialmarines/colonialmarines.dm
ihatethisengine: Just do that on hijack as I suggested

Segrain:
This should happen _after_ the dropship takeoff and it should skip the weeds on the dropship itself.
Frankly, much simpler solution would be to simply hardcode regular and forsaken hive as allies. Even should a regular xeno somehow end back on the planet, why should they be hostile to each other?

PR #3637, opened 2023-06-15
removed the ability to plant c4 on zombies

Cthulhu80:
# About the pull request

Removes the ability to plant c4 on zombies.

"Player c4d three zombies, says not to do so in the join message as well as other places by Somebody (SomeRole) on 2023-06-15 05:43:53"

# Changelog

:cl:
fix: removed the ability to c4 zombies.
/:cl:


code/game/objects/items/explosives/plastic.dm
```diff
@@ -208,6 +208,11 @@
 			to_chat(user, SPAN_WARNING("ARE YOU OUT OF YOUR MIND?!"))
 			return FALSE
 
+	// Apparently this is metagaming? Lol.
```
harryob, line 211: delete comment

code/game/objects/items/explosives/plastic.dm
```diff
@@ -208,6 +208,11 @@
...
+	if(iszombie(target))
+		to_chat(user, SPAN_WARNING("[target] skin falls off as you attempt to plant the c4"))
```
harryob, line 213:
```suggestion
		to_chat(user, SPAN_WARNING("[target]'s skin falls off as you attempt to plant the C4!"))
```

harryob (changes requested):

Drulikar (approved):

Drulikar (changes requested):
Code as it is fine; but lets just make this an admin toggle instead (so also check if that setting is true before this iszombie check). Not only was the quote regarding an event rule, but this honestly is only an issue for the first few zombies. I don't have any strong feelings one way or another what the default is.

PR #3605, opened 2023-06-12
added camera shake and stun to OB.

Cthulhu80:
# About the pull request

Added a camera shake and stun feature based on the user's distance to the OB.

# Explain why it's good for the game

OBs should feel like OBs, and adding this feature just helps make OBs feel like an actual large explosion. 

# Changelog

:cl:
add: added camera shake and stun to OB.
/:cl:

PR #3658, opened 2023-06-17
added camera shake to ob

Cthulhu80:
# About the pull request

Added a camera shake and stun feature based on the user's distance to the OB. This is a copy of an existing branch that I somehow managed to mess up -> https://github.com/cmss13-devs/cmss13/pull/3605

OBs should feel like OBs, and adding this feature just helps make OBs feel like an actual large explosion. 

# Changelog

:cl:
add: added camera shake and stun to OB.
/:cl:


code/modules/cm_marines/orbital_cannon.dm
```diff
@@ -398,6 +398,51 @@ var/list/ob_type_fuel_requirements
 		return TRUE
 	return FALSE
 
+/obj/structure/ob_ammo/warhead/proc/handle_ob_shake(turf/epicenter)
```
Zonespace27, line 401: codedoc

code/modules/cm_marines/orbital_cannon.dm
```diff
@@ -398,6 +398,51 @@ var/list/ob_type_fuel_requirements
...
+			frequency = 1
+			max_shake_factor = 8
+			max_knockdown_time = 3
+		if("explosive")
+			frequency = 3
+			max_shake_factor = 15
+			max_knockdown_time = 6
+		if("cluster")
+			frequency = 2
+			max_shake_factor = 1
```
Zonespace27, line 421: use subtypes correctly

code/modules/cm_marines/orbital_cannon.dm
```diff
@@ -398,6 +398,51 @@ var/list/ob_type_fuel_requirements
...
+
+	for(var/mob/living/M in urange(radius_size, epicenter))
```
Zonespace27, line 423: no one letter variables

code/modules/cm_marines/orbital_cannon.dm
```diff
@@ -398,6 +398,51 @@ var/list/ob_type_fuel_requirements
...
+		var/total_shake_factor = abs(max_shake_factor * distance_percent)
+
+		// it's of type cluster.
+		if(!max_knockdown_time)
+			shake_camera(M, 0.5, total_shake_factor, frequency)
+			continue
+
+		shake_camera(M, total_shake_factor, frequency)
+
+		var/total_stun_time = max_knockdown_time * distance_percent
```
Zonespace27, line 438: this doesn't need to be a var

Zonespace27 (changes requested):

code/modules/admin/tabs/event_tab.dm
```diff
@@ -897,7 +894,12 @@
 		if(alert("Are you SURE you want to do this? It will create an OB explosion!",, "Yes", "No") != "Yes") return
```
Drulikar, line 894: Since the qdel is removed from above, we ought to be sure the warhead is deleted if the admin changes their mind. Make sure it is either qdeleted or has QDEL_IN set for all scenarios.

code/modules/cm_marines/orbital_cannon.dm
```diff
@@ -398,10 +401,40 @@ var/list/ob_type_fuel_requirements
...
+		var/distance_percent = ((radius_size - distance) / radius_size)
+
+		var/total_shake_factor = abs(max_shake_factor * distance_percent)
+
+		// it's of type cluster.
+		if(!max_knockdown_time)
+			shake_camera(user, 0.5, total_shake_factor, frequency)
+			continue
+
+		shake_camera(user, total_shake_factor, frequency)
```
Drulikar, line 422: Are your parameters correct here? total_shake_factor is currently positionally set to be steps, and frequency is positionally set to be strength. Use named arguments e.g. `strength = total_shake_factor`.

code/modules/cm_marines/orbital_cannon.dm
```diff
@@ -398,10 +401,40 @@ var/list/ob_type_fuel_requirements
...
+/obj/structure/ob_ammo/warhead/proc/handle_ob_shake(turf/epicenter)
+
+	var/radius_size = 30
+
+	for(var/mob/living/user in urange(radius_size, epicenter))
+
+		var/distance = get_accurate_dist(get_turf(user), epicenter)
+
+		var/distance_percent = ((radius_size - distance) / radius_size)
+
```
Drulikar, line 414: Must we have all this whitespace? Just group the variables together. Its just kinda weird that almost everything in this proc is double spaced.

Drulikar (changes requested):

code/modules/cm_marines/orbital_cannon.dm
```diff
@@ -398,10 +401,37 @@ var/list/ob_type_fuel_requirements
 		return TRUE
 	return FALSE
 
+// proc designed for handling ob camera shakes, takes the target location as input and calculates camera shake based off user location.
```
Drulikar, line 404: Codedoc it (use /// not //)

code/modules/cm_marines/orbital_cannon.dm
```diff
@@ -353,6 +353,9 @@ var/list/ob_type_fuel_requirements
 /obj/structure/ob_ammo/warhead
 	name = "theoretical orbital ammo"
 	var/warhead_kind
+	var/frequency
+	var/max_shake_factor
+	var/max_knockdown_time
```
Drulikar, line 358: Could codedoc these (/// before each) though ultimately the minimum you should change is the name of `frequency` to something like `shake_frequency`

code/modules/cm_marines/orbital_cannon.dm
```diff
@@ -419,24 +449,32 @@ var/list/ob_type_fuel_requirements
 	var/datum/cause_data/cause_data = create_cause_data(initial(name), source_mob)
 	cell_explosion(target, clear_power, clear_falloff, EXPLOSION_FALLOFF_SHAPE_LINEAR, null, cause_data) //break shit around
 	sleep(clear_delay)
```
Drulikar, line 451: It seems a little weird this explosion doesn't cause a shake until after the sleep.

Drulikar (changes requested):

code/modules/cm_marines/orbital_cannon.dm
Cthulhu80: the code for the main explosion is right under and it only executes if the turf is valid. This is a much smaller initial explosion and I feel like this shouldn't warrant the camera shake unless the main larger explosion executes.

code/modules/cm_marines/orbital_cannon.dm
```diff
@@ -419,24 +449,32 @@ var/list/ob_type_fuel_requirements
...
 		return
 
+	// Checks turf around the target
 	for(var/turf/T in range(2, target))
 		if(!T.density)
 			cell_explosion(target, standard_power, standard_falloff, EXPLOSION_FALLOFF_SHAPE_LINEAR, null, cause_data)
+			handle_ob_shake(target)
 			sleep(double_explosion_delay)
 			cell_explosion(target, standard_power, standard_falloff, EXPLOSION_FALLOFF_SHAPE_LINEAR, null, cause_data)
-			return
```
Drulikar, line 434: Why this change? This will mean HE is now multiple explosions and thus a big buff to its power.

code/modules/cm_marines/orbital_cannon.dm
Drulikar: I suppose, but cell_explosion causes a sfx, and all other instances of cell_explosion here are with a ob_shake. So its weird this is the only one that has a sound, then delay, then there's a shake from a different explosion.

Just experiment with this. If you're happy with it as is, just let me know here and its resolved.

Drulikar (changes requested):

code/modules/cm_marines/orbital_cannon.dm
Drulikar: I suppose, but cell_explosion causes a sfx, and all other instances of cell_explosion here are with a ob_shake. So its weird this is the only one that has a sound, then delay, then there's a shake from a different explosion.

Just experiment with this. If you're happy with it as is, just let me know here and its resolved.

Cthulhu80: I'm happy with it.

code/modules/cm_marines/orbital_cannon.dm
Cthulhu80: That would be an accidental bug that I introduced, gonna fix that.

Drulikar (approved):
Ready for test merge (See Morrow's request here https://github.com/cmss13-devs/cmss13/pull/3605#pullrequestreview-1475161424 )

Drulikar:
Sounds like Morrow is happy with the previous test merge, so its good to go then.

PR #3669, opened 2023-06-18
Flashlight gun attachments turn off when they should

morrowwolf:
# About the pull request

This PR makes flashlight gun attachments turn off when they should.

Previously, the light source was just taken away without actually turning off the light.

Bug/inconsistency bad

<details>

</details>

# Changelog

:cl: Morrow
fix: Flashlight gun attachments turn off when they should
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_helpers.dm
+++ b/code/modules/projectiles/gun_helpers.dm
@@ -167,8 +167,9 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
 		var/obj/item/attachable/attachment = attachments[slot]
 		if (!attachment || !attachment.light_mod)
 			continue
-		bearer.SetLuminosity(0, FALSE, src)
-		SetLuminosity(attachment.light_mod)
+
+		attachment.activate_attachment(src, bearer)
+
 		return TRUE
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/gun_helpers.dm
```diff
@@ -167,8 +167,9 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
 		var/obj/item/attachable/attachment = attachments[slot]
 		if (!attachment || !attachment.light_mod)
 			continue
-		bearer.SetLuminosity(0, FALSE, src)
-		SetLuminosity(attachment.light_mod)
+
+		attachment.activate_attachment(src, bearer)
```
fira, line 171: could use `turn_off = TRUE` to short circuit turning it off if it's actually off

morrowwolf: Does this work without casting the attachment to the flashlight one? (not that that would be bad per se 🤷 )

fira: Yeahhh not all of them have it, scratch that

PR #3671, opened 2023-06-19
LZ-624 Alternative Colony Nightmare Insert

Steelpoint:
# About the pull request

Adds a new map based off of LV-624, at this stage it only changes the main colony. 

All equipment spawns, intel and akin are identical to LV-624's original colony. 

I'm putting the map up for review to see where it stands. 

Varity is the spice of life. A new colony to play around in might help invigorate veteran players and those who feel a bit tired of the same map being used so often. 

Screenshot of the SDMM file of the map. 
<details>

Screenshot 2023-06-21 00 53 10
"Casualties of war"

</details>

Updated SDMM image as of: 26/06

# Changelog
:cl:
mapadd: New nightmare colony insert for LV-624 that replaces the colony with an entirely new version. 
/:cl:


Steelpoint:
Not sure what's causing the errors.

Steelpoint:
Still working on the map. Not sure why I keep getting test errors. The test keeps stating the game is failing to delete a marker on the base LV map that I suspect is being overwritten by the nightmare insert. But if I remove the marker from the OG map it just complains about the next marker.

Steelpoint:
Maps in working order now and the test errors are fixed.

code/game/objects/effects/landmarks/landmarks.dm
```diff
@@ -216,23 +216,25 @@
 
 /obj/effect/landmark/yautja_teleport
 	name = "yautja_teleport"
+	/// The index we registered as in mainship_yautja_desc or yautja_teleport_descs
+	var/desc_index
 
 /obj/effect/landmark/yautja_teleport/Initialize(mapload, ...)
 	. = ..()
 	var/turf/T = get_turf(src)
+	desc_index = T.loc.name + T.loc_to_string()
```
Drulikar, line 225:
```suggestion
	var/turf/turf = get_turf(src)
	desc_index = turf.loc.name + turf.loc_to_string()
```
See https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#variables

Drulikar (changes requested):
So at this point is there still a need to remove the van? I'm not seeing what error you were experiencing with a van except the one you had already fixed.

Steelpoint:
Van is back. Lets see if it works.

code/game/objects/effects/landmarks/landmarks.dm
Steelpoint: Honestly, I was given the code by a proper coder who said it might fix the issue, I don't actually know what you want me to do here since I can't really understand the code.

Steelpoint: Oh, its a suggested change, ok then.

Steelpoint:
Looks like the Van still has issues with being deleted.

morrowwolf (changes requested):
This is not going to be added as a nightmare insert.

I suggest you look at making this a whole new map (even if LV-624 caves are tagged on the top).

Steelpoint:
Assuming everything works, I think this is now a standalone map.

Steelpoint:
The PR gave me some experience but I don't think this is to the standard to become a unique map.

I'll likely revisit to add the hydro+medical as a nightmare insert in the future.

PR #3677, opened 2023-06-19
Flare nerf (& Flare Acid Changes)

BeagleGaming1:
# About the pull request

Makes standard flares last 1/3 as long as before
(I hate acid code)

Flares stay there for near half an hour unless you melt them currently, this makes it so you need to use flares more often, and flares will melt faster if they have already been there for some time.

# Changelog

:cl:
add: Melting flares makes the flares burn out faster, instead of melting like normal
balance: Flares last ~10 mins, and flares in the air last ~5 mins
/:cl:


morrowwolf (commented):
Balance good

fira (approved):
finally

PR #3682, opened 2023-06-20
Fix XM88 streak end notification

sg2002:
# About the pull request
XM88 has a notification on the hit streak end, but the end user currently never sees it except for when the gun is dropped.

That's because it used visible_message and those don't work with things in hand. So, you would only see it when you drop the gun. Also I've changed it so that when the gun is dropped that message only appears when there was an actual streak going.

REEEEE?!

# Changelog
:cl: sg2002
fix: XM88 now correctly notifies the user on the hit streak end.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/lever_action.dm
+++ b/code/modules/projectiles/guns/lever_action.dm
@@ -518,7 +518,7 @@ their unique feature is that a direct hit will buff your damage and firerate
 		return
 	SIGNAL_HANDLER
 	if(streak > 0)
-		to_chat(user, SPAN_WARNING("\The [src] beeps as it loses its targeting data, and returns to normal firing procedures."))
+		to_chat(user, SPAN_WARNING("[src] beeps as it loses its targeting data, and returns to normal firing procedures."))
 	streak = 0
 	lever_sound = initial(lever_sound)
 	lever_message = initial(lever_message)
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/guns/lever_action.dm
```diff
@@ -513,10 +513,12 @@ their unique feature is that a direct hit will buff your damage and firerate
 		levered = FALSE
 	return empty_chamber(user)
 
-/obj/item/weapon/gun/lever_action/xm88/reset_hit_buff(one_hand_lever) //why does this need a user arg when it doesn't use user at all?
+/obj/item/weapon/gun/lever_action/xm88/reset_hit_buff(mob/user, one_hand_lever)
 	if(!(flags_gun_lever_action & USES_STREAKS))
 		return
 	SIGNAL_HANDLER
+	if(streak > 0)
+		to_chat(user, SPAN_WARNING("\The [src] beeps as it loses its targeting data, and returns to normal firing procedures."))
```
harryob, line 521:
```suggestion
		to_chat(user, SPAN_WARNING("[src] beeps as it loses its targeting data, and returns to normal firing procedures."))
```

harryob (changes requested):

harryob (approved):

PR #3709, opened 2023-06-24
Hardy Weeds Upgrade Weeds

Drulikar:
# About the pull request

This PR makes it so hardy weeds can upgrade weeds much like hive weeds do. They now allow special structures (hive core and hive cluster - morpher and recovery need hive weeds) since both it was weird they wouldn't (the turf is already checks anyways if it supports special structures such that you cannot build special on semi-weedable turfs or worse), and a gardener would just hurt the hive making areas where you couldn't build those special structures.

Gardeners should not be punished for using their special weeds. They grow fruit faster and have more health, so they are better than normal weeds. It makes no sense for them to not support special structures (the turf dictates that). Be careful though, you will replace nodes that are inferior (see testing gif)!

<details>

weeds

</details>

# Changelog
:cl: Drathek
balance: Gardener's hardy weeds now upgrade normal weeds (just like hive weeds upgrade weeds).
balance: Gardener's hardy weeds now don't prevent special structures (core and pylons) but they are still only allowed if the turf allows them.
/:cl:

### DIFF
--- a/code/__DEFINES/xeno.dm
+++ b/code/__DEFINES/xeno.dm
@@ -132,6 +132,7 @@
 // Weed defines
 #define WEED_LEVEL_WEAK 0
 #define WEED_LEVEL_STANDARD  1.5
+#define WEED_LEVEL_HARDY 1.6
 #define WEED_LEVEL_HIVE   4
 
 #define WEED_RANGE_STANDARD  3
--- a/code/modules/cm_aliens/weeds.dm
+++ b/code/modules/cm_aliens/weeds.dm
@@ -492,20 +492,20 @@
 	overlay_node = TRUE
 	overlays += staticnode
 
-/obj/effect/alien/weeds/node/Initialize(mapload, obj/effect/alien/weeds/node/node, mob/living/carbon/xenomorph/X, datum/hive_status/hive)
+/obj/effect/alien/weeds/node/Initialize(mapload, obj/effect/alien/weeds/node/node, mob/living/carbon/xenomorph/xeno, datum/hive_status/hive)
 	if (istype(hive))
 		linked_hive = hive
-	else if (istype(X) && X.hive)
-		linked_hive = X.hive
+	else if (istype(xeno) && xeno.hive)
+		linked_hive = xeno.hive
 	else
 		linked_hive = GLOB.hive_datum[hivenumber]
 
-	for(var/obj/effect/alien/weeds/W in loc)
-		if(W != src)
-			if(W.weed_strength > WEED_LEVEL_HIVE)
+	for(var/obj/effect/alien/weeds/weed in loc)
+		if(weed != src)
+			if(weed.weed_strength > WEED_LEVEL_HIVE)
 				qdel(src)
 				return
-			qdel(W) //replaces the previous weed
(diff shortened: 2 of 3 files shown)


Dethstorm:
The idea behind hardy weeds was to provide weeds to otherwise weedless places like you've show in the video, the bonus healing and faster growth were side features iirc. Special structures on hardy weeds were SPECIFICALLY disabled since it could lead to someone building a hive core (or similar structures) in LV jungle or other weedless places. If you allow for things like hive cores and clusters to be built on top of hardy weeds (even if this is restricted to weedable turfs), you end up removing the hardy weed node, which you then need to replant. I get where this is coming from, but I would just have special structures take over hardy weeds without being able to place these structures on hardy weeds, just to prevent any wacky shenanigans. Otherwise what you could do is make the hardy weed node breakable by hand, which would solve the issue of hardy weeds restricting placement of special structures.

Drulikar:
> The idea behind hardy weeds was to provide weeds to otherwise weedless places like you've show in the video, the bonus healing and faster growth were side features iirc. Special structures on hardy weeds were SPECIFICALLY disabled since it could lead to someone building a hive core (or similar structures) in LV jungle or other weedless places. If you allow for things like hive cores and clusters to be built on top of hardy weeds (even if this is restricted to weedable turfs), you end up removing the hardy weed node, which you then need to replant. I get where this is coming from, but I would just have special structures take over hardy weeds without being able to place these structures on hardy weeds, just to prevent any wacky shenanigans. Otherwise what you could do is make the hardy weed node breakable by hand, which would solve the issue of hardy weeds restricting placement of special structures.

As stated, the turf is what already prevents a special structure. The reason why I am changing it to allow special structures on these weeds, is because if you spread these weeds to improve the weeds in caves and such where you are allowed to place special structures, these weeds will then prevent special structures. This ability should not be used to troll or punish your hive; it should only benefit the hive.

See image:
image
Blue arrows are what I tried and where, and what error I got for trying to place that special structure on a turf that doesn't allow it (despite having the hardy weeds there). What enabling special structures does though is allow the hardy weeds to get special structures where they are allow rather than just preventing them regardless (e.g. the red checkmark).

I do not see any PRs related to the hardy/gardener weeds stating they increasing healing, nor do I see anything in the code that would facilitate this. They are just weeds that are allowed to *spread* on semi-weedable turfs, have twice the health, and grow fruit 20% faster.

This PR allows them to spread and upgrade normal weeds (like hive weeds do), and disables that special structure restriction.

If you think theres something I'm overlooking, please checkout this PR, test it, and point out anything I may have overlooked with these changes.

Dethstorm:
Yeah nahh, my bad dude, I misread some shit. I could swear bonus healing was a planned feature, but yeah it might just be the faster growth and increased health. If structure placement is dictated by the turf itself then yeah the change makes total sense. Disregard my smooth-brained comment, I read special structures on hardy weeds and jumped to conclusions.

Drulikar:
I was mistaken thinking the spikes and acid pillars were considered special structures, but then this means the only two structures are arbitrarily blocked are the hive core and hive cluster on hardy weeds. Egg morpher and recovery node require hive weeds, and both hive core and hive cluster already check the turf for fully weedable.

PR #3710, opened 2023-06-24
[s] Require affirmative input from the admin to run sdql2 verbs

Zonespace27:
"Powerful or dangerous admin verbs should have some prompt or forced runtime input to lower the attack surface once somebody finds a href exploit since topics can trigger commands and verbs."

Thanks MSO

https://github.com/tgstation/tgstation/pull/76276


fira (approved):
Sounds a bit backwards to me (why have it  a verb at all?) but i guess this works

harryob (approved):

PR #3711, opened 2023-06-24
Made filled resin holes and weed nodes destroyable by xenos

JackieEstegado:
# About the pull request

Made weed nodes destroyable with the plant weeds button.
Made filled resin holes destroyable by harm intent click, with a hive-wide message telling about it to avoid griefing going unnoticed. It does constitute a bit of a buff, due to it being usable for extinguishing yourself.
Both have a short windup delay to avoid accidental use.

If a resin hole gets filled it becomes unremovable by xenos until something triggers it, no reason for it to work like that, was just annoying if a badly placed hole got filled.

Pretty much the same for weeds, nice to be able to remove weeds when trying to be stealthy, not giving away your presence (For example the west LZ2 wall on BR when you are doing a Mining Hive and some XX drone weeded it).

Tested every possible shenanigans I could think of relating to these two things.

# Changelog
:cl:
qol: Weed nodes are now removable with the 'plant weeds' ability.
balance: Acid filled resin holes can now be destroyed by xenos via clicking on them with harm intent.
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/trap.dm
+++ b/code/modules/cm_aliens/structures/trap.dm
@@ -217,6 +217,20 @@
 			to_chat(X, SPAN_XENONOTICE("[src] is occupied by a child."))
 			return XENO_NO_DELAY_ACTION
 
+	if(X.a_intent == INTENT_HARM)
+		to_chat(X, SPAN_XENONOTICE("You start tearing away at the hole."))
+		xeno_attack_delay(X)
+		if(!do_after(X, 30, INTERRUPT_NO_NEEDHAND, BUSY_ICON_HOSTILE, src))
+			return XENO_NO_DELAY_ACTION
+		var/area/A = get_area(src)
+		if (X.hivenumber == hivenumber)
+			Destroy()
+		to_chat(X, SPAN_XENONOTICE("You destroy the trap."))
+		for(var/mob/living/carbon/xenomorph/Xeno in GLOB.living_xeno_list)
+			if(Xeno.hivenumber == hivenumber)
+				to_chat(Xeno, SPAN_XENOMINORWARNING("One of your Hive's traps at [A.name] has been torn apart by [X]!"))
+		return XENO_NO_DELAY_ACTION
+
 	if((!X.acid_level || trap_type == RESIN_TRAP_GAS) && trap_type != RESIN_TRAP_EMPTY)
 		to_chat(X, SPAN_XENONOTICE("Better not risk setting this off."))
 		return XENO_NO_DELAY_ACTION
--- a/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
@@ -27,8 +27,16 @@
 		return
 
 	var/obj/effect/alien/weeds/node/N = locate() in T
-	if(N && N.weed_strength >= X.weed_level)
-		to_chat(X, SPAN_WARNING("There's a pod here already!"))
+	if(N)
+		if(N.weed_strength > X.weed_level)
+			to_chat(X, SPAN_WARNING("There's a pod here already!"))
+		else
+			to_chat(X, SPAN_WARNING("You start removing the resin node."))
+			if(!do_after(X, 10, INTERRUPT_ALL|BEHAVIOR_IMMOBILE, BUSY_ICON_BUILD))
+				return
+			playsound(X.loc, "alien_resin_break", 25)
+			N.Destroy()
+			to_chat(X, SPAN_WARNING("You removed the resin node."))
 		return
 
 	var/obj/effect/alien/resin/trap/resin_trap = locate() in T


code/modules/cm_aliens/structures/trap.dm
```diff
@@ -217,6 +217,20 @@
 			to_chat(X, SPAN_XENONOTICE("[src] is occupied by a child."))
 			return XENO_NO_DELAY_ACTION
 
+	if(X.a_intent == INTENT_HARM)
```
Drulikar, line 220: Please fix the single letter vars as you need to modify code. See https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#variables

In VSC its very easy to do this, just ctrl+f, highlight the entire proc block, press the paragraph looking button to restrict search to selection, match case, and match word, and then dropdown to reveal replace.

code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
```diff
@@ -27,8 +27,16 @@
 		return
 
 	var/obj/effect/alien/weeds/node/N = locate() in T
-	if(N && N.weed_strength >= X.weed_level)
-		to_chat(X, SPAN_WARNING("There's a pod here already!"))
+	if(N)
```
Drulikar, line 30: Please fix the single letter vars as you need to modify code.

code/modules/cm_aliens/structures/trap.dm
```diff
@@ -217,6 +217,20 @@
...
+		to_chat(X, SPAN_XENONOTICE("You start tearing away at the hole."))
+		xeno_attack_delay(X)
```
Drulikar, line 222: I don't think there's really any point to add an attack delay before a do_after.

code/modules/cm_aliens/structures/trap.dm
```diff
@@ -217,6 +217,20 @@
...
+		if(!do_after(X, 30, INTERRUPT_NO_NEEDHAND, BUSY_ICON_HOSTILE, src))
```
Drulikar, line 223: Use the time macros. E.g. `3 SECONDS` See https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#use-our-time-defines

code/modules/cm_aliens/structures/trap.dm
```diff
@@ -217,6 +217,20 @@
...
+			return XENO_NO_DELAY_ACTION
+		var/area/A = get_area(src)
+		if (X.hivenumber == hivenumber)
+			Destroy()
+		to_chat(X, SPAN_XENONOTICE("You destroy the trap."))
+		for(var/mob/living/carbon/xenomorph/Xeno in GLOB.living_xeno_list)
+			if(Xeno.hivenumber == hivenumber)
+				to_chat(Xeno, SPAN_XENOMINORWARNING("One of your Hive's traps at [A.name] has been torn apart by [X]!"))
```
Drulikar, line 231: To match the wording of `/obj/effect/alien/resin/trap/proc/trigger_trap`:
```suggestion
				to_chat(Xeno, SPAN_XENOMINORWARNING("You sense one of your Hive's traps at [A.name] has been torn apart by [X]!"))
```

code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
```diff
@@ -27,8 +27,16 @@
...
+		if(N.weed_strength > X.weed_level)
+			to_chat(X, SPAN_WARNING("There's a pod here already!"))
+		else
+			to_chat(X, SPAN_WARNING("You start removing the resin node."))
+			if(!do_after(X, 10, INTERRUPT_ALL|BEHAVIOR_IMMOBILE, BUSY_ICON_BUILD))
+				return
+			playsound(X.loc, "alien_resin_break", 25)
+			N.Destroy()
```
Drulikar, line 38: Use qdel instead of directly calling what qdel calls.

code/modules/cm_aliens/structures/trap.dm
```diff
@@ -217,6 +217,20 @@
...
 			return XENO_NO_DELAY_ACTION
 
+	if(X.a_intent == INTENT_HARM)
+		to_chat(X, SPAN_XENONOTICE("You start tearing away at the hole."))
+		xeno_attack_delay(X)
+		if(!do_after(X, 30, INTERRUPT_NO_NEEDHAND, BUSY_ICON_HOSTILE, src))
+			return XENO_NO_DELAY_ACTION
+		var/area/A = get_area(src)
+		if (X.hivenumber == hivenumber)
+			Destroy()
```
Drulikar, line 227: Use qdel instead of directly calling what qdel calls.

code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
```diff
@@ -27,8 +27,16 @@
...
 
 	var/obj/effect/alien/weeds/node/N = locate() in T
-	if(N && N.weed_strength >= X.weed_level)
-		to_chat(X, SPAN_WARNING("There's a pod here already!"))
+	if(N)
+		if(N.weed_strength > X.weed_level)
+			to_chat(X, SPAN_WARNING("There's a pod here already!"))
+		else
+			to_chat(X, SPAN_WARNING("You start removing the resin node."))
+			if(!do_after(X, 10, INTERRUPT_ALL|BEHAVIOR_IMMOBILE, BUSY_ICON_BUILD))
```
Drulikar, line 35: Use the time macros. E.g. `1 SECONDS` See https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#use-our-time-defines

Drulikar (changes requested):
I'm not really keen on this, but I'm hardly the last say.

However, I would like this to be tied in with a queen setting that a queen can opt-in to having much like the existing construction_toggle and destruction_toggle options have levels for no-one, leaders, and everyone. Preferably where the setting is off by default (existing behavior).

code/modules/cm_aliens/structures/trap.dm
Drulikar: E.g. 
image

JackieEstegado: Have the queen option for both of them, in separate toggles?

Drulikar: Search for one of those options and you'll see what I mean. It's a single setting with 3 possible states.

PR #3717, opened 2023-06-24
Fix Deterred Crashsite Offset

### LINKED ISSUE #3593: Dropship has a chance to land in AA protected areas.
### Testmerges

#3470, #3552, #3298, #3588, #3576

### Description of the bug

The dropship has a chance to land in the AA protected areas even when they have been deterred the first time.

Well, it shouldn't. It should land somewhere else.

1. Set AA to a protected area.
2. Crash land the dropship to said protected area.
3. AA delays the crash but does not deter it from landing there
4. Dropship lands at AA protected area.

Observed twice on the live server.

Broken

### Issue Bingo

Drulikar:
# About the pull request

This PR fixes an oversight where a deterred crashsite does not get positioned the same as a non-deterred one.

Fixes #3593 or atleast should much better mitigate it making it so a deterred crashsite positions the same as a non-deterred.

<details>

</details>

# Changelog
:cl: Drathek
fix: Fixed the crashsite offset for a hijack shuttle that gets deterred by the MGAD System
/:cl:

### DIFF
--- a/code/modules/shuttle/dropship_hijack.dm
+++ b/code/modules/shuttle/dropship_hijack.dm
@@ -1,3 +1,5 @@
+#define HIJACK_CRASH_SITE_OFFSET_X -5
+#define HIJACK_CRASH_SITE_OFFSET_Y -11
 
 /datum/dropship_hijack
 	var/obj/docking_port/mobile/shuttle
@@ -90,8 +92,8 @@
 
 	var/obj/docking_port/stationary/marine_dropship/crash_site/target_site = new()
 	crash_site = target_site
-	crash_site.x = target.x - 5
-	crash_site.y = target.y - 11
+	crash_site.x = target.x + HIJACK_CRASH_SITE_OFFSET_X
+	crash_site.y = target.y + HIJACK_CRASH_SITE_OFFSET_Y
 	crash_site.z = target.z
 
 	target_site.name = "[shuttle] crash site"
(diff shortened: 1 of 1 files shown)


code/modules/shuttle/dropship_hijack.dm
```diff
@@ -116,8 +118,10 @@
 		remaining_crash_sites -= target_ship_section
 		var/new_target_ship_section = pick(remaining_crash_sites)
 		var/turf/target = get_crashsite_turf(new_target_ship_section)
-		crash_site.Move(target)
 		marine_announcement("A hostile aircraft on course for the [target_ship_section] has been successfully deterred.", "IX-50 MGAD System")
+		crash_site.x = target.x + HIJACK_CRASH_SITE_OFFSET_X
+		crash_site.y = target.y + HIJACK_CRASH_SITE_OFFSET_Y
+		crash_site.z = target.z
```
fira, line 124: there shouldn't be any direct loc / coordinates setting basically ever, use Move() or forceMove()

for the above one it can even just be passed as constructor argument

fira (changes requested):

code/modules/shuttle/dropship_hijack.dm
Drulikar: No real point to set up a constructor if the crash site is attempted to be reused so then we need different code for that case. Should just be the same method for both procs.

fira (approved):

PR #3725, opened 2023-06-25
Removes CL from ID Computer & updates "centcom"

realforest2001:
# About the pull request
People were using the CL's preset in ID computer to gain access to CLs office when this shouldn't really be possible. This stops that, and also makes it so switching the computer to WY mode actually does something.

<details>

</details>

# Changelog
:cl:
add: Added proper assignments for wey-yu mode on ID computers.
del: Removed CL from ID Computer unless on WY mode.
/:cl:


Drulikar (changes requested):
This all appears working *except* that the access tab (the first one) is not altered when in wey_yu mode. Namely the grant all/deny all buttons behave significantly differently in wey_yu mode but the individual regions displayed are still the marine ones.

I don't know exactly what regions *should* be displayed other than the default, but the grant all access behavior is different than the individual regions when grant all should just be a select all of the individual options button.

In marine mode, grant all grants 34 items, but selecting every entry individually is only 31 items (you can get this up to 35 items if you grant all and then also grant region on Almayer Command because the Commander's quarters doesn't work if you click it - only via region).

In wey yu mode, grant all grants 6 items, none of which are the 32-33 you can individually select.

If you try to max access via both modes you can get up to 41 entries.

By default when spawned via create_humans, a CL has this 13 items in their access list: 
image

PR #3726, opened 2023-06-26
New facehugger strain: Watcher

Hopekz:
Super easy to understand:

The watcher loses the ability to hide in exchange for letting them see further.

I don't have a sprite for this at the moment and probably won't for this PR so I have made them stand out by making them more orange because I think marines being able to see if a hugger is a Watcher might be important.
image

Currently they see 2 extra tiles like the runner but in testmerging I think it may change to 3 tiles (view range 10 instead of view range 9) unless I hit the perfect sweet spot already. **EDIT: This number is now 3**

**the way I see it watcher is designed for 2 main playstyles:**

1: Backlining to try to get someone not paying attention or busy (right behind marine frontline like medics doing medic stuff or someone resupplying/playing with inventory) or deep in the backline to try to find someone vulnerable. 

2: Hanging around the xeno side frontline with increased situational awareness to be exactly where you need to be for xeno assisted caps. For example as a watcher during the TM I have an easier time watching a warrior fling someone and capitalizing a hug off a fling because I don't have to go out until the fling happens etc

update:
------
The activation of the view range now works similar to the runner; it is toggle-able
dreamseeker_tyE3nVA9V0

Fixed exploit that allowed watchers to hide despite not having the ability.

Update 2:
--------
The watcher now does not take weeds over time. 
Essentially implemented this:

Discord_9s19Vo2olV

**Why:** 
The damage over time is meant so that the hugger cannot hide under some unweeded item the whole round to hug some random unsuspecting marine. The watcher cannot hide and is meant to be a backlining caste.
This allows the watcher to stalk people outside of weeds and wait for the perfect opportunity to hug when their target is distracted or busy.

Update 3:
--------
Added my codersprite to the watcher.
These were created by me so hopefully a spriter or my spriter replaces these with a better one if they suck.

dreamseeker_O6I1iXuo21

:cl: Hopek
add: New facehugger strain:  Watcher. The Watcher loses their ability to hide in exchange to see further and the ability to no longer take damage outside of weeds.
/:cl:


code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
```diff
@@ -240,3 +240,31 @@
...
+
+/datum/xeno_mutator/watcher/apply_mutator(datum/mutator_set/individual_mutators/mutator_set)
+	. = ..()
+	if(!.)
+		return
+
+	var/mob/living/carbon/xenomorph/facehugger/facehugger = mutator_set.xeno
+
+	facehugger.client.change_view(10, src)
+	facehugger.color = "#FFAD33" // This is here to make them look distinct until a coder decides to replace my coder solution with a unique sprite.
```
Hopekz, line 266:
```suggestion
	facehugger.color = "#FFAD33" // This is here to make them look distinct until a spriter decides to replace my coder solution with a unique sprite.
```

morrowwolf:
Honestly, just keep the same sprite. That orange does not look good.

code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
```diff
@@ -240,3 +240,31 @@
...
+
+/datum/xeno_mutator/watcher/apply_mutator(datum/mutator_set/individual_mutators/mutator_set)
+	. = ..()
+	if(!.)
+		return
+
+	var/mob/living/carbon/xenomorph/facehugger/facehugger = mutator_set.xeno
+
+	facehugger.client.change_view(9, src)
+	facehugger.color = "#FFAD33" // This is here to make them look distinct until a spriter decides to replace my coder solution with a unique sprite.
```
Hopekz, line 266:
```suggestion
```

sleepynecrons (changes requested):
morrow already mentioned the orange and that got changed but please avoid just doing recolors in the future. it still needs a unique sprite though; something that changes up the silhouette and distinguishes itself from a normal facehugger at a glance.

maybe since they can't hide you could make them bigger and give them longer legs and some sunken areas for "eyes".

morrowwolf (commented):
Conceptually this is approved. I believe the author wants some time to TM and play with how the zoom works which is fine. We are also waiting on a new sprite.

While waiting on the sprite feel free to ask for TM if you can convince a maintainer to keep an eye on it.

Angoryskaven:
Found two bugs:

-1: Upon choosing the Watcher strain, if you are have toggled the "Hide" ability, it will carry over onto Watcher, allowing you to hide, at least for the first jump;

-2: When reconnecting, the Watcher's sight range is reverted to that of a normal hugger

Hopekz:
update:
The activation of the view range now works similar to the runner; it is toggle-able
image

Hopekz:
> Found two bugs:
> 
> -1: Upon choosing the Watcher strain, if you are have toggled the "Hide" ability, it will carry over onto Watcher, allowing you to hide, at least for the first jump;
> 
> -2: When reconnecting, the Watcher's sight range is reverted to that of a normal hugger

1: Fixed

2: Fixed; it is now a toggle with a button like the runner so they can just toggle it again when reconnected.

Hopekz:
Update 2:
--------
The watcher now does not take weeds over time. 
Essentially implemented this:

Discord_9s19Vo2olV

**Why:** 
The damage over time is meant so that the hugger cannot hide under some unweeded item the whole round to hug some random unsuspecting marine. The watcher cannot hide and is meant to be a backlining caste.
This allows the watcher to stalk people outside of weeds and wait for the perfect opportunity to hug when their target is distracted or busy.

Hopekz:
Update 3:
--------
Added my codersprite to the watcher.
These were created by me so hopefully a spriter or my spriter replaces these with a better one.

dreamseeker_O6I1iXuo21

Hopekz:
Might reopen this in the future when CM dev picks back up closer to how it was when I started. (turnaround back to 2-4 days)

Personally for me it is a bit mentally exhausting at the moment to be open to modify something for a month so assuming I haven't moved on to another project I'll reopen this when I come back to CM dev.

harryob (changes requested):
resolve conflicts

code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
```diff
@@ -67,12 +67,11 @@
...
 
 /mob/living/carbon/xenomorph/facehugger/Life(delta_time)
-	if(!client && !aghosted && away_timer > XENO_FACEHUGGER_LEAVE_TIMER)
+	if(stat != DEAD && !lying && !(mutation_type == FACEHUGGER_WATCHER) && !(locate(/obj/effect/alien/weeds) in get_turf(src)))
 		// Become a npc once again
 		new /obj/item/clothing/mask/facehugger(loc, hivenumber)
 		qdel(src)
 		return
-	if(stat != DEAD && !lying && !(locate(/obj/effect/alien/weeds) in get_turf(src)))
 		adjustBruteLoss(1)
```
harryob, line 75: this is after a return so will never be called

harryob (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
```diff
@@ -217,3 +215,35 @@
...
+
+/datum/xeno_mutator/watcher/apply_mutator(datum/mutator_set/individual_mutators/mutator_set)
+	. = ..()
+	if(!.)
+		return
+
+	var/mob/living/carbon/xenomorph/facehugger/facehugger = mutator_set.xeno
+
+	facehugger.viewsize = 10
+	facehugger.layer = 4 // Mob layer is 4 but we don't have a define for it? It's commented out in the layers.dm file
```
harryob, line 245: that's because MOB_LAYER is a dm define, you can just use it

code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
```diff
@@ -67,12 +67,10 @@
...
 
 /mob/living/carbon/xenomorph/facehugger/Life(delta_time)
-	if(!client && !aghosted && away_timer > XENO_FACEHUGGER_LEAVE_TIMER)
+	if(stat != DEAD && !lying && !(mutation_type == FACEHUGGER_WATCHER) && !(locate(/obj/effect/alien/weeds) in get_turf(src)))
 		// Become a npc once again
 		new /obj/item/clothing/mask/facehugger(loc, hivenumber)
 		qdel(src)
-		return
-	if(stat != DEAD && !lying && !(locate(/obj/effect/alien/weeds) in get_turf(src)))
 		adjustBruteLoss(1)
```
harryob, line 74: why are we adjusting brute loss after qdel(src)

harryob (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
Hopekz: alright now I 100% fixed the conflict

code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
Hopekz: done

Hopekz:
From what I saw with TM 2 with how people play watcher and how it felt to play I conclude on my end that it is good for merge.

Actually seen watchers do the "stalk people until they're distracted or busy with something" playstyle or assisting other backliner casts in capping as intended.

code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
```diff
@@ -67,13 +67,13 @@
 		PF.flags_can_pass_all = PASS_ALL^PASS_OVER_THROW_ITEM
 
 /mob/living/carbon/xenomorph/facehugger/Life(delta_time)
+	if(stat != DEAD && !lying && !(mutation_type == FACEHUGGER_WATCHER) && !(locate(/obj/effect/alien/weeds) in get_turf(src)))
+		adjustBruteLoss(1)
+		return
```
realforest2001, line 72: This has stopped facehuggers losing health at all when off weeds from what I can see.

realforest2001 (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
Hopekz: Going to check if this is the case but the "!(mutation_type == FACEHUGGER_WATCHER)" part of the if statement should only leave the watcher alone as intended because it should only pass these through if they are NOT the watcher.

Hopekz: Yeah that's what it is. So it isn't that you don't take damage it's that the status bug doesn't update your damage until your status changes vs regularly 

See this GIF as it shows how it updates the damage all at once a few seconds after you get back on weeds.

dreamseeker_J1MEMbb6vs

ihatethisengine: I got reports in dchat so it's not just status bar bug. You can run without weeds for infinite amount of time, then rest and explode. So you did broke something.

ihatethisengine: Probably because you added `return` instead of `return ..()`

ihatethisengine: https://github.com/cmss13-devs/cmss13/issues/4062

ihatethisengine, line 72:
```suggestion
		return ..()
```

ihatethisengine: You basically don't call parent because of this return so everything is broken.

ihatethisengine:
oops sorry for creating so many reviews

code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
Hopekz: Fixed; good catch @ihatethisengine

harryob (approved):
tm fine, needs sprite approval

sleepynecrons (changes requested):
the current sprite is an improvement but it's just a cut and paste of the xenoid hugger head with no attempt at blending it into the watcher sprite. please fix. it also still needs a change in silhouette; maybe do something with the feelers/legs at the front considering they don't move in the running animation. 

you could also do something with the tail but it will obviously require some animation work

example
image

Hopekz:
Went ahead and removed the watcher sprite entirely because it didn't pass sprite review and I couldn't get it replaced. 
Shame but it is what it is.

Thank you!

PR #3732, opened 2023-06-26
Dropship Missile Refactor

blackdragonTOW:
# About the pull request
Refactors Dropship Missiles to not use Spawn().

While proposing a new CAS feature, this part of the code uses a function we shouldn't be using.  With a suggestion from Harry I simply swapped out the parts that we could replace without having to do a bigger overhaul.
<details>

</details>

# Changelog
:cl:
refactor: refactored dropship_ammo so that missiles don't use Spawn()
/:cl:

### DIFF
--- a/code/modules/cm_marines/dropship_ammo.dm
+++ b/code/modules/cm_marines/dropship_ammo.dm
@@ -287,7 +287,7 @@
 
 /obj/structure/ship_ammo/rocket/widowmaker/detonate_on(turf/impact)
 	impact.ceiling_debris_check(3)
-	addtimer(CALLBACK(GLOBAL_PROC, GLOBAL_PROC_REF(cell_explosion), impact, 300, 40, EXPLOSION_FALLOFF_SHAPE_LINEAR, null, create_cause_data(initial(name), source_mob), 0.5 SECONDS) //Your standard HE splash damage rocket. Good damage, good range, good speed, it's an all rounder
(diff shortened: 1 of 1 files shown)


code/modules/cm_marines/dropship_ammo.dm
```diff
@@ -317,9 +315,8 @@
 
 /obj/structure/ship_ammo/rocket/keeper/detonate_on(turf/impact)
 	impact.ceiling_debris_check(3)
-	spawn(5)
-		cell_explosion(impact, 450, 100, EXPLOSION_FALLOFF_SHAPE_EXPONENTIAL, null, create_cause_data(initial(name), source_mob)) //Insane fall off combined with insane damage makes the Keeper useful for single targets, but very bad against multiple.
-		qdel(src)
+	addtimer(CALLBACK(GLOBAL_PROC, GLOBAL_PROC_REF(cell_explosion), impact, 450, 100, EXPLOSION_FALLOFF_SHAPE_EXPONENTIAL, null, create_cause_data(initial(name)), source_mob), 0.5 SECONDS)) //Insane fall off combined with insane damage makes the Keeper useful for single targets, but very bad against multiple.
```
BeagleGaming1, line 318:
```suggestion
	addtimer(CALLBACK(GLOBAL_PROC, GLOBAL_PROC_REF(cell_explosion), impact, 450, 100, EXPLOSION_FALLOFF_SHAPE_EXPONENTIAL, null, create_cause_data(initial(name)), source_mob), 0.5 SECONDS) //Insane fall off combined with insane damage makes the Keeper useful for single targets, but very bad against multiple.
```

blackdragonTOW: Yep, I'm dumb, I miss counted and then added the closer in the wrong place. Hurr Durr.

fira:
uh, wouldn't the fire also need to be delayed?

BeagleGaming1 (approved):
code seems good

PR #3733, opened 2023-06-27
XM88 Spotter Variant for spotter

Hopekz:
**What is it?:**
The spotter gets a single-round variant of the XM88 rifle meaning that the gun itself stores 1 round inside the magazine.
dreamseeker_vJ984pio4G
___
**What makes this special?**
This rifle can fire never before seen tracking rounds.
The rifle itself has a unremovable scope that has a 4x IFF zoom similar to the sniper.
There are only 40 rounds of tracking rounds available (1 full ammo belt)

___
"**1 shot XM rifle? That sounds terrible!**"
The point of this gun is to use the tracking bullets to spot targets for the sniper. Any target hit with the tracking bullets will be spotted for 20 seconds (may change to 15 after a TM, not 100% sure). 
This is essentially a more "fun" and "involved" way for spotters to do their job for the sniper while giving them something fun to do (shooting a rifle). 
One thing this allows a spotter to do versus just using their scope is marking multiple targets at once so the sniper gets to decide what they feel like shooting at. They can even mix and match their playstyle like for example shoot 1 alien with the rifle then whip out their binos to mark another target so they'd have 2 targets marked using 2 different methods.

The regular XM88 can't hold these new tracker rounds and if they can through some means that I haven't discovered then it is a bug.

____
"**Is it overpowered!?**"

Here is what needs to happen after the rifle is fired:
1: You need to free up a hand to grab another bullet
2: Grabbing a bullet causes you to hold the gun in 1 hand, this causes the gun to unscope
3: Put the bullet in the gun
4: Chamber the round (can do after step 6)
5: hold the rifle in both hands again
6: Scope in the rifle again

In practice I think I have created a solution for this that is fairly balanced but a TM will verify it.

_____
**Flavortext lore description:**

The distinctiveness of the S6 Solo Scope's fit to the XM88 Scout Rifle springs from the deep-rooted secrets of ARMAT's engineering. The XM88, under its metallic skin, hides an exclusive, proprietary optics mount, a feat of craftsmanship from ARMAT. This bespoke design meshes flawlessly with the Solo Scope, creating a harmonious union that is the mark of an expert rifleman. Upon first glance, you might think it's just a unique mount, but it's more than that. The optics mount is engineered with an advanced mechanical lock-in mechanism, fortified with a responsive electronic interlink. When the S6 Solo Scope is placed onto the XM88, the lock-in mechanism secures the scope firmly, ensuring zero shift during recoil. Then, the electronic interlink takes over - the microprocessors within the XM88 and the Solo Scope commence a rapid series of high-speed data transfers, fine-tuning the scope's range settings to match the current atmospheric conditions, as detected by the XM88's internal barometric and temperature sensors.

This ingenious pairing of mechanical design and electronic interfacing enables the rifleman to unlock the full lethal precision the XM88 offers. And it's in this masterful interplay between steel and silicon, the convergence of mechanics and electronics, that the XM88 and the S6 Solo Scope truly become a force to be reckoned with in the right hands.

The sophisticated technology underlying the Solo Scope's functionality is best matched to a single-shot rifle like the XM88, and there's a crucial reason for that. Automatic or semi-automatic rifles may seem more powerful due to their rapid-fire capabilities, but this often comes with an unavoidable trade-off: accuracy. With each successive shot, recoil accumulates, leading to a shift in the rifle's position which can throw off the aim considerably, rendering even the best scopes less effective.

However, the XM88's single-shot mechanism sidesteps this issue elegantly. After each shot, the shooter manually reloads, which provides a moment for the rifle to return to its original position, minimizing the impact of recoil on aim.

Further, the Solo Scope is integrated with a Recoil Compensation System (RCS). This sophisticated feature uses high-speed gyroscopes to detect the minute shifts from the rifle's recoil. The RCS then rapidly calculates and adjusts the scope's reticle, compensating for the minute change in angle. Yet, such a system requires a brief moment to reset and recalibrate after each shot, a time luxury that isn't available in rapid-fire firearms.

The XM88, with its single-shot operation, offers just enough pause for the RCS to perform its recalibration, ensuring maximum accuracy for the next shot. This perfect harmony between the rifle and scope, built on the back of ARMAT's precision engineering, makes the XM88 and the S6 Solo Scope an unmatched combination in terms of accuracy and precision.
"

:cl: Hopek
add: Added the XM88 Spotter Variant for spotter. This gun fires new tracking bullets that spots individuals similar to using the spotting binoculars.
/:cl:


code/modules/projectiles/ammo_datums.dm
```diff
@@ -1443,15 +1443,25 @@
...
-	name = "tracking lever-action bullet"
+	name = ".458 tracker round"
 	icon_state = "redbullet"
 	damage = 70
 	penetration = ARMOR_PENETRATION_TIER_3
 	accuracy = HIT_ACCURACY_TIER_1
 	handful_state = "tracking_lever_action_bullet"
 
+/datum/ammo/bullet/lever_action/tracker/proc/tracker_remove(mob/M, icon)
+	REMOVE_TRAIT(M, TRAIT_SPOTTER_LAZED, TRAIT_SOURCE_EQUIPMENT(2000))
```
fira, line 1454: What's 2000 here?

code/modules/projectiles/magazines/lever_action.dm
```diff
@@ -74,6 +74,28 @@ Similar to shotguns.dm but not exactly.
...
+	caliber = ".458 tracker"
+	gun_type = /obj/item/weapon/gun/lever_action/xm88/spotter
+	max_rounds = 1
+
+/obj/item/ammo_magazine/internal/lever_action/xm88/spotter/Initialize(mapload, spawn_empty)
+	. = ..()
+	src.color = "#FFCC00" // Yes recoloring things is the new coder sprites what about it
+
+// This si the box that comes in the spotter kit
+/obj/item/ammo_magazine/internal/lever_action/xm88/spotter/box
```
fira, line 94: Don't use internal magazines for this, these are meant for internal firearms usage (holding rounds as a revolver barrel or shotgun tube, etc) and not as usable items

code/modules/projectiles/magazines/lever_action.dm
```diff
@@ -74,6 +74,28 @@ Similar to shotguns.dm but not exactly.
...
+	icon_state = "458-box"
+	item_state = "458-box"
+	default_ammo = /datum/ammo/bullet/lever_action/tracker
+	caliber = ".458 tracker"
+	gun_type = /obj/item/weapon/gun/lever_action/xm88/spotter
+	max_rounds = 1
+
+/obj/item/ammo_magazine/internal/lever_action/xm88/spotter/Initialize(mapload, spawn_empty)
+	. = ..()
+	src.color = "#FFCC00" // Yes recoloring things is the new coder sprites what about it
```
fira, line 91: This one shouldn't be visible so why change the color?

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1443,15 +1443,25 @@
 //unused and not working. need to refactor MD code. Unobtainable.
 //intended mechanic is to have xenos hit with it show up very frequently on any MDs around
 /datum/ammo/bullet/lever_action/tracker
-	name = "tracking lever-action bullet"
+	name = ".458 tracker round"
```
fira, line 1446: If you want to change that you'll also need to change the ammunition and boxes etc to match, eg. `/obj/item/ammo_magazine/lever_action/tracker` which claims to be "45-70". Yes i know this is almost the same caliber but consistency is important

fira (changes requested):

fira:
As an extra observation, i'm not sure 40 bullets is actually that much if it's non refillable, but that's more for balance people to figure out

code/modules/projectiles/ammo_datums.dm
Hopekz: It's the CAS tracker for the bullet.

The way I see it I had 4 solutions for CAS ID for this PR:
1: Give every single bullet a CAS ID round-start which would increment the CAS ID number by +40
2: Give a incremented CAS ID every time a tracking bullet fires
3: Set the CAS ID to some really high number that I don't think can be reached in-game like 2000
4: Rework the gun systems to support having a CAS ID linked to the gun itself for my one-off item.

I chose number 3 so I set the CAS ID to 2000 because I don't think we'll ever get to 2000 laser binos and rangefinders on the map.

code/modules/projectiles/magazines/lever_action.dm
Hopekz: Funny enough the reason that I put that in there is intentional if I remove the internal it causes the box to spawn fine but the rounds inside the box will be the SOCOM rounds not the tracer rounds.

Maybe some kind of bug with how the parents over-write ammo box contents but the way I solved this was to change the parent before we get to lever_action and "internal" works for my needs.

dreamseeker_7hcTAkPHf7

code/modules/projectiles/magazines/lever_action.dm
Hopekz: This is just a way to recolor the child one step above. I can change this if really needed and I don't see this making any impact.

code/modules/projectiles/ammo_datums.dm
Hopekz: This is actually a typo by me because the regular .458 rounds don't fit in the spotter rifle so I don't want to give users the idea that it will because they're called the same caliber.

Hopekz:
> As an extra observation, i'm not sure 40 bullets is actually that much if it's non refillable, but that's more for balance people to figure out

I had no idea how to test that part for balance alone so I only set it to 40 because that is what the standard ammo belt holds.

The way I figure that 40 may be a good number is because this is a gun and people miss guns but you're right in practice maybe 40 will need to change.

code/modules/projectiles/guns/lever_action.dm
```diff
@@ -541,3 +541,35 @@ their unique feature is that a direct hit will buff your damage and firerate
 #undef FLOATING_PENETRATION_TIER_2
 #undef FLOATING_PENETRATION_TIER_3
 #undef FLOATING_PENETRATION_TIER_4
+
+
+/obj/item/weapon/gun/lever_action/xm88/spotter
+	name = "XM88 Spotter Variant"
```
Zonespace27, line 547:
```suggestion
	name = "\improper XM88 Spotter Variant"
```
Sounds clunky as a gun name

Zonespace27 (changes requested):

code/modules/projectiles/magazines/lever_action.dm
Drulikar: See `/obj/item/ammo_magazine/lever_action/tracker`. If you inherit `/obj/item/ammo_magazine/lever_action/xm88` then of course its going to contain `/datum/ammo/bullet/lever_action/xm88` unless you alter `default_ammo`

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1443,16 +1443,26 @@
...
 	damage = 70
 	penetration = ARMOR_PENETRATION_TIER_3
 	accuracy = HIT_ACCURACY_TIER_1
 	handful_state = "tracking_lever_action_bullet"
 
-/datum/ammo/bullet/lever_action/tracker/on_hit_mob(mob/M, obj/item/projectile/P, mob/user)
-	//SEND_SIGNAL(user, COMSIG_BULLET_TRACKING, user, M)
-	M.visible_message(SPAN_DANGER("You hear a faint beep under [M]'s [M.mob_size > MOB_SIZE_HUMAN ? "chitin" : "skin"]."))
+/datum/ammo/bullet/lever_action/tracker/proc/tracker_remove(mob/target, icon)
+	REMOVE_TRAIT(target, TRAIT_SPOTTER_LAZED, TRAIT_SOURCE_EQUIPMENT(2000))
```
morrowwolf, line 1454: Mirror'ing what Fira said, having this just set at 2000 looks hyper cursed.

morrowwolf (commented):
Willing to TM but this is going to need some changes.

40 rounds isn't anything in a long round. Likely need a way to reload of some sort.

And do what Fira asked

Hopekz:
Might reopen this in the future when CM dev picks back up closer to how it was when I started. (turnaround back to 2-4 days)

Personally for me it is a bit mentally exhausting at the moment to be open to modify something for a month so assuming I haven't moved on to another project I'll reopen this when I come back to CM dev.

Probably wouldn't be as bad if the last 5+ of my PR's here haven't been just reviewed out to change the spelling of some variable or proc or how I write out words like me typing out T instead of turf. Waiting past the usual turnaround I'm used to just to be told that is discouraging.

Even in this specific PR what seems to be blocking a TM is a number looking weird at 2000 so I'd have to try to make it look more pretty just to see if people like the idea of the feature for a TM in probably another week or two after I modify the PR.

PR #3740, opened 2023-06-27
Fix Tableflip scanning logic + infinite loop

fira:
# About the pull request

Infinite loop observed live.

This fixes three things with the table flipping:

 * The table scan at top shouldn't be in the loop, since it's always done from the starting `loc`
 * The table scan reuses `table`, meaning that the second part would essentially "follow turns" through already-flipped tables incorrectly. This is both incorrect and would cause the infinite loop.
 * The table scan is now capped at 8 tables in either direction. This should mean you can't flip a 12 long table from edge, but can from the middle. Unintended feature, but it makes sense...

Flipping 5 wide flat tables still works.

# Changelog
:cl:
fix: Fixed an issue with table flips that could make some tables incorrectly unflippable, and cause infinite loops. It is no longer possible to flip tables more than 8 long in either direction from flipping point. 
/:cl:

### DIFF
--- a/code/game/objects/structures/tables_racks.dm
+++ b/code/game/objects/structures/tables_racks.dm
@@ -326,12 +326,16 @@
 /// Checks whether a table is a straight line along a given axis
 /obj/structure/surface/table/proc/straight_table_check(direction)
 	var/obj/structure/surface/table/table = src
-	while(table)
-		// Check whether there are connected tables perpendicular to the axis
-		for(var/angle in list(-90, 90))
-			table = locate() in get_step(loc, turn(direction, angle))
-			if(table && !table.flipped)
-				return FALSE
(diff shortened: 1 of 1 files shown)


Drulikar (approved):
I was able to reproduce the examples you mentioned, and this PR does resolve them. Though its still kinda weird that crossed/tshape tables are allowed to be flipped since they visually look connected (usually) and theres no way you could realistically flip those.

Drulikar:
You can try this adaption to deal with disallowing L or T shaped tables: 
[patch.diff.txt](https://github.com/cmss13-devs/cmss13/files/11885441/patch.diff.txt)

fira:
Yeah i'll add it in, this should work

PR #3747, opened 2023-06-27
Fixes being unable to re-enter your body when you're unnested

TheGamerdk:
# About the pull request

Turns out the parent call of unbuckle() calls afterbuckle() which deletes the nest and clears ghost_of_buckled_mob, which means it can't set your ghost to can_reenter_corpse = TRUE

Bugs bad

<details>

</details>

# Changelog
:cl:
fix: You can now re-enter your body when unnested, marine mains, rejoice!
/:cl:

### DIFF
--- a/code/game/objects/structures/stool_bed_chair_nest/xeno_nest.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/xeno_nest.dm
@@ -289,11 +289,12 @@
 	if(buckled_human.stat == DEAD )
 		buckled_mob_density = FALSE
 
-	. = ..()
-
 	var/mob/dead/observer/G = ghost_of_buckled_mob
 	var/datum/mind/M = G?.mind
 	ghost_of_buckled_mob = null
+
+	. = ..() //Very important that this comes after, since it deletes the nest and clears ghost_of_buckled_mob
+
 	if(!istype(buckled_human) || !istype(G) || !istype(M) || buckled_human.undefibbable || buckled_human.mind || M.original != buckled_human || buckled_human.chestburst)
 		return // Zealous checking as most is handled by ghost code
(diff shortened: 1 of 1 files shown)


Drulikar (changes requested):
Code works, though please also consider changing the ghost prompt in `/mob/living/proc/do_ghost()` when you remain in body for the first prompt. (As in get nested > remain > ghost)
It looks like this currently:
image
And if you changed it to say `text_prompt += "\nSince you're nested, you will get a chance to reenter your body if freed."` it would look like this:
image

TheGamerdk:
> Code works, though please also consider changing the ghost prompt in `/mob/living/proc/do_ghost()` when you remain in body for the first prompt. (As in get nested > remain > ghost)
> It looks like this currently:
> image
> And if you changed it to say `text_prompt += "\nSince you're nested, you will get a chance to reenter your body if freed."` it would look like this:
> image

That’s a bug with the tgui prompt. It should scale according to contents. Out of scope

Drulikar:
You are altering the action of ghosting when nested. That is the prompt given to users when doing that action; so it is not out of scope.

Sure updating tgui_alert to adapt to the size isn't very related; but thats not what I requested. The message is needlessly wordy regardless.

TheGamerdk:
> You are altering the action of ghosting when nested. That is the prompt given to users when doing that action; so it is not out of scope.
> 
> Sure updating tgui_alert to adapt to the size isn't very related; but thats not what I requested. The message is needlessly wordy regardless.

I’m not altering anything about ghosting. I’m only touching nest unbuckling

harryob (approved):

Drulikar:
I'll take the GBP for the requested change in a separate PR then.

PR #3760, opened 2023-06-28
People with lisp now lisp on radio

Ben10083:
# About the pull request

The lisp trait did not affect a flag for speech, making it so lisping does not occur over radio, this should make it so what they say is transmitted properly.

Fixes an oversight and allows for lisp speech to be handled by the handle_speech_problems method. Also realism or something

<details>

image
Poor marine trying to alert command
</details>

# Changelog
:cl:
fix: lisps now occur over radio as well
refactor: Relevant variables relating to speech problems now use true and false instead of 1 and 0
/:cl:


code/modules/character_traits/biology_traits.dm
```diff
@@ -59,11 +59,13 @@
 		return
 
 	ADD_TRAIT(target, TRAIT_LISPING, TRAIT_SOURCE_QUIRK)
+	target.speech_problem_flag = 1
```
harryob, line 62:
```suggestion
	target.speech_problem_flag = TRUE
```

code/modules/character_traits/biology_traits.dm
```diff
@@ -59,11 +59,13 @@
...
 
 	..()
 
 /datum/character_trait/biology/lisp/unapply_trait(mob/living/carbon/human/target)
 	REMOVE_TRAIT(target, TRAIT_LISPING, TRAIT_SOURCE_QUIRK)
+	target.speech_problem_flag = 0
```
harryob, line 68:
```suggestion
	target.speech_problem_flag = FALSE
```

code/modules/mob/living/carbon/human/say.dm
```diff
@@ -301,6 +301,7 @@ for it but just ignore it.
 			message = uppertext(message)
 			verb = pick("yells like an idiot","says rather loudly")
 	if(HAS_TRAIT(src, TRAIT_LISPING))
+		handled = 1
```
harryob, line 304:
```suggestion
		handled = TRUE
```

harryob (changes requested):
i'm aware these changes make stuff inconsistent, but all of these cases (handled and speech_problem_flag should be using the TRUE/FALSE defines instead of 1/0 (they're the same thing but it's neater)

harryob (approved):
thanks!

PR #3763, opened 2023-06-29
M2C ammo resized to Medium QOL

ghostsheet:
# About the pull request

M2C should be consistent with all other heavy ammo, m56d, Smartgun ammo, HPR etc.

They don't fit in any other normal ammo belt so you don't need to worry about a pistol pouch with 6 M2C magazines

Consistency Good, QOL good, gives more gameplay option

Basically it lets you pack it into backpacks and other medium storage are, BUT not into other specific belt

# Changelog

:cl:
add: M2C ammo boxes are now Medium size item
/:cl:


morrowwolf (commented):
This is the opposite of how I'd like this trend to go.

PR #3771, opened 2023-06-29
Queen makes movement noise even while resting

BeagleGaming1:
# About the pull request

Removes the resting check from the footstep component

The queen could be dragged while resting to be moved without making any noise

# Changelog

:cl:
add: Queen makes noise when moved even while resting
/:cl:

### DIFF
--- a/code/datums/components/footstep.dm
+++ b/code/datums/components/footstep.dm
@@ -31,7 +31,7 @@
 		return
 
 	var/mob/living/LM = parent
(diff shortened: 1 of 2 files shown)


harryob (approved):

Weuyn:
okay but what if she made the sandstone door grinding sound instead

ihatethisengine:
> okay but what if she made the sandstone door grinding sound instead

God this is amazing idea

PR #3772, opened 2023-06-30
Acid runner cannot get acid stacks from nested humans

BeagleGaming1:
# About the pull request

If a human is in a nest, the acid runner won't get any acid from it
If a human is stage 4 or 5 (about to burst), the runner won't get acid either, to prevent circumventing the above change.

It's against the rules anyways

# Changelog

:cl:
add: Acid runners don't get acid from slashing nested humans
add: Very late-stage marines do not give acid
/:cl:


ihatethisengine:
Nothing mechanically stops you from unnesting them. Maybe a better solution would be to not give acid stacks from hugged or late stage hugged?

BeagleGaming1:
> Nothing mechanically stops you from unnesting them. Maybe a better solution would be to not give acid stacks from hugged or late stage hugged?

Probably a good idea, I'm gonna do this later or in another pr if I forget

Diegoflores31:
> Nothing mechanically stops you from unnesting them. Maybe a better solution would be to not give acid stacks from hugged or late stage hugged?

No more benefits to combat hugged marines please

BeagleGaming1:
> > Nothing mechanically stops you from unnesting them. Maybe a better solution would be to not give acid stacks from hugged or late stage hugged?
> 
> 
> 
> No more benefits to combat hugged marines please

It's only stage 4 and 5, which by that point the marine is moments from bursting

Diegoflores31:
> > > Nothing mechanically stops you from unnesting them. Maybe a better solution would be to not give acid stacks from hugged or late stage hugged?
> > 
> > 
> > No more benefits to combat hugged marines please
> 
> It's only stage 4 and 5, which by that point the marine is moments from bursting

I think with them not winning acid while nested is enought . Nobody will unnest just to get 100 acid and even if they did it would be against rules

ihatethisengine:
Bro the point is to replace the rule with this mechanical limitation. And yeah, before the rule runners already had to unnest their victims to get stacks, because wall nesting wasn't a thing back then. 

Also it doesn't buff hugged marines, they still get acided. Just runners don't get acid stacks from them.

Diegoflores31:
> Bro the point is to replace the rule with this mechanical limitation. And yeah, before the rule runners already had to unnest their victims to get stacks, because wall nesting wasn't a thing back then. 
> 
> Also it doesn't buff hugged marines, they still get acided. Just runners don't get acid stacks from them.

Uh no? The rule is still in place and will remain . This is just something added on top of that you cannot kill captures unless hive is getting sieged.

Drulikar (approved):
Works; and yes *at least* stage 4 is required for this to have any effect. Stage 5 is basically 1 second from pop. IMO having an embryo at all should be grounds to not gain acid stacks, but we can push the stage level even lower in the future if needed.

PR #3790, opened 2023-07-03
Changes evac pod crash chance from a define to a var

BeagleGaming1:
# About the pull request

Changes the crash from define to a variable to allow it to be changed
Adds a proc to force crash the shuttle, either if docked or in space

This allows the crash chance to be changed, instead of just being static

# Changelog

:cl:
code: evacuation pod crash chance changed to a var
code: proc added to force evacuation pods to crash
/:cl:


mullenpaul (approved):
I'm guessing that this is to allow for admins to modify things on the fly? One future change I would like to see is a chance of a dropship hijack to alter this crash landing chance - outside the scope of this work but food for thought.

BeagleGaming1:
> I'm guessing that this is to allow for admins to modify things on the fly? One future change I would like to see is a chance of a dropship hijack to alter this crash landing chance - outside the scope of this work but food for thought.

Yeah, mostly this, but also morrow plans to add ways to change pod crash chances, and it would probably be easier to edit those changes if its a variable instead of a define

PR #3791, opened 2023-07-03
Adds improvisation skill

Katskan:
# About the pull request

Adds a new skill for survivor doctor/synthetic roles, surgical improvisation. The skill removes penalty for using improvised/substitute tools and can reduce penalty for working on poor surfaces at mastery. 

Helps improve some of the gameplay experience of the survivor synthetic, and also adds a reason to take survivor doctor instead of researcher, since without this the researcher is simply better. Also, FORECON support tech gets the skill because it is soulful that a recon marine, who is expected to be more motivated/skilled, can do IB repair a little faster in the middle of nowhere. 

Tested on a private lobby to make sure it runs without breaking things and works as intended. Tested by spawning a survivor synthetic and a CMO, and having them do ribcage opening/closing/sawing on an OR table and rollerbed, and comparing times vs the expected times in code. 

# Changelog

:cl:
add: New skill, surgical improvisation 
/:cl:

### DIFF
--- a/html/changelogs/AutoChangeLog-pr-3828.yml
+++ b/html/changelogs/AutoChangeLog-pr-3828.yml
@@ -1,0 +1,5 @@
+author: "Ben10083"
+delete-after: True
+changes:
+  - rscdel: "Working Joes can no longer be fed."
+  - code_imp: "New trait: Cannot eat. Self-explanatory."


code/datums/skills.dm
```diff
@@ -868,6 +876,7 @@ SYNTHETIC
 		SKILL_JTAC = SKILL_JTAC_BEGINNER,
 		SKILL_INTEL = SKILL_INTEL_TRAINED,
 		SKILL_DOMESTIC = SKILL_DOMESTIC_MASTER,
+		SKILL_IMPROVISATION = SKILL_IMPROVISATION_MASTER
```
BeagleGaming1, line 879:
```suggestion
		SKILL_IMPROVISATION = SKILL_IMPROVISATION_MASTER,
```
Linter doesn't like it that way

BeagleGaming1:
> Tested by spawning a survivor synthetic and a CMO, and having them do ribcage opening/closing/sawing on an OR table and rollerbed, and comparing times vs the expected times in code.

What were the new times compared to the old times?

Katskan:
E.g. on the times, using a fork for wound closure would normally incur an extra 80% penalty so 3 seconds + 0.8*3 = 5.4 seconds. The CMO had the correct 5.4 seconds, the survivor synthetic was still at 3 seconds

stanalbatross:
You've only added this skill to three roles and added it to two checks, which is much more limited than the scope for a skill usually is
You could just use a trait instead of a skill and then species check if you want to check if it's a synth (with the max level skill) doing it

Diegoflores31:
I think you could rather just give the surv doctor a surgical medkit ? . Also synthsurv is already giga OP no need to make it even better

Katskan:
It only got balance approved as a skill for survivor medical roles/synthetic, so it doesn't work as a species check for synthetics because the shipside unit isn't supposed to have it unless I'm misunderstanding species. The survivor doctor having a surgical medkit is already an RNG chance thing that can happen and making it guaranteed has been denied for balance.

Katskan:
For those of you who are not familiar with how synthetic-related WL changes get balance approved, we have a discussion channel regarding changes where we can get the rubber stamp done as shown https://i.imgur.com/tzB8Esm.png

code/datums/skills.dm
```diff
@@ -338,7 +338,7 @@ CIVILIAN
 		SKILL_VEHICLE = SKILL_VEHICLE_SMALL,
 	)
 
-/datum/skills/civilian/survivor/doctor
+//survivor/doctor
```
BeagleGaming1, line 341:
```suggestion
/datum/skills/civilian/survivor/doctor
```

Issue #3719, opened 2023-06-24
Wall Nested Mobs Blocking Directional Targeting


Superwelder:
### Testmerges

#3601, #3645, #3644

### Description of the bug

Any nested mob on a wall is taking priority over non-nested mobs for targeting. This means marines standing in a tile occupied by a nested mob will be immune to directional slashing and tackling as the target of said clicks focuses on the nested mob(often unable to slash since you can't slash infected && nested mobs). This is very likely related to how bullets hit nested people when a xeno occupies the same tile as well.

### What's the difference with what should have happened?

Nested mobs should not be soaking tackles, slashes, or bullets just by virtue of being in the same tile as a non-nested mob.

### How do we reproduce this bug?

1. Nest a mob.
2. Have a marine stand in the nested tile.
3. As xeno with directional slash enabled, try to tackle the marine down using directional targeting.
4. Watch the marine prime grenades or worse inside the hive since they can not be taken down.

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Weuyn:
oh so this is why I kept missing tackles

Superwelder:
> 

Yes, this bug is particularly dangerous because xeno players often spit out devoured marines in the hive and if they just so happen to spit them out on a tile sharing a nested person, have fun!

Wintermote:
I think this same bug also causes nested marines to absorb bullets going through the tile they're on.

PR #3806, opened 2023-07-05
Removes nesting the dead

morrowwolf:
# About the pull request

This PR makes it impossible to nest the dead.

Dead people being placed on resin walls to create magical bullet shields is not the intention of nests.

<details>

</details>

# Changelog

:cl: Morrow
del: Removed nesting the dead
/:cl:


harryob (approved):

morrowwolf:
> Other note about this though is anyone alive that end up dying (impregnated or not) in a nest will still be in this situation. Also with #3719 nests just generally need target de-prioritization (out of scope unless you instead take an approach that fixes both problems).

Just minimizing intentional abuse of this for the time being. I wouldn't be against making them bullet pass through though and I don't *think* it would be too difficult.

PR #3817, opened 2023-07-05
Buffs trashbags

Zonespace27:
# About the pull request

- Trashbags can now be looked through like boxes
- Trashbags now fit normal-sized items
- Trashbags no longer fit as a belt
- Trashbags have the same storage limits as a backpack

- If you accidentally put something in the bag, the only way to get it out is by dumping it into disposals, which is bad UX
- This might be me coping, but holy hell is it annoying to not be able to put still-small things like grenade packets and shoes in a trashbag to throw away
- It occurred to me that marines might take the bag as a belt with these changes, to be a better G8 pouch, so I removed the ability to do that. This should not affect the primary users of trashbags (WJs) because they aren't allowed to remove their toolbelts in the first place 
- This makes trashbags worse backpacks, which morrow was alright with

<details>

image

</details>

# Changelog
:cl:
balance: Trashbags now hold normal items and can be looked through like a box or storage container.
balance: Trashbags no longer fit in your belt slot.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/bags.dm
+++ b/code/game/objects/items/storage/bags.dm
@@ -31,7 +31,8 @@
 
 	w_class = SIZE_LARGE
 	max_w_class = SIZE_MEDIUM
-	storage_slots = 21
+	storage_slots = null
+	max_storage_space = 21 //equivalent to an IMP backpack
 	can_hold = list() // any
 	cant_hold = list(/obj/item/disk/nuclear, /obj/item/weapon/throwing_knife)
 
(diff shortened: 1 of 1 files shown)


morrowwolf (commented):
This is a balance nightmare.

If it can fit in your hand people will bring it and use it as extended storage and this can hold *a ton* and *easily*.

The reason we have the requirement to only remove stuff by putting it in a trashcan is to avoid people using it as storage. Either make it so only Working Joes can take stuff out or reduce the size dramatically (which would include not allowing medium items).

PR #3836, opened 2023-07-07
Adds improvisation traits

Katskan:
# About the pull request

Adds a new skill for survivor doctor/synthetic roles, surgical improvisation. The skill removes penalty for using improvised/substitute tools and can reduce penalty for working on poor surfaces at mastery. 

Helps improve some of the gameplay experience of the survivor synthetic, and also adds a reason to take survivor doctor instead of researcher, since without this the researcher is simply better. Also, FORECON support tech gets the skill because it is soulful that a recon marine, who is expected to be more motivated/skilled, can do IB repair a little faster in the middle of nowhere. 

Tested on a private lobby to make sure it runs without breaking things and works as intended. Tested by spawning a survivor synthetic and a CMO, and having them do ribcage opening/closing/sawing on an OR table and rollerbed, and comparing times vs the expected times in code. 

# Changelog

:cl:
add: New trait, surgical improvisation 
/:cl:


code/modules/surgery/surgery_steps.dm
```diff
@@ -111,8 +112,8 @@ affected_limb, or location vars. Also, in that case there may be a wait between
 			for(var/obj/surface in get_turf(target)) //Otherwise, get the lowest surface modifier of objects on their turf.
 				if(surface_modifier > surface.surgery_duration_multiplier)
 					surface_modifier = surface.surgery_duration_multiplier
-
-		step_duration *= surface_modifier
+			if(!(HAS_TRAIT(user, TRAIT_FASTLEARNER)))
```
morrowwolf, line 115: No go for fast learner

code/modules/surgery/surgery_steps.dm
```diff
@@ -102,7 +102,8 @@ affected_limb, or location vars. Also, in that case there may be a wait between
 
 	if(ispath(tool_type)) //Tool speed modifier. This means hand & any item are 100% efficient as surgical tools.
 		tool_modifier = tools[tool_type]
-		step_duration *= tool_modifier
+		if(!(HAS_TRAIT(user, TRAIT_IMPROVISER)))
+			step_duration *= tool_modifier
```
morrowwolf, line 106: Just half the malus rather than completely remove it, regular tools should still be better

morrowwolf (changes requested):

Katskan:
Renegged on agreed end-result prior to starting work.

PR #3844, opened 2023-07-09
Makes the speed for an embryo to burst a config + embryo buff

BeagleGaming1:
# About the pull request

Changed it from a number to a config.
Changed `larva_gestation_multiplier` to 2, as it only processes once every 2 seconds, so it should increase the counter by 2 at a time.
Changed an `else if` to `if` as there could never be a situation where the stage was equal to 4 but not less than 5.

Makes it easier to change, and keeps people guessing what stage they are at when hugged.

People were operating under the assumption that it was 7.5 minutes. However, due to `process` only running once every 2 seconds, it ended up taking twice as long.

Stage 4 was meant to make stasis slightly less effective, but due to a bug it just never happened.

# Changelog

:cl:
config: Moved time to burst to the config
balance: Fixed time to burst, lowering it from 15 minutes to 7.5 minutes by default
balance: Stasis bags used on late-stage mobs properly slow growth down less
/:cl:


BeagleGaming1:
chance? what? am I losing it?

config/example/config.txt
```diff
@@ -239,3 +239,6 @@ GAMEMODE_ROUNDS_NEEDED 5
 
 ## Default gamemode to auto-switch back to after a round has concluded
 GAMEMODE_DEFAULT extended
+
+## How long the mob will take to chestburst in seconds
```
Zonespace27, line 243: This is wrong. The game will interpret 450 as 45 seconds (because byond does things in ds)

code/modules/mob/living/carbon/xenomorph/Embryo.dm
```diff
@@ -76,6 +76,8 @@
 
 /obj/item/alien_embryo/proc/process_growth()
 	var/datum/hive_status/hive = GLOB.hive_datum[hivenumber]
+	//The total time the person is hugged divided by stages until burst
```
Zonespace27, line 79:
```suggestion
	/// The total time the person is hugged divided by stages until burst
```

code/modules/mob/living/carbon/xenomorph/Embryo.dm
```diff
@@ -76,6 +76,8 @@
...
+	var/per_stage_hugged_time = CONFIG_GET(number/embryo_burst_timer)/5
```
Zonespace27, line 80:
```suggestion
	var/per_stage_hugged_time = CONFIG_GET(number/embryo_burst_timer) / 5
```

Zonespace27 (changes requested):

ihatethisengine:
More shadownerfs incoming

config/example/config.txt
BeagleGaming1: It does not, the counter for processing ticks every second

BeagleGaming1:
I just noticed something strange with the embryo timer

config/example/config.txt
Zonespace27: Scuffed

morrowwolf:
TM'ing because I'm paranoid

morrowwolf:
Seems fine let's goooooooooo

PR #3853, opened 2023-07-10
Removes Rejuvinate and adds restricted proc logs

realforest2001:
# About the pull request
Removes R_REJUVINATE as it's only used in two places and we don't need it.
Adds some alerts where procs are blocked from being force called, and informs staff if someone attempts to brute force them.

<details>

</details>

# Changelog
:cl:
admin: Removed the R_REJUVINATE permissions flag and replaces instances where it is used with R_MOD.
admin: Added logs if a restricted proc is proccalled.
/:cl:


code/modules/admin/topic/topic.dm
```diff
@@ -982,7 +982,7 @@
 		message_admins("[key_name_admin(usr)] has sent [key_name_admin(M)] to the thunderdome. (Observer.)", 1)
 
 	else if(href_list["revive"])
-		if(!check_rights(R_REJUVINATE)) return
+		if(!check_rights(R_MOD)) return
```
harryob, line 985:
```suggestion
		if(!check_rights(R_MOD)) 
			return
```

harryob (changes requested):

harryob:
also when we're finally free of 514 remind me to change the strings being passed here into defines using __TYPE__ and __PROC__ ty

harryob (approved):

Issue #3863, opened 2023-07-11
The High Explosive OB has been unintentionally nerfed


ondrej008:
### Testmerges

#3676 #3783  #3784 #3729 #3726

### Description of the bug

PR #3658 nerfed the High Explosive OB by introducing a stun to the OB.

Xenomorphs are stunned *just* before the OB hits, reducing their damage taken by half.

### What's the difference with what should have happened?

The PR reduced the HE OB's damage by half.

### How do we reproduce this bug?

1. Load version before the PR, blow up some xenos but not so much that they all die.
2. Measure the damage taken.
3. Load version after the PR, blow up some xenos but not so much that they all die.
4. Measure the damage taken.
5. Compare the damages taken before and after the PR, notice how they take less damage after the stun PR was merged.

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #3864, opened 2023-07-11
Fixes the damage caused by the HE OB

ondrej008:
# About the pull request

Fixes issue #3863

This PR fixes an unintentional nerf to the HE OB.

BEFORE:
Mature Queen, Praetorian, Crusher, Ravager at the edge of the screen from a HE OB
image
Prae gets obliterated, Ravager goes into crit, Crusher has 435 health, Queen has 726 health.
AFTER:
same setup
Prae gets obliterated, Ravager gets obliterated, Crusher has 253 health, Queen has 539 health.

# Changelog

:cl:
fix: The HE OB now deals the correct amount of damage to xenos, before it dealt half damage caused by xenos being forced to rest before it hit.
/:cl:

### DIFF
--- a/code/modules/cm_marines/orbital_cannon.dm
+++ b/code/modules/cm_marines/orbital_cannon.dm
@@ -430,7 +430,6 @@ var/list/ob_type_fuel_requirements
 	icon_state = "ob_warhead_1"
 	shake_frequency = 3
 	max_shake_factor = 15
-	max_knockdown_time = 6
 
 	var/clear_power = 1200
 	var/clear_falloff = 400


fira:
Why not just increase the damage then

ondrej008:
> Why not just increase the damage then

this would also work but why are people knocked down before the OB hits it makes zero sense
with incendiary the OB spawns all at once but the HE OB will already knock you down just because it's an explosion

ondrej008:
> > Why not just increase the damage then
> 
> Clearly a better solution would be to add some flags or something to some explosions to ignore «resting» state. Halving damage is too much in many cases.
wouldn't people just add this onto nearly every explosion then?
I think resting just before an explosion hits is pretty satisfying and can save someone who otherwise would be toast, would feel like a shame to remove it from some types of explosions

PR #3902, opened 2023-07-15
You now lose hiding when melee attacking

Ben10083:
# About the pull request

Hiding now has 0.5 second cooldown, if you melee attack it will be 2 second cooldown and you lose hiding.

For melee users, having to try to hit a target that is hiding can be an exercise in futility, with many aspects such as hitbox, placing weapon on table by mistake, and other factors, leading to a overall unfair advantage. (Other abilities unaffected.)

<details>

https://github.com/cmss13-devs/cmss13/assets/91219575/238673b5-6dff-4362-ab32-ef4b3e208dcc

</details>

# Changelog

:cl:
balance: Xenomorphs lose their hiding ability for 2 seconds after a melee attack. 0.5 second cooldown applied to hide ability.
/:cl:


morrowwolf:
Fill this out please:

Gameplay:
Does it feel good to use?
Does it feel good to play against?
Does it affect other systems?

Roleplay:
Does it increase immersion?
Does it increase communication/interaction between players?

Simulation:
Does it fit in the CM universe aesthetically?
Does it fit in the lore of CM?
Does it make sense “realistically”

Ben10083:
Gameplay:
Does it feel good to use?
Hiding as it stands makes it so you appear under items instead of over them, this however does NOT mean you could 'hide' under a bullet. As such, unless you are playing against predators or synths, you will not feel much of a impact with this change, outside of the obvious element of no longer hiding
Does it feel good to play against?
If you ARE having to fight a xenomorph using melee, this change would make it so it is more fair, as they can no longer use hide to hide their sprite to make it hard for you to retaliate against them (by going under table to make it so you lose your item, for example.)
Does it affect other systems?
To my knowledge, hiding only affects what layer the xenomorph is on.

Roleplay:
Does it increase immersion?
A xenomorph no longer would be slashing you, running under a table, and staying at the table to avoid getting hit.
Does it increase communication/interaction between players?
No, this is a combat-related change

Simulation:
Does it fit in the CM universe aesthetically?
Yes
Does it fit in the lore of CM?
Yes
Does it make sense “realistically”
Yes, when it is time to start attacking someone, the time to hide is over. If the battle ends up not going your way, the cooldown should be low enough where you can activate hide once more and use it to help you escape. Overall this change should only deter those who use hide actively in combat for the purpose of hiding their sprite.

code/modules/mob/living/carbon/xenomorph/mutators/behavior_delegate.dm
```diff
@@ -65,6 +65,9 @@
 
 /datum/behavior_delegate/proc/melee_attack_additional_effects_self()
 	SEND_SIGNAL(bound_xeno, COMSIG_XENO_SLASH_ADDITIONAL_EFFECTS_SELF)
+	var/datum/action/xeno_action/onclick/xenohide/hide = get_xeno_action_by_type(bound_xeno, /datum/action/xeno_action/onclick/xenohide)
```
morrowwolf, line 68: This feels like it should be in specific behavior delegates for xenos with hide rather than on the base behavior delegate

morrowwolf (changes requested):

code/modules/mob/living/carbon/xenomorph/mutators/behavior_delegate.dm
Ben10083: Implemented, hellhound and runner given new behavior datums to allow this.

PR #3910, opened 2023-07-16
Synthflesh and Predalien Butchering of Synthetics Revamp

Ben10083:
# About the pull request

Predalien butcher now makes synthetic meat from synths, and cannot make meat from Working joes (still get gibbed though). At request of the council, new meat subtype under synthflesh for when this occurs. Meat is slightly toxic.

Working Joes turning into edible meat has terrifying undertones and shouldn't happen, synthetics now make special meat when butchered by predalien

# Changelog
:cl:
fix: Working Joes no longer spawn meat when butchered by Abominations
fix: Synthetics now are butchered into synthflesh when butchered by Abominations
add: special meat subtype 'synthflesh' if Synthetics get butchered. Uses same recipes as synthmeat
imageadd: synthflesh sprite
/:cl:


Ben10083:
The Forbidden meat 
image

harryob (changes requested):
doing type checks in a loop isn't ideal. *ideally* these vars would be stored on the species tbh. you could do that. you should

PR #3929, opened 2023-07-19
Fix the M56D ignoring collision when dismounting the user

### LINKED ISSUE #3926: M56D Can be used to clip through walls
### Testmerges

#3676: 'New Varadero revamp' by AnuvKH at commit 22e1d929bf #3783: 'Project ARES - Apollo Controller' by realforest2001 at commit 0c6722c470 #3729: 'Ravager and Hedgehog changes' by GoldenDarkness55 at commit 16b30ac8f2 #3850: 'Donator Kits' by realforest2001 at commit 3190146bfe #3879: 'Yautja Rework Holder' by realforest2001 at commit 442019a3d5 #3749: 'Re-balances acid damage' by Diegoflores31 at commit 19f32dfa4e #3909: 'Ports TGMC autofire' by Zonespace27 at commit b1699379ff #3908: 'Changes to squad color code' by BeagleGaming1 at commit 270557fdcd

### Description of the bug

You are able to wrench the M56D while manning it, rotating yourself and the gun, you are 1 tile behind the gun, which means, if you place the gun adjacent to a wall and rotate it you are able to clip through the wall and pass over them. You can go into space and probably into places where the area is intended to be inaccassable by all means.

I believe you should instead be ejected to the tile M56D is in rather than 1 tile behind the handles. You also probably shouldn't be able to rotate the gun with a wrench while you are using it.

1. Setup your M56D adjacently to a wall, grab the handles.
2. Equip a wrench and rotate the gun where your character is now on the same tile as the wall.
3. Unbuckle yourself from the M56D and now you can traverse to the other side of the wall.

### Issue Bingo

Awantje:
# About the pull request

Fixes #3926

It used a force move to dismount from the gun causing these issues. Now, if you can exit behind you will exit behind the gun.

You could exit into turfs with a density of 1 using this. And you should not be able to.

<details>

Testing it. With cades and a window frame. To ensure it works.
https://imgur.com/a/eT2LBjW

</details>

# Changelog
:cl:
fix: The M56D no longer let's you violate density code when exiting it.
/:cl:

### DIFF
--- a/code/modules/cm_marines/smartgun_mount.dm
+++ b/code/modules/cm_marines/smartgun_mount.dm
@@ -906,7 +906,7 @@
 	user.visible_message(SPAN_NOTICE("[user] lets go of \the [src]."),SPAN_NOTICE("You let go of \the [src], letting the gun rest."))
 	user.unfreeze()
 	user.reset_view(null)
-	user.forceMove(get_step(src, reverse_direction(src.dir)))
+	user.Move(get_step(src, reverse_direction(src.dir)))
 	user.setDir(dir) //set the direction of the player to the direction the gun is facing
 	user_old_x = 0 //reset our x
 	user_old_y = 0 //reset our y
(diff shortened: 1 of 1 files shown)


Awantje:
(Revert was needed I was checking the changes between step and move because gitlens showed the original used step instead of move.)

Drulikar (approved):

PR #3933, opened 2023-07-19
Removes neuro gas screen spin effect

GoldenDarkness55:
# About the pull request
Removes the nauseating screen spin effect of neuro gas.

Has no real effect on gameplay and seems to exist only to torment the player.
The screen spin isn't severe enough to throw off aim nor does it make things harder to spot. As far as I can tell the only effect it has is giving the player nausea.
After several minutes (!!!) of my screen spinning around from getting hit by a neuro glob I was seriously considering ahelping for an aheal, but opted to stand in a corner and go do something else irl until the effect passes, because any more and I'd hurl. I do not believe motion sickness meds and a bucket should be a requirement for CM.

Can replace with some other, actually mechanically impairing effect if required, but I can't be arsed to test the absolutely insane durations the gas seems to apply. Also would happily expand this to remove the hallucination audio aside from the lurker one if allowed to, because between vomiting and having to take off my headphones before I go deaf from max volume CAS blaring into my ears, it's pretty hard to tell which is worse.

<details>

</details>

# Changelog

:cl:
del: Removes the screen spinning (dizzy) effect of neuro gas
/:cl:

### DIFF
--- a/code/datums/effects/neurotoxin.dm
+++ b/code/datums/effects/neurotoxin.dm
@@ -40,7 +40,6 @@
 // General effects
 	affected_mob.last_damage_data = cause_data
 	affected_mob.apply_stamina_damage(stam_dam)
-	affected_mob.make_dizzy(12)
 
 // Effect levels (shit that doesn't stack)
 	switch(duration)
@@ -102,7 +101,6 @@
 		step(affected_mob, pick(CARDINAL_ALL_DIRS))
 		affected_mob.apply_effect(5, DAZE) // Unable to talk and weldervision
 		affected_mob.make_jittery(25)
-		affected_mob.make_dizzy(55)
 		affected_mob.emote("pain")
 		stumble = FALSE
 		addtimer(VARSET_CALLBACK(src,stumble,TRUE),3 SECONDS)


Diegoflores31:
Dizzy effect can be healed with chems.

ihatethisengine:
It's soulful, plus depending on severity it can affect your aim.

GoldenDarkness55:
> Dizzy effect can be healed with chems.

yeah bro everyone should carry tea or have the game bully him irl with nausea.



> It's soulful, plus depending on severity it can affect your aim.

I fail to see the soul in effects that make me want to ghost or alt f4. And no, I've had it at high severity hence why it lasted for minutes, it does not throw aim off at all. That only happens with sharp screen movements like buck PB shakes that were removed.
Screen blur or literally any other vision effect would both be more soulful and impactful so I really dunno why you feel the need to defend the only one that affects people irl

Diegoflores31:
The dizzy effect is on all ss13 servers. For YEARS i havent seen anyone ever complaining about irl effects specially with how infrequent the shake is . If you want it different go ahead and do a rework dont just nuke it out because it anoys you

GoldenDarkness55:
Here's the full 5 minute effect anyways https://streamable.com/ln4sv1 and god forbid you catch a whiff of more neuro during this cause it'll potentially never go away.

> For YEARS i havent seen anyone ever complaining about irl effects specially with how infrequent the shake is

You get to see it now then because someone has implemented an abominably long shake effect. Might wanna google motion sickness if you're really this ignorant.


> If you want it different go ahead and do a rework dont just nuke it out because it anoys you

The only rework I'd be willing to do regarding this is deleting it as offscreen globs were already removed once for good reason. But no, there's actually no need to replace it with anything because as I said - it is mechanically insignificant and serves only to harass the player.

Drulikar (changes requested):
I am pretty sensitive to motion sickness, but this effect doesn't feel too bad - just maybe potentially too long. So how about we implement another realistic way to mitigate this in game:
- Add a check inside the while loop of `/mob/proc/dizzy_process()` where if you're buckled or resting, your pixel_x and pixel_y get set to 0.
- Add a to_chat message at the end (when it leaves the while loop) mentioning the dizziness has passed (so you know to get up).
- Optionally lower the max dizziness value from 1000 to maybe 500 in /mob/proc/make_dizzy(amount)

GoldenDarkness55:
> I am pretty sensitive to motion sickness, but this effect doesn't feel too bad - just maybe potentially too long. So how about we implement another realistic way to mitigate this in game:
> 
> * Add a check inside the while loop of `/mob/proc/dizzy_process()` where if you're buckled or resting, your pixel_x and pixel_y get set to 0 instead (or cap the movement to something very small so you can still see the tiniest amount of it).
> * Add a to_chat message at the end (when it leaves the while loop) mentioning the dizziness has passed (so you know to get up).
> * Optionally lower the max dizziness value from 1000 to maybe 500 in `/mob/proc/make_dizzy(amount)`

Setting it to 500 doesn't seem to change much, possibly makes it even worse.
As for the lying down, that's just confusing for everyone around the marine as they'll try to shake, drag, etc and achieves the same thing as standing in a corner and alt tabbing or minimizing the window.

If the effect absolutely has to stay can it not be shortened to a sane value instead - around half a minute to match the jittery effect? Should be infinitely more bearable that way.

Drulikar:
> If the effect absolutely has to stay can it not be shortened to a sane value instead - around half a minute to match the jittery effect? Should be infinitely more bearable that way.
Reducing the max from 1000 to 500 is exactly that - cutting the maximum duration and intensity in half.

And actually, looks like resting already is implemented such that it eats away at the dizziness faster: 
image

GoldenDarkness55:
> > If the effect absolutely has to stay can it not be shortened to a sane value instead - around half a minute to match the jittery effect? Should be infinitely more bearable that way.
> 
> Reducing the max from 1000 to 500 is exactly that - cutting the maximum duration and intensity in half. So as I said you could neuter the visual effects when resting/buckled. As far as people shaking you - communicate.
> 
> Please use VSC and find all references of this variable so you can understand how it is implemented, and why I am suggesting what I am suggesting.
> 
> And actually, looks like resting already is implemented such that it eats away at the dizziness faster

Isn't 1000 to 500 still 500 seconds at max though? As for intensity I tested it after changing the value to 500 and as I said it is not much different.
I guess it's acceptable with that rest code, but is there some easy way to give the dizzy players a message to lie down every few seconds?

Drulikar:
The value is not seconds, and as I showed just above it is reduced by either 3 or 15 every `/mob/living/carbon/human/handle_regular_status_updates` which is called every stun_callback ((stunned/recovery_constant) * 2 SECONDS if stunned), knocked_down_callback ((knocked_down/recovery_constant) * 2 SECONDS if knocked down), knocked_out_callback ((knocked_out/recovery_constant) * 2 SECONDS if knocked out), and /mob/living/carbon/human/Life (2 SECONDS).

So assuming you aren't stunned, knocked down, or knocked out: The value is effectively 0.6 seconds for every 1 dizziness or 0.13 seconds for every 1 dizziness if resting.

I do recommend also doing the other two things I suggested; so far you only have implemented the third optional one I suggested.

To periodically suggest resting, inside `/mob/proc/dizzy_process()` when you aren't resting throw in a prob to send a to_chat message to suggest it.

GoldenDarkness55:
> The value is not seconds, and as I showed just above it is reduced by either 3 or 15 every `/mob/living/carbon/human/handle_regular_status_updates` which is called every stun_callback ((stunned/recovery_constant) * 2 SECONDS if stunned), knocked_down_callback ((knocked_down/recovery_constant) * 2 SECONDS if knocked down), knocked_out_callback ((knocked_out/recovery_constant) * 2 SECONDS if knocked out), and `/mob/living/carbon/human/Life` (2 SECONDS).
> 
> So assuming you aren't stunned, knocked down, or knocked out: The value is effectively 0.6 seconds for every 1 dizziness or 0.13 seconds for every 1 dizziness if resting.
> 
> I do recommend also doing the other two things I suggested; so far you only have implemented the third optional one I suggested.
> 
> To periodically suggest resting, inside `/mob/proc/dizzy_process()` when you aren't resting throw in a low prob to send a to_chat message to suggest it; or add var that you can assign world.time to so you only send it every so often.

I see, thank you for the explanation. Done the other 2 now, added the prob message and updated title, description and CL to match.

Drulikar (changes requested):

code/modules/mob/mob.dm
```diff
@@ -609,16 +609,22 @@ note dizziness decrements automatically in the mob's Life() proc.
 	is_dizzy = 1
 	while(dizziness > 100)
 		if(client)
+			if(buckled || resting)
+				client.pixel_x = 0
+				client.pixel_y = 0
+		else
```
Drulikar, line 615: The else should be an else for `if(buckled || resting)`; not if theres no client. Everything that is client.whatever needs a client to be valid.

Drulikar (changes requested):

code/modules/mob/mob.dm
GoldenDarkness55: whoops. fixed

Drulikar (approved):
Code changes should be fine, just be sure to test it out to make sure its doing what you want now and undraft when you're ready.

PR #3936, opened 2023-07-19
Fixes doubling of gadget points for second+ synth

Katskan:
# About the pull request

Fixes a bug where if any synthetic that joins will get 45 extra gadget points more than the previous, so 2nd synthetic gets 90 points, 3rd would have 135, and so on. Tested in a private server and working properly.

Makes it so late join synthetics or additional joins from events don't get massive gadget budget to spend on like 3 compact defibrillators or start taking things to hand out to marines. 

Went into private server running the branch, spawned 3 shipside USCM synthetics. Sequentially vended the coin, then took 3 gadgets, then did the same with the next. Pre-patch this would result in the 2nd and 3rd synthetics getting huge point budgets, with this patch correctly only gave 45 points to each synthetic. 

</details>

# Changelog

:cl:
fix: Fixed synthetics getting an extra 45 points more than any previous who had used the gadget vendor
/:cl:


TeDGamer:
`available_points = available_points = 45`

Shouldn't it be condensed to available_points = 45

harryob (approved):

PR #3939, opened 2023-07-19
Lesser Drones

morrowwolf:
# About the pull request

This PR adds lesser drones that can be joined as from ghost.

You can have up to 1/3rd of the hive as lesser drones.

Lesser drones can build basic resin walls/membranes/doors and can spread resin nodes.

Lesser drones can give off weak pheromones.

Lesser drones are very weak and die quickly. They will always gib to avoid intel bloat/clutter.

When lesser drones die they do not change your last time of death so you do not lose your queue spot. Perfect to play while waiting to get in.

You can spawn as a lesser drone by using the verb in the ghost tab or clicking on the hive core.

Once you die as a lesser drone you must wait 3 minutes to become one again.

Having squishy and fast xenos with very little consequence of loss adds to the "horde-like" feel of xenos.

Having something for xenos to do while waiting as a ghost is good.

Makes the game feel more alive.

# Changelog

:cl: Morrow
add: Added lesser drones
/:cl:


Ben10083:
image

Finally some good castes

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -1047,6 +1051,37 @@
 	playsound(hugger, 'sound/effects/xeno_newlarva.ogg', 25, TRUE)
 	hugger.generate_name()
 
+/datum/hive_status/proc/update_lesser_drone_limit()
+	lesser_drone_limit = Ceiling(totalXenos.len / 3)
```
ihatethisengine, line 1055: The issue here is that totalXenos list includes all xenos, including facehuggers and lesser drones. So every time you join as lesser drones, you essentially increase the limit for both facehuggers and lesser drones. 

Also, same issue with facehuggers, they increase the limit for both parties as well.

morrowwolf: This is a consideration and one I am happy to leave as is.

code/game/gamemodes/cm_initialize.dm
```diff
@@ -548,6 +548,46 @@ Additional game mode variables.
...
+		to_chat(xeno_candidate, SPAN_WARNING("The selected hive does not have a hive core to spawn from!"))
+		return
+
+	if(!hive.living_xeno_queen)
+		to_chat(xeno_candidate, SPAN_WARNING("The selected hive does not have a Queen!"))
+		return
+
+	if(!hive.living_xeno_queen.ovipositor)
+		to_chat(xeno_candidate, SPAN_WARNING("The selected hive does not have a Queen on Ovipositor!"))
+		return
```
Drulikar, line 585: Tests here are inconsistent with can_spawn_as_lesser_drone. This also ultimately results in calling can_spawn_as_lesser_drone via spawn_lesser_drone. So that is to say the checks `if(!hive.living_xeno_queen)` and `if(!hive.living_xeno_queen.ovipositor)` should be moved from here to can_spawn_as_lesser_drone.

code/modules/mob/dead/observer/observer.dm
```diff
@@ -413,8 +413,9 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 		if(ghost && !is_admin_level(z))
 			ghost.timeofdeath = world.time
 
-			// Larva queue: We use the larger of their existing queue time or the new timeofdeath except for facehuggers
-			var/new_tod = isfacehugger(src) ? 1 : world.time
+			// Larva queue: We use the larger of their existing queue time or the new timeofdeath except for facehuggers/lesser drones
+			var/new_tod = (isfacehugger(src) || islesserdrone(src)) ? 1 : world.time
```
Drulikar, line 417: You need to also update this same check in the proc `/mob/proc/ghostize(can_reenter_corpse = TRUE, aghosted = FALSE)`: https://github.com/cmss13-devs/cmss13/pull/3939/files#diff-08897ff01ebb08d739f4a3e87da529ac87080df7a6bfbe17fdfd85d5a2e5b401L370

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -1047,6 +1051,37 @@
...
+
+/datum/hive_status/proc/can_spawn_as_lesser_drone(mob/dead/observer/user)
+	if(!GLOB.hive_datum || ! GLOB.hive_datum[hivenumber])
+		return FALSE
+	if(jobban_isbanned(user, JOB_XENOMORPH)) // User is jobbanned
+		to_chat(user, SPAN_WARNING("You are banned from playing aliens and cannot spawn as a xenomorph."))
+		return FALSE
```
Drulikar, line 1062: Technically this check is currently redundant since the only thing `check_xeno_late_join` checks is if they are banned; but its ok to double check ever this proc gets called without that check.

Drulikar (changes requested):

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
ihatethisengine: Just to illustrate the issue, during chance claim round at some point xenos had 14 actual xenos, 10/10 lesser drones and 5/(?, likely also 10) facehuggers. If all facehugger slots were filled, limits would be even higher. Just pointing out.

Drulikar: Its a bit messy but there is `/datum/hive_status/proc/get_xeno_counts()` that is excluding all that have counts_for_slots false and those amounts can be tallied up like they are for the hive status UI `/datum/hive_status_ui/proc/update_xeno_counts(send_update = TRUE)` so if you do decide to go that way, can likely replicate that code or adapt it in a way that its more cached so hivestatus tgui and this can share the same data.

morrowwolf:
Two schools of thought on building:

Either make them build walls/doors/membranes every now and then and have them focus on filling haps

OR

Be more support-y with sticky/speed resin.

Currently leaning towards the first for now. May change.

Drulikar:
Nothing immediately stands out to me as needing changes code wise as it is now but can review again once it's in a state your happy with.

morrowwolf:
Need sprite reviewed anyway but good to go otherwise

Drulikar:
Other notes:
- Alt clicking a hive core still asks what hive to join when there are multiple
- Compact building list button doesn't resize correctly (its vertically too small with just 3 items)
image

Venuska1117:
> Two schools of thought on building:
> 
> Either make them build walls/doors/membranes every now and then and have them focus on filling gaps
> 
> OR
> 
> Be more support-y with sticky/speed resin.
> 
> Currently leaning towards the first for now. May change.

you can always try making support strain. ;)

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -1047,6 +1051,58 @@
...
+
+	if(lesser_drone_limit <= current_lesser_drone_count)
+		to_chat(user, SPAN_WARNING("\The [GLOB.hive_datum[hivenumber]] cannot support more lesser drones! Limit: <b>[current_lesser_drone_count]/[lesser_drone_limit]</b>"))
+		return FALSE
+
+	if(user.action_busy)
+		to_chat(user, SPAN_WARNING("You are already attempting to join the game!"))
+		return
+
+	user.action_busy = TRUE
```
harryob, line 1097: action_busy is a private var to the do_after proc just check that the user has a client after the alert

harryob (changes requested):

harryob (approved):

morrowwolf:
I may have done these pass flags incorrectly but we're gonna find out

harryob (approved):

Hopekz:
after playing around with them seems like it would add to their gameplay to be able to place down sticky resin because that's like a manual labor job that other builders wouldn't have to do so it would probably be appropriate for a lesser drone to do

PR #3940, opened 2023-07-19
Facehuggers now bypass time of death checks when joining as Larva

Ben10083:
# About the pull request

Facehuggers ~~(and other ghost roles? (#3939))~~[this functionality will have to be added in said PR] now bypass the five minute time of death check, letting them join as a larva via queue or.. join as larva button.

With infection timer at 7.5 minutes, this means that after you die as facehugger, a large portion of the infection time is spent as unjoinable, which isnt a problem for a host YOU infected, but for others, you would get skipped over due to time of death. This allows for more leniency when playing as low-survivablity ghost roles. 

Code added can also be expanded, I am sure the new variable can fit in other places.

# Changelog
:cl:
add: Facehuggers now bypass time of death checks when being considered for larva.
code: new variable for observers to handle bypass of time of death checks
/:cl:

### DIFF
--- a/code/__HELPERS/game.dm
+++ b/code/__HELPERS/game.dm
@@ -268,7 +268,7 @@
 
 		// copied from join as xeno
 		var/deathtime = world.time - cur_obs.timeofdeath
-		if(deathtime < XENO_JOIN_DEAD_TIME && ( !cur_obs.client.admin_holder || !(cur_obs.client.admin_holder.rights & R_ADMIN)) )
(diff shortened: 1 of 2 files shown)


Drulikar (changes requested):
You likely also want to change the same check in `/mob/living/proc/do_ghost()`: https://github.com/cmss13-devs/cmss13/pull/3940/files#diff-08897ff01ebb08d739f4a3e87da529ac87080df7a6bfbe17fdfd85d5a2e5b401L417 Unless you want ghosting to have the same effect; not just death.

Ben10083:
Thanks for heads up regarding ghosting.

Drulikar (approved):

PR #3944, opened 2023-07-19
AI Core APC power storage increase

Ben10083:
# About the pull request

AI Core APC now has a Hyper capacity cell like telecommunications

With ARES taking as much power as telecommunications (greedy bastard), having just a high capacity cell meant that it would only be able to operate without power for a very limited amount of time. This makes the power situation at AI core much more stable.

# Changelog
:cl:
maptweak: AI Core APC now uses a hyper capacity cell
/:cl:


Nanu308:
You should look into what the `/hardened` means

Ben10083:
> You should look into what the `/hardened` means

image

So setting the cell should still work.

Nanu308:
im too tired sometimes

PR #3947, opened 2023-07-20
Ghosts now notified when larva are about to chestburst out of un-nested

Ben10083:
# About the pull request

This PR tells ghosts when someone is about to be chestbursted (only if not nested)

It is good to have interesting events like chestbursts be told to ghosts before it happens, letting them observe the event in advance

image

# Changelog
:cl:
qol: Ghosts now informed when a non-nested host is about to chestburst
/:cl:


code/modules/mob/living/carbon/xenomorph/Embryo.dm
```diff
@@ -218,6 +218,18 @@
 		to_chat(new_xeno, "Talk in Hivemind using <strong>;</strong> (e.g. ';My life for the queen!')")
 		playsound_client(new_xeno.client, 'sound/effects/xeno_newlarva.ogg', 25, 1)
 
+	// Inform observers to grab some popcorn if it isnt nested
+	if(!HAS_TRAIT(affected_mob, TRAIT_NESTED))
```
Drulikar, line 222: This all should be tabbed over to be underneath the picked check, else this will be spammed whenever there isn't a candidate as it repeats to find one. It should only happen when there is a candidate.

code/modules/mob/living/carbon/xenomorph/Embryo.dm
```diff
@@ -218,6 +218,18 @@
...
+		var/area/burst_area = get_area(src)
+		if(burst_area)
+			for(var/mob/dead/observer/observer as anything in GLOB.observer_list)
+				to_chat(observer, SPAN_DEADSAY("A <b>[new_xeno.hive.prefix]Larva</b> is about to chestburst out of <b>[affected_mob]</b> at \the <b>[burst_area]!</b> [OBSERVER_JMP(observer, affected_mob)]"))
+			to_chat(src, SPAN_DEADSAY("A <b>[new_xeno.hive.prefix]Larva</b> is about to chestburst out of <b>[affected_mob]</b> at \the <b>[burst_area]!</b>"))
+		else
+			for(var/mob/dead/observer/observer as anything in GLOB.observer_list)
+				to_chat(observer, SPAN_DEADSAY("A <b>[new_xeno.hive.prefix]Larva</b> is about to chestburst out of <b>[affected_mob]!</b> [OBSERVER_JMP(observer, affected_mob)]"))
+			to_chat(src, SPAN_DEADSAY("A <b>[new_xeno.hive.prefix]Larva</b> is about to chestburst out of <b>[affected_mob]!</b>"))
```
Drulikar, line 231: Up to you, but all of this can just be combined into one loop if you just have a burst_area string that is either "" or <b>area</b> and the chat just includes [burst_string].

Drulikar (changes requested):

code/modules/mob/living/carbon/xenomorph/Embryo.dm
Ben10083: This seems to be what we typically do, so keeping it AS IS for now

code/modules/mob/living/carbon/xenomorph/Embryo.dm
Drulikar: Sorry it looks like its now setting stage to 6; last I remember it didn't here so no spamming. Regardless though theres no popcorn to be had if the larva just burrows because theres no one.

Ben10083: True, this works out anyway.

Drulikar: So actually, feel free to undo that indentation. So if we actually make it to stage 6 without a candidate the dchat message can help someone to actually take over that larva via join as xeno

Alternatively, keep it indented and tinker with having that call repeat during phase 6 if we have no mind yet. I swore it repeated the call but not seeing atm how it would repeat try to get a candidate.

Ben10083: Alright, will change back

Drulikar (approved):

PR #3985, opened 2023-07-24
Xeno endgame content part 1: pylons and larva

morrowwolf:
# About the pull request

This PR:

Allows xenos to capture comms relays after 90 minutes into the game.

Once a comms relay is captured, each comm relay gives 10% of the current hive total every five minutes. (Note: I'd like more impact, this is just part 1)

To capture a comms relay you must place a cluster nearby. This cluster then morphs into a pylon with CAS/mortar protection.

After a certain point in the game it can stale due to sieging or otherwise failure for either side to want to fight. This should create urgency and change the focus of the round if stagnating.

image

# Changelog

:cl: Morrow, Sleepynecrons (beautiful sprites)
add: Xenos can now capture comms relays for bonuses after 90 minutes into the round.
/:cl:


code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -367,17 +368,24 @@
...
+
+/datum/hive_status/proc/setup_evolution_announcements()
 	for(var/time in GLOB.xeno_evolve_times)
 		if(time == "0")
 			continue
 
 		addtimer(CALLBACK(src, PROC_REF(announce_evolve_available), GLOB.xeno_evolve_times[time]), text2num(time))
 
+/datum/hive_status/proc/setup_pylon_limits()
+	hive_structures_limit[XENO_STRUCTURE_PYLON] = length(GLOB.all_static_telecomms_towers) ? length(GLOB.all_static_telecomms_towers) : 2
```
Drulikar, line 387:
```suggestion
	hive_structures_limit[XENO_STRUCTURE_PYLON] = length(GLOB.all_static_telecomms_towers) || 2
```

Drulikar (changes requested):

Tarhalindur:
There should be some indication on the xeno examine text of the comms relay that you can do this, I think - especially since the changelog doesn't say how. Maybe there could also be a xeno announcement when the ability becomes available.

code/game/machinery/telecomms/presets.dm
```diff
@@ -282,6 +292,72 @@ GLOBAL_LIST_EMPTY(all_static_telecomms_towers)
...
+
+	var/obj/effect/alien/resin/special/cluster/cluster_parent = parent_node.resin_parent
+
+	var/list/held_children_weeds = parent_node.children
+	var/cluster_loc = cluster_parent.loc
+	var/linked_hive = cluster_parent.linked_hive
+
+	parent_node.children = list()
+
+	qdel(cluster_parent)
```
Drulikar, line 327: cluster_parent here can be a /obj/effect/alien/resin/special/pylon/core. Maybe you intend to delete `weeded_turf.weeds.parent` instead of `weeded_turf.weeds.parent.resin_parent`

morrowwolf:
Not fully deleting when hijacking
Increase resin ceiling range a bit
Make resin hive/hardy ???
Cut overlay flick time in half, seems to be close to working correctly
Forsaken should not be able to corrupt the towers, easier to make it so can't corrupt if in endgame probably

code/modules/cm_aliens/structures/special/pylon_core.dm
```diff
@@ -34,9 +37,27 @@
...
 
-	var/obj/effect/alien/weeds/node/pylon/W = locate() in loc
-	if(W)
-		qdel(W)
+	if(node)
+		qdel(node)
+
+	if(activated)
+		for(var/obj/effect/alien/resin/structure as anything in linked_hive.hive_structures)
+			if(!istypestrict(structure, /obj/effect/alien/resin/special/pylon))
```
Drulikar, line 45: Seems this must be null checked because it is trying to get the .type of something deleted.

Segrain: It actually runtimes anyway.
image

code/modules/cm_aliens/structures/special/pylon_core.dm
```diff
@@ -34,9 +37,27 @@
...
+				continue
+			var/obj/effect/alien/resin/special/pylon/pylon = structure
+			pylon.activated = FALSE
```
Segrain, line 48:
```suggestion
		if(XENO_STRUCTURE_PYLON in linked_hive.hive_structures)
			for(var/obj/effect/alien/resin/special/pylon/pylon as anything in linked_hive.hive_structures[XENO_STRUCTURE_PYLON]
				pylon.activated = FALSE
```

ihatethisengine:
What stops xenos from not hijacking until they get as much larva as possible?

Zonespace27 (changes requested):

morrowwolf:
Usual disclaimer: Not trying to ignore review just don't want to go touching stuff before finalizing the direction I want to go after seeing it in game a bit.

Will get to review probably later this week, thank you guys you are the best 🙏

code/game/machinery/telecomms/presets.dm
morrowwolf: Errr I don't think so, this should delete the cluster itself rather than the node attached to the cluster. I fixed the hive core being eaten issue (I think)

Drulikar: I am currently unable to reproduce that issue.

code/game/machinery/telecomms/presets.dm
```diff
@@ -212,12 +212,27 @@ GLOBAL_LIST_EMPTY(all_static_telecomms_towers)
 	freq_listening = list(COLONY_FREQ)
 	var/toggle_cooldown = 0
 
+	/// Tower has been taken over by xenos, is not usable
+	var/corrupted = FALSE
+
+	/// Held image for the current overlay on the tower from xeno corruption
+	var/image/corruption_image = null
```
harryob, line 219:
```suggestion
	var/image/corruption_image
```

code/modules/cm_aliens/structures/special/pylon_core.dm
```diff
@@ -87,8 +87,78 @@
...
+			var/datum/hive_status/checked_hive = GLOB.hive_datum[hivenumber]
+			if(!length(checked_hive.totalXenos))
+				continue
+
+			if(checked_hive == linked_hive)
+				xeno_announcement(SPAN_XENOANNOUNCE("We have lost our control of the tall's communication relay at [get_area(src)]."), hivenumber, XENO_GENERAL_ANNOUNCE)
+			else
+				xeno_announcement(SPAN_XENOANNOUNCE("Another hive has lost control of the tall's communication relay at [get_area(src)]."), hivenumber, XENO_GENERAL_ANNOUNCE)
+
+	. = ..()
```
harryob, line 117:
```suggestion
	return ..()
```

harryob (changes requested):
nits

harryob (approved):

PR #4001, opened 2023-07-26
Lower dropship equipment costs

GoldenDarkness55:
# About the pull request

Lowers manufacturing costs for useless/unused equipment.

CAS weaponry always was and will be a priority, therefore the transport and other QOL additions either need to be really strong/useful to warrant purchasing them, or have costs so negligible that it doesn't feel like griefing if you do.

LZ detectors and spotlights are exceedingly useless and a massive noobtrap/missclick bait, if they're not being removed they should at least be almost costless.
Rappel's a cool system, but sadly extremely underused and as far as I know never had much of an impact on a round - I see no point in gatekeeping fun situations for both sides behind that price cost. Encourages teamwork and communication, everyone wins here.
M56 similar to the above - sovlful, encourages fun and holdouts, but rather useless in practice and will likely get the user killed. A fun way to go though.

If some hidden worth of these ends up being discovered and they actually make a dent on the balance, the prices can be raised up accordingly.

<details>

</details>

# Changelog
:cl:
balance: Dropship machine gun cost lowered from 300 to 50.
balance: Dropship rappel module cost lowered from 500 to 50.
balance: Dropship spotlight cost lowered from 300 to 50.
balance: Dropship LZ detector cost lowered from 400 to 50.
/:cl:

### DIFF
--- a/code/modules/cm_marines/dropship_equipment.dm
+++ b/code/modules/cm_marines/dropship_equipment.dm
@@ -289,7 +289,7 @@
 	density = FALSE
 	equip_categories = list(DROPSHIP_WEAPON, DROPSHIP_CREW_WEAPON)
 	icon_state = "mg_system"
-	point_cost = 300
+	point_cost = 50
 	var/deployment_cooldown
 	var/obj/structure/machinery/m56d_hmg/mg_turret/dropship/deployed_mg
 	combat_equipment = FALSE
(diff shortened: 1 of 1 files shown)


ihatethisengine:
I was thinking of removing the requirement to have a rappel belt for rappeling because it's a limited thing and kind of discourages active usage this way.

GoldenDarkness55:
> I was thinking of removing the requirement to have a rappel belt for rappeling because it's a limited thing and kind of discourages active usage this way.

Not very familiar with how the rappel system works, but I remember hearing about some trouble related to lack of these the last time marines were going to try, so that'd be nice to see aswell.

morrowwolf (changes requested):
Up the 25s to 50s just because it looks better that way.

PR #2057, opened 2022-12-21
HPR mag antipowergame

stanalbatross:
# About the pull request

you can no longer take bullets out of HPR mags

# Explain why it's good for the game

stops people taking 500 HPR mags to act as tardis ammo storage for non-HPR weapons

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl:
balance: you can no longer remove bullets from HPR mags
/:cl:

PR #4003, opened 2023-07-26
Brings automatic weaponry to the marine and surv arsenal

Zonespace27:
# About the pull request
Gives full auto to the following guns:

m41a (and mk1)
m39
HPR (when bipodded)
m46c in IFF off
type71
mar40
nsg
every civvie SMG except the nailgun
laser uzi

All newly automatic guns fire 20% slower in automatic mode, to balance for the fact that it is much easier to re-aim than burst.

This creates a dynamic between the three firemodes:

Semi:
- Best accuracy
- Lowest DPS

Burst:
- Okay accuracy
- Highest DPS

Auto:
- Okay (it depends) accuracy, but you are able to adjust for a moving target unlike burst.
- Slightly lower DPS 

Guns not having autofire feels... pretty rough. This PR will bring CM to a more modern time, while reducing the number of players diagnosed with carpal tunnel yearly.

I went through and tested pretty much every base gun, but I cannot guarantee that they all handle well. Lemme know if something's off.

# Changelog
:cl:
balance: The M41A, M41Amk1, m39, m46c (iff off), mar-30/40/60, type 71, laser uzi, and every SMG now has automatic fire. To compensate for the ability to re-aim (unlike burst), the guns fire 20% slower while automatic. This does not apply to already-existing automatic weaponry.
balance: The HPR can now fire in full auto when bipodded.
fix: Adding/removing attachments now keeps the current firemode, if possible.
fix: The dualtube and spec sniper now can switch tubes/toggle laser again
fix: Fixed a long-standing exploit with bipods and scopes.
fix: Fixed guns jamming when shooting UI elements.
fix: Fixed a bunch of guns unintentionally having burst.
balance: You can transfer ammo between HPR ammo boxes by hitting one with another.
/:cl:


harryob (approved):

morrowwolf:
Alright list of stuff that needs done before this is merged:
MK1 kits removed from everywhere but CIC armory
Extended mags removed from any automatic weapons
Ammo cans no longer carryable on back
Higher slowdown on the ammo can backpack

More to come probably.

SarahJohnson489:
Reminder that HPR mags cannot be emptied by hand and that back 10x24 ammunition boxes were used to refill the partial mags since you can only swap to a full for the queen's sortie so many times before you're all partials. The rack packs are limited instead of being in prep vendors like the welderpacks if this is going to take that essential function away.

SarahJohnson489:
Telling you that there's a serious problem with the PR is as on-topic as it gets they aren't like other limited magazines where it's trivial to consolidate your ammo in the field. It was a working system and this PR breaks the working system. You cannot refute that. Y'all just ban anyone that isn't a yesman anywhere else if they tell you there's a problem with a pr.

Zonespace27:
> Telling you that there's a serious problem with the PR is as on-topic as it gets they aren't like other limited magazines where it's trivial to consolidate your ammo in the field. It was a working system and this PR breaks the working system. You cannot refute that. Y'all just ban anyone that isn't a yesman anywhere else if they tell you there's a problem with a pr.

Oh, sorry, misread your initial comment. I will bring this up in case anything needs to be changed with HPR mags.

SarahJohnson489:
Yeah it's kind of a mess. Normal mag partials you'd just pull some bullets out of one partial mag and slap them in the other. Stan stopped removing bullets from hpr mags in https://github.com/cmss13-devs/cmss13/pull/2057 then Morrow nuked hpr mags from squad prep in https://github.com/cmss13-devs/cmss13/pull/3534 which otherwise would have let you have more fulls to fallback for the main sorties cause of course they did so we'd wear ammo boxes on the back as a workaround to consolidate the HPR mags caused by prolonged combat. Now that's going away. 
It'd be kinda okay if imp rack backpacks were scale * 5 in the squad prep vendors even if that's being punished with a slowdown penalty cause of Stan and Morrow's pet peeve. Or if you could slap one hpr mag with the other to transfer bullets without being able to put them in m41/m4ra but they're not so it bricks the hpr performance a bit for a rather silly reason that needs an even sillier workaround as things get even more spaghetti from the fallout.

morrowwolf:
This needs a ton more work to balance properly. The change of skill floor and consistency of full auto has made marines incredibly strong.

Ammo economy can only do so much when marines are running over hives and this is likely compounded by previous balance decisions in the last few months to offset xeno strength.

Give me a week to put together a list of things that need changed before we try this again.

Zonespace27:
Update: we've agreed to try a slower rollout of autofire. First up: all SMGs now have autofire without a penalty to firerate.

morrowwolf (approved):
balance good

code/modules/projectiles/projectile.dm
```diff
@@ -1143,11 +1143,16 @@
 	// Need to do this in order to prevent the ping from being deleted
 	addtimer(CALLBACK(I, TYPE_PROC_REF(/image, flick_overlay), src, 3), 1)
 
+/// Zonennote: debug attempt for bullet lag
```
harryob, line 1146: 🚨 zonennote 🚨

Zonespace27: it worked so I'm keeping the change, removing the comment

harryob (approved):

PR #4005, opened 2023-07-26
Xeno interactions with reinforced walls and window frames

GoldenDarkness55:
# About the pull request

Allows acid below boiler level to melt reinforced window frames and reinforced walls.
Allows t2s to make holes in reinforced walls.

Reinf frames are already slashable, forgot to make them meltable when I PR'd slash fixes way back. Makes no sense to require boiler acid for these.

Since marines were buffed to be capable of making reinforced walls planetside and no one updated xenos to be able to break them, the meta thing to do has been to simply wall off flanks to FOB or tcomms with these as it prevents any t2s from accessing the area - it's way too easy and cheap to remove player interaction in this way. The intended way to protect flanks is to have people and/or sentry guns there, not wall them off else that'd be done by the mapper to begin with.
It also admittedly makes little sense to restrict slashing in this way, this PR makes it more consistent with bolted and welded doors aswell as window frames and other structures - the only difference between walls and RWalls being that they take longer to break due to more health.

<details>

</details>

# Changelog
:cl:
balance: Xenos no longer require boiler acid to melt reinforced walls.
balance: Xenos no longer require t3 claws to make holes in reinforced walls.
qol: Xenos no longer require boiler acid to melt reinforced window frames.
/:cl:

### DIFF
--- a/code/game/turfs/walls/r_wall.dm
+++ b/code/game/turfs/walls/r_wall.dm
@@ -10,7 +10,7 @@
 
 	walltype = WALL_REINFORCED
 
-	claws_minimum = CLAW_TYPE_VERY_SHARP
+	claws_minimum = CLAW_TYPE_SHARP
 
 /turf/closed/wall/r_wall/attackby(obj/item/W, mob/user)
 	if(hull)
@@ -179,7 +179,7 @@
 	if(hull)
 		return 0
 	else
-		return 2
+		return 1
 
 
 //Just different looking wall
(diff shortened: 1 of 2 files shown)


ihatethisengine:
tbh I'd make rwalls meltable but not slashable by non-t3s. Walls should be somewhat useful, they have their downsides. Slashing time isn't really important because you can stack as many xenos as you want and remove every wall in 5 seconds.

GoldenDarkness55:
> tbh I'd make rwalls meltable but not slashable by non-t3s. Walls should be somewhat useful, they have their downsides. Slashing time isn't really important because you can stack as many xenos as you want and remove every wall in 5 seconds.

That'd prevent lurkers from opening them though. If xenos stack like that it's probably for fob sieging which is fine imo, there's still the timed action to actually rip the wall open.

Diegoflores31:
this will make any surv gameplay like 100% unsurvivable outside of LZ camping

PR #4011, opened 2023-07-26
fixes xeno wound icon delay

realkhad:
# About the pull request

when xenos were getting back up from a stun/rest their bloody wounds would briefly disappear until the next time their wound icon got updated
now the wound icon is updated right as soon as they get up

a heavily wounded xeno getting up from a stun and appearing full health for a bit shouldn't be a thing

https://github.com/cmss13-devs/cmss13/assets/54692343/233604be-bc88-427e-9397-4059bba5c01a

# Changelog
:cl:Khadd
fix: xenos now update their wound icon when getting up
/:cl:

### DIFF
--- a/code/modules/mob/mob_status_procs.dm
+++ b/code/modules/mob/mob_status_procs.dm
@@ -94,9 +94,6 @@
 	handle_regular_status_updates(FALSE)
 	update_canmove()
 	knocked_down_timer = null
-	if(isxeno(src) && !resting) // !resting because we dont wanna prematurely update wounds if they're just trying to rest
-		var/mob/living/carbon/xenomorph/xeno = src
-		xeno.update_wounds()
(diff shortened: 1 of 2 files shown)


code/modules/mob/mob_status_procs.dm
```diff
@@ -94,6 +94,9 @@
 	handle_regular_status_updates(FALSE)
 	update_canmove()
 	knocked_down_timer = null
+	if(isxeno(src) && !resting) // !resting because we dont wanna prematurely update wounds if they're just trying to rest
```
morrowwolf, line 97: This should really be overridden for xenos only

morrowwolf (changes requested):

harryob (approved):

PR #4014, opened 2023-07-26
Replaces heavy revolver knockback with screenshake

GoldenDarkness55:
# About the pull request

Replaces heavy revolver knockback with heavy screenshake that extends to t3 and t4, maintains 4 tiles as max range for this effect. See video section for demonstration.

I trust I don't have to explain why the ability to permastun t1s and t2s in a secondary gun you can easily conceal is problematic. It's questionably balanced as a specialist's (scout) primary gun, let alone on a revolver.

Replaced it with a heavy screenshake effect, effects of which demonstrated in the video section - completely throws off clicks making xenos unable to effectively fight back or use abilities (ravs charging sideways, queen 1v1 hype) without leaving them entirely helpless to escape. Unique, balanced effect that keeps the gun's niche = good.

Might've been better to use a new var to set the range, but it doesn't seem to affect much as far as targets within screen range are concerned + makes sense given it's heavier ammo and fits with marksman ammo extending the range aswell.

<details>

https://streamable.com/2le3t9

</details>

# Changelog
:cl:
balance: Replaced heavy revolver knockback with screenshake
balance: Heavy revolver accurate range set to 4 tiles
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -570,6 +570,7 @@
 	name = "heavy revolver bullet"
 
 	damage = 35
+	accurate_range = 4
 	penetration = ARMOR_PENETRATION_TIER_4
 	accuracy = HIT_ACCURACY_TIER_3
 
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -573,8 +573,10 @@
 	penetration = ARMOR_PENETRATION_TIER_4
 	accuracy = HIT_ACCURACY_TIER_3
 
-/datum/ammo/bullet/revolver/heavy/on_hit_mob(mob/M, obj/item/projectile/P)
-	knockback(M, P, 4)
+/datum/ammo/bullet/revolver/heavy/on_hit_mob(mob/M, obj/item/projectile/fired_projectile)
+	if(fired_projectile.distance_travelled > accurate_range)
+		return
+	shake_camera(M, 6, 4)
 
 /datum/ammo/bullet/revolver/incendiary
 	name = "incendiary revolver bullet"


Greg-Joffer:
PR bad. sure you remove stun but you go ahead and make the game unplayable with the screenshake.

morrowwolf (commented):
This is getting a hard "meh" from me. Slowing down fire rate when using heavies or moving around stats would be better than creating an extraordinarily annoying effect that effects even the largest xenos.

GoldenDarkness55:
> This is getting a hard "meh" from me. Slowing down fire rate when using heavies or moving around stats would be better than creating an extraordinarily annoying effect that effects even the largest xenos.

True, it's not ideal.
Slowing down RoF based on ammo type feels a bit snowflake'y. I guess I could make revolver slower in general (same as slug?), but increase damage stats across all ammo to reach similar DPS. Maybe buff it to the level of other pistols while im at it since iirc its worse at least as far as normal ammo is concerned. That works?

morrowwolf:
> > This is getting a hard "meh" from me. Slowing down fire rate when using heavies or moving around stats would be better than creating an extraordinarily annoying effect that effects even the largest xenos.
> 
> True, it's not ideal. Slowing down RoF based on ammo type feels a bit snowflake'y. I guess I could make revolver slower in general (same as slug?), but increase damage stats across all ammo to reach similar DPS. Maybe buff normal and marksman ammo to around mod88 DPS levels while im at it, since iirc its far worse than other pistols. That works?

Keep the scope clean: you want heavy rounds to not be able to hard stun a xeno. How do we get there? Lower fire rate of revolvers using heavy ammo OR lower the effects of specifically heavy rounds. These are easy changes. 

I even made it so you can easily overwrite a proc to change knockback effects for ammo types. 
/datum/ammo/proc/knockback_effects()

GoldenDarkness55:
> Keep the scope clean: you want heavy rounds to not be able to hard stun a xeno. How do we get there? Lower fire rate of revolvers using heavy ammo OR lower the effects of specifically heavy rounds. These are easy changes.

Im saying I could expand the scope to fix other existing issue and to not add a weird snowflake modifier that can be avoided.
Messing with the knockback is subpar since the issue persists even with minimal stuns, and if it doesn't stun then it becomes borderline useless.
As for RoF while that'd work, I've no idea how to assign specific firerate to bullets and I'd wager it gets even trickier because of how the revolver handles magazines.

GoldenDarkness55:
If tweaking the revo stats and buffing it up to par isn't acceptable and since the ammo RoF won't code itself, how about just removing the heavy rounds entirely? Considering it'd just be "meh" with lower RoF and remain frustrating due to smartscope iff, I'm not really sure it's a feature that's worth keeping in the marine arsenal.
Could also try removing cc but raising its damage and giving it sharp falloff beyond 3 tiles. @morrowwolf

ihatethisengine:
Just remove it for real, chainstun is extremely cheap and cheesy.

Wintermote:
If this is added can you include a shaking effect to the xeno sprite affected by the screenshake so there is visual confirmation to the player shooting that their bullets are doing what they want them to do?

PR #4015, opened 2023-07-27
Eggsac no longer able to plant eggs outside hive weeds

GoldenDarkness55:
# About the pull request

Re-adds the weed restriction for egg planting to eggsac carrier.

It was removed for a VERY good reason that's more than apparent immediately after this was merged. An egg behind every corner has a massive impact on balance with marines no longer being able to chase any xeno without an instant round end, to say nothing about any pushback ability and especially warrior lunge being an instant round ender now. Even without that simply opening the egg, waiting several seconds and opening the door will cause the hugger to leap several tiles.
It was absurdly overpowered, which is why it was removed in the first place. It is even more so now, given the shorter gestation period making this the easiest and best way to perma marines xenos ever had.

Stop bringing back old issues.

<details>

</details>

# Changelog
:cl:
balance: Makes eggsac carrier unable to plant eggs outside of hive weeds again.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/egg_item.dm
+++ b/code/modules/mob/living/carbon/xenomorph/egg_item.dm
@@ -98,7 +98,7 @@
 		to_chat(user, SPAN_XENOWARNING("[src] must be planted on [lowertext(hive.prefix)]weeds."))
 		return
 
-	if(!hive_weeds && user.mutation_type != CARRIER_EGGSAC)
+	if(!hive_weeds)
 		to_chat(user, SPAN_XENOWARNING("[src] can only be planted on [lowertext(hive.prefix)]hive weeds."))
 		return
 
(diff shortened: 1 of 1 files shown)


CGosling:
I disagree with the reasons here as they're almost no different than hugger traps,  beyond the one tile activation proc.

They were not removed for balance either, it was simply part of the techweb overhaul about special structures and weeds.

morrowwolf:
>massive impact on balance

True, but also, not relevant remotely.

>simply opening the egg, waiting several seconds and opening the door will cause the hugger to leap several tiles

This is an underlying issue, fix this rather than nerfing availability of eggs.


I'm OK with eggsac carriers getting this for now. We'll revisit this in a couple weeks.

GoldenDarkness55:
> This is an underlying issue, fix this rather than nerfing availability of eggs.

Its not an issue by itself, since it requires setup and forces people to be wary when pushing the actual hive. It only becomes one when coupled with this massive egg overabundance all over the map, which is an issue by itself.

> We'll revisit this in a couple weeks.

Should probably merge them with normal carriers aswell at that time since there's no point to playing base now, assuming they ain't gonna be outright deleted.

PR #4023, opened 2023-07-27
Sentinel buffs

GoldenDarkness55:
# About the pull request

Follows-up on my old sentinel rework and buffs it into a more adequate state as it's severely underperforming.

Having been the one to rework them, while I don’t regret it and think they’re in a far better place now than with the old braindead 1 ability stunlock kit, I believe they’re underperforming as t1s

The initial design goal was to turn the lil annoying bodyblocker into an offensive support caste with strong abilities for a t1 - but unable to make good use of them by itself, thus heavily reliant on teamwork.
In practice they became

- too reliant on others

they struggle in 1v1 situations far more than any other t1 caste

- often unable to make an effective/satisfactory use of their kit

Their entire kit is about supporting and working with other caste, which requires them walking out with bulkier t3s and leads to them eating a ton of bullets that they’re ill-suited to tank. Compounded with said teamwork often being unreliable (either because there were too many bullets and the sent had to fall back unable to follow the t3, or the players just didn’t mesh very well etc) and their slow healing it can be frustrating to play when on such a bad streak of achieving nothing and having to rest.
Therefore the "hibernate" ability, to mitigate the downtime and better allow it to actually support the tankier frontline xenos without having to take ages to heal inbetween.
Also considering sentinels are supposed to guard the hive I think it’s cool and fitting to have an ability centered on hibernating somewhere.

Speed increase to let and encourage them to support backliners aswell as make it easier to work near the front.

Tackle buffs - addresses the paralyzing slash interaction - currently the window to tackle successfully is extremely tight, you’d have to start tackling the moment the marine falls down. This results in average player being entirely unable to chain these two abilities properly and even in the cases where he played good and should be winning a 1v1, the marine still comes out on top because the sent didn’t play perfectly - balance for 1% bad.

Rest I believe to be self explanatory.

<details>

</details>

# Changelog
:cl:
add: Added a new ability to sentinel - "Hibernate" - causes them to enter a 10 second sleep that restores their health and plasma to full. Requires weeds. 50 second cd and 150 plasma cost.
balance: Sentinel neuro spit cooldown raised 1.5 to 2s, superslow 2.5 to 5s
balance: Sentinel scatter spit scatter lowered 60 to 45, cooldown lowered 8s to 5s, stun time increased 0.7s to 0.8s and added a 1.5s slow
balance: Sentinel paralyzing slash stuns after 3 seconds instead of 4.
balance: Sentinel speed raised from -0.8 to -1
balance: Sentinel tackle now requires 3 attempts instead of 4.
/:cl:

### DIFF
--- a/code/__DEFINES/weapon_stats.dm
+++ b/code/__DEFINES/weapon_stats.dm
@@ -66,7 +66,7 @@ var/accuracy_mult_unwielded
 ////SCATTER////
 */
 
-#define SCATTER_AMOUNT_NEURO 60
+#define SCATTER_AMOUNT_NEURO 45
 #define SCATTER_AMOUNT_TIER_1 15
 #define SCATTER_AMOUNT_TIER_2 10
 #define SCATTER_AMOUNT_TIER_3 8
--- a/code/modules/mob/living/carbon/xenomorph/abilities/sentinel/sentinel_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/sentinel/sentinel_abilities.dm
@@ -6,7 +6,7 @@
 	macro_path = /datum/action/xeno_action/verb/verb_slowing_spit
 	action_type = XENO_ACTION_CLICK
 	ability_primacy = XENO_PRIMARY_ACTION_1
-	xeno_cooldown = 1.5 SECONDS
+	xeno_cooldown = 2 SECONDS
 	plasma_cost = 20
 
 // Scatterspit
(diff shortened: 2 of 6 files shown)


private-tristan:
wouldn't hibernate make sentinels unable to hear hivemind/word of queen?

GoldenDarkness55:
Seems it'll go stale and autolock. Whoever wants should feel free to revive if that happens. If the buff won't bring it up to par, might be worthwhile to test changing shotgun neuro to root instead of KD and give the spit a short non-stacking KD on a somewhat short cd + swap superslow to slow on low duration.

> wouldn't hibernate make sentinels unable to hear hivemind/word of queen?
It would, yes. It's probably an easy fix for proper coders, but I've no idea how to work around that.

Butlerblock:
> It would, yes. It's probably an easy fix for proper coders, but I've no idea how to work around that.

You could make it a paralyze(or whatever variable cm uses for stuns) as well as a mute(not sure if this exists for xenos) instead of a sleep. While this will allow them to have vision around them, it would indicate better to xenos around them that the sentinel is hibernating and needs moving if in danger.

morrowwolf:
Not feeling this one.

PR #4038, opened 2023-07-29
Squad Leaders Can Pick Between B12 and M4 Armour

Steelpoint:
# About the pull request

Squad Leaders can now chose between picking either a set of B12 or M4 armour, whereas currently they can only pick a set of B12 armour with the option to buy a set of M4 armour for 30 points.

SLs can no longer buy M4 armour. 

Functionally the difference between M4 and B12 armour is a small boost in defence versus one extra inventory space. This change will increase loadout flexibility for Squad Leaders between those wanting a little more defence versus those wanting a bit more inventory space. 

B12 armour is no longer sufficiently unique enough that its presence is an indication of someone being a valuable target as now many roles can easily acquire the armour. Ergo a SL picking between either set of armour will not render themselves any more or less identifiable as a Squad Leader to xenomorphs. 

It also felt egregious that Team Leaders could get M4 armour for free but SLs could not. 

<details>

</details>

# Changelog

:cl:
balance: Squad Leaders may now choose between selecting their standard issue B12 armour, or a set of M4 armour. They can only select one. 
balance: Squad Leaders can no longer buy a set of M4 armour per this change.
/:cl:


morrowwolf:
Jobs should be easily identifiable by their gear and look based on their default options. Make it a buyable option sure, otherwise no.

PR #4041, opened 2023-07-29
Unnested Larva Maturation Prompts

Drulikar:
# About the pull request

This PR adds 10s tgui_alerts asking an observer to take over a maturing larva (an embryo that reached stage 5 in a host) if the host is unnested. They will automatically orbit the bursting mob when prompted. It will make 5 attempts at this. If the host is nested, the process will proceed as it has without any prompts.

This PR also fixes the pool flash not occurring for bursts from hosts.

For the pool flash, that should have just been occurring in the first place.

For the prompts:
- This allows observers to opt out of accepting a larva spawn in what they deem an impossible scenario - or at least a scenario they don't want to risk their place in queue for.
- This now makes up to 5 attempts from the queue (technically 6 if you count a player hugger 2nd dibs) to put a player into a bursting larva rather than a single attempt. A side effect to this though is it can potentially increase the time to autoburst up to one minute while we're querying candidates.
- This ensures that the new larva has an active player that didn't step away a moment ago (the queue allows a 5 minute window for inactivity)

<details>

</details>

# Changelog
:cl: Drathek
add: Added prompts for xeno candidates to optionally take over a larva when an unnested host is ready to burst
fix: Fixed larva spawning in hosts not triggering the preference unpool taskbar flash.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Embryo.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Embryo.dm
@@ -260,6 +260,8 @@
 
 		if(new_xeno.client)
 			new_xeno.client.change_view(world_view_size)
+			if(new_xeno.client.prefs?.toggles_flashing & FLASH_POOLSPAWN)
+				window_flash(new_xeno.client)
 
 		SSround_recording.recorder.track_player(new_xeno)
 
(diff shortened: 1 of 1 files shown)


cm13live:
This pull request has been mentioned on **CM-SS13**. Please provide feedback here:

https://forum.cm-ss13.com/t/rework-larva-queue/3050/73

ihatethisengine:
Have you considered adding the pop-up only for unnested host spawns? It's kinda excessive for nested, because even if you are away for a minute in the hive, it usually doesn't change anything. And even if you away for longer, the game will ghost you automatically.

harryob (approved):

PR #4045, opened 2023-07-29
Lesser Drone now Has 3 Guarantied Slots

Ben10083:
# About the pull request

Lesser Drones now will always have at least 3 slots, no matter the state of the hive

Lesser Drones as supposed to be a swarm, and at this stage of the game many Xenos are likely ghosting. This gives them a bigger chance to participate in rounds (at least the ones where Xenos are on the backfoot). This should make it so lesser drones have a baseline of 3 slots, plus additional limit based on hive size.

# Changelog
:cl:
add: Lesser Drone Slots now have a minimum, currently set to 3.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
+++ b/code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
@@ -1050,7 +1050,7 @@
 	hugger.timeofdeath = user.timeofdeath // Keep old death time
 
 /datum/hive_status/proc/update_lesser_drone_limit()
-	lesser_drone_limit = max(Ceiling(length(totalXenos) / 3), lesser_drone_minimum)
+	lesser_drone_limit = lesser_drone_minimum + Ceiling(length(totalXenos) / 3)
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -1052,7 +1054,7 @@
 	hugger.timeofdeath = user.timeofdeath // Keep old death time
 
 /datum/hive_status/proc/update_lesser_drone_limit()
-	lesser_drone_limit = Ceiling(totalXenos.len / 3)
+	lesser_drone_limit = lesser_drone_minimum + Ceiling(length(totalXenos) / 3)
```
Zonespace27, line 1057: This math is bad and will lead to hives having +3 drones at any size, this needs to be like so (mobile, so shit formatting)

max(minimum drones, ceiling of total xenoes / 3)

Zonespace27 (changes requested):

PR #4051, opened 2023-07-29
Fix corpses merging with weeds when hung

### LINKED ISSUE #4048: Hanged Corpses Shifting After Weeds Consume Them
### Testmerges

#3985, #4003, #4024, #3726, #4035

### Description of the bug

Predators hanging corpses causes them to shift as though they were in space. The problem is that once they go perma and the weeds 'consume' them, their bodies still shift while covered with weeds. Although hilarious, it does break suspension of disbelief.

Corpses should not shift under weeds like an ocean wave pushes a coconut on a beach.

1.Have a corpse shifting, likely from a predator hanging the corpse.
2.Place weeds on the corpse.
3.Wait for the weeds to cover the corpse and not stop the shifting.
4. Smile and have a good day.

### Issue Bingo

Drulikar:
# About the pull request

This PR fixes any scenario where COMPONENT_CANCEL_MOB_PULL is signaled, namely predator hanging, to prevent merging with weeds.

The chain of do_afters for hanging now checks if their victim gets anchored (in case they started to hang right as weeds were merging them), and cutting down will now call the Moved proc to send signals indicating the mob moved (to wake up weed_food).

Fixes #4048 

<details>

</details>

# Changelog
:cl: Drathek
fix: Fixed hung mobs merging with weeds
/:cl:

### DIFF
--- a/code/datums/components/weed_food.dm
+++ b/code/datums/components/weed_food.dm
@@ -193,6 +193,8 @@
 			return FALSE
 	if(!parent_turf?.weeds)
 		return FALSE
+	if(SEND_SIGNAL(parent_mob, COMSIG_ATTEMPT_MOB_PULL) & COMPONENT_CANCEL_MOB_PULL)
+		return FALSE
 
 	if(unmerged_time == world.time)
 		return merge_with_weeds() // Weeds upgraded, re-merge now re-using the apperance
@@ -244,6 +246,9 @@
 	if(parent_buckle)
 		UnregisterSignal(parent_buckle, COSMIG_OBJ_AFTER_BUCKLE)
 		parent_buckle = null
+
+	if(SEND_SIGNAL(parent_mob, COMSIG_ATTEMPT_MOB_PULL) & COMPONENT_CANCEL_MOB_PULL)
+		return FALSE
 
 	absorbing_weeds = parent_turf?.weeds
 	if(!absorbing_weeds)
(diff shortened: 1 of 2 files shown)


code/datums/components/weed_food.dm
```diff
@@ -193,6 +193,8 @@
 			return FALSE
 	if(!parent_turf?.weeds)
 		return FALSE
+	if(SEND_SIGNAL(parent_mob, COMSIG_ATTEMPT_MOB_PULL) & COMPONENT_CANCEL_MOB_PULL)
```
harryob, line 196: signals should generally have one point of origin ( https://hackmd.io/@tgstation/SignalsComponentsElements#Signals-How-they-work ), why not just send additional signals here that're registered in /proc/get_hung()?

harryob (changes requested):

code/datums/components/weed_food.dm
Drulikar: Look again at how that signal is used, its usage is to test whether the item is pullable; not an event that something occured like most signals like you're describing.

Drulikar: Basically this is an abstract way asking any components that have registered COMSIG_ATTEMPT_MOB_PULL to deny the request. Otherwise pred hanging needs to remove this signal registration, and guard against its mobs in some other way.

harryob: i don’t really like the pattern but it’s fine you didn’t start it

harryob (approved):

PR #4053, opened 2023-07-30
More Xeno Alliances Tweaks

ihatethisengine:
# About the pull request
1) xeno_message no longer sends the message to allied hives.

2) All hive's alliances break on Queen's death (instead of essentially just ceasing to function).

3) If another hive breaks alliance with you, you break alliance with them automatically.

4) Some code improvements.

1) Less confusing. I meant to add a "notify_allies" var to the proc, but I haven't found a single message that should be sent to allies.

2) Less confusing. Essentially all alliances already don't function if queen is dead, so this way it's just more clear for everyone.

3) You don't have to keep hive alliance status open all the time in case allied queen decides to betray you. Simply a QoL.

4) Just good

<details>

</details>

# Changelog
:cl: ihatethisengine
add: All hive's alliances break on Queen's death.
qol: If a hive breaks an alliance with another, the second hive also breaks the alliance.
fix: xeno_message no longer sends the message to allied hives
/:cl:


harryob (changes requested):
string nitpick

BeagleGaming1 (approved):
Code seems good

PR #4055, opened 2023-07-30
New Healer Drone Ability: Sacrifice

ihatethisengine:
# About the pull request

New healer drone skill — sacrifice. If things get really bad, the healer can instantly transfer 75% of its current health and remove all debuffs from the target, but for the great cost of the healer’s own life.

The sacrifice ability is  an emergency button. You don’t have to use it, but at some point you may find that sacrifice is a right thing to do, so it’s a nice option to have. I want to emphasize that this skill is not a necessary part of the healer gameplay. You don’t have to trade your life for the life of another xeno. But you can if you are willing to.

Also healer drones gets a counter of the health it has transferred. If healer drone transfers at least 10 000 health points, it will be able to respawn after using the sacrifice ability (like acid runner after kaboom).

Saving fellow xenos is the most satisfying part of healer drone’s gameplay. This ability allows to save others in really dire situations, but for the great cost. It’s still fitting, because drone already can use heal ability until it goes into crit or dies, in some situations it’s a worthy risk. Also rewards long living healer drones with the opportunity to use the ability essentially for free.

I have my doubts about this ability, not going to lie, but I'll leave it to your judgement if it's a good addition to the game.

<details>

</details>

# Changelog
:cl: ihatethisengine
balance: New healer drone ability — sacrifice. Healer drone can instantly transfer 75% of its current health and remove all debuffs from the target, but for the great cost of the healer’s own life.
/:cl:


cm13live:
This pull request has been mentioned on **CM-SS13**. Please provide feedback here:

https://forum.cm-ss13.com/t/new-healer-drone-ability-sacrifice/3359/11

code/modules/mob/living/carbon/xenomorph/mutators/strains/drone/healer.dm
```diff
@@ -148,6 +151,9 @@
 	var/salve_applied_recently = FALSE
 	var/mutable_appearance/salve_applied_icon
 
+	var/trasnferred_amount = 0
```
harryob, line 154: transferred ?

code/modules/mob/living/carbon/xenomorph/mutators/strains/drone/healer.dm
```diff
@@ -173,3 +179,118 @@
...
+	var/max_range = 1
+	var/transfer_mod = 0.75 // only transfers 75% of current healer's health
+	macro_path = /datum/action/xeno_action/verb/verb_healer_sacrifice
+	action_type = XENO_ACTION_CLICK
+	ability_primacy = XENO_PRIMARY_ACTION_5
+
+/datum/action/xeno_action/verb/verb_healer_sacrifice()
+	set category = "Alien"
+	set name = "Sacrifice"
+	set hidden = 1
```
harryob, line 224:
```suggestion
	set hidden = TRUE
```

code/modules/mob/living/carbon/xenomorph/mutators/strains/drone/healer.dm
```diff
@@ -173,3 +179,118 @@
...
+
+	if(!xeno.check_state())
+		return
+
+	if(target.stat == DEAD)
+		to_chat(xeno, SPAN_WARNING("[target] is already dead!"))
+		return
+
+	if(target.health >= target.maxHealth)
+		to_chat(xeno, SPAN_WARNING("\The [target] is already at max health!"))
```
harryob, line 251:
```suggestion
		to_chat(xeno, SPAN_WARNING("[target] is already at max health!"))
```

harryob (changes requested):

code/modules/mob/living/carbon/xenomorph/mutators/strains/drone/healer.dm
ihatethisengine: Oh my god how did this happen, I'll fix the vars after I get some sleep

ihatethisengine:
As a solution to the potential issue of the healer drone rendering the SADAR AP rocket useless, I think we can just remove 1 second of the stun instead of removing it completely.

code/modules/mob/living/carbon/xenomorph/mutators/strains/drone/healer.dm
```diff
@@ -173,3 +179,118 @@
...
+/datum/behavior_delegate/drone_healer/proc/modify_transferred(amount)
+	transferred_amount += amount
+	if(transferred_amount > max_transferred_amount)
+		transferred_amount = max_transferred_amount
+	if(transferred_amount < 0)
+		transferred_amount = 0
+
+/datum/behavior_delegate/drone_healer/append_to_stat()
+	. = list()
+	. += "Trasnferred health amount: [transferred_amount]"
```
BeagleGaming1, line 196: You made the typo here too

ihatethisengine:
Apparently xeno-primary-action-five is already taken by the special structure ability. So right now there is no hotkey for the sacrifice ability (you still can macro it but oh well). Not sure what can I do about it. I could remove lesser resin fruits because they are quite useless (and I meant to remove them in my original PR on gitlab) and were added just because healer didn't have enough abilities. But it would be a nerf in a way :(

Need your advice lads.

ihatethisengine:
Apparently xeno-primary-action-five is taken by the special structure ability. That means the sacrifice ability doesn't have a hotkey right now (you still can macro it but eh). Don't know what to do with this. Can remove lesser fruits because they are very weak anyway, but that would be a nerf in a way. Not sure what to do.

ihatethisengine:
This actually might be a bug, because I don't see any references in code that the order construction ability should work as XENO_PRIMARY_ACTION_5.

ihatethisengine:
Ok, I found a solution. Now healer just uses the version of order construction that isn't set to primary action 5 (just like queen's).

code/modules/mob/living/carbon/xenomorph/mutators/strains/drone/healer.dm
```diff
@@ -173,3 +183,122 @@
...
+/datum/behavior_delegate/drone_healer/on_life()
+	if(!bound_xeno)
+		return
+	if(bound_xeno.stat == DEAD)
+		return
+	var/image/holder = bound_xeno.hud_list[PLASMA_HUD]
+	holder.overlays.Cut()
+	var/percentage_transferred = round((transferred_amount / max_transferred_amount) * 100, 10)
+	if(percentage_transferred)
+		holder.overlays += image('icons/mob/hud/hud.dmi', "xenoenergy[percentage_transferred]")
```
ihatethisengine, line 213: Thinking about moving this into a separate proc that triggers on modify_transferred(). But not sure if that would be better, because you can heal many times in two seconds and updating HUD every time sounds like a waste + inconsistent.

ihatethisengine:
Also thinking 10 000 requirement might be even too high. But on the other hand, the ability is quite strong, so not sure.

morrowwolf (changes requested):
A tier 1 with such a strong heal and stun removal is too powerful even for the life of the player. Plus, I don't believe this is really telegraphed hard enough. I appreciate your work but all in all I don't think this would be good.

PR #4087, opened 2023-08-02
Increases the availability and minimum burn time for Star Shells

VileBeggar:
# About the pull request

raises the minimum burn time for star shell fragments to 30 seconds from 5 and increases their quantity in squad vendors from scale * 2 to scale * 4.

these grenades are fairly innocuous but rarely, if ever, used outside the free packet you get with the m79. their short burn time coupled with the fact you are giving up space that could be used for actual grenades that have a lot more utility place them into the "not really worth it" category.

hopefully with a higher minimum burn time and a bit more abundance in the squad vendor they'll see some more use, whether in an ugl or the m79.

# Changelog
:cl:Vile Beggar
balance: Star Shells now have a higher minimum burn time and are more plentiful in squad prep.
/:cl:


morrowwolf (approved):

harryob (approved):
if this somehow gets annoying i will revert it

PR #4104, opened 2023-08-03
Combistick now uses a chain

Zonespace27:
# About the pull request
image

Pred combistick now uses a visible chain.

Balance ramifications:

- Someone picking up the combi (still has a do_after for non-preds mind you) will sever the chain with a message to the pred
- You can no longer yank any combi-stick regardless of if it's actually yours
- Combistick will be pulled with you when you would get more than 6 tiles from it
- You will automatically yank back a combistick that tries to get put into a container

- Preventing chain weirdness while still allowing it to be recovered by marines if the pred really isn't good about holding onto the combi
- Side effect of me making the code a lot better
- Side effect of the tether effect
- Preventing chain weirdness

# Changelog
:cl:
add: Combisticks now use a proper chain instead of an invisible magic one.
/:cl:

### DIFF
--- a/code/modules/cm_preds/yaut_bracers.dm
+++ b/code/modules/cm_preds/yaut_bracers.dm
@@ -955,19 +955,10 @@
 	if(.)
 		return
 
-	for(var/obj/item/weapon/yautja/combistick/C in range(7))
-		if(C in caller.contents) //Can't yank if they are wearing it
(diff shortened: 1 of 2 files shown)


realforest2001 (approved):
Approved on the WL end.

harryob (approved):

Issue #1450, opened 2022-11-06
You can fill ammobelts weapon slot if using a dump method


fira:
## Difference between expected and actual behavior

Dumping should honor the weapon slot restriction.

## Steps to reproduce

The Flaregun belt has 16 slots for flares and one for a flaregun.
If you dump fill it from flareboxes, it will fill it with 17 flares. 
Upon taking a flare out, you can't put it back in.

## Issue bingo (not required to be all checked)

- [x] Issue could be reproduced at least once
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Huffie56:
that PR fixed a similar issue?
https://github.com/cmss13-devs/cmss13/pull/580

PR #4110, opened 2023-08-05
fix prevent belt flare being filled by dump in action leaving no space for flare gun.

Huffie56:
# About the pull request
fixes: https://github.com/cmss13-devs/cmss13/issues/1450

<details>

</details>

# Changelog
:cl:
fix: prevent belt flare being filled by dump in action leaving no space for flare gun.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -17,7 +17,8 @@
 	var/flap = TRUE
 
 /obj/item/storage/belt/gun/dump_into(obj/item/storage/M, mob/user)
-	var/holster_gun = FALSE
+	//var/holster_gun = FALSE
+	/*
 	for(var/slot in holster_slots)
 		if(!holster_slots[slot]["gun"]) //Open holster.
 			holster_gun = FALSE
@@ -25,15 +26,16 @@
 		if(holster_slots[slot]["gun"]) //Gun in holster.
 			holster_gun = TRUE
 			break
-
-	if(!holster_gun && length(contents) >= (storage_slots-1))
+	*/
+	if(length(holstered_guns) < 1 && length(contents) >= (storage_slots-1))
 
 		to_chat(user, SPAN_WARNING("[src] is full."))
 		return FALSE
 	return ..()
 
 /obj/item/storage/belt/gun/handle_item_insertion(obj/item/W, prevent_warning = 0, mob/user)
-	var/holster_gun = FALSE
+	//var/holster_gun = FALSE
+	/*
 	for(var/slot in holster_slots)
 		if(!holster_slots[slot]["gun"]) //Open holster.
 			holster_gun = FALSE
@@ -41,8 +43,8 @@
 		if(holster_slots[slot]["gun"]) //gun in holster.
 			holster_gun = TRUE
 			break
-
-	if(W.type == /obj/item/device/flashlight/flare && holster_gun == FALSE && contents.len >= (storage_slots-1))
+	*/
+	if(W.type == /obj/item/device/flashlight/flare && length(holstered_guns) < 1 && contents.len >= (storage_slots-1))
 		return FALSE
 	return ..()
 
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -30,6 +30,12 @@
 		return FALSE
 	return ..()
 
+/obj/item/storage/belt/gun/flaregun/has_room(obj/item/W as obj, W_class_override = null)
+
+	if(istype(new_item, /obj/item/device/flashlight/flare) && length(holstered_guns) > 0 && length(contents) >= storage_slots)
+		return FALSE
+	return ..()
+
 /obj/item/storage/belt/equipped(mob/user, slot)
 	switch(slot)
 		if(WEAR_WAIST, WEAR_J_STORE, WEAR_BACK)
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -16,7 +16,7 @@
 	///TRUE Means that it closes a flap over its contents, and therefore update_icon should lift that flap when opened. If it doesn't have _half and _full iconstates, this doesn't matter either way.
 	var/flap = TRUE
 
-/obj/item/storage/belt/gun/flaregun/dump_into(obj/item/storage/Main, mob/user)
+/obj/item/storage/belt/gun/flaregun/dump_into(obj/item/storage/origin_storage, mob/user)
 
 	if(length(holstered_guns) < 1 && length(contents) >= (storage_slots-1))
 
@@ -24,9 +24,9 @@
 		return FALSE
 	return ..()
 
-/obj/item/storage/belt/gun/flaregun/handle_item_insertion(obj/item/Warn, prevent_warning = 0, mob/user)
+/obj/item/storage/belt/gun/flaregun/handle_item_insertion(obj/item/new_item, prevent_warning = FALSE, mob/user)
 
-	if(Warn.type == /obj/item/device/flashlight/flare && length(holstered_guns) < 1 && contents.len >= (storage_slots-1))
+	if(istype(new_item, /obj/item/device/flashlight/flare) && length(holstered_guns) < 1 && length(contents) >= (storage_slots-1))
 		return FALSE
 	return ..()
 
@@ -767,7 +767,7 @@
 	for(var/i = 1 to storage_slots)
 		new /obj/item/weapon/throwing_knife(src)
 
-/obj/item/storage/belt/knifepouch/_item_insertion(obj/item/W, prevent_warning = 0)
+/obj/item/storage/belt/knifepouch/_item_insertion(obj/item/W, prevent_warning = FALSE)
 	..()
 	playsound(src, 'sound/weapons/gun_shotgun_shell_insert.ogg', 15, TRUE)
 
@@ -990,7 +990,7 @@
 			to_chat(usr, SPAN_WARNING("[src] can't hold any more ammo."))
 		return FALSE
 
-/obj/item/storage/belt/gun/_item_insertion(obj/item/W, prevent_warning = 0)
+/obj/item/storage/belt/gun/_item_insertion(obj/item/W, prevent_warning = FALSE)
 	if(isgun(W))
 		holstered_guns += W
 		for(var/slot in holster_slots)
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -15,6 +15,18 @@
 	cant_hold = list(/obj/item/weapon/throwing_knife)
 	///TRUE Means that it closes a flap over its contents, and therefore update_icon should lift that flap when opened. If it doesn't have _half and _full iconstates, this doesn't matter either way.
 	var/flap = TRUE
+	var/holster_gun = FALSE
+
+/obj/item/storage/belt/proc/dump_into(obj/item/storage/M, mob/user)
+	if(holster_gun == FALSE && contents.len >= (storage_slots-1))
+		to_chat(user, SPAN_WARNING("[src] is full."))
+		return FALSE
+	..()
+
+/obj/item/storage/belt/proc/handle_item_insertion(obj/item/W, prevent_warning = 0, mob/user)
+	if(holster_gun == FALSE && contents.len >= (storage_slots-1))
+		return FALSE
+	..()
 
 /obj/item/storage/belt/equipped(mob/user, slot)
 	switch(slot)
@@ -895,11 +907,14 @@
 
 /obj/item/storage/belt/gun/on_stored_atom_del(atom/movable/AM)
 	if(!isgun(AM))
+		holster_gun = FALSE
 		return
 	holstered_guns -= AM
+	holster_gun = TRUE
 	for(var/slot in holster_slots)
 		if(AM == holster_slots[slot]["gun"])
 			holster_slots[slot]["gun"] = null
+
 			update_gun_icon(slot)
 			return
 
@@ -1452,7 +1467,7 @@
 	name = "\improper M276 pattern M82F flare gun holster rig"
 	desc = "The M276 is the standard load-bearing equipment of the USCM. It consists of a modular belt with various clips. This version is for the M82F flare gun."
 	storage_slots = 17
-	storage_holster = TRUE
+	//storage_holster = TRUE
 	max_storage_space = 20
 	icon_state = "m82f_holster"
 	item_state = "s_marinebelt"
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -1452,6 +1452,7 @@
 	name = "\improper M276 pattern M82F flare gun holster rig"
 	desc = "The M276 is the standard load-bearing equipment of the USCM. It consists of a modular belt with various clips. This version is for the M82F flare gun."
 	storage_slots = 17
+	storage_holster = TRUE
 	max_storage_space = 20
 	icon_state = "m82f_holster"
 	item_state = "s_marinebelt"
@@ -1476,7 +1477,7 @@
 /obj/item/storage/belt/gun/flaregun/attackby(obj/item/W, mob/user)
 	if(istype(W, /obj/item/storage/box/m94))
 		var/obj/item/storage/box/m94/M = W
-		dump_ammo_into_belt_holster(M,user)
+		dump_into(M,user)
 	else
 		return ..()
 
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -16,7 +16,7 @@
 	///TRUE Means that it closes a flap over its contents, and therefore update_icon should lift that flap when opened. If it doesn't have _half and _full iconstates, this doesn't matter either way.
 	var/flap = TRUE
 
-/obj/item/storage/belt/gun/dump_into(obj/item/storage/Main, mob/user)
+/obj/item/storage/belt/gun/flaregun/dump_into(obj/item/storage/Main, mob/user)
 
 	if(length(holstered_guns) < 1 && length(contents) >= (storage_slots-1))
 
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -32,8 +32,8 @@
 
 /obj/item/storage/belt/gun/flaregun/has_room(obj/item/W as obj, W_class_override = null)
 
-	if(istype(new_item, /obj/item/device/flashlight/flare) && length(holstered_guns) > 0 && length(contents) >= storage_slots)
-		return FALSE
+	if(istype(new_item, /obj/item/device/flashlight/flare) && length(holstered_guns) < 1 && length(contents) >= (storage_slots-1))
+		return FALSE //No slot open because gun in holster.
 	return ..()
 
 /obj/item/storage/belt/equipped(mob/user, slot)
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -1476,7 +1476,9 @@
 /obj/item/storage/belt/gun/flaregun/attackby(obj/item/W, mob/user)
 	if(istype(W, /obj/item/storage/box/m94))
 		var/obj/item/storage/box/m94/M = W
+		storage_slots = 16
 		dump_into(M,user)
+		storage_slots = 17
 	else
 		return ..()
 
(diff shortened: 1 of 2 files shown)


Huffie56:
need help no idea why my new proc isn't working

code/game/objects/items/storage/storage.dm
```diff
@@ -721,6 +723,45 @@ W is always an item. stop_warning prevents messaging. user may be null.**/
 			to_chat(user, SPAN_WARNING("[ammo_dumping] is empty."))
 	return TRUE
```
Huffie56, line 725: this isn't working

Huffie56: @Drulikar

code/game/objects/items/storage/storage.dm
```diff
@@ -30,7 +30,7 @@
 	var/list/content_watchers //list of mobs currently seeing the storage's contents
 	var/storage_flags = STORAGE_FLAGS_DEFAULT
 	var/has_gamemode_skin = FALSE ///Whether to use map-variant skins.
-
+	var/storage_holster = FALSE
```
Drulikar, line 33: This is not appropriate here because `/obj/item/storage` doesn't have the concept of a holstered gun. That is a var and behavior for a `/obj/item/storage/belt/gun`.

Also it being a var is unnecessary because you can just have the proc `dump_into` overridden for either a `/obj/item/storage/belt/gun` or `/obj/item/storage/belt/gun/flaregun` to have the new behavior.

code/game/objects/items/storage/storage.dm
```diff
@@ -722,6 +722,10 @@ W is always an item. stop_warning prevents messaging. user may be null.**/
 	return TRUE
 
 /obj/item/storage/proc/dump_into(obj/item/storage/M, mob/user)
+	if(storage_holster == TRUE && contents.len >= (storage_slots-1))
```
Drulikar, line 725: This doesn't test if whether a gun is already holstered, and you'd only be able to check for this in either a `/obj/item/storage/belt/gun` or a `/obj/item/storage/belt/gun/flaregun` because it is `/obj/item/storage/belt/gun` that defines `var/list/holstered_guns`. In otherwords, currently if you have a gun holstered, then this proc will still leave an empty space for a gun that is already present.

Drulikar (changes requested):

code/game/objects/items/storage/storage.dm
Huffie56: i was thinking if i but it in /obj/item/storage/belt/gun the proc wouldn't be able to access it since it's in /obj/item/storage
also i wanted to create a new child for the proc if possible...

code/game/objects/items/storage/storage.dm
Huffie56: i see didn't think of that case. thanks that will avoid use to do another fix in the future...
need to activate my brain  for this.

Huffie56:
tried a bunch of option nothing work like i want.....

code/game/objects/items/storage/belt.dm
```diff
@@ -15,6 +15,18 @@
...
+/obj/item/storage/belt/proc/dump_into(obj/item/storage/M, mob/user)
+	if(holster_gun == FALSE && contents.len >= (storage_slots-1))
+		to_chat(user, SPAN_WARNING("[src] is full."))
+		return FALSE
+	..()
+
+/obj/item/storage/belt/proc/handle_item_insertion(obj/item/W, prevent_warning = 0, mob/user)
+	if(holster_gun == FALSE && contents.len >= (storage_slots-1))
+		return FALSE
+	..()
```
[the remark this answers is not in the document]
Huffie56: oh damn i really need to pay more attention to the how use proc lost two hours trying to go around this......

Huffie56:
it's working but i need to try to tight the code better....

code/game/objects/items/storage/belt.dm
```diff
@@ -15,6 +15,34 @@
...
+/obj/item/storage/belt/gun/dump_into(obj/item/storage/M, mob/user)
+	for(var/slot in holster_slots)
+		if(!holster_slots[slot]["gun"]) //Open holster.
+			holster_gun = FALSE
+			break
+		if(holster_slots[slot]["gun"]) //Gun in holster.
+			holster_gun = TRUE
+			break
+
+	if(holster_gun == FALSE && contents.len >= (storage_slots-1))
```
Drulikar, line 29:
```suggestion
	if(!holster_gun && length(contents) >= (storage_slots-1))
```

code/game/objects/items/storage/belt.dm
```diff
@@ -15,6 +15,34 @@
...
+		if(!holster_slots[slot]["gun"]) //Open holster.
+			holster_gun = FALSE
+			break
+		if(holster_slots[slot]["gun"]) //gun in holster.
+			holster_gun = TRUE
+			break
+
+	if(W.type == /obj/item/device/flashlight/flare && holster_gun == FALSE && contents.len >= (storage_slots-1))
+		return FALSE
+	..()
```
Drulikar, line 45:
```suggestion
	return ..()
```

code/game/objects/items/storage/belt.dm
```diff
@@ -1452,6 +1481,7 @@
 	name = "\improper M276 pattern M82F flare gun holster rig"
 	desc = "The M276 is the standard load-bearing equipment of the USCM. It consists of a modular belt with various clips. This version is for the M82F flare gun."
 	storage_slots = 17
+	//storage_holster = TRUE
```
Drulikar, line 1484:
```suggestion
```
Don't need this commented code.

Drulikar (changes requested):

code/game/objects/items/storage/belt.dm
```diff
@@ -15,6 +15,34 @@
...
+			holster_gun = TRUE
+			break
+
+	if(holster_gun == FALSE && contents.len >= (storage_slots-1))
+		to_chat(user, SPAN_WARNING("[src] is full."))
+		return FALSE
+	..()
+
+/obj/item/storage/belt/gun/handle_item_insertion(obj/item/W, prevent_warning = 0, mob/user)
+	for(var/slot in holster_slots)
```
Drulikar, line 35: Also is there any reason why you're doing it this way vs just checking the list of holstered guns like in the original suggestion on discord? Did you find the flaregun not get added to that list?

code/game/objects/items/storage/belt.dm
```diff
@@ -15,6 +15,34 @@
...
+		if(!holster_slots[slot]["gun"]) //Open holster.
+			holster_gun = FALSE
+			break
+		if(holster_slots[slot]["gun"]) //gun in holster.
+			holster_gun = TRUE
+			break
+
+	if(W.type == /obj/item/device/flashlight/flare && holster_gun == FALSE && contents.len >= (storage_slots-1))
```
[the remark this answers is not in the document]
Huffie56: it's working as intended the idea is to only allow flare to be inserted.

code/game/objects/items/storage/belt.dm
```diff
@@ -15,6 +15,34 @@
...
+		if(holster_slots[slot]["gun"]) //Gun in holster.
+			holster_gun = TRUE
+			break
+
+	if(holster_gun == FALSE && contents.len >= (storage_slots-1))
+		to_chat(user, SPAN_WARNING("[src] is full."))
+		return FALSE
+	..()
+
+/obj/item/storage/belt/gun/handle_item_insertion(obj/item/W, prevent_warning = 0, mob/user)
```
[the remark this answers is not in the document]
Huffie56: i mean i could change this random var name to another  longer var name why not.

code/game/objects/items/storage/belt.dm
Drulikar: Like in my original snippet, why not look at the list `holstered_guns` rather than `holster_slots`. `holster_slots` is for the inventory representation. `holstered_guns` is just a list of guns in the gun belt. If the issue is just that the flaregun doesn't get counted as a gun to be entered into holstered_guns, then thats likely more the problem (e.g. `/obj/item/storage/belt/gun/_item_insertion(obj/item/W, prevent_warning = 0)` needs to have an override for /obj/item/storage/belt/gun/**flare_gun**/_item_insertion that'll treat a flaregun as a gun to be inserted in that list. But anyways, I'd just try it out and see if that simplifies things for you or not.

code/game/objects/items/storage/belt.dm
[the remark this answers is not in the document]
Drulikar: Then if you want to be flaregun specific and not just implement this more generically for gun belts, then this proc should be flaregun specific.

Change both your overridden procs from `/obj/item/storage/belt/gun/dump_into` and `/obj/item/storage/belt/gun/handle_item_insertion` to `/obj/item/storage/belt/gun/flaregun/dump_into` and `/obj/item/storage/belt/gun/flaregun/handle_item_insertion`

Drulikar (changes requested):

code/game/objects/items/storage/belt.dm
Huffie56: your right i must have lost the holstered_guns sugestion while trying stuff around.
i see flaregun isn't consider a gun.... that could cause other issue in other part of the code?
i will change the insertion item to only be for flargun as it's it's only purpose...
thanks.

code/game/objects/items/storage/belt.dm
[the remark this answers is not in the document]
Huffie56: dump into has a logic that can still work for other gun belt i think?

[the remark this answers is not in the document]
Huffie56: yea let's do that since dump into is only use for flare belt....

code/game/objects/items/storage/belt.dm
```diff
@@ -16,6 +16,20 @@
 	///TRUE Means that it closes a flap over its contents, and therefore update_icon should lift that flap when opened. If it doesn't have _half and _full iconstates, this doesn't matter either way.
 	var/flap = TRUE
 
+/obj/item/storage/belt/gun/flaregun/dump_into(obj/item/storage/Main, mob/user)
```
Drulikar, line 19: Two things:
- `Main` makes no sense to be the variable name; this is kinda more confusing than just a random letter like it used to be.
- We use snake_case; not camelCase or PascalCase. Your variable here is PascalCase. https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#type-paths-must-be-snake-case

So something more like `obj/item/storage/origin_storage` would be more appropriate.

code/game/objects/items/storage/belt.dm
```diff
@@ -16,6 +16,20 @@
...
+
+	if(length(holstered_guns) < 1 && length(contents) >= (storage_slots-1))
+
+		to_chat(user, SPAN_WARNING("[src] is full."))
+		return FALSE
+	return ..()
+
+/obj/item/storage/belt/gun/flaregun/handle_item_insertion(obj/item/Warn, prevent_warning = 0, mob/user)
```
Drulikar, line 27: Three things:
- `Warn` makes no sense to be the variable name; this is kinda more confusing than just a random letter like it used to be.
- We use snake_case; not camelCase or PascalCase. Your variable here is PascalCase. https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#type-paths-must-be-snake-case
- Use `FALSE` instead of `0` for booleans.

So something more like `obj/item/new_item` would be more appropriate.

code/game/objects/items/storage/storage.dm
```diff
@@ -455,16 +454,16 @@ That's done by can_be_inserted(). Its checks are whether the item exists, is an
 The stop_warning parameter will stop the insertion message from being displayed. It is intended for cases where you are inserting multiple
 items at once, such as when picking up all the items on a tile with one click.
 user can be null, it refers to the potential mob doing the insertion.**/
-/obj/item/storage/proc/handle_item_insertion(obj/item/W, prevent_warning = 0, mob/user)
-	if(!istype(W))
+/obj/item/storage/proc/handle_item_insertion(obj/item/Warn, prevent_warning = 0, mob/user)
```
Drulikar, line 457: Three things:
- `Warn` makes no sense to be the variable name; this is kinda more confusing than just a random letter like it used to be.
- We use snake_case; not camelCase or PascalCase. Your variable here is PascalCase. https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#type-paths-must-be-snake-case
- Use `FALSE` instead of `0` for booleans.

So something more like `obj/item/new_item` would be more appropriate.

code/game/objects/items/storage/storage.dm
```diff
@@ -721,24 +720,25 @@ W is always an item. stop_warning prevents messaging. user may be null.**/
 			to_chat(user, SPAN_WARNING("[ammo_dumping] is empty."))
 	return TRUE
 
-/obj/item/storage/proc/dump_into(obj/item/storage/M, mob/user)
+/obj/item/storage/proc/dump_into(obj/item/storage/Main, mob/user)
```
Drulikar, line 723: Two things:
- `Main` makes no sense to be the variable name; this is kinda more confusing than just a random letter like it used to be.
- We use snake_case; not camelCase or PascalCase. Your variable here is PascalCase. https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#type-paths-must-be-snake-case

So something more like `obj/item/storage/origin_storage` would be more appropriate.

code/game/objects/items/storage/storage.dm
```diff
@@ -721,24 +720,25 @@ W is always an item. stop_warning prevents messaging. user may be null.**/
...
+	if(!has_room(Main.contents[1])) //Does it have room for the first item to be inserted?
 		to_chat(user, SPAN_WARNING("[src] is full."))
 		return
 
-	to_chat(user, SPAN_NOTICE("You start refilling [src] with [M]."))
+	to_chat(user, SPAN_NOTICE("You start refilling [src] with [Main]."))
 	if(!do_after(user, 1.5 SECONDS, INTERRUPT_ALL, BUSY_ICON_GENERIC))
 		return
-	for(var/obj/item/I in M)
+	for(var/obj/item/I in Main)
```
Drulikar, line 738: Still have single letter vars. E.g. replace `I` with `new_item` or something.

code/game/objects/items/storage/belt.dm
```diff
@@ -16,6 +16,20 @@
...
+
+	if(Warn.type == /obj/item/device/flashlight/flare && length(holstered_guns) < 1 && contents.len >= (storage_slots-1))
```
Drulikar, line 29: Two things (I'd just make it a suggestion, but Warn is also requested to be changed):
- `length(contents)` instead of `contents.len` Faster, and handles null.
- `istype` instead of `.type ==` e.g. `istype(new_item, /obj/item/device/flashlight/flare)`
As I mentioned previously in a requested change, doing a .type comparision like this is effectively what the istypestrict macro is doing. We want to allow subtypes, so use istype.

code/game/objects/items/storage/storage.dm
```diff
@@ -721,24 +720,25 @@ W is always an item. stop_warning prevents messaging. user may be null.**/
...
 		if(!has_room(I))
```
Drulikar, line 739: Currently this will return TRUE even though your handle_item_insertion will return false because that is the one where you put the logic to keep a space for the flaregun.

So effectively right now if you fill a flaregun belt without a flaregun holstered, the extra flares spill out onto the ground. Sort of funny, but I doubt you intend this.

https://github.com/cmss13-devs/cmss13/assets/76988376/be9b8b79-bb78-49ea-8d22-c349508cae32

So:
- Either you should be overriding the logic for has_room for a `/obj/item/storage/belt/gun/flaregun` (possibly instead of handle_item_insertion)
- Or you also put the handle_item_insertion call into the if to break (e.g. `if(!has_room(new_item) || !(handle_item_insertion(new_item, TRUE, user))`
- Or just place handle_item_insertion into its own if, put the flare back into the pouch, and then break the loop.

Be sure to test your changes and think of possible edge cases like I did here.

Drulikar (changes requested):

code/game/objects/items/storage/storage.dm
Huffie56: My bad put it there as a placeholder i replace it with origin_storage like you suggest.

code/game/objects/items/storage/belt.dm
Huffie56: My bad put it there as a placeholder i replace it with origin_storage like you suggest.

code/game/objects/items/storage/belt.dm
Huffie56: i replace the Warn by new_item and replace the case of = 0 with = FALSE

code/game/objects/items/storage/belt.dm
Huffie56: replace contents.len to what you propose.
change to istype like you proposed.

code/game/objects/items/storage/storage.dm
Huffie56: replace Warn like you proposed.
and replace = 0 by = FALSE like you proposed

code/game/objects/items/storage/storage.dm
Huffie56: replaced by new_item like you asked.

code/game/objects/items/storage/storage.dm
Huffie56: i tested it with or without the flare gun holster and i didn't saw it ... weird
i had this happening in previous version when i tested it so i am not surprise....
unsure what to do i will look into it.

Huffie56: i try to reproduce it after fixing everything else you commented
and i can't reproduce what you show on the video all i get is a the rig is full.....

Huffie56: oh i did reproduce it when i hold the belt in my hand.... weird...

Drulikar: It will correctly think its full if its actually full. But if its not actually full (because it has an empty space for the gun) it will dump extras on the floor. My video has it on the belt slot in hand - but shouldn't matter.

Huffie56:
i tested this fix and it seem to work now just need to check if i can't simplify the code i added and change the single letter variable.

code/game/objects/items/storage/belt.dm
```diff
@@ -16,6 +16,26 @@
...
+
+/obj/item/storage/belt/gun/flaregun/handle_item_insertion(obj/item/new_item, prevent_warning = FALSE, mob/user)
+
+	if(istype(new_item, /obj/item/device/flashlight/flare) && length(holstered_guns) < 1 && length(contents) >= (storage_slots-1))
+		return FALSE
+	return ..()
+
+/obj/item/storage/belt/gun/flaregun/has_room(obj/item/W as obj, W_class_override = null)
+
+	if(istype(W, /obj/item/device/flashlight/flare) && length(holstered_guns) < 1 && length(contents) >= (storage_slots-1))
```
Drulikar, line 35:
```suggestion
/obj/item/storage/belt/gun/flaregun/has_room(obj/item/new_item, W_class_override = null)

	if(istype(new_item, /obj/item/device/flashlight/flare) && length(holstered_guns) < 1 && length(contents) >= (storage_slots-1))
```

The other `obj/item/W` you modified because of changing `0` to `FALSE` are okay to leave until those procs actually get modified.

code/game/objects/items/storage/belt.dm
```diff
@@ -753,7 +773,7 @@
 	for(var/i = 1 to storage_slots)
 		new /obj/item/weapon/throwing_knife(src)
 
-/obj/item/storage/belt/knifepouch/_item_insertion(obj/item/W, prevent_warning = 0)
+/obj/item/storage/belt/knifepouch/_item_insertion(obj/item/W, prevent_warning = FALSE)
```
Drulikar, line 776:
```suggestion
/obj/item/storage/belt/knifepouch/_item_insertion(obj/item/new_item, prevent_warning = FALSE)
```

The other `obj/item/W` you modified because of changing `0` to `FALSE` are okay to leave until those procs actually get modified.

Drulikar (changes requested):
Just a couple suggestions you can apply. I will re-test when you feel its ready and undrafted.

code/game/objects/items/storage/storage.dm
Huffie56: your right i manage to reproduce it after my third just didn't update here should work tested it with part empty and nothing fall on the floor...

Huffie56:
should be fine tried to remove as many single letter variable as i could.

Drulikar (approved):

PR #4117, opened 2023-08-06
Changes to Organ-Stabilizing property

BeagleGaming1:
# About the pull request

Changes Organ-Stabilizing from an on/off flag to being dependent on the organs damage

Makes organ damage matter more if you get injured enough

# Changelog

:cl:
balance: Changes to research Organ-stabilizing property
/:cl:


morrowwolf (changes requested):
I think having it be a reduction in effect rather than complete turn off would be better. It would be less effective as the organ damage became worse.

Peri should be a way to limp away as someone with organ damage. I'd like a negative effect of some sort associated.

cm13-github:

code/modules/organs/organ_internal.dm
```diff
@@ -8,17 +8,28 @@
 /datum/internal_organ
 	var/name = "organ"
 	var/mob/living/carbon/human/owner = null
-	var/vital //Lose a vital limb, die immediately.
-	var/damage = 0 // amount of damage to the organ
+	// If you lose a vital organ, you die immediately
```
BeagleGaming1, line 11:
```suggestion
	/// If you lose a vital organ, you die immediately
```

morrowwolf:
Gonna come back to this in the future

PR #4129, opened 2023-08-07
Lights turn on at roundstart

BeagleGaming1:
# About the pull request

Lights on the ship are off when the round starts then turn on after 4 to 15 seconds

https://www.youtube.com/watch?v=WCtkibZqSpk

# Changelog
:cl:
add: Ship lights turn on when the round starts
/:cl:


code/controllers/subsystem/ticker.dm
```diff
@@ -264,6 +264,10 @@ SUBSYSTEM_DEF(ticker)
 		INVOKE_ASYNC(C, TYPE_PROC_REF(/obj/structure/closet, select_gamemode_equipment), mode.type)
 	for(var/obj/structure/machinery/vending/V in machines)
 		INVOKE_ASYNC(V, TYPE_PROC_REF(/obj/structure/machinery/vending, select_gamemode_equipment), mode.type)
+	for(var/obj/structure/machinery/power/apc/apc in machines)
```
BeagleGaming1, line 267: I hope this is the right place to put this

morrowwolf:
TM'ing so I can see how long 4-15 seconds really is with start up. I'll take a closer look at where in ticker you put this later

code/controllers/subsystem/ticker.dm
harryob, line 267: this should be an SSticker.OnRoundstart()

code/modules/power/apc.dm
```diff
@@ -156,11 +156,18 @@ GLOBAL_LIST_INIT(apc_wire_descriptions, list(
 		update_icon()
 		addtimer(CALLBACK(src, PROC_REF(update)), 5)
 
+	if(is_mainship_level(z) && !(SSticker.current_state == GAME_STATE_PLAYING || SSticker.current_state == GAME_STATE_FINISHED))
+		operating = 0
```
harryob, line 160: these should be FALSE/TRUE defines

code/modules/power/apc.dm
```diff
@@ -156,11 +156,18 @@ GLOBAL_LIST_INIT(apc_wire_descriptions, list(
...
+
 	start_processing()
 
 	if(!start_charge && is_ground_level(z) && prob(10))
 		set_broken()
 
+/obj/structure/machinery/power/apc/proc/late_activation()
+	operating = 1
```
harryob, line 168: as above

harryob (changes requested):

morrowwolf (changes requested):
I think this would be cooler if it was by light rather than by area APC so as to have kind of a delayed flicker to the lighting of each area

PR #4149, opened 2023-08-09
Mt addings

Dorodomki:
# About the pull request
adds the changes morrowr asked

adds more content for the MT Locker and CE vendor

# Changelog

:cl:
Adds: overalls, hardhats, First aid medkits, welder chestrig to the MT engineer locker, overalls for CE, 
/:cl:


morrowwolf (changes requested):
So again, remove all of the chief engineer locker changes. They should be added to the vendor.

No, you can't just remove the regular hazardvest and make it blue. Have both or don't touch it.

The firstaid pouch needs to be only spawning on the ship, not groundside so it needs to be in the subsection for not ground level.

PR #4151, opened 2023-08-10
Alters ASay chat color back to the old MSay color

blackdragonTOW:
Changing Asay back to the Msay color that many were used to for a very long time.

Because this color doesn't blend in with other major chat colors (such as dsay, and others), and it reminds us of simpler times.

<details>

</details>

# Changelog

:cl:
admin: Adjustment of ASay color
/:cl:


realforest2001 (changes requested):
This is currently making tickets appear brown too, please can you make it so tickets remain purple so they're easier to identify.

blackdragonTOW:
Removed changes to headminsay colors as those do not alter the asay color of hosts in asay.

Additionally, as it stands before this change, ASay is purple.  Tickets are purple.  There hasn't been confusion about those colors being the same, so the change here is fine.   Tickets are a function of ASay and move through that channel, thus the coloring.
To make tickets and asay separate we would have to add a great deal that would be outside the scope of this color change PR and would need to be atomized with another PR.

blackdragonTOW:
What if we did.........this....

blackdragonTOW:
Okay, this should achieve the following goals:
Headmins have different colors in ASay
ASay chat messages are the OG mod colors
Ahelps are still purple
ASay and AHelps are now distinct colors.

code/modules/admin/tabs/admin_tab.dm
```diff
@@ -255,9 +255,9 @@
 
 	log_adminpm("ADMIN: [key_name(src)] : [msg]")
 
-	var/color = "adminsay"
+	var/color = "mod"
 	if(ishost(usr))
```
realforest2001, line 259:
```suggestion
	if(check_rights(R_PERMISSIONS))
```

realforest2001 (approved):

code/modules/admin/tabs/admin_tab.dm
```diff
@@ -255,9 +255,9 @@
 
 	log_adminpm("ADMIN: [key_name(src)] : [msg]")
 
-	var/color = "adminsay"
-	if(ishost(usr))
-		color = "headminsay"
+	var/color = "mod"
+	if(check_rights(R_PERMISSIONS))
```
harryob, line 259:
```suggestion
	if(check_rights(R_PERMISSIONS, show_msg = FALSE))
```
or you'll spam non-permissioned users when they try to talk

harryob (changes requested):

harryob (approved):

PR #4162, opened 2023-08-11
Stops new specs using all gear

realforest2001:
# About the pull request
This PR changes the skill level of squad specialists so that rather than being able to use absolutely everything, they can't use anything at all without selecting a class. Not selecting a kit was already against the rules, this now makes it mechanically enforced.

Less workload for staff, and removes any confusion from squad specs.
<details>

</details>

# Changelog
:cl:
fix: Squad Specialists can no longer bypass restrictions by not selecting a kit to use.
/:cl:


harryob (approved):
why the hell was it ever the opposite way around

PR #4175, opened 2023-08-12
Admin Proc for Distress Signal TGUI and Fix

Ben10083:
# About the pull request

Readds functionality for silent ert (erts with no ARES announcements) and coverts some prompts to TGUI

TGUI pretty. Also helps event staff get reinforcements without notifying everyone ingame

# Changelog
:cl:
fix: Prompt allowing staff to call for ERT with an announcement fixed to actually appear
ui: Prompts for admin-calling ERT converted to TGUI.
/:cl:


code/modules/admin/tabs/event_tab.dm
```diff
@@ -218,21 +218,22 @@
...
-		if("Cancel")
-			qdel(chosen_ert)
-			return
-		if("No")
-			is_announcing = FALSE
+	var/is_announcing = tgui_alert(usr, "Would you like to announce the distress beacon to the server population? This will reveal the distress beacon to all players.", "Announce distress beacon?", list("Yes", "No"), 20 SECONDS)
+	if(!is_announcing)
+		qdel(chosen_ert)
+		return
+	else if(is_announcing == "No")
```
harryob, line 225: you don’t need an else if the previous condition returns

code/modules/admin/tabs/event_tab.dm
```diff
@@ -218,21 +218,22 @@
...
+		is_announcing = FALSE
```
harryob, line 226: you dont need an else here

code/modules/admin/tabs/event_tab.dm
```diff
@@ -218,21 +218,22 @@
...
-		if("Cancel")
-			qdel(chosen_ert)
-			return
-		if("Current Location")
-			override_spawn_loc = get_turf(usr)
+	var/prompt = tgui_alert(usr, "Spawn at their assigned spawn, or at your location?", "Spawnpoint Selection", list("Spawn", "Current Location"), 0)
+	if(!prompt)
+		qdel(chosen_ert)
+		return
+	else if(prompt == "Current Location")
```
harryob, line 235: you don’t need an else here

harryob (changes requested):

harryob (approved):

PR #4193, opened 2023-08-16
stops not being signed into byond from being a major pain during development

harryob:
lets you not use a byond account for testing

no player facing changes


BeagleGaming1 (approved):
code seems good

Issue #3894, opened 2023-07-14
Research: Defibrillating can't be created higher than level 1


Weuyn:
### Testmerges

none that'd effect it

### Description of the bug

Defibrillating gets stronger the higher level it is and has a special healing effect at potency 7 or higher since Warcrimes, but it is functionally impossible to actually create a chemical with so much defibrillating because the property is limited in Create mode to level 1. 

While technically you could amplify 6 times from level 1 to 7, this would cost like, 30 more points than it should, and 30 points is half of your round.

### What's the difference with what should have happened?

The property should have whatever flag it has that limits it to only level 1 removed.

### How do we reproduce this bug?

1. Get defib
2. put it in a create mode chemical
3. try to increase potency
4. receive a chat log message: 'Max level for defibrillating is 1.'
...

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Huffie56:
you got you answer with the chat log.
it as been nerf previously probably because it was too strong?

Weuyn:
Defibrillating has never been nerfed, its always been sorta bad because it lets xenos drag farther and cap easier, and has had problems with not alerting/ grabbing ghosts properly so marines are left playerless.

I mean warcrimes was merged a year ago, two years ago? And no one noticed defibrillating was impossible to make in sufficient potency to activate its touch effect, that's how little it gets used.

It's a bug for sure, a relic of when amplifying it did nothing and level 1 was as strong as level 10.

Segrain:
One day I will get to cleaning up all the utter mess that those "warcrimes" made...

PR #4195, opened 2023-08-17
Research Trait fixes

Diegoflores31:
# About the pull request

ALL traits will now use Potency instead of level to calculate their effects . I have noticed there was some traits that were not using this and as a result they were not working at all when used with modifiers like Boosting. This PR corrects that behaviour to work as it should .

PD : STATS UNTOUCHED THIS PR JUST FIXES THE BEHAVIOUR ALL NUMBERS ARE THE SAME AS BEFORE .
PD 2 : potency = level / 2 which means i have to add a (potency * POTENCY_MULTIPLIER_MEDIUM) to keep the numbers the same

bug bad . research = funny

<details>

</details>

# Changelog
:cl:
fix: Fixes Booting , Cryometabolizing , and other metabolites  not working on some traits
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry_properties/chem_property.dm
+++ b/code/modules/reagents/chemistry_properties/chem_property.dm
@@ -12,6 +12,7 @@
 	var/updates_stats = FALSE //should the property change other variables in the reagent when added or removed?
 	/// Should reagent with this property explode/start fire when mixed more than overdose threshold at once?
 	var/volatile = FALSE
+	var/potency = 1
 
 /datum/chem_property/Destroy()
 	holder = null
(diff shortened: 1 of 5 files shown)


Weuyn:
Allowing boosting to affect ciphering is a terrible idea, ciphering costs many times times as much to amplify to make it harder to spam hives (it had a unique modifier that no other property has, except the egg plasma one iirc).

Amplifying boosting from 1 to 5 costs only 7 points. Amplifying ciphering from 1 to 5 costs like 40? I think? It would be really funny, but not balanced.

fire/explosive mod traits don't care about boosting anyway, boosting only has an effect when metabolized?

Diegoflores31:
> Allowing boosting to affect ciphering is a terrible idea, ciphering costs many times times as much to amplify to make it harder to spam hives (it had a unique modifier that no other property has, except the egg plasma one iirc).

only corrupted are allowed to be friendly at marines anyways so spamming hives is not really concern . and boosting is not something you get every round . its almost as rare as fire penetrating and its outright  dumb that such a high tier, hard to get modifier just doesnt work with like half of traits

Huffie56:
would it fixed this issue @Diegoflores31 ?
https://github.com/cmss13-devs/cmss13/issues/3894

Diegoflores31:
> would it fixed this issue @Diegoflores31 ? #3894

No . That issue is not a bug sadly it is "intended" because of how research warcrimes if anything i will rather do something about it on a different PR

Unknownity:
This should require clarification from a developer regarding Boosting not affecting some properties whether that is actually intended or not.
@Segrain

Diegoflores31:
Well the description clearly says that boost other properties . And wiki says the same . Its counter intuitive that it just applies to some chems and not to others

Diegoflores31:
i realized i did a lil bit of shitcode will rewrite this PR tomorrow

Issue #4188, opened 2023-08-16
alamo's control hijack impossible to fix


piterskiy2:
### Testmerges

N/A

### Description of the bug

you can't properly restore alamo's controls after queen locks it down. instead, after finishing the long circle action, it just tells you that the control's are on cooldown for X minutes.

### What's the difference with what should have happened?

certainly not this

### How do we reproduce this bug?

1. queen locks down DS
2. try to unlockdown DS
3. epic fail
...

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

piterskiy2:
> Pretty sure there is a cooldown before you can even try to fix a dropship . But fr it should tell you BEFORE attempting the long ass 3minute thing

after waiting for 5 minutes it just fixed itself, i didn't even finish the circle action for the second time

Huffie56:
so this look like a qol just need to put the check for cooldown before you try to take back control....

PR #4209, opened 2023-08-19
Fix add a check when trying to regain control of locked DS to avoid waste of time.

Huffie56:
# About the pull request
fixes: https://github.com/cmss13-devs/cmss13/issues/4188
add a check before they start the long process of unlock the DS after queen lock it to avoid people wasting time.

just a little fix or qol or both...
<details>

image

</details>

# Changelog
:cl:
fix: add a check before people start the long process of unlock the DS so that people don't waste time.
/:cl:

### DIFF
--- a/code/modules/shuttle/computers/dropship_computer.dm
+++ b/code/modules/shuttle/computers/dropship_computer.dm
@@ -147,6 +147,8 @@
 		return
 
 	if(dropship_control_lost && skillcheck(user, SKILL_PILOT, SKILL_PILOT_EXPERT))
+		if(remaining_time > 60)
+			to_chat(user, SPAN_WARNING("The shuttle is not responding, try again in [remaining_time] [units]."))
 		to_chat(user, SPAN_NOTICE("You start to remove the Queens override."))
 		if(!do_after(user, 3 MINUTES, INTERRUPT_ALL, BUSY_ICON_HOSTILE))
 			to_chat(user, SPAN_WARNING("You fail to remove the Queens override"))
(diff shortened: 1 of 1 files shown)


code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -147,6 +147,11 @@
 		return
 
 	if(dropship_control_lost && skillcheck(user, SKILL_PILOT, SKILL_PILOT_EXPERT))
+		var/remaining_time = timeleft(door_control_cooldown) / 10
+		var/units = "seconds"
```
harryob, line 151: redundant var just put it in the string

harryob (changes requested):

code/modules/shuttle/computers/dropship_computer.dm
Huffie56: i am not sure how i do that...
not even sure in what string you want me to put it....

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -147,6 +147,11 @@
...
+		if(remaining_time > 60)
+			to_chat(user, SPAN_WARNING("The shuttle is not responding, try again in [remaining_time] [units]."))
```
harryob, line 153:
```suggestion
			to_chat(user, SPAN_WARNING("The shuttle is not responding, try again in [remaining_time] seconds."))
```

Huffie56: ah i see thanks for the clarification.

PR #4216, opened 2023-08-20
balance: change IO armor to be based off light armor with extra storage and slight buff on light.

Huffie56:
# About the pull request
No IO use the IO armor but prefer light armor so i decided to rebase IO armor on light armor with storage space and extra light.

it's sad to make a specific armor for IO but they don't use it...
<details>

</details>

# Changelog
:cl:
balance: change IO armor to be based off light armor with extra storage and slight buff on light.
/:cl:


Huffie56:
> FYI it seems that the XM4 would switch to using the M3-L sprite by being an M3-L subtype which seems a little weird.

i can modify the description even more but i am not changing the sprite... could be an lighten version of the M4...

Steelpoint:
I feel like if this armour is going to be changed to just be light armour with more inventory space, that it should also apply to the base M4 Armour for consistencies sake. 

Personally I'd just suggest making the IO's variant of the M4 armour just have the same slowdown as light armour. If anyone on the Almayer is going to be issued the actually good armour, it'd be the overpaid IOs. Though that might get denied for being powercreep.

Googles-Hands:
I just wanted to say that I don't know where you got the idea from that IO's don't use their armour.
As a matter of fact, I see much much less use of the service jacket. Thats just from what I ecperienced, maybe every IO runs stark naked around the map when I don't play.

And as for the people that don't take the armour, so what, they will always take the service jacket due to it having no slowdown, no matter how you edit the IO armour.
It's like normal PFC's in service jackets, nothing you can do about that, and trying to change the existing equipment to suit that playerbase is not something I consider good.

Huffie56:
my point is just that every IO with hours under their belt will take light armor over IO armor as speed currently is way to good mechanic wise....

Steelpoint:
From my personal observations many veteran IOs will take Light Armour over their unique armour

PR #4227, opened 2023-08-22
General Quarters via ARES no longer cares if Red alert or not

Ben10083:
# About the pull request

ARES general quarter call no longer cares if it is Red Alert already, but has a 10 minute cooldown between uses

Many been telling me how they like General Quarters, but usually can't use it as Red Alert has been called already, this makes it so it can still be used (albeit no mechanical effect), but a cooldown has been added to prevent spam.

# Changelog
:cl:
qol: General Quarters via ARES can now be called regardless of alert level.
/:cl:

### DIFF
--- a/code/game/machinery/ARES/ARES_procs.dm
+++ b/code/game/machinery/ARES/ARES_procs.dm
@@ -508,7 +508,7 @@ GLOBAL_LIST_INIT(maintenance_categories, list(
 		// -- Emergency Buttons -- //
 		if("general_quarters")
 			if(!COOLDOWN_FINISHED(src, ares_quarters_cooldown))
-				to_chat(usr, SPAN_WARNING("It has not been long enough since last General Quarters call!"))
+				to_chat(usr, SPAN_WARNING("It has not been long enough since the last General Quarters call!"))
 				playsound(src, 'sound/machines/buzz-two.ogg', 15, 1)
 				return FALSE
 			if(security_level < SEC_LEVEL_RED)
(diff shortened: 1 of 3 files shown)


code/game/machinery/ARES/ARES_procs.dm
```diff
@@ -506,16 +507,18 @@ GLOBAL_LIST_INIT(maintenance_categories, list(
...
 		// -- Emergency Buttons -- //
 		if("general_quarters")
-			if(security_level == SEC_LEVEL_RED || security_level == SEC_LEVEL_DELTA)
-				to_chat(usr, SPAN_WARNING("Alert level is already red or above, General Quarters cannot be called."))
+			if(!COOLDOWN_FINISHED(src, ares_quarters_cooldown))
+				to_chat(usr, SPAN_WARNING("It has not been long enough since last General Quarters call!"))
 				playsound(src, 'sound/machines/buzz-two.ogg', 15, 1)
 				return FALSE
-			set_security_level(2, no_sound = TRUE, announce = FALSE)
+			if(security_level != SEC_LEVEL_RED && security_level != SEC_LEVEL_DELTA)
```
Diegoflores31, line 514: you could also use 
if(security_level < SEC_LEVEL_RED) 
since each sec level is a number
green = 0
blue = 1 
red = 2 
delta = 3

code/game/machinery/ARES/ARES_procs.dm
```diff
@@ -506,16 +507,18 @@ GLOBAL_LIST_INIT(maintenance_categories, list(
...
+				set_security_level(2, no_sound = TRUE, announce = FALSE)
```
Diegoflores31, line 515: use defines for this proc. it should be 
set_security_level(SEC_LEVEL_RED, no_sound = TRUE, announce = FALSE)

Diegoflores31 (commented):
code Lgtm , UI dunno

harryob (approved):

PR #4239, opened 2023-08-23
Fix wire cutter on a plant bug

### LINKED ISSUE #3367: Bug using a wire-cuter on a plant will create a sample but will increase the growth stage of the plant.
### Testmerges

not relevant

### Description of the bug

using a wire-cuter on a plant will create a sample as expected but will increase the growth stage of the plant instead of reducing is health or any set back.
image

at the minimum taking a part out of the plant to have sample shouldn't make the plant growth...

1.use a wire-cuter on a plant
2.get a sample.
3.observe that the growth stage as increase compare to an untouched plant...
...

### Issue Bingo

MikeKuwait:
# About the pull request

Fixes #3367 

Sampled plant's growth are reduced to a probability of 10 percent.

<details>

</details>

# Changelog

:cl:

fix: Sampled plant's growth are reduced

/:cl:


Weuyn:
this is awful

Diegoflores31:
i dont think this is what the bug is refering to

MikeKuwait:
> i dont think this is what the bug is refering to

In the bug issue it says that if the plant is sampled, it shouldn't make the plant growth. Unfortunately in the code, there is no proc that stops growth of a plant while also allowing it age. This is the only solution I came up with, unless there is something I didn't understand with the issue?

MikeKuwait:
> the issue is that the act of sampling "suposedly" makes the plant age increase . the expected behaviour is that you should just sample and thats it ( and plant has a chance lot loose health and shit) . i am unsure if it is a real bug or not in the first place because it could be just coincidency .

Oh I see what you mean now that I re-read the issue properly. I somehow read it as the plant should stop growing if sampled. Thanks for clarifying!

PR #4268, opened 2023-08-26
3291 - Fixes the ability to manipulate a dummy from a distance using the tablet

### LINKED ISSUE #3291: You can interact with the dummy tablet whilst being away from it.
### Testmerges

#3273, #3271, #3088, #3284 and #3290.

### Description of the bug

I was able to de-limb and preform other actions on the dummy whilst not in the usually interact-able range (outside a 3x3 grid).

If outside the range where you should be able to interact with something, it should not allow you to use the tablet.

1. Pickup Tablet.
2. Use in hand to open UI.
3. Drop tablet and walk away.
4. Attempt to use the functions of the tablet.

### Issue Bingo

Xufysz:
# About the pull request

Closes #3291 

This code fixes an issue where you can manipulate a dummy at a distance.

N/A, an issue was raised for it.

https://github.com/cmss13-devs/cmss13/assets/8283338/40e61fb6-c833-4faa-a0fe-e2423ba7870e

# Changelog

:cl:
fix: Fixes admin training dummy interaction distance
/:cl:

### DIFF
--- a/code/game/objects/items/devices/dummy_tablet.dm
+++ b/code/game/objects/items/devices/dummy_tablet.dm
@@ -12,7 +12,12 @@
 	linked_dummy = null
 	. = ..()
 
-/obj/item/device/professor_dummy_tablet/proc/is_adjacent_to_dummy(mob/user as mob)
+/**
+ * Checks if the user is adjacent to the dummy.
+ * @param user The user to check.
+ * @return TRUE if the user is adjacent to the dummy, FALSE otherwise.
+*/
+/obj/item/device/professor_dummy_tablet/proc/is_adjacent_to_dummy(mob/user)
 	if (get_dist(linked_dummy, user) > 1)
 		to_chat(user, "You are too far away to use the tablet.")
 		return FALSE
@@ -27,7 +32,7 @@
 	interact(user)
 
 /obj/item/device/professor_dummy_tablet/interact(mob/user as mob)
-	if (linked_dummy == null)
+	if (isnull(linked_dummy))
 		return
 
 	if (!is_adjacent_to_dummy(user))
--- a/code/game/objects/items/devices/dummy_tablet.dm
+++ b/code/game/objects/items/devices/dummy_tablet.dm
@@ -13,10 +13,12 @@
 	. = ..()
 
 /**
- * Checks if the user is adjacent to the dummy.
- * @param user The user to check.
- * @return TRUE if the user is adjacent to the dummy, FALSE otherwise.
-*/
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -155,7 +155,7 @@
 						return held_weapon.afterattack(target,src)
 
 			var/disarm_chance = rand(1, 100)
-			var/attacker_skill_level = attacking_mob.skills ? skills.get_skill_level(SKILL_CQC) : SKILL_CQC_MAX // No skills, so assume max
+			var/attacker_skill_level = skills && attacking_mob.skills ? skills.get_skill_level(SKILL_CQC) : SKILL_CQC_MAX // No skills, so assume max
```
BeagleGaming1, line 158: What was this change for?

Xufysz: When playing as CO, and trying to disarm the dummy, skills was null. Im assuming skills is referring to the dummy, which would make sense. Either way seems like a check was needed for it as it was erroring.

code/game/objects/items/devices/dummy_tablet.dm
```diff
@@ -12,6 +12,13 @@
 	linked_dummy = null
 	. = ..()
 
+/obj/item/device/professor_dummy_tablet/proc/is_adjacent_to_dummy(mob/user as mob)
```
harryob, line 15:
```suggestion
/// Checks if the provided mob is next to the dummy, returns TRUE if they are, and returns FALSE if they are not. Sends a message to the user if they are too far.
/obj/item/device/professor_dummy_tablet/proc/is_adjacent_to_dummy(mob/user)
```
`as user` is verb specific syntax, and we DMdoc new procs per https://github.com/tgstation/tgstation/blob/master/.github/guides/AUTODOC.md

code/game/objects/items/devices/dummy_tablet.dm
```diff
@@ -20,6 +27,12 @@
 	interact(user)
 
 /obj/item/device/professor_dummy_tablet/interact(mob/user as mob)
+	if (linked_dummy == null)
```
harryob, line 30:
```suggestion
	if (!linked_dummy)
```
alternatively, you can is `isnull(linked_dummy)`, but just checking for falsiness is fine

harryob (changes requested):

harryob (approved):

PR #4269, opened 2023-08-26
LV-624 Move Position of Hidden Floor Safe

Steelpoint:
# About the pull request

Moves the positioning of the hidden floor safe in the Commandant's office from being placed in a wall to being in the centre of the tile it is on.

Floor safe should be on the floor, not hidden under a wall. 

<details>

</details>

# Changelog
:cl:
maptweak: Hidden floor safe in LV-624 is now positioned correctly. 
/:cl:

### DIFF
--- a/maps/map_files/LV624/LV624.dmm
+++ b/maps/map_files/LV624/LV624.dmm
@@ -10305,7 +10305,6 @@
 "aSQ" = (
 /obj/structure/safe/floor{
 	name = "safe";
-	pixel_x = 30;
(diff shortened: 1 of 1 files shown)


realforest2001 (approved):

Steelpoint:
Just to dissuade any issues with the change. As it stands the hidden safe is only revealed when you remove the floor tile where it is hidden, however the offset means when you remove the floor tiles, the safe magically appears inside a wall and not under the floor. Which makes no sense.

PR #4277, opened 2023-08-28
Machete in-hand activation

JackieEstegado:
# About the pull request

Trying to clear vines by clicking on them is extremely infuriating due to their thin hitboxes. This will give marines a much much comfier way to clear vines, while also making the machete actually be better for doing it than the easier to store hatchet.
Works by checking the tile directly ahead of the user, and if it finds any cutable objects it will try to cut them.
The delay on cutting was set in a way so as to reflect the speed at what a semi-good player would be able to cut at with clicking (accounting for usual server tick slowdown).

Also there appears to be a ton of invisible vines littered all over LV, someone should fix that.

QOL, less annoying to do an important job on maps like LV-624.

<details>

https://github.com/cmss13-devs/cmss13/assets/15560820/e0ccb094-4f1a-49cd-a8cb-08b4428bfa33

</details>

# Changelog
:cl:
qol: Added in-hand activation for machetes, which will cause one to start cutting vines and grass on the tile directly in-front.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/blades.dm
+++ b/code/game/objects/items/weapons/blades.dm
@@ -32,6 +32,25 @@
 	icon_state = "machete"
 	w_class = SIZE_LARGE
 
+/obj/item/weapon/claymore/mercsword/machete/attack_self(mob/user)
+	if(user.action_busy)
+		return
+
+	var/turf/root = get_turf(user)
+	var/facing = user.dir
+	// List containing the 3 tiles in front of the user
+	var/list/in_front = list(get_step(root, facing),  get_step(root, turn(facing, 45)),  get_step(root, turn(facing, -45)))
+
+	// We check each tile in front of us, if it has flora that can be cut we will attempt to cut it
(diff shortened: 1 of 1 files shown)


code/game/objects/items/weapons/blades.dm
```diff
@@ -32,6 +32,25 @@
...
 	w_class = SIZE_LARGE
 
+/obj/item/weapon/claymore/mercsword/machete/attack_self(mob/user)
+	if(user.action_busy)
+		return
+
+	var/turf/root = get_turf(user)
+	var/facing = user.dir
+	// List containing the 3 tiles in front of the user
+	var/list/in_front = list(get_step(root, facing),  get_step(root, turn(facing, 45)),  get_step(root, turn(facing, -45)))
```
morrowwolf, line 42: Definitely don't want this reaching out past the tile in front of the user, so just one tile there.

morrowwolf (changes requested):

harryob (approved):

PR #4290, opened 2023-08-30
overlay limbs for improved client performance

harryob:
from:
image

to:
image

lower number is obviously good, especially on clients because this can make being on the front with a bunch of humans incredibly laggy

for some reason we use vis_contents for limbs, which simplifies their icon updating. however it does also make clients lag the fuck out for whatever reason so we shouldn't do that

also the fact that we copy these to ghosts doesn't help and can really start multiplying things

can also remove iterating through visobjs in gFI, so ended up porting over the readability changes from mothblocks on tg here:
https://github.com/tgstation/tgstation/pull/60285

god willing no player facing changes


Zonespace27 (approved):
Good for tm

PR #4300, opened 2023-09-01
Adds synthetic night vision

Katskan:
# About the pull request

Standardizes vision between ship and colony synthetics.

Removes meta, improves individual style options, increases lore accuracy by integrating the overwhelmingly universal night optic pick into innate traits.

# Changelog

:cl:
balance: Merged vision toggle options between ship and colony synthetic types.
balance: Removed 15 points from WY experimental tool vendor coin redemption
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/synthetic.dm
+++ b/code/modules/mob/living/carbon/human/species/synthetic.dm
@@ -89,10 +89,6 @@
 	knock_down_reduction = 3.5
 	stun_reduction = 3.5
 
-	inherent_verbs = list(
-		/mob/living/carbon/human/synthetic/proc/toggle_HUD,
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/species/synthetic.dm
```diff
@@ -89,6 +90,7 @@
 	stun_reduction = 3.5
 
 	inherent_verbs = list(
```
morrowwolf, line 92: You can just remove this list and it will no longer override the parent list and instead inherit from the list on line 45

morrowwolf (changes requested):

code/modules/mob/living/carbon/human/species/synthetic.dm
Katskan: Should be done and good now

PR #4305, opened 2023-09-01
Removes debuffs from the folded M39 stock

VileBeggar:
# About the pull request

removes the debuffs from the folded M39 stock, making it equal to a stockless M39.

currently there's no reason to fold your stock, as it inflicts a heavy wielded scatter penalty (from 9 scatter when extended to 15 when folded) with the only benefit being a removal of the wield delay and slightly better onehanded performance.

a marine shouldn't be penalized for interacting with the folding stock. hopefully this make the attachment a bit more useful in its secondary form.

<details>
  <summary>stat changes for anyone curious how it looks like in-game</summary>
  
image
(old stock, folded)
image
</details>

# Changelog

:cl:
balance: The M39 foldable stock no longer inflicts debuffs when folded.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -1809,7 +1809,7 @@ Defined in conflicts.dm of the #defines folder.
 
 /obj/item/attachable/stock/smg/collapsible
 	name = "submachinegun folding stock"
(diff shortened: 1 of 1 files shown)


morrowwolf (approved):
Sure

PR #2733, opened 2023-02-26
Misc shuttle fixes

mullenpaul:
# About the pull request

Fixed various issues with dropships and escape pods based off recent feedback.

# Explain why it's good for the game

Smooths some of the rough edges with the dropships PR

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl:
fix: Queen can't hijack dropships on the Almayer
fix: Ported upstream fix to stop shuttle separation
fix: Resolved auto launch feature for ERT
fix: escape pod sound now correct
/:cl:

PR #4315, opened 2023-09-02
Possible fix for double dropship landing sound

morrowwolf:
# About the pull request

From time to time I will hear two landing sounds when landing in a dropship. 

I believe this is the cause of it as the landing sound *should* be handled by on_prearrival() at line 551 of code/modules/shuttle/shuttle.dm

Bug bad

<details>

</details>

# Changelog

:cl: Morrow
fix: Possible fix for double dropship landing sound
/:cl:

### DIFF
--- a/code/modules/shuttle/shuttles/dropship.dm
+++ b/code/modules/shuttle/shuttles/dropship.dm
@@ -123,7 +123,6 @@
 		if(istype(destination, /obj/docking_port/stationary/marine_dropship))
 			dropzone.turn_on_landing_lights()
 		playsound(dropzone.return_center_turf(), landing_sound, 60, 0)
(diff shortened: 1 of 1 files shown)


Drulikar:
The first iteration of the new landing sound only worked on the almayer and not for the passengers. So either this is an observer issue, or there's a timing issue where sometimes the dropship is on the almayer z level and other times not when the sound gets played.

From the blame it was introduced here: https://github.com/cmss13-devs/cmss13/pull/2733

Drulikar (changes requested):
This change seems to make it so passengers on the dropship no longer have a landing sfx. There is only a landing sfx for people at the landing site.

morrowwolf:
HMMMMM, maybe it is just an observer situation

PR #4344, opened 2023-09-06
Maybe fixes an issue with burst firing

Zonespace27:
# About the pull request
I've heard of guns *very rarely* getting stuck with their bursting var set to on. This will at least *try* to fix it. No guarantee, can't figure it out.


morrowwolf:
Sadly, no good. Problem seems to persist. This also seems to cause a few null references due to having no callback for whatever reason.

PR #4357, opened 2023-09-08
IOs get their own squads

TheGamerdk:
# About the pull request

IOs get their own squad, and they get assigned roundstart

It promotes teamwork for IOs, the lack of which is the #1 cause of IO death

In addition there's already an overwatch console in the CIC office.

You could already do this via Echo, but that's not apparant to new players

<details>

</details>

# Changelog
:cl:
add: IOs start in their own squad now
/:cl:


morrowwolf (changes requested):
No race stripes, gonna have to code a way to turn it off
Set all of the "max" types to 0 from /datum/squad (lines 45-55)
Preferably make it so this can't have an aSL as it's goofy

PR #4364, opened 2023-09-08
heavy mg ammo accuracy to 75%

TeDGamer:
# About the pull request

Currently, hmg ammo's average accuracy is base accuracy + 75 within it's accurate range with some variance. That's ~160%

This does not make sense and I believe the intention was to make the TOTAL accuracy 75%.

https://gitlab.com/cmdevs/colonial-warfare/-/merge_requests/2573

"Accuracy from 40 to 75"

No more accurate bullets outside the accurate range. 

I shot at a crusher 14 tiles away as a test and the final projectile hit chance was still over 100% before this change.

Base accuracy on bullet generation
image

# Changelog
:cl:
balance: Reduced M2C accuracy to be more in line with realistic accuracy intent
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1876,7 +1876,7 @@
 	accurate_range = 10
 	damage =  50
 	penetration = ARMOR_PENETRATION_TIER_6
-	accuracy = HIT_ACCURACY_TIER_9 + HIT_ACCURACY_TIER_5 // 75 accuracy
+	accuracy = -HIT_ACCURACY_TIER_2 // 75 accuracy
 	shell_speed = AMMO_SPEED_TIER_2
 	max_range = 15
 	effective_range_max = 7


morrowwolf (approved):

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1876,7 +1876,7 @@
 	accurate_range = 10
 	damage =  50
 	penetration = ARMOR_PENETRATION_TIER_6
-	accuracy = HIT_ACCURACY_TIER_9 + HIT_ACCURACY_TIER_5 // 75 accuracy
+	accuracy = -HIT_ACCURACY_TIER_2 // 75 accuracy
```
morrowwolf, line 1879: I feel like half the PRs you make I have to do some mind wrapping to get to the point where things are considered a "fix". 

Remove the comment as it's not relevant to the total configuration of the weapon (as multiple guns with various settings could potentially use any ammo datum).

This is a balance change (not one I'm against mind you). 

Please stop tagging these as fixes.

morrowwolf (changes requested):

code/modules/projectiles/ammo_datums.dm
TeDGamer: Sorry about that. Will keep in mind going forward

PR #4365, opened 2023-09-09
[Balance] Small xenos ( larva, huggers ) has a smaller chance of being hit

TeDGamer:
# About the pull request

Just a part of looking at projectile hit chances.

Noticed that while big xenos had a hit buff, that was it. Decided it might be good for small ( only larva and huggers, lessers unaffected by this change ) can get a slight nerf to their hit chance. Basically opposite of the big hit chance increase.

Well, for small creatures, just a slgihtly better dodge.

Obviously, shouldn't change too much as -10% is a small buff. But should feel a bit better than consistently dying to stray bullets without even a chance considering huggers take 3 minutes to respawn.

</details>

# Changelog
:cl:
balance: Huggers and larvas are slightly harder to hit
code: Replace some w_class magic numbers with defines
/:cl:


code/modules/projectiles/projectile.dm
```diff
@@ -799,7 +799,8 @@
 			if(X.hivenumber == hivenumber)
 				return FALSE
 
-		if(mob_size >= MOB_SIZE_BIG) . += 10
+		if(mob_size == MOB_SIZE_SMALL) . -= 10
```
morrowwolf, line 802: No in line if statements

code/modules/projectiles/projectile.dm
```diff
@@ -799,7 +799,8 @@
...
+		else if(mob_size >= MOB_SIZE_BIG) . += 10
```
morrowwolf, line 803: same here

morrowwolf (changes requested):

morrowwolf (approved):

PR #4379, opened 2023-09-09
Science Annex Yard Readdition

c4xmaniac2:
# About the pull request

This PR aims to readd yard to Science Annex. In February last year, we removed yard due to having a negative effect on Science Annex's gameplay flow. However, the complete removal of it has shown over time that it also has its own major issues. This readdition should address the flow issues that the map suffers. 

Will probably require test merging to gauge its effect, another possible solution is only allowing it to serve as a connector to the North and South.

Science Annex currently suffers badly from there being no center of the map, meaning that you have to travel all the way around the map to reach the other side. This should aid in the flow of the map, without allowing yard to turn into the meta afk and hold spot while ignoring the rest of the map. You can use yard as a hold, but you will need to use the rest of the map to push. 

<details>

image

</details>

# Changelog

:cl:
mapadd: Readds yard to Science Annex
maptweak: Redid the yard loot to be more in line with current surv loot standards.
/:cl:

### DIFF
--- a/maps/map_files/FOP_v3_Sciannex/Fiorina_SciAnnex.dmm
+++ b/maps/map_files/FOP_v3_Sciannex/Fiorina_SciAnnex.dmm
@@ -174,6 +174,15 @@
 	},
 /turf/open/floor/plating/prison,
 /area/fiorina/station/lowsec)
+"acx" = (
+/obj/structure/flora/bush/ausbushes/grassybush{
+	icon_state = "ppflowers_2"
+	},
+/turf/open/organic/grass{
+	desc = "It'll get in your shoes no matter what you do.";
+	name = "astroturf"
+	},
+/area/fiorina/station/central_ring)
 "adk" = (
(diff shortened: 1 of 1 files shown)


c4xmaniac2:
I might need to remove the m60ds or whatever they're called now that they have full auto, but we'll see how marines/survs abuse them. Pretty sure science annex has two right now anyways.

Diegoflores31:
> I might need to remove the m60ds or whatever they're called now that they have full auto, but we'll see how marines/survs abuse them. Pretty sure science annex has two right now anyways.

arent those M60D unwrenchable ? never seen survs use it before because yard is a pretty bad hold spot . if they wish to hold in such a bad spot i dont see wrong on having a lil gaming

c4xmaniac2:
Not sure wtf Merge branch 'cmss13-devs:master' into prison-yard-readdition is ngl, I just tried to upload the updated annex map file and the sprinkle

PR #4380, opened 2023-09-10
UPP rank

CapCamIII:
# About the pull request

I filled this in bottom to top just read the explain why its good and changelog it has literally everything

Presets to fill already existing ranks and complete the UPP officer corps, no reason not to.
Kapitan serves as non-WLed XO, without the BE skill and tool, Maj-Kol serve as CO ranks with possibility of including CO council status in UPP CO roles, General roles for event purposes and the ranks already existed in-code.

Conscript having its own rank being "conscript" makes it more of a meme role, vast majority of UPP soldiers are conscripts lore-wise anyway so doesn't make sense, making them E1 with standard soldiers being E2 makes more sense and cleans up ranks a bit.

UPP rank code had multiple redundant ranks, O1 and O1M for the same rank, but just doctors and normal LTs, no actual mechanical difference, same with E3S and E3M, sapper and medic, acted the exact same ingame but separated ranks for no reason and makes potentially adding new roles a bit harder.

# Changelog
:cl:
add: Adds upp presets for the unused officer ranks, Kapitan, Lt Kolonel, MajGen, LtGen, Gen, for use in admin events
add: Changes UPP conscript rank from weird conscript only rank to Private, moves UPP soldier to PFC, removing the rank gap between soldier and support
code: Cleans up UPP rank code by deleting redundant ranks
/:cl:


code/datums/factions/upp.dm
```diff
@@ -25,10 +25,20 @@
 			hud_icon_state = "lt"
 		if(JOB_UPP_SRLT_OFFICER)
 			hud_icon_state = "slt"
-		if(JOB_UPP_MAY_OFFICER)
+		if(JOB_UPP_KPT_OFFICER)
 			hud_icon_state = "may"
```
morrowwolf, line 29: hud icon should be renamed so it's less ambiguous

morrowwolf (changes requested):
Otherwise seems fine

morrowwolf:
Seems fine, hopefully nothing has been missed.

PR #4245, opened 2023-08-24
Biolock & Tristan Donor Gear

Warfan1815:
# About the pull request

A change to Biolock's donor gear (credit to Thwomper for the sprite)
&
A change to Tristan's donor gear

Changes to legacy donor gear is allowed (to the best of my knowledge)

PR made on behalf of Biolock & Tristan.

# Explain why it's good for the game

Donor gear is good

# Testing Photographs and Procedure

https://cdn.discordapp.com/attachments/1144067901343666216/1144090409501474916/image.png

</details>

# Changelog

:cl:
add: Changes donor Biolock's medium medic armour to a light custom paintjob armour and donor Tristan's reactive scientist armour to a custom marine armour

PR #4400, opened 2023-09-12
Fix fluff type check for map overrides

### LINKED ISSUE #4398: Biolock-Donor Sprite
### Testmerges

#4379, #4182, #4378

### Description of the bug

Donor armor is not showing properly. The helmet is fine.
Screenshot 2023-09-11 163907

Should show donor armor.x

1. Spawn "/obj/item/clothing/suit/storage/marine/light/fluff/biolock"

### Issue Bingo

Drulikar:
# About the pull request

This PR fixes an issue caused by a path remapping in #4245 where a typecheck for an old fluff value no longer triggered to prevent map specific camos on the fluff gear.

Note: NO_SNOW_TYPE is more than just snow, but also not all item types even perform a test on that flag to prevent calling select_gamemode_skin.

Fixes #4398 (Or any /obj/item/clothing/suit/storage/marine/light/fluff)

# Changelog
:cl: Drathek
fix: Fix some donor items breaking on maps with camos.
fix: Fix donor items not having a fingerprint atom flag and donor armors not having conductive atom flag.
/:cl:

### DIFF
--- a/code/__DEFINES/equipment.dm
+++ b/code/__DEFINES/equipment.dm
@@ -27,7 +27,7 @@
 #define ITEM_UNCATCHABLE (1<<9)
 /// Used for nonstandard marine clothing to ignore 'specialty' var.
 #define NO_NAME_OVERRIDE (1<<10)
-/// Used for armors or uniforms that don't have a snow icon state.
+/// Used for armors or uniforms that don't have a snow/desert/etc icon state set via select_gamemode_skin (not all item types currently perform this test though).
 #define NO_SNOW_TYPE (1<<11)
 
 #define INVULNERABLE (1<<12)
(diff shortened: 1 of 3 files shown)


Drulikar:
Fluff armors/helmets were also incorrectly losing FPRINT and CONDUCT flags when they set NO_NAME_OVERRIDE.
image
image
image
image

Also NO_SNOW_TYPE is a poor flag name since its more than snow, but it is also not universally used either. Many item types do not test it before calling select_gamemode_skin. Only 5/26 references test it: 
image

morrowwolf:
>Also NO_SNOW_TYPE is a poor flag name since its more than snow

Agreed

>Many item types do not test it before calling select_gamemode_skin()

Frustrating SMH

Drulikar:
Obvious solution to that would be to test the flag within the implementations of select_gamemode_skin instead of using it to prevent calling that proc, but I don't know if there was a particular reason why it was done this way in the first place.

harryob (approved):

PR #4428, opened 2023-09-17
Hivelord Mini-Buff

HeresKozmos:
# About the pull request

Hivelord can now carry an egg in each hand instead of one single egg.

It's just a little QOL thing for Hivelords that make it easier to place more eggs in the hive when it is being built.

6e66f53cc62026742662d3535aa95f9f

# Changelog
:cl:
qol: Hivelord can now hold two eggs, one per hand.
/:cl:


morrowwolf (changes requested):
QOL is not the correct way to argue for this. Look at the grand scheme of which castes can do what and give a top down look of why this should be added to the others.

PR #4446, opened 2023-09-19
changes the fallback code

harryob:
need to check if this fixes the status on https://spacestation13.com/play/quick


harryob:
this works we're registering on https://spacestation13.com/play/quick again

PR #4447, opened 2023-09-19
Adds a Commissary to the Hangar

HeresKozmos:
# About the pull request

Adds a small self-serve commissary to the hangar.

The old panic room which this is replacing was never used so I decided to replace it with a more useful room, this commissary has the five main vendors (RecVend, Hot Drinks, Snacks, Smokes and Souto) all in one place and the intent is to have an area right in the hangar where returning/departing marines can grab things such as a walkman, cigarettes or whatever else they wish.

I believe this to be much more useful than the old panic room which I have never seen actually be used in-game at all.

I ran it locally, the disposal and vent systems both work as intended and the vendors all work as intended,

image

# Changelog

:cl:
mapadd: added a commissary accessible in the hangar
/:cl:

### DIFF
--- a/maps/map_files/USS_Almayer/USS_Almayer.dmm
+++ b/maps/map_files/USS_Almayer/USS_Almayer.dmm
@@ -30090,6 +30090,13 @@
 	icon_state = "test_floor4"
 	},
 /area/almayer/hallways/starboard_umbilical)
+"dir" = (
+/obj/structure/pipes/standard/manifold/hidden/supply,
+/obj/structure/disposalpipe/segment{
+	dir = 4
+	},
+/turf/open/floor/plating/plating_catwalk,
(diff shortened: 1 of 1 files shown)


morrowwolf (changes requested):
Needs area and APC

Huffie56:
> I would disagree with the panic room being never used - it was good for dropship incursions if you were nearby.
> 
> It is also good as a holding cell for localized prisoners.
> 
> The comissary just includes a bunch of vendors which would otherwise be found in other places of the ship. I also feel like the cramped space isn't very conducive to RP since you can only fit two or so people comfortably until you start blocking people :(

good to hear it as been used can you share some story of it being used ? tag me on discord please

cm13-github:

PR #956, opened 2022-08-28
Heavy Armor Buff + Armor Linking Reactivation

Geevies:
## About The Pull Request

Enabled the armor linking system, meaning combat gloves and combat boots now benefit from wearing heavier armor. Additionally, heavy armor now has an integrated drop pouch.

## Why It's Good For The Game

Heavy armor is weak, because the increased armor only slowed you down and made you an easier target for foot and hand slashers. Now, with the hands and feet being protected more, as well as the increase in storage capacity (and potential storage items), it seems a little more worth taking now.

## Changelog

:cl:
balance: Enabled the armor linking system, meaning combat gloves and boots now benefit from wearing heavier armor.
balance: Gave Heavy armor an integrated drop pouch, which can store more and larger items.
/:cl:

PR #4451, opened 2023-09-20
Revives Armor Linking (med/heavy armor buff)

private-tristan:
# About the pull request
a revival of #956 without the integrated drop pouch
Armor linking! wearing armor will cause your boots and gloves to share the damage resistances.

(BEFORE VS AFTER ARMOR LINKING)
for light armor you get: 
go from 15->20 bomb armor on boots and gloves

for med armor you'll get:
10 internal damage resist -> 20 internal damage resist 

for heavy armor you'll get:
20 -> 25 melee armor
10 -> 25 internal damage resist 
20 -> 35 bullet armor
15 -> 35 bomb armor
20 -> 25 bio armor

(didn't list rad/energy values because i don't think they're used??)

currently, heavy and medium armors are HEAVILY underused due to the fact that a vast majority of xenos simply aim for feet or hands, bypassing your armor and dealing the same amount of damage while being able to fracture and delimb far easier, armor linking will mitigate this, equalizing damage on every bodypart (except for helmet, which is always equivelant to med armor).

this will encourage xenos to target places that aren't just hands/feet. maybe we'll see xenos targeting head now that marines will be wearing heavy armor more often, dealing more damage for a less impactful fracture, along with possibly leading to more decaps (if they get REALLY lucky).

video demonstrating the difference between heavy, no armor, and light armor.
<details>

https://github.com/cmss13-devs/cmss13/assets/54422837/703cd552-e9ad-4758-9633-5577c83fe961

bonus video - delimbing in 3 hits through heavy armor still exists if you get lucky enough 

https://github.com/cmss13-devs/cmss13/assets/54422837/3e2a35b5-5498-429e-9cf2-e646ff3dfbc2

</details>

# Changelog
:cl: Geevies, Private Tristan
add: Armor Linking! wearing chest armor will now give its' armor stats to both your boots and gloves. 
/:cl:

### DIFF
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -211,6 +211,11 @@
 	var/gloves_blood_amt = 0 //taken from blood.dm
 	var/hide_prints = FALSE
 
+/obj/item/clothing/gloves/Initialize(mapload, ...)
+	. = ..()
+	if(armor_melee || armor_bullet || armor_laser || armor_energy || armor_bomb || armor_bio || armor_rad)
+		AddComponent(/datum/component/armor_link, WEAR_JACKET, TRUE)
+
 /obj/item/clothing/gloves/update_clothing_icon()
 	if (ismob(src.loc))
 		var/mob/M = src.loc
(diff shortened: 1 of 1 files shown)


morrowwolf:
ALRIGHT

Let's have "the talk" since this *seems* like ***the answer*** to all of our problems and a neat way to just move on.

CM is not balanced around body armor values. As with all things, it is the weakest link that breaks the chain. CM is balanced around FOOT and HAND armor. If we go forward with this marines are, frankly, going to be walking tanks of doom. Try killing a marine in heavy armor by exclusively hitting their chest as a runner or warrior. It takes *a while*. There are already instances where xenos can be fighting a decent marine and the marine will outpace them DPS-wise.

This just... ain't it. Worse, I feel it kind of cheapens the math of choosing *where* to strike. I agree it's lame that most people just target hands and feet. I think it would be good to enhance the effectiveness of wearing heavier armor but that could be in a dozen ways other than just spreading the armor to the other parts of the body.

I'm sure we'll be having this same PR again in another 6 months (as is tradition). Have a good one.

PR #4475, opened 2023-09-22
Adds drawing to tacmaps

Cthulhu80:
# About the pull request
Features 
- Gives cic (sos, xo, co) and the queen the ability to draw on the tactical map and then announce it to their respective factions
- All marine faction members now are able to access the tactical map announcement through the stat panel, which can be viewed at any time.
- Xenos in addition to having the normal tactical map will also have access to their own canvas map that can be viewed when Queen is on ovi.
- Ghosts can view the tactical maps for both factions in the Ghost tab as well as from ghost alert popups.
- Admins can view all the round's drawings and optionally delete them.
Hopefully this will spice things up for cic, queen and players alike. The idea is command and the queen would now be able to better convey what they want from their faction, which should help with general cohesion and likely expand strategic plays on both sides. (don't abuse it or harry will take it away, or make it more exclusive to the queen and co)

# Testing photographs and procedure 
Screen Shot 2023-09-23 at 6 38 25 PM
Screen Shot 2023-09-23 at 6 39 08 PM
Screen Shot 2023-09-23 at 6 40 22 PM
Screen Shot 2023-09-23 at 6 42 54 PM

# Changelog

:cl: Cthulhu80, Drathek 
add: Adds drawing to tactical maps, viewable via stat panel for marines and xeno tacmap for xenos.
fix: Corrupted (and other hives) now have separate tactical maps. 
/:cl:


Drulikar (changes requested):
In the current state, these are problems I found in no particular order (not that all need to be addressed now - there's a lot of polish problems that need to eventually be sorted out - **bolded ones I consider blocking**):
- The tacmap visor attempts to activate every visor attempt including when turning off (meaning it spams the message `The map visor is not for you to use, the map visor flickers as it powers off. x6`) and it will open the UI both when swapping to that visor, and off of that visor.
- Once the UI is opened, changing the visor to some other mode doesn't close the map so you only need to swap to it once and then indefinitely have the map available.
- The tacmap visor probably shouldn't just open up a UI window when swapped to - it just seems not in line with a visor and weirdly interrupts you. If anything it should add a button to allow you to open the map. But even just needing to switch to a different visor to then just get a button I think would feel weird. The best solution I see for this would be if the tacmap visor did the ghost tactical map hologram. But this is probably a lot to ask to get the SVG drawn on that hologram and for it to not look weird. Even if the visor just had that ghost hologram map it would be a very useful visor (probably too useful - it would need to be a very difficult to obtain visor).
- Disallowing CO from using the tacmap visor seems a bit weird. They will have the tablet instead which is better, but I don't know it seems wrong that for it to say its not for them.
- **Tacmap visor probably needs to be its own PR because this PR has too many issues that need to be sorted out.**
-----
- **Non-tgui interactions are still used (announcement message and confirmations).**
- **Old canvas view should just start with an undrawn tacmap - it should never just be an empty page (makes the tacmap visors show nothing and seem broken until there's been a tacmap announcement).**
- Tactical Map Announcement without any button to open the map seems weird. "Announcement" should probably just be changed to Update to better indicate that the tacmaps have changed maybe? It also probably should just grant a button to everyone for a duration but I imagine this may conflict with what you were trying to do via visors?
- Old canvas and tacmap buttons (when interacting with map table in command) don't have a button to return.
- Interacting with the map should just start in one of the views (likely the drawing view - or just the view last canvas view) rather than a menu asking where to go.
- Tacmap interface just isn't sized large enough - scrolling shouldn't be the default behavior - and we should figure out how to choose colors that doesn't cause a big resize (e.g. menu on top if we must so the dropdown can just cover map area)
- **It is not intuitive that updating the canvas then also has to be clicked again to be announced: The canvas needs to persist if you update it and it not be put on cooldown (and then click home or close it), or announce/update are combined into the same action.**
- **Clicking clear canvas then drawing then clear canvas doesn't clear the canvas. The clear canvas seems to only work in this state if you undo then it will behave. Undo button is also similarly wonky if you do the same thing:**
fail
- The tacmap color selection can also get out of sync in a similar way to above: 
not red
- The undo button doesn't seem to reliably keep more than one stroke saved to undo.
- The tacmap menu can crash - but I'm not exactly certain the reproduction steps I did to cause it. Fixing above three issues likely would eliminate the possibility though:
image
- **Either doing a second tacmap for the round, or resizing the window when announcing a new tacmap results in a broken result (this is "old canvas view"/tacmap visor view):** 
image
**I tried doing another one after above without resizing the window when announcing, but it has the same problem still. Problem now persists through restarts without resizing...**
-----
- Queen tacmap update message is announced basically the same as a word of the queen so xenos won't necessarily know that the text is a tacmap update other than what the queen enters into the announcement.
image
- **Tacmap drawing never seems to save for normal hive xenos after update/announcement, it always remains a blank window.** This may be related to the broken result mentioned earlier.
- **Non-queen normal hive Xenos no longer have a way to view tacmap w/ the map icons. I imagine this is what you were having difficulty getting saved - but even if we just have two views where one is a "snapshot" (with or without icons - ideally with) that is drawn on vs the existing tacmap view that would be fine.**

code/controllers/subsystem/minimap.dm
```diff
@@ -472,15 +524,131 @@ SUBSYSTEM_DEF(minimaps)
...
+
+		if ("selectColor")
+
+			toolbar_color_selection = params["color"]
+			toolbar_updated_selection = toolbar_color_selection
+			. = TRUE
+
+		if ("selectAnnouncement")
+
+			var/icon/flat_map = getFlatIcon(map_holder.map)
```
Drulikar, line 612: You were correct that the issue such as in the above image is because of this call, but I can confirm reverting #4290 allows the map to be turned into a flat icon. E.g. 
image
As far as why I'm not sure yet, but for now you can locally revert (but do not push that reversion in this PR) to test with this call working.

Even reverted though, its incorrectly coloring the icons (see how the two xenos are purpleified incorrectly compared to the normal tacmap) 
image

Drulikar (changes requested):

code/controllers/subsystem/minimap.dm
Drulikar: Additionally, the tacmap with icons should be the view when drawing so you get what you see, rather than the final result adding icons.

Drulikar: Other than the color problem for the icons, instead of reverting that pr, this patch of that PR's changes will fix the regression in getFlatIcon: 
[patch.diff.txt](https://github.com/cmss13-devs/cmss13/files/12708397/patch.diff.txt)
image
image
image

Cthulhu80:
With the issue of the undo button and clear canvas not working, it's an issue of forcing state (forgot to fix this in the new pr). If you clicked on another color and then clicked the undo button instead of clicking the undo button consecutively then it it would force a state change which would allow you to to then undo the previous line. So the undo button and clear canvas work fine, i'll fix it nonetheless though.

Cthulhu80:
> * The tacmap visor probably shouldn't just open up a UI window when swapped to - it just seems not in line with a visor and weirdly interrupts you. If anything it should add a button to allow you to open the map. But even just needing to switch to a different visor to then just get a button I think would feel weird. The best solution I see for this would be if the tacmap visor did the ghost tactical map hologram. But this is probably a lot to ask to get the SVG drawn on that hologram and for it to not look weird. Even if the visor just had that ghost hologram map it would be a very useful visor (probably too useful - it would need to be a very difficult to obtain visor).
 - please come to a conclusion on the maintainer side, getting requests to do it a bunch of different ways.

> * Disallowing CO from using the tacmap visor seems a bit weird. They will have the tablet instead which is better, but I don't know it seems wrong that for it to say its not for them.
 - The issue is with how the tgui interface is coded, when you create a new tacmap it shares the current tgui selected menu view with all users that open it (its defined on the tacmap, current_menu). Basically, when every marine accesses the tacmap they all share the current state of the menu, which is just the old canvas to view announced drawings. If the co came along and then tried to open up the global tacmap and started to draw, any marine currently viewing the marine specific  tacmap (which is any human with a skill level less than expert who has access to the helmet tacmap visor) would (to their amazement) then be able to draw and announce just like cic, and any new marine that opens the map would then close the drawing canvas for the co and they would be wondering why they are looking at the previous canvas drawing. This reminds me that I need to add a type_check for the skill level and change the menu if a marine tries to open the canvas, anywho that's issue with that. I can redesign it for it not do that if that's too egregious, but that's the reason. 

> * **Non-tgui interactions are still used (announcement message and confirmations).**
- seems ok to me, but your call.

> * **Old canvas view should just start with an undrawn tacmap - it should never just be an empty page (makes the tacmap visors show nothing and seem broken until there's been a tacmap announcement).**
-  The reason this was done this way was because I wanted to optimize the amount of times h2ml2icon and getFlatIcon are called, for there to be background image we need to generate a png asset, and doing that every time a new interface is opened doesn't seem optimal for performance hence why it's only done for cic and queen (new joins for cic need to be able to draw) and left blank for everyone else (would it be bad to just add some fluff in there?). The marines who join late won't be able to see the old tacmap asset (it wasn't distributed to them originally) but when the new one is announced it should be distributed to everyone currently in the faction. This was the most optimal way I could think of distributing the map asset as a png without sacrificing too much in performance, if there is a better way to do this or if I'm missing something please let me know. 

> * Tactical Map Announcement without any button to open the map seems weird. "Announcement" should probably just be changed to Update to better indicate that the tacmaps have changed maybe? It also probably should just grant a button to everyone for a duration but I imagine this may conflict with what you were trying to do via visors?
- yeah, please figure this out on the maintainer end cause i'm getting different requests for different things when it comes to announcing it. I can remove the announcement of it, but I feel like queens/cic would like that feature. Instead of having to announce that the tacmap has been updated. 

> * Old canvas and tacmap buttons (when interacting with map table in command) don't have a button to return.
- can't layer anything ontop of byondui and i'll add the home button to the old canvas for cic and queen.

> * Interacting with the map should just start in one of the views (likely the drawing view - or just the view last canvas view) rather than a menu asking where to go.
- I disagree, but ok

> * Tacmap interface just isn't sized large enough - scrolling shouldn't be the default behavior - and we should figure out how to choose colors that doesn't cause a big resize (e.g. menu on top if we must so the dropdown can just cover map area)
- sure

> * **It is not intuitive that updating the canvas then also has to be clicked again to be announced: The canvas needs to persist if you update it and it not be put on cooldown (and then click home or close it), or announce/update are combined into the same action.**
-ok

> * **Clicking clear canvas then drawing then clear canvas doesn't clear the canvas. The clear canvas seems to only work in this state if you undo then it will behave. Undo button is also similarly wonky if you do the same thing:**
> * The undo button doesn't seem to reliably keep more than one stroke saved to undo.
>   **In other words, the tacmap drawings will work until a single marine is groundside or you are in distress signal mode which has sensor tower icons. HTML view of this broken state appears fine:**
- again, it does, like stated before it's an issue of state (you select undo, then you selected it again and it doesn't register it as a new state). Select another color, click undo or clear and every thing will work, but as stated it is still somewhat of a bug so it does need to be fixed regardless. 

> * Queen tacmap update message is announced basically the same as a word of the queen so xenos won't necessarily know that the text is a tacmap update other than what the queen enters into the announcement.
- sure

> * **Tacmap drawing never seems to save for normal hive xenos after update/announcement, it always remains a blank window.** This may be related to the broken result mentioned earlier where any minimap icons breaks this tacmap preview.
> * **Non-queen normal hive Xenos no longer have a way to view tacmap w/ the map icons. I imagine this is what you were having difficulty getting saved - but even if we just have two views where one is a "snapshot" (with or without icons - ideally with) that is drawn on vs the existing tacmap view that would be fine.**
- that would be the getFlatIcon issue as discussed earlier, it's defaulting to empty because the flatIcon that the user is supposed to see as their background never gets generated, and also the hive issue. 2 potential causes there. I'll fix the hive one. There was a concern with performance for generating the png asset as the background for every single user, it's already done for cic users but if 90 marines do icon2html at once I don't know if that's ideal or not.

code/controllers/subsystem/minimap.dm
Cthulhu80: > Additionally, the tacmap with icons should be the view when drawing so you get what you see, rather than the final result adding icons.
 - This would mean I would have to flatten the icon every so often even if cic or the queen choose not to announce, I guess I could add a timer for how often that happens, but even then the flattened icon png would be out of sync with the current tacmap. Again, performance concern was the issue here, and also the issue of sync.

tgui/packages/tgui/interfaces/TacticalMap.tsx
```diff
@@ -1,24 +1,277 @@
...
+  );
+};
+
+const OldMapPanel = (props, context) => {
+  const { data, act } = useBackend<TacMapProps>(context);
+  return (
+    data.flatImage && (
+      <DrawnMap svgData={data.svgData} flatImage={data.flatImage} />
+    )
+  );
```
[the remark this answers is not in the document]
Cthulhu80: > So for xenos it seems data.flatImage is not truthy. For example if I change the return to this:
> 
> ```js
>   return <DrawnMap svgData={data.svgData} flatImage={data.flatImage} />;
> ```
> 
> Then at least xenos do see the drawing and the html isn't empty here: image
> 
> The code here doesn't need to change persay, but just pointing out thats why xenos always have an empty panel atm because it resolves to false and thus puts nothing.

 - that was by design, I didn't want to overlay a drawing ontop of a non-existent png asset. So if a new user joins and tries to open the tacmap, they wouldn't have access to old generated distributed png asset (icon2html) but they would if they were already in the , I was asking in discord if there was a way to check if an asset is stored on the users end (in the case of new users, it wouldn't be) and then distribute it to them, but I didn't get a response.

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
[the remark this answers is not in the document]
Cthulhu80: Same issue as described here but only switch the roles from co and marines to queen and xenos -> 

The issue is with how the tgui interface is coded, when you create a new tacmap it shares the current tgui selected menu view with all users that open it (its defined on the tacmap, current_menu). Basically, when every marine accesses the tacmap they all share the current state of the menu, which is just the old canvas to view announced drawings. If the co came along and then tried to open up the global tacmap and started to draw, any marine currently viewing the marine specific tacmap (which is any human with a skill level less than expert who has access to the helmet tacmap visor) would (to their amazement) then be able to draw and announce just like cic, and any new marine that opens the map would then close the drawing canvas for the co and they would be wondering why they are looking at the previous canvas drawing. This reminds me that I need to add a type_check for the skill level and change the menu if a marine tries to open the canvas, anywho that's issue with that. I can redesign it for it not do that if that's too egregious, but that's the reason.

So, the queen is getting her own tacmap that the xenos can't interact with and vice versa. like I said, I can change it upon request.

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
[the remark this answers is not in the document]
Cthulhu80: answered in the above comment.

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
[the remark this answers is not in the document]
Drulikar: Are you sure about that? Literally try reverting it and see the behavior - because that's what I did. If you meant marines seeing xeno drawings - then ok I didn't test that, but I still think these changes in these two files wouldn't be necessary to resolve that - rather hive.tacmap would need to not be a clone of marine's (assuming that was the issue). With how the TGUI is right now, if you have both queen and non-queen xenos both access `hive.tacmap.tgui_interact(src)` instead of queen getting her own like via below image, only queen has access to drawing abilties, and non-queen xenos can only view the last tacmap.
image
It doesn't really matter, but the changes appear to be unnecessary. Is the concern about the user finding some hacky way to manipulate the UI to go back to get to the drawing view of the interface? If so then they *actually* need to be different implementations of the interface because at the moment they are the same type and they share the same maps.

tgui/packages/tgui/interfaces/TacticalMap.tsx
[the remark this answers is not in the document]
Drulikar: > The code here doesn't need to change persay, but just pointing out thats why xenos always have an empty panel atm because it resolves to false and thus puts nothing.

code/controllers/subsystem/minimap.dm
Drulikar: > This would mean I would have to flatten the icon every so often even if cic or the queen choose not to announce, I guess I could add a timer for how often that happens, but even then the flattened icon png would be out of sync with the current tacmap. Again, performance concern was the issue here, and also the issue of sync.

It wouldn't need to be periodic - only once ever they open the new canvas.

Also leaving the view and returning back doesn't keep the drawing around - even if you update and not announce, There's no "coming back" to your drawing - or at least I certainly could not get that functionality to work in testing.

Drulikar:
> please come to a conclusion on the maintainer side, getting requests to do it a bunch of different ways.

You must have missed the part where I said:
> Tacmap visor probably needs to be its own PR because this PR has too many issues that need to be sorted out.

Who are these other maintainers anyways? If its Harry or Morrow telling you a way, there isn't anyone else but a host that can  change that decision. I also highly doubt what you were told was a demand.

---
> The issue is with how the tgui interface is coded, when you create a new tacmap it shares the current tgui selected menu view with all users that open it (its defined on the tacmap, current_menu)

Then that's still an issue in command w/ the map table. Ultimately this is because of how this particular TGUI interface is implemented. If I were to guess part of the problem is that the variable `current_menu` is part of `/datum/tacmap` rather than say a `useLocalState` variable in the javascript. You are correct that often the same device shares its TGUI interface (e.g. two people viewing the same exact camera monitor computer), but each visor should be each their own instance of the TGUI.

---
> Non-tgui interactions are still used (announcement message and confirmations).

See https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STANDARDS.md#user-interfaces

---
> The reason this was done this way was because I wanted to optimize the amount of times h2ml2icon and getFlatIcon are called, for there to be background image we need to generate a png asset, and doing that every time a new interface is opened doesn't seem optimal for performance 

My suggestion is once at round start for all factions w/ tacmaps, not every time the interface is opened by anyone. One way to do this would be where its global define is created because it accepts a function call as its initializer. In otherwords there is never a time the "old canvas" data is empty. It only needs to change when a CO/Queen change it.

---
> yeah, please figure this out on the maintainer end cause i'm getting different requests for different things when it comes to announcing it. I can remove the announcement of it, but I feel like queens/cic would like that feature. Instead of having to announce that the tacmap has been updated.

What I wrote here does not differ with what Morrow expected after your message in development, and ultimately its Morrow's call. The concept originally was for this to replace announcements with Gyazo links. Having an announcement that opens a TGUI map rather than a web browser to gyazo or imgur of an image was the expectation.
image
Regardless it feels weird for it to go to chat (thats what it does now - you have a chat announcement), but then there's no link map or anything to view. My recommendation is to sort out visors in a different PR. Just get the drawing on a map at all working in this PR. Whether you want to do this via an action button at the top of the screen (like the ghost button when you are dead), or just a link in chat is up to you. 

---
> that would be the getFlatIcon issue as discussed earlier, it's defaulting to empty because the flatIcon that the user is supposed to see as their background never gets generated, and also the hive issue. 2 potential causes there. I'll fix the hive one. There was a concern with performance for generating the png asset as the background for every single user, it's already done for cic users but if 90 marines do icon2html at once I don't know if that's ideal or not.

I think it best that the drawing is a snapshot of the state of things, but then like on master xenos should still have access to the live tacmap like they always have (its just the first tab like you gave CO). If you're intending to be doing a getFlatIcon of the tacmap every few seconds thats going to be a lot of performance loss and may be significant memory wise for a particularly long round. It doing that operation on demand when the CO/Queen makes a new map is fine though.

code/controllers/subsystem/minimap.dm
Cthulhu80: that's what I mean, say a  so,xo,co keep switching between views, like bouncing around from the old canvas to tacmap to new canvas or whatnot, each time they open up a new canvas it would need to flatten the image to display it as the background to the canvas (the idea was once it's flattened when a member of cic opens it there is a timer preventing this sort of unnecessary accidental spam from switching between views) .That's what I wanted to avoid, but if you're saying it's not a big deal then sure, can do it that way. Maybe I'm misunderstanding - and yeah it doesn't do that, once you leave the the canvas it wipes it.

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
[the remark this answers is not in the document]
Cthulhu80: I doubled check to make sure, unless i'm testing it incorrectly.
So, i'm talking about something like this.
Queen joins the game, ovis.
Queen opens up a tacmap which brings her to the home view
Second xeno joins the game (join as guest or something)
Second xeno opens up a tacmap while queen is on ovi
if it's just hive.tacmap.tgui_interact(src) (the old implementation) because of the shared nature of the current_menu the queen gets kicked out of the home view and is now looking at the old canvas view.
So yes, the concern was about manipulating the UI, they should be separate and such I made two distinct tacmaps, one for the queen and another for the normal xenos. I'm making some changes here to allow the xenos to see the byondui tacmap as well, so now i'm overriding tgui_interact and making some minor changes which will be visible in the next commit.

Cthulhu80:
> Who are these other maintainers anyways? If its Harry or Morrow telling you a way, there isn't anyone else but a host that can change that decision. I also highly doubt what you were told was a demand.
> What I wrote here does not differ with what Morrow expected after your message in development, and ultimately its Morrow's call. The concept originally was for this to replace announcements with Gyazo links. Having an announcement that opens a TGUI map rather than a web browser to gyazo or imgur of an image was the expectation. image Regardless it feels weird for it to go to chat (thats what it does now - you have a chat announcement), but then there's no link map or anything to view. My recommendation is to sort out visors in a different PR. Just get the drawing on a map at all working in this PR. Whether you want to do this via an action button at the top of the screen (like the ghost button when you are dead), or just a link in chat is up to you.
- they were bickerin' about it a few lines under the image you sent

Screen Shot 2023-09-24 at 3 23 32 PM

> My suggestion is once at round start for all factions w/ tacmaps, not every time the interface is opened by anyone. One way to do this would be where its global define is created because it accepts a function call as its initializer. In otherwords there is never a time the "old canvas" data is empty. It only needs to change when a CO/Queen change it.
- ok, i'll look into it. 

> I think it best that the drawing is a snapshot of the state of things, but then like on master xenos should still have access to the live tacmap like they always have (its just the first tab like you gave CO). If you're intending to be doing a getFlatIcon of the tacmap every few seconds thats going to be a lot of performance loss and may be significant memory wise for a particularly long round. It doing that operation on demand when the CO/Queen makes a new map is fine though.
- yup, xenos should have access to the original byondui tacmap after this next commit.

code/controllers/subsystem/minimap.dm
Drulikar: > that's what I mean, say a so,xo,co keep switching between views, like bouncing around from the old canvas to tacmap to new canvas or whatnot, each time they open up a new canvas it would need to flatten the image to display it as the background to the canvas (the idea was once it's flattened when a member of cic opens it there is a timer preventing this sort of unnecessary accidental spam from switching between views, it will grab the original one. In a few minutes it could then be flattened again for viewing by cic ) .That's what I wanted to avoid, but if you're saying it's not a big deal then sure, can do it that way. Maybe I'm misunderstanding - and yeah it doesn't do that, once you leave the the canvas it wipes it.

What you are drawing on should be what you get. In other words, what you are drawing on should have the icons from the tacmap with people in various places. I wouldn't stress too much about a CO ending up creating a flat icon (just be sure you don't keep that old image around so it can be garbage collected). But the solution to use your cooldown lockout would be to prevent accessing that page (or it refusing to draw and update the tacmap icons) until the cooldown is complete.

I understand what you were trying to do, but even still I don't imagine it to be very intuitive if you initially open to a blank map. What you draw on should already have the icons. It shouldn't need them to opt to update the map then start drawing when it allows them to draw on a blank map.

However you could do a mix of these two things: 
1. Have it "update" the map when that page is opened.
2. Immediately have the update cooldown start.
3. Separate the announce button from the update button.
4. Consider reducing it to I don't know 2 minutes.
5. You don't really need to concern yourself with an announcement cooldown in addition to the update cooldown; but if you really want to then have at it.

Drulikar: Youve already seen from discord, but just for transparency, here are the patches to get getFlatIcon to respect RESET_COLOR appearance_flags:
image

Dimension fixes + color fixes: 
[patch_all.diff.txt](https://github.com/cmss13-devs/cmss13/files/12713913/patch_all.diff.txt)
Only color fixes: 
[patch_color.diff.txt](https://github.com/cmss13-devs/cmss13/files/12713915/patch_color.diff.txt)

Do note that this means there could potentially be 2xLayer (each blip is a couple layers) more blends being performed, so this can potentially be that much more expensive of a call to perform. I still need others to evaluate what I've experimented with to see if theres anything that can be done to do it more optimally.

Cthulhu80:
major issues should be sorted, ironing out the edge cases.

Cthulhu80:
https://github.com/cmss13-devs/cmss13/assets/122310258/193810c8-12ce-4a3f-8588-7cc8261b2079

Cthulhu80:
https://github.com/cmss13-devs/cmss13/assets/122310258/bbe1435c-d5ee-4109-8df4-a9454f1e70da

I know you said leave the visor stuff for another pr, but it should close properly now

cm13-github:

cm13-github:

Cthulhu80:
Real.

cm13-github:

cm13-github:

Drulikar (changes requested):

tgui/packages/tgui/interfaces/TacticalMap.tsx
```diff
@@ -1,24 +1,288 @@
...
+  'LV-624': 'https://cm-ss13.com/w/images/6/6f/LV624.png',
+  'Ice Colony': 'https://cm-ss13.com/w/images/1/18/Map_icecolony.png',
+  'Whiskey Outpost': 'https://cm-ss13.com/w/images/7/78/Whiskey_outpost.png',
+  'Solaris Ridge': 'https://cm-ss13.com/w/images/9/9e/Solaris_Ridge.png',
+  'Fiorina Science Annex':
+    'https://cm-ss13.com/w/images/e/e0/Prison_Station_Science_Annex.png',
+  'Trijent Dam': 'https://cm-ss13.com/w/images/9/92/Trijent_Dam.png',
+  'Sorokyne Strata': 'https://cm-ss13.com/w/images/2/21/Sorokyne_Wiki_Map.jpg',
+  'Kutjevo Refinery': 'https://cm-ss13.com/w/images/0/0d/Kutjevo_a1.jpg',
+  "LV-522 Chance's Claim": 'https://cm-ss13.com/w/images/b/bb/C_claim.png',
```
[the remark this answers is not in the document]
Cthulhu80: just made it static on the tacmap.

Drulikar (changes requested):
I manged to get some time to test this in game. We're very close to this being ready for a test merge, but please try to tackle atleast the issues I mention that aren't minor. The bolded one is ultimately why I am not just TMing it now:
- New but minor: Initial wiki map is cleared merely by a CO/SO opening their tablet/computer (not even by submitting any new tactical map) it should stay on that wiki map until an actual tacmap update is sent.
- **Related to above: Opening the tacmap when flatten cooldown is satisfied results in everyone's tacmaps getting the new positional data from the flat map (but old svg). The last drawn map should only change when the CO/SO/whoever submits it.**
- New (or atleast I didn't catch it before) but minor: You can scroll on the navigation part of the tacical map options and it will make the tabs dissapear. Likely just you have a lot of extra horizontal buffer space that shouldn't be there.
- Minor/Polish: First time opening the tactical map via status panel link _still_ requires them to click on the tab to be displayed a map (but seems to persist showing the map on open for subsequent opens)
- Undo/Clear canvas buttons are _still_ unresponsive: Basically you have to un-highlight them by clicking empty space and then click them again for them to be more responsive, you can't just multi click the undo button and get it to actually undo.
- Performing an undo/clear far enough to a black color _still_ will make the color selection dropdown out of sync with what color will actually be drawn.
- Minor: If a marine keeps their window open, a new tacmap update _still_ doesn't refresh the data on the map. Nor does clicking the view tacmap button in status bar cause the window to update, and the tab stays depressed so it can't be clicked again to cause a refresh. You can only close it and re-open it.
- Two cooldowns (one of which is hidden to the CO) is funky for updating vs announcing the map. (Its also kinda incorrect that the progress bar say time to update map when its actually the time to announce). It also means that clicking update canvas _usually_ does nothing. If you want a way to refresh the flat view it ought to be its own button with its own cool down represented - not tied into the action of announcing. If you merely just want a two button press to announce the map, then make it more along the lines of what https://github.com/cmss13-devs/cmss13/blob/master/tgui/docs/component-reference.md#buttonconfirm does (To do that you merely would change the text for announce to `Confirm?` and change the text Update Canvas to `Announce`).

Drulikar:
Forgot to send the gif of the silly scrolling bug: 
scroll

Cthulhu80:
Please don't quote me on this, but I'm fairly certain most of the issue should be resolved. Will be testing some more just for good measure.

Drulikar:
All the merges with master are unnecessary. Only if there's something merged that you think could affect your local testing of this branch, or if there's a conflict to be resolved.

BroseidonMOO:
It is critical that Staff have the ability to see tacmap history to see who drew what before this is merged.

Wintermote:
Not sure why this included removal of tacmaps from SO consoles

genuinely made the role unplayable for me, the tacmaps were that important for the job

Drulikar:
> It is critical that Staff have the ability to see tacmap history to see who drew what before this is merged.

image

Drulikar:
> Not sure why this included removal of tacmaps from SO consoles
> 
> genuinely made the role unplayable for me, the tacmaps were that important for the job

I'm skeptical that moving two tiles makes the role unplayable, but as far as I understand the need for them to no longer have access has been rectified.

Drulikar:
image

Step count reduced by two!

Cthulhu80:
> Not sure why this included removal of tacmaps from SO consoles
> 
> 
> 
> genuinely made the role unplayable for me, the tacmaps were that important for the job

Resolved

code/modules/mob/dead/observer/observer.dm
```diff
@@ -898,6 +902,23 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
...
+
+	GLOB.uscm_tacmap_status.tgui_interact(src)
+
+/mob/dead/observer/verb/view_xeno_tacmap()
+	set name = "View Xeno Tacmap"
+	set category = "Ghost.View"
+
+	var/datum/hive_status/hive = GLOB.hive_datum[XENO_HIVE_NORMAL]
+	if(!hive || !length(hive.totalXenos))
+		to_chat(src, SPAN_ALERT("There seem to be no living hives at the moment"))
```
fira, line 917: This only checks XENO_HIVE_NORMAL so that wording is incorrect

html/statbrowser.js
```diff
@@ -369,6 +371,13 @@ function draw_status() {
 			document
 				.getElementById("statcontent")
 				.appendChild(document.createElement("br"));
+		} else if (
+			// hardcoded for testing purposes .includes() seems to be breaking things for some reason.
```
fira, line 375: comment?

fira (approved):
Code LGTM although I'm not the best tgui reviewer

harryob (changes requested):

html/statbrowser.js
fira: to clarify: this is final code, so either it's hardcoded for good (and not testing purposes) or this is still an outstanding problem that needs to be fixed

tgui/packages/tgui/interfaces/TacmapAdminPanel.js
```diff
@@ -0,0 +1,168 @@
...
+          tooltipPosition="bottom"
+          ml={0.5}
+          onClick={() =>
+            act('recache', {
+              uscm: is_uscm,
+            })
+          }
+        />
+      }>
+      <Fragment>
```
mullenpaul, line 119: Fragments can be denoted as
`<> </>`

mullenpaul (approved):
Things are generally OK. I can see some future work making this a bit more flexible but I would rather get this in as it stands than knock it back waiting for polish.

code/controllers/subsystem/minimap.dm
```diff
@@ -338,6 +338,170 @@ SUBSYSTEM_DEF(minimaps)
...
+ * Fetches the datum containing an announced flattend map png reference.
+ *
+ * Arguments:
+ * * faction: FACTION_MARINE or XENO_HIVE_NORMAL
+ */
+/proc/get_tacmap_data_png(faction)
+	var/list/map_list
+
+	if(faction == FACTION_MARINE)
+		map_list = GLOB.uscm_flat_tacmap_data
```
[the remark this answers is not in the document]
Drulikar: I do not recommend it until other factions are more common in the main loop, or there's a different game mode that is used enough that could benefit from command/queen outlining more of a plan.

- Its intentional that the data is not sent to the other factions to avoid finagling some way to get the enemy tactical map (but who knows whats possible when you're going to those lengths or if what I did to try to separate the data from marines <-> xenos by not entering the opposite faction in the client list for ssassets even works how I intended).

- Additionally, ghosts really ought to then have a single unified view tacmap interface like admins do. I'd prefer not to give ghosts access to the admin panel even if restricted to keep better separation; but right now they just have two verbs that open that faction's interface as if they were a player (just with more limited functionality - a single page).

code/controllers/subsystem/minimap.dm
```diff
@@ -338,6 +338,170 @@ SUBSYSTEM_DEF(minimaps)
...
+	else if(faction == XENO_HIVE_NORMAL)
+		if(!COOLDOWN_FINISHED(GLOB, xeno_flatten_map_icon_cooldown))
+			return FALSE
+		COOLDOWN_START(GLOB, xeno_flatten_map_icon_cooldown, flatten_map_cooldown_time)
+	else
+		return FALSE
+
+	var/icon/flat_map = getFlatIcon(map_holder.map, appearance_flags = TRUE)
+	if(!flat_map)
+		to_chat(usr, SPAN_WARNING("A critical error has occurred! Contact a coder.")) // tf2heavy: "Oh, this is bad!"
```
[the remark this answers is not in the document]
Drulikar: This was originally more of an IC message that was changed to be what it is now because this shouldn't happen (nor have I seen this ever get hit) because I felt it made it seem like an intended problem and it is not an intended problem.

cm13-github:

cm13-github:

code/modules/admin/tacmap_panel/tacmap_admin_panel_tgui.dm
```diff
@@ -0,0 +1,145 @@
...
+				selected_svg = GLOB.xeno_svg_tacmap_data[xeno_selection + 1]
+			selected_svg.svg_data = null
+			last_update_time = world.time
+			message_admins("[key_name_admin(usr)] deleted the <a href='?tacmaps_panel=1'>tactical map drawing</a> by [selected_svg.ckey].")
+			return TRUE
+
+/datum/tacmap_admin_panel/ui_close(mob/user)
+	. = ..()
+	uscm_selection = -1
+	xeno_selection = -1
```
harryob, line 145: these -1s look like special numbers that should probably be defines

code/modules/mob/new_player/new_player.dm
```diff
@@ -146,11 +146,12 @@
 
 				if(observer.client)
 					observer.client.change_view(world_view_size)
+					send_tacmap_assets_latejoin(observer)
```
harryob, line 149: these snowflake calls really shouldn't be necessary, it should be totally viable to send these pngs when required through SSassets

harryob (changes requested):
just this

code/modules/mob/new_player/new_player.dm
Drulikar: I need a suggestion on where you want these instead then. I can confirm it is necessary to send these assets for latejoiners - else they will have a wikimap as fallback because they don't have the current tacmap. Only until they've been present for a tacmap update will they then have a tacmap drawing thats on a flattened tactical map.

Both `tgui_interact` and `ui_static_data` are called every time the respective interface is opened by that user, so to not oversend the maps unnecessarily we'd have to either check a var stored on the mob or client, or look them up in some global list.

As they are, they are then only sent when needed. They currently are only unnecessarily sent if the player is sent back to lobby and they late join or observe again.

Drulikar: Let me know if you meant something else by this.

harryob (approved):

Issue #4445, opened 2023-09-19
Pressurized Reagent Canister disappears if removed from pouch if hands are full


RavinHobo:
### Testmerges

2f35bde249780643b5468a6767e4016b6242dc29

### Description of the bug

Pressurized Reagent Canister disappears if removed from pouch if hands are full.

### What's the difference with what should have happened?

It should be dropped on the floor. 

### How do we reproduce this bug?

1.have hands full
2.right click on pouch
3.click remove cannister
4. canister is now evaporated
...

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #4491, opened 2023-09-24
Bugfix: Reagent Canister Nullspace

Crowbar764:
# About the pull request

Removing the container from a reagent canister with your hands full would cause the container to be nullspaced. Now it is instead dropped on the floor.

Fixes https://github.com/cmss13-devs/cmss13/issues/4445

<details>

https://github.com/cmss13-devs/cmss13/assets/59719612/c2187a7d-95e9-4fc2-9a42-3ee7b0192b02

</details>

# Changelog

:cl: Casper
fix: fixed reagent canister being deleted if ejected with your hands full
/:cl:

### DIFF
--- a/code/game/objects/items/storage/pouch.dm
+++ b/code/game/objects/items/storage/pouch.dm
@@ -1052,7 +1052,10 @@
 		to_chat(usr, SPAN_WARNING("There is no container inside this pouch!"))
 		return
 
-	usr.put_in_any_hand_if_possible(inner, disable_warning = TRUE)
+	var/had_empty_hand = usr.put_in_any_hand_if_possible(inner, disable_warning = TRUE)
+	if(!had_empty_hand)
+		usr.drop_inv_item_on_ground(inner)
+
 	inner = null
 	update_icon()
 


code/game/objects/items/storage/pouch.dm
```diff
@@ -1052,7 +1052,10 @@
 		to_chat(usr, SPAN_WARNING("There is no container inside this pouch!"))
 		return
 
-	usr.put_in_any_hand_if_possible(inner, disable_warning = TRUE)
+	var/had_empty_hand = usr.put_in_any_hand_if_possible(inner, disable_warning = TRUE)
```
morrowwolf, line 1055: Use /mob/proc/put_in_hands(obj/item/W, drop_on_fail = TRUE)

Should handle dropping if there's no free hand.

morrowwolf (changes requested):

code/game/objects/items/storage/pouch.dm
Crowbar764: That is the first thing I tried but it doesn't work because that function resets the item's plane to its initial value which for this object is not the dropped item plane, see video:


https://github.com/cmss13-devs/cmss13/assets/59719612/b3363f98-bca6-4cfc-ad8e-90374f489d51

morrowwolf: Huh, right you are. You could probably put the proc in the if but this is fine.

PR #4492, opened 2023-09-24
no more anchorpoint pvt

QuickLode:
# About the pull request

hi this mainly fixes intended timelocks for 2 vital CMB roles, and removes Pvts from Anchorpoint QRF

Had a rank/timelock error due to the way the root was set, this alleviates it so no more Pvt leading the entire squad LMAO(In hindsight I feel like a Pvt is out of place to be on a QRF being sent from anchorpoint, and instead newer troops would stay on the station, may readd later tho)
Also the intended 5hr SL timelock was placed on CMB Mar to prevent weird incidents which have happened. 1hr for ICC for CL duties. 

also I'm not sure why but it seems black market CMB may be having problems with their hud? i will add it if I find a problem
bug bad 

<details>

</details>

# Changelog
:cl:
del: removed Pvts from Anchorpoint QRF (rip)
spellcheck: fixed a typo in CMB call-in
/:cl:

### DIFF
--- a/code/game/jobs/job/special/cmb.dm
+++ b/code/game/jobs/job/special/cmb.dm
@@ -6,11 +6,6 @@
 /datum/job/special/cmb/marshal
 	title = JOB_CMB_TL
 
-AddTimelock(/datum/job/special/cmb/marshal, list(
-	JOB_SQUAD_LEADER = 5 HOURS,
-))
-
-
 // CMB Investigative Synthetic
 /datum/job/special/cmb/synthetic
 	title = JOB_CMB_SYN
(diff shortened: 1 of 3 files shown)


code/game/jobs/job/special/cmb.dm
```diff
@@ -6,6 +6,11 @@
 /datum/job/special/cmb/marshal
 	title = JOB_CMB_TL
 
+AddTimelock(/datum/job/special/cmb/marshal, list(
```
morrowwolf, line 9: As far as I can tell neither of these timelocks actually talk to the emergency_call system. Instead it uses a number in emergency_call: time_required_for_job. It is default 5 hours and would need to be custom set, with both check_timelock() calls in the CMB emergency_call.

morrowwolf: All that to say, they're both currently timelocked for 5 hours for CL and SL, as far as I can tell, already

morrowwolf (changes requested):

PR #4494, opened 2023-09-24
Fixes grab delay after performing an action

Git-Nivrak:
# About the pull request
First there was an issue where you could grab a marine behind a cade to bypass grab delay and hit it faster than intended.
Which was then fixed by #4204 which in turn fucked your ability to grab things without delay and then attempted to be fixed by #4246 which only partially fixed it.

This PR fixes it with a simpler solution that doesn't affect any other click delays, or actions. By removing a mechanic which automatically attacks the "barricade" (can also be a mob, just the general name for something in the way of you grabbing someone else) when attempting to grab it fails.

Fixes an annoying bug that's been up for a while
<details>

</details>

# Changelog
:cl:
balance: xenos no longer have grab delay via any grab method
/:cl:

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -210,7 +210,7 @@
 		return TRUE
 
 	if (mods["ctrl"])
-		if (Adjacent(user) && user.next_move < world.time)
+		if (Adjacent(user))
 			user.start_pulling(src)
 		return TRUE
 	return FALSE
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -126,10 +126,11 @@
 	if (!isturf(loc))
 		return
 
-	if (world.time <= next_move && A.loc != src) // Attack click cooldown check
+	if (world.time <= next_move && A.loc != src && src.a_intent != INTENT_GRAB) // Attack click cooldown check
 		return
 
-	next_move = world.time
+	if (src.a_intent != INTENT_GRAB)
+		next_move = world.time
 	// If standing next to the atom clicked.
 	if(A.Adjacent(src))
 		click_adjacent(A, W, mods)
@@ -158,7 +159,8 @@
 			W.afterattack(A, src, 1, mods)
 	else
 		if(!isitem(A) && !issurface(A))
-			next_move += 4
+			if (a_intent != INTENT_GRAB)
+				next_move += 4
 		UnarmedAttack(A, 1, mods)
 
 /mob/proc/check_click_intercept(params,A)
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -126,11 +126,14 @@
 	if (!isturf(loc))
 		return
 
-	if (world.time <= next_move && A.loc != src && src.a_intent != INTENT_GRAB) // Attack click cooldown check
+	if(A.Adjacent(src) && src.a_intent == INTENT_GRAB && !W)
+		click_adjacent(A, W, mods)
 		return
 
-	if (src.a_intent != INTENT_GRAB)
-		next_move = world.time
+	if (world.time <= next_move && A.loc != src) // Attack click cooldown check
+		return
+
+	next_move = world.time
 	// If standing next to the atom clicked.
 	if(A.Adjacent(src))
 		click_adjacent(A, W, mods)
@@ -158,9 +161,8 @@
 
 			W.afterattack(A, src, 1, mods)
 	else
-		if(!isitem(A) && !issurface(A))
-			if (a_intent != INTENT_GRAB)
-				next_move += 4
+		if(!isitem(A) && !issurface(A) && src.a_intent != INTENT_GRAB)
+			next_move += 4
 		UnarmedAttack(A, 1, mods)
 
 /mob/proc/check_click_intercept(params,A)
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -126,7 +126,9 @@
 	if (!isturf(loc))
 		return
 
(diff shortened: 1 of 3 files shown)


Drulikar:
This also has the side effect of removing a few ways xenos bonk into barricades.

Git-Nivrak:
> This should also has the side effect of removing a few ways xenos bonk into barricades.
> 
> _Edit_ In testing though runner/lurker pounce still has the bonk handled a different way so that was my main concern. Warrior lunge now just doesn't hit the barricade.

If that's the case imo it's fine, About warrior I think that not having an unintentional slash is actually a good thing all in all

morrowwolf (changes requested):
Alright so before this is merged we need to make sure ctrl-click and grab intent click are functioning effectively the same way.

My current understanding with this change is grab intent click is still going to have the delay.

Still thinking about it otherwise. I don't have a huge issue with the ctrl-click delay 🤷

Git-Nivrak:
That is the way it was before the initial bugfix for the infinite slash problem, that is ctrl + click not having delay and grab intent click having delay, so I brought it back in line.
Now either way I think it's important to either leave it like that or make grab click have no delay as well because grabbing + disarming is something that is used a lot and the delay makes it feel really bad (Suppose you disarm someone, then you grab them to devour then by the time you are able to click them again you lost a disarm) or if you are trying to grab someone mid combat which is already hard when moving then it just doesn't go through because you have a delay from disarming / slashing.
Let me know what you decide

Drulikar:
Morrow is saying if Ctrl+click grab is not going to have a delay, then grab intent needs to not have a delay. Currently without this pr they behave the same. If this pr would be merged as is, then once again Ctrl click would be delayless and grab click would have a delay.

Git-Nivrak:
Now they are both delay-less

code/_onclick/click.dm
```diff
@@ -126,10 +126,11 @@
 	if (!isturf(loc))
 		return
 
-	if (world.time <= next_move && A.loc != src) // Attack click cooldown check
+	if (world.time <= next_move && A.loc != src && src.a_intent != INTENT_GRAB) // Attack click cooldown check
```
Drulikar, line 129: I don't have a *firm* confirmation exactly what is now exploitable by simply using grab intent when attacking, but the next_move delay must only be removed for pulling; or we keep things as it is where ctrl+click and grab intent both require next_move delay to be satisfied.

The only feasible way I personally can think to do this is to revert this change, revert the change on 132/133, possibly revert the change on 162/163 and rip all pulling logic out of click_adjacent to be a new proc called (e.g. click_adjacent_grab) around just before this at around line 128.

Drulikar (changes requested):

code/_onclick/click.dm
Git-Nivrak: This is not exploitable because it only applies to grab e.g:
You grab someone then disarm him then grab him again, if you try to disarm now it will still fail because of the next_move delay
this is because grabbing does not reset your move delay, it just ignores it so whatever delay you had before will still be there

Drulikar: Please actually test your changes and confirm what you are saying rather than dismissing feedback. This is on your branch even after locally merging master into it (to rule out maybe some bug fix that master fixed). This PR was also TM'd and there were reports of exploited delay-less attacks (thus why I said I didn't have firm confirmation since I personally didn't witness it). The same is not reproduce-able on master. See below steps and reproduction video.

Steps:
- Select grab intent
- Click an adjacent enemy with a weapon as fast as possible - you can attack without delay because this PR merely checks if intent is grab - not that the resulting action will be a pull.

https://github.com/cmss13-devs/cmss13/assets/76988376/5ec06af1-3067-4575-97b4-e49c5a500c48

Drulikar:
Previous feedback is not resolved.

Git-Nivrak:
Shouuuld be all good now
Tested the following:
Grab intent into disarm on xeno -> No delay 
Disarm into grab intent on xeno -> No delay 
Disarm into grab into disarm again -> Delay on second disarm 
No infinite bashing with items on grab intent

code/_onclick/click.dm
```diff
@@ -126,6 +126,10 @@
 	if (!isturf(loc))
 		return
 
+	if(A.Adjacent(src) && src.a_intent == INTENT_GRAB && !W)
```
Drulikar, line 129: Since this is hot code, please put the adjacent check into a var so we can just check the var rather than calling Adjacent again on line 134/138.

Ideally we correct all the single var usage too, but most of the proc's code would need to change to accomplish that.

Drulikar (changes requested):
Issue above no longer appears to be reproduceable. This should now be suitable to TM.

code/_onclick/click.dm
Git-Nivrak: Done

Drulikar (approved):
Single letter vars bug me but otherwise this needs testing to make sure there isn't something else I didn't catch.

Drulikar (changes requested):
Sorry for the delay in getting this TM'd and tested more, but it seems the grab intent rapid attacking of cades and fences is back. I can see about getting a reproduction in a bit if you like.

Drulikar:
https://github.com/cmss13-devs/cmss13/assets/76988376/c3019765-ea04-42f9-940c-ee81b028a209

Should be pretty easy to reproduce just using the grab intent and click objects.

Drulikar:
Please work out a way that both ctrl+click and grab intent behave the same - not that one way is superior to the other.

PR #4505, opened 2023-09-25
Get Flat Icon Fixes

Drulikar:
# About the pull request

This PR fixes two issues with getFlatIcon: It failing to properly resize its template dimensions to whatever icon dimensions it was passed, and it ignoring appearance_flags for RESET_COLOR and RESET_ALPHA. This was done to address issues with #4475 but can be even more performance costly than it already was because now it applies color and alpha blends per overlay/underlay layer rather than once per recursive call if there is a color/alpha to apply.

Fixes the broader potential uses for getFlatIcon such as flattening the tacmap.

<details>

Example of tacmap flattened prior to dimension fixes: 
image

Example of tacmap flatted prior to color fixes: 
image

Example of tacmap properly flattened: 
image

</details>

# Changelog
:cl: Drathek
fix: Fix getFlatIcon not resizing its template nor respect appearance_flags of RESET_COLOR and RESET_ALPHA
/:cl:

### DIFF
--- a/code/__HELPERS/icons.dm
+++ b/code/__HELPERS/icons.dm
@@ -435,11 +435,19 @@ world
 			if(layer_image.alpha == 0)
 				continue
 
+			var/apply_color = TRUE
+			var/apply_alpha = TRUE
+
 			if(layer_image == copy) // 'layer_image' is an /image based on the object being flattened.
 				curblend = BLEND_OVERLAY
 				add = icon(layer_image.icon, layer_image.icon_state, base_icon_dir)
 			else // 'I' is an appearance object.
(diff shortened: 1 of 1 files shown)


morrowwolf (changes requested):
So I'm not 100% on how the changes are working *but* would it be possible to put the more expensive stuff that's only needed for some things behind an argument or similar?

Drulikar:
With appearance_flags default FALSE
default

If I were to make the default TRUE (so all use the fix):
opt in

No changes that I can tell - but should be the older performance. The new tacmap PR must opt in to this fix else all icons are colored incorrectly. I did also try to take a stab at why the vents appear layered wrong in photos, but I'm not sure yet why they don't sort correctly. TG's photo method is drastically different.

harryob (approved):

PR #4516, opened 2023-09-26
makes LZ detector work while landed

cuberound:
# About the pull request

This PR makes it so that LZ detector works all the time not only when in air.

The fact that LZ detector worked only when in air made it near useless, when you are already headed for the LZ what good is it to see it overrun and how is seeing xeno in FOB after you leave it any good. This turns LZ detector into a tool for transport pilot to better evaluate the situation in FOB before taking off from it or heading to it.

# Changelog

:cl:

balance: LZ detector now works even when the DS is stationary

/:cl:

### DIFF
--- a/html/changelogs/AutoChangeLog-pr-4636.yml
+++ b/html/changelogs/AutoChangeLog-pr-4636.yml
@@ -1,0 +1,4 @@
+author: "Drathek"
+delete-after: True
+changes:
+  - refactor: "Refactored camera code to be less blocking, use typechecks less often, and provide somewhat more fluid descriptions to photos."


morrowwolf (changes requested):
Needs to turn off the linked network when the LZ detector is taken off the ship

cuberound:
I am a fucking bozo, did not test melting it while not installed and was trying to acces properties of fucking NULL xD

morrowwolf (changes requested):
update_equipment() is called more than just taking it on and off so when you're picking it up and such it will runtime as there's no linked_cam_console at line 529

cm13-github:

PR #4542, opened 2023-09-28
[QOL] You can acid things on surfaces

TeDGamer:
# About the pull request

Allows aciding of items again on surfaces.

Currently, it says you're too far even if you're adjacent to the item and surface the item is on.

Starts the progress to acid but then says you're too far away. This fixes that.

I remember being able to acid items on tables a while back.

Would alt click and then select the items to acid.

Without this, you need to destroy the surface first before aciding the items which is fine but annoying. 

I don't think this would change things too much destroying things on the tables without needing to touch the table. Just a minor QOL / fix from what it used to be like.

# Changelog
:cl:
qol: You can acid objects on surfaces again
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/ability_helper_procs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/ability_helper_procs.dm
@@ -119,7 +119,7 @@
 	if(!check_plasma(plasma_cost))
 		return
 
-	if(!O.Adjacent(src) || (I && !isturf(I.loc)))//not adjacent or inside something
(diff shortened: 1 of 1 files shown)


fira (changes requested):
You'll probably want to also apply that to `/mob/living/carbon/xenomorph/runner/corrosive_acid` (or be the hero that fixes the fact it's doubled up)

TeDGamer:
> You'll probably want to also apply that to `/mob/living/carbon/xenomorph/runner/corrosive_acid` (or be the hero that fixes the fact it's doubled up)

Damn, didn't realize. I'll take a look.

Issue #4309, opened 2023-09-01
Zombies can modify Sec Records


drawsstuff:
### Testmerges

#4246 #4224 #4063 #4290

### Description of the bug

Zombies can modify security records

### What's the difference with what should have happened?

Undead colonists (lacking ID mind you) should not be able to do this

### How do we reproduce this bug?

1.Spawn in as zombie
2.Kill your desired Unga Dunga (in this case I am unable to grab details because my chat log is full of zombies doing push ups
3.Examine them
4.Description should come up, with the relevant "add" button near security records
...

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

drawsstuff:
Attaching relevant recording 

https://github.com/cmss13-devs/cmss13/assets/16404218/fbd3e721-d6ce-427b-926e-ab1e1c7c3a64

Huffie56:
can't reproduce it
![image](https://github.com/cmss13-devs/cmss13/assets/117036822/94ad57cd-02f0-4bfe-b185-0d45439d5af4)

probably was caused because your zombie eyes worked as security HUD....?

Cthulhu80:
@Huffie56 it was fixed in the mentioned pr.

Huffie56:
feel free to tag me if anyone can confirmed the bug as happen recently(current month.)

Cthulhu80:
The issue isn't just zombies. If a user puts on hud glasses and then removes them it still displays it. Same thing with medical.

Huffie56:
> The issue isn't just zombies. If a user puts on hud glasses and then removes them it still displays it. Same thing with medical. Is this intended?

i need more info to respond but if you think their is a more general issue(bug or inconsistency (with the code or in term of logic)  with hug glasses you should open an issue for it.
thanks for confirming this bug is real and fixing it.

PR #4558, opened 2023-09-30
TGUI / Bugfixes / Improvements: Security Cleanup

Crowbar764:
# About the pull request

- Converts Security Records console to TGUI (inspired by /TG/'s design).
- Fixed printed fingerprint reports run-timing when read.
- Fixed newly created records having incorrect gender casing.
- Fixed the player's wanted status being set to NULL if the SecHud dropdown was cancelled.
- Fixed zombies from being able to modify security records (fixes: https://github.com/cmss13-devs/cmss13/issues/4309).
- Cleaned up the SecHud chat text.
- Cleaned up printed-out data (records, fingerprints, round-start reports).
- Refactors some things for standards and performance.
- Fixed some other bugs I can't remember.

For optimisation, some data is not always accurate when modified from an outside source. I do not want to refresh the data of 100+ players every tick.

Screenshots supplied in the dropdown.

Modern UI. Fixes some broken features. Easier to read SecHud

<details>

Main UI:
image

Comments tab:
image

Forensics scanner with prints added (bottom left):
image

SecHud:
image

Printed record:
image

</details>

# Changelog

:cl: Casper
fix: fixed record-related bugs
ui: added security records TGUI
refactor: cleaned-up SecHud text
/:cl:


Cthulhu80:
Screen Shot 2023-09-30 at 9 09 12 PM
noice

code/game/machinery/computer/security.dm
```diff
@@ -1,46 +1,321 @@
...
+			to_chat(user, "You insert [held_scanner].")
+			update_static_data_for_all_viewers()
 
 	..()
 
 /obj/structure/machinery/computer/secure_data/attack_remote(mob/user as mob)
 	return attack_hand(user)
 
-//Someone needs to break down the dat += into chunks instead of long ass lines.
+/obj/structure/machinery/computer/secure_data/ui_data(mob/user)
```
mullenpaul, line 34: All return paths for this proc should return something. Rather than explicitly having data as a list variable, just make the default return value an empty list and append data to it.

code/game/machinery/computer/security.dm
```diff
@@ -1,46 +1,321 @@
...
+
+	// Get security data, if it exists.
+	if (active_record_security)
+		data["active_record"]["criminal_status"] = active_record_security.fields["criminal"]
+		data["active_record"]["incidents"] = active_record_security.fields["incidents"]
+		data["active_record"]["comments"] = active_record_security.fields["comments"]
+
+	return data
+
+/obj/structure/machinery/computer/secure_data/ui_static_data(mob/user)
```
mullenpaul, line 62: Same as above for ui_data

code/game/machinery/computer/security.dm
```diff
@@ -1,46 +1,321 @@
...
+	for(var/datum/data/record/record_general_data as anything in GLOB.data_core.general)
+		var/crew_member = list()
+
+		crew_member["name"] = record_general_data.fields["name"]
+		crew_member["ref"] = "\ref[record_general_data]"
+		crew_member["rank"] = record_general_data.fields["rank"]
+
+		for(var/datum/data/record/record_security_data as anything in GLOB.data_core.security)
+			if (record_security_data.fields["id"] == record_general_data.fields["id"])
+				crew_member["criminal_status"] = record_security_data.fields["criminal"]
```
mullenpaul, line 81: are we ok with criminal status not being dynamically updated? With this in static data we would only see data refreshed if the UI was closed and reopened or hard refreshed

code/game/machinery/computer/security.dm
```diff
@@ -1,46 +1,321 @@
...
+			if (!istype(active_record_security, /datum/data/record) || !sanity_check(params["ref"]))
+				return
+
+			var/new_wanted_status = params["new_status"]
+			if(!new_wanted_status || !(new_wanted_status in WANTED_STATUSES))
+				return
+
+			active_record_security.fields["criminal"] = new_wanted_status
+
+			var/mob_gid = hex2num(active_record_general.fields["id"])
```
mullenpaul, line 202: I'm wondering, do we want to message admins, as a heads up, that someone has been flagged for arrest? It could help get eyes on a situation before it escalates too far

code/game/machinery/computer/security.dm
```diff
@@ -1,46 +1,321 @@
...
+				"created_by_name" = created_by_name,
+				"created_by_rank" = created_by_rank,
+				"created_at" = created_at,
+				"deleted_by" = null,
+				"deleted_at" = null
+			)
+
+			active_record_security.fields["comments"] += list(new_comment)
+
+			to_chat(ui.user, text("You have added a new comment to the Security Record of [].", active_record_security.fields["name"]))
```
mullenpaul, line 240: Do we want to log who changes the comments on a sec record? It might be belt and braces but it would help track down a griefer

code/modules/paperwork/filingcabinet.dm
```diff
@@ -99,30 +99,39 @@
  * Security Record Cabinets
  */
 /obj/structure/filingcabinet/security
-	var/virgin = 1
+	var/virgin = TRUE
```
mullenpaul, line 102: this is a bad variable name - I appreciate that it was like this but if we're touching the code can we give it a more useful name?

tgui/packages/tgui/interfaces/SecurityRecords/constants.tsx
```diff
@@ -0,0 +1,8 @@
+export const CRIMESTATUS2COLOR = {
```
mullenpaul, line 1: would it be possible to use scss rather than hard coded strings?

tgui/packages/tgui/interfaces/SecurityRecords/helpers.tsx
```diff
@@ -0,0 +1,16 @@
+import type { GenericRecord } from './types';
+
+/** Matches search by name or rank */
+export const isRecordMatch = (record: GenericRecord, search: string) => {
+  if (!search) return true;
+  const { name, rank } = record;
+
+  switch (true) {
```
mullenpaul, line 8: I'm not sure a switch case statement is really the way to do this

tgui/packages/tgui/interfaces/SecurityRecords/index.tsx
```diff
@@ -0,0 +1,435 @@
...
+            {active_record.criminal_status && (
+              <LabeledList.Item
+                verticalAlign="middle"
+                label="Status"
+                color={CRIMESTATUS2COLOR[active_record.criminal_status]}>
+                <Flex justify="space-between">
+                  <Flex.Item>{active_record.criminal_status}</Flex.Item>
+                  <Flex.Item>
+                    {wanted_statuses.map((status) => {
+                      let isActiveStatus =
```
mullenpaul, line 215: should use const over let

mullenpaul (changes requested):
Generally looks good - a few small things I feel that should be touched up. I don't know MP well enough to test it all out myself but we can try it on a TM. If you can address the minor issues at some point

code/game/machinery/computer/security.dm
Crowbar764: I used to do it that way and then I think Nanu taught me to do it this way 😆
I can change to the      .["blah"] = xyz      way if you prefer

code/game/machinery/computer/security.dm
Crowbar764: To get criminal status of every player, you have to search through the security records datacore for EVERY player to locate their record.
For 150 marine players that's just under and very roughly 22,500 checks (150 x 150), every UI update.
Doing this every tick seems like a disgusting waste of processing power for something so unimportant.

mullenpaul: That makes sense. One possibility would be to have a service go manage who is a criminal - which would reduce the scope but honestly that is way out of scope here

Cthulhu80:
fix this issue while you're at it, very simple fix. https://github.com/cmss13-devs/cmss13/issues/4309

tgui/packages/tgui/interfaces/SecurityRecords/constants.tsx
Crowbar764: I can't find anywhere in the code an example of where this has been done before. Are you sure it's possible / better than just using the string like all other components?

tgui/packages/tgui/interfaces/SecurityRecords/helpers.tsx
Crowbar764: Changed to if statements.

PR #4563, opened 2023-10-01
Egg runtime fix

JackieEstegado:
# About the pull request

Attacking an egg with a weapon that had no attack_verb would runtime the attack process, ending up not applying any damage.
Like a welder :|

Ja

# Changelog
:cl:
fix: Fixed runtime that could happen when melleing an egg.
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/egg.dm
+++ b/code/modules/cm_aliens/structures/egg.dm
@@ -208,7 +208,7 @@
 		return
 
 	user.animation_attack_on(src)
-	if(W.attack_verb.len)
+	if(W.attack_verb)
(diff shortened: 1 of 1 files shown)


code/modules/cm_aliens/structures/egg.dm
```diff
@@ -208,7 +208,7 @@
 		return
 
 	user.animation_attack_on(src)
-	if(W.attack_verb.len)
+	if(W.attack_verb)
```
morrowwolf, line 211: Do length(W.attack_verb) instead. Will stop the runtime and insure we have at least one attack_verb to choose from in the to_chat()

morrowwolf (changes requested):

morrowwolf (approved):

PR #4566, opened 2023-10-01
Fixes machinery coming out of largecrates not updating power

morrowwolf:
# About the pull request

Due to how forceMove() works it does not properly call area's Entered() proc as the areas are not changing when the contents are removed.

This is a hacky way to force add the machinery.

Non-updating machinery is lame

<details>

</details>

# Changelog

:cl: Morrow
fix: Fixed machinery coming out of largecrates not updating power
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/largecrate.dm
+++ b/code/game/objects/structures/crates_lockers/largecrate.dm
@@ -32,8 +32,8 @@
 		var/atom/movable/current_atom = contents[1]
 		current_atom.forceMove(current_turf)
 		if(istype(current_atom, /obj/structure/machinery))
-			var/obj/structure/machinery/current_machine = current_atom
-			current_machine.power_change()
+			var/area/current_area = get_area(src)
+			current_area.add_machine(current_atom)
 
 	deconstruct(TRUE)
 
--- a/code/game/objects/structures/crates_lockers/largecrate.dm
+++ b/code/game/objects/structures/crates_lockers/largecrate.dm
@@ -31,6 +31,9 @@
 	for(var/atom/movable/moving_atom in contents)
 		var/atom/movable/current_atom = contents[1]
 		current_atom.forceMove(current_turf)
+		if(istype(current_atom, /obj/structure/machinery))
+			var/obj/structure/machinery/current_machine = current_atom
+			current_machine.power_change()
 
 	deconstruct(TRUE)
 


fira:
there must be another explanation - forceMove is just a shortcut to doMove which triggers all the usual Crossed/Uncrossed/Exited/Entered

fira:
oh OK i get it, you're right, since the box itself (so its contents) are already in the area, it didn't change area as far as get_area() is concerned and Entered is not called

in my view this is a bug of doMove() which should base area logic off loc rather than get_area, but i'm not sure how safe this is to change

morrowwolf:
> oh OK i get it, you're right, since the box itself (so its contents) are already in the area, it didn't change area as far as get_area() is concerned and Entered is not called
> 
> in my view this is a bug of doMove() which should base area logic off loc rather than get_area, but i'm not sure how safe this is to change

So I came to similar conclusions, I was concerned with touching doMove() as it looked to be perf related? I'm happy to do it any way, might be worth checking for machinery in contents of movers when changing areas ??? but again that may be a big hit as well. This felt like the least likely to blow up everything solution 😭

fira:
The big question is, when you push a box, does its contents move and do they enter areas too?

Glanced at /tg/'s solution, doMove is identical but they have a system of "important recursive contents" which register to their parents for select state changes (like area transition in Entered) --> https://github.com/tgstation/tgstation/blob/master/code/game/area/areas.dm#L452

I think it's fine to leave it as hack for now because we don't have much use for all this stuff right now, might want to properly comment it 

If you want it to be more generic, you could bake it in /obj/structure/machinery/doMove so it applies in all cases

morrowwolf:
> The big question is, when you push a box, does its contents move and do they enter areas too?
> 
> Glanced at /tg/'s solution, doMove is identical but they have a system of "important recursive contents" which register to their parents for select state changes (like area transition in Entered) --> https://github.com/tgstation/tgstation/blob/master/code/game/area/areas.dm#L452
> 
> I think it's fine to leave it as hack for now because we don't have much use for all this stuff right now, might want to properly comment it
> 
> If you want it to be more generic, you could bake it in /obj/structure/machinery/doMove so it applies in all cases

So what my thought was is whenever a largecrate does like doMove() or whatever we add a check to pass Entered() to any machinery inside contents as well. I *think* that would do it but I'm not sure it'd be worth doing it that way when we can just send it once it opens not sure.

PR #4575, opened 2023-10-02
Adds discretionary arrests as capital crime

nauticall:
As part of the latest CO announcement, adds discretionary arrests into the JAS database as a capital crime. Will make this into a full PR once the trial is complete and successful.


realforest2001:
Full PR #4733

PR #4577, opened 2023-10-02
Fixes binocs destroying while zoomed FOR REAL

morrowwolf:
# About the pull request

Alright Morrow coming at you again with a hacky solution to a bug.

Due to how qdel callback works we don't get a user, if we're still zoomed though we feel we can *assume* that we are still in the direct contents of the user.

I am, admittedly, very sleepy so there could be a blatantly obvious answer that I'm missing and I hope there is. I'm not sure if it would be worth climbing locs up to an area or turf or whatever.

<details>

</details>

# Changelog

:cl: Morrow
fix: Fixed binocs destroying while zoomed
/:cl:


morrowwolf:
WAIT, I think I have a better solution with SIGNALS

fira:
registering through another obj is kinda cursed

i don't understand, can't you just make a /mob/proc/reset_zoom and call it?

PR #4584, opened 2023-10-04
Throws check tick all over the photo code

morrowwolf:
# About the pull request

So I have a general understanding of how this works in scheduling and all but I'm sure this could be done better.

This *should* help with the absolutely gnarly 3-4 second hard stops on the server when cameras are taking pictures of lots of stuff at 7 width.

Hard server freezes are bad (hypothetically)

<details>

</details>

# Changelog

:cl: Morrow
fix: Throws check tick all over the photo code to stop hard freezes
/:cl:

### DIFF
--- a/code/modules/paperwork/photography.dm
+++ b/code/modules/paperwork/photography.dm
@@ -184,6 +184,7 @@
 		for(var/atom/A in the_turf)
 			if(A.invisibility) continue
 			atoms.Add(A)
+			CHECK_TICK
 
 	// Sort the atoms into their layers
 	var/list/sorted = sort_atoms_by_layer(atoms)
@@ -209,6 +210,7 @@
 					xoff+=A:step_x
 					yoff+=A:step_y
 				res.Blend(IM, blendMode2iconMode(A.blend_mode),  A.pixel_x + xoff, A.pixel_y + yoff)
+				CHECK_TICK
 
 	// Lastly, render any contained effects on top.
 	for(var/turf/the_turf as anything in turfs)
@@ -218,6 +220,7 @@
 		var/image/IM = getFlatIcon(the_turf.loc)
 		if(IM)
 			res.Blend(IM, blendMode2iconMode(the_turf.blend_mode),xoff,yoff)
+			CHECK_TICK
 	return res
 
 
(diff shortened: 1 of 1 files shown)


code/modules/paperwork/photography.dm
```diff
@@ -184,6 +184,7 @@
 		for(var/atom/A in the_turf)
 			if(A.invisibility) continue
 			atoms.Add(A)
+			CHECK_TICK
```
Drulikar, line 187: Checking tick on just loops filtering out contents I'm skeptical is helpful.

code/modules/paperwork/photography.dm
```diff
@@ -262,6 +267,7 @@
 	var/list/turf/turfs = RANGE_TURFS(radius, target) & view(world_view_size + radius, user.client)
 	for(var/turf/T as anything in turfs)
 		mobs += get_mobs(T)
+		CHECK_TICK
```
Drulikar, line 270: Checking tick on just loops filtering out contents I'm skeptical is helpful.

code/modules/paperwork/photography.dm
```diff
@@ -218,6 +220,7 @@
 		var/image/IM = getFlatIcon(the_turf.loc)
 		if(IM)
 			res.Blend(IM, blendMode2iconMode(the_turf.blend_mode),xoff,yoff)
```
Drulikar, line 222:
```suggestion
			CHECK_TICK
			res.Blend(IM, blendMode2iconMode(the_turf.blend_mode),xoff,yoff)
```
The get flat icon just before this did a bunch of blends to worth checking tick before this blend.

Also try testing this by merging with master to get the getflaticon fixes pr in and you can better evaluate if this resolves any issues with large pictures (you may need to opt in to the more expensive get flat icon only for testing - not something the cameras should generally opt into unless we find cases where the extra work is needed for photos)

code/modules/paperwork/photography.dm
```diff
@@ -209,6 +210,7 @@
 					xoff+=A:step_x
 					yoff+=A:step_y
 				res.Blend(IM, blendMode2iconMode(A.blend_mode),  A.pixel_x + xoff, A.pixel_y + yoff)
+				CHECK_TICK
```
Drulikar, line 213: Harry when he sees single letter vars in the code context 👀

Drulikar (changes requested):

code/modules/paperwork/photography.dm
Drulikar: Also whitespace may be funky on mobile - not sure why this suggestion diff is red lining the existing code - just want a new line with check tick.

PR #4167, opened 2023-08-11
Changes to EZ Autoinjectors

BeagleGaming1:
# About the pull request

Removes skilled variant of emergency autoinjector and makes it non-refillable
Removes EZ autoinjectors from marine pouches

# Explain why it's good for the game

Makes Emergency autoinjectors more limited
Makes usage of autoinjectors require time, rather than being able to use while moving

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

I did not test it

</details>

# Changelog

:cl:
del: Removes emergency autoinjectors from squad medics
del: Removes skillless variant of the emergency autoinjector, makes default variant skillless
balance: Emergency autoinjectors cannot be refilled
balance: Replaces skillless autoinjectors in marine pouches with standard autoinjectors
/:cl:

PR #4593, opened 2023-10-06
Revert "Changes to EZ Autoinjectors (#4167)"

Git-Nivrak:
This reverts commit 49e512e5391707b1e1ad71181c9c170ec72acedc.
# About the pull request

The original PR had no proper reasoning
Either way waiting 5 seconds to use an injector is not good for the game and does not add anything and being able to use it without delay isn't incredibly broken either.

<details>

</details>

# Changelog
:cl:
balance: Reverts changes to EZ Autoinjectors
/:cl:


Drulikar:
> The original PR had no proper reasoning

It is in line with the project direction: https://forum.cm-ss13.com/t/maintainer-update-jun2023/2922
> First, we’d like to see rounds “slowed down”. Speed rounds are generally not satisfying for anyone. The average length we’d like to see is around 90-120 minutes. Thankfully, we already do a fairly good job on this but keeping it consistent and minimizing speedruns is going to be a priority. Giving time for more roleplay elements to thrive and evolve is important.

> Second, any additions to complexity for medical are likely to be approved. Our current system is pretty barebones and the skill/knowledge ceiling gives something to be desired for our more veteran players. This doesn’t explicitly need to be in the form of “realism.” The goal is to give more without making it impossible to understand. We plan on slowly introducing these new concepts and ideas so community knowledge can propagate nicely.

Also its goofy to be capturing a marine and then its injection sfx, injection sfx, injection sfx, injection sfx, injection sfx. Still other ways to easily OD, and a medic could still do the same, but its still a step in the right direction.

How does this increase medical complexity/roleplay? Well it increases the reliance on medics taking care of marines rather than marines just self medicating themselves. Its only 3 shots per though: except they're refillable groundside in many locations.

> Being able to use it without delay isn't incredibly broken either.

Then you should be indifferent to the change. This really needs to be a more compelling argument.

Git-Nivrak:
> Then you should be indifferent to the change. This really needs to be a more compelling argument.

I don't see the reasoning here, Maybe I wasn't clear - I meant that being able to use it without delay doesn't hurt the game so the delay is redundant.

The project direction lines you quoted are rather ambiguous and can fit in both how the game stands and how it will be after this revert.

As for the goofy part, yeah it might seem weird to see someone injecting quickly but "goofy" is not what the game is balanced around - It doesn't mean this has to be fully reverted, but it in it's current state it's just bad.

Cthulhu80:
this pr will start a civil war.

Steelpoint:
Not even a day and people are trying to revert Morrow's changes.

Superreallycoolguy:
morrow leaves and the vultures swarm! so sad!

PR #4596, opened 2023-10-07
Adds MP, OT and CL Tents

Steelpoint:
# About the pull request

This PR adds three new tents to the game, one for the MPs, OTs and a CL tent. The tents can be found in their respective departments (MP tent is in the armoury, OT tent is in their main room, the CL's tent is in their bedroom.

The MP tent includes an equipment vendor, recharger and a temporary holding cell. It is not intended as a proper prison cell for serving a sentence, but merely somewhere to hold an individual until they can be sent shipside.

The OT tent only has a chair, a sectioned off area and a crate, intended to allow an OT to store explosives in a secure area and hand them out as needed.

The CL tent is a bit more luxurious, it comes with the basic CL goods and a rug. 

If we need to see an expansion of FOB sizes to accommodate more tents, I can see to that in a separate PR. 

This is part of a discussion held about seeing a desire to include more shipside personnel in groundside operations. 

I should stress, these tent additions will require a policy change to give more broad permission for these occupations to leave and setup in the FOB, as it stands MPs and OTs are fairly restricted in going groundside. 

<details>

Screenshot 2023-10-07 16 19 01

</details>

# Changelog
:cl:
add: New tents have been added for the Military Police, the Ordnance Technician and the Corporate Liaison. The MP tent serves as a resting area and provides a holding cell, the OT tent contains a secure storage area for an OT to hand out equipment and the CL tent allows a CL to provide groundside corporate work in the field, and a rug. 
/:cl:


fira:
I'm honestly kinda against this. The original tents concept as per thwomper was meant to keep things simple and manageable.

The 3 tents have already been proven to sometimes be too much and take too much space on landing zones or be unpractical to deploy and use in practice.

Now not only this is adding 3 more tents, it's also adding 4x4 ones which are just huge. I don't really see fitting the actual game flow...

Steelpoint:
> I'm honestly kinda against this. The original tents concept as per thwomper was meant to keep things simple and manageable.
> 
> The 3 tents have already been proven to sometimes be too much and take too much space on landing zones or be unpractical to deploy and use in practice.
> 
> Now not only this is adding 3 more tents, it's also adding 4x4 ones which are just huge. I don't really see fitting the actual game flow...

This addition is a one aspect of a larger background discussion regarding the game in terms of shipside player engagement with the entire game from a broader perspective. The hope is to start moving towards encouraging most players to be involved in the core gameplay loop of the game instead of being separated and virtually uninvolved. MPs and OTs especially are harshly affected by this. 

I'm prepared to conduct a series of LZ passovers to add some more space for tents as required if this PR is accepted.

code/modules/tents/folded_tents.dm
```diff
@@ -144,6 +144,33 @@
...
+	name = "folded company Tent"
+	icon_state = "cmd"
+	desc = "A company liaison tent, provided with all the resources for company field work, and a luxurious rug. ENTRANCE TO THE SOUTH."
+	dim_x = 3
+	dim_y = 4
+	off_x = -2
+	template = /datum/map_template/tent/corp
+
+/obj/item/folded_tent/mp
+	name = "folded MP Tent"
```
Diegoflores31, line 157: i would suggest to not use "MP" in the name. call it detention tent or something

code/modules/tents/folded_tents.dm
```diff
@@ -144,6 +144,33 @@
...
+	icon_state = "big"
+	desc = "Field tent for USCM Military Police personnel, provides a rest area and a holding cell for individuals to be held until they can be sent shipside. Donuts included. ENTRANCE TO THE SOUTH."
+	dim_x = 4
+	dim_y = 4
+	off_x = -2
+	template = /datum/map_template/tent/mp
+
+/obj/item/folded_tent/ot
+	name = "folded OT Tent"
```
Diegoflores31, line 166: same for this

Diegoflores31 (commented):
I would also like to suggest that instead of giving this items for free it would be better that it was something Obtainable to avoid having comical amounts of tents in FOB .
CL tent = briefcase points
MP tent = requisitions
OT tent = requisitions
Also for OT tent i see a little redundant because REQ tent is supposed to fill the role of handing out equipment and in general OT should not really be deploying

FransFeiffer (approved):
Sprite's are awesome, they have my approval.

Steelpoint:
Under current rules the OT can't really deploy but the idea is there may be a background policy change to facilitate this, ergo this PR would support that change, though that does require the policy change to be accepted for this PR to work.

Locking the tents behind intel I feel might lead it to be being a trap choice. I suspect players will find it to be a troll pick if CIC buys a MP or OT tent versus saving for Foxtrot or a nuke.

Diegoflores31:
As i said before requisitions would be a good alternative . Maybe 5K per tent and for CL maybe 10 points 1 time purchase on his briefcase

Steelpoint:
@Diegoflores31 if that's the case, no one is going to buy the OT/MP tents. 

CL change is ok I guess.

This entire PR is predicated on the idea of a policy change to support these positions being able to deploy to the ground without any major restrictions, ergo if there are major mechanical restrictions on actually getting the tents then I feel this change quickly becomes redundant or a "never used" feature.

Tsunamico:
Heya! I'm doing my best to get past the issues with policy!
Most sincerely I understand the worry that is in the issues with the tents feeling like they could be bad choices for intel expenditures, but further changes in policy can bring these tents into intel too!
That is separate for CL. Maybe they get it upon request only, or through the briefcase, et cetera!
Changes for the LZs have also been discussed to propperly fit these tents in the areas. They're big, yes, but the feeling of setting up these tents in the FOB, darn, sounds great!
Also, for intel buying tents, it should be a single item to buy going for 2 points at tier 2 along with the OBs, and already spawn these new tents (sans the CL) along with a free big tent!

these are my two cents of thought in this!

Steelpoint:
I'll make the change that the CL can buy the tent from his briefcase, and that you can buy the MP and OT tent from req for a fair price.

Watermelon914:
> I'll make the change that the CL can buy the tent from his briefcase, and that you can buy the MP and OT tent from req for a fair price.

Don't do this, just keep them free.
There isn't really any gameplay reason these should cost money. Changing where the tents are obtained is fine, but you don't need to attach a price tag onto them.

I'd say keep it as is. The concerns about these being too big in the FOB is more so a separate problem that should get its own separate solution, like increasing available spacing on an LZ so that these tents can fit in a non-intrusive way.

It would be a good idea to push FOB construction in such a way that tents act like modules that build up an FOB rather than solely just being about building cades.

Diegoflores31:
I think this tents are not something that are "mandatory" for FOB as there may be many shifts when it will go unused . Req for a fair price i think is good so req has options to buy it . Get a replacement in case its broken . 
Making it optional will also make FOB different over the rounds players may choose which tents will be used instead of just deploying them all because "why not"

Steelpoint:
I'll keep the PR as it stands in terms of tent availability but I'll be open to changing if that is the wish of the maintainers.

Steelpoint:
I'll close this for now since there's no huge admin push to change MP and shipside roles, and I can re-open if this is desired in the future

PR #4600, opened 2023-10-08
changed flyby tgui

mullenpaul:
# About the pull request

Slight change to the TGUI for dropship flight controls. I've had feedback that it is a bit unintuitive. This change should resolve novice POs accidentally launching flyby in a panic.

Better UIs make everyone happier.
<details>

image

</details>

# Changelog
:cl:
ui: tweaked flyby controls to make them in line with other destinations
/:cl:

### DIFF
--- a/code/modules/shuttle/computers/dropship_computer.dm
+++ b/code/modules/shuttle/computers/dropship_computer.dm
@@ -317,12 +317,21 @@
 	.["primary_lz"] = SSticker.mode.active_lz?.linked_lz
 	if(shuttle.destination)
 		.["target_destination"] = shuttle.in_flyby? "Flyby" : shuttle.destination.name
-	.["destinations"] = list()
 
 	.["door_status"] = is_remote ? list() : shuttle.get_door_data()
-
 	.["flight_configuration"] = is_set_flyby ? "flyby" : "ferry"
 	.["has_flyby_skill"] = skillcheck(user, SKILL_PILOT, SKILL_PILOT_EXPERT)
+
+	.["destinations"] = list()
+	// add flight
+	.["destinations"] += list(
+		list(
+			"id" = DROPSHIP_FLYBY_ID,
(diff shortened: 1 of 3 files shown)


code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -359,23 +368,24 @@
...
-				to_chat(user, SPAN_WARNING("You don't have the skill to perform a flyby."))
-				return FALSE
 			var/is_optimised = FALSE
 			// automatically apply optimisation if user is a pilot
 			if(skillcheck(user, SKILL_PILOT, SKILL_PILOT_EXPERT))
 				is_optimised = TRUE
 			update_equipment(is_optimised)
-			if(is_set_flyby)
+			var/dockId = params["target"]
+			if(dockId == DROPSHIP_FLYBY_ID)
```
harryob, line 377:
```suggestion
			var/dock_id = params["target"]
			if(dock_id == DROPSHIP_FLYBY_ID)
```

harryob (changes requested):

fira (approved):

IsProbablyCatto:
Isn't tested. Flyby is now coded as a destination, which lasts 30 seconds instead of 200 seconds.

cuberound:
jeh none tested it and you just merged it... indeed does not work as intended

mullenpaul:
> Isn't tested. Flyby is now coded as a destination, which lasts 30 seconds instead of 200 seconds.

This was tested, it introduced a regression.

Flybys we're always coded as destinations. Behind the scenes the shuttle is forced to transit to it's existing landing zone. This work was too change how the UI tells the backend how to do this.

dadlips:
There is no longer a cancel flyby button, is this intented? Our only option is to wait out the full 200+ seconds, and with a medevac'd marine this can be bad. 
image

PR #4621, opened 2023-10-09
Medbelts pull from pillbottles as default

IsProbablyCatto:
# About the pull request

Changes the medical belts to pull pills from bottles by default, rather than pulling the entire bottle out. Tested as it just changes a single variable

Waking up and forgetting to swap your belt mode as doctor or CMO is frustrating, this fixes that.

Tested and works fine.

# Changelog

:cl:
qol: Medbelts pull from pillbottles as default
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -151,7 +151,7 @@
 	storage_slots = 14
 	max_w_class = SIZE_MEDIUM
 	max_storage_space = 28
-	var/mode = 0 //Pill picking mode
+	var/mode = 1 //Picking from pill bottle mode
 
 	can_hold = list(
 		/obj/item/device/healthanalyzer,


code/game/objects/items/storage/belt.dm
```diff
@@ -151,7 +151,7 @@
 	storage_slots = 14
 	max_w_class = SIZE_MEDIUM
 	max_storage_space = 28
-	var/mode = 0 //Pill picking mode
+	var/mode = 1 //Picking from pill bottle mode
```
Cthulhu80, line 154: You should probably be using defines, for example, there is a define here called SIZE_MEDIUM. https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#no-magic-numbers-or-strings

Cthulhu80: meant for this, max_storage_space = 28, storage_slots = 14, It's not even his code, hence it's a bit of an ask.

Cthulhu80: And aren't true and false just defines for 0 and 1? Might be wrong.

Cthulhu80: not saying you need to do this, but this is an example of how boolean defines are properly done 
Screen Shot 2023-10-09 at 10 11 02 PM
It would eliminate the need for you to place a comment because it would self-explanatory what the define is.

RavinHobo:
GIVE THIS MAN A MEDAL

Zonespace27 (approved):
In the future the ability to change modes needs to be more clear

HeresKozmos:
As someone with a gazillion medical hours thank you for this 🙏

PR #4623, opened 2023-10-10
Reimplements halloween pumpkin helmets event

fira:
# About the pull request

This reimplements the Halloween Pumpkins spawning on map and being carvable as helmets which was apparetnly a thing a number of years ago

Variety

# Changelog
:cl: Firartix and Frans_Feiffer
add: Readded Halloween pumpkin helmets.
/:cl:


code/modules/holidays/halloween/decorators.dm
```diff
@@ -0,0 +1,98 @@
...
+/datum/game_decorator/halloween/cobwebs/proc/decorate_turf(turf/closed/wall/almayer/T)
+	var/static/list/order = list(NORTHWEST, SOUTHEAST, NORTHEAST, SOUTHWEST) // Ordering of wall_connections
+	if(length(T.wall_connections) < 4)
+		return
+
+	var/event_progress = get_event_progress()
+	var/placement_chance = base_chance - (event_progress * ramp_chance)
+	for(var/i = 1 to 4)
+		var/diag = order[i]
+		if(T.wall_connections[i] != "5") // CORNER_CLOCKWISE | CORNER_COUNTERCLOCKWISE as string - don't ask me
```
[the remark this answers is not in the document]
fira: Well that was just refactoring but yeah, fixed

code/modules/holidays/halloween/decorators.dm
```diff
@@ -0,0 +1,98 @@
...
+/datum/game_decorator/halloween/pumpkins/decorate()
+	var/list/turf/valid_turfs = list()
+	var/list/ground_levels = SSmapping.levels_by_trait(ZTRAIT_GROUND)
+	for(var/ground_z in ground_levels)
+		var/list/turf/all_turfs = block(locate(1, 1, ground_z), locate(world.maxx, world.maxy, ground_z))
+		for(var/turf/open/turf in all_turfs)
+			if(turf.is_groundmap_turf)
+				valid_turfs += turf
+
+	var/list/turf/picked_turfs = list()
```
Drulikar, line 33: There are no valid_turfs counted for at least LV624 as the groundmap when this fires on localhost.

code/modules/holidays/halloween/decorators.dm
```diff
@@ -0,0 +1,98 @@
...
+
+/datum/game_decorator/halloween/is_active_decor()
+	return (get_event_progress() != -1)
+
+/datum/game_decorator/halloween/proc/get_event_progress()
+	. = -1
+	var/cur_day = text2num(time2text(world.timeofday, "DD"))
+	var/cur_mon = text2num(time2text(world.timeofday, "MM"))
+	if(cur_mon == 10)
+		return min(0, 31 - cur_day)
```
Drulikar: max is probably what you intended? But even then if this returned a scalar 0-1 on how haloweeny it is it probably would make more sense than -1, 0-31.

So I think you could `return 1 - (31-cur_day)/31`? to get a value 0-1 where the first is 0.032258064516129 and the 31st is 1?

code/modules/holidays/halloween/pumpkins/wearable.dm
```diff
@@ -0,0 +1,55 @@
...
+		to_chat(user, SPAN_WARNING("You can't put on a full pumpkin! Empty and carve it with a sharp object first."))
+		return FALSE
+	. = ..()
+
+/obj/item/clothing/head/pumpkin/attackby(obj/item/tool, mob/user)
+	if(!carved_icon && (tool.sharp == IS_SHARP_ITEM_ACCURATE || tool.sharp == IS_SHARP_ITEM_BIG))
+		var/choice = tgui_input_list(user, "Select the pattern to carve on your pumpkin!", "Pumpkin Carving", carvable_icons)
+		if(choice)
+			carved_icon = choice
+			name = "carved pumpkin"
```
Drulikar, line 41: It is kind of weird that the carved sprite is the same when on the ground. Only when worn on the head is the carving any different. I'd suggest merely having another set of overlays that are shifted downwards some, but they'd also need to be enlarged so I dunno.

code/modules/holidays/halloween/pumpkins/wearable.dm
```diff
@@ -0,0 +1,55 @@
...
+/obj/item/clothing/head/pumpkin/get_mob_overlay(mob/user_mob, slot)
+	var/image/pumpkin = ..()
+	if(carved_icon && slot == WEAR_HEAD)
+		var/image/overlay = overlay_image(icon, "[prefix]pumpkin_[carved_icon]")
+		pumpkin.overlays += overlay
+	return pumpkin
+
+/obj/item/clothing/head/pumpkin/corrupted
+	prefix = "cor_"
+	carvable_icons = list("cry", "sob", "sad", "why", "spooky", "ff")
```
Drulikar, line 55: The patch has a unique name/description for corrupted, but the corrupted pumpkins themselves do not.

code/modules/holidays/halloween/pumpkins/patches.dm
```diff
@@ -0,0 +1,58 @@
...
+	update_icon()
+
+/obj/structure/pumpkin_patch/update_icon()
+	overlays?.Cut()
+	. = ..()
+	switch(pumpkin_count)
+		if(3) icon_state = "[icon_prefix]pumpkins_full"
+		if(2) icon_state = "[icon_prefix]pumpkins_half"
+		if(1) icon_state = "[icon_prefix]pumpkin"
+		else  icon_state = ""
```
Drulikar, line 27: No pumpkins left leaves it as a placeholder examine image: 
image

Drulikar (changes requested):

code/modules/holidays/halloween/decorators.dm
fira: i forgot to put /turf/open/gm (groundmap) as groundmap turf because i needed to check rivers etc, gonna add it

code/modules/holidays/halloween/pumpkins/wearable.dm
fira: Yeah, I thought about it, but i'm not really confident in my spriting abilities so i just rolled with the previous sprites Frans provided me...

code/modules/holidays/halloween/decorators.dm
fira: should be fixed. also reduced amount of pumpkins spawned

code/modules/holidays/halloween/pumpkins/patches.dm
fira: added a proper placeholder blank sprite

Drulikar (approved):
Still needs sprite approval (and maybe they can whip up the different carved states when not on a head), but seems to be working and nothing is obviously sticking out to me (other than a lack of new lines in game_decorator).

Only thing I am slightly confused about is now the difference between a decorator and now a game_decorator. Is the xmas decorator still a decorator because it does need to keep re-firing? Whereas this is is just a single-shot change so its a game_decorator?

fira:
The decorators apply to atom types, stored through a typecache for fast lookup during instanciation

This doesn't really cut it efficiently/at all for things that must consider the entire state of the game at one point, eg. the pumpkins which scan and select spots on the map in one go

Hence addition of game_decorators. It's not ideal but i'm not sure how else to bring this together

code/game/supplyshuttle.dm
Drulikar, line 1: Did you mean to commit changes to this file?

fira: Nope

code/controllers/subsystem/game_decorator.dm
```diff
@@ -0,0 +1,31 @@
...
+	if(defer_decoration)
+		RegisterSignal(SSdcs, COMSIG_GLOB_MODE_POSTSETUP, PROC_REF(defered_decoration))
+/datum/game_decorator/proc/is_active_decor()
+	return FALSE
+/datum/game_decorator/proc/defered_decoration(dcs)
+	SIGNAL_HANDLER
+	decorate()
+/datum/game_decorator/proc/decorate()
+	set waitfor = FALSE
+	return
```
harryob, line 31:
```suggestion
/datum/game_decorator
	var/defer_decoration = TRUE //! So map decoration is done post-setup after nightmare and spawners

/datum/game_decorator/New()
	if(defer_decoration)
		RegisterSignal(SSdcs, COMSIG_GLOB_MODE_POSTSETUP, PROC_REF(defered_decoration))

/datum/game_decorator/proc/is_active_decor()
	return FALSE

/datum/game_decorator/proc/defered_decoration(dcs)
	SIGNAL_HANDLER
	decorate()

/datum/game_decorator/proc/decorate()
	set waitfor = FALSE
	return
```

harryob (changes requested):

PR #4652, opened 2023-10-12
reduces timer on joining ert after death to 30 seconds

CapCamIII:
# About the pull request

reduces timer

Having to wait a full minute to join an ERT is annoying, it was better b4 when timer from ERT was a minute as well, but 30 second ERT means if u die just b4 ERT goes you cant join regardless.

if ppl are ghosting bc they want ert then they are stupid.

<details>

</details>

# Changelog
:cl:
balance: Timer on attempting to join ERT after death is now 30 seconds down from 1 minute
/:cl:

### DIFF
--- a/code/datums/emergency_calls/emergency_call.dm
+++ b/code/datums/emergency_calls/emergency_call.dm
@@ -168,7 +168,7 @@
 		return
 	var/deathtime = world.time - usr.timeofdeath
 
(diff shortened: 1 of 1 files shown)


harryob (approved):
we should probably make a snitch log at some point for jannies if someone ghosts when there's nothing wrong with them

Zonespace27 (changes requested):
image

Must be at least 30 seconds.

Issue #4609, opened 2023-10-08
Xenos can see raw HTML tags on papers


DoomLaser9:
### Testmerges

N/A

### Description of the bug

Viewing a paper with text as a xeno results in a lot of garbled text, but which contains all of the raw HTML from the paper. The text is also no longer colored and formatted properly (a xeno should still be able to see the overall color/format of the paper).

Sometimes the paper is also just completely blank when viewed.

### What's the difference with what should have happened?

When viewing a paper as a xeno. The text positioning/formatting/colors should be the same as when a human sees it, and there shouldn't be HTML tags in the output.

### How do we reproduce this bug?

1. Have human write `123` on paper.
2. Look at paper as human, see red `123`.
3. Be xeno.
4. Look at paper.
5. See that it has a lot more text and nothing is red.
6. Also, deciphering the paper (viewing multiple times) results in: `<font face="Verdana" color=red>123</font>`

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Cthulhu80:
I feel like Xenos shouldn't even be able to examine the paper.

Cthulhu80:
blank page came from the output of the stars proc, since it was returning formatted html with random asterisks everywhere. The browser couldn't understand or display it most of the time, depending on what was modified to be a asterisks.

PR #4654, opened 2023-10-12
fixes xenos seeing raw html on paper

Cthulhu80:
# About the pull request
currently work in progress, fixes the bug as outlined here #4609

bug bad

https://github.com/cmss13-devs/cmss13/assets/122310258/e768a5d6-be35-4be7-b845-3850e7ef7cc6

# Changelog
:cl: 
fix: xenos viewing paper no longer displays unformatted raw HTML
/:cl:


code/modules/mob/mob_helpers.dm
```diff
@@ -131,6 +131,41 @@ var/global/list/limb_types_by_name = list(
 		index++
 	return output_message
 
+// proc that parses an html input string and scrambles the non-html string contents
+/proc/stars_decode_html(message)
+	if(length(message) == 0)
```
harryob, line 136:
```suggestion
	if(!length(message))
```

code/modules/mob/mob_helpers.dm
```diff
@@ -131,6 +131,41 @@ var/global/list/limb_types_by_name = list(
...
+	var/parsing_message = FALSE
+
+	// the current string value that needs to be scrambled.
+	var/current_string_to_scramble = ""
+
+	for(var/character_index in 1 to length(message))
+		if(message[character_index] == ">")
+			parsing_message = TRUE
+			continue
+		else if(message[character_index] == "<")
```
harryob, line 153:
```suggestion
		if(message[character_index] == "<")
```
don't need an else if the previous condition returns

code/modules/mob/mob_helpers.dm
```diff
@@ -131,6 +131,41 @@ var/global/list/limb_types_by_name = list(
...
+			parsing_message = FALSE
+			if(length(current_string_to_scramble) == 0)
```
harryob, line 155:
```suggestion
			if(!length(current_string_to_scramble))
```

harryob (changes requested):
nit picks

code/modules/mob/mob_helpers.dm
```diff
@@ -131,6 +131,41 @@ var/global/list/limb_types_by_name = list(
...
+
+	// the current string value that needs to be scrambled.
+	var/current_string_to_scramble = ""
+
+	for(var/character_index in 1 to length(message))
+		if(message[character_index] == ">")
+			parsing_message = TRUE
+			continue
+		if(message[character_index] == "<")
+			parsing_message = FALSE
```
DoomLaser9, line 154: You also need to check for quotation marks inside of HTML tags. Consider a tag such as `<tag property=">Injection()" />`. So when you are inside an HTML tag and run into a quotation mark (either single `'` or double `"`), then you advance until you find the matching quotation mark. Perhaps you could add a couple booleans for this, e.g. `need_single_quote` and `need_double_quote`.

Maybe it would be a bit overkill but it's probably good to be safe.

code/modules/mob/mob_helpers.dm
```diff
@@ -131,6 +131,41 @@ var/global/list/limb_types_by_name = list(
...
+				continue
+			var/scrambled_string = stars(current_string_to_scramble)
+			replaced_scrambled_strings[current_string_to_scramble] += scrambled_string
+			current_string_to_scramble = null
+			continue
+		if(parsing_message)
+			current_string_to_scramble += message[character_index]
+
+	for(var/unscrambled_message_key in replaced_scrambled_strings)
+		message = replacetext(message, unscrambled_message_key, replaced_scrambled_strings[unscrambled_message_key])
```
DoomLaser9, line 165: Carefully-crafted user input would make this replace text inside of HTML tags. This should be reworked to be a bit more similar to how `/proc/stars` works, where the final message is being appended to in the original `for` loop. In this case, you could be appending `scrambled_string` from the loop before to a variable representing the final output. I suspect that would also be better performance-wise compared to repeatedly searching through the whole text for replacements.

Your current code also means that duplicates of `unscrambled_message_key` would get the same star text, while doing it iteratively would preserve the original behavior of not duplicating star text instances.

DoomLaser9 (commented):
Two things: HTML attribute handling and replacetext issues

Cthulhu80:
going to be busy, i'm closing this pr for now. I'll get to it later.

Cthulhu80:
branch being real funky, closing and making a new branch.

code/modules/mob/mob_helpers.dm
Cthulhu80: I don't think the duplication matters too much, or the performance since this is very cold code. If a maintainer disagrees then I'll refactor it.

code/modules/mob/mob_helpers.dm
Cthulhu80: Alright.

PR #4668, opened 2023-10-13
Fixes disconnected ahelpers runtime

fira:
# About the pull request

If client disconnected, due to hard del, instances get replaced by null. This still goes through the timer CALLBACK somehow (?), causing invocation of the proc with src nulled out. 

Every time this happens are two different runtimes, one for add_verb and one for deltimer.

Untested

# Changelog

Internal change


harryob (approved):
i hate this

PR #4670, opened 2023-10-13
Gives ERTs a variant of the regular medkit that isn't locked to skills in packs.

CapCamIII:
# About the pull request

A bunch of ERTs are designed with having their meds in medkits, which was obviously not noticed or cared about when making !4443 so I made a variant of the regular kit that has no lock, only ERT roles(and 2 HG roles on WO) have it so its not a real balance concern.

Storage nesting is not an issue with normal preset loadouts like ERTs(or at all :D) as they are designed around and only have access to specific loadouts.

<details>

</details>

# Changelog
:cl:
balance: ERTs now have a variant of the standard medkit that has no storage lock.
fix: Lots of ERTs now correctly have a medkit spawning in their gear
/:cl:


code/game/objects/items/storage/firstaid.dm
```diff
@@ -98,6 +98,10 @@
 /obj/item/storage/firstaid/regular/empty/fill_preset_inventory()
 	return
 
+/obj/item/storage/firstaid/regular/response
+	required_skill_for_nest_opening = SKILL_MEDICAL
+	required_skill_level_for_nest_opening = SKILL_MEDICAL_DEFAULT
```
Zonespace27, line 103: Change the medkit desc to not have the skill bit

Zonespace27 (changes requested):

code/game/objects/items/storage/firstaid.dm
CapCamIII: Done @Zonespace27 :D

Issue #4606, opened 2023-10-08
Molotov Safe Area inhibitor


DoomLaser9:
### Testmerges

N/A

### Description of the bug

`The improvised firebomb's safe-area accident inhibitor prevents you from priming the grenade!`

Molotovs shouldn't have any inhibitors.

### What's the difference with what should have happened?

It doesn't make any sense for a glass bottle with a paper sticking out of it to somehow have advanced electronics to detect that it's on the ship and prevent it from priming. Molotovs should always work.

It's not easy to get molotovs on the ship anyway, so there shouldn't be any worries about someone barbecuing briefing at the beginning of the round.

### How do we reproduce this bug?

1. Acquire molotov.
2. Be on ship.
3. Attempt to use molotov.

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Wintermote:
We honestly never even had serious problems with people griefing with grenades on the almayer anyway, and the odd time someone actually set a grenade off it's almost always been an accident.

I'm not even sure why the IFF was implemented in the first place besides thematics, but considering most newer players won't even know how to get or make molotovs, I don't see why it should have IFF.

Accidents are fun anyway

Huffie56:
not a bug as forest say. 
but if you have a better more RP way to do safe area inhibitor for molotov cocktail your welcome to post it on the forum.

PR #4697, opened 2023-10-16
molotov

Diegoflores31:
# About the pull request
Molotovs no longer have IFF antigrief inhibitors.
there is no room for goverment microchips on a glass bottle full of welding fuel.

bug bad ?
#4606

<details>

</details>

# Changelog
:cl:
fix: Molotovs no longer have IFF inhibitors

/:cl:


fira:
I don't think this is a bug, the molotovs have no reason to get special treatment. The protection is a purely OOC one - it could be written anything else and has nothing to do with electronics. If you think the restriction is dumb, the game already has a setting to switch antigrief to only new players, so take it up to management to change it.

realforest2001:
Not a bug. As fira said if it's that much of an issue you can change the display to read as some other reason you can't firebomb the marines.
Antigrief config is not presently on the table to change.

PR #4700, opened 2023-10-17
Bandaid the null clients in GLOB.clients

Drulikar:
# About the pull request

Apparently we are somehow getting nulls into GLOB.clients now, and I don't know why. Stat panel is the most egregious spammer about this fact, but a rather slow subsystem will also periodically check too.

We don't need tens of thousands of runtimes.

<details>

</details>

# Changelog
:cl: Drathek
fix: Bandaided a problem with nulls in GLOB.clients
/:cl:


code/controllers/subsystem/inactivity.dm
```diff
@@ -1,20 +1,25 @@
-#define INACTIVITY_KICK 6000 //10 minutes in ticks (approx.)
+#define INACTIVITY_KICK 10 MINUTES
 
 SUBSYSTEM_DEF(inactivity)
 	name = "Inactivity"
-	wait = INACTIVITY_KICK
+	wait = INACTIVITY_KICK * 0.5
```
fira, line 5: why * 0.5 ?

code/controllers/subsystem/inactivity.dm
```diff
@@ -1,20 +1,25 @@
...
-		for(var/i in GLOB.clients)
-			var/client/C = i
-			if(C.admin_holder && C.admin_holder.rights & R_ADMIN) //Skip admins.
-				continue
-			if (C.is_afk(INACTIVITY_KICK))
-				if (!istype(C.mob, /mob/dead))
-					log_access("AFK: [key_name(C)]")
-					to_chat(C, SPAN_WARNING("You have been inactive for more than 10 minutes and have been disconnected."))
-					qdel(C)
+	// Maybe we should just get a SS to bandaid all important global lists like this?
```
fira, line 11: that's what /tg/ does, they have a "maintenance" subsystem for that. it might be better to port it, or do one of our own, then we can run it much faster. null in client lists will break down various things, not just cause runtimes

code/controllers/subsystem/inactivity.dm
Drulikar: To clear nulls atleast every 5 minutes. Tg clears them every 6 ticks but even with tens of thousands of runtimes flowing in from the stat panel encountering a mil client, the game didn't grind to a halt.

code/controllers/subsystem/inactivity.dm
Drulikar: Ok - are there any other lists you know are problematic it should also clean up even if at a lower frequency?

code/controllers/subsystem/inactivity.dm
Drulikar: Also worth noting, a delay of 10 minutes firing means that you are giving players 10-20 minutes to be inactive. A fire time of 5 minutes means they can be 10-15 minutes inactive.

code/controllers/subsystem/inactivity.dm
Drulikar: Since GLOB.clients is all we are concerned about for now, I will consider null clients to be inactive and they will be kicked from the list here ;)

harryob (approved):

PR #4720, opened 2023-10-19
Fix a source of bad icon operations

Drulikar:
# About the pull request

This PR updates the stat panel examine of human mobs to use a dummy copy of the examined mob to fix runtimes related to a bad icon operation from a workaround in icon2html for humans. It is not a perfect copy to limit the upper end of cost for rendering humans, but should be fairly accurate visually and forced to face south. Initial testing seems to report 0 to 0.2 seconds per mob.

This PR doesn't eliminate the possibility of an icon2html of a human, but I am unsure if its wise to just remove that workaround. Even if it was just removed for this situation, the resulting icon would be empty for the human mob.

Fixes an issue that can result in tens of thousands of runtimes.

<details>

image
image

</details>

# Changelog
:cl: Drathek
fix: Fixed a cause of bad icon operations
/:cl:

### DIFF
--- a/code/__HELPERS/icons.dm
+++ b/code/__HELPERS/icons.dm
@@ -774,14 +774,14 @@ world
 
 	/* Debugguing due to gamebreaking performance in the Blend calls made by getFlatIcon (think 200ms+ per icon) and overtimes */
 	if(istype(thing, /atom))
-		var/atom/dir = thing
-		log_debug("costly_icon2html called on ref=[REF(dir)], instance=[dir], type=[dir.type], with [length(dir.overlays)] overlays, finished in [(REALTIMEOFDAY-start_time) / 10]s")
+		var/atom/D = thing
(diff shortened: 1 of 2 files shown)


fira:
Found another source of human icon2html (which also crashes) at `/datum/reagents/proc/handle_reactions`
What do we do about it?

fira (approved):

Drulikar:
> Found another source of human icon2html (which also crashes) at `/datum/reagents/proc/handle_reactions`
> What do we do about it?

If I can reproduce it maybe I can find some other solution, but it might be best we just remove the workaround mrstonedone put in. It'll just return a blank icon since humans require overlays to also be rendered to have any limbs etc - so they must use getflaticon or something similar that's going through all their overlays to actually appear as anything.

harryob (approved):

PR #4737, opened 2023-10-20
adds a round id box to the issue template

harryob:
so we can look at logs  a little easier

no playerfacing changes


fira:
✅  good point

PR #4745, opened 2023-10-20
fix reqs vendors sprites

fira:
# About the pull request

Snowflake Reqs sprites for vendors are instanciated in map rather than using the dedicated subtypes. This causes the sprite to revert to normal vendor rack when you use it. Bleh.

Witness: 
image

image

The attachie vendor just wasn't messed up. I'm pretty sure this already got fixed before and was just reverted by sloppy map merging.

CM Dev gave me these cool sprites, I'm going to use all the cool sprites!!!

See above.

# Changelog
:cl:
fix: Fixed incorrect Reqs vendors visuals on the Almayer. They now blend in with the walls again.
/:cl:

### DIFF
--- a/code/game/machinery/vending/vendor_types/requisitions.dm
+++ b/code/game/machinery/vending/vendor_types/requisitions.dm
@@ -192,7 +192,7 @@
 				turf_to_vent_to = H.loc
 	return turf_to_vent_to
 
-/obj/structure/machinery/cm_vending/sorted/cargo_guns/blend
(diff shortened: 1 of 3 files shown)


Drulikar (approved):
Map diff bot doesn't even see a change, but regardless better to remove mapped in icon state overrides.

fira:
Inheritance is wrong and this removes the snowflake "vend to counter" feature of the vendors

fira:
Should be fixed

harryob (approved):

PR #4754, opened 2023-10-21
turns options button into a "tab"

harryob:
i think it just looks nicer aesthetically. if someone disagrees particularly thats fine

:cl:
ui: the options button is now part of the statbrowser tabs proper
/:cl:


fira (approved):

Drulikar:
This works, though TBH it would be better here or something: 
image
But that will involve passing that data to the stat panel.

Drulikar:
My statement above are still my thoughts, but this code as is is functional.

PR #4756, opened 2023-10-21
Minor Code Maintenance Bundle (deletes, standards, runtimes)

fira:
# About the pull request

Just some bundled code maintenance bits:
 * annotations and variable names 
 * loc -> forceMove / New -> Initialize, GLOB variables, per standard
 * add some missing ref clearing in Destroy
 * dead code
 * fix minor runtimes 
 * runtime fix in gear presets spamming logs
 * fixes a typo in bottles making them use only 1 of the 4 sprites

Standards, bugfixing

# Changelog
:cl:
fix: Spawned bottles now use all 4 of their sprites as intended rather than always the same.
/:cl:


code/modules/mob/dead/observer/observer.dm
```diff
@@ -477,6 +478,9 @@ Works together with spawning an observer, noted above.
 		ghost.mind.player_entity.update_panel_data(round_statistics)
 		ghost.mind.original = src
 
+	// If we're ghosting as part of previous mob deletion, do not keep refs
+	if(QDELETED(src) && mind?.original == src)
+		mind.original = null
```
Drulikar, line 483: I'm inclined to think this would be better to do in just Destroy rather than if we're calling ghostize.

code/modules/gear_presets/uscm.dm
```diff
@@ -780,8 +780,6 @@
 	new_human.equip_to_slot_or_del(new /obj/item/storage/pouch/general/large(new_human), WEAR_R_STORE)
 	new_human.equip_to_slot_or_del(new /obj/item/device/binoculars(new_human), WEAR_L_HAND)
 
-	new_human.back.pickup()
```
Drulikar, line 783: Both radio backpack presets did this some reason but yeah it would throw a runtime - is there anything broken with the radio backpack if you ckey into these mobs now that they don't perform pickup?

code/modules/mob/dead/observer/observer.dm
fira: It would be better ideally, but this is on purpose: there is actually handling for it in `/mob/Destroy` but by design it can only take effect if ghostize failed. If ghostize succeeds we are left with two choices: this, or nulling it BEFORE ghostize and hoping it doesn't break anything.

code/modules/gear_presets/uscm.dm
fira: I honestly didn't even bother checking what the breakable might be because the thing runtimes anyway and it's been like that for 5 months -- but it's true it might still trigger ID Changes in `/obj/item/storage/backpack/marine/satchel/rto/pickup`

code/modules/mob/dead/observer/observer.dm
fira: I guess I could refactor ghostize to return success and act upon it instead... Thoughts?

Drulikar: Up to you, I just figure assignment to null can just be done always in Destroy (namely when it goes through mob/Destroy) rather than conditionally here.

Drulikar: This is existing code:
image

So to me it seems redundant to be nulling mind here when in a few lines it will be nulled?

code/modules/gear_presets/uscm.dm
Drulikar: https://github.com/cmss13-devs/cmss13/pull/727 is the origin of that pickup call.

We probably need that pickup function to fire ever a mind enters this mob (e.g. Login?) and or you can try passing new_human as the argument for pickup to see if that then properly registers the signals.

code/modules/mob/dead/observer/observer.dm
fira: Well that's what i was explaining above, I don't understand what you mean

If ghostize didn't fail then the non-ghost mob is still stored as mind.original, hence it needing to be cleared

If ghostize fails, both mind.current and mind.original are still set and need to be cleared in that snippet

ghostize cannot know at present whether it should or not clear mind.original without adding eg. a variable passed to it, which i figured was essentially simpler to just use QDELETED

we cannot clear mind in Destroy() because it is still valid and needed for the purpose of ghostize

Drulikar: There is nothing done about a "ghostize" failure, but this is also extremely minor.

My statement is simply that it is redundant to: Enter Mob Destroy -> Enter Ghostize -> Attempt to clear mind -> Return to Mob Destroy -> Attempt to clear mind

What I am saying is Ghostize clearing mind ONLY for a qdel is redundant because we know we should already be in a Destroy call that will clear it. So whats the point of even checking if we're deleted if the Deletion proc will know it is deleted and can just clear it once and for all?

fira:
per discussion on discord will do a more in depth fix for mind.original with weakrefs in a separate PR

code/modules/gear_presets/uscm.dm
fira: should be fixed, just made phone setting happen on equipped() aswell

Drulikar (approved):
Code changes appear fine but this really ought to be tested, I don't know what all scenarios I would need to test to ensure this isn't a regression.

cm13-github:

cm13-github:

PR #4757, opened 2023-10-21
Synthetic pain sources change

QuickLode:
# About the pull request

Changes up text for pain and should removes slow from zombie 

"But why do Synths feel pain when internals damaged" I have that 'fixed' in another PR where synths will get powercell failure from getting internals hurt
"But why do Synths feel pain when nearing crit" this is supposed to simulate a catastrophic chassis shutdown
"But why does the beserker rav still pain the synth?" well its a big ass scythe unlike the little rat claws or tail so I thought this is fine for now.
Synthetics don't feel pain, and would not be crippled in the same way as a human would be.

<details>
Wasn't able to fully test this because I only have 1 client!
</details>

# Changelog
:cl:
fix: Synthetic's no longer feel pain from zombie claws
spellcheck: clarifies Synthetic pain during a crippling slash
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/Lurker.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Lurker.dm
@@ -80,8 +80,9 @@
 	if (next_slash_buffed)
 		to_chat(bound_xeno, SPAN_XENOHIGHDANGER("You significantly strengthen your attack, slowing [target_carbon]!"))
 		if(IS_SYNTHETIC)
-			to_chat(target_carbon, SPAN_XENOHIGHDANGER("[bound_xeno] targets a critical joint, damaging you!"))
+			to_chat(target_carbon, SPAN_XENOHIGHDANGER("[bound_xeno] targets a critical joint, temporarily causing a malfunction!"))
 			original_damage *= buffed_slash_damage_ratio
+			target_carbon.set_effect(get_xeno_stun_duration(target_carbon, 3), SUPERSLOW)
 			next_slash_buffed = FALSE
 		else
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/Lurker.dm
```diff
@@ -79,13 +79,19 @@
...
 	if (next_slash_buffed)
 		to_chat(bound_xeno, SPAN_XENOHIGHDANGER("You significantly strengthen your attack, slowing [target_carbon]!"))
-		to_chat(target_carbon, SPAN_XENOHIGHDANGER("You feel a sharp pain as [bound_xeno] slashes you, slowing you down!"))
-		original_damage *= buffed_slash_damage_ratio
-		target_carbon.set_effect(get_xeno_stun_duration(target_carbon, 3), SUPERSLOW)
-		next_slash_buffed = FALSE
-		var/datum/action/xeno_action/onclick/lurker_assassinate/ability = get_xeno_action_by_type(bound_xeno, /datum/action/xeno_action/onclick/lurker_assassinate)
-		if (ability && istype(ability))
-			ability.button.icon_state = "template"
+		if(IS_SYNTHETIC)
```
harryob, line 82: aren’t you checking for a constant value? this is just a bitfield

harryob (changes requested):

morrowwolf:
Just FWIW (and I'm not a maintainer anymore so whatever) a lot of this synth specific stuff should probably just be passed to species to handle so we can OOP a bit and not have all of these if(synth) changes all over iunno

PR #4762, opened 2023-10-21
Add Item & Footprints offsets

fira:
# About the pull request

This:
* Adds transverse offsets to blood footprints
* Adds notable pixel offsets to a few items
* Adds a very slight pixel offset to all items
* Enables rotation for thrown flashlights
* Cause objects exiting a surface (rack/table) to regenerate offset instead of being stuck at center
* Stops random offsets from overwriting mapped offsets

The goal is to have map visuals more organic when we have a lot of objects on the ground - targeting in particular items you find readily in dense areas such as Reqs or a FOB.

Consider this for example, the blood footprints are all aligned, in more extreme situations (eg WO) it makes an actual "grid" which i personally find very immersion breaking
image

Adding a slight offset helps counter that:
image

# Changelog
:cl:
add: Bloody footprints are now slightly offset to break long visual straight lines.
fix: Items do not align back to the center of turfs anymore when picked from a surface (table or rack)
add: Some more items now have offsets on the map display, and they all can be slightly offset.
/:cl:

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -156,6 +156,11 @@
 
 	var/list/inherent_traits
 
+	/// How much to offset the item alongside X visually
+	var/ground_offset_x = 2
+	/// How much to offset the item alongside X visually
+	var/ground_offset_y = 2
+
 /obj/item/Initialize(mapload, ...)
 	. = ..()
 
@@ -174,6 +179,8 @@
 
 	if(flags_item & MOB_LOCK_ON_EQUIP)
 		AddComponent(/datum/component/id_lock)
+
+	scatter_item()
 
 /obj/item/Destroy()
 	flags_item &= ~DELONDROP //to avoid infinite loop of unequip, delete, unequip, delete.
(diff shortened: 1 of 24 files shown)


code/game/objects/items.dm
```diff
@@ -156,6 +156,11 @@
 
 	var/list/inherent_traits
 
+	/// How much to offset the item alongside X visually
+	var/ground_offset_x = 2
+	/// How much to offset the item alongside X visually
```
fira, line 161:
```suggestion
	/// How much to offset the item alongside Y visually
```

code/game/objects/items.dm
```diff
@@ -156,6 +156,11 @@
 
 	var/list/inherent_traits
 
+	/// How much to offset the item randomly etiher way alongside X visually
+	var/ground_offset_x = 0
+	/// How much to offset the item randomly etiher way alongside Y visually
```
harryob, line 161:
```suggestion
	/// How much to offset the item randomly either way alongside X visually
	var/ground_offset_x = 0
	/// How much to offset the item randomly either way alongside Y visually
```

harryob (changes requested):
1 nit

harryob (approved):

PR #4778, opened 2023-10-24
Removes thermal goggles from spawning in game

Steelpoint:
# About the pull request

Removes a pair of very overpowered items from two maps (Soro and Kut) as well as a RNG spawner. 

# Explain why it's good for the game

Items usually reserved for whiteout or other high-strength entities should not be available in normal gameplay. 

Vision and sightlines are a highly contentious area of balance, the game is carefully crafted around the idea that humans have limited visibility and they typically have to expend some effort or resource to increase that visiblity. Certain Marines have access to night vision, but this is either a rare kit (spec) or at greater cost (NVG optic that has a high cost + power drain).

Free thermal visors that not only provide night vision, but thermal vision, for free and forever, are highly overpowered in the hands of a marine. Normally reserved for Xenos, Predators or rare elite ERTs or whiteout operators, this is not a item I believe should be accessible to an average marine or survivor. 

(Don't forget this could be combined with binocs for extra long vision)

Furthermore this is a case of "the rich get richer", its more likely a veteran player will know of these items locations, ergo more experienced players will find it, and thus gain an ever stronger advantage of other players. 

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
maptweak: Removed in game access to thermal goggles from Kutjevo and Sorokyne, as well as removing a RNG spawner that can have them be spawned in. 
/:cl:

PR #4780, opened 2023-10-25
Thermal Alternate PR

c4xmaniac2:
# About the pull request

This PR is an alternate to https://github.com/cmss13-devs/cmss13/pull/4778 which strives to maintain the kit in the game. These items have been in the game for half a decade, and are/have been a non-issue for this entire time. What happens is someone (usually a surv) gets them, then they die and no one takes them because no one thinks they matter. This PR makes it harder to gain access to these items by placing them further afield, and often time placing them inside of a safe further increasing the opportunity cost and time sink required to achieve these items.

I do not think there is a warranted reason to remove thermals from normal play, they've been in the game and actively used for half a decade and I've hardly seen any complaints about them in all of that time. The spawns that were easily accessed have been moved further away, often put into safes so the risk is extremely high. I think this type of gear balancing (time and opportunity cost combined with risk vs reward) should be considered as the norm for dealing with survivor gear rather than simply removing or randomizing gear.

<details>

image
Science annex has had all of its random goggle spawners removed, and the old thermal spawn in chapel has been returned. With the removal of the west bridge in yard, its still in a high risk area to go for.

image
One of the thermal goggles has been moved into the NE of soro and put into a safe. Getting access to this item will be a major risk.

image
One of kutjevo's goggles have been moved into a safe and to the rear of construction. High risk area because its almost always weeded within 3min of the game starting, and you'll need to guess the safe's combo to get it.

image
The other kutjevo goggles have been moved to where the golden HG spawn is. Super far into the map, you'll need to take a major risk to get it.

</details>

# Changelog

:cl:
balance: Rebalanced the thermal placement on Soro, Science Annex, and Kutjevo
/:cl:

### DIFF
--- a/maps/map_files/FOP_v3_Sciannex/Fiorina_SciAnnex.dmm
+++ b/maps/map_files/FOP_v3_Sciannex/Fiorina_SciAnnex.dmm
@@ -4537,11 +4537,6 @@
 	icon_state = "bluefull"
 	},
 /area/fiorina/station/power_ring)
-"cFX" = (
-/obj/structure/largecrate/supply/supplies,
-/obj/effect/spawner/random/goggles/lowchance,
-/turf/open/floor/plating/prison,
-/area/fiorina/station/power_ring)
 "cGg" = (
 /obj/structure/machinery/gibber,
 /turf/open/floor/prison{
@@ -23497,13 +23492,6 @@
 	icon_state = "floor_plate"
 	},
 /area/fiorina/station/power_ring)
-"ohs" = (
-/obj/structure/surface/table/reinforced/prison,
-/obj/effect/spawner/random/goggles/lowchance,
-/turf/open/floor/prison{
-	icon_state = "redfull"
-	},
-/area/fiorina/station/security)
 "ohF" = (
 /obj/structure/platform/kutjevo/smooth,
 /turf/closed/wall/mineral/bone_resin,
@@ -28706,12 +28694,6 @@
 	icon_state = "floor_plate"
 	},
 /area/fiorina/station/lowsec)
-"rru" = (
-/obj/effect/spawner/random/goggles/midchance,
-/turf/open/organic/grass{
-	name = "astroturf"
-	},
-/area/fiorina/station/research_cells)
 "rrP" = (
 /obj/effect/spawner/random/tool,
 /turf/open/floor/prison{
@@ -30529,16 +30511,6 @@
 	icon_state = "floor_plate"
 	},
 /area/fiorina/tumor/aux_engi)
-"sAx" = (
-/obj/structure/bed{
-	icon_state = "abed"
-	},
-/obj/effect/spawner/random/goggles/lowchance,
-/turf/open/floor/prison{
-	dir = 6;
-	icon_state = "whitepurple"
-	},
-/area/fiorina/station/research_cells)
 "sAK" = (
 /obj/item/device/flashlight/lamp/tripod,
 /turf/open/floor/prison{
@@ -31707,8 +31679,8 @@
 /area/fiorina/maintenance)
 "tpt" = (
 /obj/structure/closet/wardrobe/chaplain_black,
-/obj/effect/spawner/random/goggles,
 /obj/effect/landmark/objective_landmark/close,
+/obj/item/clothing/glasses/thermal/syndi/bug_b_gone,
 /turf/open/floor/prison/chapel_carpet,
 /area/fiorina/station/chapel)
 "tpO" = (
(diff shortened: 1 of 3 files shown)


fira:
If they're useful primarily for surv specifically you could just make their position radomized and/or have them break down after a certain amount of use

Steelpoint:
Full randomization is my belief in the way forward, I'd want to discourage 'secrets' that only veteran players will find, or those who delve into the code. 

That, and all items should be very far into the map, perhaps even literally in the hive spots. Perhaps you can reward it by making the items a subtype that is unmeltable so the xenos can't deny the winning marines a reward.

Skatuk:
Can we get thermals for each map thanks - CLF Main

c4xmaniac2:
> Full randomization is my belief in the way forward, I'd want to discourage 'secrets' that only veteran players will find, or those who delve into the code.
> 
> That, and all items should be very far into the map, perhaps even literally in the hive spots. Perhaps you can reward it by making the items a subtype that is unmeltable so the xenos can't deny the winning marines a reward.

Full randomization has already proven to be a failure with the science annex changes done to it over a year ago. Survivor balance should be made with opportunity cost and risk versus reward in mind. If I spend 3 minutes going to an area and finding/obtaining an item I should not be rewarded with a random item (especially no item reward). Think back to how Halo 1, 2, 3, and Reach balanced the placement of power weapons/pick ups. The energy sword is placed in the center of the map where it has the highest risk, highest opportunity cost, and highest time investment to gain. This is also how quake and other similar games balance power pick ups. They're not randomized. Every side knows where everything is, its up to the individual to create a path from their already randomized spawn to acquire the tools they wish to use to survive. These items are being primarily placed in the sandbox for the survivors to use - with it being kept in mind that marines could get a hold of them, but the primary purpose of them being placed in the sandbox is for the survivor vs xeno TDM phase. Giving power weapons/pickups statuses such as unmeltable so marines could pick up a power item placed in the hive does not follow with any valid game design rule. This essentially would function as a "win more" mechanic, and would still be rushed by veteran players. This defeats the whole point. 

I can go into more detail, outside of maps which have had their loot meticulously balanced like chances claim, they should follow a kind of "strata." Where low valueable loot (think common gear) is available close to where most survs spawn and initially very safe areas like LZs. This kind of loot would be pistols, some rifle/shotgun ammo and very basic medical stuff. This would quantify a "Green zone." Extending out towards the center of the map would begin the "yellow zones" where risk vs reward begins to apply more. Here you would have better weapon spawns, access to medical kit, perhaps minor stims or something of the sort in harder to reach areas (more time consuming). From here you get to highly dangerous areas, the "Red Zones." These areas are close to the hives and have the best loot. Think flamers, maybe a stim like runnerblood-neuroblood, tac shotties, thermals and so on. The difficulty is that you do not want to encourage hive diving, and so these spawns must be placed accordingly to avoid survivors rushing and intercepting larva/young queens moving between hive spawn locations. So placing good loot inside of hive spawns should be avoided. 

image

Here's a pretty crude example on what I proposed to Morrow a while ago for rebalancing science annex to get rid of the randomized loot spawns. If Zonespace wants to approve it too, I can start to work on it when I have time. I want to do some other stuff like adding in gates to delay progress into certain areas for survivors and so on, but that's a topic not relevant to this discussion. My point in all of this is that fixes such as randomizing loot or deleting kit is a bandaid fix that doesn't solve the problem, nor does it make for fun gameplay. Power items make the sandbox more fun, not less. This is not a competitive game at the end of the day, and we don't need to streamline everything. I believe you can put the most comically overpowered things into your map so long as the risk and time cost to acquire the item is equivalent to the value of the item.

Steelpoint:
My concern that pre-placement of powerful items will only benefit veteran players/powergamers, and that it won't actually resolve any issue no matter where you put the item unless it spawns underneath the Queen.

No matter where else you put a overpowered item on the map, all it will take is a 60 second detour for a survivor to run straight there, pick it up, then rejoin the other survivors

"Hey guys, I'm going to go grab the OP items, you can go set up in the meta hold" is how I expect this change to go. 

Alternativity the only change will be the usual suspects taking an extra 5 to 15 seconds to grab their overpowered gear than currently. 

Unless you go full randomization, or you make these items break when Marines arrive, then I think the items should still be removed as simply placing the items further inside a map makes no real effective change.

Zonespace27 (changes requested):
This is a little out of scope, but could you make these take power a la NVGs, except that they drain much faster?

c4xmaniac2:
> This is a little out of scope, but could you make these take power a la NVGs, except that they drain much faster?

Beyond my ability, sorry.

cm13-github:

Steelpoint:
If this PR does not go ahead can I put forward re-opening the removal version as the current overpowered items are not at all balanced with the current game state.

c4xmaniac2:
Why not just implement this fix and then add then batteries later?

PR #4787, opened 2023-10-26
Renames var/final

BeagleGaming1:
# About the pull request

It kept giving fake "errors"

It annoyed me

# Changelog

No player facing changes

### DIFF
--- a/code/datums/quadtree.dm
+++ b/code/datums/quadtree.dm
@@ -10,14 +10,14 @@
 	var/z_level
 
 	/// Don't divide further when truthy
-	var/final
+	var/final_divide = FALSE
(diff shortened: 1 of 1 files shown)


morrowwolf:
Removing soul

fira (approved):

harryob (commented):
this should be fixable on the dreamchecker end with https://github.com/SpaceManiac/SpacemanDMM/issues/375 as it is valid dm. if it’s not all too easy to fix we can just merge this, i suppose

harryob (approved):
ah sure, sdmm cut a new release so i wanna use that

PR #4792, opened 2023-10-27
Crusher Charge Change

Steelpoint:
# About the pull request

Crushers charge ability now has no windup time, it now has a range limit of 4 tiles. 

The charge ability is extremely non viable for a crusher to use, the windup time makes it almost virtually impossible for a crusher to land a hit on a target, which makes the crusher have very little ability to effect a strong initial attack 

This change will make the charge a far more skill based ability that gives the crusher the ability to better dictate the opening of an attack, as well as make the charge far more useful as an actual offensive ability. The loss of the range of the attack will ensure that there is a hard limit to how far a charger can charge and avoid situations of the crusher instantly hitting someone across the screen. 

<details>

</details>

# Changelog
:cl:
balance: Crushers charge ability no longer has a windup timer, but is now limited to a range of 4 tiles
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/crusher/crusher_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/crusher/crusher_abilities.dm
@@ -8,13 +8,13 @@
 	xeno_cooldown = 140
 	plasma_cost = 20
 	// Config options
-	distance = 9
+	distance = 4
 	knockdown = TRUE
 	knockdown_duration = 2
 	slash = FALSE
 	freeze_self = FALSE
 	windup = TRUE
-	windup_duration = 12
+	windup_duration = 0
 	windup_interruptable = FALSE
 	should_destroy_objects = TRUE
 	throw_speed = SPEED_FAST


Yellows91:
I just want to leave a concern here

T3 crusher is meant for SIEGE warfare. Why remove its long range ram that is meant to hit cades and break marine cadelines.

Steelpoint:
Crushers can still engage Marine barricades with the change to how the charge works. The change puts more emphasis on a Crusher's ability to actually engage in combat outside of either smashing a cade 2 slashes and retreating, or hiding behind a resin door and praying a marine stands still until you hit them before you slowly retreat back to cover.

GoldenDarkness55:
Imo it'd be better to lower the charge time significantly and -1 range on it, which fixes both "offscreen" charges (you walk on the tile as crusher finishes cast time) and the issue with it being impossible to hit, without introducing new problems - too short range and a guaranteed charge hit which would be really silly and annoying considering CD reduction per slash

Zonespace27:
Thank you for making this PR, but I am not interested in making this change at this time.

PR #4802, opened 2023-10-29
extends invis wall

harryob:
re-adds two blocks of invis wall accidentally removed in https://github.com/cmss13-devs/cmss13/pull/632 1 morbillion years ago on soro


realforest2001 (approved):
All G

PR #4804, opened 2023-10-29
Fix runtimes in toxic water targeting unliving

fira:
# About the pull request

You've heard this story before. This causes runtimes with Queen Eye and probably other camera mobs. Observers exempt due to a dead check later down.

No player facing changes.


harryob (approved):
ismob() i hate you

PR #4807, opened 2023-10-30
Brings back NVG

ihatethisengine:
# About the pull request

Re-adds NVG (Night Vision Goggles) back to gameplay, removes NVO (Night Vision Optics (visor thing)) from gameplay.

Honestly, I was going to try to fix NVO, but in the end I decided it is not really worth it. Too many issues: 
1) Too inconvenient to use. Yeah switching between visors is a real pain.
2) Too inconvenient to recharge. You need someone to bring recharger, otherwise you won't be able to use the NVO more than for few minutes. 
3) NVO are way less balanced. You can use the binoculars with NVO, which is a HUGE advantage out of nowhere. Also, you can't break NVO as a xeno, and I think being able to break NVGs was a nice little counterplay.
4) It's really hard to tell if marine is using NVO, even if you know where to look. It's just 1 blinking pixel that doesn't really change the shape of the helmet. There is also some tiny green glow, but it's probably only relevant in HvH, as xeno you won't see it. Really, I was playing xeno for last months and I think I noticed NVO only once. The sprite is not very good either and it's also broken (it doesn't blink when you are facing north iirc).

While everything I mentioned can be fixed, I'm not sure it's worth it. Maybe NVO matches the aesthetics a little better, but it requires too much repairs to actually work and I'm not willing to spend that much time and energy fixing someone else's project (and I doubt that _someone_ will fix it either). And I cannot make new sprites for NVO anyway. So this is why I propose a revert to the thing that worked flawlessly and everyone was happy with.

<details>
I did not test it, but no one read this anyway.

</details>

# Changelog
:cl: ihatethisengine
add: Night Vision Goggles are obtainable again.
del: Night Vision Optics are no longer obtainable.
/:cl:

### DIFF
--- a/code/datums/supply_packs/gear.dm
+++ b/code/datums/supply_packs/gear.dm
@@ -63,3 +63,15 @@
 	containertype = /obj/structure/closet/crate/ammo
 	containername = "fulton recovery device crate"
 	group = "Gear"
+
+/datum/supply_packs/nvg
+	name = "M2 Night Vision Goggles Crate (x3)"
+	contains = list(
+		/obj/item/prop/helmetgarb/helmet_nvg,
+		/obj/item/prop/helmetgarb/helmet_nvg,
+		/obj/item/prop/helmetgarb/helmet_nvg,
+	)
+	cost = 60
+	containertype = /obj/structure/closet/crate/supply
+	containername = "M2 Night Vission Goggles Crate"
+	group = "Gear"
--- a/code/game/machinery/vending/vendor_types/intelligence_officer.dm
+++ b/code/game/machinery/vending/vendor_types/intelligence_officer.dm
@@ -7,7 +7,7 @@ GLOBAL_LIST_INIT(cm_vending_gear_intelligence_officer, list(
 		list("SUPPLIES", 0, null, null, null),
 		list("Power Control Module", 5, /obj/item/circuitboard/apc, null, VENDOR_ITEM_REGULAR),
 		list("Binoculars", 5, /obj/item/device/binoculars, null, VENDOR_ITEM_REGULAR),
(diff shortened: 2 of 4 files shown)


cuberound:
kinda sad, being able to use it with binos was the only good part about NVO

Nanu308:
This just seems like a revert of a change Morrow made, and i am more in favor of Marines not having NVG's readily available at all since they are incredibly strong and darkness is meant to be in favor of the xenos.

This goes against that gameplay design entirely, so i am adding the DNM label to it because i personally do not wish to see this ingame.

ihatethisengine:
> This just seems like a revert of a change Morrow made, and i am more in favor of Marines not having NVG's readily available at all since they are incredibly strong and darkness is meant to be in favor of the xenos.
> 
> This goes against that gameplay design entirely, so i am adding the DNM label to it because i personally do not wish to see this ingame.

Well, we have NVGs in one form or another for over a year (or two?) already, so unless you want to remove NVG (including Morrow's version) whatsoever, you should merge this PR since this version favors xenos way more:
1) Can't use with binos (NVGs with binos is an insane advantage, not comparable to anything).
2) Xenos can break NVGs, which gives them some counter-play.
3) Sprite is way more obvious, so you can notice an NVG user before they start shooting you.

Only significant downside of Morrow's version is recharging, but it doesn't affect xeno gameplay directly. I am saying this as a certified xenomain. So again, unless you want to remove NVG whatsoever, this version is more fitting for the gameplay than Morrow's.

cm13-github:

iloveloopers:
> This just seems like a revert of a change Morrow made, and i am more in favor of Marines not having NVG's readily available at all since they are incredibly strong and darkness is meant to be in favor of the xenos.
> 
> This goes against that gameplay design entirely, so i am adding the DNM label to it because i personally do not wish to see this ingame.

"this PR changes NVGs to a previously weaker state which will make it more avaliable to marines" yeah this makes total sense

Issue #4777, opened 2023-10-24
Vehicle interiors can't hear exteriors


Doubleumc:
### Testmerges

N/A (localhost)

### Round ID

N/A (localhost)

### Description of the bug

The majority of a vehicle's interior cannot hear any external noises. The northeast corner can potentially hear it, but the volume is so quiet that most noises are inaudible.
![image](https://github.com/cmss13-devs/cmss13/assets/14267245/add631d4-41f4-49be-a01a-1db10b52d764)
![image](https://github.com/cmss13-devs/cmss13/assets/14267245/5d9d6903-4f2a-4d90-ad84-44c6d428b614)

I used the following external test apparatus. The queen's armor was set to 100 and I was seated within the APC:
![image](https://github.com/cmss13-devs/cmss13/assets/14267245/6b995f55-7956-4856-ae29-82338ebefd81)

Within the highlighted **CAN HEAR** area, external audio noises are played at an extremely low volume and are logged by the `Dump Playing Client Sounds` command. Note the bullet sounds being logged while inside the APC:
![image](https://github.com/cmss13-devs/cmss13/assets/14267245/82ccc78d-eab4-4544-883a-cfa06b9627f9)

Inside the APC but outside the **CAN HEAR** area, only the interior ambience is logged:
![image](https://github.com/cmss13-devs/cmss13/assets/14267245/18c5a4a8-702b-4039-8e1f-43079858eae4)

Prior to commit https://github.com/cmss13-devs/cmss13/commit/25255039c600e3e7b7e8aff60bc23611e49806a2 interior audio worked as expected, the interiors could hear exterior noises.

After commit https://github.com/cmss13-devs/cmss13/commit/25255039c600e3e7b7e8aff60bc23611e49806a2 and before commit https://github.com/cmss13-devs/cmss13/commit/0808500777f390bbe1d59142e955678344ef6d3e interiors had audio issues. Anyone within the **CAN HEAR** area could hear interior and exterior noises. Anyone inside the vehicle but outside the **CAN HEAR** area couldn't hear external **or** internal noises, such as internal gunfire or emotes.

After commit https://github.com/cmss13-devs/cmss13/commit/0808500777f390bbe1d59142e955678344ef6d3e interiors had different audio issues. Internal noises work as expected, anyone in the interior can hear them. External noises can only be heard within the **CAN HEAR** area, and at a very low volume.

### What's the difference with what should have happened?

External audio should be consistently audible within a vehicle interior.

[...]

PR #4815, opened 2023-10-31
Vehicles hear exterior audio

Doubleumc:
# About the pull request

As described in issue #4777 while vehicle interiors are intended to hear external sounds, it is largely nonfunctional. Sounds are either not heard or are inaudibly quiet. This is due to several smaller issues working together.

First, interiors are not detected as interiors. In commit https://github.com/cmss13-devs/cmss13/commit/0808500777f390bbe1d59142e955678344ef6d3e the types stored in the `used_turfs` list were changed to weakrefs, but the `is_interior()` function still checked for the old type, causing it to always return `FALSE`: https://github.com/cmss13-devs/cmss13/blob/5fbe14390f80225a5d295f07b9acf781fb8aa2b2/code/controllers/subsystem/interior.dm#L54-L58
Fixed by checking for the correct type.

Second, the `get_interior_by_coords()` function returned an object on success, but returned `FALSE` on failure - which still counted as an object. Since shuttles add turfs to `used_turfs` but don't have actual interiors, they would pass `is_interior()` but fail `get_interior_by_coords()`. The sound code expected any object received at this point to be valid, but using `FALSE` would thrown an error: https://github.com/cmss13-devs/cmss13/blob/5fbe14390f80225a5d295f07b9acf781fb8aa2b2/code/game/sound.dm#L82-L84
Fixed by returning `null` on failure.

Third, the audio was now playing but in the wrong part of the interior. It was offset to the top-right corner because `get_middle_coords()` got the difference between the bounds, but never actually averaged them: https://github.com/cmss13-devs/cmss13/blob/5fbe14390f80225a5d295f07b9acf781fb8aa2b2/code/modules/vehicles/interior/interior.dm#L321
Fixed by dividing the difference by two.

Fourth, the sound playback zone was oriented incorrectly because the width and height were transposed when creating the zone. Rectangles were defined as width then height, but the sound subsystem was passing height then width: https://github.com/cmss13-devs/cmss13/blob/5fbe14390f80225a5d295f07b9acf781fb8aa2b2/code/controllers/subsystem/sound.dm#L44
Fixed by passing width then height.

Vehicle interiors were intended to hear external sounds. This restores that functionality.
<details>

</details>

# Changelog
:cl:
fix: vehicle interiors can hear exterior noises
/:cl:

### DIFF
--- a/code/controllers/subsystem/interior.dm
+++ b/code/controllers/subsystem/interior.dm
@@ -42,7 +42,7 @@ SUBSYSTEM_DEF(interior)
 			continue
 		if(x >= bounds[1].x && x <= bounds[2].x && y >= bounds[1].y && y <= bounds[2].y)
 			return current_interior
-	return null
+	return
 
 /// Checks if an atom is in an interior
 /datum/controller/subsystem/interior/proc/in_interior(loc)
--- a/code/controllers/subsystem/interior.dm
+++ b/code/controllers/subsystem/interior.dm
@@ -42,7 +42,7 @@ SUBSYSTEM_DEF(interior)
 			continue
 		if(x >= bounds[1].x && x <= bounds[2].x && y >= bounds[1].y && y <= bounds[2].y)
 			return current_interior
-	return FALSE
+	return null
 
 /// Checks if an atom is in an interior
 /datum/controller/subsystem/interior/proc/in_interior(loc)
@@ -51,7 +51,7 @@ SUBSYSTEM_DEF(interior)
 	if(!isturf(loc))
 		loc = get_turf(loc)
 
-	var/datum/turf_reservation/interior/reservation = SSmapping.used_turfs[loc]
+	var/datum/weakref/reservation = SSmapping.used_turfs[loc]
 
 	if(!istype(reservation))
 		return FALSE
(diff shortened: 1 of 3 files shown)


fira:
What do you mean by "FALSE is still counted as an object" ? FALSE is literally 0 which is falsey, this shouldn't change anything for the purposes of what's done here

Doubleumc:
I might be using the wrong words here. What I mean is the sound code takes the return of `get_interior_by_coords()` (`VI`), checks if it exists (`VI?`), and uses the `ready` var from it (`VI?.ready`): https://github.com/cmss13-devs/cmss13/blob/5fbe14390f80225a5d295f07b9acf781fb8aa2b2/code/game/sound.dm#L83-L84

It doesn't check if its `TRUE` or `FALSE`, it checks if it exists or not. And `FALSE`/`0` exists, but can't be used in this way:
image

code/controllers/subsystem/interior.dm
```diff
@@ -42,7 +42,7 @@ SUBSYSTEM_DEF(interior)
 			continue
 		if(x >= bounds[1].x && x <= bounds[2].x && y >= bounds[1].y && y <= bounds[2].y)
 			return current_interior
-	return FALSE
+	return null
```
harryob, line 45:
```suggestion
	return
```

harryob (changes requested):
works for me, thanks for the fix

harryob (approved):

PR #4816, opened 2023-10-31
Boots Storage Refactor

fira:
# About the pull request

While discussing with Beagle on the state of boots equipping shitcode, i figured it was a good spot to make a component  which can hopefully be reused later down the line. 

Enter `simple_atom_storage` whose behavior is as follow: 
 * You can apply it to any `/atom` - tracking what's in independently of contents.
 * The storage can store a certain amount of movables.
 * You can only store movables from a list of types in it.
 * Mobs can store and retrieve from it.
 * A sound is played when interacting with it.

If anything it's here just as a practical component example and in hope of reuse. Sure it's no awesome /obj/item/storage but it works for anything that needs to store 1-# movables. Hook signals to user input and voila.

Also this stops knives from being taken out of boots in cryo, which I can't be bothered to write a handler for and was dumb anyway littering cryo list

Example code, modularization

# Changelog
:cl:
qol: Knives won't be removed from boots in cryo storage.
/:cl:


code/datums/components/simple_atom_storage.dm
```diff
@@ -0,0 +1,68 @@
...
+	var/maximum_slots
+	var/sound_file
+
+/datum/component/simple_atom_storage/Initialize(list/allowed_types, maximum_slots, sound_file)
+	. = ..()
+	if(!isatom(parent))
+		return COMPONENT_INCOMPATIBLE
+	src.allowed_types = allowed_types
+	src.maximum_slots = maximum_slots
+	src.sound_file = sound_file
```
Birdtalon, line 15: Hey, for my own information could you explain why you'd use `src.` here wouldn't it be implied?

fira: Because the proc argument variable has the same name
So src.var refers to the datum one while the second one refers to the proc arg one
It's a pattern that seems used a lot in constructors here, i think because it avoids having to make up slightly different variable names which can be confusing

Birdtalon: Brilliant thank you 👍

cm13-github:

code/datums/components/simple_atom_storage.dm
harryob: > Brilliant thank you 👍

fira is correct, and it's explicitly requested in our style guide too (https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#when-passing-vars-through-new-or-initializes-arguments-use-srcvar)

cm13-github:

PR #4837, opened 2023-11-04
Makes rear dropship doors solid again and refactors multi door filler turf logic

Das15:
# About the pull request

Fixes a bug with rear drop ship doors, which allowed people to see through them (you can see space for 1 tick on departure though, dunno how to fix this detail). Refactored filler turfs logic in door parent, so we can avoid redundant code.

Also removed some unnecessary code from door class.

Bugs and old, unnecessary code bad, therefore they should be gone.
- Checked rear drop ship doors on start of the round, 
- Launched the ship, 
- Checked the doors again,
- On arrival checked the doors once again.

# Changelog
:cl: MrDas
fix: You can no longer see through solid rear drop ship doors.
code: Moved filler turf logic from almayer airlocks to its parent door.
/:cl:


code/game/machinery/doors/door.dm
```diff
@@ -137,64 +147,57 @@
...
 			else
 				qdel(src)
 	return
 
-
 /obj/structure/machinery/door/get_explosion_resistance()
 	if(density)
 		if(unacidable)
-			return 1000000
+			return 1000000 //I have no idea why do we use magic number here...
```
Birdtalon, line 193: I am making a big assumption here but I'm guessing this is just here to be a "big number" in order to negate explosions. I reckon it's just been copypasted ever since.
https://github.com/cmss13-devs/cmss13/blob/4ff511287574c8613cb080d5038c8693b3d849b8/code/__DEFINES/misc.dm#L35

Explosion max power define is 5000 and looks like it's used in this calculation.

https://github.com/cmss13-devs/cmss13/blob/4ff511287574c8613cb080d5038c8693b3d849b8/code/datums/autocells/explosion.dm#L168-L172

code/game/machinery/doors/door.dm
```diff
@@ -137,64 +147,57 @@
...
 		try_to_activate_door(user)
-		return 1
+		return TRUE
 
 /obj/structure/machinery/door/emp_act(severity)
-	if(prob(20/severity) && (istype(src,/obj/structure/machinery/door/airlock) || istype(src,/obj/structure/machinery/door/window)) )
-		open()
+	if(prob(20/severity))
+		if((istype(src, /obj/structure/machinery/door/airlock) || istype(src, /obj/structure/machinery/door/window)))
+			open()
```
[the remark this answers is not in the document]
Das15: I wonder why do we even need this check in the first place? If we want to make it not work with non-powered doors, we can just check for use_power, the class already has open() implemented.

code/game/machinery/doors/door.dm
Das15: Yup, sounds like something defines would be good at, I will leave a comment for now, maybe someone will take care of this, hopefully.

ihatethisengine:
You are supposed to see through lifeboats doors (which is a subtype of reardoor), so make sure it's working

Das15:
> You are supposed to see through lifeboats doors (which is a subtype of reardoor), so make sure it's working

Yup, checked just in case, lifeboat doors work as they are supposed to

code/game/machinery/doors/door.dm
```diff
@@ -7,72 +6,85 @@
...
 
 /obj/structure/machinery/door/initialize_pass_flags(datum/pass_flags_container/PF)
 	..()
 	if (PF)
 		PF.flags_can_pass_all = NONE
 
+/// Also refreshes filler_turfs list.
+/obj/structure/machinery/door/proc/change_filler_opacity(new_opacity)
+	// I have no idea why do we null opacity first before... changing it
+	for(var/turf/filler_turf in filler_turfs)
```
harryob, line 56:
```suggestion
	for(var/turf/filler_turf as anything in filler_turfs)
```

code/game/machinery/doors/door.dm
```diff
@@ -7,72 +6,85 @@
...
+		filler_turf.set_opacity(null)
+
+	filler_turfs = list()
+	for(var/turf/filler in locate_filler_turfs())
```
harryob, line 60:
```suggestion
	for(var/turf/filler as anything in locate_filler_turfs())
```

code/game/machinery/doors/door.dm
```diff
@@ -7,72 +6,85 @@
...
+		change_filler_opacity(opacity)
 
-//process()
-	//return
+/// Finds turfs which should be filler ones.
+/obj/structure/machinery/door/proc/locate_filler_turfs()
+	var/turf/filler_temp
+	var/list/located_turfs = list()
+
+	for(var/i = 1, i < width, i++)
```
harryob, line 80:
```suggestion
	for(var/i = 1 in 1 to width - 1)
```

code/game/machinery/doors/door.dm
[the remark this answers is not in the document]
harryob: please do, istype(src is a sin unto itself

code/game/machinery/doors/door.dm
```diff
@@ -256,20 +256,16 @@
...
 	return TRUE
 
-
-/obj/structure/machinery/door/proc/update_flags_heat_protection(turf/source)
-
-
 /obj/structure/machinery/door/proc/autoclose()
-	var/obj/structure/machinery/door/airlock/A = src
-	if(!A.density && !A.operating && !A.locked && !A.welded && A.autoclose)
+	var/obj/structure/machinery/door/airlock/airlock = src
```
harryob, line 267: this typecast should be pointless and you can just check density operating locked etc below. if you can't this proc shouldn't be a `/obj/structure/machinery/door` proc it should ee a `/obj/structure/machinery/door/airlock` proc

code/modules/shuttle/shuttles/dropship.dm
```diff
@@ -55,6 +55,14 @@
 /obj/docking_port/mobile/marine_dropship/proc/get_door_data()
 	return door_control.get_data()
 
+/// Is called on_departure and on_arrival to prevent transparent solid multitile airlocks
+/obj/docking_port/mobile/marine_dropship/proc/refresh_multitile_airlock_filler_turfs()
+// Yes, there is 1 tick where you can see through solid airlocks, but it's the simplest solution I could've think of
+	for(var/place in shuttle_areas)
+		for(var/obj/structure/machinery/door/doors in place)
+			if(doors.width > 1)
+				doors.handle_multidoor()
```
harryob, line 64: this pattern ends up iterating through every atom in the world, unfortunately (as areas do not actually keep the stuff in their area in a list).  you should probably put this logic in /obj/structure/machinery/door/onShuttleMove()

harryob (changes requested):

code/modules/shuttle/shuttles/dropship.dm
Das15: For some reason if I try to put the logic in onShuttleMove() it doesn't work even though filler_turfs get set properly, afterShuttleMove() however works similarly to this, so I'll be using that.

harryob (approved):

cm13-github:

PR #4853, opened 2023-11-06
fixes merge conflict markers

harryob:
title

no playerfacing changes apart from no merge conflict markers getting through ci again


Drulikar (approved):
Code changes look fine.

code/modules/mapping/merge_conflicts.dm
```diff
@@ -0,0 +1,16 @@
...
+	icon = 'icons/landmarks.dmi'
+	icon_state = "merge_conflict_marker"
+	desc = "If you are seeing this in-game: someone REALLY, REALLY, REALLY fucked up. They physically mapped in a fucking Merge Conflict Marker. What the shit."
+
+///We REALLY do not want un-addressed merge conflicts in maps for an inexhaustible list of reasons. This should help ensure that this will not be missed in case linters fail to catch it for any reason what-so-ever.
+/obj/merge_conflict_marker/Initialize(mapload)
+	. = ..()
+	var/msg = "HEY, LISTEN!!! Merge Conflict Marker detected at [AREACOORD(src)]! Please manually address all potential merge conflicts!!!"
+	to_chat(world, SPAN_BOLDANNOUNCE("[msg]"))
+	warning(msg)
```
Drulikar, line 16: Although, this doesn't scream loud enough so will go unnoticed until someone actually tests in game and reads chat during startup. This doesn't currently result in a unit test failure; so it would be better if it threw it as a runtime instead of a debug message for check_runtimes to pick up on.

I tried testing two branches off of this branch both editing the basket ball on the Almayer: 
image
image
image
image

```suggestion
	CRASH(msg)
```

If you do above then: 
image

Drulikar (changes requested):

code/modules/mapping/merge_conflicts.dm
harryob: > Although, this doesn't scream loud enough so will go unnoticed until someone actually tests in game and reads chat during startup. This doesn't currently result in a unit test failure; so it would be better if it threw it as a runtime instead of a debug message for check_runtimes to pick up on.
> 
> I tried testing two branches off of this branch both editing the basket ball on the Almayer: image image image image
> 
> If you do above then: image

sure, but it is also picked up in run linters currently

Drulikar (approved):

realforest2001 (approved):

PR #4828, opened 2023-11-02
/tg/ Status effects part 1 - fluid status updates

fira:
image
CM Dev figured it out 9 years ago and nobody listened and kept tacking illogical conditions

# About the pull request

Builds on previous "prelude" PRs in the following steps:
 * Ports /tg/ body_position and mobility_flags
 * Fixes some interaction requirements to use stun/mobility rather than lying/knocked_down
 * Ports /tg/ granular status updates, ie. status propagating through handlers and signals. This includes changes to resting, buckling, and lying down human transforms.
 * Wires our status effect system to it directly
 * Removes `update_canmove` from existence completely as not needed anymore

Because step 1 and 2 require updating all the gameplay logic using them, this PR modifies a lot of files.

Part 2 will move the actual status effects to /tg/ status_effects, resolving our timing problems.

# Testing Photographs and Procedure

<details>
<summary>Testing Checklist!</summary>

- [x] Basic Movement
- [x] Admin Freeze Prevents Movement
- [x] Resting, Getting Up
- [x] Xenos change icon when resting
- [x] Buckling, including bed rotation and propelled chairs
- [x] Crawling Movement including sprite movement
- [x] Aggressive, Choke Grabbing, and Fireman carry apply rotation
- [x] Xeno Pounce and Abduct properly freeze both target and caster 
- [x] Double dropship seats density update
- [x] Explosive knockout on Humans
- [x] Xeno burrow density and movement interactions
- [x] Xeno nest interactions, specifically confirm density changes work
- [ ] Xeno nest bullet hits doublecheck with snowflake trait check
- [x] Combat Xeno ~~knockouts~~ knockdown and sprite updates
- [x] ~~Sleeping, Waking up, Usage of items while sleeping~~ - Can't really test this we have almost no sleep code
- [x] Arbitrary buckling rotations
- [x] Admin-set transforms work with buckling/lying
- [ ] All the broken objects that will only be found out in Testmerge

</details>

# Changelog

:cl:
code: Ported basic /tg/ Status Backend.
add: Human transform changes such as lying down, knock down, buckling, are now animated.
fix: Some statuses will now take effect immediately instead of waiting for a life tick, notably Resting.
balance: Many interaction requirements were changed to eg. fail upon stuns rather than if lying down.
/:cl:

PR #4869, opened 2023-11-07
/tg/ Status Effects Prelude A - Updated Trait Backend

fira:
# About the pull request

Content from #4828 broken down in an individual PR

This does only a few things:
 * Adds /tg/ ASSERT which includes reasons
 * Updates the Traits backend from /tg/
 * Refactors cloak cancellation with signals because it used REMOVE_TRAIT_ALLSOURCES which is both bad and defunct

Tested trait backend as part of main PR


code/game/objects/items/handcuffs.dm
```diff
@@ -29,20 +29,6 @@
 	if(!C.handcuffed)
 		place_handcuffs(C, user)
 
-/obj/item/handcuffs/obj/structure/MouseDrop(mob/living/carbon/human/H)
```
Zonespace27, line 32: cursed

code/modules/mob/mob.dm
```diff
@@ -513,6 +513,32 @@
 
 	return do_pull(AM, lunge, no_msg)
 
+/mob/proc/stop_pulling()
+	if(pulling)
```
Zonespace27, line 517: early return

code/modules/mob/mob.dm
```diff
@@ -513,6 +513,32 @@
...
+		var/mob/M = pulling
```
Zonespace27, line 518: one letter var

code/modules/mob/mob.dm
```diff
@@ -513,6 +513,32 @@
 
 	return do_pull(AM, lunge, no_msg)
 
+/mob/proc/stop_pulling()
```
Zonespace27, line 516: code doc

code/modules/mob/mob.dm
```diff
@@ -513,6 +513,32 @@
...
+		pulling.pulledby = null
+		pulling = null
+
+		grab_level = 0
+		if(client)
+			client.recalculate_move_delay()
+			// When you stop pulling a mob after you move a tile with it your next movement will still include
+			// the grab delay so we have to fix it here (we love code)
+			client.next_movement = world.time + client.move_delay
+		if(hud_used && hud_used.pull_icon)
```
Zonespace27, line 528:
```suggestion
		if(hud_used?.pull_icon)
```

Zonespace27 (changes requested):

code/game/objects/items/handcuffs.dm
fira: It is, that's why i'm getting rid of it on the spot

Zonespace27 (approved):

PR #4871, opened 2023-11-07
/tg/ Status Effects Prelude C - Mob Density Refactor

fira:
# About the pull request

Part of #4828

Implements the undense trait. This both lets us stack and unstack sources of undensity properly avoiding issues eg. we've had with nests in the past, and lays some groundwork for updates when main PR gets rid of update_canmove 

Minor testing, mostly tested as part of parent PR

# Changelog
:cl:
fix: Density effects should now properly update when stacked and unstacked.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -16,6 +16,7 @@
 
 	attack_icon = image("icon" = 'icons/effects/attacks.dmi',"icon_state" = "", "layer" = 0)
 
+	register_init_signals()
 	initialize_incision_depths()
 	initialize_pain()
 	initialize_stamina()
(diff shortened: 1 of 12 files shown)


code/modules/mob/living/living.dm
```diff
@@ -463,3 +464,9 @@
 	switch(stat)
 		if(DEAD)
 			SEND_SIGNAL(src, COMSIG_MOB_STAT_SET_DEAD)
+
+/mob/living/proc/update_density()
```
Zonespace27, line 468: ///

Zonespace27 (changes requested):

PR #4881, opened 2023-11-08
Should fix pumpkins, PUMPKINS EVERYWHERE

fira:
# About the pull request

Commit from https://github.com/PvE-CMSS13/PvE-CMSS13/commit/0aa95b1d3692b63eed6b59ba5bca27db4f5faae6

Turns out i'm a monkey and didn't think about This One Small Thing when refactoring, so the decorator keeps applying even without being active. Neat huh.

Pumpkins overstepping their bounds. It's not much of a holiday if it's on all the time...

# Changelog

:cl: MorrowWolf
fix: Fixed Pumpkins appearing past season.
/:cl:


Drulikar (approved):
A new way to blame Morrow for changes, cherry picked commits :D

Anyways, this code looks fine. Other way would be to have the logic determining quantity of pumpkins when it gets -1 to do nothing, but this is more performant.

PR #4884, opened 2023-11-09
Colonial Space Grunts

Releasethesea:
# About the pull request

Adds the tabletop game known as Colonial Space Grunts, this will enable roleplay for shipside roles when they have downtime.

yes i have permission to add this in game, it was done by a good friend of mine who sadly doesnt play much anymore. 

Colonial space grunts is a simple to learn and easy to get into tabletop game, it can be played in a single game, ive ran several events with this in the past and all of them have had positive reception, so i figured id make it permenantly in game

<details>

</details>

# Changelog
:cl:
add: Adds Colonial Space Grunts to rec vend aswell as several pens and Dice for playing the game.
/:cl:


code/modules/paperwork/paper.dm
```diff
@@ -906,3 +906,8 @@
 
 	info = parsepencode(template, null, null, FALSE)
 #undef MAX_FIELDS
+
+/obj/item/paper/Almayer/CSG
```
Birdtalon, line 910: Not sure if there is anything in the style guide but type paths are usually all lower case

fira: yeah please use colonial_space_grunts or space_grunts or grunts_manual something similar

code/modules/paperwork/paper.dm
```diff
@@ -906,3 +906,8 @@
...
+	name = "Colonial Space Grunts"
+	desc = "A tabletop game based around the USCM, easy to get into, simple to play, and most inportantly fun for the whole squad."
+	info = "<div> <img style='align:middle' src='ColonialSpaceGruntsEZ.png'>"
```
fira, line 913: Just one more thing - you need to use SSassets link generator or this won't work on the live server due to CDN setup
You can do this by using `SSassets.transport.get_asset_url("name_of_the_file")`
Because this would be done at runtime, you need to set it within an Initialize proc

code/modules/paperwork/paper.dm
fira: You can remove the /Almayer too i think

fira (changes requested):

code/modules/paperwork/paper.dm
fira: eg instead of just doing info =   you'd do 

```
/obj/item/paper/Almayer/CSG/Initialize(mapload, ...)
  . = ..()
  info = "...."
```

fira (approved):

PR #4886, opened 2023-11-09
Refactor of ammo and it's children file to clean a bit BE code.

Huffie56:
# About the pull request
Make it so that all child of ammo can use BE code instead of just bullet child.

<details>

</details>

# Changelog
:cl:
refactor: moved the two proc that handled BE to ammo object instead of bullet object allowing more gun to BE in the future.
refactor: removed some instance where variable where change for the value they already had.
/:cl:


Huffie56:
noticed that damage isn't set by define value ... not sure what to do with this but i could refactor this here if needed...

code/datums/ammo/bullet/pistol.dm
```diff
@@ -167,8 +166,6 @@
 
 /datum/ammo/bullet/pistol/highpower
 	name = "high-powered pistol bullet"
-	headshot_state = HEADSHOT_OVERLAY_MEDIUM
```
fira, line 170: why remove the overlays?

Huffie56: because it is already set the same way  in is parent /datum/ammo/bullet/pistol

Huffie56:
https://discord.com/channels/150315577943130112/964684928161808384/1175457155604828251

Birdtalon:
Lets say at some point in the future we want to allow BE with other types of ammo. The user feedback and applied effects of these procs don't fit all `/datum/ammo`  so it would have to be refactored at some point. When are we going to see BE with `/datum/ammo/xeno/acid` 😄. But to be honest I'm not sure I see the benefit of moving these out of the scope of `datum/ammo/bullet` at this time.

Huffie56:
> Lets say at some point in the future we want to allow BE with other types of ammo. The user feedback and applied effects of these procs don't fit all `/datum/ammo` so it would have to be refactored at some point. When are we going to see BE with `/datum/ammo/xeno/acid` 😄. But to be honest I'm not sure I see the benefit of moving these out of the scope of `datum/ammo/bullet` at this time.

yea i am probably going to close this PR and do something else around those file their is lot's to do on them....
turning random damage into define come to mind...

Huffie56:
not a real point to this PR closing it down

PR #4868, opened 2023-11-07
Adds possibility to customize crates

fira:
# About the pull request

You can now customize non-secure crates using packaging wrap, by tagging them and changing their appearance to a list of common crates. The tag works as a labeler label, and is also displayed in the overwatch supply drop alert.

# Explain why it's good for the game
If anything, it's that these Squad crates are just sitting there unused. I don't know if this feature will be used much by live Reqs.

# Changelog
:cl:
add: You can now use package wrap to change standard crates visuals.
add: You can now label crate with package wrap, which will also be shown during overwatch supply drop alert.
mapadd: Added more packaging wrap to Requisitions.
/:cl:

PR #4889, opened 2023-11-10
Echo pad droppod fix

fira:
# About the pull request

I manage to screw something up with every PR it seems - see #4868 

Updated droppod constructor arguments in droppod cosmetic PR (to make it so you have the crate falling down sky) but only did so for Overwatch pods, not Echo's.

This causes a deleted droppod to be launched anyway and remain on the ground. AWKWARD.

# Changelog
:cl:
fix: Fixed ghost droppods appearing when launching crates by Echo pad.
/:cl:

### DIFF
--- a/code/game/supplyshuttle.dm
+++ b/code/game/supplyshuttle.dm
@@ -327,7 +327,7 @@ var/datum/controller/supply/supply_controller = new()
 		M.count_niche_stat(STATISTICS_NICHE_CRATES)
 
 	playsound(C.loc,'sound/effects/bamf.ogg', 50, 1)  //Ehh
-	var/obj/structure/droppod/supply/pod = new()
+	var/obj/structure/droppod/supply/pod = new(null, C)
 	C.forceMove(pod)
 	pod.launch(T)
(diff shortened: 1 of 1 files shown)


code/game/supplyshuttle.dm
```diff
@@ -327,7 +327,7 @@ var/datum/controller/supply/supply_controller = new()
 		M.count_niche_stat(STATISTICS_NICHE_CRATES)
 
 	playsound(C.loc,'sound/effects/bamf.ogg', 50, 1)  //Ehh
-	var/obj/structure/droppod/supply/pod = new()
+	var/obj/structure/droppod/supply/pod = new(null, C)
```
Segrain, line 330:
```suggestion
	var/obj/structure/droppod/supply/pod = new(package = C)
```

fira: I wish but unfortunately you can't use kwargs in atom new() due to us using the New->Initialize scheme losing them

image
image

harryob (approved):

PR #4892, opened 2023-11-10
Fix missing delay on climbing objects

fira:
# About the pull request

When you climb a structure such as railings or platform, the movement delay is not currently applied. 
This means that while you're waiting for the climbing windup, the movement comes off cooldown, you climb, and move immediately a second time walking. 

More than concerns about balance or consistency, i'm opening this PR because it's janky: it just *feels* really weird especially given that for a movement that is some distance away, BYOND will NOT apply the movement glide effrect and instead "teleport" you. This would be fine for Req Railings you jump over, but eg. lifeboat pumps platforms you constantly walk on/off simply by moving around.

# Changelog
:cl:
fix: Climbing rails and platforms now respects movement delay, preventing from "teleporting" 2 tiles at once.
/:cl:

### DIFF
--- a/code/game/objects/structures.dm
+++ b/code/game/objects/structures.dm
@@ -145,6 +145,8 @@
 	user.forceMove(TT)
 	for(var/atom/movable/thing as anything in grabbed_things) // grabbed things aren't moved to the tile immediately to: make the animation better, preserve the grab
 		thing.forceMove(TT)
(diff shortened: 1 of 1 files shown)


fira:
this should probably use next_move instead but i'm a bit stupid, will advise

fira:
Yeah no next_move is for actions, this is correct

PR #4917, opened 2023-11-13
Early JSON logging implementation

fira:
# About the pull request

A simple JSON logging implementation for feeding in logstash/opensearch because:
 * I'm sick of seeing unreadable logs containing HTML and href tokens 
 * ADMIN_JMP() overeuse alos contributes to HTML logs
 * We're missing "classic" event logging for some big events such as job assignments or round results
 * We don't currently have a way to do aggregations on said important events anyway
 
While our current usage of OpenSearch with legacy HTML-hybrid logs is great for fulltext searching about anything, it immensely falls short when you want to obtain data in a structured fashion (eg ahelp tickets text), filtering by id (eg ahelp lines by sender), or want to do aggregations (eg round results)

Just a sample impl for testing this workflow, it's incomplete, not well enough thought out (defines? port another backend instead?), and only contains a few log lines right now

### DIFF
--- a/code/__HELPERS/_time.dm
+++ b/code/__HELPERS/_time.dm
@@ -38,6 +38,10 @@ var/rollovercheck_last_timeofday = 0
 		wtime = world.time
 	return time2text(wtime - GLOB.timezoneOffset, format)
 
+/// Show the current date and time in ISO 8601 format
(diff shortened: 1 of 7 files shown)


fira:
Will reopen that when there's more to add

PR #4953, opened 2023-11-17
Hivecores perform multiple spawns at once

### LINKED ISSUE #4945: Join as Larva Delay
### Testmerges

unsure, just remembered I had this clipped

### Round ID

19554

### Description of the bug

> 8 buried larva
> click join as larva
> 3rd in queue
> wait

I assume it's partially caused by  #4502 disalowing you to directly join as larva when there's a queue
https://github.com/cmss13-devs/cmss13/assets/54422837/20a917ee-f219-41c0-8660-6596bb940b10
https://github.com/cmss13-devs/cmss13/assets/54422837/71077a52-eeff-45c3-8e60-3dbcd2a7dda4

> 8 buried larva
> click join as larva
> become larva

1. see that there are buried larvas while observing
2. join as larva
3. get put in queue??? 
4. wait.
...
not exactly sure how to reproduce it, but i'd assume you'd need to have a larva queue already so that #4502 doesn't allow observers to join.

### Issue Bingo

Drulikar:
# About the pull request

This PR simply makes it so hive cores try to dequeue as many larva as they have stored_larva each larva spawn delay rather than a single larva.

Fixes #4945 

1. Spawn as queen
2. Create core
3. Ensure hive has 2 stored_larva
4. Ghost (Have two admin clients at this point observing)
5. Ensure both clients have xeno preference set and then lose larva protections (assuming admin)
6. Both will spawn at once
7. Try again with 1 stored_larva
8. One will spawn and other has queue message updated

# Changelog
:cl: Drathek
fix: Tweaked larva queue spawning: Now spawns as many larva as possible each cycle rather than one.
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/special/pylon_core.dm
+++ b/code/modules/cm_aliens/structures/special/pylon_core.dm
@@ -272,12 +272,14 @@
 		if(spawning_larva || (last_larva_queue_time + spawn_cooldown * 4) < world.time)
 			last_larva_queue_time = world.time
 			var/list/players_with_xeno_pref = get_alien_candidates(linked_hive)
-			if(length(players_with_xeno_pref))
-				if(spawning_larva && spawn_burrowed_larva(players_with_xeno_pref[1]))
-					// We were in spawning_larva mode and successfully spawned someone
-					count_spawned = 1
-				// Update everyone's queue status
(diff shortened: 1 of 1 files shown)


fira (approved):
Looks good

PR #4341, opened 2023-09-05
Adds Custom External Webbing For IO | L-Armour To IO Vendor

Steelpoint:
# About the pull request

Adds an alternative armour to the IO, the 'customized external webbing'. Can be vended as a third armour option in lieu of the other options.

The armour is functionally identical to the service jacket in terms of defence and speed malus (that being no malus). 

The IO vendor can now vend a standard suit of light armour. 

# Explain why it's good for the game

The new external webbing offers a new way for IO's to customize their appearance. The service jacket tends to look very out of place or odd when combined with the usual IO helmet setups, the webbing may help some IO's who want the benefit of speed but want to look somewhat aesthetically fitting for the role.  

The addition of light armour to the vendor is intended to speed up IO deployment times by decreasing their need to raid a squad prep for light armour. As many IOs prefer L-Armour over the XM4 armour. 

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Screenshot 2023-09-05 23 17 19
IO wearing ext webbing.

</details>

# Changelog
:cl:
add: Intel Officers can now chose a unique 'custom external webbing' as a armour option. It is functionally identical to a service jacket, having a more unique alternative appearance that may be more fitting for some players.
add: The IO vendor now offers the option to vend a standard issue set of light armour. For the IO players who prefer it and want to save time on going to a squad prep room. 
/:cl:

PR #4963, opened 2023-11-19
Adds external webbing vest to the IO's vendor as an alternative to the service jacket

kirieee:
# About the pull request

Adds an external webbing vest to the IO vendors. 

I think that the service jacket looks a little goofy for going into a combat zone, but I want to keep the utility and speed it provides. The external webbing has the same stats, it's just a different (and, in my opinion, better) look than wearing a service jacket into battle. 

Here's some pictures of the onmob and regular sprite: 
image
image

Opened game. Webbing was in the vendor. :3

# Changelog

:cl:
add: External webbing vest added to IO vendors
/:cl:


fira:
But the webbing has no protection while the jacket does - or am I missing something here

kirieee:
I've just checked. The webbing vest has no protection at all, however the service jacket has LOW protection in melee and bullet. I was under the assumption the stats were the same, but I still don't think that this is much of a problem. I'm somewhat inexperienced when it comes to balance so I might be wrong

Steelpoint:
The asthetics of external webbing do look far more appealing than the service jacket, but the loss of armour will be a blow and make IOs even more vulnerable to damage if they take this instead of a jacket.

I did toy with the idea of adding a custom-custom external webbing that basically functioned identically to a service jacket but with the external webbing sprites as noted here: https://github.com/cmss13-devs/cmss13/pull/4341

I mostly closed it due to feeling mixed about adding another armour option but if this idea is warranted you could consider something like I did

kirieee:
I think I'll try editing the external webbing and putting a custom version in, then, just to fully smooth things over. I'm not entirely sure how to make a whole new clothing item like that but I'll figure it out, hopefully with some help

Steelpoint:
> I think I'll try editing the external webbing and putting a custom version in, then, just to fully smooth things over. I'm not entirely sure how to make a whole new clothing item like that but I'll figure it out, hopefully with some help

You can look over how I did it, the PR itself was fine I just ran into unrelated merge conflicts and a disinterest to finish the PR.

Huffie56:
if your item isn't OP i would suggest you to add it to the clothing surplus vendor instead of just IO vendor...

kirieee:
Adding it to the clothing vendors would essentially be the same as adding a service jacket to the vendors, too. I'm not opposed! But I think that some maintainers would be?

Huffie56:
> Adding it to the clothing vendors would essentially be the same as adding a service jacket to the vendors, too. I'm not opposed! But I think that some maintainers would be?

as long as it's not OP and in small number i don't see a reason to not add them but yes maintainer have to approve it.
maybe just ask them on discord in dev channel?

PR #4964, opened 2023-11-19
Fix nullspaced closet runtime

Zonespace27:
# About the pull request
image

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets.dm
+++ b/code/game/objects/structures/crates_lockers/closets.dm
@@ -159,8 +159,9 @@
 /obj/structure/closet/proc/take_damage(damage)
 	health = max(health - damage, 0)
 	if(health <= 0)
-		for(var/atom/movable/A as anything in src)
-			A.forceMove(src.loc)
+		if(loc)
+			for(var/atom/movable/movable as anything in src)
+				movable.forceMove(loc)
 		playsound(loc, 'sound/effects/meteorimpact.ogg', 25, 1)
 		qdel(src)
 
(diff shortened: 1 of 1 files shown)


fira:
Wait, what? This doesn't make sense. How does the closet take damage from the proejctile if it's in nullspace?

fira:
Ohhhhhhhh I think I get it

Shrapnel flies -> scan_a_turf -> Hits the closet -> Closet takes damage and falls below 0 HP -> Closet forceMoves the contents out of it -> Contents movement triggers Crossed with the shrapnel again and scan_a_turf causing the bullet to hit the closet again -> recurse until every contents are forceMoved off, one per frame -> last frame deletes the closet -> previous frame keeps going with copied `as anything in src` list -> loc is null, runtime -> keep unstacking frames

I think that's it because looking at dd.log there are actually 22 runtimes, as it unwinds the stack and fails to move them at every recursion level

It's not directly the bullet code's fault because the closet isn't yet deleted when hit checks happen - but the legacy behavior to rescan the whole turf when Crossed is really dumb. The closet's behavior shouldn't make it recurse bullet hits onto itself

fira:
But yeah if that's it this change won't fix it, you'd need to check `loc` INSIDE the for loop

Zonespace27:
> But yeah if that's it this change won't fix it, you'd need to check `loc` (or QDELETED) INSIDE the for loop
> 
> Edit: An easy way out could be to early return from take_damage if the health is already <0 but i don't know how sane that is

did both

fira (approved):
🤯

PR #4966, opened 2023-11-19
Fix human explosion runtime

Zonespace27:
# About the pull request
image

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -193,8 +193,9 @@
 
 	//Focus half the blast on one organ
 	var/mob/attack_source = last_damage_data?.resolve_mob()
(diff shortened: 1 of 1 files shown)


fira:
I don't think this is the correct way to fix this. There are several other places that also rely on `pick(limbs)` in chemical code that would have the same issue, and limbs should never be null, the limbs include chest

Presumably the only way this would end up with nulled/empty limbs is if the humans was deleted to begin with

Zonespace27:
> I don't think this is the correct way to fix this. There are several other places that also rely on `pick(limbs)` in chemical code that would have the same issue, and limbs should never be null, the limbs include chest
> 
> Presumably the only way this would end up with nulled/empty limbs is if the humans was deleted to begin with

Gah, forgot that the chest was a limb. Will look into further, this isn't the right fix.

PR #4970, opened 2023-11-19
Fix plushie/fountain pen/family photo pickup runtime

Zonespace27:
# About the pull request

image

I'm not 100% on why this happens. My best guess is it's the usual "clients don't follow the rules" stuff that lets them attempt to take the plushie out of their satchel before they're fully spawned in and started.


fira:
That one is my fault

fira (approved):

fira (approved):

PR #4973, opened 2023-11-20
Fix null limb owner runtime.

Zonespace27:
# About the pull request
image

The rest of the code seems to always check for owner, so I suppose having an owner isn't guaranteed?


fira:
`owner` should always be set i think? i had a similar problem not long ago further down in `get_limb_icon_key` where `species` was unexpectedly null. the fix was #4805 - the limbs were not deleted and not actually cleared

just like #4966 i think it usually means the whole damned thing is deleted and improperly used to begin with

PR #4976, opened 2023-11-20
More lifeboat launch options

zzzmike:
# About the pull request

The default option when launching lifeboats will now have ARES make an announcement, followed by a 10 second delay before doors close. The current functionality where it instantly launches is also preserved as the Emergency Launch option.

It's sad that currently someone can't make an announcement that they are about to launch a lifeboat if they don't have the command tablet. This gives them the option.

<details>

I did test it, triggered hijack and launched lifeboat using the new option. It worked on local server.

</details>

# Changelog
:cl: zzzmike, drathek, ihatethisengine2
add: Lifeboat launch now has an ARES announcement followed by a 10 second delay before doors close. The current launch functionality is preserved as Emergency Launch.
/:cl:


code/modules/shuttle/computer.dm
```diff
@@ -305,9 +305,16 @@
...
-				if(tgui_alert(user, "Early launch the lifeboat?", "Confirm", list("Yes", "No"), 10 SECONDS) == "Yes")
-					to_chat(user, SPAN_NOTICE("[src]'s screen blinks and says \"Early launch accepted\"."))
-					lifeboat.evac_launch()
+				var/response = tgui_alert(user, "Early launch the lifeboat?", "Confirm", list("Yes", "No", "Instantly"), 10 SECONDS)
+				switch(response)
+					if("Yes")
+						to_chat(user, "[src]'s screen blinks and says \"Early launch accepted\".")
+						shipwide_ai_announcement("Launch command received. " + (lifeboat.id == MOBILE_SHUTTLE_LIFEBOAT_PORT ? "Port" : "Starboard") + " lifeboat doors will close in 10 seconds.")
+						addtimer(CALLBACK(lifeboat, TYPE_PROC_REF(/obj/docking_port/mobile/crashable/lifeboat, evac_launch)), 10 SECONDS)
+					else if("Instantly")
```
ihatethisengine, line 314:
```suggestion
					if("Instantly")
```
It's a switch, else is redundant (and probably should break the code).

zzzmike: yeah I ended up adding that to fix an error and then it worked locally but not with linter... anyways thanks, resolved

code/modules/shuttle/computer.dm
```diff
@@ -305,10 +305,17 @@
...
-				if(tgui_alert(user, "Early launch the lifeboat?", "Confirm", list("Yes", "No"), 10 SECONDS) == "Yes")
-					to_chat(user, SPAN_NOTICE("[src]'s screen blinks and says \"Early launch accepted\"."))
-					lifeboat.evac_launch()
-					return
+				var/response = tgui_alert(user, "Launch the lifeboat?", "Confirm", list("Yes", "No", "Emergency Launch"), 10 SECONDS)
+				switch (response)
+					if ("Yes")
+						to_chat(user, "[src]'s screen blinks and says \"Launch command accepted\".")
+						shipwide_ai_announcement("Launch command received. " + (lifeboat.id == MOBILE_SHUTTLE_LIFEBOAT_PORT ? "Port" : "Starboard") + " Lifeboat doors will close in 10 seconds.")
+						addtimer(CALLBACK(lifeboat, TYPE_PROC_REF(/obj/docking_port/mobile/crashable/lifeboat, evac_launch)), 10 SECONDS)
```
fira, line 313: missing return here no?

code/modules/shuttle/computer.dm
```diff
@@ -305,10 +305,17 @@
 					to_chat(user, SPAN_NOTICE("[src]'s screen says \"Unable to launch, fuel insufficient\"."))
 					return
 
-				if(tgui_alert(user, "Early launch the lifeboat?", "Confirm", list("Yes", "No"), 10 SECONDS) == "Yes")
-					to_chat(user, SPAN_NOTICE("[src]'s screen blinks and says \"Early launch accepted\"."))
-					lifeboat.evac_launch()
-					return
+				var/response = tgui_alert(user, "Launch the lifeboat?", "Confirm", list("Yes", "No", "Emergency Launch"), 10 SECONDS)
```
fira, line 308: need something to prevent spam launching. i don't know if this has undesirable effects on shuttle side (though nothing really prevents it) but at least the announcements would get duplicated

fira (changes requested):

ihatethisengine:
I already mentioned this in the contrib channel, but will say it again: I think we should keep it simple, two buttons for the same action does sound like overkill, better just remove "instant" launch (why do you really need it, besides cheesy abductions of queen), while keeping regular launch with the announcement and the timer.

zzzmike:
> I already mentioned this in the contrib channel, but will say it again: I think we should keep it simple, two buttons for the same action does sound like overkill, better just remove "instant" launch (why do you really need it, besides cheesy abductions of queen), while keeping regular launch with the announcement and the timer.
> 
> This is my opinion, I am not a dev.

I understand your side of it for sure, I just don't want to force them to have a delay in case of a situation where a 10 second delay would mean everyone dies. But yes, ideally it would be used very rarely. One situation might be when people don't know about this change yet - which meant they ended up in a situation where they expect and need the current "instant" functionality on the live environment. Could look at revisiting it down the road for sure.

code/modules/shuttle/computer.dm
zzzmike: thanks, fixed

fira (changes requested):

code/__DEFINES/shuttles.dm
```diff
@@ -1,6 +1,7 @@
 //shuttle mode defines
 #define SHUTTLE_IDLE    "idle"
 #define SHUTTLE_IGNITING   "igniting"
+#define SHUTTLE_IGNITINGSOON "ignitingsoon"
```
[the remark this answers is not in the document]
zzzmike: thanks, fixed

zzzmike:
> I'm unsure what the impact will be of the access changes here but I don't think they're hugely important since this is at the end of the round anyway.

Sorry @Birdtalon I should've probably posted my commit message as a comment here to make it more clear. Basically, CC and CL have marine_command in their access, and I believe it's only so that it's more fun for them OOCly to be able to walk in to CIC and RP with people without having to ask to be let in. And I REALLY don't think it's intended for them to be able to launch lifeboats and potentially leave people behind. Anyways, commit message below:

So, I searched up the file history and saw that the last person to change this - morrow - initially had this restricted to access_marine_senior and then went to access_marine_command. I spoke to him about it and he said "Senior command ID was too restrictive, command ID is probably not restrictive enough, I meant to change it again but uuuh it wasn't my PR anymore and I quit." After looking at who this would affect, I think "marine_senior or marine_dropship" is a good compromise.

zzzmike:
@fira @BeagleGaming1 if you're able to re-review so it doesn't have Changes Requested slapped on it forever I would appreciate it, thanks ❤️

PR #4982, opened 2023-11-20
Research fire chemicals buff

SubjectD9341:
Stronger custom fire chems
# About the pull request

This PR is to buff the ability of research in making fire chems.

Currently. Fire chems are ridiculously expensive and inferior in all ways compared to your standard MST/NST meta stim.
For example, if you were to make a fueling10, oxi10, flowing5 (required to make the chemical work with a flamethrower). You would create a chemical that's but barely stronger than regular naphtalm.

With this PR, my hope is to bring this very neat ability of research into something more viable in the current state of the game. And hopefully, push current players in doing something else than always the same tired MST/NST stims

It allows research to produce viable flamer fuels more easily than before

<details>

image

 
Here is pictured an oxidizing 10, fueling 10 chemical with the changes to the oxidizing property. The fuel is of course, much stronger but still allows certain castes to pat themselves out alone and survive it.
</details>

# Changelog
:cl:
balance: made the oxidizing property stronger.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry_properties/prop_positive.dm
+++ b/code/modules/reagents/chemistry_properties/prop_positive.dm
@@ -756,7 +756,7 @@
 	description = "The chemical is oxidizing, increasing the intensity of chemical fires. However, the fuel is also burned slightly faster because of it."
 	rarity = PROPERTY_COMMON
 	value = 1
-	intensity_per_level = 6
+	intensity_per_level = 10
 	duration_per_level = -2
 
 	intensitymod_per_level = 0.2
--- a/code/modules/reagents/chemistry_reagents/other.dm
+++ b/code/modules/reagents/chemistry_reagents/other.dm
@@ -494,7 +494,7 @@
 	falloff_modifier = -0.1
 	burncolor = "#ff9900"
 	chemclass = CHEM_CLASS_RARE
-	properties = list(PROPERTY_FUELING = 5, PROPERTY_OXIDIZING = 3, PROPERTY_VISCOUS = 4, PROPERTY_TOXIC = 1)
+	properties = list(PROPERTY_FUELING = 5, PROPERTY_OXIDIZING = 2, PROPERTY_VISCOUS = 4, PROPERTY_TOXIC = 1)
 
 /datum/reagent/space_cleaner
 	name = "Space cleaner"


RenaRenaRe:
Just so you know, this would affect base chemicals too like oxygen and ethanol.
Most notably Welding Fuel which currently is the only base chem with decent flamer fuel stats because for some strange reason it has hard coded stats combined with the stats from properties. So I believe with this PR welding fuel would go from 18 intensity, 34 duration, 4 range to 30 intensity, 34 duration, 4 range. This is better than UT-napthal. Possible solution to this though is to remove the weird hard coded stats, not sure why it even has them in the first place since no other chem has it.

Other noteworthy effect on base chems is Napalm would have 45 intensity, 29 duration, 1 range, which is better than UT-Napthal at the cost of having to increase the fuel pressure for decent range. CLF3 would have 90 intensity with no duration and range which is pretty funny.

Proper flamer chems like UT-napthal don't use Oxidising property so this shouldn't affect them luckily.

SubjectD9341:
> Just so you know, this would affect base chemicals too like oxygen and ethanol. Most notably Welding Fuel which currently is the only base chem with decent flamer fuel stats because for some strange reason it has hard coded stats combined with the stats from properties. So I believe with this PR welding fuel would go from 18 intensity, 34 duration, 4 range to 30 intensity, 34 duration, 4 range. This is better than UT-napthal. Possible solution to this though is to remove the weird hard coded stats, not sure why it even has them in the first place since no other chem has it.
> 
> Other noteworthy effect on base chems is Napalm would have 45 intensity, 29 duration, 1 range, which is better than UT-Napthal at the cost of having to increase the fuel pressure for decent range. CLF3 would have 90 intensity with no duration and range which is pretty funny.
> 
> Proper flamer chems like UT-napthal don't use Oxidising property so this shouldn't affect them luckily.


I have checked just now the CLF3 and napalm, did some tests with them ingame to check how they functioned, thankfully, they don't work properly for flamethrower nor inside spray bottles (the famous bootleg flamethrower).
If you are worried that napalm/CLF3 can just be modified by research to make it flamethrower viable, fret not, it's not something they are able to achieve (you cannot relate new properties on a base chem)
 
You were entirely correct concerning the welding fuel, therefore, I reduced the oxidizing property from level 3 to level 2, now welding fuel should only have an intensity of 2, keeping it weaker than base napthal.

I also checked OT but discerned no differences

SarahJohnson489:
Isn't it a problem with the regular custom tank itself capping values preventing you from overcompensating with higher fuel pressure with a cheaper chem and being physically unable to do it at all without a metal obsessed OT existing at all to vend tanks on the researcher side? Arguably the recipe difficulty for the quantities you want to pump out if you're doing it alone if you're fuel pressure coping. Arguably having poor testing feedback and not being able to test it yourself at all if the req flamer kits are all gone after funny man nuked their kits from pfc vendors?  Flamer kit accessibility being nuked by funny man did more damage than anything else cause there's only 0-3 left by the time you get the chem done. Getting Relate bingo and never finding fire penetration in a round long enough for it and having a live pyro is just the sad consequence of the RNG system.

A problem with marines not direct flaming then multiple marines chasing so it can't go roll and pat contributing to the perception of poor performance despite now living in a post-nozzle society?

Sort of a problem with the difficulty of distributing it to the marines who actually have flamers since there's no real effective way to to reach them if they don't respond to you before first drop? Which also excludes late joins cause we don't have a message board by the alamo which marines wouldn't read the notes on it even if we did.
Then there's the general lack of knowledge of intel by CIC, riflemen and riflemen unwilling to take initiative to escort intel gang resulting in few to no grants, chems, papers, or bodies. Small hit from no PO lowpop, PO's not using fultons, and shipside not delivering or notifying bodies. Like, it's doable people are just leaving research hanging. Short rounds and little to no additional credits is hard when there's 5 people jockeying for one basket of points.
Which then ties into the fundamental flaw of  3 properties for one actual effect chem punishing you by +7-14 or more credits, not getting credits for scanning chems not on the list, and the pricing/credit system in general. Four if you're leaving something to slap fire penetration onto or having to find a 4+ chem paper that isn't godawful to fix up.

Other than the fact that you have to find it, anti-parasitic, bonemend, hypometabolic, death timer, explosive, or defib chems are all good at keeping people in the round and contributing just underutilized and difficult to get into the hands of those that need it in significant quantities with the knowledge to use it properly and the courage to actually administer it. Electrogenetic lets you make a straight better epi. Stronger ACR/NGN alongside keloderm and merabic so anyone that managed to live is quickly healed to full and anyone that died is quickly back to full preventing the common push fatigue wipes. All increasingly compounding as the round lets you get more and more of them done, mass produced, and actually in their hands not ignored on the fob or supply drop floor. There's good stuff to be made and used the hangup is in the literal terrorists raising clearance too high w/o cause, difficulty getting the other "endgame" chem(s), credit system and teamwork req bottlenecking it like other strong but agonizing jobs for which there's no real fix that wouldn't get the sad state of research beaten down even harder.

Research flame chems are turbowrecked by caps on the custom flamer tank and the flame intensity softcap through transitioning the flameshape from the diamond pattern to the star pattern on other ordnance which is only fair cause it would get out of hand if they weren't kept shit. Wouldn't even last a week before it got beaten with the nerf bat into a sadder state than before.

Tanks work with less than 10/10/5 It just sucks more, higher tank psi, takes longer via waiting for credits to improve parameters, or is a larger pain to make if you fail the recipe rng. A little wait makes that not matter by just making it easy enough to mass produce so they can just take extra tanks to cycle through.
The credit penalty at level 6-7 isn't even that bad and you don't have to . The problem with the system comes from everything else whether it's the RNG or the credit penalties and the quantities used for a bad recipe bogging it down. 
Works I guess


P.S. 1 hour rounds suck. All my homies hate <1hr evac/victory speedruns.

SarahJohnson489:
Don't know why I expect better on a project that affects everyone playing the game.

Zonespace27:
Thank you for making this PR. Unfortunately, it will be closed.

Research is in an incredibly poor spot as it stands. The entire job largely depends on RNG to get interesting things made, which is not fun or engaging. Furthermore, the heavy reliance on RNG makes it incredibly difficult to balance, as there's no guarantee for when you might get any given property. Research needs a full rework to have a more consistent vision that depends less on RNG.

Until a rework of research is proposed and made, I will not be accepting any new features or balance changes for research.

PR #4990, opened 2023-11-21
Fix resin doors not closing on mob/living if they are merged with weeds

Drulikar:
# About the pull request

This PR simply changes the mob/living check in `/obj/structure/mineral_door/resin/Close()` to also account for whether they are merged with the weeds. As soon as xeno corpses are also merging with weeds this will handle them too.

Although I think this is a very rare situation, it can be especially confusing if a fully weeded corpse that intentionally is hard to see is causing a door to not close (seemingly arbitrarily).

<details>

https://github.com/cmss13-devs/cmss13/assets/76988376/4770ff65-c484-4e88-9381-19dfad22ffae

</details>

# Changelog
:cl: Drathek
fix: Resin doors will now close on dead mobs that are merged with weeds (currently only human).
fix: Resin doors will now restart their closing timer each open making the delay to close consistent.
code: Added a TRAIT_MERGED_WITH_WEEDS that is set whenever the mob is currently merged with weeds.
/:cl:

### DIFF
--- a/code/modules/cm_aliens/XenoStructures.dm
+++ b/code/modules/cm_aliens/XenoStructures.dm
@@ -422,14 +422,22 @@
 		if(!isSwitchingStates && state == 1)
 			Close()
 
+/obj/structure/mineral_door/resin/proc/close_blocked()
+	for(var/turf/turf in locs)
+		for(var/mob/living/living_mob in turf)
+			var/datum/component/weed_food/weed = living_mob.GetComponent(/datum/component/weed_food)
+			if(!weed?.merged)
+				return TRUE
+	return FALSE
+
 /obj/structure/mineral_door/resin/Close()
 	if(!state || !loc) return //already closed
 	//Can't close if someone is blocking it
-	for(var/turf/turf in locs)
-		if(locate(/mob/living) in turf)
-			spawn (close_delay)
-				Close()
-			return
+	if(close_blocked())
+		spawn(close_delay)
+			Close()
+		return
+
 	isSwitchingStates = 1
 	playsound(loc, "alien_resin_move", 25)
 	flick("[mineralType]closing",src)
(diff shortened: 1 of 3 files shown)


code/modules/cm_aliens/XenoStructures.dm
```diff
@@ -422,14 +422,22 @@
 		if(!isSwitchingStates && state == 1)
 			Close()
 
+/obj/structure/mineral_door/resin/proc/close_blocked()
+	for(var/turf/turf in locs)
+		for(var/mob/living/living_mob in turf)
+			var/datum/component/weed_food/weed = living_mob.GetComponent(/datum/component/weed_food)
```
fira, line 428: You're not supposed to use GetComponent at all, it defeats the whole point of components, use a signal

code/modules/cm_aliens/XenoStructures.dm
```diff
@@ -422,14 +422,22 @@
...
 	if(!state || !loc) return //already closed
 	//Can't close if someone is blocking it
-	for(var/turf/turf in locs)
-		if(locate(/mob/living) in turf)
-			spawn (close_delay)
-				Close()
-			return
+	if(close_blocked())
+		spawn(close_delay)
+			Close()
```
fira, line 438: Infinite recursion with a spawn on top is kind of ehhh, should replace this with a timer while modifying there

fira (changes requested):

code/modules/cm_aliens/XenoStructures.dm
Drulikar: If so, then other uses of GetComponent should be refactored and the comments for GetComponent should reflect this. Regardless simple enough to convert over to the style of signal handling that is frowned upon.

fira: Well it's in the ECS documentation

image

You could use a new trait but leveraging the component sounds good to me, it's just the whole point of them is to delegate the behavior to it through signal/proc rather than acting upon it conditionally in the calling code

fira (approved):

PR #4768, opened 2023-10-22
LV624 Fog Removal + River Change

Steelpoint:
# About the pull request

Several major but interconnected changes have been made to 624.

All fog on the map is removed, meaning the entire map is accessible from the start.

The river flowing through the centre of the map inflicts significant slowdown to any mob that tries to move through it, making it extremely dangerous to traverse it while enemies are in range.

Three new bridges have been placed around the map (one by LZ2 comms, one north of hydro, one by eastern river), the bridges are guaranteed to be present and are fully intact, offering a traversable area to move through.

The areas around these bridges have been expanded for xeno weeding.

Nightmare inserts regarding fog or river bridges have been removed from the map.

I've removed the floodlights in tfort and replaced them with the breakable flood lights, so xenos can kill the lights instead of them being on forever

# Explain why it's good for the game

The Fog is a ancient 2016 concept that we have long passed needing.

My changes are aimed at shaking up the meta of a map that has stagnated for a decade. The addition of three bridges, plus the massive slowdown of the river, will crate artificial chock points that the Marines and Xenos will want to fight over to establish control over the enemies territory. While the river can still be crossed, it represents a death sentence if caught by a enemy. 

Fruthermore the extreme slowdown will encourage both sides to take and hold a bridge.

The expanded weeding grounds of the bridges will allow xenos to establish better river defences and even allow them to confront the marines in the south river area on first contact. 

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Screenshot 2023-10-22 18 00 59

Screenshot 2023-10-22 18 01 07

Screenshot 2023-10-22 18 01 26

</details>

# Changelog
:cl:
maptweak: Fog on 624 has been removed
maptweak: River on 624 now inflicts a massive movement slowdown on anyone cross it
maptweak: Three bridges have been placed around 624 allowing easy access across the river
maptweak: Expanded the areas xenos can weed around the river to allow better defences. 
maptweak: Flood lights around t-fort are now breakable by xenos
/:cl:

PR #4997, opened 2023-11-22
LV624 Fog Removal + River Change V2

Steelpoint:
# About the pull request

Continuation of this prior PR, but without the increased slowdown on the river or removal of the floodlight at T-fort: https://github.com/cmss13-devs/cmss13/pull/4768

----

All fog on the map is removed, meaning the entire map is accessible from the start.

Three new bridges have been placed around the map (one by LZ2 comms, one north of hydro, one by eastern river), the bridges are guaranteed to be present and are fully intact, offering a traversable area to move through.

The areas around these bridges have been expanded for xeno weeding.

Nightmare inserts regarding fog or river bridges have been removed from the map.

The Fog is a ancient 2016 concept that we have long passed needing. 

The expanded weeding grounds of the bridges will allow xenos to establish better river defences and even allow them to confront the marines in the south river area on first contact. 

<details>

Screenshot 2023-10-22 18 00 59

Screenshot 2023-10-22 18 01 07

Screenshot 2023-10-22 18 01 26

</details>

# Changelog
:cl:
maptweak: Fog on 624 has been removed
maptweak: Three bridges have been placed around 624 allowing easy access across the river
maptweak: Expanded the areas xenos can weed around the river to allow better defences. 
/:cl:


irakligabunia:
The Fog is a ancient 2016 concept that we have long passed needing.
Explain why it's good for the game
I don't get how that's an explanation, why does it need to go?

Steelpoint:
When 624 was first released, it featured no fog at all. The entire map was accessible from the very start. However, the administration of the day had a rule in place stating that Marines could not attack the Xeno positions across the river prior to a certain time elapsing, however it was difficult to enforce this rule as both sides were always itching to start a fight.

Ergo, a code solution to an admin problem was implemented, the river fog, making it impossible for Marines to attack across the river.

Over the many years between 2016 and 2023, the game has changed substantially, in-between restricting marine deployments until 20 minutes has passed, new mechanics, tunnels, evo changes, etc, etc. 

River fog is a old administration block that has long since run its need, we no longer have a rule against "meta rushes" across the river, ergo the fog is no longer needed.

----

Furthermore I have two further points to support its removal.

Firstly, the current iteration of 624 results in situations where both sides, xenos and marines, are sitting at the fog waiting for it to drop so both sides can actually start fighting. We have a situation of Marines needing to wait 20 mins to drop, then you need to wait 10+ minutes for the fog to drop. This is egregious. 

Secondly, no other map ever released after 624 features fog blockages in any capacity outside of Chances and Soro, which only use Fog to stop survivors camping in the landing zones. No other map sans 624 uses fog to block marines or xenos from fighting. 

----

I should further note that this removal will open up the opportunity for xenos to actually contest the colony on first drop. As it stands, the current meta is for xenos to build up at beach. Now, thanks to my proposed change, xenos will not only be able to set up some fortification on the south river, but they can expand and try to defend/skirmish south of the river as they will now be able to retreat north at any time.

99.9% of the time, xenos never contest south colony until post-fog. This PR makes this a viable strategy.

Steelpoint:
Done

maps/Nightmare/maps/LV624/scenario.json
```diff
@@ -1,19 +1,9 @@
...
-		{ "type": "def", "values": { "mainpath": "right"  } }
-	]
-},
 {
 	"type": "pick", "name": "LV624 Global Event",
 	"choices": [
 		{ "weight": 2, "type": "def", "values": { "lvevent": "none"       } },
-		{ "weight": 4, "type": "def", "values": { "lvevent": "last_stand" } },
-		{ "weight": 2, "type": "def", "values": { "lvevent": "fallen_ship", "mainpath": "bridge"  } },
-		{ "weight": 2, "type": "def", "values": { "lvevent": "fallen_ship", "mainpath": "right" } }
```
fira, line 16: wait no don't remove these, roll only the lvevent, we still need fallen_ship to be an option

```
{ "weight": 4, "type": "def", "values": { "lvevent": "fallen_ship" } },
```

everything looks good otherwise 👍

Steelpoint:
Why I should not do changes at midnight. Changes made

Nanu308:
Need this tested some more, and personally not convinced on these changes as it stands after playing when the last PR was in TM.

fira (approved):
Code/Defs good

Nanu308:
I've played it now a bit, spoken to some people as well to hear about how the rounds flow from the ones i could not participate in and I'm gonne say i oppose this change entirely.

My reasons are as follows:
With the fog removed the center of the colony sees a lot less attention from both sides as there is not a good reason to go there anymore when you have direct access to the beaches, which means we get a lot more unused map area. 

For xenos, they need to cover more grounds to setup defenses, or, they setup the hive in hydro or heavily fortify it so then the center of the map gets a bit more use, but it leads to potentially very short rounds if either side looses a lot of forces.

Besides that, since there is no good time to setup defenses for the marines and the map is rather small from where xenos normally face off the marines in this TM run the risk of being rushed down entirely if the front collapses, shortening round length a lot when it occurs.

I'll be leaving the DNM label on it for now, i don't think this is a good change for LV.

Steelpoint:
Perhaps an alternative would be to better funnel both sides in via removing some of the eastern and western river entrances so there's more of an incentive for both sides to push the centre of the colony. 

The issue of xenos setting up in hydro does not strike me as a major issue. Firstly, xenos were always able to do this (they just never did since they'd have no escape route), secondly this is a 'all in' strategy since if the xenos get pushed out they are likely to be routed. I witnessed this occur at least two times where the xenos suffered mass casualties into a defeat since their main hive was too close to the marines and it was easier for marines to concentrate themselves since they were so close to the FOB. 

I still personally think the fog is an egregious concept that no other map needs to make the map function, and I'd like to get it removed, even if I have to make further changes to make it happens. 624 never shipped with the fog to start with, and I still think its worth returning to that state.

PR #5011, opened 2023-11-22
Système d'énergie de commandement - 1

LC4492:
# About the pull request
This PR is focused on setting the CIC Power Substation on where the Synthetic Closet originally is, and move that same closet to where the station once was, improving mobility by reducing the amount of time the synthetics spend on walking to their closet and also improving protection to the power station by placing it in the secure hall of CIC.

If the explanation above is good enough, then I suppose its good for the game.

image
New power substation
image
New synthetic closet

</details>

# Changelog
:cl:
maptweak: Sets the CIC power station to where the synthetic closet is currently located, and moves the synthetic closet to where the power station is currently located. Improves APC positioning, adds more lights to specific areas where there should be more, fixes stuff that were blocking the view of signs and making them ugly.
/:cl:


fira:
Are the changes in Cryo intentional?

LC4492:
> Are the changes in Cryo intentional?

Yes. When I made the substation side, I needed to change the position of some airlocks to everything fit sweetly, that's when I stopped and wanted to do some little changes, like just adding some extra window and positioning the chambers better.

Nanu308 (changes requested):
Brig has this because people were memeing and causing issues with the APC's, i do not think it should be expanded to other parts of the ship. I think we should strive to keep it like in Alien Isolation where APC's are in almost every single case located in the rooms themselves and not bunched up together.

Moving the Synth closet is fine, and moving the SMES storage in is also fine, but the room that used to house the synth spot looks really boxy now so I'd do some tweaks to make it look better.

There are some good map fixes tho besides this. 👍

cm13-github:

PR #5038, opened 2023-11-25
ASO Reqs Access Removal

fira:
# About the pull request

Removes Reqs Machine access from ASO. They are still free to waltz about Reqs due to Command access, or add the access if needed.

ASO's job is, to quote the Game's Entry text, `to oversee the hangar crew, the intel officers, the engineering department, and requisition department.`

Three of these departments do not require active oversight soooooo you very frequently get into the situation where Reqs is manned by 1 ASO + 1 QM + CTs (up to 3).

You might know that a proactive minmaxing QM will handle the majority of the work alone. The same now applies to ASO: often the ASO drives the QM out of a job as super-supervisor.

This is a **light attempt** at reminding the ASO that it is an oversight role. Hope is also that without being able to push the buttons itself, ASO will be railroaded into engaging in interaction and RP with rest of the crew. Much like with SOs, it can still give itself Reqs access if there is a shortage of personnel. 

image

# Changelog
:cl:
del: The ASO, as an overseeing role, now starts without Reqs Access. It can still enter Req Bay, and grant itself access if there is a lack of personel. 
/:cl:


BadAtThisGame302:
You said it yourself that he's supposed to oversee Req and he doesn't even get full access to a department he oversees? What if there's no CTs or QM and someone has to man the Req line, do they have to give the guy who's allegedly the boss access to his own department? 

Also another point that you made is that the ASO drives the QMs out of a job, which is the exact opposite from what I've seen, because people rarely play ASO to run Req out of the three departments they oversee, I usually see ASOs run Intel Overwatch and focus on Intel rather than Req...

fira:
Like I said I think 3 times they can just give themselves access if needed. This is just slight incentive to not overstep crate work as one of the ship's top officers.

fira:
Whatever, this doesn't address the real problem anyway

PR #5050, opened 2023-11-27
Fixes Kutjevo space tiles

### LINKED ISSUE #5013: Kutjevo Refinery space tiles under wall
### Testmerges

#4866, #4934, #4960, #4986, #5007, #4828

### Round ID

19637

### Description of the bug

Its space tile when there should be floor.
cmbug3974

There should be floor.

1. Go to the spot
2. Destroy walls
3. See space instead of floor
...

### Issue Bingo

SabreML:
# About the pull request

Fixes #5013 by replacing a couple of space tiles under a window with plating.
(They were in a 'Nightmare' submap, so I didn't need to touch the main map file.)

I also checked the other maps for any more hidden space tiles, and found a few out of bounds in Whiskey Outpost, so I figured I'd add those too.

They don't actually seem to do anything bad, but either way spess tiles shouldn't be visible on a ground map.
<details>
<summary>Screenshots</summary>

**Before:**
scrnshot1

**After:**
scrnshot2

</details>

# Changelog
:cl:
fix: Fixed a few space tiles under a window in Kutjevo Refinery.
/:cl:

### DIFF
--- a/maps/map_files/Kutjevo/sprinkles/35.communications.dmm
+++ b/maps/map_files/Kutjevo/sprinkles/35.communications.dmm
@@ -108,10 +108,6 @@
 	dir = 1
 	},
 /area/template_noop)
-"qI" = (
-/obj/structure/window/framed/kutjevo,
-/turf/open/space/basic,
-/area/template_noop)
 "qM" = (
 /obj/structure/machinery/light,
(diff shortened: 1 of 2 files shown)


Nanu308 (approved):
lgtm

PR #5051, opened 2023-11-27
Kutjevo Tunnels Upgrade

HeresKozmos:
# About the pull request

As part of my ongoing series of PRs to update tunnel placements, especially on older maps, Kutjevo has had its original five tunnels replaced with 7 new tunnels with more of a focus on the center and right sides of the map.

Kutjevo's tunnels I find are mediocre at best, to start with there's a very limited quantity of five tunnels, which was like Sorokyne until recently, and of which three of them I believed to be mediocre at best and the other two to be not particularly useful most times.

The scrubland tunnel is out in the open with no natural cover and it is rather close to the LZ, this attempts to move it further backwards and make it a little more out of the way. 

The stony fields tunnel is not terrible but it suffers from a lack of use since there was already a better option between botany and medical and from personal experience I find the majority of landings are at the north LZ rather than the south LZ. Since it's so out of the way it can become a free shortcut to the comms tower in the late game should it spawn at Hydroelectric.

The afforementioned botany-medical tunnel isn't bad, it's just plainly in the open and it doesn't last particularly long once a foothold has been established in medical or botany. The two existing tunnels inside the caves are mediocre at best and from my personal experience any normal hive tends to be just far enough away from them to make an in-Hive tunnel a requirement.

I've opted to give both medical and botany their own "dedicated" tunnels in the medical park and the NE corner of botany. I have moved the scrubland tunnel further east behind some cover. The other tunnels provide better access to the caves and the eastern construction area in general.

(The lines are just visual aids to make finding the tunnel spots easier, they are not indicative of other functions*)

Kutjevo's Existing Tunnels:
Kutjevo Tunnel Map V2 0

Kutjevo's New Tunnels:
Kutjevo Tunnel Map V3 0

# Changelog

:cl:
mappadd: Added 7 new tunnels to Kutjevo
maptweak: Removed the 5 original tunnels on Kutjevo
/:cl:

### DIFF
--- a/maps/map_files/Kutjevo/Kutjevo.dmm
+++ b/maps/map_files/Kutjevo/Kutjevo.dmm
@@ -563,6 +563,10 @@
 	},
 /turf/open/auto_turf/sand/layer0,
 /area/kutjevo/exterior/scrubland)
+"aKg" = (
+/obj/structure/tunnel,
+/turf/open/auto_turf/sand/layer0,
+/area/kutjevo/interior/colony_South)
 "aKl" = (
 /obj/structure/machinery/landinglight/ds1{
 	dir = 1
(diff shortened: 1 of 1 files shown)


Nanu308:
Need @Zonespace27 approval, will go over after that

Nanu308 (approved):
lgtm!

PR #4716, opened 2023-10-18
Eggsac carrier revival

Birdtalon:
[Forum Thread](https://forum.cm-ss13.com/t/eggsac-carrier-revival/4711)

# About the pull request

The concept of this PR is to find a middle ground between the current eggsac carrier and the pre #4459 eggsac carrier.

This PR will make the following changes. (From this point "normal weeds" can be substituted for "off hive weeds" Placing eggs on hive weeds is **unchanged**.)

- Eggsac carrier can once again place eggs on normal weeds.
- Eggsac carrier can only place 4 eggs at once on normal weeds.
- If the Eggsac places more than 4 eggs on normal weeds, their oldest placed egg will die. No hugger release.
- Eggs placed on normal weeds have a maximum lifetime of 5 MINUTES after which they will die. No hugger release.
- Eggs placed on normal weeds have a 1 MINUTE life whilst the carrier is further than 14 tiles away from them.
- If the carrier dies, all of their sustained eggs die.

# Explain why it's good for the game

Eggsac carrier at the moment is in a bit of a poor place and has gone from being very strong to quite poor. Considering the limitation of having to place only on hive weeds.
The majority of feedback I read against eggsac carrier was with the quantity of eggs able to be placed, as well as the locations they can be placed in, all across the map and with impunity.

This PR aims to address those concerns to make the eggsac both less infuriating to play against while still being satisfying to play as a frontline support or as a stealthy trapper.

Eggs can no longer be placed all over the map because of the 4 egg limit off weeds, so the carrier has to think where they want to impact the map. The carrier also has to stay within a reasonable distance to where they are impacting with their eggs which localises their impact to their immediate play area. The carrier also has to be more reactive to current events as they cannot place an egg which then becomes useful 30 minutes later.

Killing the carrier also has a small reward as in addition to removing a xeno from the game, the eggs they are sustaining are cleared. If a carrier is supporting the front and dies, the marines don't then have to clear X number of eggs AFTER the kill.

# Testing Photographs and Procedure

1. Spawn as egg carrier.
2. Plant egg on normal weeds.
3. Move 15+ tiles away.
4. Wait 1 minute
5. OR Wait 5 minutes within 14 tiles.
6. Kill the carrier.

The 5 minute lifetime of the eggs will also clear the eggs in the case that the carrier is admin deleted, or some other weird stuff happens which doesn't result in a death. This will also catch carriers de-evolving

<details>
<summary>Screenshots & Videos</summary>

[...]

PR #5054, opened 2023-11-28
Egg Changes

Zonespace27:
# About the pull request
Now that https://github.com/cmss13-devs/cmss13/pull/4716 's been merged, here's the changes I've wanted to make for eggs for a bit now.

- Eggs can no longer be planted in vehicles
- Eggs can no longer be planted on tiles that would obstruct the egg. 

- In the vast majority of cases, you cannot tell that a vehicle you're entering is infested with you-instantly-lose eggs. While normally getting hit with an egg is reasonably predictable (there are weeds == possibility of eggs), having no indication of it until you go inside and get hugged isn't fun or balanced.
- Eggs being able to layer under things like grass is technically a "feature", but not one I would like to keep. Traps serve the exact same purpose, but are much harder to mass produce (as intended). Eggs should be cheaper but more detectable, which doesn't align them being able to be near-fully hidden.
<details>

image

</details>

# Changelog
:cl:
balance: You cannot plant eggs inside vehicles.
balance: Eggs can no longer be planted on tiles with objects that would obscure them.
/:cl:


fira (approved):

code/modules/mob/living/carbon/xenomorph/egg_item.dm
```diff
@@ -106,6 +106,12 @@
 		to_chat(user, SPAN_XENOWARNING("[src] cannot be planted inside a vehicle."))
 		return
 
+	for(var/obj/object in T.contents)
+		var/obj/effect/alien/egg/xeno_egg = /obj/effect/alien/egg
+		if(object.layer > initial(xeno_egg.layer))
+			to_chat(user, SPAN_XENOWARNING("[src] cannot be planted below objects that would obscure it."))
```
fira, line 112: Ehh i don't know, what if the object gets put after? I wasn't a fan of previous one but i think it made more sense

Releasethesea:
since this is TM'd i assume i can leave feedback on it, so far my experience has been very inconsistent i can place eggs in some corners and some others i cant ive also not been able to place eggs when there is a railing next to them, aswell as from what ive seen papers and other items on the floors, like roller chairs

code/modules/mob/living/carbon/xenomorph/egg_item.dm
Zonespace27: Missed this earlier; I don't think it'll be a problem, there aren't many things that'd layer over an egg's sprite that xenoes can create or move.

Birdtalon (approved):
TM has been going quite a while with no obvious unintended effects.

PR #5059, opened 2023-11-28
Eggsac carrier bugfix

Birdtalon:
# About the pull request

So in one of various refactors of #4716 I removed some important code from `Burst()` for the new egg type which removes it from the carrier's sustained eggs. This is bad because we keep trying to call procs on eggs which have already burst etc.

This removes the egg from the carrier's sustain list properly on Burst().
I also added a check in Life() on the carrier to remove it from their sustained eggs. If something funky happens we don't want to impact the player from playing

<details>

</details>

# Changelog
🆑 
fix: Eggsack carrier eggs get properly removed from sustained list upon bursting.
/🆑

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/mutators/strains/carrier/eggsac.dm
+++ b/code/modules/mob/living/carbon/xenomorph/mutators/strains/carrier/eggsac.dm
@@ -67,8 +67,6 @@
 			remove_egg_owner(my_egg)
 			my_egg.start_unstoppable_decay()
 			to_chat(bound_xeno, SPAN_XENOWARNING("You can only sustain [egg_sustain_cap] eggs off hive weeds! Your oldest placed egg is decaying rapidly."))
-		else
-			eggs_sustained -= my_egg
 
 	for(var/obj/effect/alien/egg/carrier_egg/my_egg as anything in eggs_sustained)
 		//Get the distance from us to our sustained egg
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/mutators/strains/carrier/eggsac.dm
```diff
@@ -63,9 +63,12 @@
...
 		var/obj/effect/alien/egg/carrier_egg/my_egg = eggs_sustained[1]
-		remove_egg_owner(my_egg)
-		my_egg.start_unstoppable_decay()
-		to_chat(bound_xeno, SPAN_XENOWARNING("You can only sustain [egg_sustain_cap] eggs off hive weeds! Your oldest placed egg is decaying rapidly."))
+		if(my_egg)
+			remove_egg_owner(my_egg)
+			my_egg.start_unstoppable_decay()
+			to_chat(bound_xeno, SPAN_XENOWARNING("You can only sustain [egg_sustain_cap] eggs off hive weeds! Your oldest placed egg is decaying rapidly."))
+		else
+			eggs_sustained -= my_egg
```
fira, line 71: Bit confused, how does that work? Won't that be `eggs_sustained -= null` ?

Birdtalon: Yes you're right. It doesn't work and shouldn't be here.

fira:
As per what harry posted on discord I think you need both 🆑 either as text or emojis for it to work

Birdtalon:
> As per what harry posted on discord I think you need both 🆑 either as text or emojis for it to work

Oh yea I didn't even notice one wasn't an emoji version!

fira (approved):

realforest2001 (approved):

PR #5079, opened 2023-11-30
Consistent LZ names.

blackdragonTOW:
# About the pull request

Alters the lz1/lz2 names of each map placed object to have LZ1 or LZ2 in the front of the name, that way there's never any confusion over which is which when looking at the DS consoles.  "We're going to LZ2" should never be a point of confusion for new POs anymore.

Information good.  

<details>

</details>

# Changelog
:cl:
maptweak: Standardized the names of LZs to include the name of the LZ.
/:cl:

### DIFF
--- a/maps/map_files/BigRed/BigRed.dmm
+++ b/maps/map_files/BigRed/BigRed.dmm
@@ -119,7 +119,7 @@
 /area/bigredv2/outside/space_port)
 "aau" = (
 /obj/docking_port/stationary/marine_dropship/lz1{
-	name = "Communications Landing Zone"
+	name = "LZ1: Communications Landing Zone"
 	},
 /turf/open/floor/plating,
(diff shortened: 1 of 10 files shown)


HeresKozmos:
Pretty based

fira:
Well if argue this makes the "Landing Zone" in the name redundant

blackdragonTOW:
The name edits will only show up in the dropship console and remote consoles, having good flavor names is nice but if the pilot doesnt know which is LZ1 and which is LZ2 it can lead to confusion. 

If the lz1 or lz2 objects are not assigned a custom name, they default to Landing Zone 1, or Landing Zone 2, respectively.  Hence why the only changes here are for LZs that have custom names.

Here is a direct example of what I mean:
Capture

Nanu308 (approved):
lgtm

PR #5083, opened 2023-11-30
Functional M5 Gasmask 2: Electric Boogaloo

Ediblebomb:
# About the pull request
Makes the M5 Gasmask actually work
70% NVG code 30% trial and error

### This time I *probably* won't fuck things up.
#5072 
Consistency, the regular gasmask protects you against CN20 Nerve Gas, whereas the helmet gasmask previously did not.
<details>

https://github.com/cmss13-devs/cmss13/assets/95509996/8f4c2eed-4ad6-477d-83d4-1b87630d7ff7

</details>

# Changelog
:cl: Ediblebomb
add: M5 Helmet Gasmask now functions as a gasmask when in a marine helmet (and yes, gasmasks do still have some functionality)
/:cl:

### DIFF
--- a/code/game/objects/items/props/helmetgarb.dm
+++ b/code/game/objects/items/props/helmetgarb.dm
@@ -502,11 +502,11 @@
 
 	var/obj/item/Helmet = HelmetInternalInv.master_object
 
-	if(!istype(Helmet, /obj/item/clothing/head/helmet/marine))
+	if(!istype(Helmet, /obj/item/clothing/head/helmet))
 		return
 
-	Helmet.flags_inventory = BLOCKSHARPOBJ | COVERMOUTH | COVEREYES | ALLOWINTERNALS | BLOCKGASEFFECT | ALLOWREBREATH | ALLOWCPR
-	Helmet.flags_inv_hide = HIDEEARS|HIDEFACE|HIDELOWHAIR
+	Helmet.flags_inventory = initial(Helmet.flags_inventory)+BLOCKGASEFFECT
+	Helmet.flags_inv_hide = initial(Helmet.flags_inv_hide)+HIDEFACE
 	..()
 
 /obj/item/prop/helmetgarb/helmet_gasmask/on_exit_storage(obj/item/storage/internal/HelmetInternalInv)
@@ -515,11 +515,11 @@
 
 	var/obj/item/Helmet = HelmetInternalInv.master_object
 
-	if(!istype(Helmet, /obj/item/clothing/head/helmet/marine))
+	if(!istype(Helmet, /obj/item/clothing/head/helmet))
 		return
 
-	Helmet.flags_inventory = BLOCKSHARPOBJ
-	Helmet.flags_inv_hide = HIDEEARS
+	Helmet.flags_inventory = initial(Helmet.flags_inventory)
+	Helmet.flags_inv_hide = initial(Helmet.flags_inv_hide)
 	return..()
 
 /obj/item/prop/helmetgarb/trimmed_wire
--- a/code/game/objects/items/props/helmetgarb.dm
+++ b/code/game/objects/items/props/helmetgarb.dm
@@ -496,30 +496,30 @@
 	desc = "The USCM had its funding pulled for these when it became apparent that not every deployed enlisted was wearing a helmet 24/7; much to the bafflement of UA High Command."
 	icon_state = "helmet_gasmask"
 
-/obj/item/prop/helmetgarb/helmet_gasmask/on_enter_storage(obj/item/storage/internal/HelmetInternalInv)
-	if(!istype(HelmetInternalInv))
+/obj/item/prop/helmetgarb/helmet_gasmask/on_enter_storage(obj/item/storage/internal/helmet_internal_inventory)
+	if(!istype(helmet_internal_inventory))
 		return
 
-	var/obj/item/Helmet = HelmetInternalInv.master_object
+	var/obj/item/clothing/head/helmet/helmet_item = helmet_internal_inventory.master_object
 
-	if(!istype(Helmet, /obj/item/clothing/head/helmet))
+	if(!istype(helmet_item))
 		return
 
-	Helmet.flags_inventory = initial(Helmet.flags_inventory)+BLOCKGASEFFECT
-	Helmet.flags_inv_hide = initial(Helmet.flags_inv_hide)+HIDEFACE
+	helmet_item.flags_inventory|=BLOCKGASEFFECT
+	helmet_item.flags_inv_hide|=HIDEFACE
 	..()
 
-/obj/item/prop/helmetgarb/helmet_gasmask/on_exit_storage(obj/item/storage/internal/HelmetInternalInv)
-	if(!istype(HelmetInternalInv))
+/obj/item/prop/helmetgarb/helmet_gasmask/on_exit_storage(obj/item/storage/internal/helmet_internal_inventory)
+	if(!istype(helmet_internal_inventory))
 		return
 
-	var/obj/item/Helmet = HelmetInternalInv.master_object
+	var/obj/item/clothing/head/helmet/helmet_item = helmet_internal_inventory.master_object
 
-	if(!istype(Helmet, /obj/item/clothing/head/helmet))
+	if(!istype(helmet_item))
 		return
 
-	Helmet.flags_inventory = initial(Helmet.flags_inventory)
-	Helmet.flags_inv_hide = initial(Helmet.flags_inv_hide)
+	helmet_item.flags_inventory = initial(helmet_item.flags_inventory)
+	helmet_item.flags_inv_hide = initial(helmet_item.flags_inv_hide)
 	return..()
 
 /obj/item/prop/helmetgarb/trimmed_wire
--- a/code/game/objects/items/props/helmetgarb.dm
+++ b/code/game/objects/items/props/helmetgarb.dm
@@ -498,6 +498,8 @@
 
 /obj/item/prop/helmetgarb/helmet_gasmask/on_enter_storage(obj/item/storage/internal/helmet_internal_inventory)
 	..()
+	if(!istype(helmet_internal_inventory))
+		return
 	var/obj/item/clothing/head/helmet/helmet_item = helmet_internal_inventory.master_object
 
 	if(!istype(helmet_item))
@@ -508,13 +510,15 @@
 
 /obj/item/prop/helmetgarb/helmet_gasmask/on_exit_storage(obj/item/storage/internal/helmet_internal_inventory)
 	..()
+	if(!istype(helmet_internal_inventory))
+		return
 	var/obj/item/clothing/head/helmet/helmet_item = helmet_internal_inventory.master_object
 
 	if(!istype(helmet_item))
 		return
 
-	helmet_item.flags_inventory = initial(helmet_item.flags_inventory)
-	helmet_item.flags_inv_hide = initial(helmet_item.flags_inv_hide)
+	helmet_item.flags_inventory &= ~(BLOCKGASEFFECT)
+	helmet_item.flags_inv_hide &= ~(HIDEFACE)
 
 /obj/item/prop/helmetgarb/trimmed_wire
 	name = "trimmed barbed wire"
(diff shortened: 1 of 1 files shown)


code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,32 @@
...
+	if(!istype(HelmetInternalInv))
+		return
+
+	var/obj/item/Helmet = HelmetInternalInv.master_object
+
+	if(!istype(Helmet, /obj/item/clothing/head/helmet))
+		return
+
+	Helmet.flags_inventory = initial(Helmet.flags_inventory)+BLOCKGASEFFECT
+	Helmet.flags_inv_hide = initial(Helmet.flags_inv_hide)+HIDEFACE
```
Drulikar, line 509: If your intention is to only add these bit flags (as in turning these flags on but leaving whatever is already set as is), you would do it like this:
```suggestion
	Helmet.flags_inventory |= BLOCKGASEFFECT
	Helmet.flags_inv_hide |= HIDEFACE
```

See https://www.byond.com/docs/ref/#/operator/|

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,32 @@
 	desc = "The USCM had its funding pulled for these when it became apparent that not every deployed enlisted was wearing a helmet 24/7; much to the bafflement of UA High Command."
 	icon_state = "helmet_gasmask"
 
+/obj/item/prop/helmetgarb/helmet_gasmask/on_enter_storage(obj/item/storage/internal/HelmetInternalInv)
+	if(!istype(HelmetInternalInv))
+		return
+
+	var/obj/item/Helmet = HelmetInternalInv.master_object
```
Drulikar, line 503: Use snake_case; not PascalCase.

See https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#type-paths-must-be-snake-case

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,32 @@
...
+
+	Helmet.flags_inventory = initial(Helmet.flags_inventory)+BLOCKGASEFFECT
+	Helmet.flags_inv_hide = initial(Helmet.flags_inv_hide)+HIDEFACE
+	..()
+
+/obj/item/prop/helmetgarb/helmet_gasmask/on_exit_storage(obj/item/storage/internal/HelmetInternalInv)
+	if(!istype(HelmetInternalInv))
+		return
+
+	var/obj/item/Helmet = HelmetInternalInv.master_object
```
Drulikar, line 516: Use snake_case; not PascalCase.

See https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#type-paths-must-be-snake-case

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,32 @@
...
+
+	if(!istype(Helmet, /obj/item/clothing/head/helmet))
+		return
+
+	Helmet.flags_inventory = initial(Helmet.flags_inventory)+BLOCKGASEFFECT
+	Helmet.flags_inv_hide = initial(Helmet.flags_inv_hide)+HIDEFACE
+	..()
+
+/obj/item/prop/helmetgarb/helmet_gasmask/on_exit_storage(obj/item/storage/internal/HelmetInternalInv)
```
Drulikar, line 512: Use snake_case; not PascalCase.

See https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#type-paths-must-be-snake-case

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,32 @@
 	desc = "The USCM had its funding pulled for these when it became apparent that not every deployed enlisted was wearing a helmet 24/7; much to the bafflement of UA High Command."
 	icon_state = "helmet_gasmask"
 
+/obj/item/prop/helmetgarb/helmet_gasmask/on_enter_storage(obj/item/storage/internal/HelmetInternalInv)
```
Drulikar, line 499: Use snake_case; not PascalCase.

See https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#type-paths-must-be-snake-case

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,32 @@
...
+	Helmet.flags_inv_hide = initial(Helmet.flags_inv_hide)+HIDEFACE
+	..()
+
+/obj/item/prop/helmetgarb/helmet_gasmask/on_exit_storage(obj/item/storage/internal/HelmetInternalInv)
+	if(!istype(HelmetInternalInv))
+		return
+
+	var/obj/item/Helmet = HelmetInternalInv.master_object
+
+	if(!istype(Helmet, /obj/item/clothing/head/helmet))
```
Drulikar, line 518: Sort of moot, but if you changed `var/obj/item/Helmet` earlier to be `var/obj/item/clothing/head/helmet/helmet_item` you could simplify this check to be just `if(!istype(helmet_item))`

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,32 @@
 	desc = "The USCM had its funding pulled for these when it became apparent that not every deployed enlisted was wearing a helmet 24/7; much to the bafflement of UA High Command."
 	icon_state = "helmet_gasmask"
 
+/obj/item/prop/helmetgarb/helmet_gasmask/on_enter_storage(obj/item/storage/internal/HelmetInternalInv)
+	if(!istype(HelmetInternalInv))
+		return
+
+	var/obj/item/Helmet = HelmetInternalInv.master_object
+
+	if(!istype(Helmet, /obj/item/clothing/head/helmet))
```
Drulikar, line 505: Sort of moot, but if you changed `var/obj/item/Helmet` earlier to be `var/obj/item/clothing/head/helmet/helmet_item` you could simplify this check to be just `if(!istype(helmet_item))`

Drulikar (changes requested):

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,32 @@
...
+	if(!istype(helmet_internal_inventory))
+		return
+
+	var/obj/item/clothing/head/helmet/helmet_item = helmet_internal_inventory.master_object
+
+	if(!istype(helmet_item))
+		return
+
+	helmet_item.flags_inventory|=BLOCKGASEFFECT
+	helmet_item.flags_inv_hide|=HIDEFACE
```
Drulikar, line 509:
```suggestion
	helmet_item.flags_inventory |= BLOCKGASEFFECT
	helmet_item.flags_inv_hide |= HIDEFACE
```

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,32 @@
...
+	..()
```
Drulikar, line 510: The norm for this proc appears to be to have the parent call first. At the moment with it here your returns earlier will prevent the parent call from occurring.

Move this to be the first line.

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,32 @@
 	desc = "The USCM had its funding pulled for these when it became apparent that not every deployed enlisted was wearing a helmet 24/7; much to the bafflement of UA High Command."
 	icon_state = "helmet_gasmask"
 
+/obj/item/prop/helmetgarb/helmet_gasmask/on_enter_storage(obj/item/storage/internal/helmet_internal_inventory)
+	if(!istype(helmet_internal_inventory))
+		return
```
Drulikar, line 501: This test is likely unnecessary since the parent call isn't performing any test like this.

You should be able to remove these lines without issue. If this actually can occur, we would probably rather it runtime so the cause can be fixed rather than just testing and ignoring silently.

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,32 @@
...
+	if(!istype(helmet_item))
+		return
+
+	helmet_item.flags_inventory|=BLOCKGASEFFECT
+	helmet_item.flags_inv_hide|=HIDEFACE
+	..()
+
+/obj/item/prop/helmetgarb/helmet_gasmask/on_exit_storage(obj/item/storage/internal/helmet_internal_inventory)
+	if(!istype(helmet_internal_inventory))
+		return
```
Drulikar, line 514: This test is likely unnecessary since the parent call isn't performing any test like this.

You should be able to remove these lines without issue. If this actually can occur, we would probably rather it runtime so the cause can be fixed rather than just testing and ignoring silently.

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,32 @@
...
+
+	var/obj/item/clothing/head/helmet/helmet_item = helmet_internal_inventory.master_object
+
+	if(!istype(helmet_item))
+		return
+
+	helmet_item.flags_inventory = initial(helmet_item.flags_inventory)
+	helmet_item.flags_inv_hide = initial(helmet_item.flags_inv_hide)
+	return..()
```
Drulikar, line 523: The norm for this proc appears to be to have the parent call first. At the moment with it here your returns earlier will prevent the parent call from occurring.

Move this to be the first line just as `..()` since this proc isn't normally returning anything.

Drulikar (changes requested):

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,26 @@
...
+
+/obj/item/prop/helmetgarb/helmet_gasmask/on_exit_storage(obj/item/storage/internal/helmet_internal_inventory)
+	..()
+	var/obj/item/clothing/head/helmet/helmet_item = helmet_internal_inventory.master_object
+
+	if(!istype(helmet_item))
+		return
+
+	helmet_item.flags_inventory = initial(helmet_item.flags_inventory)
+	helmet_item.flags_inv_hide = initial(helmet_item.flags_inv_hide)
```
Drulikar, line 517: Technically it might be better to return the flags this way; but I'm not sure if there's anything else altering helmet flags at the moment. Though this way also assumes we are the only source of either of these flags.

```suggestion
	helmet_item.flags_inventory &= ~(BLOCKGASEFFECT)
	helmet_item.flags_inv_hide &= ~(HIDEFACE)
```

So optional, just pointing it out.

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,26 @@
 	desc = "The USCM had its funding pulled for these when it became apparent that not every deployed enlisted was wearing a helmet 24/7; much to the bafflement of UA High Command."
 	icon_state = "helmet_gasmask"
 
+/obj/item/prop/helmetgarb/helmet_gasmask/on_enter_storage(obj/item/storage/internal/helmet_internal_inventory)
+	..()
+	var/obj/item/clothing/head/helmet/helmet_item = helmet_internal_inventory.master_object
```
Drulikar, line 501:
```suggestion
/obj/item/prop/helmetgarb/helmet_gasmask/on_enter_storage(obj/item/storage/internal/helmet_internal_inventory)
	..()
	if(!istype(helmet_internal_inventory))
		return
	var/obj/item/clothing/head/helmet/helmet_item = helmet_internal_inventory.master_object
```
It appears this is an incorrect assumption the argument passed will be specifically an `obj/item/storage/internal`:
image
So my review https://github.com/cmss13-devs/cmss13/pull/5083#discussion_r1414834147 ended up to be incorrect. I missed that the argument was changed to be `obj/item/storage/internal` instead of just `obj/item/storage`

code/game/objects/items/props/helmetgarb.dm
```diff
@@ -496,6 +496,26 @@
...
+
+	if(!istype(helmet_item))
+		return
+
+	helmet_item.flags_inventory |= BLOCKGASEFFECT
+	helmet_item.flags_inv_hide |= HIDEFACE
+
+/obj/item/prop/helmetgarb/helmet_gasmask/on_exit_storage(obj/item/storage/internal/helmet_internal_inventory)
+	..()
+	var/obj/item/clothing/head/helmet/helmet_item = helmet_internal_inventory.master_object
```
Drulikar, line 511: "

```suggestion
/obj/item/prop/helmetgarb/helmet_gasmask/on_exit_storage(obj/item/storage/internal/helmet_internal_inventory)
	..()
	if(!istype(helmet_internal_inventory))
		return
	var/obj/item/clothing/head/helmet/helmet_item = helmet_internal_inventory.master_object
```

Drulikar (changes requested):
Changes appear to work in game other than the runtime that one of my suggestions caused.

PR #5084, opened 2023-11-30
adds returns to prevent runtimes in /datum/component/healing_reduction/process(delta_time) hopefuly

cuberound:
# About the pull request

when there is no parent the function deleted src but then kept going and asked for parent of null, it calls return now instead

runtime bad 

<details>

</details>

# Changelog
:cl:
fix: fixed a runtime in /datum/component/healing_reduction/process(delta_time)
/:cl:

### DIFF
--- a/code/datums/components/healing_reduction.dm
+++ b/code/datums/components/healing_reduction.dm
@@ -70,7 +70,7 @@ Humans will take continuous damage instead.
 	var/atom/parent_atom = parent
 	parent_atom.remove_filter("healing_reduction")
 
-/datum/component/healing_reduction/proc/stat_append(mob/mob_with_list, list/stat_list)
+/datum/component/healing_reduction/proc/stat_append(mob/target_mob, list/stat_list)
 	SIGNAL_HANDLER
 	stat_list += "Healing Reduction: [healing_reduction]/[max_buildup]"
 


code/datums/components/healing_reduction.dm
```diff
@@ -33,15 +33,18 @@ Humans will take continuous damage instead.
...
+
 	healing_reduction = max(healing_reduction - healing_reduction_dissipation * delta_time, 0)
 
+	if(healing_reduction <= 0)
+		qdel(src)
+		return
+
 	if(ishuman(parent)) //deals brute to humans
 		var/mob/living/carbon/human/H = parent
 		H.apply_damage(healing_reduction_dissipation * delta_time, BRUTE)
```
Drulikar, line 46:
```suggestion
		var/mob/living/carbon/human/human_parent = parent
		human_parent.apply_damage(healing_reduction_dissipation * delta_time, BRUTE)
```

Drulikar: Since we're already here, should bring this code in lines w/ the code style guidelines: https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#use-descriptive-and-obvious-names


`A` below could also be atom_parent; but its outside of the context for me to make it as a suggestion.

Drulikar (changes requested):

cuberound:
was a bit tired yesterday when I found it, all vars in the file should be ranamed

code/datums/components/healing_reduction.dm
cuberound: can I make comment like this on PRs of others? if so how?

Drulikar: See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request#starting-a-review

Drulikar (changes requested):

PR #5088, opened 2023-12-01
Minor sound changes and clarity.

zzzmike:
# About the pull request

Adds the armorequip noise to backpack, uniform, shoes. Changes the whoosh (disarm-stun) noise to only trigger when you stun someone. Adds whoosh to tablestun, and changes tablestun text to be more clear to new players that it stuns.

In human vs human fights, sometimes your back/uniform/shoes get stolen. The armorequip sound is a good clue to draw attention to when this happens, both for the victim and the attacker, so adding it to a few more important gear pieces should be a benefit in these fringe cases. The effect is subtle enough, and the situation rare enough, that I don't think it would be too much. The whoosh sound is currently triggering when you "disarm" someone even when they are holding nothing in hand, which doesn't make sense. Now it's reserved for when a disarm-stun goes off, or a tablestun because tablestun being silent doesn't make sense either. From my experience talking to players, tablestun is not very well known outside of veteran players who familiarized themselves with that niche interaction, so adding a sound and changing the text to make the effect clear should be a benefit.

<details>

</details>

# Changelog

:cl:
ui: tablestun text is more clear that it stuns
soundadd: equip/unequip noise added to uniform, backpack, shoes. whoosh added to tablestun.
sounddel: for disarms, whoosh only triggers on stun. otherwise, the miss sound plays
/:cl:

### DIFF
--- a/code/game/objects/structures/tables_racks.dm
+++ b/code/game/objects/structures/tables_racks.dm
@@ -275,6 +275,7 @@
 			if(user.a_intent == INTENT_HARM)
 				if(user.grab_level > GRAB_AGGRESSIVE)
 					if (prob(15)) M.apply_effect(5, WEAKEN)
+					playsound(loc, 'sound/weapons/thudswoosh.ogg', 25, 1, 7)
 					M.apply_damage(8, def_zone = "head")
 					user.visible_message(SPAN_DANGER("[user] slams [M]'s face against [src]!"),
 					SPAN_DANGER("You slam [M]'s face against [src]!"))
@@ -285,8 +286,9 @@
 			else if(user.grab_level >= GRAB_AGGRESSIVE)
 				M.forceMove(loc)
 				M.apply_effect(5, WEAKEN)
-				user.visible_message(SPAN_DANGER("[user] throws [M] on [src]."),
-				SPAN_DANGER("You throw [M] on [src]."))
+				playsound(loc, 'sound/weapons/thudswoosh.ogg', 25, 1, 7)
+				user.visible_message(SPAN_DANGER("[user] throws [M] on [src], stunning them!"),
+				SPAN_DANGER("You throw [M] on [src], stunning them!"))
 		return
 
 	if(HAS_TRAIT(W, TRAIT_TOOL_WRENCH) && !(user.a_intent == INTENT_HELP))
--- a/code/modules/clothing/shoes/marine_shoes.dm
+++ b/code/modules/clothing/shoes/marine_shoes.dm
@@ -21,6 +21,8 @@
 	var/armor_stage = 0
 	items_allowed = list(/obj/item/attachable/bayonet, /obj/item/weapon/throwing_knife, /obj/item/weapon/gun/pistol/holdout, /obj/item/weapon/gun/pistol/clfpistol, /obj/item/tool/screwdriver, /obj/item/tool/surgery/scalpel, /obj/item/weapon/straight_razor)
 	var/knife_type
+	pickup_sound = "armorequip"
+	drop_sound = "armorequip"
 
 /obj/item/clothing/shoes/marine/Initialize(mapload, ...)
 	. = ..()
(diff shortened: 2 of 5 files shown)


zzzmike:
tablehit1 actually seems quite reasonable at 25%, much more reasonable compared to playing the sound outside of byond

cuberound:
before some maintainer comes along to tell you himself, rename the W M B G vars :)

code/modules/mob/mob_grab.dm
```diff
@@ -88,7 +88,7 @@
 
 /obj/item/grab/proc/progress_passive(mob/living/carbon/human/user, mob/living/victim)
 	user.grab_level = GRAB_AGGRESSIVE
-	playsound(src.loc, 'sound/weapons/thudswoosh.ogg', 25, 1, 7)
+	playsound(src.loc, 'sound/weapons/Genhit.ogg', 25, 1, 7)
```
Birdtalon, line 91: I don't think this one fits really.

code/game/objects/structures/tables_racks.dm
```diff
@@ -275,18 +275,20 @@
 			if(user.a_intent == INTENT_HARM)
 				if(user.grab_level > GRAB_AGGRESSIVE)
 					if (prob(15)) M.apply_effect(5, WEAKEN)
+					playsound(loc, 'sound/weapons/tablehit1.ogg', 25, 1, 7)
```
Birdtalon, line 278: Aren't we already playing this sound already in this block?

Birdtalon (commented):
Couple of sounds I'm not sure really fit. I think the uniform, boots backpack sound additions are all good.

code/modules/mob/mob_grab.dm
zzzmike: It's supposed to be the sound of them hitting the ground, as aggressive grab forces the floored status. But I'm open to suggestions if you have another sound in mind👍

code/game/objects/structures/tables_racks.dm
zzzmike: Woops, my bad, fixed 😳

Drulikar (changes requested):
While you're at it you may as well add a sfx when fireman carrying a person. Can just be thudswoosh again; unless you find something else you like for a grab chain.

code/modules/clothing/shoes/marine_shoes.dm
```diff
@@ -21,6 +21,8 @@
 	var/armor_stage = 0
 	items_allowed = list(/obj/item/attachable/bayonet, /obj/item/weapon/throwing_knife, /obj/item/weapon/gun/pistol/holdout, /obj/item/weapon/gun/pistol/clfpistol, /obj/item/tool/screwdriver, /obj/item/tool/surgery/scalpel, /obj/item/weapon/straight_razor)
 	var/knife_type
+	pickup_sound = "armorequip"
```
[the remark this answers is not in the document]
zzzmike: changed it now so I'm only adding an unequip sound (the purpose of the change in the first place is for a sound to play if it's unexpectedly dropped straight to ground, like if someone strips it in briefing when you aren't expecting it)

code/modules/mob/mob_grab.dm
zzzmike: no prob, will leave as is

zzzmike:
> While you're at it you may as well add a sfx when fireman carrying a person. Can just be thudswoosh again; unless you find something else you like for a grab chain.

Disarm-stun is thudswoosh, passive-grab is thudswoosh, aggressive grab is thudswoosh... I don't think I can bring myself to add more, am gonna leave this alone for the moment, maybe it can be looked at in a separate PR or something. At least it's obvious from the animation when you carry them - whereas it's not super obvious when you passive or aggressive grab (there's slight animation but I feel like it can easily be missed in the heat of combat)

code/modules/clothing/under/marine_uniform.dm
```diff
@@ -17,6 +17,8 @@
 	armor_rad = CLOTHING_ARMOR_NONE
 	armor_internaldamage = CLOTHING_ARMOR_LOW
 	flags_jumpsuit = UNIFORM_SLEEVE_ROLLABLE
+	pickup_sound = "armorequip"
```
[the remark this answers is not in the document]
zzzmike: changed it now so I'm only adding an unequip sound (the purpose of the change in the first place is for a sound to play if it's unexpectedly dropped straight to ground, like if someone strips it in briefing when you aren't expecting it)

zzzmike:
mob_grab is back to default as per the review - thanks

PR #5096, opened 2023-12-01
Weeds are conserved under walls being upgraded and destroyed

### LINKED ISSUE #5060: Upgrading resin walls to thick resin walls no longer allows nesting
### Testmerges

n/a

### Round ID

_No response_

### Description of the bug

When upgrading a resin wall to a thick resin wall the wall loses its ability to nest.

This is caused by upgrading the wall removes the hive weeds underneath so this return is hit.

https://github.com/cmss13-devs/cmss13/blob/c8d3a20fe67ebc51ae5dc591f4dad690f51be44d/code/modules/mob/living/carbon/xenomorph/XenoProcs.dm#L670-L674

The wall should retain the ability to nest hosts.

1. Make a normal resin wall on hive weeds
2. Upgrade with hivelord to thick resin
3. Try to nest a human

### Issue Bingo

Birdtalon:
# About the pull request

Fixes #5060

Side effect of this PR is that when resin walls are destroyed then do not remove the weeds underneath. However those weeds should still be linked to a nearby node and recede as normal when that is destroyed.

<details>

</details>

# Changelog
:cl:
fix: Upgraded resin walls can now nest hosts.
/🆑

### DIFF
--- a/code/game/turfs/walls/walls.dm
+++ b/code/game/turfs/walls/walls.dm
@@ -88,8 +88,6 @@
 			if(istype(found_object, /obj/structure/sign/poster))
 				var/obj/structure/sign/poster/found_poster = found_object
 				found_poster.roll_and_drop(src)
-			if(istype(found_object, /obj/effect/alien/weeds))
-				qdel(found_object)
 
 		var/list/turf/cardinal_neighbors = list(get_step(src, NORTH), get_step(src, SOUTH), get_step(src, EAST), get_step(src, WEST))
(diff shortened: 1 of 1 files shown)


fira (approved):

realforest2001 (approved):

Drulikar (approved):
Changes appear to work, though in testing of this I notice wherever a wall is destroyed the lighting of the turf is weird. Behavior isn't unique to this PR though: 
image

PR #4996, opened 2023-11-22
Brown Gloves/Boots Surplus Vendors

567Turtle:
# About the pull request

adds brown gloves and boots to the surplus vendors

# Explain why it's good for the game

It's a drip or drown world out there, and our marines need a life-vests.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

image

</details>

# Changelog
:cl:
add: Brown boots and gloves are now vendable from the surplus vendors.
/:cl:

PR #5102, opened 2023-12-02
Adds Jungle boots to the surplus vendors

Googles-Hands:
# About the pull request
Adds the Jungle Combat Boots to the marine vendor.

This completes #4996 and let's you save even more loadout points to spend on candy.

CMjungleboots

# Changelog

:cl: Googles_Hands
add: Jungle boots are now vendable from the surplus vendors.
/:cl:


fira:
Well I don't know... Wouldn't have approved #4996 to begin with if I knew they were in loadouts to begin with...
The more loadout items you make available everywhere else the less of a special thing they become and the more they just become an hinderance for being in loadout system at all

fira:
People having more loadout points to spend is not neccessarily better vs what loadouts are trying to achieve. If it was we'd just increase the points, no need to dump half the codebase in every vendor.

My point is the more things are made readily available the less interesting they are as loadout item, effectively reducing outfiting possibilities since anyone can now take them instead. I think of loadouts not just as drip but as providing objects of interest or with some RP value. 

Imagine if we made a plush vendor with all the new plushs. What do you think's gonna happen? Well nobody will have "theirs", instead you'd have 80 random ones scattered everywhere in briefing room.

So yeah i don't have immediate concerns for just this PR, but i don't think this reasoning is one we'd want to follow in the future...

fira (approved):

PR #5106, opened 2023-12-02
Boot icon fixes.

silencer-pl:
# About the pull request

This MR moves boots into an overlay under the uniform and touches up boot on mob sprites to look like actual boots, since they now don't have to be cut off for the uniforms.

The only visible effect on CM of this change is that the CL and Synth can now wear boots with their black skirt suit, see pictures.

The more significant effect is that this properly layers socks/boots/shoes under uniforms, meaning that should someone downstream say add uniforms with skirts, boots/stockings will not overlay on top of them.

This also fixes the RMC Boots and makes the item icon properly update to reflect whether they have a knife or not (as they were missing a proc that did that)

Finally it also removes a garbage var and and an empty check using it.

Due to how the global part of update_icon() handles item_state paths, removing the -1 onmob icons breaks onmob icons that appear on spawned-in npcs. I'm sure fixing that will break it in 100 other places, so I opted for the -1 icons instead.

Proper, clear layering that does not require dmi shenanigans is good for both spriters and coders.
Onmob icons now properly represent what the boots actually are, which makes the icons universal which is also good.
Happy downstream that does not have to implement this as a separate fix :)

<details>

Skirt and boots:

image
image

RMC Boot icon:

image

</details>

# Changelog
:cl:silencer_pl
fix: Boots now properly render under skirts/shorts/uniforms etc without icon edits
fix: RMC Boot icons now properly reflect if an object is holstered in them
/:cl:

### DIFF
--- a/code/modules/clothing/shoes/marine_shoes.dm
+++ b/code/modules/clothing/shoes/marine_shoes.dm
@@ -248,6 +248,12 @@
 	)
 	flags_atom = NO_NAME_OVERRIDE
 
+/obj/item/clothing/shoes/royal_marine/update_icon()
+	if(stored_item)
+		icon_state = "[initial(icon_state)]-1"
+	else
+		icon_state = initial(icon_state)
+
 /obj/item/clothing/shoes/royal_marine/knife
 /obj/item/clothing/shoes/royal_marine/knife/Initialize(mapload, ...)
 	. = ..()


silencer-pl:
Indicentally, with how /obj/proc/update_icon() is handled, I dont think item_state actually does antyhing at present, I went into more detail about it at https://discord.com/channels/150315577943130112/745447048261795890/1180603291365806120

silencer-pl:
Closing. I can't wait 9 days for a review of 3 lines of code, considering my other MR involving sprites was just let to rot, I can take the hint. This has been merged into my downstream, feel free to reopen the fix when someone feels like actually putting this in.

PR #5107, opened 2023-12-02
Componentizes rechargers & unwrenching

Zonespace27:
# About the pull request
Converts rechargers into a component.
Also converts the ability to unwrench things into an element due to it being needed to have feature parity w/ recharger componentization.

Marked as draft until I add cell components to everything that can be put into a recharger.

I want to be able to add a form of recharger to more things, so componentization is necessary.

🆑 
fix: You can no longer unwrench certain atmos parts.
\🆑


fira (changes requested):
Looks good to me overall

cm13-github:

cm13-github:

code/_onclick/human.dm
```diff
@@ -86,6 +86,7 @@
 	return HANDLE_CLICK_PASS_THRU
 
 /atom/proc/attack_hand(mob/user)
+	SEND_SIGNAL(src, COMSIG_ATOM_ATTACK_HAND, user)
```
fira, line 89: bit of a problem with this is it's unreliable since not everything calls parent

code/datums/components/cell.dm
```diff
@@ -16,49 +16,65 @@
...
+		charge = min(initial_charge, max_charge)
 	src.hit_charge = hit_charge
 	src.max_recharge_tick = max_recharge_tick
 	src.charge_drain = charge_drain
 	src.display_charge = display_charge
 	src.charge_examine_range = charge_examine_range
 	src.cell_insert = cell_insert
+	if(cell_insert)
+		inserted_cell = new cell_insert_default_cell(parent)
+	src.examine_string = examine_string
```
fira, line 60: you could use `src.examine_string = examine_string || src.examine_string` (or just a if) so you don't have to duplicate the default string in arguments above

code/datums/components/recharger.dm
```diff
@@ -0,0 +1,228 @@
+/datum/component/recharger
+	dupe_mode = COMPONENT_DUPE_UNIQUE
+	/// Ref to the inserted item, if any
+	var/obj/item/inserted_item
+	/// If we're currently recharging anything
+	var/recharging = FALSE
+	/// A typecache of what items can be inserted into the recharger
+	var/list/valid_items
+	/// A list of the default acceptable items to be inserted into the recharger, in case one isn't passed in during init
+	var/static/list/default_valid_items = list(
```
fira, line 10: is this neccessary? wouldn't cells be able to work on any /obj/item? hell with minor tweaks i'm sure it'd be applicable to synths

code/datums/components/recharger.dm
```diff
@@ -0,0 +1,228 @@
...
+
+	if(charge_overlay)
+		var/obj/structure/machinery/machine_parent = parent
+		machine_parent.overlays -= charge_overlay
+		QDEL_NULL(charge_overlay)
+
+	if(custom_item_overlay)
+		var/obj/structure/machinery/machine_parent = parent
+		machine_parent.overlays -= charge_overlay
+		QDEL_NULL(charge_overlay)
```
fira, line 85: guess you mean custom_item_overlay here

code/datums/components/recharger.dm
```diff
@@ -0,0 +1,228 @@
...
+		return ELEMENT_OBJ_STOP_UNWRENCH
+
+/datum/component/recharger/proc/on_attack(datum/source, obj/item/weapon, mob/living/user, params)
+	SIGNAL_HANDLER
+
+	if(inserted_item)
+		to_chat(user, SPAN_WARNING("There's already something recharging inside [parent]."))
+		return COMPONENT_NO_AFTERATTACK
+
+	var/obj/obj_parent = parent
```
fira, line 109: type would be machinery here i think

code/game/objects/items/tools/experimental_tools.dm
```diff
@@ -274,7 +276,7 @@
 		to_chat(user, SPAN_WARNING("You don't seem to know how to use \the [src]..."))
 		return
 
-	if(!pdcell || pdcell.charge == 0)
+	if(SEND_SIGNAL(src, COMSIG_CELL_CHECK_CHARGE) & COMPONENT_CELL_CHARGE_INSUFFICIENT)
```
fira, line 279: That's gonna sound dumb but if there's no cell at all then it works
Might want to reverse such signals so the default is having no cell if no component is attached

fira (changes requested):

code/_onclick/human.dm
Zonespace27: This is unfortunately a bit of an issue that's hard to solve (see atom/attackby() and etc.) and is out of scope to fix in this PR

code/game/objects/items/tools/experimental_tools.dm
Zonespace27: The dialysis device will always have a cell component, and the proc inside the component for COMSIG_CELL_CHECK_CHARGE will return COMPONENT_CELL_CHARGE_INSUFFICIENT if there isn't a cell when one is needed.

code/datums/components/recharger.dm
Zonespace27: It's not necessary, but I do think it's good to keep it this way for now. This allows for certain rechargers to only allow certain items, or for an item that can't be placed in a recharger

fira: indeed 👍

code/game/objects/items/tools/experimental_tools.dm
fira: Yeah i meant the component, but it's true it's always present, makes sense 👍

code/_onclick/human.dm
fira: Could perhaps just toss in a comment about it instead, so future coders don't figure that out the hard way

fira (approved):

code/_onclick/human.dm
Zonespace27: done

Drulikar (changes requested):
I don't know all the possible things that are in game that might be recharged but I can confirm defibs and smart gun batteries recharge fine. However, it doesn't seem like you can recharge M2 Nightvision goggles using a cell (specifically a `/obj/item/cell/hyper`) because that is still accessing the .charge of the cell which in this case is -1. So there are still lingering references to charge that aren't using the cell component.

image

PR #5110, opened 2023-12-03
fix(preferences): fixes saving of certain preferences

Tsurupeta:
# About the pull request

Fixed the saving of certain prefs, like DISABLING RADIAL MENUS (important!!!) among others.

I'm too lazy to list them out so heres a screenshot with list of prefs that are fixed.

image

# Changelog

:cl:
fix: fixed saving of certain preferences.
/:cl:

### DIFF
--- a/code/modules/client/preferences_savefile.dm
+++ b/code/modules/client/preferences_savefile.dm
@@ -201,9 +201,7 @@
 	S["autofit_viewport"] >> auto_fit_viewport
 	S["adaptive_zoom"] >> adaptive_zoom
 	S["tooltips"] >> tooltips
-
 	S["key_bindings"] >> key_bindings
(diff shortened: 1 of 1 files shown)


Drulikar (dismissed):

code/modules/client/preferences_savefile.dm
```diff
@@ -207,6 +202,12 @@
 	S["adaptive_zoom"] >> adaptive_zoom
 	S["tooltips"] >> tooltips
 
+	S["key_bindings"] >> key_bindings
+	check_keybindings()
```
fira, line 206: Is this just for the save_prefs in check_keybindings? That doesn't look like a proper fix to me, it's still done before the sanitizing, should rather defer the saving

PR #5111, opened 2023-12-03
Fixes disarm chance calculation

zzzmike:
# About the pull request

I was testing stuff out for a different PR and thought the disarm chance (for human vs human) seemed weird at times, so I forced it to output attacker_skill_level and defender_skill_level to chat every time a disarm was done. And then found out it's not calculating it properly. This seems to fix it in limited testing but might need more testing. I'm kinda surprised it wasn't noticed earlier cause there's two signs it's wrong in certain situations even with skill_level variables being invisible: the "tackled" message should only appear when the attacker's CQC level is two or higher (otherwise it should say pushed/shoved), and also, anyone that has a CQC advantage is still disarm-stunning everyone for the same amount of time as anyone else on average ('on average' because there was apparently some variance with a tick offset bug before the TG status effect testmerge) - the code implies it should be longer.

Bugs should be fixed. From a balance perspective I don't know if it 'should' be fixed from what it is currently in it's bugged state... but regardless, it's a bug.

<details>

</details>

# Changelog
:cl:
fix: disarm chance wasn't calculating properly
/:cl:


code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -155,8 +155,8 @@
 						return held_weapon.afterattack(target,src)
 
 			var/disarm_chance = rand(1, 100)
-			var/attacker_skill_level = skills && attacking_mob.skills ? skills.get_skill_level(SKILL_CQC) : SKILL_CQC_MAX // No skills, so assume max
-			var/defender_skill_level = skills ? skills.get_skill_level(SKILL_CQC) : SKILL_CQC_MAX // No skills, so assume max
+			var/attacker_skill_level = attacking_mob.skills ? attacking_mob.skills.get_skill_level(SKILL_CQC) : SKILL_CQC_DEFAULT // No skills, so assume default
```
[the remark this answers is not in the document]
zzzmike: > assuming maximum if no skills is the intended behavior AFAIK, it's only used for things like admin-set infinite skills. having no skills allow you to bypass skillchecks.

Ah ok, I set it back. I didn't think there was an intended use-case for it being max. I was able to trigger it by creating someone in the create humans admin verb, and thought it made a lot more sense to assume default in that case.

fira (approved):

code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -155,8 +155,8 @@
 						return held_weapon.afterattack(target,src)
 
 			var/disarm_chance = rand(1, 100)
-			var/attacker_skill_level = skills && attacking_mob.skills ? skills.get_skill_level(SKILL_CQC) : SKILL_CQC_MAX // No skills, so assume max
-			var/defender_skill_level = skills ? skills.get_skill_level(SKILL_CQC) : SKILL_CQC_MAX // No skills, so assume max
+			var/attacker_skill_level = attacking_mob.skills && skills ? attacking_mob.skills.get_skill_level(SKILL_CQC) : SKILL_CQC_MAX // No skills, so assume max
+			var/defender_skill_level = skills && attacking_mob.skills ? skills.get_skill_level(SKILL_CQC) : SKILL_CQC_MAX // No skills, so assume max
```
Drulikar, line 159: Either I don't understand what you're trying to do by doing it this way, but I would think you are merely wanting to either get each mob's skill level or if they don't have skills to assume max. I don't know why whether the src having skill or not has to be taken into account when determining the skill level of two different mobs.

```suggestion
			var/attacker_skill_level = attacking_mob.skills ? attacking_mob.skills.get_skill_level(SKILL_CQC) : SKILL_CQC_MAX // No skills, so assume max
			var/defender_skill_level = skills ? skills.get_skill_level(SKILL_CQC) : SKILL_CQC_MAX // No skills, so assume max
```

Drulikar (changes requested):

code/modules/mob/living/carbon/human/human_attackhand.dm
zzzmike: @Drulikar I think it would be better if either the attacker or defender has an undefined CQC skill, to assume max for both participants, since something is a bit strange for that situation to happen. If the attacker or defender is meant to have a CQC advantage, it should be defined. Even in on-the-fly events where someone is meant to have a CQC advantage, it could be defined by admin variable edit in-game, I assume. That's why I wrote it that way.

Drulikar: So think again what Fira mentioned:
> assuming maximum if no skills is the intended behavior AFAIK, it's only used for things like admin-set infinite skills. having no skills allow you to bypass skillchecks.

So you're saying an admin set infinite skill person disarming say a rifleman who does have skills set, we need to assume now the rifleman has max SKILL_CQC because one of the parties has no skills set?

zzzmike: I took what Fira said to mean "Sometimes admins spawn characters with infinite skills to avoid headaches with them not being able to, say, open a pill bottle, or build a cade, or fly a ship, etc. which is different then "I purposefully want this character to be godly at CQC to the point of being immune to being disarm-stunned - I thought they would be specifically setting the CQC level to max in that case.

Drulikar: I still don't see how one party missing skills should result in the party not missing skills returning a different value for their skills.

zzzmike: I look at it as "If one party is missing skills, assume an equal CQC level only for purposes of disarms between them" because in the case I mentioned earlier, did the admin intend for the person they spawned to be immune to disarm stuns? And intend for 50% of the disarms they do to stun for 14 seconds each time? If they intended that much of a massive difference to the standard disarm values, why did they not specifically define the CQC level?

Drulikar: If that was the intention you'd just put the other person's skill level as the missing person's skills so they have matching skills.

Really think about what you are doing and evaluate how poor this code is as written.

Birdtalon: Agree with @Drulikar on this one. It makes no sense to me to set the defender's skill based on the attacker's skill. Treat each mob independently.

fira: Yes you're both right, my bad for stamping this approved here

Drulikar (approved):

PR #5122, opened 2023-12-04
Pylons now count only groundside marines, xenos get major boost to evo speed for 3 minutes after hijack

ihatethisengine:
# About the pull request
Pylons give larva only up to 50% of groundside (weighted) marines, instead of 40% of all humans.
Xenos get major boost to evo speed for 3 minutes after hijack, also hijacking removes ovi cd.

Larva surge works this way (but for shipside), and pylons were kinda meant to work this way too. Current implementation encourages xenos to delay hijack as long as possible to get the maximum larva, get evo etc. which is not good. It shouldn't break balance too much, you should be able to defeat marines if you have 40% of their force.

<details>

</details>

# Changelog
:cl: ihatethisengine
balance: Pylons give larva only up to 50% of groundside (weighted) marines, instead of 40% of all humans.
balance: Xenos get major boost to evo speed for 3 minutes after hijack. Hijacking resets ovi cooldown.
/:cl:


code/modules/cm_aliens/structures/special/pylon_core.dm
```diff
@@ -197,8 +197,17 @@
...
-	if(real_total_xeno_count > (length(GLOB.alive_human_list) * ENDGAME_LARVA_CAP_MULTIPLIER))
+	var/groundside_humans_weighted_count = 0
+	for(var/mob/living/carbon/human/current_human as anything in GLOB.alive_human_list)
+		if(!(isspecieshuman(current_human) || isspeciessynth(current_human)))
+			continue
+		var/datum/job/job = RoleAuthority.roles_for_mode[current_human.job]
+		if(!job)
+			continue
+		var/turf/turf = get_turf(current_human)
+		if(!is_ground_level(turf?.z))
```
fira, line 208: that's the opposite???

fira (changes requested):

code/modules/cm_aliens/structures/special/pylon_core.dm
ihatethisengine: Happens to the best, I wanted to invert is shipside level, but found this proc

ihatethisengine:
Why dont it work :sob:

Birdtalon:
I'd say this needs balance approval since xeno endgame relies largely on pylon larva at the moment to siege and then execute a successful hijack. Based on my gut feeling this PR could reduce the larva gain by anything from 20-50% some rounds which I would say is not insignificant.

ihatethisengine:
We can up the number to 50%. Theoretically all groundside marines (except doctors) should have weight of 1, so it shouldn't be an issue. Realistically if you have 20 xenos vs 40 marines it's xeno major, especially since you get new xenos when old xenos die.

GoldenDarkness55:
less numbers = you're even more encouraged to wait for evo.
Just add some kinda check for queen ovi on DS that speeds evo up tenfold for a minute or smth

cm13-github:

cm13-github:

cm13-github:

cm13-github:

Zonespace27 (changes requested):
Little bit out of scope, but could you please add something that increases evo speed for a period after the queen launches the DS?

Lgtm otherwise

ihatethisengine:
I'll try.

ihatethisengine:
Done

code/modules/cm_aliens/structures/special/pylon_core.dm
```diff
@@ -179,10 +179,9 @@
 	activated = TRUE
 	addtimer(CALLBACK(src, PROC_REF(give_larva)), XENO_PYLON_ACTIVATION_COOLDOWN, TIMER_UNIQUE|TIMER_OVERRIDE|TIMER_LOOP|TIMER_DELETE_ME)
 
+/// Looped proc via timer to give larva after time
 #define ENDGAME_LARVA_CAP_MULTIPLIER 0.4
 #define LARVA_ADDITION_MULTIPLIER 0.10
-
-/// Looped proc via timer to give larva after time
```
Drulikar, line 185: Revert this incorrect autodoc change moving it somereason to ENDGAME_LARVA_CAP_MULTIPLIER.

code/modules/cm_aliens/structures/special/pylon_core.dm
```diff
@@ -195,16 +194,24 @@
...
+		if(!(isspecieshuman(current_human) || isspeciessynth(current_human)))
+			continue
+		var/datum/job/job = GLOB.RoleAuthority.roles_for_mode[current_human.job]
+		if(!job)
+			continue
+		var/turf/turf = get_turf(current_human)
+		if(!is_ground_level(turf?.z))
+			groundside_humans_weighted_count += GLOB.RoleAuthority.calculate_role_weight(job)
+	if(real_total_xeno_count > (groundside_humans_weighted_count * ENDGAME_LARVA_CAP_MULTIPLIER))
 		return
```
Drulikar, line 210: Since you cited "Current implementation encourages xenos to delay hijack as long as possible to get the maximum larva" you'd have more effect if hitting this cap in some way was communicated to the hive. I'm not really suggesting it should have a text announcement (if it does it should only do it infrequently and not be spammy); but likely would be good to think of maybe a way for this to be indicated in the hive status.

code/modules/mob/living/carbon/xenomorph/hive_status.dm
```diff
@@ -1285,6 +1284,10 @@
 	xeno_message(SPAN_XENOANNOUNCE("You sense that [english_list(defectors)] turned their backs against their sisters and the Queen in favor of their slavemasters!"), 3, hivenumber)
 	defectors.Cut()
 
+/datum/hive_status/proc/override_evilution(evil, override)
+	if(SSxevolution)
+		SSxevolution.override_power(hivenumber, evil, override)
```
Drulikar, line 1289: Is it possible for this subsystem to not be present?

I mean the check is okay, but it seems unnecessary? I'm more concerned what situation won't have the subsystem?

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -299,7 +298,14 @@
 		hivenumber = xeno.hivenumber
 	xeno_message(SPAN_XENOANNOUNCE("The Queen has commanded the metal bird to depart for the metal hive in the sky! Rejoice!"), 3, hivenumber)
 	xeno_message(SPAN_XENOANNOUNCE("The hive swells with power! You will now steadily gain pooled larva over time."), 2, hivenumber)
-	GLOB.hive_datum[hivenumber].abandon_on_hijack()
+	var/datum/hive_status/hive = GLOB.hive_datum[hivenumber]
+	hive.abandon_on_hijack()
+	var/original_evilution = hive.evolution_bonus
+	hive.override_evilution(10, TRUE)
```
Drulikar, line 304: This should be a define somewhere rather than hard coded in this proc.

Drulikar (changes requested):

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -299,7 +298,14 @@
...
+	if(hive.living_xeno_queen)
+		var/datum/action/xeno_action/onclick/grow_ovipositor/ovi_ability = get_xeno_action_by_type(hive.living_xeno_queen, /datum/action/xeno_action/onclick/grow_ovipositor)
+		ovi_ability.reduce_cooldown(5 MINUTES)
```
Drulikar, line 307: It would be better to reduce by the cooldown amount for better future proofing. E.g. maybe ovi cooldown nerfed to 10 minutes, then this will only reduce the time an amount not reset it.
```suggestion
		ovi_ability.reduce_cooldown(ovi_ability.xeno_cooldown)
```
Untested suggestion, but should be correct.

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -299,7 +298,14 @@
...
+	addtimer(CALLBACK(hive, TYPE_PROC_REF(/datum/hive_status, override_evilution), original_evilution, FALSE), 3 MINUTES)
```
Drulikar, line 308: The 3 MINUTES should also be a define.

Drulikar (changes requested):

code/modules/cm_aliens/structures/special/pylon_core.dm
ihatethisengine: oops it was an accident

code/modules/mob/living/carbon/xenomorph/hive_status.dm
ihatethisengine: I just copypasted it from another place, not sure

code/modules/cm_aliens/structures/special/pylon_core.dm
ihatethisengine: I added messages, but I am afraid it still might be spammy. Especially because pylons provide larvae separately, so time between procs may differ significantly. Any better ideas?

ihatethisengine: I probably can add some kind of timer that will check if hit_larva_limit is still FALSE after few minutes and send the message only then... Seems kinda lame, but cannot think of a better solution

ihatethisengine: and still there can be a situation, when I send the message and the hit_larva instantly becomes true and sends another message =\\\

ihatethisengine: It's almost like I need to use one timer for both messages and queue them correctly, hmmm

Drulikar: As I mentioned, I think it would be better to be information in the hive_status rather than messages:
> I'm not really suggesting it should have a text announcement (if it does it should only do it infrequently and not be spammy); but likely would be good to think of maybe a way for this to be indicated in the hive status.

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -263,8 +263,9 @@
 		hijack(xeno)
 		return
 
+#define HIJACK_EVILUTION_BUFF 10
+#define HIJACK_EVILUTION_TIME 3 MINUTES
```
Drulikar, line 267: Move these to https://github.com/cmss13-devs/cmss13/blob/a730d2b5cbf69388d919396e7a94b03cd5dcb8e9/code/__DEFINES/xeno.dm#L171-L178 with autodoc comments to be with the rest of most tweakable settings for xenos. Or a different define file if you find one more appropriate.

Drulikar (changes requested):

code/modules/cm_aliens/structures/special/pylon_core.dm
Drulikar: E.g. here: 
image

ihatethisengine: yeah I can use tgui I code via VM. I don't think anyone would check hivestatus to know pylon status though

Drulikar: Almost every xeno uses this UI. And if theres new text near burrowed larva they will see it, because people look at that number all the time.

ihatethisengine: Alright I figure something out a bit later

Drulikar (changes requested):

code/__DEFINES/xeno.dm
```diff
@@ -169,20 +169,26 @@
...
-#define XENO_COMM_ACQUISITION_TIME (55 MINUTES)
+#define XENO_COMM_ACQUISITION_TIME (1 MINUTES)
 
 /// The time it takes for a pylon to give one larva while activated
-#define XENO_PYLON_ACTIVATION_COOLDOWN (5 MINUTES)
+#define XENO_PYLON_ACTIVATION_COOLDOWN (1 MINUTES)
 
 /// The time until you can re-corrupt a comms relay after the last pylon was destroyed
-#define XENO_PYLON_DESTRUCTION_DELAY (5 MINUTES)
+#define XENO_PYLON_DESTRUCTION_DELAY (1 MINUTES)
```
Drulikar, line 178: Revert

code/modules/cm_aliens/structures/special/pylon_core.dm
```diff
@@ -178,11 +178,12 @@
 
 	activated = TRUE
 	linked_hive.check_if_hit_larva_from_pylon_limit()
-	addtimer(CALLBACK(src, PROC_REF(give_larva)), XENO_PYLON_ACTIVATION_COOLDOWN, TIMER_UNIQUE|TIMER_OVERRIDE|TIMER_LOOP|TIMER_DELETE_ME)
+	addtimer(CALLBACK(src, PROC_REF(give_larva)), XENO_PYLON_ACTIVATION_COOLDOWN, TIMER_UNIQUE|TIMER_OVERRIDE|TIMER_LOOP|TIMER_DELETE_ME|TIMER_NO_HASH_WAIT)
```
Drulikar, line 181: Unnecessary, but doesn't hurt either. You only need to TIMER_NO_HASH_WAIT if the duration changes, and in this case it can't change.

code/modules/cm_aliens/structures/special/pylon_core.dm
```diff
@@ -178,11 +178,12 @@
...
 
-#define LARVA_ADDITION_MULTIPLIER 0.10
 /// Looped proc via timer to give larva after time
 /obj/effect/alien/resin/special/pylon/endgame/proc/give_larva()
+	xeno_announcement(SPAN_XENOANNOUNCE("Test"), linked_hive.hivenumber, XENO_GENERAL_ANNOUNCE)
+
```
Drulikar, line 186: Revert

Drulikar (changes requested):

code/__DEFINES/xeno.dm
```diff
@@ -177,6 +177,17 @@
 /// The time until you can re-corrupt a comms relay after the last pylon was destroyed
 #define XENO_PYLON_DESTRUCTION_DELAY (5 MINUTES)
 
+/// Evolution boost during hijack
+#define XENO_HIJACK_EVILUTION_BUFF 10
+
+/// For how long the buff lasts
+#define XENO_HIJACK_EVILUTION_TIME 3 MINUTES
```
fira, line 184: use ( ) around time defines to avoid people tripping up on math operators - like XENO_PYLON_DESTRUCTION_DELAY above

fira (approved):
Code LGTM

cm13-github:

cm13-github:

cm13-github:

cm13-github:

code/modules/cm_aliens/structures/special/pylon_core.dm
```diff
@@ -190,24 +188,13 @@
...
-	var/list/hive_xenos = linked_hive.totalXenos.Copy()
-
-	for(var/mob/living/carbon/xenomorph/xeno in hive_xenos)
-		if(!xeno.counts_for_slots)
-			hive_xenos -= xeno
-
-	var/real_total_xeno_count = length(hive_xenos) + linked_hive.stored_larva
-
-	if(real_total_xeno_count > (length(GLOB.alive_human_list) * ENDGAME_LARVA_CAP_MULTIPLIER))
+	if(!linked_hive.check_if_hit_larva_from_pylon_limit())
```
Drulikar, line 191: I believe you have this reversed from what you want.

Drulikar (changes requested):

Drulikar (approved):
Code changes look fine.

PR #5134, opened 2023-12-05
DEBUG: src.pulledby.pulling == src

fira:
Live debug for an elusive and rare issue invalidating this relationship, apparently causing some people to render warriors unable to keep grabs on things. Possibly affects all forms of grabbing, this debug also aims at determining that.

It's very rare and as of now we can't pinpoint the location so hoping this gives some more insight into how this is triggered to fix it...


Birdtalon (approved):
lets roll

PR #5143, opened 2023-12-06
PO can remove queens override again, bit by bit edition

ihatethisengine:
# About the pull request
PO can perform a 20 seconds do_after to remove additional 20 seconds of the lockdown.

You used to be able to remove lockdown completely if you perform a 3 minutes do_after, but it was lost during the port to new shuttle system. Now it's back in another, better form! Gives marines a chance to remove lockdown sooner if they control the cockpit. In best case scenario you can remove the lock in 5 minutes.

<details>

</details>

# Changelog
:cl: ihatethisengine
balance: PO can remove queens dropship override again.
/:cl:

### DIFF
--- a/code/modules/shuttle/computers/dropship_computer.dm
+++ b/code/modules/shuttle/computers/dropship_computer.dm
@@ -146,7 +146,7 @@
 
 	if(dropship_control_lost)
 		var/remaining_time = timeleft(door_control_cooldown) / 10
-		to_chat(user, SPAN_WARNING("The shuttle is not responding due to an unauthorazed access attempt, the system will automatically remove the lockout in about [remaining_time] seconds."))
+		to_chat(user, SPAN_WARNING("The shuttle is not responding due to an unauthorazed access attempt, a large lockout timers reads that the lockout will be automatically removed in [remaining_time] seconds."))
 		if(!skillcheck(user, SKILL_PILOT, SKILL_PILOT_EXPERT))
 			return
 		if(user.action_busy || override_being_removed)
@@ -163,7 +163,7 @@
 				break
 			remaining_time = timeleft(door_control_cooldown) / 10 - 20
 			if(remaining_time > 0)
-				to_chat(user, SPAN_NOTICE("You partly remove the lockout, about [remaining_time] seconds left."))
+				to_chat(user, SPAN_NOTICE("You partly remove the lockout, only [remaining_time] seconds left."))
 				door_control_cooldown = addtimer(CALLBACK(src, PROC_REF(remove_door_lock)), remaining_time SECONDS, TIMER_STOPPABLE|TIMER_UNIQUE|TIMER_OVERRIDE|TIMER_NO_HASH_WAIT)
 	override_being_removed = FALSE
 	if(dropship_control_lost)
--- a/code/modules/shuttle/computers/dropship_computer.dm
+++ b/code/modules/shuttle/computers/dropship_computer.dm
@@ -142,22 +142,32 @@
 	if(shuttle.mode == SHUTTLE_CRASHED)
 		to_chat(user, SPAN_NOTICE("\The [src] is not responsive"))
 		return
-	var/remaining_time = timeleft(door_control_cooldown) / 10
-	to_chat(user, SPAN_WARNING("The shuttle is not responding, try again in [remaining_time] seconds."))
-	if(dropship_control_lost && skillcheck(user, SKILL_PILOT, SKILL_PILOT_EXPERT))
-		to_chat(user, SPAN_NOTICE("You start to remove the Queens override."))
-		if(do_after(user, 20 SECONDS, INTERRUPT_ALL, BUSY_ICON_HOSTILE))
+
+	if(dropship_control_lost)
+		var/remaining_time = timeleft(door_control_cooldown) / 10
+		to_chat(user, SPAN_WARNING("The shuttle is not responding due to the Queen's override, the system will automatically remove the override in about [remaining_time] seconds."))
+		if(!skillcheck(user, SKILL_PILOT, SKILL_PILOT_EXPERT))
+			return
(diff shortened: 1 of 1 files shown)


ihatethisengine:
Need help with the timer, it isn't deleted properly for some reason even if I do

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -142,16 +142,22 @@
 	if(shuttle.mode == SHUTTLE_CRASHED)
 		to_chat(user, SPAN_NOTICE("\The [src] is not responsive"))
 		return
-
+	var/remaining_time = timeleft(door_control_cooldown) / 10
+	to_chat(user, SPAN_WARNING("The shuttle is not responding, try again in [remaining_time] seconds."))
```
Drulikar, line 146: At the moment this message sort of suggests we can't do anything about it. If it was done in character more it probably wouldn't have this effect; but I'm not sure what would be automatically done to allow it to fix itself (that we can speed up).

Also the estimates should probably qualified with `about` or `around` even if they actually are totally accurate timer amounts because it would be weird to know precisely how many seconds you need to clear off goop.

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -142,16 +142,22 @@
...
 	if(dropship_control_lost && skillcheck(user, SKILL_PILOT, SKILL_PILOT_EXPERT))
-		var/remaining_time = timeleft(door_control_cooldown) / 10
-		if(remaining_time > 60)
-			to_chat(user, SPAN_WARNING("The shuttle is not responding, try again in [remaining_time] seconds."))
-			return
```
Drulikar, line 150: Shouldn't the mentions of Queens override be `Queen's override` (if we don't just call it something else)

Drulikar (changes requested):
This is a good approach to basically put it in milestones to chip away at it so it's not super detrimental to get interrupted; but I think you ought to have it repeat the do_after again much like filling sandbags or clearing snow. Likely as simple as adding a while loop we break out of if interrupted but double check those implementations.

code/modules/shuttle/computers/dropship_computer.dm
ihatethisengine: Have any suggestions?

code/modules/shuttle/computers/dropship_computer.dm
ihatethisengine: I think it's better now. Also I made a for cycle instead of notorious while(true)

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -143,15 +144,33 @@
...
+				return
+			if(!dropship_control_lost)
+				to_chat(user, SPAN_NOTICE("The Queen's override is already removed."))
+				break
+			remaining_time = timeleft(door_control_cooldown) / 10 - 20
+			if(remaining_time > 0)
+				to_chat(user, SPAN_NOTICE("You partly remove the Queen's override, about [remaining_time] seconds left."))
+				door_control_cooldown = addtimer(CALLBACK(src, PROC_REF(remove_door_lock)), remaining_time SECONDS, TIMER_STOPPABLE|TIMER_UNIQUE|TIMER_OVERRIDE|TIMER_NO_HASH_WAIT)
+			else
+				break
```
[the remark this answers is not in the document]
ihatethisengine: > will be re-checked each iteration unlike that for loop as far as I understand.

My point was you don't really need to recheck it every time, but I think your solution is better anyway.

code/modules/shuttle/computers/dropship_computer.dm
ihatethisengine: What do you think now?

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -143,15 +144,31 @@
 		to_chat(user, SPAN_NOTICE("\The [src] is not responsive"))
 		return
 
-	if(dropship_control_lost && skillcheck(user, SKILL_PILOT, SKILL_PILOT_EXPERT))
+	if(dropship_control_lost)
 		var/remaining_time = timeleft(door_control_cooldown) / 10
-		if(remaining_time > 60)
-			to_chat(user, SPAN_WARNING("The shuttle is not responding, try again in [remaining_time] seconds."))
+		to_chat(user, SPAN_WARNING("The shuttle is not responding due to an unauthorazed access attempt, a large lockout timers reads that the lockout will be automatically removed in [remaining_time] seconds."))
```
Drulikar, line 149:
```suggestion
		to_chat(user, SPAN_WARNING("The shuttle is not responding due to an unauthorized access attempt. In large text it says the lockout will be automatically removed in [remaining_time] seconds."))
```
Spelling/grammar/conciseness.

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -143,15 +144,31 @@
...
+			if(!do_after(user, 20 SECONDS, INTERRUPT_ALL, BUSY_ICON_HOSTILE))
+				to_chat(user, SPAN_WARNING("You fail to remove the lockout!"))
+				override_being_removed = FALSE
+				return
+			if(!dropship_control_lost)
+				to_chat(user, SPAN_NOTICE("The lockout is already removed."))
+				break
+			remaining_time = timeleft(door_control_cooldown) / 10 - 20
+			if(remaining_time > 0)
+				to_chat(user, SPAN_NOTICE("You partly remove the lockout, only [remaining_time] seconds left."))
```
Drulikar, line 166:
```suggestion
				to_chat(user, SPAN_NOTICE("You partially bypass the lockout, only [remaining_time] seconds left."))
```

Drulikar (changes requested):

PR #5147, opened 2023-12-06
smartpistol buff, removes AP gives higher damage

irakligabunia:
# About the pull request

increases raw damage from the smartpistol, lowers its armor piercing value

as it currently stands, su-6 is a pistol meant for backliners to protect their patients or people they are defending. 
it has way too low damage to handle frontline castes but enough armor piercing for it, if it had higher damage
it has way too low damage to handle backline castes, but armor piercing is useless for it
i take it as: 
MOD88 should be used for AP, frontline
M4A3 is a good inbetween, hollowpoint and standard magazines makes it a specialized weapon
VP78 is in a rough spot, but as of right now it is a close range, frontline, armor piercing pistol

smartpistol does not have any place, i want to see if this will help the people using it without severely impacting the game

its a number change

</details>

# Changelog

:cl: stalkerino
balance:  gives smartpistol 40 dmg, lowers its AP to 5
/:cl:

### DIFF
--- a/code/datums/ammo/bullet/pistol.dm
+++ b/code/datums/ammo/bullet/pistol.dm
@@ -259,7 +259,7 @@
 	flags_ammo_behavior = AMMO_BALLISTIC
 
 	accuracy = HIT_ACCURACY_TIER_8
-	damage = 30
-	penetration = 20
+	damage = 40
+	penetration = 5
 	shrapnel_chance = SHRAPNEL_CHANCE_TIER_2
 


Blundir:
this isn't a buff at all...

Zonespace27:
Tried it out on a local, seems alright overall

Zonespace27:
Thank you for the pull request, but I will be closing this for the following reason:
- "as it currently stands, su-6 is a pistol meant for backliners to protect their patients or people they are defending." I overall disagree with this statement. Friendly fire is far less an issue in the backline, and many other IFF-less guns will do just as good a job or better. The place of this gun, like most IFF guns, is in the middle or in the back of the marine frontline ungaball. This change will make runners and similar t1s be further shredded by standing too close to the frontline, without carving out a proper niche for itself.

Issue #5103, opened 2023-12-02
Warrior grab being dropped after lunge


Birdtalon:
### Testmerges

#4986 #4939 #5054 #4842 #5067 #5073 #5022 #5018 #4077 #5100

### Round ID

19770

### Description of the bug

Warrior lunge drops every grab for no apparent reason. Admin PM'd another warrior in this round and they did not have the same issue so it appears to be localised to the mob.

YT link as video too large for github
https://youtu.be/i2tHPa13RsQ

### What's the difference with what should have happened?

The marine should not be able to get up and run away instantly.

### How do we reproduce this bug?

Not sure what causes it but after that you can just lunge a marine and they will get up instantly. This has happened to me across multiple rounds, but not with any pattern or consistency.

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #5151, opened 2023-12-06
Fixes Warrior Pulls breaking in future when lunging a M56D/M2C user from front

fira:
# About the pull request

So that one took a lot of debugging...

To keep it as short as possible, when you lunge a M56D/M2C user from front as Warrior, they get stunned. The stun cancels their HMG interaction, causing them to step away. This means they leave the grab range, breaking it. Warrior code incorrectly re-sets pulledby, resulting in a mob "pulledby" one that doesn't pull it.

In practice once this happens, this means that the "incorrectly" pulled mob will reset any active grabs from the Warrior when it moves. Negating the use of lunges completely.

# Changelog

:cl:
fix: Fixed a logic error in Warrior code causing lunging some M2C/M56D users to semi-permanently brick lunge.
/:cl:


Birdtalon:
Fixes #5103

PR #5154, opened 2023-12-07
Oppressor cooldown changes

InsaneRed:
# About the pull request

Lowers the oppressor tail_abudct (the hook) to 15 seconds of cooldown and makes the windup faster.

Makes punch shave off cooldown from the abduct for 5 seconds 

All have been tested but i would like this to get testmerged first so i can actually see the results in game, nothing is set in stone and i want to edit this further so the cd / cd reduction isnt too powerful, they're just numbers ive decided were good enough to atleast make the caste decent for the time being.

Oppressor has been a snoozer strain for a while now where you cast an ability, and IF it hits you get to play the game otherwise you wait 20 seconds and thats just not fun. Especially for what the ability is, a 20 second cooldown is not worth it.

 I've talked with a few people that all agree that the downtime for what you "could" do with oppressor is not worth it. And i have to agree with them, the caste feels boring to play and its basically half dead due to the amnout of downtime you have between abilities compared to how everything else works. The idea of this is to make it so its not busted out of its brain but atleast not an observer++ strain so you can feel more involved in the gameplay.

</details>

# Changelog

:cl:
balance: Oppressor tail abduct changed to 15 seconds and lowers the windup to 7 deciseconds
balance:  Changes around the punch effect so that instead of having to meet demonic standards, you only need to punch to lower your tail/hook on oppressor.
fix: You will now automatically punch chest if the target you are aiming at is delimbed instead of doing nothing
/:cl:


code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_powers.dm
```diff
@@ -412,19 +412,18 @@
...
-			tail_lash_action.reduce_cooldown(5 SECONDS)
 	else
 		target_carbon.apply_armoured_damage(get_xeno_damage_slash(target_carbon, damage), ARMOR_MELEE, BRUTE, target_limb? target_limb.name : "chest")
 		step_away(target_carbon, oppressor_user, 2)
 
 
 	shake_camera(target_carbon, 2, 1)
+	var/datum/action/xeno_action/activable/prae_abduct/abduct_action = get_xeno_action_by_type(oppressor_user, /datum/action/xeno_action/activable/prae_abduct)
+	var/datum/action/xeno_action/activable/tail_lash/tail_lash_action = get_xeno_action_by_type(oppressor_user, /datum/action/xeno_action/activable/tail_lash)
+	if(abduct_action && abduct_action.action_cooldown_check())
```
Birdtalon, line 423: `if(abduct_action && !abduct_action.action_cooldown_check())`

code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_powers.dm
```diff
@@ -412,19 +412,18 @@
...
+		abduct_action.reduce_cooldown(5 SECONDS)
+	if(tail_lash_action && tail_lash_action.action_cooldown_check())
```
Birdtalon, line 425: `if(tail_lash_action && !tail_lash_action.action_cooldown_check())`

Birdtalon (changes requested):

Birdtalon (approved):
Will need balance approval. But I think a good change to make Oppressor have much less downtime throughout the game and make missing a hook less punishing.

code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_abilities.dm
```diff
@@ -169,7 +169,7 @@
 
 	// Config
 	var/fling_dist = 3
-	var/windup = 2
```
Zonespace27, line 172: Revert this and it's good to go for a TM

Zonespace27 (changes requested):

Zonespace27 (approved):
been fine from what i've seen

PR #5156, opened 2023-12-07
Zerker cannot use eviscerate below 5 rage

ihatethisengine:
# About the pull request
Zerker cannot use eviscerate below 5 rage

While the idea of an ability scaling with your rage sounds pretty cool, in practice there is never a reason to use eviscerate below 5 rage. It's just suboptimal. You just put yourself into more risk standing still for two seconds and gain basically nothing, while in these 2 seconds you could easily get enough stacks to actually do something. The only reason people use eviscerate below 5 rage is misclick. 

<details>
I tested it fr fr

</details>

# Changelog
:cl: ihatethisengine
qol: berserker ravager can use eviscerate only at maximum rage stacks.
balance: berserker eviscerate now has a flat 2 range, flat 70 damage, and flat wind-up speed of 1 second.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/ravager/ravager_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/ravager/ravager_powers.dm
@@ -362,57 +362,49 @@
 	if(!xeno.check_state())
 		return
 
-	var/damage = base_damage
-	var/range = 1
-	var/windup_reduction = 0
+	if(xeno.mutation_type != RAVAGER_BERSERKER)
+		return
+
 	var/lifesteal_per_marine = 50
 	var/max_lifesteal = 250
 	var/lifesteal_range =  1
 
-	if (xeno.mutation_type == RAVAGER_BERSERKER)
-		var/datum/behavior_delegate/ravager_berserker/behavior = xeno.behavior_delegate
-		if (behavior.rage == 0)
-			to_chat(xeno, SPAN_XENODANGER("You cannot eviscerate when you have 0 rage!"))
-			return
-		damage = damage_at_rage_levels[Clamp(behavior.rage, 1, behavior.max_rage)]
-		range = range_at_rage_levels[Clamp(behavior.rage, 1, behavior.max_rage)]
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/abilities/ravager/ravager_powers.dm
```diff
@@ -362,57 +362,49 @@
...
 
 	ADD_TRAIT(xeno, TRAIT_IMMOBILIZED, TRAIT_SOURCE_ABILITY("Eviscerate"))
 	xeno.anchored = TRUE
 
-	if (do_after(xeno, (activation_delay - windup_reduction), INTERRUPT_ALL | BEHAVIOR_IMMOBILE, BUSY_ICON_HOSTILE))
-		xeno.emote("roar")
-		xeno.spin_circle()
+	if(!do_after(xeno, activation_delay, INTERRUPT_ALL|BEHAVIOR_IMMOBILE, BUSY_ICON_HOSTILE))
+		xeno.anchored = FALSE //sanity check
+		return ..()
```
ihatethisengine, line 388: here

usnpeepoo:
just because it's sub-optimal doesn't mean it has no use, the lifesteal mechanic with eviscerate still works on less rage evisc's.  

If the berserker would want the health gain faster they could do so instead of being forced to be at max rage and potentially being in a bad position with rage lock out (from being max rage) this pr would remove those potential options.

ihatethisengine:
While technically true, realistically you won't want to stay still for whole two seconds if you really need that life steal. This may only be viable in some extremely rare situations where a few marines are somehow unable to either shoot you or move away within two seconds and you want that healing so badly for some reason and fine with dealing so tiny damage with the ability. I believe we can just disregard these extremely rare situations for the sake of many cases when people used eviscerate below max rage due to some mistake and got killed. 

Also, frankly, I think you can heal more from slashes in these two seconds than you can heal from eviscerate even if you somehow got 5 marines (for max heal) standing still and doing nothing near you for 2 seconds.

cm13-github:

Zonespace27:
I don't think this is necessary currently. Thank you for the PR.

PR #5158, opened 2023-12-07
fixes is_interior(), fixing xeno construction on dropships during transport

harryob:
huh? how'd this not get reported?

:cl:
fix: xenos can construct in dropships during transport again
/:cl:

### DIFF
--- a/code/controllers/subsystem/interior.dm
+++ b/code/controllers/subsystem/interior.dm
@@ -51,10 +51,16 @@ SUBSYSTEM_DEF(interior)
 	if(!isturf(loc))
 		loc = get_turf(loc)
 
-	var/datum/weakref/reservation = SSmapping.used_turfs[loc]
+	var/datum/weakref/reservation_weakref = SSmapping.used_turfs[loc]
+
+	if(!reservation_weakref)
+		return
+
+	var/datum/turf_reservation/interior/reservation = reservation_weakref.resolve()
 
 	if(!istype(reservation))
 		return FALSE
+
 	return TRUE
 
 #undef INTERIOR_BORDER_SIZE


SabreML:
Wow, that was fast.

Doubleumc:
If this does what I think it does (reserved-but-not-interior turfs don't inherently count as interior) you can revert part of #4815 https://github.com/cmss13-devs/cmss13/pull/4815/files#diff-150200d3d67a0db8d052db8dfcfa37b39396e514b3bef9d708dcce5d7e1bdd7aL45-R45

Drulikar:
> If this does what I think it does (reserved-but-not-interior turfs don't inherently count as interior) you can revert part of #4815 https://github.com/cmss13-devs/cmss13/pull/4815/files#diff-150200d3d67a0db8d052db8dfcfa37b39396e514b3bef9d708dcce5d7e1bdd7aL45-R45

I don't think there's currently any change there. if `.` was assigned it would matter; but the default return value should already be 0 aka FALSE

Doubleumc:
> I don't think there's currently any change there. if `.` was assigned it would matter; but the default return value should already be 0 aka FALSE; though maybe its actually null - either case it should be falsey - just not sure what it returns for isnull.

It makes a difference because the return (`VI`) is subjected to a null check, which 0/FALSE would pass.

https://github.com/cmss13-devs/cmss13/blob/98f032b7d7434f4a06edbf5f2d59b0b877440138/code/game/sound.dm#L82-L84

Also described here: https://github.com/cmss13-devs/cmss13/pull/4815#issuecomment-1789698825

If reserved-but-not-interior turfs counting as interior is undesired, reverting this change would cause exceptions to be thrown if one managed to be created again.

PR #5164, opened 2023-12-08
Projectile ref clears

Doubleumc:
# About the pull request

Forgot to clear a reference I set in https://github.com/cmss13-devs/cmss13/pull/4986 and cleared some others I saw.

Not familiar enough with the Destroy/QDEL pipeline to know what else, if anything, should be done with `weapon_cause_data` and `bullet_traits` since they hold references themselves.

I will clean up after myself.
<details>

</details>

# Changelog
No player-facing changes.

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -96,7 +96,10 @@
 	starting = null
 	permutated = null
 	path = null
+	vis_source = null
+	process_start_turf = null
 	weapon_cause_data = null
+	bullet_traits = null
 	firer = null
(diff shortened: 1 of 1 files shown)


Doubleumc:
> weapon_cause_data can be discarded without qdel i'm not really seeing where bullet_traits have refs, do you have a specific location in mind?

My understanding was `bullet_traits` were components, which do a lot of things but at the very least contained a `parent` ref to the holding projectile. Not familiar enough to know if that is handled gracefully - *probably* fine as-is since its been working without this change for a long time.

fira (approved):

PR #5174, opened 2023-12-09
Vanguard tweaks

InsaneRed:
# About the pull request
This catches up vanguard to current gameplay as it was last changed approximately 4 years ago

Currently Vanguard is supposed to be a frontlining tier 3 who dashes in > stays in and gets some damage in and goes out  (thats why the shield is a thing) and you're supposed to be able to stay there because your abilities (pierce and dash) resets your shield. But with the recent addition of just more damage in general be it m56d, full auto mode or just the amnout of extra damage you can receive from the front compared 4 years ago.

The strain currently struggles and is near useless other then the occasional go in and lucky fling.

I've changed up the dash to reset your shield once you hit ONE person, rather then two so that you dont instantly die while going in, the shield is very situational as it will most likely decay before you can even go in.

Listening to people who play vanguard, i've also increased the root to 2.5 seconds (this is buffed so when the prae has the shield) the marine can still shoot back when they're rooted so i dont think the duration is a big problem (this is going to be a test merge so i will be watching)

it just works

</details>

# Changelog

:cl:
balance: Vanguard dash now restores your shield if you hit ANYONE instead of 2 people.
balance: Vanguard buffed root now roots you for 2.5 seconds, unbuffed for 1 second
qol: Vanguard's pierce has now a hit sound for better feedback
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_abilities.dm
@@ -60,8 +60,6 @@
 	// Fling config
 	var/fling_dist_unbuffed = 3
 	var/fling_dist_buffed = 6
-	var/stun_power = 0.5
-	var/weaken_power = 0.5
 
 	// Root or do a punch-like-effect.
 	var/root_toggle = TRUE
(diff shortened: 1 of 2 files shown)


Zonespace27 (approved):
Seems fine.

PR #5175, opened 2023-12-09
Xeno Throwing Refactor, prevents canceling mob throws by moving

fira:
# About the pull request

This refactors xeno throw procs to be less of a pain to work with, and introduces automatic immobilizing during some xeno throws. They end after the throw using throw callbacks.

This is important because if you are not immobilized during the throw, you can just move and step from your current location, canceling it - mostly visible on flings and vanguard cleave

Xeno throws that apply a stun would already prevent moving.

Because they were only used by Xeno Pounce and Crusher Tumble, i don't know how robust the callbacks are. It's worth Testmerging.

More reliable throws, throw procs which are easier to work with for new contributors by baking in the immobilization

Done baseline testing with things such as Runner Pounce and Vanguard Cleave. Not all abiltiies are affected because they used different handlers, such as Warrior Fling.

# Changelog
:cl:
fix: Some Xeno throws now immobilize their targets, ensuring they do not walk out of the toss mid-flight.
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -415,6 +415,8 @@ GLOBAL_LIST(trait_name_map)
 #define TRAIT_SOURCE_XENO_ACTION_CHARGE "t_s_xeno_action_charge"
 ///Status trait coming from a xeno nest
 #define XENO_NEST_TRAIT "xeno_nest"
+///Status trait from a generic throw by xeno abilities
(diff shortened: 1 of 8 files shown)


Birdtalon:
I will likely refactor warrior fling to use this after it's merged.

Birdtalon (changes requested):
Makes sense to me, could you just add a simple code doc for the procs you created?

Birdtalon (approved):

PR #5188, opened 2023-12-11
Standartize the price of all three binculars type for SG,medic,comtech, IO

Huffie56:
# About the pull request
The logic for standartizing it go as fallow:
no reason to not make lower tier of bino available if you have access to higher tier of it.
no reason to not give access to higher tier if lower ranks have access to it.

concerning the price change for medic i found it weird that it was the only one with this price tag so i figure better smooth things up. and their is no reason for medic to need higher tier binocular as they role isn't to do JTAC...

Give access to more items while keeping it logical 
Giving people a cheaper option than higher tier one
also those bino are free in the prep vendor and have a few public spot where they spawn.
<details>

</details>

# Changelog
:cl:
balance: For SG,IO,medic,comtech standartize the binoculars section as fallow (Binoculars=5, Range Finder =10, Laser Designator=15).
/:cl:


JackieEstegado:
Keep in mind, Medic and CT have jtac skill as they are intended to perform jtac if sl or ftl is not available.

Zonespace27 (approved):

PR #5198, opened 2023-12-12
Compass Headgear now shows your direction when examined

Contrabang:
# About the pull request

When the compass headgear is now examined, it will tell you the direction your character is facing (north, south, east, or west). This only works when you're on a planet, this means being groundside and not on a map that is situated in space.

I've considered changing the description, but I've left it because I couldn't think of a better one.

When I used the compass in game, I expected to actually tell me which direction I'm facing, but I was a bit disappointed when it didn't. So, I change it to work in a realistic way.

<details>

On LV-624, where the compass works (left). On the ship, were the compass does not work (right).
image

On maps like Fiorina where are situated in space, the compass will not work.
image

</details>

# Changelog
:cl:
add: The compass headgear will now tell you what direction you're facing when on a planet.
code: Added a new ground environmental trait for a ground level in space.
/:cl:
My discord is: contrabang, if you would like to contact me, that is the best place.

### DIFF
--- a/code/__DEFINES/maps.dm
+++ b/code/__DEFINES/maps.dm
@@ -32,6 +32,7 @@ require only minor tweaks.
 #define ZTRAIT_FOG "Fog"
 #define ZTRAIT_LOCKDOWN "Lockdown"
 #define ZTRAIT_BASIC_RT "BasicRT"
+#define ZTRAIT_IN_SPACE "InSpace" // Is our ground_level considered in space or on a space station
 
 // boolean - weather types that occur on the level
 #define ZTRAIT_SNOWSTORM "weather_snowstorm"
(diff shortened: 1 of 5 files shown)


Contrabang:
~~I forgot to change the title, embarrassing~~

BadAtThisGame302:
Shouldn't you also change the desc of the compass? "It always faces north"... well now it won't always face 'north'.

Contrabang:
> Shouldn't you also change the desc of the compass? "It always faces north"... well now it won't always face 'north'.

I mean, the needle of a compass always does point north, that's how you determine your rotation. Its a bit of a silly description, but I can't think of a better one, so I left it

harryob (approved):

BottleBook:
>Its a bit of a silly description, but I can't think of a better one
What about "It always points towards the north"?

PR #5202, opened 2023-12-12
Adds a way for SLs/SOs/XOs to recommend medals that the CO can then approve/deny.

TheGamerdk:
# About the pull request

See title

Will likely make medals more used, and means the CO doesn't have to scramble to get medal recommendations in the last 3 minutes.

<details>

How to give recommendations:
image
OR
image

The actual panel for COs:
image

</details>

# Changelog

:cl:
add: SLs, SOs and XOs can now recommend people for medals! Do this by right-clicking your headset, or using the new button in the object panel of the Status Panel!
/:cl:


code/datums/medal_awards.dm
```diff
@@ -366,3 +463,185 @@ GLOBAL_LIST_INIT(xeno_medals, list(XENO_SLAUGHTER_MEDAL, XENO_RESILIENCE_MEDAL,
...
+			var/choice = tgui_alert(user, "Would you like to change the medal text?", "Medal Citation", list("Yes", "No"))
+			var/medal_citation = recommendation.reason
+			if(choice == "Yes")
+				medal_citation = strip_html(input("What should the medal citation read?", "Medal Citation", null, null) as message|null, MAX_PAPER_MESSAGE_LEN)
+
+			var/confirm_choice = tgui_alert(user, "Are you sure you want to give a medal to [recommendation.recipient_name]?", "Medal Confirmation", list("Yes", "No"))
+			if(confirm_choice != "Yes")
+				return
+			var/atom/user_machine = user_locs[user]
+			if(give_medal_award_prefilled(get_turf(user_machine), user, recommendation.recipient_name, recommendation.recipient_rank, recommendation.recipient_ckey, medal_citation, medal_type, recommendation.recommended_by_ckey, recommendation.recommended_by_name))
```
Drulikar, line 624: Using the turf makes it so `The floor prints a medal.` and it will print it out wherever you started the process - not where you end up. It should just be the tablet passed and then when printed it can print to the turf where the tablet is located.

code/datums/medal_awards.dm
```diff
@@ -366,3 +463,185 @@ GLOBAL_LIST_INIT(xeno_medals, list(XENO_SLAUGHTER_MEDAL, XENO_RESILIENCE_MEDAL,
...
+				GLOB.medal_recommendations -= recommendation
+				qdel(recommendation)
+				user_machine.visible_message(SPAN_NOTICE("[user_machine] prints a medal."))
+				. = TRUE
+
+		if("deny_medal")
```
Drulikar, line 630: This should either have a confirmation OR you use `Button.Confirm` in the TGUI. Its a rather small button that you can easily misclick next to the approve/edit button.

code/datums/medal_awards.dm
```diff
@@ -366,3 +463,185 @@ GLOBAL_LIST_INIT(xeno_medals, list(XENO_SLAUGHTER_MEDAL, XENO_RESILIENCE_MEDAL,
...
+		if(recipient_name == recommendation_giver.real_name)
+			continue
+		recipient_ranks[recipient_name] = record.fields["rank"]
+		possible_recipients += recipient_name
+	var/chosen_recipient = tgui_input_list(recommendation_giver, "Who do you want to recommend a medal for?", "Medal Recommendation", possible_recipients)
+	if(!chosen_recipient)
+		return FALSE
+
+	// Write a citation
+	var/reason = strip_html(input("Why does this person deserve a medal?", "Medal Recommendation", null, null) as message|null, MAX_PAPER_MESSAGE_LEN)
```
Drulikar, line 492: Ideally TGUI input is used instead rather than these old prompts.

code/datums/medal_awards.dm
```diff
@@ -366,3 +463,185 @@ GLOBAL_LIST_INIT(xeno_medals, list(XENO_SLAUGHTER_MEDAL, XENO_RESILIENCE_MEDAL,
...
+				return
+			if(!user_locs[user])
+				return
+			if(recommendation.recipient_name == user.real_name)
+				return
+
+			var/choice = tgui_alert(user, "Would you like to change the medal text?", "Medal Citation", list("Yes", "No"))
+			var/medal_citation = recommendation.reason
+			if(choice == "Yes")
+				medal_citation = strip_html(input("What should the medal citation read?", "Medal Citation", null, null) as message|null, MAX_PAPER_MESSAGE_LEN)
```
Drulikar, line 618: Ideally TGUI input is used instead rather than these old prompts.

code/datums/medal_awards.dm
```diff
@@ -366,3 +463,185 @@ GLOBAL_LIST_INIT(xeno_medals, list(XENO_SLAUGHTER_MEDAL, XENO_RESILIENCE_MEDAL,
...
+			var/medal_type = params["medal_type"]
+			if(!(medal_type in GLOB.human_medals))
+				return
+			var/datum/medal_recommendation/recommendation = locate(recommendation_ref) in GLOB.medal_recommendations
+			if(!recommendation)
+				return
+			if(!user_locs[user])
+				return
+			if(recommendation.recipient_name == user.real_name)
+				return
```
Drulikar, line 613: They will probably figure out why this doesn't work at some point, but you should to_chat a message why this doesn't work. As is it just makes the UI seem like its bugged because it doesn't do anything when you press the button.

Drulikar (changes requested):

code/datums/medal_awards.dm
TheGamerdk: This is a sanity check. It shouldn't be a

code/datums/medal_awards.dm
TheGamerdk: give_medal_award_prefilled takes the loc where the medal is placed.
The message is just below (without the use of get_turf)

code/datums/medal_awards.dm
TheGamerdk: It should remove their whitelist instead

code/game/machinery/computer/communications.dm
```diff
@@ -130,7 +130,7 @@
 				cooldown_message = world.time
 
 		if("award")
-			print_medal(usr, src)
+			open_medal_panel(usr, src)
```
Drulikar, line 133: Note: As far as I can tell this is impossible to access atm because the only communications computer on the almayer is in  the secure command hallway and that doesn't satisfy the `if(!istype(loc.loc, /area/almayer/command/cic)) //Has to be in the CIC. Can also be a generic CIC area to communicate, if wanted.` test.

code/game/machinery/computer/groundside_operations.dm
```diff
@@ -235,7 +235,7 @@
 			log_announcement("[key_name(usr)] has announced the following: [input]")
 
 		if("award")
-			print_medal(usr, src)
+			open_medal_panel(usr, src)
```
Drulikar, line 238: Note: As far as I can tell this is impossible to access atm because the interface no longer has "award" as an option in its `ui_interact`.

code/datums/medal_awards.dm
```diff
@@ -366,3 +463,189 @@ GLOBAL_LIST_INIT(xeno_medals, list(XENO_SLAUGHTER_MEDAL, XENO_RESILIENCE_MEDAL,
...
+		user.visible_message("ERROR: ID card not registered for [user.real_name] in USCM registry. Potential medal fraud detected.")
+		return
+
+	switch(action)
+		if("grant_new_medal")
+			if(!user_locs[user])
+				return
+			if(give_medal_award(get_turf(user_locs[user])))
+				var/atom/user_machine = user_locs[user]
+				user_machine.visible_message(SPAN_NOTICE("[user_machine] prints a medal."))
```
Drulikar, line 599: Note: Tablet in hand won't ever display this message because visible_message doesn't handle items in mob contents; but I wouldn't worry about it.

code/datums/medal_awards.dm
```diff
@@ -366,3 +463,189 @@ GLOBAL_LIST_INIT(xeno_medals, list(XENO_SLAUGHTER_MEDAL, XENO_RESILIENCE_MEDAL,
...
+
+/datum/ic_medal_panel/tgui_interact(mob/user, datum/tgui/ui)
+	ui = SStgui.try_update_ui(user, src, ui)
+	if(!ui)
+		ui = new(user, src, "IcMedalsPanel", "Medals Panel")
+		ui.open()
+		ui.set_autoupdate(FALSE)
+
+/datum/ic_medal_panel/ui_state(mob/user)
+	return GLOB.not_incapacitated_state
```
Drulikar, line 550: At the moment you can use the tablet or computer, drop it down (in case of tablet), and run as far as way from it as you like and still interact with it.

I would suggest using a mixture of maybe not_incapacitated_and_adjacent_state and inventory_state, however as a global datum this panel has no loc (unlike the cic tablet where its ui is part of the device).

An alternative would be to implement `/datum/ic_medal_panel/ui_status/(mob/user)` but you would still need to communicate what item or machine opened this interface and test that.

So it may be best when the UI that opened this UI closes it also attempts to close any ic_medal_panel it opened.

code/datums/medal_awards.dm
```diff
@@ -632,6 +633,9 @@ GLOBAL_DATUM_INIT(ic_medals_panel, /datum/ic_medal_panel, new)
 			var/datum/medal_recommendation/recommendation = locate(recommendation_ref) in GLOB.medal_recommendations
 			if(!recommendation)
 				return
+			var/confirm = tgui_alert(user, "Are you sure you want to give a deny this medal recommendation?", "Medal Confirmation", list("Yes", "No"))
```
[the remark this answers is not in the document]
TheGamerdk: no this make complete sense

code/game/machinery/computer/communications.dm
TheGamerdk: this is just a Ctrl-F replace. I can change it if you want, but seems sorta out-of-scope

code/game/machinery/computer/groundside_operations.dm
TheGamerdk: see above

code/datums/medal_awards.dm
TheGamerdk: Fixed

code/game/machinery/computer/communications.dm
Drulikar: Its just a note; I'm not sure why it is how it is and why literally no one has ever reported this problem before.

code/datums/medal_awards.dm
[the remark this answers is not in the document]
Drulikar: Uh; maybe read it aloud.

Corrected examples:
Are you sure you want to deny this medal recommendation?
Are you sure you want to give this medal recommendation a denial?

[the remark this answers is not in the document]
TheGamerdk: This was a joke.

code/datums/medal_awards.dm
TheGamerdk: Fixed

code/datums/medal_awards.dm
Drulikar: The change here didn't resolve it for the tablet, but still works for machines; so its whatever.

Drulikar:
The `ui_host` implementation is still acceptable, but IMO worse than when originally approved.

code/datums/medal_awards.dm
```diff
@@ -160,7 +161,102 @@ GLOBAL_LIST_INIT(human_medals, list(MARINE_CONDUCT_MEDAL, MARINE_BRONZE_HEART_ME
...
+	var/citation = strip_html(reason)
+	if(!citation)
+		return FALSE
+
+	// Get mob information
+	var/recipient_rank = _recipient_rank
+	var/posthumous = TRUE
+	var/recipient_ckey = _recipient_ckey
+	var/mob/recipient_mob
+	var/found_other = FALSE
```
fira, line 189: why use intermediate variables? you can use the args directly (and wihtout a _)

code/datums/medal_awards.dm
```diff
@@ -180,8 +276,9 @@ GLOBAL_LIST_INIT(human_medals, list(MARINE_CONDUCT_MEDAL, MARINE_BRONZE_HEART_ME
 		user.visible_message("ERROR: ID card not registered for [user.real_name] in USCM registry. Potential medal fraud detected.")
 		return
 
-	if(give_medal_award(get_turf(printer)))
-		user.visible_message(SPAN_NOTICE("[printer] prints a medal."))
+	GLOB.ic_medals_panel.user_locs[user] = printer
```
fira, line 279: You can't do that, that's DOUBLE reference fraud in a single line
Need to either setup deletion handlers or use weak_reference instead

fira (changes requested):

code/datums/medal_awards.dm
TheGamerdk: because I copy pasted everything. will change

Grithok:
If the ASO, a staff officer, tries to give a medal out, they receive this message.

'Only Staff Officers, Executive Officers and Squad Leader are permitted to give medal recommendations!'

ItsVyzo:
Perhaps Expand to include department heads, say a doctor or CT/MT goes above and beyond

code/game/objects/items/devices/radio/headset.dm
```diff
@@ -405,6 +405,24 @@
...
+	set src in usr
+
+	var/mob/living/carbon/human/wearer = usr
+	if(!istype(wearer))
+		return
+	var/obj/item/card/id/id_card = wearer.wear_id?.GetID()
+	if(!istype(id_card))
+		return
+	if(!(id_card.rank in list(JOB_SO, JOB_XO, JOB_SQUAD_LEADER)))
+		to_chat(wearer, SPAN_WARNING("Only Staff Officers, Executive Officers and Squad Leader are permitted to give medal recommendations!"))
```
fira, line 421: SLs

Drulikar: They mean `Squad Leader` should be plural

code/datums/medal_awards.dm
TheGamerdk: Fixed

harryob (approved):

PR #5204, opened 2023-12-12
Requisitions elevator lighting fix

NateDross:
# About the pull request

This PR fixes the elevator lighting by changing the way the turf cloning logic works to use the pre-existing ChangeTurf function, which will result in the static lighting properties being copied properly. By doing this change, we can disable the base lighting properties from the shuttle area and have the static lighting system apply to the elevator turf in requisition which is consistent with how the room lighting already works.

The existing elevator has many visual effect issues, requiring other forms of lighting to compensate for the problem, such as placing movable lights (flashlight), or other non-static lighting sources. Without these work arounds the elevator will be completely black after the first time the space turfs are cloned.

<details>

## Before

https://github.com/cmss13-devs/cmss13/assets/149045778/f15fca7b-3a77-45ff-9632-63ee371eff64

## After

https://github.com/cmss13-devs/cmss13/assets/149045778/f99ae1af-f4ba-4a64-a66d-c13d5b2e13f8

</details>

# Changelog
:cl:
fix: Requisitions elevator lighting fix
/:cl:

### DIFF
--- a/code/game/supplyshuttle.dm
+++ b/code/game/supplyshuttle.dm
@@ -17,29 +17,25 @@ GLOBAL_DATUM_INIT(supply_controller, /datum/controller/supply, new())
 /area/supply
 	ceiling = CEILING_METAL
 
-/area/supply/station //DO NOT TURN THE lighting_use_dynamic STUFF ON FOR SHUTTLES. IT BREAKS THINGS.
+/area/supply/station
 	name = "Supply Shuttle"
 	icon_state = "shuttle3"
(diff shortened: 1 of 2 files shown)


fira (approved):
I've been looking for it for months and i never realized there were turfs instanciations here!!!!! I feel so stupid!!!! AGHHH thanks so much

Zonespace27 (approved):

Birdtalon (approved):
No issues noted from testmerges.

PR #5205, opened 2023-12-12
Fixes CAS killing players with ammo crates instead of bullets

### LINKED ISSUE #3211: Death by GAU gives cause of death as ammo crate instead of gun
### Testmerges

#3073, #3088, #2652, #3179, #3185, #3192

### Description of the bug

During CAS runs, if a mob dies to gunfire from a GAU cannon run, the cause of death (in deadchat) will be misattributed to "**PGU-100 [ammo type] 30mm ammo crate**" instead of the proper "**GAU-21 30mm cannon**".

Most relevant active testmerge is #3185. Not sure if it's the cause though.

GAU
image

Death message should say the gun instead of the ammo used to kill something/someone. Should be obvious as to why.

1. Equip a dropship with a GAU.
2. Do a flyby CAS fire mission.
3. Kill a mob with the GAU fire.

### Issue Bingo

SabreML:
# About the pull request

Fixes #3211 by making GAU-21 and Laser Cannon ammo pass the weapon to the `cause_data`, rather than the ammo crate/battery.

The dropship is meant to be firing the ammunition inside the crate, not the crate itself. (Although that is fun to imagine)

<details>

</details>

# Changelog

:cl:
fix: Fixed the death message from GAU-21 and Laser Cannon strikes saying that the player was killed by the ammo crate.
/:cl:

### DIFF
--- a/code/modules/cm_marines/dropship_ammo.dm
+++ b/code/modules/cm_marines/dropship_ammo.dm
@@ -12,8 +12,8 @@
 	var/fire_mission_delay = 4
 	/// Time to impact in deciseconds
 	var/travelling_time = 100
-	/// Type of equipment that accept this type of ammo.
-	var/equipment_type
+	/// Type of dropship equipment that accepts this type of ammo.
+	var/obj/structure/dropship_equipment/equipment_type
 	/// Ammunition count remaining
 	var/ammo_count
 	/// Maximal ammunition count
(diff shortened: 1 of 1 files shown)


fira (changes requested):
i can't agree with that fix because it breaks weapon subtyping

if you want to do that make the weapon be passed down detonate_on and use it directly rather than using equipment_type and initial

SabreML:
> i can't agree with that fix because it breaks weapon subtyping
> 
> 
> if you want to do that make the weapon be passed down detonate_on and use it directly rather than using equipment_type and initial

Thanks for the suggestion, I'll make the change as soon as I'm back later today. 👍

fira (approved):
That works!

PR #5207, opened 2023-12-12
Xeno telegraph effect can now be any rgb colour

Birdtalon:
# About the pull request

I didn't want it to be locked into the basic colours we have based on the sprites, so instead you can supply a colour code to the constructor to create any colour you want. Defaults to white if none provided. Eventually would be good to roll this out also the animated sprites.

When we add new castes/xenos we can diversify the colour palette for different effects instead of everything being red or brown.

https://github.com/cmss13-devs/cmss13/assets/25027759/be086245-eda3-4f27-8ce0-525df59674a1

<details>

</details>

# Changelog
:cl:
add: Xenomorph telegraph effects can now be any rgb colour.
/:cl:


InsaneRed:
these damn liberal xenos

InsaneRed:
and they thems dyed abilities

Birdtalon:
I'd quite like to actually use the animated telegraphs sometime. Eventually I'll get around to converting those also. They're unused currently.

FransFeiffer (approved):
As far as sprites go this sprite is fine so take my approval

code/game/objects/effects/aliens.dm
```diff
@@ -489,29 +489,30 @@
...
-	icon_state = "xeno_telegraph_brown"
+	color = COLOUR_BROWN
 
 /obj/effect/xenomorph/xeno_telegraph/green
-	icon_state = "xeno_telegraph_green"
+	color = COLOUR_GREEN
 
-/obj/effect/xenomorph/xeno_telegraph/brown/abduct_hook
+/obj/effect/xenomorph/xeno_telegraph/abduct_hook
 	icon_state = "xeno_telegraph_abduct_hook_anim"
```
Drulikar, line 512: Either this needs to inherit brown again, or it should set the color to brown. Or is the intention to change its color to none?

Drulikar (changes requested):

code/game/objects/effects/aliens.dm
Birdtalon: The icon for this is brown by default so as long as no colour argument is provided it will be brown. I could change the type path back to brown if that would be clearer? Eventually would like to get these all converted to white with colour arguments but need to find a spriter to do all the sprite work.

Drulikar: That's fine if this one then isn't using a white sprite to be colored, but then why set the one below this when it didn't used to have a color set to brown (it only inherited a path that was brown in name but that path had a white color)?

code/game/objects/effects/aliens.dm
```diff
@@ -489,30 +489,32 @@
...
 /obj/effect/xenomorph/xeno_telegraph/brown
-	icon_state = "xeno_telegraph_brown"
+	color = COLOUR_BROWN
 
 /obj/effect/xenomorph/xeno_telegraph/green
-	icon_state = "xeno_telegraph_green"
+	color = COLOUR_GREEN
 
-/obj/effect/xenomorph/xeno_telegraph/brown/abduct_hook
+///The below paths /abduct_hook and /lash have a brown icon state by and do not have a color overlay by default.
```
Drulikar, line 511:
```suggestion
/// This has a brown icon state and does not have a color overlay by default.
```

code/game/objects/effects/aliens.dm
```diff
@@ -489,30 +489,32 @@
...
+/obj/effect/xenomorph/xeno_telegraph/abduct_hook
 	icon_state = "xeno_telegraph_abduct_hook_anim"
 
-/obj/effect/xenomorph/xeno_telegraph/brown/lash
+/obj/effect/xenomorph/xeno_telegraph/lash
```
Drulikar, line 515:
```suggestion
/// This has a brown icon state and does not have a color overlay by default.
/obj/effect/xenomorph/xeno_telegraph/lash
```

Drulikar (changes requested):

Drulikar (approved):

PR #5211, opened 2023-12-13
DS doorgun no longer drops the gun when destroyed

cuberound:
# About the pull request

this bug has been around for a looong time, someone might be tempted to abuse it with the MG changes

moveable m56d with 1500 rounds that can shoot over cover is not intended

<details>

</details>

# Changelog

:cl:
fix: DS doorgun no longer drops the gun when destroyed
/:cl:


Drulikar (changes requested):
This does alter the behavior for a deployed turret, but not for an undeployed turret (that's equipped to a dropship - xenos can destroy it via acid but it will drop a gun as usual), nor for the unequipped box that you would equip to a dropship (still drops a gun as usual).

So likely you also need to alter `/obj/structure/dropship_equipment/mg_holder`

Also, when a destroyed system is picked up by a powerloader to uninstall it, it has the wrong icon_state (its invisible): 
image
I believe it should instead be: `sentry_system_destroyed`

.vscode/settings.json
```diff
@@ -9,7 +9,7 @@
 		"**/.pnp.*": true
 	},
 	"editor.codeActionsOnSave": {
-		"source.fixAll.eslint": true
+		"source.fixAll.eslint": "explicit"
```
Drulikar, line 12: Can revert this change. It likely will conflict. Instead merge upstream/master into your branch and this won't be something detected as a change.

Drulikar: Handled by merging upstream master myself.

Drulikar (approved):

PR #5216, opened 2023-12-14
Fixed hard hat user inventory head icon

NateDross:
# About the pull request
In the recent PR #5189 the action item for the hard hat was fixed. However because the icon was being turned on in hand instead of on head, it was overlooked that the user inventory head icon was also incorrect at times, specifically when turned on while the user is wearing it. This PR is related to the previous change mentioned because the `update_icon` call must occur first, due to the fact that both `update_button_icon` and `update_inv_head` both reference the icon and item state properties.

It is desirable for the icon state to be consistent in all situations. We want the light icon to show as shining when light is being emitted.

<details>

This update was tested both in hand and on head, as well as off states. The bug is reliably observable when the hard hat is activated while on head, however when in hand and placed on head in the on state, the bug does not occur. Upon various testing I have been unable to cause the icon to be in an inconsistent state.

## Before
before

## After
fixed

</details>

# Changelog
:cl:
fix: Fixed hard hat user inventory head icon
/:cl:

### DIFF
--- a/code/modules/clothing/head/hardhat.dm
+++ b/code/modules/clothing/head/hardhat.dm
@@ -57,10 +57,11 @@
 		return
 
 	set_light_on(toggle_on)
-	if(user == loc)
-		user.update_inv_head()
 
 	update_icon()
+
+	if(user == loc)
+		user.update_inv_head()
 
 	for(var/datum/action/current_action as anything in actions)
 		current_action.update_button_icon()


NateDross:
I wasn't sure how to properly word it but this fixes the actual sprite.

beforesprite

fira (approved):

PR #5218, opened 2023-12-14
SStechtree startup optimisation

SabreML:
# About the pull request

Optimises the initialisation time of the 'Tech Tree' subsystem.

Currently, `/datum/controller/subsystem/techtree/Initialize()` is changing the turf of every tile on the tech tree's z-level to `/turf/closed/void`, presumably so that it all appears black rather than space-y. However since it's calling `ChangeTurf()` on the *entire* z-level (300x226, or 67,800 turfs), this is pretty performance intensive.

This PR fixes that by just... not changing the turfs, instead only adding a few `/turf/closed/void`s around the edge of the tech tree area so that it still looks the same in-game. This changes the subsystem's init time from 6.4 seconds(ish), to 0.15 seconds(ish).

**Before:**
before 1

**After:**
after 1

Players viewing the tech tree through the console won't notice anything different, as the remaining layer of `/turf/closed/void`s block their view of the rest of the z-level.

About 6 seconds knocked off of the server's init time. (10%!) *(At least on my end)*

<details>
<hr>

### Before/after init times from a few different tests:
**Before:**
before 1
before 2
before 3

**After:**
after 1
after 2
after 3

<hr>

### Teleporting to it as an observer, there *is* a visible difference:

**Before:**
(null)scrnshot2

**After:**
(null)scrnshot1

<hr>

### But viewing it IC, it all works exactly the same as before:

https://github.com/cmss13-devs/cmss13/assets/57483089/14fc7b05-5279-489f-99df-e9acc83b1420

<hr>

### The new `/turf/closed/void` tiles highlighted for reference:

techtreebounds

</details>

# Changelog
:cl:
code: Made the Tech Tree subsystem initialise faster.
/:cl:

### DIFF
--- a/code/controllers/subsystem/techtree.dm
+++ b/code/controllers/subsystem/techtree.dm
@@ -34,17 +34,6 @@ SUBSYSTEM_DEF(techtree)
 		var/datum/space_level/zpos = SSmapping.add_new_zlevel(tree.name, list(ZTRAIT_TECHTREE))
 		tree.zlevel = zpos
 
-		var/zlevel = zpos.z_value
-		var/turf/z_min = locate(1, 1, zlevel)
-		var/turf/z_max = locate(world.maxx, world.maxy, zlevel)
-
-
-
-		for(var/t in block(z_min, z_max))
-			var/turf/Tu = t
-			Tu.ChangeTurf(/turf/closed/void, list(/turf/closed/void))
-			new /area/techtree(Tu)
-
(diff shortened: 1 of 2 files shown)


fira:
Good idea! Ideally I think we should eventually even port tech trees to reservations or just a tgui, but it's already a big step forward

Birdtalon (approved):
Makes sense.

PR #5221, opened 2023-12-15
Shuttle & Predator ship lighting fixes

NateDross:
# About the pull request

This PR contains the following changes
 * Corrected visual issues on tiles on predator ship
 * Corrected floorlight issues after turning off due to overlays being cut and not reapplied in open turfs `update_icon` function
 * Corrected shuttle ships lighting
 * Fixed where shuttle lights turned off after taking on, but not back on, as well as the shuttle light could never be shut off
* Added the `marine_dropship/on_arrival` callback, so that dropship equipment have the expected event triggered. This event was not triggering for the spotlights because it was missing.

This PR fixes various lighting related issues which are explained above as well as better explained in the images below. Overall these changes bring a better lighting experience for both marines and predators.

## Important notes
It is important to note that the shuttle only appears graphically correct when the base alpha is above 128. I tried playing with the colour matrix and other methods to have them consistent in all situations, but I am unsure if this is even a use-case which happens. Please let me know if there is a need to handle this situation and I can spend more time on it, but the alpha masking changes only apply to the shuttles using the `use_alphamasking` property.

Additionally, these changes add a method which should be used to get lighting effects for base lighting instead of referencing the property directly from the area. It is named `get_turf_lighting_effect`

Additionally, while I have not been able to spot any graphical issues, the changes to the `lighting_effect` property do affect the base lighting for **all tiles**, so any feedback on things to look out for would be welcomed. I have checked the ship over and common areas on surface maps but it can be tricky to spot things.
<details>

#### Before
colour_blending_before
#### After
colour_blending_after

#### Before
light_tile_offstate_before
#### After
light_tile_offstate_after

#### Before
shuttle_lightsoff_before
shuttle_lighting_surface_lightsoff_before

#### After
dropship_lighting_after
shuttle_lighting_surface_lightsoff_after

</details>

# Changelog
:cl:
fix: Fixed marine shuttle and predator ship lighting issues
fix: Fixed floorlight lighting issues
fix: Fixed marine shuttle spotlight toggling issues
/:cl:

### DIFF
--- a/code/game/area/shuttles.dm
+++ b/code/game/area/shuttles.dm
@@ -10,6 +10,11 @@
 	unique = FALSE
 
 	base_lighting_alpha = 255
+	area_has_alphamasking = TRUE
+
+	base_lighting_multiply_types = list(
+		/turf/open/floor/plating
+	)
 
 
 ///area/shuttle/Initialize()
--- a/code/game/area/shuttles.dm
+++ b/code/game/area/shuttles.dm
@@ -13,7 +13,7 @@
 	area_has_alphamasking = TRUE
 
 	base_lighting_multiply_types = list(
-		/turf/open/floor/plating
+		/turf/open/floor/plating,
 	)
 
 
--- a/code/game/objects/objs.dm
+++ b/code/game/objects/objs.dm
@@ -37,6 +37,10 @@
 	/// set when a player uses a pen on a renamable object
 	var/renamedByPlayer = FALSE
 
+	//The alphamasking is used for objects where the turf below it has lighting. It is only used in combination with base lighting, when the "area_has_alphamasking" property is enabled.
+	var/use_alphamasking = FALSE
+	var/mutable_appearance/alphamask_appearance = null
+
 
 /obj/Initialize(mapload, ...)
 	. = ..()
@@ -47,6 +51,8 @@
 	if(buckled_mob)
 		unbuckle()
 	. = ..()
+	if(alphamask_appearance != null)
+		QDEL_NULL(alphamask_appearance)
 	remove_from_garbage(src)
 
 /obj/vv_get_dropdown()
--- a/code/game/turfs/open.dm
+++ b/code/game/turfs/open.dm
@@ -18,8 +18,12 @@
 
 	update_icon()
 
+
 /turf/open/update_icon()
 	overlays.Cut()
+
+	var/area/area = get_area(src)
+	area.add_base_lighting_for_turf(src)
 
 	add_cleanable_overlays()
 
--- a/code/game/turfs/turf.dm
+++ b/code/game/turfs/turf.dm
@@ -97,16 +97,11 @@
 
 	//Get area light
 	var/area/current_area = loc
-	if(current_area?.lighting_effect)
-		overlays += current_area.lighting_effect
+	if(current_area?.get_turf_lighting_effect(src))
+		overlays += current_area.get_turf_lighting_effect(src)
 
 	if(opacity)
 		directional_opacity = ALL_CARDINALS
-
-	//Get area light
-	var/area/A = loc
-	if(A?.lighting_effect)
-		overlays += A.lighting_effect
 
 	return INITIALIZE_HINT_NORMAL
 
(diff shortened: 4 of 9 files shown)


cuberound:
could you possibly add those floor ligts that are used on.. BR LZ I think? might help with the hangar lighting

NateDross:
> could you possibly add those floor ligts that are used on.. BR LZ I think? might help with the hangar lighting

This PR focuses around code changes to improve the lighting and would be out of scope. I'm not sure who would be best to talk to about mapping changes but I am just looking to improve the lighting code.

NateDross:
@fira

In response to the questions on discord:

98d6b07bc34025169cf3caaddda20c88


* I have a few things to say code wise

    Assuming the outcome this PR is working to achieve is desired, I am happy to make any code changes you see fit. This is my first PR outside of minor code changes and want to adhere to the project standards.
* Why do we need to seperately multiply mask contents?

    The base lighting area is effectively 1.0, so if we are to add the darkness together we will get something like (1.0+0.2) = 1.2 (clamped to 1.0)
    When multiplying we will overcome this and have something like (0.2*1.0)=0.2. But this explanation is a bit vague and while I could just say "it works" and point to the screenshots, that is not really a great explanation, so I will try to point to other resources online to explain it.

    While these references are not specific to byond, the blending principles remain the same.
    - https://docs.unrealengine.com/5.1/en-US/material-blend-modes-in-unreal-engine/
        - Additive: `This means that there is no darkening; since all pixel values are added together, blacks will just render as transparent.`
        - Modulate (multiply): `The behavior is very similar to the Multiply Blend Mode in Photoshop, and produces a darkening` 
    - https://helpx.adobe.com/substance-3d-designer/substance-compositing-graphs/nodes-reference-for-substance-compositing-graphs/atomic-nodes/blend/blending-modes-description.html
        - Multiply: `As the value of each pixel is comprised between 0 and 1, the result is always equal or lower (darker) compared to the original.`
    - https://en.wikipedia.org/wiki/Blend_modes
        - Add: `Since this always produces the same or lighter colors than the input it is also known as 'plus lighter'.`
        - Multiply: `Wherever either layer was brighter than black, the composite is darker; since each value is less than 1, their product will be less than each initial value that was greater than zero. `
    - https://photoshoptrainingchannel.com/blending-modes-explained/#multiply
        - Multiply: `This Blending Mode multiplies the luminosity of the base color by the blend color. The resulting color is always a darker color. White produces no change, while the black pixels remain.`
    - https://tips.clip-studio.com/en-us/articles/4142
        - Add (Glow): `Any color that you apply on a layer that's on Add (Glow) makes that color brighter because it reduces contrast and intensifies brightness.`
        - Multiply: `Multiply is the blend mode I often use to add shadows in the drawing. In this mode, any color you put into a layer adds tint to the layer below it, creating a new shade or combined colors.`

It is complicated to explain and also I am not exactly a lighting expert, however based upon what resources I can find as well as my observation in game, this is the desired effect here, but even so this is limited to only the effected turfs in the shuttle area. Basically in this situation we are trying to darken the turfs, not brighten them. Currently as shown in the screenshots they are maximum brightness due to the base alpha settings. If we are to remove those settings then it does look correct, but the entire interior of the shuttle is dark. We could actually just solve the problem by having light sources in the shuttle as well.

- Why is this a turf dependent and area wide setting

In the shuttle (area), the turf selected is the only one where this lighting issue is visible
- Why only for the shuttles:

This same issue can occur anywhere base lighting happens beside non-base lighting areas, however I am not aware of many places where this occurs other than some admin areas, so I only enabled these settings for the shuttle.
- What if contents change
        
I will assume you mean the area contents and not shuttle, because these changes do not affect the interior of the shuttle.
If the contents change than the person changing them will need to consider the lighting properties, and at the time adjust the settings to be appropriate to look graphically correct for those changes


I hope these answers are sufficient and satisfy what you are asking, I do not mean to just blast you with a pile of links but it is a little difficult to explain, however based on various readings as well as the observed results from these code changes I believe it is correct to solve the lighting issues shown in the "before" screenshots on the shuttle.

NateDross:
I will put this PR up in smaller pieces, as well as investigate blend properties which do not involve scanning the contents.

PR #5226, opened 2023-12-16
Dynamic price for omnisentryguns

cuberound:
# About the pull request

Adds dinamic price to sentrguns, starts at 300 and goes up 100 points after each purchase. Aims to keep the anti spam meshures in place (makes them stricter in fact), while allowing players to purchase few of them for reasonable price.
Please before you put angry comments read this. 

Omnisentryguns are not used at all, and that is a shame, due to the changes that were were put in place to prevent sentryspam(they also weakened the sentrgun a lot along the way, it was not just price change) , but made them too expensive to buy just one. 

This PR changes it by setting the initial price to 300 and it goes up 100 points after each purchase.
This makes it so that no pilot has reason to not buy at least one.
It is balanced around 3 sentrguns for 400 points each, 5 sentrguns 500 points each ( old price) and 11 sentryguns 800 points each(that means for 8800 points, same as now), and if you try to buy more then current roundstart maximum of 11, they will cost more then they do now on avarage. 

This should enshure that FTLs and engineers can count with 1- 3 sentryguns each round and plan acordingly, decide if they want some for FOB, comms and hopfuly have one left to set up frontal outpost. 

In terms of effect on xeno gameplay, the sentrygun is quite weak after the nerfs, alone it can be taken down by basicly any two xenos with just a bit of efford and fold when t3 or defender attacks it. That means it serves the intended purpose, covering medics and engineers from skirmish castes, while slightly helping agains bigger threats while it is part of defended outpost. 

Please mind that the sentryguns can not be moved, so when marines overextend past it or are forced to retreat from it the sentry is good as dead.

In terms of pilot gameplay it kind of nerfs experienced pilots who get loads of calls as they will be expected to spend some points on sentryguns and have less ammo as resoult and less expirienced pilots who would not be able to make use of the ammo anyway are given another tool that is easier to use and makes them contribute to the round.

<details>

</details>

# Changelog

:cl:
balance: omnisentrygun price down to 300, but it goes up 100 points after each purchase
/:cl:


Cthulhu80:
It should probably cap out at a certain price.

code/modules/cm_marines/vehicle_part_fabricator.dm
```diff
@@ -58,6 +61,8 @@
 		return
 	visible_message(SPAN_NOTICE("[src] starts printing something."))
 	spend_point_store(cost)
+	if(part_type==/obj/structure/ship_ammo/sentry)
+		omnisentry_price +=omnisentry_price_scale
```
Zonespace27, line 65: the fact that we have to check the path (use `ispath()` instead of this) means that something is being done fundamentally wrong. Please change this to be generally in line with OOP.

code/modules/cm_marines/vehicle_part_fabricator.dm
```diff
@@ -28,7 +30,8 @@
 
 /obj/structure/machinery/part_fabricator/dropship/ui_data(mob/user)
 	return list(
-		"points" = get_point_store()
+		"points" = get_point_store(),
+		"omnisentrygun_price" =omnisentry_price
```
Zonespace27, line 34:
```suggestion
		"omnisentrygun_price" = omnisentry_price
```

tgui/packages/tgui/interfaces/PartFabricator.js
```diff
@@ -57,18 +57,35 @@ const GeneralPanel = (props, context) => {
...
-                      tooltip={Ammo.desc}
-                      tooltipPosition="left"
-                      onClick={() =>
-                        act('produce', {
-                          path: Ammo.path,
-                          cost: Ammo.cost,
-                        })
-                      }
-                    />
+                    Ammo.path === '/obj/structure/ship_ammo/sentry' ? (
```
Zonespace27, line 60: same here.

Zonespace27 (changes requested):

code/modules/cm_marines/vehicle_part_fabricator.dm
```diff
@@ -61,7 +61,7 @@
 		return
 	visible_message(SPAN_NOTICE("[src] starts printing something."))
 	spend_point_store(cost)
-	if(part_type==/obj/structure/ship_ammo/sentry)
+	if(istype(part_type,/obj/structure/ship_ammo/sentry))
```
Zonespace27, line 64: `ispath()`

cm13-github:

cm13-github:

PR #5246, opened 2023-12-18
Adds  antispam check similar to what sentrygun has to m56d

cuberound:
# About the pull request

(doing this on a bench so it will need some cleanup sorry for that) adds check to placing the m56d that checks if there is another m56d in surounding area to prevent mg spam, numbers and stuff are placeholder it will get cleaned up ( I will be happy if you help me with it)

m56d spam has come to be an issue recently, I am part of it and do not want it to end up like omni sentryguns did (compleatly useless garbage) so I try to take better aproach to the balance then just AP or firerate nerf that would not solve the issue

prevents m56d massing and encorages m56d users to spread out and cover diferent chokes instead of hosing down on and obliterating anyone who comes in the way

<details>

</details>

# Changelog
:cl:
balance: m56d can not be placed within five tiles of another one
/:cl:


code/modules/cm_marines/smartgun_mount.dm
```diff
@@ -44,6 +44,8 @@
 	icon_state = "M56D_gun_e"
 	var/rounds = 0 // How many rounds are in the weapon. This is useful if we break down our guns.
 	var/has_mount = FALSE // Indicates whether the M56D will come with its folding mount already attached
+	var/defense_check_range = 2
+	var/can_be_near_defense = FALSE
```
Birdtalon, line 48: This variable seems pointless if all you're going to do is check `if(!can_be_near_defense)` and it never changes may as well not have it and remove those checks also.

code/modules/cm_marines/smartgun_mount.dm
```diff
@@ -84,7 +86,11 @@
 
 /obj/item/device/m56d_gun/attack_self(mob/user)
 	..()
-
+	if(!can_be_near_defense)
+		for(var/obj/structure/machinery/machine in urange(defense_check_range, loc))
+			if(istype(machine, var/obj/structure/machinery/m56d_hmg) || istype(machine, var/obj/structure/machinery/m56d_post)
```
Birdtalon, line 91: Missing `)` and check your syntax of `istype()`

https://www.byond.com/docs/ref/#/proc/istype

code/modules/cm_marines/smartgun_mount.dm
```diff
@@ -125,6 +131,11 @@
 
 	if(!do_after(user, 1 SECONDS, INTERRUPT_ALL|BEHAVIOR_IMMOBILE, BUSY_ICON_BUILD))
 		return
+	if(!can_be_near_defense)
+		for(var/obj/structure/machinery/machine in urange(defense_check_range, loc))
+			if(istype(machine, var/obj/structure/machinery/m56d_hmg) || istype(machine, var/obj/structure/machinery/m56d_post)
```
Birdtalon, line 136: Missing `)` and `istype()` as above.

Birdtalon (changes requested):

cuberound:
is 2 enough or should I increase?

ihatethisengine:
I personally believe you should make it at least 5, 2 is definitely not enough.

Drulikar (changes requested):
This works except you can skip the check by placing a mount within the not allowed radius and then place the gun on the mount. So the same check has to apply to `m56d_post` as well.

code/modules/cm_marines/smartgun_mount.dm
```diff
@@ -307,6 +315,10 @@
 
 	if(istype(O,/obj/item/device/m56d_gun)) //lets mount the MG onto the mount.
 		var/obj/item/device/m56d_gun/MG = O
+		for(var/obj/structure/machinery/machine in urange(MG.defense_check_range, loc,1))
```
Drulikar, line 318:
```suggestion
		for(var/obj/structure/machinery/machine in urange(MG.defense_check_range, loc, TRUE))
```

code/modules/cm_marines/smartgun_mount.dm
```diff
@@ -44,6 +44,7 @@
 	icon_state = "M56D_gun_e"
 	var/rounds = 0 // How many rounds are in the weapon. This is useful if we break down our guns.
 	var/has_mount = FALSE // Indicates whether the M56D will come with its folding mount already attached
+	var/defense_check_range = 5
```
Drulikar, line 47: Autodoc
```suggestion
	///The distance this has to be away from other m56d_hmg and m56d_post to be placed.
	var/defense_check_range = 5
```

code/modules/cm_marines/smartgun_mount.dm
```diff
@@ -44,6 +44,7 @@
 	icon_state = "M56D_gun_e"
 	var/rounds = 0 // How many rounds are in the weapon. This is useful if we break down our guns.
 	var/has_mount = FALSE // Indicates whether the M56D will come with its folding mount already attached
```
Drulikar, line 46: Autodocc
```suggestion
	///How many rounds are in the weapon. This is useful if we break down our guns.
	var/rounds = 0
	///Indicates whether the M56D will come with its folding mount already attached
	var/has_mount = FALSE 
```

Drulikar (changes requested):

Drulikar (approved):
Code looks fine. Added to TM so we can get more of an idea of its balance. Undraft when you are happy with it.

fira (commented):
i don't think you need to include the POSTS in your range check that's just gonna be confusing

cuberound:
> i don't think you need to include the POSTS in your range check that's just gonna be confusing
you in fact do, as the post is also m56d just before you use screwdrive to finish it into the stage when it can shoot, so without it you can place down two posts, screw one finished and be unable to finish second one or interact with it at all/ be able to finish it and bypass the checks

Drulikar:
If it needs tweaking in the future it can be handled then; but this seems reasonable to me when already sentries have a similar mechanic.

PR #5253, opened 2023-12-19
Fixes fire resist spamming

### LINKED ISSUE #5124: Ability to remove multiple fire stacks in one resist
### Testmerges

#4986, #4939, #5054, #4842, #5067, #5073, #5018, #5100, #5101, #5108

### Round ID

Across multiple rounds.

### Description of the bug

When on fire as a xeno, you can use the 'resist' verb multiple times before the stun ends to remove multiple fire stacks in a single go. Can probably do the same as human.

You shouldn't be able to use the 'resist' verb while stunned

1. Get lit on fire, either with blue or normal fire
2. Press resist a lot really fast really hard!!!!
3. You're not on fire anymore

### Issue Bingo

### LINKED ISSUE #5247: Xeno double resisting while on fire
### Testmerges

#4842: '/tg/ Status Effects Part 2 - datum, KD, KO, Stuns' by fira at commit 595086537a this one maybe????? i dunno i dont code bruh..

### Round ID

20011 ( many more rounds before this one I was just too lazy to report )

### Description of the bug

xenos can double resist while on fire reducing their firestacks 2x faster than they should normally be able to ( game breaking !!!!!!!!!)

when they resist they shouldn't be able to keep resisting with lower/no cool down 

1. play xeno
2.get lit on fire
3.spam b
...

### Issue Bingo

SabreML:
# About the pull request

Fixes #5247 and fixes #5124 by changing the `if(!is_mob_incapacitated(TRUE))` check near the end of `/mob/living/verb/resist()` to `if(mobility_flags & MOBILITY_MOVE)`.
This was causing the bug because `is_mob_incapacitated()` doesn't check for the `WEAKEN` effect, which is what is added by `resist_fire()`.
(Editing `is_mob_incapacitated()` so that it checks for it would probably work too, but since TG uses the `MOBILITY_MOVE` thing and this system is ported from there, I figured I'd just go with that.)

Fixes fire being too easy to remove.

<details>

**Before:**

https://github.com/cmss13-devs/cmss13/assets/57483089/0b5b6777-4286-4eb1-9293-09ffcaa4fb4b

**After:**

https://github.com/cmss13-devs/cmss13/assets/57483089/84097909-a081-4c4f-8400-59e955b8a443

(Ignore the "There are no pipes within range" message, that's from one of my macros.)

</details>

# Changelog
:cl:
fix: Fixed being able to remove multiple stacks of fire by spamming resist.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Xenomorph.dm
@@ -1015,20 +1015,17 @@
 		return
 	var/breakouttime = legcuffed.breakouttime
 
-	next_move = world.time + 100
-	last_special = world.time + 10
+	next_move = world.time + 10 SECONDS
+	last_special = world.time + 1 SECONDS
 
 	var/displaytime = max(1, round(breakouttime / 600)) //Minutes
-	to_chat(src, SPAN_WARNING("We attempt to remove [legcuffed]. (This will take around [displaytime] minute(s) and we must stand still)"))
-	for(var/mob/O in viewers(src))
-		O.show_message(SPAN_DANGER("<B>[usr] attempts to remove [legcuffed]!</B>"), SHOW_MESSAGE_VISIBLE)
-	if(!do_after(src, breakouttime, INTERRUPT_NO_NEEDHAND^INTERRUPT_RESIST, BUSY_ICON_HOSTILE))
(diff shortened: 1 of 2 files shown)


fira (approved):

Drulikar (approved):
Changes appear to work and not currently seeing any bugs caused by this.

PR #5255, opened 2023-12-19
Additional ui_act checks for tacmap

Drulikar:
# About the pull request

This PR adds additional checks for the ui_acts when using the tacmap interface.

Should shore up most flaws allowing weird usage of it.

<details>

image

</details>

# Changelog

No player facing changes.

### DIFF
--- a/code/controllers/subsystem/minimap.dm
+++ b/code/controllers/subsystem/minimap.dm
@@ -1,3 +1,6 @@
+#define CANVAS_COOLDOWN_TIME 4 MINUTES
+#define FLATTEN_MAP_COOLDOWN_TIME 3 MINUTES
+
 /**
  *  # Minimaps subsystem
  *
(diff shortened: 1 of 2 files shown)


code/controllers/subsystem/minimap.dm
```diff
@@ -500,6 +503,8 @@ SUBSYSTEM_DEF(minimaps)
 		qdel(svg_store_overlay)
 		debug_log("SVG coordinates for [faction] are not implemented!")
 
+#define can_draw(faction, user) (faction == FACTION_MARINE && skillcheck(user, SKILL_LEADERSHIP, SKILL_LEAD_EXPERT) || faction == XENO_HIVE_NORMAL && isqueen(user))
```
fira, line 506: Can you add parens to explicit order here? a && b || c && d is ambiguous

fira (approved):

code/controllers/subsystem/minimap.dm
Drulikar: Order of precedence is here: https://www.byond.com/docs/ref/#/operator but sure.

code/controllers/subsystem/minimap.dm
```diff
@@ -832,60 +844,68 @@ SUBSYSTEM_DEF(minimaps)
...
+		if("selectColor")
 			var/newColor = params["color"]
 			if(newColor)
 				toolbar_color_selection = newColor
 				toolbar_updated_selection = newColor
 			action_queue_change += 1
 
-		if ("onDraw")
+		if("onDraw")
 			updated_canvas = FALSE
```
mullenpaul, line 870: do we want to return after these options?

Drulikar: As in the same prevention checks? We could - I just didn't think it necessary because they don't actually *do* anything outside of the UI; and it seems moot to ensure the UI has good values if they're messing with the UI already.

`updated_canvas` resets the announcement button in the UI so you have to click it twice (because the canvas needs to lose focus to apply its last change)
`toolbar_updated_selection` getting set to a non-color enum tells the CanvasLayer an action to perform.
`action_queue_change` getting set to any value that is not the previous value is to prevent the above being performed multiple times.

harryob (approved):

PR #5273, opened 2023-12-21
adds forced scream emote after warrior lunge

Cthulhu80:
# About the pull request

~~Created a new emote called panic, it's very similar to scream and inherits virtually all of its properties and methods however the ogg files are of course different.~~
Added a scream emote whenever a warrior lunges and grabs a human mob, also added one more ogg audio file to the scream list.

~~Initially I just wanted to add a forced a scream emote whenever a warrior lunged at a human mob, but I sort of found the existing scream ogg sounds to be kind of lack luster. However, they are iconic in their own right and removing them or modifying them would be a detriment to the game. I chose to create a new emote entirely rather than adding audio files to the existing one. The emote which is similar to scream but the sounds are more "authentic" and erm, immersive. So far it's just two sounds for each gender but this will be changed once I find more audio files to add.~~

Warrior lunges at marines should be a traumatic experience that gives them ptsd. 

# Changelog

:cl:
add: human mobs now scream whenever they get lunged at by a warrior.
soundadd: added a new scream sound and put it into the existing scream list. 
/:cl:


code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -245,6 +245,20 @@
...
+	message = "panics!"
+
+/datum/emote/living/carbon/human/scream/panic/get_sound(mob/living/user)
+	if(ishuman_strict(user))
+		if(user.gender == MALE)
+			return get_sfx("male_panic_scream")
+		else
+			return get_sfx("female_panic_scream")
+		if(isyautja(user))
+			return get_sfx("pred_pain")
```
Drulikar, line 260: This is inaccessible code.

Drulikar (changes requested):

Drulikar (changes requested):
Already discussed on discord, but just for parity here:
- The screams need to be much shorter
- The female screams seem especially out of place
- There should be no delay to the start of the clips (trim dead space so when you play the sfx it immediately makes the noise rather than a second later or whatever)

Cthulhu80:
> Already discussed on discord, but just for parity here:
> 
> * The screams need to be much shorter
> * The female screams seem especially out of place
> * There should be no delay to the start of the clips (trim dead space so when you play the sfx it immediately makes the noise rather than a second later or whatever)

Put the second sound into the scream list, otherwise removed all the other ones and just have the regular scream emote when the warrior lunges.

Drulikar (approved):

Birdtalon:
Can you put in your changelog that you're adding a new sound?

Cthulhu80:
> Can you put in your changelog that you're adding a new sound?

Done

PR #5278, opened 2023-12-22
Xenos now see marines on the tacmap during hijack if xenos outnumber marines for more than 25%

ihatethisengine:
# About the pull request
Xenos now see marines on the tacmap during hijack if xenos outnumber marines for more than 25%.

Hide and seek is boring. Queen need to be on ovi to access tacmap, so marines would still have a chance.

<details>
I am lazy but it worked when I tested.

</details>

# Changelog
:cl: ihatethisengine
add: Xenos now see marines on the tacmap during hijack if xenos outnumber marines for more than 25%.
/:cl:


code/modules/mob/living/carbon/human/death.dm
```diff
@@ -75,14 +75,22 @@
...
 				continue
-			if(last_living_human)
+			shipside_humans_count++
+			if(last_living_human && see_humans_on_tacmap)
 				last_living_human = null
 				break
 			last_living_human = cur_human
-		if(last_living_human && (GLOB.last_qm_callout + 2 MINUTES) < world.time)
+
+		if(!see_humans_on_tacmap && shipside_humans_count < main_hive.totalXenos.len * 0.75)
```
Contrabang, line 90: This means they would need to outnumber them by 33%, doesn't it?
For example as it currently is, 12 xenos needs less than 9 humans remaining. 12/9 = 133%

If you wanted 25 percent, you'd need to place it like this on the opposite side
```suggestion
		if(!see_humans_on_tacmap && shipside_humans_count * 1.25 < main_hive.totalXenos.len)
```
or adjust it like this so that its the reciprocal of 1.25
```suggestion
		if(!see_humans_on_tacmap && shipside_humans_count < main_hive.totalXenos.len * 0.8)
```

eg. 15 xenos requires less than 12 humans, 15/12 = 125%
This may seem counterintuitive at first, but 1.25 (5/4) and 0.8 (4/5) are reciprocals, so 1.25 * 0.8 = 1.

ihatethisengine: :nerd_face:

fira (commented):
Looks good but what about corrupted / Xeno ERT etc ? Should they be revealed on low human count too ?

ihatethisengine:
I think this code let you see basically everyone, just like observers can see all factions on the tacmap. Anyway, iirc greenos and feral don't count towards round end, so it shouldn't be really relevant.

Skatuk:
Does the last guy dissapear from the tacmap? You already have the QM giving you directions. And its the last guy against 20 xeno's

ihatethisengine:
Not really a situation I care about in particular. If there is one guy against twenty xenos, I'd rather have them to die quickly than to run around for 10-15 minutes. QM message happens only once per two minutes and by the time you get there they might already be gone. Not a situation I want to make exceptions for.

code/controllers/subsystem/minimap.dm
```diff
@@ -689,6 +689,11 @@ SUBSYSTEM_DEF(minimaps)
 	if(faction == FACTION_NEUTRAL && isobserver(user))
 		faction = allowed_flags == MINIMAP_FLAG_XENO ? XENO_HIVE_NORMAL : FACTION_MARINE
 
+	if(is_xeno && xeno.hive.see_humans_on_tacmap)
+		allowed_flags = MINIMAP_FLAG_ALL
```
Drulikar, line 693:
```suggestion
		allowed_flags |= MINIMAP_FLAG_USCM|MINIMAP_FLAG_PMC|MINIMAP_FLAG_UPP|MINIMAP_FLAG_CLF
```

Alternatively create:
`#define MINIMAP_FLAG_HUMANS MINIMAP_FLAG_USCM|MINIMAP_FLAG_PMC|MINIMAP_FLAG_UPP|MINIMAP_FLAG_CLF` in `code\__DEFINES\minimap.dm` and use `allowed_flags |= MINIMAP_FLAG_HUMANS`

code/controllers/subsystem/minimap.dm
```diff
@@ -689,6 +689,11 @@ SUBSYSTEM_DEF(minimaps)
...
+		faction = FACTION_NEUTRAL
```
Drulikar, line 694:
```suggestion
```
Why did you think this was necessary?

Drulikar (changes requested):

code/modules/mob/living/carbon/human/death.dm
```diff
@@ -75,14 +75,22 @@
...
 				continue
-			if(last_living_human)
+			shipside_humans_count++
+			if(last_living_human && see_humans_on_tacmap)
 				last_living_human = null
 				break
 			last_living_human = cur_human
-		if(last_living_human && (GLOB.last_qm_callout + 2 MINUTES) < world.time)
+
+		if(!see_humans_on_tacmap && shipside_humans_count < main_hive.totalXenos.len * 0.8)
```
Drulikar, line 90:
```suggestion
		if(!see_humans_on_tacmap && shipside_humans_count < length(main_hive.totalXenos) * 0.8)
```

Drulikar (changes requested):

code/controllers/subsystem/minimap.dm
ihatethisengine: I copypasted this from observer tacmap :sob:

code/controllers/subsystem/minimap.dm
```diff
@@ -689,6 +689,10 @@ SUBSYSTEM_DEF(minimaps)
 	if(faction == FACTION_NEUTRAL && isobserver(user))
 		faction = allowed_flags == MINIMAP_FLAG_XENO ? XENO_HIVE_NORMAL : FACTION_MARINE
 
+	if(is_xeno && xeno.hive.see_humans_on_tacmap)
+		allowed_flags |= MINIMAP_FLAG_USCM|MINIMAP_FLAG_PMC|MINIMAP_FLAG_UPP|MINIMAP_FLAG_CLF
+		targeted_ztrait = ZTRAIT_MARINE_MAIN_SHIP
```
Drulikar, line 694:
```suggestion
	if(is_xeno && xeno.hive.see_humans_on_tacmap && targeted_ztrait != ZTRAIT_MARINE_MAIN_SHIP)
		allowed_flags |= MINIMAP_FLAG_USCM|MINIMAP_FLAG_PMC|MINIMAP_FLAG_UPP|MINIMAP_FLAG_CLF
		targeted_ztrait = ZTRAIT_MARINE_MAIN_SHIP
		map_holder = null
```
We must also null out `map_holder` so the different z-level minimap is fetched below. The additional condition is just to make this only occur once.

But just VVing it to null I can then get it to work:
image

Drulikar (changes requested):

Drulikar (approved):

SabreML:
It may have been mentioned already, but for some reason the aft section of the ship is cut off on the map:
image

So if someone's hiding at lifeboats for example, they probably won't be visible.

Drulikar:
> It may have been mentioned already, but for some reason the aft section of the ship is cut off on the map:
> 
> So if someone's hiding at lifeboats for example, they probably won't be visible.

Out of scope for this PR to resolve unless ihate really wants to - but also pretty minor. Also the drawn on map (should the queen do so) is more zoomed out too.

Drulikar:
If you do want to tackle it though, its as simple as increasing the size of `icons/ui_icons/minimap.dmi` and then changing these values: 
image
For example 512x512 would look like this: 
image

But you will want to see how this looks on all maps in all interfaces (squad computers, CAS UI, etc).

ihatethisengine:
Let's get this merged first, then when I have time I will look into it and make a separate PR.

code/modules/mob/living/carbon/human/death.dm
```diff
@@ -75,14 +75,22 @@
...
 				continue
-			if(last_living_human)
+			shipside_humans_count++
+			if(last_living_human && see_humans_on_tacmap)
 				last_living_human = null
 				break
 			last_living_human = cur_human
-		if(last_living_human && (GLOB.last_qm_callout + 2 MINUTES) < world.time)
+
+		if(!see_humans_on_tacmap && shipside_humans_count < length(main_hive.totalXenos) * 0.8)
```
Drulikar, line 90: This change will require #5122 but make the ratio (0.8) be a define like in that PR near the other defines. Also count the xenos using get_real_total_xeno_count() from that PR as well.

Drulikar (changes requested):

PR #5281, opened 2023-12-23
Fix queen death not readmitting banished xenos fully

Drulikar:
# About the pull request

This PR fixes either an oversight, or a recent bug that I've not been able to find the source, of where the death of a queen resets the hive's banished keys list, but doesn't fully unbanish those xenos leaving them in a state that they can't be banished properly again nor readmitted.

A queen should retain her ability to banish and readmit all xenos freely and not be beholden to the will of some former queen.

<details>

free

</details>

# Changelog
:cl: Drathek
fix: Fix queen death not fully readmitting any banished xenos
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/Queen.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Queen.dm
@@ -792,12 +792,23 @@
 		return TRUE
 
 /mob/living/carbon/xenomorph/queen/death(cause, gibbed)
-	if(hive.living_xeno_queen == src)
+	if(src == hive?.living_xeno_queen)
 		hive.xeno_queen_timer = world.time + XENO_QUEEN_DEATH_DELAY
-		hive.banished_ckeys   = list() // Reset the banished ckey list
+
+		// Reset the banished ckey list
+		if(length(hive.banished_ckeys))
+			for(var/mob/living/carbon/xenomorph/target_xeno in hive.totalXenos)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/Queen.dm
```diff
@@ -792,13 +792,24 @@
...
 
 /mob/living/carbon/xenomorph/queen/death(cause, gibbed)
-	if(hive.living_xeno_queen == src)
+	if(src == hive?.living_xeno_queen)
 		hive.xeno_queen_timer = world.time + XENO_QUEEN_DEATH_DELAY
-		hive.banished_ckeys   = list() // Reset the banished ckey list
+
+		// Reset the banished ckey list
+		if(length(hive.banished_ckeys))
+			for(var/mob/living/carbon/xenomorph/target_xeno in hive.totalXenos)
```
Cthulhu80, line 800: You know more than me, but why can't we as anything?

Drulikar: It's not that we can't, but this proc is called 0-a few times ever in a round so we might as well enforce the variables contained in the list are properly typed. In other words this is very cold code that is not important to be as performant as possible. As anything means we skip the type checking; but not doing it on a typed iterator means we won't risk a runtime should there be an entry that doesn't cast to that type.

fira (approved):

PR #5303, opened 2023-12-25
Fixes synths being unlungeable while in "critical state"

InsaneRed:
# About the pull request

Synths were given immunity to dragging from old code back when they could go into 'crit' so they dont get dragged to hell, however they no longer have a crit state but still keep the immunity lunges also do not work because of this. this is not intended.

bugfix

<details>
yeah i tested it also no you cant get dragged while dead

</details>

# Changelog
:cl:
fix: Synths are no longer immune to lunges / dragging while in 'critical state' since they dont go into crit.
/:cl:


fira:
https://github.com/cmss13-devs/cmss13/commit/76198c8cf96ba9542b06abc77350771475ce2aef

fira (approved):
given the other introduced check was removed as in above commit, i'm assuming this is a leftover

PR #5309, opened 2023-12-26
sentry laptop now uses camera manager

mullenpaul:
# About the pull request

The sentry laptop had a bunch of custom code for handling gunsight cameras. This PR refactors it to exclusively use the camera_manager component.

# Explain why it's good for the game

Reduction of duplicated code.
# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
refactor: sentry laptop now uses camera manager component
/:cl:

PR #5298, opened 2023-12-25
Polish Dropship Weapons UI

Drulikar:
# About the pull request

This PR is a follow up to #4812 polishing minor issues that were discovered but not addressed in that PR.

- Fire missions can now be scrolled in target acquisition
- The target selection in equipment view is now the same as other panels (and can be scrolled now too)
- Fixed deploying equipment like the spotlight and sentry
- ~~Partially~~ fixed camera view for sentry ~~(map size doesn't always get set correctly some reason)~~
- Fixed the extra scrolling of fire missions
- Fixed the scrolling of target selection being a fixed quantity of 10 targets
- Improved menuing for the fire mission and strike sub menus
- Fixed strike ready messaging
- Map and Cam buttons are now mutually exclusive to one another (would just break the byond UI)
- Camera now resets when a CAS flare is destroyed (unless fire mission is underway)
- Camera now resets when sentry undeploys (if it was being used)
- Tweaked some button placements
- Fixed medevac layout
- Fixed direct fire (strike & equipment)
- Fixed camera views not handling /datum/component/overlay_lighting correctly (fixed by using TILE_BOUND planes; but will note it won't work completely correct if on a byond version prior to 515.1609 because of https://www.byond.com/forum/post/2873835)
- Fixed some hard deletes in `/client/proc/clear_map` based on https://github.com/tgstation/tgstation/pull/61562
- Renamed nvgon and nvgoff to NV-ON and NV-OFF

# Explain why it's good for the game

Fixes issues such as (but not limited to): 
target
image

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

image
image
image
light

</details>

# Changelog
:cl: Drathek
ui: Polished various aspects of the new dropship weapons UI
fix: Fixed CAS direct firing
fix: Fixed Medevac buttons not moving the dropship (still currently requires manual winch)
fix: Fixed camera_manager resizing the view incorrectly because of overlay_lighting
refactor: Ported some hard delete fixes for maps.
/:cl:

PR #5310, opened 2023-12-26
camera console now uses camera manager

mullenpaul:
# About the pull request

Continuing the work of https://github.com/cmss13-devs/cmss13/pull/5309 - camera code now uses camera manager component.

Reduction of duplicated code.
<details>

</details>

# Changelog
:cl:
refactor: camera consoles now use camera manager component
/:cl:


fira (approved):

Drulikar (commented):
Not really worth doing this change until we resolve the issue where camera_manager is not refreshing byondui correctly.

(As in it not scaling when screen_loc dimensions change) 
image

code/game/camera_manager/camera_manager.dm
```diff
@@ -87,6 +87,7 @@
 	RegisterSignal(parent, COMSIG_CAMERA_SET_AREA, PROC_REF(set_camera_rect))
 	RegisterSignal(parent, COMSIG_CAMERA_SET_TARGET, PROC_REF(set_camera))
 	RegisterSignal(parent, COMSIG_CAMERA_CLEAR, PROC_REF(clear_camera))
+	RegisterSignal(parent, COMSIG_CAMERA_REFRESH, PROC_REF(refresh_area_camera))
```
Drulikar, line 90:
```suggestion
	RegisterSignal(parent, COMSIG_CAMERA_REFRESH, PROC_REF(refresh_camera))
```

code/game/camera_manager/camera_manager.dm
```diff
@@ -113,6 +115,10 @@
 	target_height = null
 	show_camera_static()
 
+/datum/component/camera_manager/proc/refresh_area_camera()
+	SIGNAL_HANDLER
+	update_target_camera()
```
Drulikar, line 120:
```suggestion
/datum/component/camera_manager/proc/refresh_camera()
	SIGNAL_HANDLER
	if(render_mode == RENDER_MODE_AREA)
		update_area_camera()
		return
	update_target_camera()
```

code/game/camera_manager/camera_manager.dm
Drulikar, line 1: While we are still on 514 (also may just be required entirely for the dropship regardless of being on 515 but I've not sorted out that issue yet), add three instances of this snippet:
```
// Workaround for https://www.byond.com/forum/post/2873835
cam_background.name = cam_background.name == "background" ? "background_1" : "background"
```
- Inside `if(last_camera_turf == newturf)` in update_target_camera
- Inside `if(last_camera_turf == new_location)` in update_area_camera
- After `cam_background.fill_rect(1, 1, size_x, size_y)` in render_objects

Alternatively here's a patch with these changes: [patch.diff.txt](https://github.com/cmss13-devs/cmss13/files/13783587/patch.diff.txt)

code/game/machinery/computer/camera_console.dm
```diff
@@ -160,55 +134,18 @@
 		if(!selected_camera)
 			return TRUE
 
-		update_active_camera_screen()
+		SEND_SIGNAL(src, COMSIG_CAMERA_SET_TARGET, selected_camera, selected_camera.view_range, selected_camera.view_range)
```
Drulikar, line 137: It is currently moot to assign `w` and `h` because `/datum/component/camera_manager/proc/set_camera` merely sets those into `target_width` and `target_height` which isn't used for anything.

code/game/machinery/computer/camera_console.dm
```diff
@@ -160,55 +134,18 @@
...
 	var/user_ref = WEAKREF(user)
 	var/is_living = isliving(user)
 	// Living creature or not, we remove you anyway.
 	concurrent_users -= user_ref
 	// Unregister map objects
-	user.client.clear_map(map_name)
+	SEND_SIGNAL(src, COMSIG_CAMERA_UNREGISTER_UI, user)
 	// Turn off the console
 	if(length(concurrent_users) == 0 && is_living)
 		current = null
```
Drulikar, line 151: Our current is nulled out, but the camera manager's current is not nulled out.
```suggestion
		current = null
		SEND_SIGNAL(src, COMSIG_CAMERA_CLEAR)
```
Without this, a closed camera_console when re-opened is in a weird state where it thinks we aren't currently viewing a camera but the view is on a camera: 
image

Though I will note that fixing this behavior sort of makes the other bug where it is a incorrectly sized map moot.

Drulikar (changes requested):

code/game/camera_manager/camera_manager.dm
mullenpaul: I think we could capture this in a #ifdef statement - that way we won't have a regression on migrating to 515. I'm also thinking we can add a warning or todo for 515 to clean up.

Drulikar: Sure though dropship_weapons needs this workaround regardless of being on 515. I'm just not completely satisfied with the workaround either since I found a situation where the workaround does the opposite of what it ought to do... So there's something really funky with the dropship_weapons.

But the name toggling should be pretty cheap and harmless anyways to not be concerned about removing this from 515 code with an compiler define anyways.

Drulikar: I have now solved the issue with the dropship_weapons usage of camera_manager by setting all the planes to be TILE_BOUND. The name toggling is still necessary for any byond build prior to 515.1509 because of https://www.byond.com/forum/post/2873835 but its not a large problem.

code/game/machinery/computer/camera_console.dm
mullenpaul: Reworked this so that it uses the target_width and target_height being passed in via signal. It slightly reduces decoupling between the component and the camera.

code/game/camera_manager/camera_manager.dm
```diff
@@ -191,7 +200,7 @@
 	// Cameras that get here are moving, and are likely attached to some moving atom such as cyborgs.
 	last_camera_turf = get_turf(cam_location)
 
-	var/list/visible_things = current.isXRay() ? range(current.view_range, cam_location) : view(current.view_range, cam_location)
+	var/list/visible_things = current.isXRay() ? range("[target_width]x[target_height]", cam_location) : view("[target_width]x[target_height]", cam_location)
```
Drulikar, line 203: This change seems to just produce incorrect results in every view:
image
Despite it having a 7 and 7 for width and height, it made a view of 5x7.

If its not merely because you don't have the fix setting target_width and target_height in `/datum/component/camera_manager/proc/set_camera_rect(source, x, y, z, w, h)` like in #5298 then I'm not sure.

Drulikar (changes requested):

code/game/camera_manager/camera_manager.dm
Drulikar: I also recommend using 515.1609+ so you don't have the viewport sizing bug.

Drulikar: Reverting only that line the views seem to be properly displayed: 
image

cm13-github:

mullenpaul:
I'll resolve conflicts today

cm13-github:

Drulikar (approved):
Ultimately this makes `set_camera_rect` proc not do what it is intended to do (target_width and target_height both aren't used for anything) but your version does atleast have no in game problems as far as I can see.

PR #5316, opened 2023-12-27
Fixes the Hive Status error when the Queen dies

### LINKED ISSUE #4330: Xeno Hive Status Screen Error
### Testmerges

N/A

### Description of the bug

The screen shows a TGUI error when the queen dies if you keep the xeno hive status panel open.
I have been able to reproduce locally as both a xeno and a ghost, as long as the screen is open when the queen dies (it does not error if the queen is deleted).

Either the screen to close, or continue to function correctly, instead of runtiming.

1. Have the xeno hive status screen open
2. Have queen die, but not get deleted
3. Hive status panel runtimes

### Issue Bingo

SabreML:
# About the pull request

Fixes the Hive Status window showing an error message when the Queen dies.

For xeno players, the interface just closes.
For observing players, the interface will show 'The Hive has no Queen!' where the queen's location was previously.
(The bug actually causing the error should be fixed too now)

(Note: I'm not quite sure what the `tgui-panel.bundle.css` diff is from, but it kept modifying itself whenever I compiled so I figured it was something related to this.)

Fixes #4330

<details>

**Testing queen spawn and death as the queen, as an observer, and as a xeno:**

https://github.com/cmss13-devs/cmss13/assets/57483089/ba3b2960-d4ce-4be4-af6b-6daae2d10ac1

</details>

# Changelog
:cl:
fix: Fixed the Hive Status window showing an error message when the Queen dies.
/:cl:


Birdtalon (approved):

Drulikar (changes requested):
The hive status is accessed by more than just players in the game. So if anything the observer check should be before this UI_CLOSE return

(Testing not using this PR):
For example, a hive with a lone xeno viewed by observer: 
image
But of course the player can't access it: 
image

Tested on this PR: Nothing happens when an observer clicks view hive status.

You'd be better off simply adding a null check in the TGUI than this.

SabreML:
> The hive status is accessed by more than just players in the game. So if anything the observer check should be before this UI_CLOSE return (but that won't prevent a crash)
> 
> (snip)
> 
> You'd be better off simply adding a null check in the TGUI than this.

For some reason, I had thought that observers couldn't see the hive status either if there was no queen.
I'll move the `ui_status()` check down so that it doesn't get closed for observers, and add something to the frontend.

fira (approved):

cm13-github:

cm13-github:

Drulikar (approved):

Issue #4880, opened 2023-11-08
Emergency shutters dont work


Releasethesea:
### Testmerges

#4475, #4812, #4793, #4866, #4863, #4872

### Round ID

19438

### Description of the bug

Emegency shutters are opening in the wrong areas all over the ship

### What's the difference with what should have happened?

they should only open where they are at, not outside of the area, heres my crude drawings showing what happens, this occurs are multiple locations through the ship which it shouldnt be doing

![fireshutter2](https://github.com/cmss13-devs/cmss13/assets/135743398/516ca215-c354-4c43-9d66-95029d55adf6)
![fireshutter](https://github.com/cmss13-devs/cmss13/assets/135743398/8679f893-6026-4b2d-8f53-bd40cdb3cd68)

### How do we reproduce this bug?

1. toggle the fire alarm
2.look outside of the area 
3.see its not properly aligned with the area/where its ment to be
...

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Huffie56:
emergency shutter button is suppose to only activate shutter in the area they are in..
so yes it's a bug maybe it's a mapping issue?

Huffie56:
after testing this is a bug and not a simple mapping issue.

Cthulhu80:
yeah, this is fucked.

Cthulhu80:
https://github.com/cmss13-devs/cmss13/assets/122310258/58193a27-e532-43f0-9754-d6f65d4f4d5b

Huffie56:
so yes commenting out looking out work for some situation but on other we need code to look out for adjacent shutter to close.
![image](https://github.com/cmss13-devs/cmss13/assets/117036822/fba43120-f2a5-487c-9259-55df4cb60b45)
let's imagine an extreme example where red is "area A"(red)and is surrounded by other areas with their shutter at their access 
@Cthulhu80

Huffie56:
you need in some way to only restrict it to only check adjacent area from floor not from walls.

Huffie56:
remapping area could also be a solution in most case...

Cthulhu80:
wat

Cthulhu80:
I mean, idk man, whatever ya want me to do. I ain't no mapper though. I think you are, right? Might be a task for you then.

[...]

PR #5325, opened 2023-12-28
Fixes fire alarm closing shutters for adjacent area doors

Cthulhu80:
# About the pull request

From the code this looks intentional.... is it even a bug? #4880.

having all adjacent shutters close in a given area just looks a bit silly. 

shutters

# Changelog

:cl:
fix: fixes adjacent shutters closing after a fire alarm is pulled.
/:cl:


Huffie56:
so yes commenting out looking out work for some situation but on other we need code to look out for adjacent shutter to close.
image
let's imagine an extreme example where red is "area A"(red)and is surrounded by other areas with their shutter at their access point(orange).
when i push the fire alarm in area A code need to look to adjacent area to close orange shutter.

Cthulhu80:
Said this on discord but, maybe instead we can instead have it so that all connected fire shutters to any adjacent shutter close as well. That way it doesn't look funky and closes all the nearby room shutters.

code/game/machinery/doors/firedoor.dm
```diff
@@ -54,11 +54,11 @@
...
-	for(var/direction in GLOB.cardinals)
-		A = get_area(get_step(src,direction))
-		if(istype(A) && !(A in areas_added))
-			A.all_doors.Add(src)
-			areas_added += A
+	// for(var/direction in GLOB.cardinals)  why are we even doing this?!
+	// 	A = get_area(get_step(src,direction))
+	// 	if(istype(A) && !(A in areas_added))
+	// 		A.all_doors.Add(src)
+	// 		areas_added += A
```
Drulikar, line 61:
```suggestion
```
Just remove it.

code/game/machinery/doors/firedoor.dm
```diff
@@ -54,11 +54,11 @@
 	A.all_doors.Add(src)
 	areas_added = list(A)
```
Drulikar, line 55: Refactor out the need for the list `areas_added`

Drulikar (changes requested):

fira (approved):
probably intended for atmos so who cares

Drulikar (changes requested):
Works but it is likely desirable to just remove hallways fire alarms because there's just going to still be situations like this where the hallway is just a gigantic region so it having a firealarm affects things super far away - and even then it doesn't share fire doors that are shared by a different area: 
image
image

Technically in this situation it ought to have closed the doors in "Port Fore Lifeboat Fuel Pump" and "Medical Upper" (and maybe others because this area spans a good distance) to be a proper fire barrier. So if you opt to not remove hallway firealarms, then maybe consider adding the old code back but have it fire alarm that entire zone (but no further than it?) but I really do not expect this to work well. Technically the old code would have handled this situation better but obvious that has its own flaws. Alternatively we break up the areas more, and we add a way for firedoors to be linked to multiple areas in sets.

Also the fire alarms themselves don't change state visually in any way (most code bases change them to a flashing red state). We do have a state for this, just this code isn't setting it. 
image

Cthulhu80:
> Works but **it is likely desirable to just remove hallways fire alarms** because there's just going to still be situations like this where the hallway is just a gigantic region so it having a firealarm affects things super far away - and even then it doesn't share fire doors that are shared by a different area: 
> 
> image
> 
> image
> 
> 
> 
> Technically in this situation it ought to have closed the doors in "Port Fore Lifeboat Fuel Pump" and "Medical Upper" (and maybe others because this area spans a good distance) to be a proper fire barrier. So if you opt to not remove hallway firealarms, then maybe consider adding the old code back but have it fire alarm that entire zone (but no further than it?) but I really do not expect this to work well. Technically the old code would have handled this situation better but obvious that has its own flaws. Alternatively we break up the areas more, and we add a way for firedoors to be linked to multiple areas in sets.
> 
> 
> 
> **Also the fire alarms themselves don't change state visually in any way (most code bases change them to a flashing red state). We do have a state for this, just this code isn't setting it.** Be sure it changes the state of all firealarms in an area if its possible for multiple firealarms to exist in a single area, and be sure it returns back to whatever code green/blue/red state. 
> 
> image
> 
> 

Sure yeah I can do that, @Huffie56, if u want to remove the hallway fire alarms and shutters. I can finish up the pr after. Feel free to commit your changes to this pr.

Drulikar:
I'll note its probably fine to leave the shutters in hallways - I'm okay with there being no current way to deploy them. Its just this implementation off fire alarms in hallways isn't desirable to have them deploy so if we only have fire alarms in rooms it should work better with this implementation. But maybe maintainer will think we should find a better code solution to this.

Cthulhu80:
> I'll note its probably fine to leave the shutters in hallways - I'm okay with there being no current way to deploy them. Its just this implementation of fire alarms in hallways isn't desirable to have them deploy so if we only have fire alarms in rooms it should work better with this implementation. But maybe maintainer will think we should find a better code solution to this.

I think the optimal solution would have to involve some sort of mapping, which again I don't know how to do. @Huffie56, what do you think?

Huffie56:
My personal opinion is that removing current check is fine.

we also need to remind that their isn't only hallway that share emergency shutters.
that's why we had the code check for adjacent area in the first place...

one of the most used example is med bay and the OR's when you pull for treatment center room you want all the emergency for OR's to go down to...

just test in medbay and if you find this ok then go....

if i am not  very clear apology i am a bit tired...

Cthulhu80:
I'm going to be gone for a bit @Huffie56, if u want to take this bug over. Closing the pr.

PR #5326, opened 2023-12-28
rto backpack can't be accessed on the move

mullenpaul:
# About the pull request

Makes the radio backpack not accessible on the move, like the standard backpacks.

Right now the radio backpack is a straight upgrade to the USCM satchel. It has the same storage size and can be readily accessible. This makes the radio backpack almost always better. The radio functionality should have some sort of trade-off.

<details>

</details>

# Changelog
:cl:
balance: RTO backpacks can't be accessed on the move.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/backpack.dm
+++ b/code/game/objects/items/storage/backpack.dm
@@ -496,6 +496,7 @@ GLOBAL_LIST_EMPTY_TYPED(radio_packs, /obj/item/storage/backpack/marine/satchel/r
 	icon_state = "rto_backpack"
 	item_state = "rto_backpack"
 	has_gamemode_skin = FALSE
+	worn_accessible = FALSE
 	actions_types = list(/datum/action/item_action/rto_pack/use_phone)
 
 	flags_item = ITEM_OVERRIDE_NORTHFACE


iloveloopers:
What happened to the feature freeze?

ihatethisengine:
Frankly, I believe players should be rewarded for willingness to communicate, not punished. You already spend 5 points to buy the backpack, which is not huge, but anyway. And it's a backpack, most players don't carry backpacks whatsoever, because the slot is way more useful for a second weapon. Adding another downside would simply lead to the pack being used being used even less frequently (which is already rare). And I don't think there is a valid reason to do so, the backpack, again, even not that useful and communication is a thing we want to be present more, not less.

InsaneRed:
There is no reason to punish rto users anymore then we do already by forcing them to not take a second weapon and its incentive to run it

Sylphily:
i know when given the option between carrying a backpack and a RTO phone, i end up taking the RTO because its useful to call and decent for storage, but if its just a delay to use it might as well take a bag over it for more storage, or take the other IMP Rack instead, in the end i don't think it boons will end up outweighing this downside

Knuff-les:
Adding the delay on the backpack doesn't do anything for balance, it’d just make an already rarely used feature of the game used even less. The tradeoffs for the backpack already exist. Less space than the regular backpack, and you lose a slot for an extra weapon/equipment. This PR is nothingburger imo

Doom-external:
Why not consider making it so that people with no/low JTAC skill have a cooldown to open it while RTOs and SLs don't. Encourage roles that should still be communicating to use it.

PR #5329, opened 2023-12-28
Fixes xenos seeing raw html on paper

### LINKED ISSUE #4609: Xenos can see raw HTML tags on papers

Cthulhu80:
# About the pull request
Continuation of #4654 , to fix #4609 . This fix should make it so that when xenos examine paper it no longer displays a blank page and the raw html was replaced with the formatted scrambled message. 

raw html is rather ugly

https://github.com/cmss13-devs/cmss13/assets/122310258/c279d8c8-e81e-477f-bcd0-cadc2bd32b4e

# Changelog

:cl:
fix: fixes xenos being able to view raw html on paper.
/:cl:


Cthulhu80:
For parity or anyone too lazy to view the comments on the old pr, they are as follows:

**You also need to check for quotation marks inside of HTML tags. Consider a tag such as <tag property=">Injection()" />. So when you are inside an HTML tag and run into a quotation mark (either single ' or double "), then you advance until you find the matching quotation mark. Perhaps you could add a couple booleans for this, e.g. need_single_quote and need_double_quote.
Maybe it would be a bit overkill but it's probably good to be safe.**

**Carefully-crafted user input would make this replace text inside of HTML tags. This should be reworked to be a bit more similar to how /proc/stars works, where the final message is being appended to in the original for loop. In this case, you could be appending scrambled_string from the loop before to a variable representing the final output. I suspect that would also be better performance-wise compared to repeatedly searching through the whole text for replacements.

Your current code also means that duplicates of unscrambled_message_key would get the same star text, while doing it iteratively would preserve the original behavior of not duplicating star text instances.**

code/modules/mob/mob_helpers.dm
```diff
@@ -131,6 +131,66 @@ GLOBAL_LIST_INIT(limb_types_by_name, list(
...
+		if(message[character_index] == ">")
+			parsing_message = TRUE
+			output_message += message[character_index]
+			continue
+		if(message[character_index] == "<")
+			parsing_message = FALSE
+			if(length(current_string_to_scramble))
+				var/scrambled_string = stars(current_string_to_scramble)
+				output_message += scrambled_string
+				current_string_to_scramble = null
```
Drulikar, line 186: This could just be
`current_string_to_scramble = ""`
(included in patch)

Drulikar (changes requested):
At the moment this works, though it would be nice for at least the <style> block to not get scrambled:

image

Instead of:

image


(Normal view:)
image

If you would like to try or adapt my experiment supporting this, `git apply` this patch: 
[patch.diff.txt](https://github.com/cmss13-devs/cmss13/files/13801282/patch.diff.txt)


Test procedure:
1. Load USS Runtime
2. `Spawn "faxmachine`
3. Send a blank fax
4. Admin reply to it with custom message w/ HTML
5. Inspect reply with a xeno

code/modules/mob/mob_helpers.dm
```diff
@@ -131,6 +131,71 @@ GLOBAL_LIST_INIT(limb_types_by_name, list(
...
+				output_message += current_char
+				continue
+
+		if(current_char == ">")
+			parsing_message = TRUE
+			output_message += current_char
+			if(current_tag == "<style" || findtext(current_tag, "<style ") == 1) // findtext because HTML doesn't care about anything after whitespace
+				escaped_tag = TRUE
+			else if(escaped_tag && (current_tag == "</style" || findtext(current_tag, "</style ") == 1)) // 1 for findtext because we only care about the start of the string matching
+				escaped_tag = FALSE
```
Drulikar, line 182:
```suggestion
			if(findtext(current_tag, "<style") == 1) // findtext because HTML doesn't care about anything after whitespace
				escaped_tag = TRUE
			else if(escaped_tag && findtext(current_tag, "</style") == 1) // 1 for findtext because we only care about the start of the string matching
				escaped_tag = FALSE
```
Case insensitive and just cleaner.

Drulikar (changes requested):

Drulikar (approved):

PR #5338, opened 2023-12-29
Fixes Nuke failing to detonate

fira:
# About the pull request

Under very specific circumstances (seen 2x or more on live), the nuke broken logic of checking eligbile mobs for explosion will cause it to crash completely and fail to detonate. The problem is primarily the `loc` check at top, which is not a sufficient safeguard.

Also moves the explosion delay so that people die when the nuke explode and not before...

Spawn in nuke and testers, safety off, proccall explode

# Changelog
:cl:
fix: Fixed Nuke failing to detonate under specific circumstances. It may involve a little rodent.
fix: Nuke explosion will now kill mobs on its actual explosion rather than at the start of the cinematic sequence.
/:cl:


Drulikar (approved):
Code looks fine and looks like it still works, but I am unable to reproduce the issue on master.

Drulikar:
Was able to reproduce and confirm it works.

PR #5342, opened 2023-12-30
Fixes mobs in crit being able to understand other languages 

### LINKED ISSUE #3778: You can understand Xenos when you're about to die.
### Testmerges

#3644 #3676 #3690 #3755

### Description of the bug

Screenshot_14
as you can see here, I could understand Xenos when I was about to die as a human.

A human shouldn't understand what xenomorphs are talking about, I guess...

1. Get closer to losing consciousness before you die.
2. Make some Xenomorphs talk close to your body.
3. Observe the chat.

### Issue Bingo

Cthulhu80:
# About the pull request

Fixes #3778 , mobs should no longer be able to comprehend other languages while in crit. 

bug bad

# Changelog

:cl:
fix: Fixes mobs being able to understand other languages while in crit. 
/:cl:

### DIFF
--- a/code/modules/mob/hear_say.dm
+++ b/code/modules/mob/hear_say.dm
@@ -202,18 +202,21 @@
 
 /mob/living/hear_say(message, verb, datum/language/language, alt_name, italics, mob/speaker, sound/speech_sound, sound_vol)
 	if(client && mind && stat == UNCONSCIOUS)
-		hear_sleep(src, message, src == speaker, Adjacent(speaker))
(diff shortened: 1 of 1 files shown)


Drulikar (approved):
Doesn't make it impossible but code looks fine.

Cthulhu80:
> Doesn't make it impossible but code looks fine.

I mean, just makes it the same as when a living human mob hears a xeno. Same probability of asterisks.

PR #5357, opened 2024-01-01
Fix OT explosive assemblies not taking containers they are supposed to

JackieEstegado:
# About the pull request

OT assemblies were not accepting added 120u containers, even though it should be taking anything up to it.
Due to some weird order of operation that goes against DM documentation.
For example here the (M15) assembly is not taking a 120u bucket.

Screenshot_31
This is essentially evaluating: 120+(0>120)
which equals 120, since its not 0 or a negative number it interprets it as True.

Pre-open source bugs in the code is bad.

<details>

</details>

# Changelog
:cl:
fix: Fixed a bug where OT explosive assemblies were not accepting reagent containers when they should have.
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/explosive.dm
+++ b/code/game/objects/items/explosives/explosive.dm
@@ -150,7 +150,7 @@
 			return
 		else
 			if(W.reagents.total_volume)
-				if(W.reagents.maximum_volume + current_container_volume > max_container_volume)
+				if((W.reagents.maximum_volume + current_container_volume) > max_container_volume)
 					to_chat(user, SPAN_DANGER("\the [W] is too large for [name]."))
 					return
 				if(user.temp_drop_inv_item(W))


SabreML:
I tried testing it myself since it seemed like a really weird bug, and for whatever reason I don't seem to be able to reproduce the issue on my end.

Testing on a clean master branch:

https://github.com/cmss13-devs/cmss13/assets/57483089/e19606b2-59fa-4316-b51f-ad0c1d2d987e

Drulikar (changes requested):
I don't think parenthesis are solving a problem here - I suspect more likely one of arguments sometimes is not exactly those integers you expect. The order of operations in the reference would not be changed by what you've done, nor do any tests report anything of this kind; and I really don't think most things would work if the operator order functioned that way. 
image

If you are able to reproduce the problem reliably, experiment with https://www.byond.com/docs/ref/#/proc/round

More information about how byond handles numbers (floats) is documented here: https://www.byond.com/docs/ref/#/{notes}/numbers

JackieEstegado:
Seems to have been a 514 bug. Works fine on 515.10

PR #5358, opened 2024-01-01
Character name max length increase by 2.

blackdragonTOW:
# About the pull request
Bumping the max name length by two.  I've seen a few people who drop the quotations on their nickname to save characters to fit the name they want.  This would help allow them to have quotations on their nick again.

Letting people have the names they want without having to gimmick the quotation character count.

<details>

</details>

# Changelog
:cl:
qol: increased name character limit by two.
/:cl:


fira:
I don't know, it's bit of an endless chase, there's no reason adding 2 more will stop people from doing that with longer names

blackdragonTOW:
> I don't know, it's bit of an endless chase, there's no reason adding 2 more will stop people from doing that with longer names

100% there's genuinely nothing stopping that from happening, and could totally be a "just one more lane bro" type fix but I keep feeling really sad when I see un-quoted nicknames due to character limits.

realforest2001:
I, and I think a few others, would like to see the character name field split into three, for forename, nickname and surname. That would allow automatic quotation marks to be included too. Problem with that is balancing lengths becomes harder, without forcing it to be checked to a max limit after assembly. Splitting the fields is probably preferable to arbitrarily adding two characters to the limit though.

Birdtalon:
So I'm going to merge this for this reason and I get that would be frustrating for someone's name not to be able to fit. But ideally what Forest suggests above is the more maintainable outcome in the long term.

> > I don't know, it's bit of an endless chase, there's no reason adding 2 more will stop people from doing that with longer names
> 
> 100% there's genuinely nothing stopping that from happening, and could totally be a "just one more lane bro" type fix but I keep feeling really sad when I see un-quoted nicknames due to character limits.

Birdtalon (approved):

PR #5360, opened 2024-01-01
Splits hivemind messages into their own chat category

SabreML:
# About the pull request

Separates xeno hivemind chat messages into their own toggleable category.

(It seems like TGUI-chat handles a new category being added by just making it enabled for everything, so it'll need to be manually disabled in custom 'OOC'/'Admin'/etc. tabs. (although better that than having it be *disabled* by default))

A lot of the time when observing it can be very hard to make out what's being said in the hivemind chat with how similar its colour is to local xeno chat, charlie radio, JTAC radio, colony radio, etc., especially if chat is moving quickly. This doesn't *really* solve that, but it does make it possible to set up a custom chat tab just for hivemind/xeno messages. (Or to stop seeing the hivemind I guess, if someone wanted to do that.)

<details>

https://github.com/cmss13-devs/cmss13/assets/57483089/5a5f0515-3712-46d3-88c3-24d662da044d

</details>

# Changelog
:cl:
ui: Separated xeno hivemind chat messages into their own toggleable category, separate from 'Radio'.
/:cl:

### DIFF
--- a/code/__DEFINES/chat.dm
+++ b/code/__DEFINES/chat.dm
@@ -6,6 +6,7 @@
 #define MESSAGE_TYPE_SYSTEM "system"
 #define MESSAGE_TYPE_LOCALCHAT "localchat"
 #define MESSAGE_TYPE_RADIO "radio"
+#define MESSAGE_TYPE_HIVEMIND "hivemind"
 #define MESSAGE_TYPE_INFO "info"
 #define MESSAGE_TYPE_WARNING "warning"
 #define MESSAGE_TYPE_DEADCHAT "deadchat"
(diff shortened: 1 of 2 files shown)


fira (approved):
I was gonna suggest using message flags instead but oh my god xeno hivemind uses show_message..........

SabreML:
> I was gonna suggest using message flags instead but oh my god xeno hivemind uses show_message..........

Yeah it's uh... It works?

fira:
Using message type in to_chat is just faster and less ambiguous

realforest2001 (approved):

Birdtalon (approved):

PR #5364, opened 2024-01-01
Adds prescription values to eyewear that did not have them previously

Staykeu:
## About the pull request

Adds prescription lens effects to items such as the SG/Scout/Sniper sights, also removes old unused prescription variants of the med, sec, sci and sensor HUDs and replaces them with the default version supporting prescription effects.

## Explain why it's good for the game

QoL change that makes life easier for those that have characters with impaired vision, allows the use of these eyepieces without obscuring the mask slot, be it reserved for smokes, scarfs, coifs, etc.
<details>

N/A

</details>

# Changelog
:cl: Stakeyng
qol: combines prescription glasses and eyewear for varying roles
/:cl:


567Turtle:
just add `prescription = true` to `/obj/item/clothing/glasses/hud/` instead of adding it to each hud individually

567Turtle:
stinky

567Turtle:
actually wait I'm wrong, you added `prescription = true` to `/obj/item/clothing/glasses/hud/` so I'm pretty sure you don't have to add it to all the other branches of `/obj/item/clothing/glasses/hud/`

567Turtle:
Correct me if I'm wrong but you don't need to have `prescription = TRUE` in
`/obj/item/clothing/glasses/night/m42_night_goggles/spotter`, 
`/obj/item/clothing/glasses/night/m42_night_goggles/upp` 
and `/obj/item/clothing/glasses/night/m42_night_goggles/m42c`
because you already gave `prescription = TRUE` to `/obj/item/clothing/glasses/night/m42_night_goggles`

Additionally, I would just add `prescription = TRUE` to `/obj/item/clothing/glasses/hud` and then add `prescription = FALSE` to the huds that you DONT want to have a prescription (which is only one as of writing this comment, the jensen shades). This would remove the need to restate `prescription = TRUE` 15 different times for each and every branch off of hud.

567Turtle:
`/obj/item/clothing/glasses/night/m42_night_goggles/upp` and `/obj/item/clothing/glasses/night/m42_night_goggles/m42c` still have an unnecessary `prescription = TRUE`

567Turtle:
The security jensen glasses have prescription with the current code. If you don't want them to (you didn't previously) add `prescription = FALSE` to them.

Staykeu:
> The security jensen glasses have prescription with the current code. If you don't want them to (you didn't previously) add `prescription = FALSE` to them.

i didn't add them because it was an unnecessary addition to an item that gets rarely used, same goes for the second update, it simply doesnt matter

PR #5314, opened 2023-12-27
Adds a 'Time Since Death' counter to the status tab

SabreML:
# About the pull request

Adds a 'Time Since Death' counter to the status tab, which shows in minutes and seconds how long it's been since the player died.
(The counter isn't shown to players who joined as observers, just ghosts.)

(also for the record I asked and this is fine in regards to the freeze: https://discord.com/channels/150315577943130112/745447048261795890/1189329387397578912)

# Explain why it's good for the game

Primarily, it would be very useful for tracking how long until it's possible to spawn as a lesser drone/facehugger without having to keep clicking the buttons to see the failure message.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

**Normal:**
after normal
<hr>

**With larva queue:**
after with queue
<hr>

**With hijack:**
after with hijack
<hr>

**With both:**
after
<hr>

</details>

# Changelog
:cl:
qol: Added a 'Time Since Death' counter to the status tab for ghosts.
/:cl:

PR #5374, opened 2024-01-03
'Time Since Death' hour wrapping fix

SabreML:
# About the pull request

Tweaks the 'Time Since Death' counter added in #5314 so that it doesn't wrap around to '00:00' when it reaches one hour, instead adding an hour counter to the number.

I tried to see if I could make the minutes counter continue above 60 instead, but the best method I could figure out involved writing a whole time parsing thing, plus something to add zero padding so that it would display '61:02' and not '61:2', so I decided to just add a `>=` check instead.

Technically this does just push the wrapping threshold back a bit rather than actually fixing it, but if a round's been going for more than 24 hours then there's probably bigger problems.

Less confusing for ghosted players.

<details>

https://github.com/cmss13-devs/cmss13/assets/57483089/cdc88e6b-ff1d-4a5a-a04f-b6f7a8abf4c4

(Runtime is unrelated, something to do with the DB I think?)

</details>

# Changelog
:cl:
fix: Made the 'Time Since Death' counter show hours if it passes 1 hour, rather than wrapping the minutes back to 0.
/:cl:

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -1235,17 +1235,11 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 			. += ""
 
 	if(timeofdeath)
-		var/time_since_death = debug_var
+		var/time_since_death = world.time - timeofdeath
 		var/format = (time_since_death >= 1 HOURS ? "hh:mm:ss" : "mm:ss")
 
 		. += "Time Since Death: [gameTimestamp(format, time_since_death)]"
 
-/mob/dead/observer/var/debug_var
-/mob/dead/observer/verb/debug_thing()
-	set category = "Ghost"
-	set name = "debug thing"
-
-	debug_var += 10 MINUTES
 
 /proc/message_ghosts(message)
 	for(var/mob/dead/observer/O as anything in GLOB.observer_list)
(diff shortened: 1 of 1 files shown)


code/modules/mob/dead/observer/observer.dm
```diff
@@ -1235,7 +1235,10 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 			. += ""
 
 	if(timeofdeath)
-		. += "Time Since Death: [duration2text_sec(world.time - timeofdeath)]"
+		var/time_since_death = world.time - timeofdeath
+		var/format = (time_since_death >= 1 HOURS ? "hh:mm:ss" : "mm:ss")
+
+		. += "Time Since Death: [gameTimestamp(format, time_since_death)]"
```
Drulikar, line 1241: I don't think you want to use gameTimestamp instead of just time2text because I don't think we want the timezone accounted for like gameTimestamp does. All other game calculations against timeofdeath are just against world.time.

Also your example video doesn't show the counter moving any when it should be growing every second time passes.

You can simply right click your mob, view variables, and edit your time of death value to a low number (meaning far in the past including negative) for testing purposes rather than needing to code a temporary debugging verb for this.

Drulikar (changes requested):

code/modules/mob/dead/observer/observer.dm
SabreML: Oh for some reason I didn't think you could VV edit something to a negative value. That definitely would've made testing a lot easier.

Drulikar (approved):
Code changes look fine.

PR #5388, opened 2024-01-06
Makes TGUI transfer over when changing bodies

SabreML:
# About the pull request

Makes any open TGUI windows transfer over when a player's mind is transferred to a new body. *(At least if it's done through `/datum/mind/proc/transfer_to()`)*
This is mainly useful for xeno players who want to evolve, since it means they won't need to re-open any interfaces they had open.

The changes in `code/controllers/subsystem/minimap.dm` are there so that the tacmap image remains visible to the player after a transfer, since the client changing bodies seems to remove it from their [`screen`](https://www.byond.com/docs/ref/#/client/var/screen) list.

(This *could* have hidden side effects of interfaces staying open when they shouldn't, so I'd appreciate this being testmerged for a bit just in case.)

QOL addition so that you don't have to manually re-open them yourself.

<details>

https://github.com/cmss13-devs/cmss13/assets/57483089/70549c92-915b-4881-95ca-d1625a346f4b

</details>

# Changelog
:cl:
qol: Made open TGUI windows transfer over when a xeno player evolves.
/:cl:


fira:
Noticed that before, I think this should just flat out be in transfer_to. Even NanoUI has UI transfer already there, it just never was done for TGUI, unless there's some obscure reason it doesn't work...

SabreML:
> Noticed that before, I think this should just flat out be in transfer_to. Even NanoUI has UI transfer already there, it just never was done for TGUI, unless there's some obscure reason it doesn't work...

It probably would be better to try that instead, although I'm not sure how I could check for an open tacmap from `/datum/mind` instead of `/mob/living/carbon/xenomorph`. (The map needs to be re-registered, otherwise the ui just shows a black screen where it should be.)

fira:
Could hook COMSIG_MOB_NEW_MIND

SabreML:
I've rewritten the PR in the latest commit with your suggestions since they made a lot of sense, and from some more testing everything seems to be working the same. (The failed checks are from a github server error)

harryob (approved):
we'll see how it goes

PR #5409, opened 2024-01-09
Fixes stack of flags anchoring after placing one down

### LINKED ISSUE #5145: Anchoring bug for "/obj/item/stack/flag"
### Testmerges

#4986,#4939, #5054, #4842, #5067, #5018, #5127, #5134

### Round ID

19833

### Description of the bug

Placing flags from a stack of flags will lead to the remaining flags in a stack to turn anchored when you drop or throw the stack of flags.

The stack of flags should not anchor themselves.

1. Place flags into the ground from /obj/item/stack/flag/
2. Keep a few flags in hand.
3. Drop or throw away remaining flags.
4. See how the flags will remain anchored and you cannot unanchor them outside of VV, with exception of the place flags.
5. 
bilde
bilde
bilde

...

### Issue Bingo

Vicacrov:
# About the pull request

Currently, when you put down a flag, the stack of flags in your hand becomes anchored instead of the placed down flag. Fixes #5145

The actual flag gets anchored, not the stack of flags in your hands.

<details>

https://github.com/cmss13-devs/cmss13/assets/49321394/fb9092a5-9c2e-4d86-a591-e9f793a62a2e

</details>

# Changelog

:cl:
fix: Placing flags now anchors the actual flag, not the stack of flags in your hand.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/flags.dm
+++ b/code/game/objects/items/stacks/flags.dm
@@ -71,7 +71,7 @@
 	var/obj/item/stack/flag/newflag = new src.type(T)
 	newflag.amount = 1
 	newflag.upright = TRUE
-	anchored = TRUE
+	newflag.anchored = TRUE
 	newflag.name = newflag.singular_name
 	newflag.icon_state = "[newflag.base_state]_open"
 	newflag.visible_message("<b>[user]</b> plants [newflag] firmly in the ground.")


fira (approved):
This is genius

Birdtalon (approved):

PR #5414, opened 2024-01-09
Brings back UPP Random Names for the UPP Survs.

BadAtThisGame302:
# About the pull request

Title.

I got inspiration from last round's Trijent Dam's UPP Insert, where I decided to make this PR because UPP with English names is not roleplay friendly or calling someone "Comrade Jeff" or "Master Serzhant Jack" make sense especially because the UPP is meant to disassociate you with your previous character and turn you into a different faction who completely play differently and act differently.  

The one thing I guess I can see is that you know if your fellow survivors are bald or not, but that's not really a reason that should be taken into account, realistically. Apart from that I really don't see a positive in having your character name be the UPP Survivors name, unless you RP as a UPP remnant or deserter. 

This goes for especially flavor texts that are meant for Marines or for Survs, like "USCM's Greatest Hero" is a flavor text I saw one time on a UPP Survivor, doesn't really make sense does it?

<details>

</details>

# Changelog
:cl:
code: Brought back the UPP Survivor random name generator for the Trijent Dam UPP Nightmare Insert.
/:cl:


Steelpoint:
I should note that the reason this was implemented was that the previous head maintainer mandated the change, stating that they believed a player should not have their choice of name taken away from them for a survivor roll. Their preferred change would be if the game offered the player the ability to selecte a non-USCM name for these kinds of roles instead.

irakligabunia:
"The one thing I guess I can see is that you know if your fellow survivors are bald or not, but that's not really a reason that should be taken into account, realistically. Apart from that I really don't see a positive in having your character name be the UPP Survivors name, unless you RP as a UPP remnant or deserter."
People like playing a character they want to and for that reason I think this PR would be handling the majority because of the minority

BadAtThisGame302:
> "The one thing I guess I can see is that you know if your fellow survivors are bald or not, but that's not really a reason that should be taken into account, realistically. Apart from that I really don't see a positive in having your character name be the UPP Survivors name, unless you RP as a UPP remnant or deserter." People like playing a character they want to and for that reason I think this PR would be handling the majority because of the minority

I would be inclined to agree, if not for the completely different faction and the completely different way the UPP is supposed to act from the USCM. Not the mention if you know or hang out with some people and are a UPP surv, they'll talk to you like you are best buds, even though your factions are in a cold war and rarely have fraternization amongst eachother.

harryob (changes requested):
as before, i still would rather we had a box in the "Special Roles" tab in character preferences. if you need it, i can help out with this

PR #4588, opened 2023-10-04
Nerfs sticky resin

Mister-moon1:
# About the pull request
Adds a build cap to sticky resin

# Explain why it's good for the game
Sticky resin right now is pretty awful ingame, not because it's bad but because it promotes the act of hivelords spending 20 minutes placing sticky resin on every single tile in caves, it should be used more strategically and develop where the combat is, rather than coating every square inch in 5 layers of it. 

Overall it will provide a more pleasant time for marines, increase the usage of hive clusters for the slowdown of hiveweeds and it will encourage builders to play more strategically

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl:
balance: Added a build cap to sticky resin of 25
/:cl:

PR #5416, opened 2024-01-10
Maximum Limit to Sticky Resin

blackdragonTOW:
# About the pull request

Caps Sticky Resin to 50 per xeno.  

50x50 wide fields of sticky+node is obnoxious to deal with both as a marine moving through it, and as a builder who has to tear down the field to be able to build walls.

<details>

</details>

# Changelog
:cl:
balance: Max Deployable Sticky Resin set to 50.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/resin_constructions.dm
+++ b/code/modules/mob/living/carbon/xenomorph/resin_constructions.dm
@@ -239,6 +239,7 @@ GLOBAL_VAR_INIT(resin_lz_allowed, FALSE)
 	construction_name = "sticky resin"
 	cost = XENO_RESIN_STICKY_COST
 	build_time = 1 SECONDS
+	max_per_xeno = 50
 
 	build_path = /obj/effect/alien/resin/sticky
 


Birdtalon:
Thank you for the PR, I do have some comments though.

#4588 attempted to do the same thing and was closed by Zonespace back in October 23 for the following reasons:

- Hivelords spending N minutes filling rooms with sticky resin is N minutes spent not building the things that actually matter, walls and fortifications. If they choose to spend their time building a more situationally-dependent room, then so be it.
- Sticky resin is incredibly cheap to destroy, being destroyed in 1 hit from almost every melee weapon and any explosion.
- It isn't very good UX to have to keep track of 25 cheap, destructible turf objects and destroy them if they're placed poorly (resin spikes are not an exception to this either)

I see that your limit is 50 instead of 25 from #4588 but please may you explain why this PR should be merged when it doesn't address any of the above feedback except for increasing the number of built resins?

blackdragonTOW:
> Thank you for the PR, I do have some comments though.
> 
> #4588 attempted to do the same thing and was closed by Zonespace back in October 23 for the following reasons:
> 
> > * Hivelords spending N minutes filling rooms with sticky resin is N minutes spent not building the things that actually matter, walls and fortifications. If they choose to spend their time building a more situationally-dependent room, then so be it.
> > * Sticky resin is incredibly cheap to destroy, being destroyed in 1 hit from almost every melee weapon and any explosion.
> > * It isn't very good UX to have to keep track of 25 cheap, destructible turf objects and destroy them if they're placed poorly (resin spikes are not an exception to this either)
> 
> I see that your limit is 50 instead of 25 from #4588 but please may you explain why this PR should be merged when it doesn't address any of the above feedback except for a larger number of built resins?

Sure!
Firstly, I never knew about that specific PR.
Secondly, in our current state of Feature Freeze there is no feasible way to overhaul builders without violating the lock.  This solves the problem without a huge overarching overhaul.
Thirdly, we've historically preferred mechanical changes to things rather than requiring admin intervention.  And in cases where we've had a builder coat a quarter of the map in buildables like this we've tapped them on the shoulder to tell them to calm down. (Feel free to chat up one of the hosts).  This would prevent such situations from happening where even zoomed out ghost view cannot see the edges of the sticky/node field. 
Fourthly, there's no good justifiable game reason for the mass fields beyond "It's funny." It creates a detrimental game experience that simply isn't fun for either side. 
Finally, the cap is high enough that if you're using sticky resin in a deliberate targeted manner you should never bump into the cap, but if you're just filling turfs for the sake of filling turfs, you'll bump into the cap after filling your local choke (which is what you should be doing with them in the first place).

That said to address the points directly.  The limit should be high enough that you shouldn't need to mentally track how many you have if you're not spamming them.   The cheapness of destruction conflicts with spikes, but still "one attack" is a fact for sure, but imagine that across the entirety of lambda labs on every single turf.  As for Hivelords building where they please, they still can, and the limit will nudge them towards working on productive defenses rather than things that are "funny".

Drulikar (changes requested):
Is it possible you could implement this in such a way that when you place your 51st resin, the oldest is replaced?

This likely means adding a new var to `/datum/resin_construction` to indicate this type of resin can be replaced when capped.

Then in `/mob/living/carbon/xenomorph/proc/build_resin` before hitting `return SECRETE_RESIN_FAIL` you instead check if the resin allows replacement of old - and if so you delete the oldest (likely first entry of you'd have to see) item in built_structures[RC.build_path] to skip returning `SECRETE_RESIN_FAIL`?

Technically this can result in their 51st deleting a sticky and then should the placement of the new fail for whatever reason they're now at 49 but its fine; we don't need to restore the old one.

PR #5424, opened 2024-01-11
refactor: turned machinery of distillery/main.dm into "props"

Huffie56:
move all the machinery of distillery/main.dm into machinery.dm and remove all their added functionality.
# About the pull request

# Explain why it's good for the game
This is obsolete code from an unfinished project need to go to clear the code.
# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
del: deleting machinery called centrifuge in distillery/main.dm and deleted the distellery/main.dm file and folder.
refactor: move all the machinery of distillery/main.dm into machinery.dm and remove all their added functionality.
/:cl:

PR #5425, opened 2024-01-11
remove all of machinery from distillery/main.dm from all the maps exept big red.

Huffie56:
# About the pull request
Part of the plan to remove those obsolete machinery those must be purged...

Part of the plan to remove those obsolete machinery
<details>

</details>

# Changelog
:cl:
maptweak: remove all of machinery from distillery/main.dm from all the maps except big red.
/:cl:

### DIFF
--- a/maps/map_files/DesertDam/Desert_Dam.dmm
+++ b/maps/map_files/DesertDam/Desert_Dam.dmm
@@ -12676,7 +12676,7 @@
 /turf/open/floor/plating,
 /area/desert_dam/exterior/valley/valley_crashsite)
 "aMk" = (
-/obj/structure/machinery/mill,
+/obj/structure/machinery/autolathe,
 /turf/open/floor/plating,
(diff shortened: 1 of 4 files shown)


Nanu308 (changes requested):
They are still used as map props for looks, change the machinery to keep the icon/desc and name then swap out the filepath for something else.

Huffie56:
> They are still used as map props for looks, change the machinery to keep the icon/desc and name then swap out the filepath for something else.

i am not sure what you mean there.....
but this PR is to remove them all from maps that don't have a PR on them currently.

I already have a PR to turn them into prop.(tackling this file from two separate angle...)
https://github.com/cmss13-devs/cmss13/pull/5424

so if you think we should keep them as prop (when they use auto-lathe sprite and just have a different name and description)
just tell me i will close this PR it's fine...

Huffie56:
closing because they will just be turn into prop not removed totally...

PR #5450, opened 2024-01-13
Alamayer Perma chair faces the right way

private-tristan:
# About the pull request
chairs in permabrig now face south (towards their table) instead of north, towards the wall

perma prisoners should be able to sit at their chair and be facing the paper and pen instead of the wall??
# Changelog
:cl:
maptweak: Alamayer PermaBrig Chairs face the right way
/:cl:


Nanu308 (approved):
lgtm

PR #5464, opened 2024-01-15
Fixes mobs unable to move after being flung while buckled.

### LINKED ISSUE #5420: Character unable to move/not responding to keyboard movement inputs
### Testmerges

N/A to my knowlage

### Round ID

20343

### Description of the bug

During evac, I had died to a ravager while reviving another marine, during my mauling I was unable to move and subsequently died. I was revived by another medic on the trip back up to the Almayer where I was unable to move but could equip/unequip and look around directionally with my cursor. I had to be dragged off the alamo and after a moment or two of struggling I pinged an admin who suggested fixes that unfortunately had no effect on my problem. buckling to a folding chair nearby, disconnecting and rejoining, and giving them control of my character to see if it was a client-side issue. eventually I requested the admin delete/cryo my character body so I could sit out the hijack where I found my ghost was able to freely move about unlike my body. I'm not sure if its unrelated or not but I have been experiencing beyond crashes whenever a crate landed at fob for this round. 
_Beyond version 515.1625 stable branch_

I would assume expected behavior would be my character moving in response to my keyboard input WASD ect.

1. die on alamo before takeoff, specifically one tile to the left of the last left inboard row of seats
2.have your revive defib interrupted by the alamo landing then be revived successfully after which.
3.I don't know if its reproduceable but this was pretty much the string of events the "anchored" me
...

### Issue Bingo

Birdtalon:
# About the pull request

Buckled mobs didn't have their `TRAIT_IMMOBILIZED` removed properly since we return early from `launch_towards()` if `buckled`
https://github.com/cmss13-devs/cmss13/blob/939e283561d2446d5df4b60faba08097eaea1f79/code/modules/mob/living/living.dm#L401-L405

Therefore any callbacks in end_throw_callbacks are never invoked. In the case of xeno throws we rely on end_throw_callbacks to remove our immobilised trait

Fixes #5420

<details>

</details>

# Changelog
:cl:
fix: Fixed buckled mobs unable to move after being thrown by a xeno.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -405,6 +405,7 @@
 		return
 	if(buckled)
 		LM.invoke_end_throw_callbacks(src)
+		return
 	if(pulling)
 		stop_pulling() //being thrown breaks pulls.
 	if(pulledby)
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -401,9 +401,10 @@
 /mob/living/launch_towards(datum/launch_metadata/LM)
 	if(src)
 		SEND_SIGNAL(src, COMSIG_MOB_MOVE_OR_LOOK, TRUE, dir, dir)
-	if(!istype(LM) || !LM.target || !src || buckled)
-		LM.invoke_end_throw_callbacks(src)
+	if(!istype(LM) || !LM.target || !src)
 		return
+	if(buckled)
+		LM.invoke_end_throw_callbacks(src)
 	if(pulling)
 		stop_pulling() //being thrown breaks pulls.
 	if(pulledby)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/living.dm
```diff
@@ -402,6 +402,7 @@
 	if(src)
 		SEND_SIGNAL(src, COMSIG_MOB_MOVE_OR_LOOK, TRUE, dir, dir)
 	if(!istype(LM) || !LM.target || !src || buckled)
+		LM.invoke_end_throw_callbacks(src)
```
Drulikar, line 405: Are you certain all the conditions that were returning for here allow us to be able to even call the end throw callbacks? E.g. if the istype fails, will it even be something that implements this proc? And are there any callbacks added to the end throw callbacks that need src (move_atom) or the target to not be falsey?

You probably only want the buckled return, not the others.

Drulikar (changes requested):

code/modules/mob/living/living.dm
Birdtalon: I admit I didn't even think of that I saw buckled and just whacked it in there.

Drulikar (approved):
Code changes look fine.

PR #5469, opened 2024-01-16
Adds a semi-big whistle cooldown.

ClairionCM:
Adds a 10 second cooldown to the whistles to make them less spammy
# About the pull request

This PR gives the whistles a 10 second cooldown to have people spam them less. It also sends a message telling the player how long they need to wait before they can use it again.

Right now what often ends up happening is people deleting all of the whistles because they are VERY loud and VERY spammy. This should make that less of an issue.

<details>
It works :D

image

</details>

# Changelog

:cl:
qol: made whistles less spammy
/:cl:

### DIFF
--- a/code/game/objects/items/devices/whistle.dm
+++ b/code/game/objects/items/devices/whistle.dm
@@ -30,7 +30,7 @@
 
 /obj/item/device/whistle/proc/whistle_playsound(mob/user)
 	if(!COOLDOWN_FINISHED(src, spam_cooldown))
-		to_chat(user, SPAN_DANGER("\ You are out of breath after using [src]! Wait [COOLDOWN_SECONDSLEFT(src, spam_cooldown)] second(s)."))
+		to_chat(user, SPAN_DANGER("You are out of breath after using [src]! Wait [COOLDOWN_SECONDSLEFT(src, spam_cooldown)] second\s."))
 		return
 
 	user.visible_message(SPAN_WARNING("[user] blows into [src]!"))
--- a/code/game/objects/items/devices/whistle.dm
+++ b/code/game/objects/items/devices/whistle.dm
@@ -8,7 +8,8 @@
 	actions_types = list(/datum/action/item_action)
 
 	var/volume = 60
-	var/spamcheck = 0
+	var/spam_cooldown_time = 10 SECONDS
+	COOLDOWN_DECLARE(spam_cooldown)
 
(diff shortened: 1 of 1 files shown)


code/game/objects/items/devices/whistle.dm
```diff
@@ -28,14 +29,14 @@
 		..()
 
 /obj/item/device/whistle/proc/whistle_playsound(mob/user)
-	if (spamcheck)
+	if(!COOLDOWN_FINISHED(src, spam_cooldown))
+		to_chat(user, SPAN_DANGER("\ You are out of breath after using [src]! Wait [COOLDOWN_SECONDSLEFT(src, spam_cooldown)] second(s)."))
```
Zonespace27, line 33:
```suggestion
		to_chat(user, SPAN_DANGER("You are out of breath after using [src]! Wait [COOLDOWN_SECONDSLEFT(src, spam_cooldown)] second\s."))
```

Zonespace27 (changes requested):

code/game/objects/items/devices/whistle.dm
ClairionCM: Thanks Zonespace. I dont know how to code!

Drulikar (approved):
Code changes look fine.

Antixenobuff:
We have no problem with whistle spamming. Plus, it's always funny and encouraging to see one SL spamming his whistle after a bad Queen scree/OB to indicate all marines to hard push that zone. It makes you feel actually in a warzone and adds non verbal communication. One whistle every 10 seconds will remain almost unnoticed. If what worries you it's someone spamming the whistle with zero reason, there are two options: shove and take his whistle or ahelp it (that's the reason admins are). You can aslo do both.

Wintermote:
You can solve this in a better way by emulating what was done with chat spam: Too much use within a defined period, therafter introduces a cooldown. e.g. use it 5x in 10 seconds but the 6th time has a 10 second cooldown.

That allows you to spam the whistle when necessary but alleviates someone just sitting in the dropship blowing the whistle for 10 minutes straight.

ClairionCM:
Yeah I know the cooldown being on the item is weird and I'll try to fix it when I manage to figure it out. I'm really bad at coding and also really busy most of the time so that might take a while.
But yeah I00% agree that it should be tied to the mob instead of the whistle object. The issue has basically been solved since now its really impractical to spam whistles but I should fix it even if its just to clean up the code a bit more.

PR #5475, opened 2024-01-16
Fixes the equipment attach point being on the wrong plane

Vicacrov:
# About the pull request

Currently, the equipment attach points (the ones on the Alamo and the Normandy) are on the basic atom layer, -6. This causes certain effects, such as hive clusters or egg morphers layer below it (that are on `FLOOR_PLANE`, -7).

See the images for a before and after.

No weird layering. Hive clusters are now visible when placed on the same tile as attach points.

<details>

Before: 

image

After:

image

</details>

# Changelog
:cl:
fix: Special xeno structures no longer layer below the Alamo and Normandy's equipment attach points.
/:cl:

### DIFF
--- a/code/modules/dropships/attach_points/attach_point.dm
+++ b/code/modules/dropships/attach_points/attach_point.dm
@@ -7,6 +7,7 @@
 	unacidable = TRUE
 	anchored = TRUE
 	layer = ABOVE_TURF_LAYER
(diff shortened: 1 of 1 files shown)


harryob (approved):
oh, yeah. shouldn't be on that one regardless because of ambient occlusion, good catch

PR #5481, opened 2024-01-17
Fix m56d and m2c crushing, labeling, and damage retention

### LINKED ISSUE #5477: M56Ds inmune to crusher charge
### Testmerges

#5086 #5337 #5228 #5007 #5407 #5435 #5397 #5469 #5280 #5464

### Round ID

20433

### Description of the bug

M56D machineguns seem to be inmune to momentum-based charger crusher charges.

Used to be able to crush them down with enough momentum just as M2Cs, not anymore.

1. Be charger
2. Have assembled M56D
3. Charge it max speed
...

### Issue Bingo

Drulikar:
# About the pull request

This PR addresses 5 things regarding the m56d and m2c:
1. Both now properly retain labeler labels (though technically the m56d post can lose its label - gun takes priority)
2. The m56d now retains its damage when disassembled
3. The m56d's post now has an implementation for handle_charge_collision so they can croosh not fully assembled guns
4. It appeared possible to duplicate the m56d when a charger crushed it if it both was destroyed by the impact and the update_health resulted in the gun being refunded; this is no longer possible since the charge logic returns early if it is destroyed
5. Vehicles running over a m56d post now undeploys the m56d correctly if it had a gun attached (it would downgrade to just a post)

Fixes #5477 but I don't think this *ever* worked. Rather the issue is purely when the gun is in this state: 
image

Additionally, if we're going to have a mechanic to weld to fix the gun; you shouldn't be able to circumvent this by simply disassembling it... However, since the destruction of the m56d didn't have a guaranteed chance to destroy the gun, technically this results in a fully repaired gun again if you can make another post for it. It would need to work more like the m2c if we wanted it to remain in a broken state in this situation.

https://youtu.be/rg1mLFjhsEY (though I forgot to record crowbar disassembly for the m56d too)

# Changelog
:cl: Drathek
fix: The M56D and the M2C now retain their labels when disassembled
fix: The M56D can now be charger crushed when not fully assembled
fix: The M56D being run over by a vehicle now undeploys the gun correctly (so it retains the gun if it had a gun)
fix: Fix an edge case that could duplicate the M56D when charger crushed
balance: The M56D now retains the health of the gun through all of its different disassembled states
/:cl:

### DIFF
--- a/code/modules/cm_marines/smartgun_mount.dm
+++ b/code/modules/cm_marines/smartgun_mount.dm
@@ -137,13 +137,14 @@
 			to_chat(user, SPAN_WARNING("This is too close to [machine]!"))
 			return
 
-	var/obj/structure/machinery/m56d_post/M = new /obj/structure/machinery/m56d_post(user.loc)
-	M.setDir(user.dir) // Make sure we face the right direction
-	M.gun_rounds = src.rounds //Inherit the amount of ammo we had.
-	M.gun_mounted = TRUE
-	M.anchored = TRUE
-	M.update_icon()
-	transfer_label_component(M)
+	var/obj/structure/machinery/m56d_post/post = new(user.loc)
+	post.setDir(user.dir) // Make sure we face the right direction
+	post.gun_rounds = src.rounds //Inherit the amount of ammo we had.
+	post.gun_mounted = TRUE
+	post.gun_health = health // retain damage
+	post.anchored = TRUE
+	post.update_icon()
+	transfer_label_component(post)
 	to_chat(user, SPAN_NOTICE("You deploy \the [src]."))
 	qdel(src)
 
@@ -223,8 +224,8 @@
 		return
 
 	to_chat(user, SPAN_NOTICE("You deploy \the [src]."))
-	var/obj/structure/machinery/m56d_post/M = new /obj/structure/machinery/m56d_post(user.loc)
-	transfer_label_component(M)
+	var/obj/structure/machinery/m56d_post/post = new(user.loc)
+	transfer_label_component(post)
 	qdel(src)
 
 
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/carbon/xenomorph/mutators/strains/crusher/charger.dm
```diff
@@ -488,37 +488,78 @@
...
+		return
+
+	CrusherImpact()
+	var/datum/effect_system/spark_spread/sparks = new
+	update_health(charger_ability.momentum * 15)
+	sparks.set_up(1, 1, loc)
+	sparks.start()
+	xeno.visible_message(
+		SPAN_DANGER("[xeno] rams \the [src]!"),
+		SPAN_XENODANGER("You ram \the [src]!")
```
harryob, line 502:
```suggestion
		SPAN_DANGER("[xeno] rams [src]!"),
		SPAN_XENODANGER("You ram [src]!")
```

code/modules/mob/living/carbon/xenomorph/mutators/strains/crusher/charger.dm
```diff
@@ -488,37 +488,78 @@
...
+		HMG.health = health
+		transfer_label_component(HMG)
+		HMG.rounds = rounds //Inherent the amount of ammo we had.
+		HMG.update_icon()
+		qdel(src)
+	else
+		var/obj/item/device/m56d_gun/HMG = new(loc) // note: find a better way than a copy pasted else statement
+		HMG.health = health
+		transfer_label_component(HMG)
+		HMG.rounds = rounds //Inherent the amount of ammo we had.
```
harryob, line 523:
```suggestion
		HMG.rounds = rounds //Inherit the amount of ammo we had.
```
but honestly this amount of comments is so unnecessary (obviously not your fault)

code/modules/cm_marines/smartgun_mount.dm
```diff
@@ -137,13 +137,14 @@
...
-	var/obj/structure/machinery/m56d_post/M = new /obj/structure/machinery/m56d_post(user.loc)
-	M.setDir(user.dir) // Make sure we face the right direction
-	M.gun_rounds = src.rounds //Inherit the amount of ammo we had.
-	M.gun_mounted = TRUE
-	M.anchored = TRUE
-	M.update_icon()
-	transfer_label_component(M)
+	var/obj/structure/machinery/m56d_post/post = new(user.loc)
+	post.setDir(user.dir) // Make sure we face the right direction
+	post.gun_rounds = src.rounds //Inherit the amount of ammo we had.
```
harryob, line 142:
```suggestion
	post.gun_rounds = rounds //Inherit the amount of ammo we had.
```

harryob (changes requested):
minor nits

harryob (approved):

PR #5505, opened 2024-01-20
Queen give_evo changes

InsaneRed:
# About the pull request

changes the required_larva to 1 instead of 3

with our pop we never have 3 burrowed larva, i still think it wont be able to get used even with the limit being one because of how our que system and etc works but atleast this should make it get used more compared to the 3 burrowed limit

<details>

</details>

# Changelog
:cl:
balance: You now only need 1 burrowed larva to trade it for evolution as Queen instead of 3 burrowed larva.
/:cl:


Birdtalon (approved):

Birdtalon:
Balance approving as over 1 week has passed & is a relative small change.

PR #5514, opened 2024-01-21
Gardener QOL changes

InsaneRed:
# About the pull request
Gives bigger fruit eaten/destroyed/picked texts as its currently too small to be seen in chat because  is usually always moving and being spammed.
Makes it so you can put fruits next to eachother, but not ontop of eachother.
Lets the player know you can use resin surge to speed up the growth of the fruits.
eaten fruits now delete faster so you can replace them faster, the old 3 seconds made it so if you ate a fruit and instantly replaced it at max fruits, it'd rather delete another one because the eaten fruit still exited for a whole 3 seconds.

QOL changes good

<details>

</details>

# Changelog
:cl:
qol: Bigger text when a gardener's fruit gets picked/destroyed/consumed
qol: Gardeners can now put their fruits next to eachother.
grammar: Adds new information to the strain buy text
code: The fruits now get deleted 1 second after they get eaten, instead of 3 seconds.
balance: Gardener can now transfer plasma
fix: Properly gives the corresponding message when you eat a fruit from the ground, now "consumed" instead of "destroyed"
/:cl:


Birdtalon (approved):

code/modules/cm_aliens/structures/fruit.dm
```diff
@@ -148,7 +148,7 @@
 	mature = FALSE
```
Drulikar, line 148: Please also set picked true here so the message can be corrected from destroyed to consumed if you eat fruit w/o picking up the fruit.

Drulikar (changes requested):
Everything else looks fine.

code/modules/cm_aliens/structures/fruit.dm
```diff
@@ -213,6 +214,7 @@
 	if(recipient && !QDELETED(recipient))
 		recipient.gain_health(heal_amount)
 		to_chat(recipient, SPAN_XENONOTICE("We recover a bit from our injuries, and begin to regenerate rapidly."))
+		to_chat(bound_xeno, SPAN_XENOHIGHDANGER("One of our picked resin fruits has been consumed."))
```
Drulikar, line 217: This isn't the correct place to put this. This means only `/obj/effect/alien/resin/fruit/greater` will indicate they are consumed.

In other words healing fruit now has the consume message, but no other fruit does. Instead this message ought to be in either `finish_consume` or `delete_fruit` or `Destroy`

My recommendation is to make these changes: 
image

```suggestion
```

code/modules/cm_aliens/structures/fruit.dm
```diff
@@ -146,9 +146,10 @@
 /obj/effect/alien/resin/fruit/proc/finish_consume(mob/living/carbon/xenomorph/recipient)
 	playsound(loc, 'sound/voice/alien_drool1.ogg', 50, 1)
 	mature = FALSE
+	picked = TRUE
 	icon_state = consumed_icon_state
 	update_icon()
```
Drulikar, line 151:
```suggestion
	update_icon()
	if(!QDELETED(bound_xeno))
		to_chat(bound_xeno, SPAN_XENOHIGHDANGER("One of our picked resin fruits has been consumed."))
```

Drulikar (changes requested):

Drulikar (approved):

PR #5517, opened 2024-01-21
increases recovery node healing from 2 per sec to 8 per sec

cuberound:
# About the pull request
you can delete this if #5518 gets approved

increases value to make recovery nodes heal 8 hp per sec from current 2hp, simple

Recovery nodes right now are just noob trap, they heal 10 hp per 5 sec that amounts to 2 hp per sec, preaty far from being worth the space, they need 3x3 open area, time efford and risk to heal near them. the PR increases the healing so that it is at least a bit worth it. More tools for players to use means more fun. Even with the buff it is far from OP as even square of 4 nodes heals 32 per second to single target, while needing loads of space resting in middle of that open space and quite some efford to build, also there is hardcap on how many healing nodes can have.

<details>

</details>

# Changelog

:cl:
balance: increases healing node heal from 10hp every 5 sec to 20 hp every 5 sec
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/special/recovery_node.dm
+++ b/code/modules/cm_aliens/structures/special/recovery_node.dm
@@ -30,4 +30,4 @@
 				SPAN_HELPFUL("You feel a warm aura envelop you."))
 	if(!do_after(picked_candidate, heal_cooldown, INTERRUPT_MOVED, BUSY_ICON_MEDICAL))
 		return
-	picked_candidate.gain_health(heal_amount*picked_candidate.caste.heal_resting)
+	picked_candidate.gain_health(heal_amount)
--- a/code/modules/cm_aliens/structures/special/recovery_node.dm
+++ b/code/modules/cm_aliens/structures/special/recovery_node.dm
@@ -30,4 +30,4 @@
 				SPAN_HELPFUL("You feel a warm aura envelop you."))
 	if(!do_after(picked_candidate, heal_cooldown, INTERRUPT_MOVED, BUSY_ICON_MEDICAL))
 		return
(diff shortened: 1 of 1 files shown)


ihatethisengine:
Thing is you don't get health from a single source usually. While 32 hp per second doesn't seem to be huge on its own, you should also take HP regen from weeds and recov pheros into account.

Drulikar (changes requested):
Lets just double the healing for now. I don't think it needs to go this far.

I prefer this to the pheros because then most of the time the recovery node will be useless as just a source of pheros if theres any other source of recovery (basically always queen leaders or eye).

code/modules/cm_aliens/structures/special/recovery_node.dm
```diff
@@ -30,4 +30,4 @@
 				SPAN_HELPFUL("You feel a warm aura envelop you."))
 	if(!do_after(picked_candidate, heal_cooldown, INTERRUPT_MOVED, BUSY_ICON_MEDICAL))
 		return
-	picked_candidate.gain_health(heal_amount)
+	picked_candidate.gain_health(heal_amount*picked_candidate.caste.heal_resting)
```
Drulikar, line 33: If this affected more than just a few castes maybe. But it would need to cut the heal amount probably back down to 10. For this as is it feels weird that only hellhounds, lurkers, runners, and warriors to be getting 1.4-2.5x the healing boost here. Its not like they don't get that bonus in the first place, they'd just be double dipping here on it more.

Drulikar (changes requested):

Drulikar (approved):
Give this some time and if you still feel it needs more we can push it further. But we probably will need the additional building distance requirement at that point.

PR #5519, opened 2024-01-22
Research Fire Properties rework/rebalance

iloveloopers:
# About the pull request
This PR rebalances fire properties.
Fire properties are now exempt from the penalty that makes them ridiculously expensive to create after level 5.
Oxidizing no longer has -2 duration
Fueling no longer has -2 intensity
Flowing no longer has -1 intensity and -1 duration
All fire chems have been adjusted to remain the same as before
Welding fuel hardcoded snowflake properties removed
OT incinerator tank for normal flamer has had it's caps increased to 40 intensity and 40 duration (same as blue flame)

Research fire properties right now are very expensive for something that is slightly better than normal flame, theres almost no point in making fire chems when you could spend those precious credits on something actually useful (MST NST stims) and is guaranteed to be used by marines unlike a fire chem, which requires the marine to have a flamer in the first place.
You're now able to make a blueflame tier chem with easy complexity for around 25-30 points + clearance 3.
You can now make a maxcap flame (only usable by pyro spec) for around 50 points + clearance 5.

I tested it yeah. Might want to TM first tho
I humbly ask mantainers to please give this a thought and to not instantly close.

# Changelog
:cl:
balance: Oxidizing no longer has -2 duration.
balance: Fueling no longer has -2 intensity.
balance: Flowing no longer has -1 intensity and -1 duration.
balance: Welding fuel hardcoded snowflake properties removed, all other fire chems adjusted to match old stats.
balance: Tank for normal flamer has had it's caps increased to 40 intensity and 40 duration (same as blue flame).
balance: Custom Chemical burncolors will now always be either white or pink.
/:cl:


RenaRenaRe:
Just so you're aware, changing the property levels for chems like welding fuel, thermite and napalm affects OT grenades, rockets, etc. So this is an unintentional(?) OT nerf. Your statement "All fire chems have been adjusted to remain the same as before" is only true for the flamer stats, so you've maintained flamer stats at the cost of nerfing the stats for things like grenades and rockets

iloveloopers:
> Just so you're aware, changing the property levels for chems like welding fuel, thermite and napalm affects OT grenades, rockets, etc. So this is an unintentional(?) OT nerf. Your statement "All fire chems have been adjusted to remain the same as before" is only true for the flamer stats, so you've maintained flamer stats at the cost of nerfing the stats for things like grenades and rockets

added OT stats to the main flame chems used (they override properties).
reverted alcohol changes cause they're weak enough already and don't matter.

code/modules/projectiles/magazines/flamer.dm
```diff
@@ -161,6 +161,8 @@
 	max_rounds = 100
 	max_range = 5
 	fuel_pressure = 1
+	max_duration = 40
+	max_intensity = 40
```
cuberound, line 165:
```suggestion
```
it already inherits this value from /obj/item/ammo_magazine/flamer_tank

iloveloopers:
OT values fixed for thermite and napalm. (they're true to the wiki now)
welding fuel and CLF3 OT values removed
Welding fuel is objectively worse than both thermite and napalm and nobody uses it so it stays as is

Nanu308:
Willing to testmerge this and take a look myself on the balance aspect, but not promising that it's gonne get merged or accepted as it is

iloveloopers:
normal flamer tank max_duration raised to 40, meaning you no longer need to get the OT to print you custom tanks for you to reach the blueflame cap.
if Nanu considers this to remove too much department cooperation, I am willing to revert (though the only thing that the OT does is press a button on the armylathe and an MT can do it aswell).

Segrain:
I do believe that there has been quite public announcement: changes to research are not currently accepted. I do not understand what makes you think that this is somehow an exception.

even that procedural point aside, this PR has been on testmerge for a few days by now, and after observing multiple times how it ends being used I can only say "no, please no, _absolutely_ not", for reasons which are mostly not code review and do not quite belong here. Just in case management would disagree and want to scream at me about it, I am not closing the PR outright with extreme prejudice, but if you will approve and merge this without telling me, I will scream at you too.

iloveloopers:
removed the exemption of fire properties from the penalty that is added above level 5 on create mod, price is now slightly increased.
custom flames burncolor will now always be either white or pink (50/50 chance) to avoid confusion by xenos.

Segrain makes some fair points, though I believe the ultimate decision will be made by management

PR #5536, opened 2024-01-24
Fixes xeno tunnel tracking

### LINKED ISSUE #5529: Tunnel tracker does not track the tunnel you click on
### Testmerges

#5086: 'Adds the XM51 Breaching Scattergun' by VileBeggar at commit 1a300eff40 #5007: 'Project ARES TM Holder (v4)' by realforest2001 at commit 9ef47f593f #5493: 'React Migration - part 2' by mullenpaul at commit 707f3585f6 #5392: 'Yautja Weapon Gimmicks, Round Two' by realforest2001 at commit bcd8b2b63d #5490: 'Requisition Centralization [Part 1]' by ItsVyzo at commit f75675449d #5525: 'Fix Weird Inventories' by Drulikar at commit 4787757a5f #5515: 'Nerfs GAU damage by one fifth.' by Birdtalon at commit 3586ae2981

### Round ID

20529 and 20530

### Description of the bug

If you alt click the queen locator you can switch it to track a tunnel. However if you try to track a tunnel it will not necessarily track the tunnel you clicked on and will instead track a different tunnel.  It seems like it's due to https://github.com/cmss13-devs/cmss13/pull/5465, I think if you try to track the furthest tunnel away from you (the last in the list) it will instead track the most recently created tunnel (what _used_ to be the last in the list before that PR).

Tracking a tunnel should actually track that tunnel.

1. Be a xeno
2. Track a tunnel
3. Notice that it does not actually point to the correct tunnel

### Issue Bingo

SabreML:
# About the pull request

I swear I do actually test my PRs 😢.

Fixes #5529 by refactoring the 'Queen locator' to track targets with `/datum/weakref`s, rather than their index in a global list.

Bugfix. (Honestly no idea how I missed this one)

It makes the code much simpler and easier to understand (arguably).

<details>

https://github.com/cmss13-devs/cmss13/assets/57483089/97552c21-c28a-49a8-a4aa-6ff03209062e

</details>

# Changelog
:cl:
fix: Fixed the xeno 'queen locator' giving the wrong location when a tunnel was selected.
/:cl:

### DIFF
--- a/code/_onclick/hud/screen_objects.dm
+++ b/code/_onclick/hud/screen_objects.dm
@@ -505,7 +505,11 @@
 	name = "queen locator"
 	icon = 'icons/mob/hud/alien_standard.dmi'
 	icon_state = "trackoff"
-	var/list/track_state = list(TRACKER_QUEEN, 0)
+	/// A weak reference to the atom currently being tracked.
+	/// (Note: This is null for `TRACKER_QUEEN` and `TRACKER_HIVE`, as those are accessed through the user's hive datum.)
+	var/datum/weakref/tracking_ref = null
+	/// The 'category' of the atom currently being tracked. (Defaults to `TRACKER_QUEEN`)
+	var/tracker_type = TRACKER_QUEEN
 
 /atom/movable/screen/queen_locator/clicked(mob/living/carbon/xenomorph/user, mods)
 	if(!istype(user))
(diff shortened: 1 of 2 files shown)


SabreML:
I haven't quite finished testing this one yet since I had to prepare food (I melted a spatula!), so it's drafted for a little while.

code/_onclick/hud/screen_objects.dm
```diff
@@ -520,35 +524,39 @@
...
+		for(var/mob/living/carbon/xenomorph/leader in user.hive.xeno_leader_list)
+			options["Xeno Leader [leader]"] = list(WEAKREF(leader), TRACKER_LEADER)
 
 		var/list/sorted_tunnels = sort_list_dist(user.hive.tunnels, get_turf(user))
-		var/tunnel_index = 1
-		for(var/obj/structure/tunnel/tunnel in sorted_tunnels)
-			options["Tunnel [tunnel.tunnel_desc]"] = list(TRACKER_TUNNEL, tunnel_index)
-			tunnel_index++
+		for(var/obj/structure/tunnel/tunnel as anything in sorted_tunnels)
+			options["Tunnel [tunnel.tunnel_desc]"] = list(WEAKREF(tunnel), TRACKER_TUNNEL)
```
Drulikar, line 538: Lets only convert them to weakrefs when you've actually picked one.

It can be a lot of extra work that we don't really need, and the reference should only last as long as the input list is open (feel free to add like a 30 SECONDS or whatever timeout to the tgui_input_list call if you want).

If you're concerned about making a weakref to something that got deleted between the time the person was making a selection, you can just ignore that (it'll make a weakref that on first resolve then your weakref will return null on for you to untrack) or you can add an extra if QDELETED check when we're making a selection.

Drulikar (changes requested):

code/_onclick/hud/screen_objects.dm
SabreML: Yeah that would work better, thanks for the suggestion!

I don't think it'll be necessary to add any extra checks to it, since like you said if the atom gets deleted it'll try to resolve, fail, then revert back to tracking the queen.

Drulikar (approved):
Code looks fine. Not currently heard or seen any issues with it TM'd either.

PR #5545, opened 2024-01-24
scouts can no longer attach 2x-4x scopes on their gun

InsaneRed:
# About the pull request
makes it so you cant attach the 2x-4x scopes on the gun

its not fun being stunned from 2/3 screens away, especially with burstfire it becomes neigh uncounterable. scout should be semi cqc check below for the video

<details>

https://github.com/cmss13-devs/cmss13/assets/47158596/f025e43b-8d50-4e49-ac46-1d5505bd4dfc

</details>

# Changelog
:cl:
balance: Scouts can no longer put 2x-4x scopes on their gun
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/specialist/scout.dm
+++ b/code/modules/projectiles/guns/specialist/scout.dm
@@ -42,8 +42,6 @@
 		/obj/item/attachable/verticalgrip,
 		/obj/item/attachable/angledgrip,
 		/obj/item/attachable/lasersight,
-		/obj/item/attachable/scope,
-		/obj/item/attachable/scope/mini,
 		/obj/item/attachable/flashlight/grip,
 	)
 


HeresKozmos:
cool change

cuberound:
better change to prevent this without limiting scout gameplay, make the stun amo stun only when in certain range, like the heavy revovler bullets, easy

Nanu308:
I agree with offscreen combat sucking at times, especially with no indication of it/warning, but I don't think this is a good way to solve backliners being stunlocked from range, i believe another suggestion which was mentioned might be better and that is limiting the range of when it can stun, or even make it have a "stun timer" falloff depending on how far away it gets fired from.

I think there's a higher chance of something like i've mentioned above leading to this PR being accepted.

Nanu308 (changes requested):
see comment

PR #5548, opened 2024-01-25
XM4 Armor Slowdown Change

MobiusWon:
# About the pull request
Changes the XM4 armor slowdown to `SLOWDOWN_ARMOR_LIGHT (0.35)` instead of the current `SLOWDOWN_ARMOR_MEDIUM (0.55)`.

As it stands the XM4 armor is a "noobtrap", with majority of experienced Intelligence Officers (IOs) opting to use either the Service Jacket or Light Armor from the squad prep vendors. Changing the XM4 slowdown to match the slowdown of the light armor makes it a viable option, while not making it overpowered.

IOs rely on speed and awareness to survive in most cases. This will also match the general "fluff" of IOs having advanced gear a slight cut above normal Marines. This should not have any major effects on balance vs. Xenos, as IOs already have the capability to use Light Armor for the same speed. Regardless, most IOs are going to wind up dead in a colony corner anyways if they get jumped, so the differences in the armor values are likely negligible. This merge would really just cut down on the prep time for IOs.

N/A

# Changelog
:cl:
qol: XM4 slowdown = SLOWDOWN_ARMOR_LIGHT (0.35) changing from the current 0.55

:cl:

### DIFF
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -282,6 +282,7 @@
 	name = "\improper XM4 pattern intelligence officer armor"
 	uniform_restricted = list(/obj/item/clothing/under/marine/officer, /obj/item/clothing/under/rank/qm_suit, /obj/item/clothing/under/marine/officer/intel)
 	specialty = "XM4 pattern intel"
+	slowdown = SLOWDOWN_ARMOR_LIGHT
 
 /obj/item/clothing/suit/storage/marine/MP
 	name = "\improper M2 pattern MP armor"


Nanu308:
So, the IO Armor compared to the Light armor has;
 Four Storage Slots of space, 
 Better armor - tiny bit better than medium even, 
 Better light range & power,

The argument "This merge would really just cut down on the prep time for IOs." is a bad one and you could solve it then by just adding the light armor/different armors to the IO vendor itself instead of doing a buff.

The other argument of "IOs already have the capability to use Light Armor for the same speed" is a weird one to me, because if they swap armor they are giving up the benefits of the IO armor.

And the argument about it not affecting balance much depends on ones view, skirmishers would be affected, the time potentially for marines to gain intel would be affected, and more.

I'm not convinced of the arguments provided here, because this just seems like a powercreep buff.

MobiusWon:
@Nanu308 What you have referenced are my supporting arguments not my primary argument. My primary argument is that any IOs with reasonable experience do not take the XM4 armor and highly recommend new IO players against taking it, solely because of the slowdown, rendering it effectively useless. If no one uses the asset then why is it in game to at all? This patch would correct that issue by eliminating the main reason no one uses it. I believe it is a stretch to consider this power creep when no one uses it. This is _atleast_ worth a test merge. As far as skirmishers are concerned we’re talking about a negligible difference of maybe 2–3 hits to crit an IO. I know because I have worn both armors and tested this in a live game.

Steelpoint:
The actual armour of the XM4 is of dubious benefit for an IO. The main advantage of armour is the effect in minimising the effect of friendly fire or explosives, for IO's you generally prefer speed over anything else. So the trade-off for more experienced IO players is you either take the jacket for no armour but high speed, or light armour for decent speed and a bit of armour.  The issue therein is that xm4 is not desirable as a IO prizes speed over armour. 

This issue has been brought up before and the prior stance was that XM4 was "fine" as is even though its in a state where generally only new players take it, and veteran players take Light armour or no armour.

PR #5571, opened 2024-01-27
increases exxecution speed of CAS FM, number is up for a debate

cuberound:
# About the pull request

reduces time betwean FM execution steps, increses it speed from 3 tiles per second to 5 tiles per second

after I added all the extra warnings it is near imposible to hit xeno with it, especily considering that even when the xeno is in the area where the CAS is being fired, the FM moves so slowly that even the slowest xenos can outrun it easily , this aims to make CAS more powerful (and also cool looking), by making it move faster so that when you end up in the area where the CAS is being fired at even after all the warnings, you can not just walk away from it without getting hit, still have chance when you are close to edge of the FM

<details>

</details>

# Changelog
:cl:
balance: increseas FM movement speed to 5 tiles per second from 3 tiles per second
/:cl:

### DIFF
--- a/code/game/cas_manager/datums/cas_fire_mission.dm
+++ b/code/game/cas_manager/datums/cas_fire_mission.dm
@@ -160,7 +160,7 @@
 		return "Weapon [weapon_string] has not enough ammunition to complete this Fire Mission"
 	return "Unknown Error"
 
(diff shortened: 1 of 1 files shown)


Drulikar:
I don't think this solves the issue caused by the drastically increased amount of warning. You can book it even if you don't know where it is. I was hopeful sometimes people would be warned when they aren't close to it and then run closer to it sometimes but we also made the 2nd warning earlier too.

But tinker with things by testing locally with two clients.

mullenpaul:
Should this instead be a var on the weapons console that could be edited via VV? This would allow for more detailed experimentation rather than lots of back and forth testing. It would allow us to run an 'event' round where we pin the execution speed to a new number and see what we think. This would allow for general feedback from both xenos and marines

PR #5583, opened 2024-01-28
Removes Squad Optics as Default Helmet Visors

Staykeu:
# About the pull request

Removes the useless "squad optic" as a default attachment for marine helmets.

The squad optic is functionally useless and only serves as a cosmetic item, most players don't use the squad optic and most of the time it only serves as another optic you have to flick through to get to the optic you want. Those who still want to use the squad optic can get them from squad prep rooms, it's not fully removed.

<details>

</details>

# Changelog
:cl: Stakeyng
qol: Removes the need to cycle through the squad optic to get to other optics (welding/medical/nv).
/:cl:

### DIFF
--- a/code/game/machinery/vending/vendor_types/requisitions.dm
+++ b/code/game/machinery/vending/vendor_types/requisitions.dm
@@ -385,6 +385,7 @@
 		list("M10 Pattern Marine Helmet", 20, /obj/item/clothing/head/helmet/marine, VENDOR_ITEM_REGULAR),
 		list("M10 Pattern Technician Helmet", 20, /obj/item/clothing/head/helmet/marine/tech, VENDOR_ITEM_REGULAR),
 		list("M10 Pattern Corpman Helmet", 20, /obj/item/clothing/head/helmet/marine/medic, VENDOR_ITEM_REGULAR),
+		list("Squad Optic", 10, /obj/item/device/helmet_visor, VENDOR_ITEM_REGULAR),
 		list("M3 Pattern Carrier Marine Armor", 20, /obj/item/clothing/suit/storage/marine/medium/carrier, VENDOR_ITEM_REGULAR),
 		list("M3 Pattern Padded Marine Armor", 20, /obj/item/clothing/suit/storage/marine/medium/padded, VENDOR_ITEM_REGULAR),
 		list("M3 Pattern Padless Marine Armor", 20, /obj/item/clothing/suit/storage/marine/medium/padless, VENDOR_ITEM_REGULAR),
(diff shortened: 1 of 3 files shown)


mullenpaul (approved):

Drulikar (changes requested):
The visor does work - however headsets currently also provide this HUD for USCM so it is currently redundant assuming you have a headset.

I am half tempted to just remove the HUD from headsets however for now in this PR please instead simply address the reason why you were making this PR in the first place: Revert your current changes and instead **if you already have a HUD, have the visor toggle skip that visor.**

This means it should also skip a medhud visor if you have an eye item giving the medical hud. Perform the check by looking at the the data in mob_hud.dm.

Wintermote:
Does this also change the current functionality so squad optics can be removed with a screwdriver after install?

Staykeu:
> The visor does work - however headsets currently also provide this HUD for USCM so it is currently redundant assuming you have a headset.
> 
> I am half tempted to just remove the HUD from headsets however for now in this PR please instead simply address the reason why you were making this PR in the first place: Revert your current changes and instead **if you already have a HUD, have the visor toggle skip that visor.**
> 
> This means it should also skip a medhud visor if you have an eye item giving the medical hud. Perform the check by looking at the the data in mob_hud.dm.

I would argue this is unnecessary, all of it. The squad HUD is still available in squad prep's vending, and while the headset is in its current state (where it should stay), there is no reason to have a squad HUD.

Staykeu:
> Does this also change the current functionality so squad optics can be removed with a screwdriver after install?

Currently, the squad optic is built into the helmet, built-in visors cannot be removed. Attached visors CAN be removed, so if you attach a squad visor, it can be removed.

harryob:
> I would argue this is unnecessary, all of it.

not how requested changes work

Staykeu:
> > I would argue this is unnecessary, all of it.
> 
> not how requested changes work

so you just fucking close it? awesome, amazing

Drulikar:
Feel free to open another PR with the requested change. It both keeps the functional visor in the helmet, and resolves your issue needing to perform an extra skip over a hud you already have from your headset, AND would do the same for medhuds too should someone also be wearing one (e.g. prescription maybe).

Staykeu:
> Feel free to open another PR with the requested change. It both keeps the functional visor in the helmet, and resolves your issue needing to perform an extra skip over a hud you already have from your headset, AND would do the same for medhuds too should someone also be wearing one (e.g. prescription maybe).

https://github.com/cmss13-devs/cmss13/pull/5600 harry helped me out with it, this is the new pr with the changes you requested

PR #5586, opened 2024-01-29
XM4 Armor Integral Motion Detector

MobiusWon:
# About the pull request

Adds a motion detector action to the XM4 Intelligence Officer (IO) Armor.

1. Currently there is no difference between the standard M4 "RTO" armor and the XM4 IO armor. The integral motion detector makes the X for "Experimental" meaningful. 
2. The XM4 currently goes _mostly_ unused by IOs, this would change that.
3. The Motion Detector is already an essential item for IOs, this change distinguishes it from the other armor choices in game and adds a uniqueness to their piece of kit.
4. I envision the XM4 as the "safety" choice for IOs: it has extra storage, slower, and (with this feature) includes a Motion Detector to improve situational awareness. The other common choices for IOs are the Service Jacket I.E. the "speed" choice, and the M3-L armor a "middle ground" option.
5. This should not have any meaningful impact on balance vs. skirmishers as the Motion Detector is already an essential item for IOs.
6. Keep in mind the XM4 is only attainable by IOs and ~cannot be worn or purchased by anyone else.~

NOTE: This is my first "major" feature. I am open to suggestions, including "balance" related stuff.

https://github.com/cmss13-devs/cmss13/assets/32654903/445352b4-bfec-460f-95d8-2eadea27b953

Edit: Tested to confirm that the MD turns off upon user death. ✔️
# Changelog
:cl:
NervanCatos
add: Adds integral Motion Detector to XM4 IO Armor
balance: Uniform utility accessories (I.E. Webbing) no longer can be equipped while the XM4 is worn. Armor slots increased to 5 to compensate.
/:cl:


code/modules/clothing/suits/marine_armor.dm
```diff
@@ -282,6 +282,77 @@
 	name = "\improper XM4 pattern intelligence officer armor"
 	uniform_restricted = list(/obj/item/clothing/under/marine/officer, /obj/item/clothing/under/rank/qm_suit, /obj/item/clothing/under/marine/officer/intel)
 	specialty = "XM4 pattern intel"
+	// XM4 Integral Motion Detector
+	actions_types = list(/datum/action/item_action/toggle, /datum/action/item_action/intel/toggle_motion_detector)
+	var/motion_detector = 0
+	var/obj/item/device/motiondetector/xm4/MD
```
Birdtalon, line 288: Try to not use single letter vars or abbreviations.

code/modules/clothing/suits/marine_armor.dm
```diff
@@ -282,6 +282,77 @@
 	name = "\improper XM4 pattern intelligence officer armor"
 	uniform_restricted = list(/obj/item/clothing/under/marine/officer, /obj/item/clothing/under/rank/qm_suit, /obj/item/clothing/under/marine/officer/intel)
 	specialty = "XM4 pattern intel"
+	// XM4 Integral Motion Detector
+	actions_types = list(/datum/action/item_action/toggle, /datum/action/item_action/intel/toggle_motion_detector)
+	var/motion_detector = 0
```
Birdtalon, line 287: Code doc this as one can't tell what it is at a glace. Also if its supposed to be boolean then use `FALSE`
```
/// This var does something
var/my_var = 0
```

Birdtalon (dismissed):

code/modules/clothing/suits/marine_armor.dm
```diff
@@ -282,6 +282,77 @@
...
+	var/obj/item/device/motiondetector/xm4/MD
+	var/long_range_cooldown = 2
+	var/recycletime = 120
+
+/obj/item/clothing/suit/storage/marine/medium/rto/intel/Initialize(mapload, ...)
+	MD = new(src)
+	. = ..()
```
Birdtalon, line 294:
```suggestion
	. = ..()
	MD = new(src)
```

Should call parent first in Initialization

code/game/objects/items/devices/motion_detector.dm
```diff
@@ -191,6 +191,13 @@
 	if(ishuman(A.loc))
 		return A.loc
 
+/obj/item/device/motiondetector/xm4
+
+/obj/item/device/motiondetector/xm4/get_user()
+	var/atom/A = loc
+	if(ishuman(A.loc))
+		return A.loc
```
Birdtalon, line 199: Doesn't this just do the same as the proc  on L183-185 we're casting to an /atom but not operating on it so I don't see why we need this override?

code/modules/clothing/suits/marine_armor.dm
```diff
@@ -282,6 +282,77 @@
...
+			MD.refresh_blip_pool()
+
+		long_range_cooldown--
+		if(long_range_cooldown)
+			return
+		long_range_cooldown = initial(long_range_cooldown)
+		MD.scan()
+
+/obj/item/clothing/suit/storage/marine/medium/rto/intel/proc/toggle_motion_detector(mob/user)
+	to_chat(user, "[icon2html(src, usr)] You [motion_detector? "<B>disable</b>" : "<B>enable</b>"] \the [src]'s motion detector.")
```
Birdtalon, line 341: This should probably use a `SPAN()` for formatting.

code/modules/mob/living/carbon/human/human_helpers.dm
```diff
@@ -248,6 +248,22 @@
...
+			var/obj/item/device/motiondetector/md = i
+			md.toggle_active(src, old_active = TRUE, forced = TRUE)
+		if(istype(i, /obj/item/clothing/suit/storage/marine/medium/rto/intel))
+			var/obj/item/clothing/suit/storage/marine/medium/rto/intel/xm4 = i
+			if(xm4.motion_detector)
+				xm4.motion_detector = FALSE
+				var/datum/action/item_action/intel/toggle_motion_detector/TMD = locate(/datum/action/item_action/intel/toggle_motion_detector) in xm4.actions
+				TMD.update_icon()
+				xm4.motion_detector()
+
```
Birdtalon, line 266: - Why the prox sensor check again?
- You can just include this behahiour in the for loop on L235

MobiusWon:
@Birdtalon To answer any “why” you have, this code is borrowed from the integral md that the smartgun has. That being said, I can make the changes you mentioned!

code/modules/clothing/suits/marine_armor.dm
MobiusWon: How do I make a code doc? Or do you mean just add more comments?

kiVts: notice how review above has /// instead of //, this is code doc. makes things show up when hovering over it in vsc iirc

MobiusWon: Oh, I see what you mean. That’s neat, thanks!

MobiusWon:
Update: Removed some unnecessary duplicates of code, added some documentation, added suggestions from @Birdtalon, and added a unique description for the XM4 armor.

MobiusWon:
> “Later variants of USCM standard armor included integrated motion trackers, thereby leaving a Marine's hands free to ready their weapon.[[7]] The armor-integrated motion tracker appeared in the corner of a Marine's head-up display and its appearance was very similar to what was displayed on the screen of the handheld tracker. The HUD motion tracker functioned the same as the handheld M314, with a moving object shown on the tracker's display as a series of contours of probable loci and with the range and bearing given to the target. The HUD motion tracker would also display the direction of an objective that the user had active.”

Neat lore that supports this feature. The XM4 could be a prototype version of the integrated motion tracker, that after successful field testing led to it being standard issue.

Staykeu:
i really dont want this PR to die, but it looks like everyone in the maintainer team is intentionally ignoring it or outright disapproving of it

if this PR goes under i'd recommend trying again but instead of an integral motion detector the XM4 comes with an integral data detector, i feel like that has a higher chance to appeal (or at least just not tick off potential biases within the maint team)

MobiusWon:
> Code seems fine but I do not approve of the PR for design & balance reasons.

What needs to change for it to be balanced? First thing that comes to mind is reducing storage.

code/modules/clothing/suits/marine_armor.dm
```diff
@@ -282,6 +282,106 @@
...
+	var/mob/living/carbon/human/human = user
+	var/obj/item/clothing/under/uniform = human.w_uniform
+	if(slot == WEAR_JACKET)
+		uniform?.valid_accessory_slots -= ACCESSORY_SLOT_UTILITY
+
+/obj/item/clothing/suit/storage/marine/medium/rto/intel/unequipped(mob/user, slot) //When unequipped this adds the ACCESSORY_SLOT_UTILITY back as a valid accessory
+	. = ..()
+	var/mob/living/carbon/human/human = user
+	var/obj/item/clothing/under/uniform = human.w_uniform
+	uniform?.valid_accessory_slots += ACCESSORY_SLOT_UTILITY
```
Drulikar, line 384: This should only be done if it originally had that slot; and only if the slot is WEAR_JACKET. See the patch I gave you previously.

code/modules/clothing/suits/marine_armor.dm
```diff
@@ -282,6 +282,106 @@
...
+	if(accessory)
+		to_chat(user, SPAN_WARNING("[src] can't be worn with [accessory]."))
+		return FALSE
+
+	return TRUE
+
+/obj/item/clothing/suit/storage/marine/medium/rto/intel/equipped(mob/user, slot, silent) //When XM4 is equipped this removes ACCESSORY_SLOT_UTILITY as a valid accessory for the uniform
+	. = ..()
+	var/mob/living/carbon/human/human = user
+	var/obj/item/clothing/under/uniform = human.w_uniform
```
Drulikar, line 376: Minor, but this should be after the slot check like in the patch I gave you previously.

code/modules/clothing/suits/marine_armor.dm
```diff
@@ -282,6 +282,106 @@
...
+		playsound(loc,'sound/items/detector_turn_off.ogg', 25, 1)
+	motion_detector = !motion_detector
+	var/datum/action/item_action/intel/toggle_motion_detector/TMD = locate(/datum/action/item_action/intel/toggle_motion_detector) in actions
+	TMD.update_icon()
+	motion_detector()
+
+/obj/item/clothing/suit/storage/marine/medium/rto/intel/proc/motion_detector()
+	if(motion_detector)
+		START_PROCESSING(SSobj, src)
+	if(!motion_detector)
```
Drulikar, line 357:
```suggestion
	else
```

Drulikar (changes requested):

code/modules/clothing/suits/marine_armor.dm
MobiusWon: So the issue here is that in the off chance in an edge case it would add the slot to a uniform that didn't originally allow it?

code/game/objects/items/devices/motion_detector.dm
```diff
@@ -191,6 +191,14 @@
 	if(ishuman(A.loc))
 		return A.loc
 
+/obj/item/device/motiondetector/xm4
+
+///Forces the blue blip to appear around the detected mob
+/obj/item/device/motiondetector/xm4/get_user()
+	var/atom/A = loc
+	if(ishuman(A.loc))
+		return A.loc
```
Drulikar, line 200:
```suggestion
	var/atom/holder = loc
	if(ishuman(holder.loc))
		return holder.loc
```

code/modules/clothing/suits/marine_armor.dm
```diff
@@ -283,6 +283,117 @@
...
+	name = "Toggle Motion Detector"
+	action_icon_state = "motion_detector"
+	button.name = name
+	button.overlays.Cut()
+	button.overlays += image('icons/mob/hud/actions.dmi', button, action_icon_state)
+
+/datum/action/item_action/intel/toggle_motion_detector/action_activate()
+	. = ..()
+	var/obj/item/clothing/suit/storage/marine/medium/rto/intel/recon = holder_item
+	recon.toggle_motion_detector(usr)
```
Drulikar, line 319:
```suggestion
	recon.toggle_motion_detector(owner)
```
No need to use `usr` when a `/datum/action` has its owner mob set. Generally should avoid usr.

Drulikar (changes requested):
Couple minor things, but otherwise seems to be working.

Also I will note that:
>  cannot be worn ... by anyone else

Isn't true; there's nothing preventing anyone else from wearing this armor. But I don't know if that needs to change.

MobiusWon:
> Couple minor things, but otherwise seems to be working.
> 
> Also I will note that:
> 
> > cannot be worn ... by anyone else
> 
> Isn't true; there's nothing preventing anyone else from wearing this armor. But I don't know if that needs to change.

Want me to change that?

Staykeu:
> > Couple minor things, but otherwise seems to be working.
> > Also I will note that:
> > > cannot be worn ... by anyone else
> > 
> > 
> > Isn't true; there's nothing preventing anyone else from wearing this armor. But I don't know if that needs to change.
> 
> Want me to change that?

probably could make the armor ID lockable like spec armor, therefore, when an IO permas their armor unlocks and can be repurposed

Drulikar (approved):

PR #5595, opened 2024-01-30
Fix greenfire not mixing well with lighting

### LINKED ISSUE #5594: Green fire appearance is incorrect
### Testmerges

N/A

### Round ID

N/A Tested locally

### Description of the bug

Green fire for some reason is not appearing as it should be in game. I currently do not see any difference other than the obvious of different duration and color values.

~~I've briefly investigated this and its not obvious to me the problem.~~ If the fire is near a different light, it will appear correctly. It is its own lights blowing out the color.

It ought to look like its sprite:
image
image

But in reality only green fire is the only one that has this weird appearance:

image

1. Start test game
2. Spawn as a pyro spec
3. Fire green fire
4. Compare green fire to any other fire

### Issue Bingo

Drulikar:
# About the pull request

The current green fire sprites doesn't behave well with green lights - it just blows out all definition in the sprite. There were some unused greenfire sprites in the file so I swapped to those. However there were only 3 of them, so I copied the current white fire and applied basically the same palette to it to make it green. It has the same 1 2 1 2 delay.

Fixes #5594

https://github.com/cmss13-devs/cmss13/assets/76988376/f9cc780b-5994-41b5-a7a1-374e59e52e34

# Changelog
:cl: Drathek
imageadd: Changed green fire sprites to work better with lights
/:cl:


harryob (approved):
recolor

PR #5600, opened 2024-01-31
Helmet HUDs skip over HUDs that are already in effect on the player

Staykeu:
# About the pull request

Makes it so that if you have an active HUD, a helmet visor that does the same thing will not be considered when switching through visors in your helmet. Credit to harryob for his help with the code.

Visors are usually a pain to switch through when their functionality isn't needed, take for example, the squad optic, which does the same job as the headset (HUD-wise). This change makes it so you don't need to switch through unneeded HUDs.

<details>

</details>

# Changelog
:cl: Stakeyng
qol: Removes the need to cycle through helmet visors that contain HUD effects that are already active on your character.
/:cl:

### DIFF
--- a/code/modules/clothing/head/helmet.dm
+++ b/code/modules/clothing/head/helmet.dm
@@ -670,8 +670,15 @@ GLOBAL_LIST_INIT(allowed_helmet_items, list(
 		for(var/hud_type in total_visors)
 			if(hud_type == active_visor)
 				if(length(total_visors) > iterator)
-					active_visor = total_visors[(iterator + 1)]
-					toggle_visor(user, total_visors[iterator], TRUE)
+					var/obj/item/device/helmet_visor/current_visor = total_visors[iterator]
+					var/obj/item/device/helmet_visor/next_visor = total_visors[iterator + 1]
+
+					if(!isnull(GLOB.huds[next_visor.hud_type].hudusers[user]))
+						iterator++
+						continue
+
+					active_visor = next_visor
+					toggle_visor(user, current_visor, silent = TRUE) // disables the old visor
 					toggle_visor(user)
 					return active_visor
 				else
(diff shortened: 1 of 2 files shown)


code/modules/clothing/head/helmet.dm
```diff
@@ -680,15 +687,15 @@ GLOBAL_LIST_INIT(allowed_helmet_items, list(
...
-	if(total_visors[1])
-		active_visor = total_visors[1]
+	for(var/obj/item/device/helmet_visor/new_visor in total_visors)
+
+		if(!isnull(GLOB.huds[new_visor.hud_type].hudusers[user]))
+			continue
+
+		active_visor = new_visor
 		toggle_visor(user)
 		return active_visor
```
harryob, line 698:
```suggestion

		return FALSE
```

harryob (changes requested):
for sanity

code/modules/clothing/head/helmet.dm
```diff
@@ -680,15 +687,16 @@ GLOBAL_LIST_INIT(allowed_helmet_items, list(
...
 
-	active_visor = null
-	recalculate_visors(user)
-	return FALSE
-
+		return FALSE
```
harryob, line 699:
```suggestion
	return FALSE
```

harryob (changes requested):

harryob (approved):
im fine with this but given i my hand in creating the pr we should get someone else to look it over

Drulikar (changes requested):
This is close, but it doesn't fully handle the situation where the next visor is not valid - it skips disabling the current visor. But this isn't your fault; this is actually fairly complex and probably needs to be refactored because its the logic in toggle_visor that's complicating everything. But I believe my suggestions below address the issue I discovered.

For example, equip a helmet w/ medical visor and squad visor. Remove your headset. Toggle visor to squad. Equip your medical glasses. Try to toggle the visor: Icon will change to no visor, but it fails to remove the squad visor and now thinks it cannot toggle to any visors.

However, this requires modification of lines outside of suggestable scope so best I can streamline this is to provide a patch: 
[patch.diff.txt](https://github.com/cmss13-devs/cmss13/files/14130498/patch.diff.txt)

To apply it copy this file to the root of your fork, then open terminal in VSC and enter `git apply patch.diff.txt`

The changes should result in this: 
image

Feel free to ask any questions why I did what I did and to test it.

Drulikar (approved):

PR #5607, opened 2024-01-31
fixes medal saving

harryob:
posthumous medals would never save because the ckey is nulled immediately on mob changes. this also applies to medals given when you're disconnected for whatever reason

:cl:
fix: medals should save more reliably
/:cl:


Drulikar (approved):
I need to do the Fira request and make this all use weakrefs instead at some point.

Issue #5229, opened 2023-12-16
Metal Barricades Inconsistency


nuklearcellphoneg:
### Testmerges

NA

### Round ID

NA

### Description of the bug

Basically- you can repair/dismantle walls foldable barricades plasteel barricades and reinforced walls one after the other & simultaneously. 

Metal barricades are the exception to the rule as repairing one prevents starting new repairs. I don't think it'd be that huge of a buff considering gas & acid still forces engies to move away. Furthermore you can repair for example about 9 plasteel barricades and one metal barricade but if you repair the metal barricade first you can only repair one metal barricade. Bad design!

### What's the difference with what should have happened?

Should be able to repair multiple metal barricades.

### How do we reproduce this bug?

1.Repair multiple plasteel cades or walls
2.Enjoy
3.Repair multiple metal cades
4.Don't enjoy

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Weuyn:
this was added intentionally and plasteel cades were overlooked

ihatethisengine:
Definitely not a bug. On the contrary, it's more likely that the ability to repair several folding cades (not only plasteel) is a bug. Not that I want anyone to fix this, but buffing engineers by allowing them to repair multiple metal cades at once seems unreasonable either.

nuklearcellphoneg:
I'm quite unsure- again it's the sole exception to the rule- for example you can dismantle up to four walls simultaneously from a corner. There's also the issue how whoever did it coded it- you can repair 9 plast + 1 metal cade simultaneously but the second you repair the metal cade it prevents repairing new foldable cades. That's just bad implementation imo.

As it stands if metal cades weren't made consistent then technically the optimal gameplay plan is to have your last line of defense to be 100% plasteel cades with its edges reinforced walls/walls- outside cades should all be made of pure metal. Then again I assume that if you made a line of pure foldable cades you'd be able to repair 100% of them simultaneously as well.

ihatethisengine:
Dismantling walls is irrelevant for balance, but repairing cades is crucial for FOB siege gameplay. You are not going to realistically have 9 plasteel cades because of the cost so it's also not a big deal, but I agree it was probably an oversight rather than deliberate decision.

Huffie56:
hopefully this will fix it or part of it...
[...]

PR #5609, opened 2024-01-31
refactor barricades files to hopfully have a more constant interaction when trying to fix them with welder.

Huffie56:
# About the pull request
fixes: https://github.com/cmss13-devs/cmss13/issues/5229
1-replace a bunch of single letter variable and in the same movement standardize it for all the parent/child object...
2-create a proc that is a group of check to see if you can start fixing a barricade with a welder.
3-implement this proc on metal,plasteel and deployable to have all the check be the same as we want them to all react the same way...

i tested welding different different cades at the same time seem to work i had some weirdness but it's working.
<details>

</details>

# Changelog
:cl:
refactor: refactored files with the aim of making fixing barricade with a welder more constant.
fix: being unable to repair more than one metal barricade at the time.
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade/barricade.dm
+++ b/code/game/objects/structures/barricade/barricade.dm
@@ -487,10 +487,10 @@
 	if(health == maxhealth)
 		to_chat(user, SPAN_WARNING("[src] doesn't need repairs."))
 		return FALSE
-
+/* remove this check because we want people being able to weld multiple cades at the same time?
 	if(user.action_busy)
 		to_chat(user, SPAN_WARNING("You are currently doing something else"))
-		return FALSE
+		return FALSE*/
 
 	if(damage_state >= welder_fix && damage_state != null && welder_fix != null)
 		to_chat(user, SPAN_WARNING("[src] has sustained too much structural damage to be repaired."))
--- a/code/game/objects/structures/barricade/barricade.dm
+++ b/code/game/objects/structures/barricade/barricade.dm
@@ -470,7 +470,9 @@
 // testing making a prop
 
 /obj/structure/barricade/proc/attackby_welder(obj/item/item, mob/user)
-
+	if(!HAS_TRAIT(item, TRAIT_TOOL_BLOWTORCH))
+		to_chat(user, SPAN_WARNING("You need a stronger blowtorch!"))
+		return
 
 	if(health == maxhealth)
 		to_chat(user, SPAN_WARNING("[src] doesn't need repairs."))
--- a/code/game/objects/structures/barricade/barricade.dm
+++ b/code/game/objects/structures/barricade/barricade.dm
@@ -482,16 +482,16 @@
 /obj/structure/barricade/proc/attackby_welder(obj/item/item, mob/user)
 	if(!HAS_TRAIT(item, TRAIT_TOOL_BLOWTORCH))
 		to_chat(user, SPAN_WARNING("You need a stronger blowtorch!"))
-		return
+		return FALSE
 
 	if(health == maxhealth)
 		to_chat(user, SPAN_WARNING("[src] doesn't need repairs."))
-		return
-
+		return FALSE
+// if this block fixing multiple cades i should rework this.
 	if(user.action_busy)
 		to_chat(user, SPAN_WARNING("You are currently doing something else"))
-		return
+		return FALSE
 
-	if(damage_state == welder_fix && damage_state != null)
+	if(damage_state >= welder_fix && damage_state != null)
 		to_chat(user, SPAN_WARNING("[src] has sustained too much structural damage to be repaired."))
-		return
+		return FALSE
--- a/code/game/objects/structures/barricade/barricade.dm
+++ b/code/game/objects/structures/barricade/barricade.dm
@@ -42,7 +42,7 @@
 	var/burn_flame_multiplier = 1
 	var/repair_materials = list()
 	var/metallic = TRUE
-	/// var that say what state a barricade can't be fix by welder. null mean it can be repair no matter the state.
+	/// Lower limit of damage beyond which the barricade cannot be fixed by welder. Compared to damage_state. If null it can be repaired at any damage_state.
 	var/welder_fix = null
 
 /obj/structure/barricade/Initialize(mapload, mob/user)
--- a/code/game/objects/structures/barricade/barricade.dm
+++ b/code/game/objects/structures/barricade/barricade.dm
@@ -487,11 +487,11 @@
 	if(health == maxhealth)
 		to_chat(user, SPAN_WARNING("[src] doesn't need repairs."))
 		return FALSE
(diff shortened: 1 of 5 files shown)


Huffie56:
seem to work but since it's the first time i am making a proc  from zero...

code/game/objects/structures/barricade/deployable.dm
```diff
@@ -26,16 +26,9 @@
...
-			return
-		if(user.action_busy)
-			return
-		var/obj/item/tool/weldingtool/welder = item
-		if(health == maxhealth)
-			to_chat(user, SPAN_WARNING("[src] doesn't need repairs."))
-			return
+		attackby_welder(item, user)
 
+		var/obj/item/tool/weldingtool/welder = item
```
Birdtalon, line 31: Casting here is unnecessary as we don't operate on welder after this.

code/game/objects/structures/barricade/barricade.dm
```diff
@@ -466,3 +475,23 @@
...
+		to_chat(user, SPAN_WARNING("[src] doesn't need repairs."))
+		return
+
+	if(user.action_busy)
+		to_chat(user, SPAN_WARNING("You are currently doing something else"))
+		return
+
+	if(damage_state == welder_fix && damage_state != null)
+		to_chat(user, SPAN_WARNING("[src] has sustained too much structural damage to be repaired."))
+		return
```
Birdtalon, line 497: I'm not really sure about this. Wouldn't you want to compare `damage_state` being greater than `welder_fix`?
I.E. if it's damaged more than the minimum threshold for repairing?
Secondly can `damage_state` ever equal null anyway?

```
/obj/structure/barricade/proc/update_damage_state()
	var/health_percent = round(health/maxhealth * 100)
	switch(health_percent)
		if(0 to 25) damage_state = BARRICADE_DMG_HEAVY
		if(25 to 50) damage_state = BARRICADE_DMG_MODERATE
		if(50 to 75) damage_state = BARRICADE_DMG_SLIGHT
		if(75 to INFINITY) damage_state = BARRICADE_DMG_NONE
```

code/game/objects/structures/barricade/deployable.dm
Huffie56: it's required for the proc bellow i think?

Birdtalon: It's not you can just pass item.

code/game/objects/structures/barricade/deployable.dm
```diff
@@ -26,16 +26,9 @@
...
-		if(!HAS_TRAIT(item, TRAIT_TOOL_BLOWTORCH))
-			to_chat(user, SPAN_WARNING("You need a stronger blowtorch!"))
-			return
-		if(user.action_busy)
-			return
-		var/obj/item/tool/weldingtool/welder = item
-		if(health == maxhealth)
-			to_chat(user, SPAN_WARNING("[src] doesn't need repairs."))
-			return
+		attackby_welder(item, user)
```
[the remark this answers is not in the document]
Huffie56: those check should		return FALSE if they are true...
i modify them.

code/game/objects/structures/barricade/barricade.dm
Huffie56: good eyes.
so since damage_state value go from 0-3 from none to heavy i am going to make a check >= welder_fix
since welder_fix hold the limit from when you can't repair a cade.

i don't think damage_state can be null it's probably just a check to prevent an error?..(didn't made that check i assume stuff are there for a reason...)

code/game/objects/structures/barricade/deployable.dm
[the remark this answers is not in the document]
Huffie56: the idea is when one of those check it stop the code there and return chat to the user to explain why the action isn't happening...

Huffie56:
Main issue that's left is my proc that doesn't behave the way i want it to...
when a condition is meet it just spam in chat and doesn't stop the action of welding after....

code/game/objects/structures/barricade/barricade.dm
```diff
@@ -466,3 +475,21 @@
...
+/obj/structure/barricade/proc/attackby_welder(obj/item/item, mob/user)
+	if(!HAS_TRAIT(item, TRAIT_TOOL_BLOWTORCH))
+		to_chat(user, SPAN_WARNING("You need a stronger blowtorch!"))
+		return FALSE
+
+	if(health == maxhealth)
+		to_chat(user, SPAN_WARNING("[src] doesn't need repairs."))
+		return FALSE
+
+	if(damage_state >= welder_fix && damage_state != null && welder_fix != null)
```
Birdtalon, line 491: This will runtime if either `damage_state` and `welder_fix` are `null`.

If you need to check a value is null you should do it before you try to compare it.
i.e.
```
if(isnull(value) || isnull(another_value))
    return
if(value > another_value)
    do_stuff()
```
If you want to specifically check for null use `isnull()`
```suggestion
	if(!(isnull(damage_state) || isnull(welder_fix)) && damage_state >= welder_fix)
```
This will return if either `damage_state` or `welder_fix` is null or both are not null and `damage_state >= welder_fix`

Birdtalon (changes requested):

code/game/objects/structures/barricade/barricade.dm
Huffie56: if welder fix is null it mean barricade as no limit on the state in wish you can repair it.
so my check should return true if it's null and if damage state is null and damage_state>= welder fix...

code/game/objects/structures/barricade/barricade.dm
```diff
@@ -35,6 +42,8 @@
 	var/burn_flame_multiplier = 1
 	var/repair_materials = list()
 	var/metallic = TRUE
+	/// var that say what state a barricade can't be fix by welder. null mean it can be repair no matter the state.
+	var/welder_fix = null
```
Birdtalon, line 46: Could we name this something more descriptive like `welder_lower_damage_limit` or something then we should be good to go,

Birdtalon (changes requested):

code/game/objects/structures/barricade/barricade.dm
```diff
@@ -35,6 +42,8 @@
 	var/burn_flame_multiplier = 1
 	var/repair_materials = list()
 	var/metallic = TRUE
+	/// var that say what state a barricade can't be fix by welder. null mean it can be repair no matter the state.
```
Birdtalon, line 45:
```suggestion
	/// Lower limit of damage beyond which the barricade cannot be fixed by welder. Compared to damage_state. If null it can be repaired at any damage_state.
```

Birdtalon (changes requested):

Birdtalon (approved):

PR #5614, opened 2024-01-31
doubles size of flarepouch

cuberound:
# About the pull request

soubles the internal size of flarepouch to 16 from 8

compared to other pouches flarepouch in comparison to other intem storing pouches is lacking, at current moment it can store one flarebox worth of flares, flarebox is mag sized (4), meanwhile mag pouch is able to hold 3 magazines. Even medium general pouch (size 5) does have more storage then flarepouch technicly. This makes flarepouch more viable and more used. there should not be worry about this allowing flarespam or stuff like that, when you want to do that current moment you take backpack filled with flare boxes not flarepouch anyway. 

<details>

</details>

# Changelog

:cl:
balance: increases flare pouch storage slots from 8 to 16 
/:cl:

### DIFF
--- a/code/game/objects/items/storage/pouch.dm
+++ b/code/game/objects/items/storage/pouch.dm
@@ -1107,8 +1107,8 @@
 	name = "flare pouch"
 	desc = "A pouch designed to hold flares. Refillable with an M94 flare pack."
 	max_w_class = SIZE_SMALL
-	storage_slots = 8
-	max_storage_space = 8
+	storage_slots = 16
+	max_storage_space = 16
 	storage_flags = STORAGE_FLAGS_POUCH|STORAGE_USING_DRAWING_METHOD
 	icon_state = "flare"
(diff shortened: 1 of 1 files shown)


Birdtalon (approved):

Drulikar:
I can agree that the flare pouch is incredibly underwhelming. Especially if you can open a flare pouch and put it in your bag and still draw flares out of it since its only 1 storage level deep. It would be awkward to hold 1.5 pouches worth because refilling will drop them on the floor if you don't keep track (and yes its amusing to me that they spill out - I don't intend to have that removed either). Ultimately this is a side upgrade to the flaregun which is kind of more a troll utility weapon that doesn't need more encouragement.

So I think the flaregun needs to be nerfed before this can happen.

Drulikar:
I am comfortable with this PR if in conjunction with #5623

PR #5628, opened 2024-02-01
handle permanent and sticky bans in database

harryob:
a long time coming

:cl:
admin: permanent and sticky bans are now handled in game, instead of being managed externally
/:cl:


Drulikar (changes requested):
I recommend fixing https://github.com/cmss13-devs/cmss13/blob/master/code/modules/tgui_panel/telemetry.dm#L89

Past that if you want sticky panel in the ESC menu and if you want to not dupe the messages in staff log (I believe its because its logging both as important_log_channel and normal log)

I think everything else is addressed and working.

Drulikar (approved):
Code changes look fine.

PR #5629, opened 2024-02-01
removes dead tor code

harryob:
this isn't enabled on the live server, nor would it ever be

:cl:
server: tor banning functionality is removed
/:cl:


SabreML:
There's config section for this too:
https://github.com/cmss13-devs/cmss13/blob/7846f4194a36d4ba0eb5b7851b82c0df1c5fa874/config/example/config.txt#L154-L155

harryob:
> There's config section for this too:
> 
> https://github.com/cmss13-devs/cmss13/blob/7846f4194a36d4ba0eb5b7851b82c0df1c5fa874/config/example/config.txt#L154-L155

it is no more

code/modules/admin/IsBanned.dm
```diff
@@ -26,19 +27,8 @@
...
-		message_admins("Failed Login: [src] - Banned: ToR")
-		return list("reason"="Using ToR", "desc"="\nReason: The network you are using to connect has been banned.\nIf you believe this is a mistake, please request help at [CONFIG_GET(string/banappeals)]")
-
-	// wait for database to be ready
 
 	. = P.check_ban(computer_id, address)
-	if(.)
-		return .
-
-	return ..() //default pager ban stuff
```
Drulikar, line 41: Are you sure this is related and you do want to change?

harryob: escaped from the atomisation chamber of #5628

Drulikar (approved):

PR #5640, opened 2024-02-02
Mouth and Eyes Surgical Incisions Show on Scanner & Open Surgical Incisions Alone Show Up On Scanner - A Fix To Issue #5639

### LINKED ISSUE #5639: Eyes and mouth incisions do not show up on the Health Analyzer
### Testmerges

irrelevant

### Round ID

irrelevant

### Description of the bug

When scanning a human, if they have an incision on the eyes/mouth zones, the scanner does not pick it up (because the scanner looks for limbs and these two are zones, not limbs; they are not even the children of the head limb).

This leads to "stealth incisions", where the patient keeps bleeding, but the scanner does not pick up why. Self-examining shows the open incision, but a doctor/medic cannot see it at a glance on other people.

The scanner should show if these two zones have an open incision. In `living_healthscan.dm`, on line 581, a simple 

```
for(var/zone in list("eyes", "mouth"))
	if(H.active_surgeries[zone])
		dat += "\t\t [capitalize(zone)]: \t <span class='scanner'>Open surgical incision</span>"
```

could probably solve this, however, I do not know how to add it to tgui, so I cannot PR this change myself.

1. Start an incision on the mouth/eyes area (you need to either add `damage` to the eyes organ or set the head limb's `disfigured` to 1)
2. Analyze it with a handheld scanner

### Issue Bingo

Warfan1815:
# About the pull request

Fixes #5639 

This is done by adding open surgical incisions as a way to have a limb show up on scanners

Also, the "mouth" and "eyes" zones of the head now have their own snowflake way to be picked up on scanners.

Incisions probably should be detectable by scanners

image
image

# Changelog

:cl:
fix: Open surgical incisions now show up on a health scanner even if there is nothing else wrong.
fix: Open surgical incisions in the head "zones" (mouth and eyes) now show up independently on health scanner
/:cl:

### DIFF
--- a/code/modules/mob/living/living_healthscan.dm
+++ b/code/modules/mob/living/living_healthscan.dm
@@ -178,7 +178,7 @@ GLOBAL_LIST_INIT(known_implants, subtypesof(/obj/item/implant))
 				bleeding_check = TRUE
 				break
 
-			if((!limb.brute_dam && !limb.burn_dam && !(limb.status & LIMB_DESTROYED)) && !bleeding_check && !internal_bleeding_check && !(implant && detail_level >= DETAIL_LEVEL_BODYSCAN ) && !(limb.status & LIMB_UNCALIBRATED_PROSTHETIC) && !(limb.status & LIMB_BROKEN) && !(limb.status & LIMB_SPLINTED) && !(limb.status & LIMB_SPLINTED_INDESTRUCTIBLE))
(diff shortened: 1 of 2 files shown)


SabreML:
Try removing the "issue" part of "Fixes issue #5639" in the description. It doesn't seem like it's doing the fancy auto-closing thing.

harryob (approved):

PR #5644, opened 2024-02-02
Re-adds the Mk1 to the SL vendor

567Turtle:
# About the pull request

Re-adds the Mk1 to the SL vendor.

Back in the elden days it was removed because people would play SL just to get the Mk1. This isn't really a problem anymore because of the changed to roleplay (DO YOUR JOB!!!). You'll still have to go to req for extra mags though.

<details>

i am not testing this

</details>

# Changelog
:cl:
add: Re-added the Mk1 to the SL vendor
/:cl:

### DIFF
--- a/code/game/machinery/vending/vendor_types/squad_prep/squad_leader.dm
+++ b/code/game/machinery/vending/vendor_types/squad_prep/squad_leader.dm
@@ -12,6 +12,7 @@ GLOBAL_LIST_INIT(cm_vending_gear_leader, list(
 		list("M56D Heavy Machine Gun", 0, /obj/item/storage/box/guncase/m56d, MARINE_CAN_BUY_KIT, VENDOR_ITEM_REGULAR),
 		list("M79 Grenade Launcher", 0, /obj/item/storage/box/guncase/m79, MARINE_CAN_BUY_KIT, VENDOR_ITEM_REGULAR),
 		list("MOU-53 Shotgun", 0, /obj/item/storage/box/guncase/mou53, MARINE_CAN_BUY_KIT, VENDOR_ITEM_REGULAR),
+		list("M41A Veteran Kit", 0, /obj/item/storage/box/guncase/m41aMK1, MARINE_CAN_BUY_KIT, VENDOR_ITEM_REGULAR),
 		list("XM88 Heavy Rifle", 0, /obj/item/storage/box/guncase/xm88, MARINE_CAN_BUY_KIT, VENDOR_ITEM_REGULAR),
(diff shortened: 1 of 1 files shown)


Drulikar:
Is the argument why this should return both that:
A) The removal of Mk1 from SL worked; It made it so people who were playing to lead were there playing it to lead not just have better stuff
and
B) Because admins are expected to ensure players are playing their role?

567Turtle:
> Is the argument why this should return both that:
> A) The removal of Mk1 from SL worked; It made it so people who were playing to lead were there playing it to lead not just have better stuff
> and
> B) Because admins are expected to ensure players are playing their role?

Notttt quite. I'm just saying that the rules were more permissive towards people sharing their duties for the Mk1 in the past then they are now. With the stricter rules now people will be discouraged from just playing SL for a gun, and if they do it'll be much easier to have them punished.

Nanu308:
I do not see the compelling argument of just "Roleplay is better now". 

There is no reason as to why this specifically wont end up degrading the fact that most people now pick the role to lead, re-adding the Mk1 would not help incentivize that SL's should be leading, but rather the opposite by giving them access to a stronger gun.

567Turtle:
> I do not see the compelling argument of just "Roleplay is better now". 
> 

In all honesty that's a fair point, I just think that the negative repocautions of this will be a lot lesser then they were in the past based off my personal experiences and opinions. SLs get held to a much higher standard now then back then. If they shaft their duties like they did back then today then I think they'll face much harder punishment ICly and OOCly because of how people (in my experience) treat SLs for a lot less.

Drulikar:
This is kind of a situation of its not broken so why fix it. I don't think this is a strong enough argument for the return of the MK1 to their vendor. Even as is there's discussions about how COs shouldn't have some of the combat tools they do. Its a different situation but the same logic can also apply here.

Wherever possible we're going to use mechanics to steer players a certain way. Rules cover grey areas, things that aren't mechanically enforced (yet), or things that just can't really be mechanically enforced. Expecting a handful of admins to police hundreds of players is already a huge task. Its often players who are going to point out an issue and then an admin can investigate, not that an admin is going to preemptively notice a problem always.

I think with a future req update this could be a moot issue if the req line moves faster and its easier for people to get limited equipment from there. I mean already its kinda moot because the person manning req maybe _does_ want to prioritize giving the SLs that gun anyways.

I think the removal was a successful change that has improved the quality of SL players that do play that role and this doesn't need to change.

iloveloopers:
Ff currently the only upside to MK1s are the big mags, wouldn't it make sense for SLs to have them so they can lead for longer? Most SLs I see carry extended MK2 mags anyways

PR #5280, opened 2023-12-23
Changes to bone repairing

BeagleGaming1:
# About the pull request

Bone glue is limited, with 5% glue used per repair
Repairing bones with a screwdriver costs 2 metal sheets per repair
Adds a machine to the USS Almayer that allows you to refill bone glue

# Explain why it's good for the game

By limiting bone glue, you will limit the amount of surgery that happens groundside without sacrificing other resources that have important use, or by having additional supplies delivered. 
If there is a limit to how much surgery can be done, it will force the groundside surgeon to prioritize and consider the severity of the injuries, instead of fully repairing every bone.
This will increase the consequences from getting injured and having broken bones, by requiring you to either work with the downside via splints, or to return shipside.
There should be no changes to shipside surgery, as the machine should be close enough to replenish without having any problems.

# Changelog

:cl:
balance: Repairing bones with bone glue costs a portion of the bone glue's resources
balance: Repairing bones with a screwdriver costs metal
del: Removed bone glue from the medilathe
spellcheck: New description for bone glue
/:cl:

PR #5650, opened 2024-02-03
Revert "Changes to bone repairing (#5280)"

AlakRacc:
# About the pull request

Reverts the PR. 

This PR does not do what it is supposed to. It does not balance groundside surgery in any way other than making it an annoyance. What could have been done different? Many things.
- Make groundside surgery only available in Colony Medbays. 
- Make surgery take longer groundside, even while in a proper surgical theatre.
- Give marines some sort of debuff if they get revived too many times.
- Make surgery give a flat amount of non-healable damage. (Akin to surgical kits in Tarkov)

This PR was just a half-baked attempt at crippling marines. It isn't very well thought out. Changes could further be made to make this PR have an actual effect on gameplay other than what it does currently. Criticism was given to this PR, and it was not taken into consideration.  

<details>

</details>

# Changelog
:cl: Alakaya
del: Removed Bonegel Cost
/:cl:


Drulikar:
This reversions corrupts the Almayer ship map (see checks). Also that's about as weak of an argument for a PR as possible.

AlakRacc:
> This reversion corrupts the Almayer ship map (see checks). Also that's about as weak of an argument for a PR as possible.

Updated. Will look into the corruption.

Drulikar:
> Make groundside surgery only available in Colony Medbays.

I mean sure we can do this - but this has nothing to do with why bonegel usage was introduced.
> Make surgery take longer groundside, even while in a proper surgical theatre.

Already the case that medbay surgery is faster than groundside surgery (at least not in a theatre). If you were to fix up a operating theatre though I believe it would be on par with shipside. But again this has nothing to do with the reason why bonegel usage was introduced.
> Give marines some sort of debuff if they get revived too many times.

They already do in the form of heart damage - but again this has nothing to do with the reason why bonegel usage was introduced.
> Make surgery give a flat amount of non-healable damage. (Akin to surgical kits in Tarkov)

I am not sure where you are going with this suggestion. Bonegel usage doesn't in some way prevent surgery. In fact the PR you propose to revert added the alternative surgery using metal to fix bones too.

---

As I have mentioned in many different places, bonegel/metal is more or less the first consumable for surgery. You could consider drugs and blood otherwise. Bonegel running out is to create a logistical problem. Shipside it practically is infinite, but when taking it groundside you need to get a supply chain with the ship to keep it stocked.

You can expect the vendors that can be used groundside to also have features removed (such as refilling stacks of splints and such) in favor of needing to be restocked with supplies from requisitions. No doubt its a harder logistical problem right now than what it will be once a disposals chute is made between the hangar and requisitions, but by no means has bonegel usage made bone repair surgery impossible.

Ultimately these changes are to make make transport pilots and requisitions more interesting jobs to play. The roadmap task that has existed some time on this is here: https://github.com/orgs/cmss13-devs/projects/6/views/1?pane=issue&itemId=23005516 I have added some additional notes to it today with what changes I want to do for it.

---

I appreciate the enthusiasm to make this PR, and the quality of what you have outlined is drastically better than what you originally opened with. But I will be closing it now to ensure you don't sink more time into something that we aren't going to accept. I hope you can trust that ultimately this will make a more interesting experience in the end. Regardless hopefully this PR has given you some experience in how the development process works. I look forward to seeing your enthusiasm in your next PR.

Issue #5610, opened 2024-01-31
tgui_input_number with {min,max}_value will not allow you to enter a number inbetween


harryob:
### Testmerges

N/A

### Round ID

N/A

### Description of the bug

tgui_input_number will always take the greatest number when a max_value is provided. you can copy paste values into the field to bypass this, but it can't be entered directly

![CleanShot 2024-01-31 at 13 00 47](https://github.com/cmss13-devs/cmss13/assets/55142896/5bf43485-770b-4ec9-a16d-796918955900)

### What's the difference with what should have happened?

permitted to enter a number between the min_value and max_value

### How do we reproduce this bug?

1. use a tgui_input_number with restricted values, such as the age field on character setup
2. attempt to enter a number
3. you will now be 90

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

ItsVyzo:
Copy/pasting is a "temporary" fix

PR #5651, opened 2024-02-03
Fix bug with RestrictedInput TGUI component

mullenpaul:
# About the pull request

Should fix bug in restricted input TGUI component. The number was being clamped on every input, which meant that applying a number of 30 into a range of 19-90, 3 would be bounded to 19 and then the 0 would make the value 190, being bounded to 90.

This change performs the check on blur.

Fixes https://github.com/cmss13-devs/cmss13/issues/5610

<details>

</details>

# Changelog
:cl:
code: rework logic of restrictedinput component to reduce checks
/:cl:

### DIFF
--- a/code/modules/tgui/tgui_number_input.dm
+++ b/code/modules/tgui/tgui_number_input.dm
@@ -145,6 +145,7 @@
 		"min_value" = min_value,
 		"preferences" = list(),
 		"title" = title,
+		"integer_only" = integer_only
 	)
 
 /datum/tgui_input_number/ui_data(mob/user)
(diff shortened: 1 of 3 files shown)


Drulikar (changes requested):
Nothing sticks out to me as an issue in code.

In testing, this seems to also make it possible to type out negative numbers as well. However, it doesn't seem possible to enter floats. I'm not sure where the error is in code, but putting a logger into the `const getClampedNumber` it tells me that the value for `allowFloats` is `undefined`. Sorry I don't know where its not getting passed correctly yet.

Drulikar (approved):
Change appear to be working fine.

Only note is if you say have a integer only input from 0-10 if you enter 5.5 it will treat the input as if it was 55 and then cap it to 10. It'd probably be better to just round or truncate after the decimal but up to you.

Drulikar:
I will tweak this in the future if no one else gets to it.

PR #5658, opened 2024-02-04
Admin logs for 3rd Party Victory events

blackdragonTOW:
# About the pull request

Adds logging in the event that W-Y, UPP, or CLF win during a xeno minor.

Would be good to track how common (or rare) these sorts of things are, which would also be good feedback on if its worthwhile to make more faction victory moments.

<details>

</details>

# Changelog
:cl:
admin: added logging for 3rd party victory events
/:cl:

### DIFF
--- a/code/game/gamemodes/colonialmarines/colonialmarines.dm
+++ b/code/game/gamemodes/colonialmarines/colonialmarines.dm
@@ -372,10 +372,16 @@
 				var/living = headcount["total_headcount"]
 				if ((headcount["WY_headcount"] / living) > MAJORITY)
 					musical_track = pick('sound/theme/lastmanstanding_wy.ogg')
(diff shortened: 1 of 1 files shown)


SabreML:
It would probably make more sense for this to use `GLOB.round_statistics` rather than a game log. I assume it would be trackable in admin tools either way, but it would be easier if it was grouped with the other stats.

blackdragonTOW:
> It would probably make more sense for this to use `GLOB.round_statistics` rather than a game log. I assume it would be trackable in admin tools either way, but it would be easier if it was grouped with the other stats.

X Minors are already tracked, but these are subsets of Xeno Minors that are only triggered off of math at the end of the round.  There's no strictly defined faction victories and each of these still count as a Xeno Minor still.

Birdtalon (approved):

PR #5664, opened 2024-02-04
Fixes trash carts not working with fultons

### LINKED ISSUE #5621: You can fulton trash carts, but the fulton recovery system cannot pick it up
### Testmerges

Irrelevant

### Round ID

Multiple rounds

### Description of the bug

The fulton device can send trash carts up in the air, but when you try to retrieve it on the dropship with the fulton recovery system, it just fails. No message, nothing, you can press the button 500 times, it does not react, and the cart does not get moved to the dropship either.

You either should not be able to fulton trash carts or the recovery system should be able to pick it up.

1. Spawn a `/obj/item/stack/fulton` and a `/obj/structure/closet/crate/trashcart`
2. Fulton the latter with the former
3. Try to pick it up with a dropship

### Issue Bingo

SabreML:
# About the pull request

Fixes #5621, which was caused by a runtime in `/obj/structure/dropship_equipment/fulton_system/proc/automate_interact()`.
The cart's entry in `possible_fultons` has a key of `"�\u{16}trash cart"`, but the `fulton_choice` passed from the UI is just `"trash cart"`. Because they don't match, `fult` is unable to be set, and an error is thrown on L1154.
https://github.com/cmss13-devs/cmss13/blob/2657502ad1c5de88fd0aef35072900a736be84fe/code/modules/cm_marines/dropship_equipment.dm#L1141-L1154

Bugfix!

<details>

**Before:**
Runtime error. *(I can't be bothered to go through the set up again just to record this)*

**After:**

https://github.com/cmss13-devs/cmss13/assets/57483089/9ee496a4-d201-43b2-95d2-963dcc7c8569

</details>

# Changelog
:cl:
fix: Fixed trash carts not being pick-uppable by the dropship's fulton recovery system.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/crates.dm
+++ b/code/game/objects/structures/crates_lockers/crates.dm
@@ -335,7 +335,7 @@
 	icon_closed = "closed_supply"
 
 /obj/structure/closet/crate/trashcart
-	name = "\improper trash cart"
+	name = "trash cart"
 	desc = "A heavy, metal trashcart with wheels."
 	icon_state = "closed_trashcart"
 	icon_opened = "open_trashcart"
@@ -437,4 +437,3 @@
 	density = TRUE
 	icon_opened = "open_mcart_y"
 	icon_closed = "closed_mcart_y"
-
--- a/code/modules/cm_marines/dropship_equipment.dm
+++ b/code/modules/cm_marines/dropship_equipment.dm
@@ -1144,9 +1144,11 @@
 
 	var/list/possible_fultons = get_targets()
 
-	var/obj/item/stack/fulton/fult = possible_fultons[fulton_choice]
 	if(!fulton_choice)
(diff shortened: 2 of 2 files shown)


harryob (approved):

SabreML:
> Tested locally, seems to work but the trash cart layers under the attachment image
(trash cart layer = 2.98, fulton attachment layer = 3.04)
<hr>

Yeah I noticed that too. As far as I can tell it's probably been an issue for a few years.
`/obj/structure/dropship_equipment` blame:
image
`/obj/structure/closet` blame:
image

It's a bit late for me now, but I can try looking into a fix tomorrow? One of them probably just needs to be moved to a different plane.

SabreML:
Done! I've put it in a second PR rather than this one just in case. (#5673)

PR #5687, opened 2024-02-09
fixes custom chemicals not working on the flamer nozzle

iloveloopers:
# About the pull request

custom chems now work on the flamer nozzle

bug fix
may impact balance minorly due to costing less when making a flamer chem cause of the lack of flowing property required when using the nozzle

https://github.com/cmss13-devs/cmss13/assets/140007537/33dc6ad4-145d-477d-a357-a7c1809eefe0

# Changelog
:cl:
fix: custom flamer chemicals now work on the flamer nozzle
/:cl:

### DIFF
--- a/code/datums/ammo/ammo.dm
+++ b/code/datums/ammo/ammo.dm
@@ -237,8 +237,8 @@
 
 		P.fire_at(new_target, original_P.firer, original_P.shot_from, P.ammo.max_range, P.ammo.shell_speed, original_P.original) //Fire!
 
-/datum/ammo/proc/drop_flame(turf/Turf, datum/cause_data/cause_data) // ~Art updated fire 20JAN17
-	if(!istype(Turf))
+/datum/ammo/proc/drop_flame(turf/turf, datum/cause_data/cause_data) // ~Art updated fire 20JAN17
+	if(!istype(turf))
 		return
 	if(locate(/obj/flamer_fire) in Turf)
 		return
--- a/code/datums/ammo/ammo.dm
+++ b/code/datums/ammo/ammo.dm
@@ -251,4 +251,4 @@
 	chemical.fire_penetrating = reference.fire_penetrating
 	chemical.burn_sprite = reference.burn_sprite
 
-	new /obj/flamer_fire(Turf, cause_data, chemical)
+	new /obj/flamer_fire(turf, cause_data, chemical)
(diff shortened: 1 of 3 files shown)


cuberound:
variables could use renaming T to turf P to projectile and so on please

iloveloopers:
> variables could use renaming T to turf P to projectile and so on please

done.
not touching the rest of the code.

code/datums/ammo/ammo.dm
```diff
@@ -237,11 +237,18 @@
...
 
-	var/datum/reagent/R = new flamer_reagent_type()
-	new /obj/flamer_fire(T, cause_data, R)
+	var/datum/reagent/reference = GLOB.chemical_reagents_list[flamer_reagent_id] //this is probably a bad way of doing this but it works
+	var/datum/reagent/chemical = new reference.type
+	chemical.intensityfire = reference.intensityfire
+	chemical.durationfire = reference.durationfire
+	chemical.burncolor = reference.burncolor
+	chemical.fire_penetrating = reference.fire_penetrating
+	chemical.burn_sprite = reference.burn_sprite
```
Drulikar, line 252:
```suggestion
	var/datum/reagent/chemical = GLOB.chemical_reagents_list[flamer_reagent_id]
```
/obj/flamer_fire/Initialize is already going to new the chemical anyways and use make_alike() no need to partially do it.

Can confirm this works w/ the nozzle just fine too, but feel free to try it yourself: 
image

code/datums/ammo/ammo.dm
```diff
@@ -84,7 +84,7 @@
 	/// that will be given to a projectile with the current ammo datum
 	var/list/list/traits_to_give
 
-	var/flamer_reagent_type = /datum/reagent/napalm/ut
+	var/flamer_reagent_id = "utnapthal"
```
Drulikar, line 87: Just a note: if we we required 515 this would be better as `var/flamer_reagent_id = /datum/reagent/napalm/ut::id` otherwise we could set it like `flamer_reagent_id = initial((/datum/reagent/napalm/ut):id)` in new or Initialize. But I wouldn't worry about this since I don't expect us to be changing the reagent ids and it should be pretty obvious if it gets renamed and things break.

Drulikar (changes requested):

Drulikar (approved):

PR #4229, opened 2023-08-22
fancy lighting

harryob:
this was a long time coming

many thanks to:
https://github.com/tgstation/TerraGov-Marine-Corps/pull/1964 for the lighting controller (A-lexa)
https://github.com/tgstation/TerraGov-Marine-Corps/pull/4747/ and https://github.com/tgstation/TerraGov-Marine-Corps/pull/7263 for the lighting (TiviPlus)
https://github.com/tgstation/tgstation/pull/54520 for the dir lighting component
tgstation/tgstation#75018 for the out of bounds fix in lighting
https://github.com/tgstation/TerraGov-Marine-Corps/pull/6678 for the emissives (TiviPlus) (but this is still a 
- [ ] to do)

:cl: TiviPlus, A-lexa, SyncIt21, Azarak, harryob, MorrowWolf, Zonespace
add: shadows are now in the game, and light follows you.
/:cl:

PR #5709, opened 2024-02-12
Reverts flare buff from #4229

Mister-moon1:
# About the pull request
Reverts the flare buff that came alongside fancy lighting from 7 range to 5

Since the fancy lighting change, other factors have come in to play, most notably the rise in the flare gun, the increase in pouch storage and the buff to armour light, making controlling lighting incredibly easy just with 1, with them being infinitely more difficult to deal with than flashlights and a lot more accessible while being safer to use with other utility.

 I believe it is fair to revert the buff especially considering how bright flares are in real life, being more of a concentrated amount of light, not overpowering a military issue flashlight 10 fold without the cone effect to boot.

With how considerable the nerf would be I doubt they will fall into the realm of uselessness, with them still being a great option for scouting ahead of what is visible and still being very easy to carry, just needing an additional flare or so for a room that would be lit up by one.

https://imgur.com/5VNOQrh

<details>

</details>

# Changelog

:cl:
balance: Reverted flare range buff 7->5 tiles
/:cl:

### DIFF
--- a/code/game/objects/items/devices/flashlight.dm
+++ b/code/game/objects/items/devices/flashlight.dm
@@ -238,7 +238,7 @@
 	desc = "A red USCM issued flare. There are instructions on the side, it reads 'pull cord, make light'."
 	w_class = SIZE_SMALL
 	light_power = 2
-	light_range = 7
+	light_range = 5
 	icon_state = "flare"
 	item_state = "flare"
 	actions = list() //just pull it manually, neckbeard.


GoldenDarkness55:
nerf rail flashlights aswell 🙏

Mister-moon1:
I believe that lighting attached to the person is fine because it is a requirement for them to be in the area, flares however are a throw and forget item that linger for a long while, producing more light than anything else and being easy to carry in large amounts

Drulikar (approved):
It seems fine enough; can tweak it later if it proves to be problematic (e.g. maybe larger but more red).

after flare
before flare

PR #5716, opened 2024-02-13
Jensenshades are now proper thermal glasses

iloveloopers:
# About the pull request
JensenShades were a type of thermals that gave you see_mobs but not night vision, with the changes to lighting they don't work anymore as you now need light to see a mob through walls even if you have see_mobs.
this turns them into proper thermal glasses with night vision

bug bad
item works now

probably works

# Changelog
:cl:
fix: Jensenshades are now proper thermal glasses
/:cl:


harryob:
kinda fine with these how they are, honestly. they're 1) meant to be security hud glasses, which this does remove and 2) i don't like rng night vis, i'm so sure we've had to hunt that down before

PR #5542, opened 2024-01-24
Xeno strains refactor

SabreML:
# About the pull request

Refactors xeno strains by taking out the old `/datum/xeno_mutator`/`/datum/mutator_set` system and replacing it with a new `datum/xeno_strain`, since as far as I'm aware mutators haven't been touched in almost 5 years.

As a side note, the DMI changes are from me deleting the icon states of the unused 'Tremor' Burrower strain, since I figured it fit within the bounds of the PR. (The strain's sprites are identical to that of the base caste.)

# Explain why it's good for the game

The new datum should be a lot more streamlined than the old system, and require much less boilerplate when making a new strain. Ideally this could even replace `/datum/behavior_delagate` with signal handlers in the future, although a few more refactors would be required before that's possible.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
refactor: Refactored xeno strains.
add: Added a keybind setting to open the 'Purchase Strain' window. (Also renamed the macro)
/:cl:

PR #5725, opened 2024-02-15
Xeno strains unit test

SabreML:
# About the pull request

Adds a unit test for xeno strains.
I could have included this in #5542, but I thought it would be better in its own PR in order to keep the checks and everything separate.

The removed sprites for the Trapper, Charger, Gardener, and Watcher strains are (as far as I'm able to tell) completely identical to that of the base caste, and are only there in order to make the old system work properly.

Currently, this includes the following tests:
- Checking that the `caste_type` variable on each `/mob/living/carbon/xenomorph` subtype has a corresponding entry in `GLOB.xeno_datum_list`. (Not really strain related but I figured no harm in including it)
- Checking that each `/datum/xeno_strain` subtype has its own `name` and `description`.
- Making sure that each strain's `actions_to_add` and `actions_to_remove` lists are added and removed from the xeno properly. (Taking into account any which are re-added.)
- Making sure that the strain's `behavior_delegate_type` is successfully added to the xeno.
- If the strain doesn't have an `icon_state_prefix`, going through the xeno's `icon` file and double checking that there aren't strain sprites in there (That is, sprites beginning with the strain's name). If there are any, then `icon_state_prefix` should probably be set to use them.
- If the strain *does* have an `icon_state_prefix`, making sure that the xeno's sprite actually got changed to the strain version.

More unit tests! (woo)

<details>

</details>

# Changelog
:cl:
code: Added a unit test for xeno strains.
/:cl:


SabreML:
damn, had to be something

harryob (approved):

PR #5728, opened 2024-02-16
Allow surgery on select tiles on lifeboats

hislittlecuzingames:
# About the pull request

Allows Surgery on select tiles on Lifeboats only. 

Video demonstrates test. 

Once life boats are "away" successfully, there is no way to do surgery on IB. 
If crash landed, still no surgery. 

I get that "if it's flying, you can't surgery." but these aren't moving up and down a combat zone. they're either stationary, crashing, or safely away. 

I limited it to two tiles. Figured that's enough. 
I also put it in the back of the boat so it's more risky being far from lifeboat controls so if they're coming, you gotta run and you'll be seen. 
Also thematic sense that the pilot needs to fly, not be concerned about blood and stuff.

Made the tiles white for visual distinction so it's very obvious that "these are surgery tiles" instead of dealing with "oh yeah, THIS tile is the one you can operate on." 

put tiles down. Made sure I could do surgery. made sure I couldn't do surgery if patient was not on said tile. 

<details>

Same video. 
https://youtu.be/rwDW8lstr1Y

https://www.youtube.com/watch?v=rwDW8lstr1Y

</details>

# Changelog

:cl:
qol: Allows surgery on specific tiles on lifeboats. 
/:cl:

### DIFF
--- a/maps/shuttles/lifeboat-port.dmm
+++ b/maps/shuttles/lifeboat-port.dmm
@@ -239,6 +239,11 @@
 	},
 /turf/template_noop,
 /area/shuttle/lifeboat)
+"BZ" = (
+/turf/open/floor/almayer{
+	icon_state = "dark_sterile"
+	},
+/area/shuttle/lifeboat)
 "Ct" = (
 /obj/structure/surface/table/reinforced/almayer_B,
 /obj/structure/machinery/computer/shuttle/lifeboat{
@@ -533,7 +538,7 @@ fr
 Rs
 Id
 jX
-nn
+BZ
 Ro
 iR
 vZ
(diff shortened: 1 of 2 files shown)


Nanu308 (changes requested):
Not a fan of this as it makes little sense to me, it's a unsuitable place for surgery.

Also, it looks really bad visually/aesthetically just plopping down two white tiles

harryob:
> I get that "if it's flying, you can't surgery." but these aren't moving up and down a combat zone. they're either stationary, crashing, or safely away.

if you want to make it so the lifeboats can have surgery performed on them while they're not in transit, do that. otherwise, i'm not particularly interested in this solution

PR #5751, opened 2024-02-18
Makes observers able to see the UI of Xeno players (and also action buttons)

SabreML:
# About the pull request

Makes auto-observing work for xenomorphs *and* humans by moving it to `/carbon`, and also makes it show the target's action buttons.

I'm pretty confident that this all works correctly, but I would still recommend testmerging it anyway.
It's ~~probable~~possible that I missed something somewhere and an observer could get stuck being unable to click anything, or something like that.

This makes auto-observing a bit more consistent, more interesting for observers, and could actually help newer players to learn how to play some castes. (Which abilities to use and when, where to target, etc.)

<details>

https://github.com/cmss13-devs/cmss13/assets/57483089/54fbb8c6-f9fa-444b-b80c-228b4e6e3480

</details>

# Changelog
:cl:
add: Added 'observe' functionality to Xenomorphs, allowing observers to view the target's UI.
add: Made observing a player also show their action buttons.
/:cl:

### DIFF
--- a/code/_onclick/observer.dm
+++ b/code/_onclick/observer.dm
@@ -14,6 +14,10 @@
 	if(..())
 		return TRUE
 
+	// If the player is auto-observing someone, and they clicked on part of the target's UI.
+	if((observe_target_client || observe_target_mob) && istype(target, /atom/movable/screen))
+		return TRUE
+
 	if (mods["shift"] && mods["middle"])
 		point_to(target)
 		return TRUE
--- a/code/_onclick/observer.dm
+++ b/code/_onclick/observer.dm
@@ -14,10 +14,6 @@
 	if(..())
 		return TRUE
 
-	// If the player is auto-observing someone, and they clicked on part of the target's UI.
-	if((observe_target_client || observe_target_mob) && istype(target, /atom/movable/screen))
-		return TRUE
-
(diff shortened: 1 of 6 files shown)


code/_onclick/observer.dm
```diff
@@ -14,6 +14,10 @@
 	if(..())
 		return TRUE
 
+	// If the player is auto-observing someone, and they clicked on part of the target's UI.
+	if((observe_target_client || observe_target_mob) && istype(target, /atom/movable/screen))
```
fira, line 18: Is this necccessary? Won't it prevent using observer jump alerts while following someone?

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -311,6 +311,29 @@
 	<BR>"}
 	show_browser(user, dat, name, "mob[name]")
 
+/mob/living/carbon/human/auto_observed(mob/dead/observer/observer)
```
fira, line 314: Please codedoc this, it's not very obvious what it does out there

SabreML: It should inherit the docs from `/mob/living/carbon/proc/auto_observed()` (at least in VSCode), but you are right.
I'll add something to explain this one better.

PR #5755, opened 2024-02-19
Fix CMB ERT & Path Based ERT Calling

Drulikar:
# About the pull request

This PR fixes the black market CMB automatic ERT and makes get_specific_call ~~more flexible and~~ efficient.

Fixes `## ERROR: get_specific_call could not find emergency call 'Inspection - Colonial Marshal Ledger Investigation Team'` caused by a grammar fix from Marshal to Marshals. This should generally be harder to cause since the call can now only be called via a path.

<details>

image
image

</details>

# Changelog
:cl: Drathek
fix: Fixed automatic CMB ERT caused by black market heat
code: get_specific_call now only accepts a path or a string of a path
/:cl:

### DIFF
--- a/code/datums/emergency_calls/emergency_call.dm
+++ b/code/datums/emergency_calls/emergency_call.dm
@@ -105,13 +105,6 @@
 		em_call.activate(quiet_launch, announce_incoming)
 		return
 
-	for(var/datum/emergency_call/cur_em_call as anything in all_calls) //Loop through all potential candidates
-		if(cur_em_call.name == call_name)
-			var/datum/emergency_call/em_call = new cur_em_call.type()
-			em_call.objective_info = info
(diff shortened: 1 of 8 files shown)


harryob:
> code: All uses of get_specific_call now use a path instead of name

out of curiosity, if you're happy to refactor all of `get_specific_call`, why preserve the old functionality at all?

Drulikar:
> out of curiosity, if you're happy to refactor all of `get_specific_call`, why preserve the old functionality at all? this kinda dependance on strings has a bad codesmell, and i don't know why we'd facilitate it if we can deprecate it

Ultimately its working code that is now is unused because all situations now are either a path or stringified path.

harryob (approved):

PR #5761, opened 2024-02-20
Auto Delete Foodless MREs

Steelpoint:
# About the pull request

MRE's that contain no food will now be automatically deleted.

Current MRE's mean the FOB, and frontline where a resupply crate lands, are littered with endless amounts of opened MRE's that are filled with the fluff items. This not only clutters up the area but it means if you are desperate for food, you have to search these half-used MRE's.

With this change, this means that a MRE on the ground will actually contain food instead of matches or water, and this will significantly declutter the FOB and frontline of food items. 

<details>

</details>

# Changelog
:cl:
add: MRE's will now auto-delete if they do not contain and food items when they are discarded. 
/:cl:

### DIFF
--- a/code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
+++ b/code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
@@ -305,7 +305,7 @@
 /obj/structure/machinery/cm_vending/sorted/cargo_guns/squad/populate_product_list(scale)
 	listed_products = list(
 		list("FOOD", -1, null, null),
-		list("MRE", round(scale * 5), /obj/item/storage/box/MRE/nofluff, VENDOR_ITEM_REGULAR),
+		list("MRE", round(scale * 5), /obj/item/storage/box/MRE, VENDOR_ITEM_REGULAR),
 		list("MRE Box", round(scale * 1), /obj/item/ammo_box/magazine/misc/mre, VENDOR_ITEM_REGULAR),
 
 		list("TOOLS", -1, null, null),
--- a/code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
+++ b/code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
@@ -305,7 +305,7 @@
 /obj/structure/machinery/cm_vending/sorted/cargo_guns/squad/populate_product_list(scale)
 	listed_products = list(
 		list("FOOD", -1, null, null),
-		list("MRE", round(scale * 5), /obj/item/storage/box/MRE, VENDOR_ITEM_REGULAR),
+		list("MRE", round(scale * 5), /obj/item/storage/box/MRE/nofluff, VENDOR_ITEM_REGULAR),
 		list("MRE Box", round(scale * 1), /obj/item/ammo_box/magazine/misc/mre, VENDOR_ITEM_REGULAR),
 
 		list("TOOLS", -1, null, null),
--- a/code/game/objects/items/storage/boxes.dm
+++ b/code/game/objects/items/storage/boxes.dm
@@ -774,23 +774,6 @@
 		isopened = 1
 		icon_state = "mealpackopened"
 
-/obj/item/storage/box/MRE/nofluff
-
-/obj/item/storage/box/MRE/nofluff/fill_preset_inventory()
-	pickfluffflavor()
-
-/obj/item/storage/box/MRE/nofluff/proc/pickfluffflavor()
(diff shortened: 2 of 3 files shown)


code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
```diff
@@ -305,7 +305,7 @@
 /obj/structure/machinery/cm_vending/sorted/cargo_guns/squad/populate_product_list(scale)
 	listed_products = list(
 		list("FOOD", -1, null, null),
-		list("MRE", round(scale * 5), /obj/item/storage/box/MRE, VENDOR_ITEM_REGULAR),
+		list("MRE", round(scale * 5), /obj/item/storage/box/MRE/nofluff, VENDOR_ITEM_REGULAR),
 		list("MRE Box", round(scale * 1), /obj/item/ammo_box/magazine/misc/mre, VENDOR_ITEM_REGULAR),
```
Huffie56, line 309: isn't this MRE Box suppose to be no fluff to?
/obj/item/storage/box/MRE/nofluff

Steelpoint: the box contains the no-fluff MREs

Huffie56: naming in the vendor was  wrong? it's "MRE" when it's should be "MRE Box" and bellow it should be the opposite?

Huffie56:
maybe change the name and description for all four so we know what MRE type (box and packet) we got?

code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
Steelpoint: MRE box is the big box with all the items. So its technically correct as far as the average player is concerned

Steelpoint:
I don't get what you want to see changed, the item names are unchanged and consistant with what players expect.

The MRE's are technically boxes but no one referes to them as a box item, whereas the big container that holds a lot of MREs is seen as a box even if its not a box.

code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
Huffie56: yea your right after looking into it it's just the path for the object that made me very confused...
sorry

Steelpoint:
Whilst I think removing the fluff items from all MREs would completely remove the FOB clutter issue, however I cannot discount the chance some people may use the fluff items in the MREs ergo I've elected to keep the standard issue MRE's the default ones.

Staykeu:
An idea I had a while ago was a sort of mid-way sprite when only fluff items remain in an MRE. That or the regular crushed up sprite being visible when only fluff items remain. I believe this would solve the issue without dialing back the amount of fluff items that are available to more roleplay-centered marines (or those who just like the free smokes (virtual nic addicts)).

I, along with a handful of other people, do enjoy the fluff items included in MREs, but I still think FOB clutter is insanely exponentialized by non-crushed up (basically useless) MREs. Though I do want to have the ability to pick out these fluff items at liberty of my own will, be they from MREs strewn along the ground or one that I dropped with.

Nanu308:
I don't think we should remove fluff items, instead, just add a option for one to crumple the MRE pack even if it contains the fluff items

Steelpoint:
The issue with that @Nanu308 is that it would need to be automated as literally no one is going to crumple their MRE manually.

Making it a automatic action is beyond my abilities

Drulikar:
Try above, that sounds fine.

Steelpoint:
I'll get to this sometime next week

Steelpoint:
Altered PR to suggested comments.

code/game/objects/items/storage/boxes.dm
```diff
@@ -754,6 +754,18 @@
 	. = ..()
 	isopened = 0
 	icon_state = "mealpack"
+	RegisterSignal(src, COMSIG_ITEM_DROPPED, PROC_REF(try_forced_folding))
+
+/obj/item/storage/box/MRE/proc/try_forced_folding(datum/source, mob/user)
+	SIGNAL_HANDLER
+
+	for(var/obj/item/reagent_container/food/snacks/packaged_meal/meal in src)
```
Drulikar, line 762:
```suggestion
	if(locate(/obj/item/reagent_container/food/snacks/packaged_meal) in src)
```

Drulikar (changes requested):
Also technically you can open it, drop it with food, remove the food, and then it won't be deleted; but not a big deal. Just figured I'd mention it.

Drulikar (approved):

PR #5724, opened 2024-02-15
XM43E1 in regular gameplay

Releasethesea:
# About the pull request

this pull request is an attempt to get the XM43E1 into normal gameplay as a side grade to the sniper, the XM43E1 is a sniper which main quirk is the ability to shoot through walls, i wanted to see it used in gameplay for a long while now since i fixxed it up previously, ive toned down the damage output greatly to be about the same as a normal sniper, compared to the normal sniper (M42A) it takes longer to kill aswell as more shots on some xenos a example of which is taking 1 aimed shot to kill a lesser drone with the M42 while the XM43 takes two. the rifle does very similar damage to the M42, the main feature of it being the ability to pierce xenos, which is why the damage is so similar

if i messed up anything in this PR PLEASE tell me and ill try to fix it, this is my first time making such a big project

# Explain why it's good for the game

variety is good for gameplay i think so having the sniper be able to choose between the XM43 and the M42A would really help with keeping the gameplay cycle fresh

editing this due to forgetting to add some stuff! the rifle does low damage due to it having piercing! by this i mean it goes through xenos even when aimed shotting, which is also the reason behind it lacking night vision goggles

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

on mob of the armour
https://imgur.com/a/VJuPzsw

helmet not on mob 
https://imgur.com/a/SHJvrNG

on mob helmet, due to it being animated im posting like this, i hope thats fine
https://imgur.com/a/BGx6CII

armour on floor/ in hand sprite
https://imgur.com/a/HNcOyym

</details>

tested included several rounds of shooting the xenos and counting the number of shots required to kill them, the normal sniper taking less time and rounds to kill than the XM43E1, wearing the armour in game, firing when not wearing armour and firing when wearing armour

# Changelog
:cl:
add: Adds the XM43E1 a optional sidegrade to snipers main rifle aswell as armour to go along with it
/:cl:

PR #5767, opened 2024-02-21
Updated XM43E1 for USCM Sniper Spec

Kaga-404:
# About the pull request

Alternate PR to #5724 without new armor sprites or items, actual sidegrade-balanced numbers, and patches for multiple old issues with the related code.

Credit to Backsea for fixing up the barrel and ammo sprites, and the purple targeting sprites which are still in the .dmi, though the gun uses the sniper's red laser currently. Alongside Zonespace for the idea of making a Wall-Penetrating Ammo subtype, with more nuance and falloff mechanics to expand on stemming from the Vulture, and all the help in messing around with the code.

The XM43E1 is an experimental anti-materiel rifle that's existed in the code for a while as a test weapon, but had several bugs and insufficient testing. This PR should hopefully allow it to be testmerged in a balanced manner as an M42A sniper sidegrade.

Its design purpose is that of a sniper weapon that fires slower, with less direct combat utility, but more supportive utility and higher impact on shots that actually land. To this purpose, in comparison to the normal sniper rifle (M42A), it:
- Has around half the magazine of the normal sniper (8 vs 15)
- Has slightly less damage potential within the magazine (1000 vs 1050)
- Has less DPS than the M42A (125 vs 140, per 3s)
- Loses out on combining ammunition types, and their special effects. It cannot set targets on fire, hit around corners with flak, or use their respective special effects from aimed shots (blindness and slows).
- Has special damage scaling on its aimed shot, causing it to lose massive amounts of potential damage per magazine when hitting small targets. 1.8x vs Non-Xenos and Runners, 2.4x vs most Xenos, and only the normal 3x vs Big Xenos. Most T3s, the Queen, and Fortified Defenders. This comes out to as little as 1800 damage per magazine from Aimed Shots, compared to Marksman M42A mags dealing up to 3150, with a max single-target potential of 3000 vs 3150.

In return, however, it has higher single-shot damage that allows for more surprising burst, and effects of its own to enable alternative support and utility, alongside special balancing to make them more balanced and interesting to play around, explained below.

Also included in this PR are several backend fixes to actually make this work properly - without this, the weapon would deal more damage when wallbanging, and also deal multiplied damage to mobs inside of walls, like a Queen building beneath herself.

The updated XM43E1's design keeps the weapon's wall-and-mob-piercing, but adds in **_falloff_** for balancing and interactivity, **_which applies to everything it hits_.** While it retains massive damage against walls and structures, able to break even Thick Resin in a single shot, normal metal walls in 4, and reinforced in 8, it now loses a huge portion of its energy when doing so, subsequently reducing its damage and penetration power.

What this means in practice is that it can no longer theoretically multiply its potential damage by up to 8x if everything lines up. In fact, it can barely even get through 2 walls, doing very little damage after that. Mobs, it can handle a bit better, but bodyblocking for more vulnerable targets is a valid strategy to try and protect allies from the AMR. You take the brunt of the hit, and increase their technical DPS, but greatly blunt the damage to the targets behind you.

The actual numbers are 20% reduction in damage for most targets hit, and 60% for dense walls (normal and resin walls) **_and Big Xenos_.** Note that this _does not include Resin Membranes or windows_, which goes into another of the XM43E1's sidegrade utilities: the ability to **_fight Boilers_**, something Sniper Spec has been lacking for years ever since the caste was updated, and a feature this weapon had been originally been designed around.

Membranes will still reduces the AMR's effectiveness temporarily, but such defenses break under sustained focus from it, forcing the Boiler to either seek out allies to cover for it (Fortify Defender escort is back!) or shift to new positions and angles, prepared by builders to force the AMR to expend more ammo.

Additionally, this also affects the Aimed Shot's potential, stacking with the existing special modifiers. Against single targets, the XM43E1 is still far inferior in most cases, only being able to deal more damage when you can only fire off one shot before needing to run, at the cost of way less actual damage and none of the special effects compared to the M42A's special ammos, but in a crowd it gains some benefits and loses others.

The main benefit is that the technical damage it's dealing is increasing. More targets, more hits, more efficiency. **_However, the Aimed Shot target can be covered by others, potentially reducing that multiplier down, or even blocking the shot completely_.** A few big xenos can entirely eat an Aimed Shot for a Queen, **_and they don't take the increased damage_.**

Wallbreaking and wallbanging potential is also heavily reduced by this new effect, as the XM43E1 should usually only be able to hit targets at most 2 walls in, and that's within pretty close ranges. At most sniping ranges it'll probably be only one. It can break a Thick Resin Wall in 1 shot, but it'll only slightly hurt the second one behind that, and have to spend another full shot to break that down, and so on. The Sniper Spec also doesn't have thermals, so while true wallbanging is still possible, it's not consistent like the PMC Sniper, nor is it nearly as effective due to the above mentioned falloff. You'll be looking at around like 50 damage after just 1 wall.

That one shot can save people, though, like the old-school masterkey supports. Gotta get someone out who got dragged behind a door? Blast away like the XM51 does these days, except from a long-range support position. Its design also emphasizes aiming for structural weakspots like Resin Membranes, glass, and doors.

Essentially, it serves as a sidegrade where you trade off superior overall stats, extremely flexible combat utility, and special Aimed Shot effects for:
- Slightly worse damage against single target, but better damage against multiple that line up, with diminished effectiveness for each target, at long ranges, and against big xenos.
- An Aimed Shot 60% as efficient as a M42A's aimed shot against runners, 80% against most xenos, and only equal efficiency against big xenos. It can potentially still hit and get that sweet multiplier through a few xenos, or if a target ducks behind a wall once the shot goes out but before it hits. However, any blocking, whether mobs or walls, will greatly reduce the final damage, and may even completely block it. No more hitting a full damage Aimed Shot through 7 xenos.
- The ability to break singular walls and obstacles with precisely aimed shots, especially resin, but at the cost of a full shot. You only get 8 per mag, and each mag costs the same as a sniper's, and it'll be 3 seconds before you can shoot something else.
- A weapon that isn't CAS/Mortar/OB to counter Boilers, by easily breaking their cover and chipping away at them, and performing better against peeking than the M42A's lower single-shot damage.
<details>

</details>

# Changelog
:cl: Kaga
add: Added the XM43E1 Anti-Materiel Rifle as a pickable sidegrade option for USCM Sniper Specs, through the item in their box. It can shred obstacles, blow a hole through most infantry and their equipment, and even hit a collateral on multiple targets, but its lack of specialized ammunition and small magazine mean that you'll need to make every shot count. Ammo can be bought from Req, at the same cost per magazine as the M42A. Credits to Backsea for the new angled ammunition sprites and the spec-weapon-picker.
balance: Completely updated the XM43E1 AMR, adding falloff to its wall-piercing properties whenever it hits something. Most things reduce its damage by 20%. Dense walls and big xenos reduce it by 60%.
/:cl:

### DIFF
--- a/code/modules/cm_marines/equipment/kit_boxes.dm
+++ b/code/modules/cm_marines/equipment/kit_boxes.dm
@@ -298,7 +298,7 @@
 	return TRUE
 
 /obj/item/spec_kit/sniper_weapon_selector/select_and_spawn(mob/living/carbon/human/user)
-	var/sniper_selection = tgui_input_list(usr, "Select your weapon", "Weapon Selection", list("M42A sniper rifle", "XM43E1 anti-materiel rifle"))
+	var/sniper_selection = tgui_input_list(user, "Select your weapon", "Weapon Selection", list("M42A sniper rifle", "XM43E1 anti-materiel rifle"))
 	if(!sniper_selection || QDELETED(src))
 		return FALSE
 	var/turf/T = get_turf(loc)


cm13-github:

cm13-github:

code/datums/elements/bullet_trait/damage_boost.dm
```diff
@@ -71,7 +61,9 @@ GLOBAL_LIST_INIT(damage_boost_vehicles, typecacheof(/obj/vehicle/multitile))
 /datum/element/bullet_trait_damage_boost/proc/handle_bullet(obj/projectile/P, atom/A)
 	SIGNAL_HANDLER
 
-	atom_type = check_type(A)
+	to_chat(world,"STEP 0: Handling Damage Boost with bullet '[P]', target '[A]', and source '[src]'.")
+	to_chat(world,"STEP 1: Last Atom Hit was [P.last_atom_signaled]. Bonus Projectile Check is [P.bonus_projectile_check].")
```
harryob, line 65: debug

code/modules/cm_marines/equipment/kit_boxes.dm
```diff
@@ -278,6 +289,30 @@
...
+//Sniper weapon selector snowflake
+/obj/item/spec_kit/sniper_weapon_selector
+	name = "\improper Sniper weapon case"
+	desc = "A large case that might contain either a classic M42A sniper rifle, or the experimental XM43E1 anti-materiel rifle. Choose wisely."
+
+/obj/item/spec_kit/sniper_weapon_selector/can_use(mob/living/carbon/human/user)
+	return TRUE
+
+/obj/item/spec_kit/sniper_weapon_selector/select_and_spawn(mob/living/carbon/human/user)
+	var/sniper_selection = tgui_input_list(usr, "Select your weapon", "Weapon Selection", list("M42A sniper rifle", "XM43E1 anti-materiel rifle"))
```
harryob, line 301:
```suggestion
	var/sniper_selection = tgui_input_list(user, "Select your weapon", "Weapon Selection", list("M42A sniper rifle", "XM43E1 anti-materiel rifle"))
```

harryob (changes requested):

Drulikar (changes requested):
Since it doesn't sound like you and Backsea communicated or collaborated together at a minimum you must attribute their work in the change log.

Ultimately this is legal to do but it's very awkward to use the commits from a pr that is still open then in your own pr and it's not from you collaborating with the other person.

Ultimately I recommend you two collaborate together on a single pr. It's fine to do changes after a pr is merged too to adjust or expand the pr. But at this point it's a competing pr that effectively stole the work from the other without full attribution or friendly permission.

Drulikar:
This PR is on hold until the PR it builds off of is either merged or closed. Feel free to continue work on it, especially getting it to a point that even compiles; but it is very likely you will need to resolve conflicts in the future.

PR #5774, opened 2024-02-21
Updated XM43E1 for Sniper Spec

Kaga-404:
# About the pull request

New Updates:
- Using Aimed Shot on a fully-focused target will now change the laser and icon to blue.
- 
- Aimed Shot Damage has been adjusted across the board with two new effects, which stack together.
- The first is **current health damage**, which allows it to stay about as dangerous against full-health Xenos, but reduces the pressure at low health, and the likelihood of going from 375 HP to crit immediately. Low-health burst sniping becomes even more unlikely from Aimed Shots when combined with the second effect, **Focused Fire**. The scaling is 20% against most xenos, and 30% against big xenos.
- **Focused Fire** triggers whenever Aimed Shot lands on a target, though it only modifies damage applied to most xenos at the moment, except Runners. This effect reduces the bonus Aimed Shot Damage by 50% when hitting a new target. However, subsequent Aimed Shot hits against the same xeno will increase the modifier, with the second going up to 75%, and then the 3 onwards at 100%. **This effect resets when landing an Aimed Shot on another target.** If you're the sniper, consider using normal shots for damage and support to preserve your focus on a priority target.
- Defenders are now excluded from the higher-tier size calculations, so they only take 200 + 10% current health damage from Aimed Shots. Focused Fire damage reduction doesn't apply to them because of this, but it might be added in if they still continue to struggle when trying to block shots for allies. However, they may also be targeted less by the ability with this update, and only take normal hits (can survive 4-5 on weeds, which takes the sniper about 16-20s of open sightlines and free shots) because the focus mark is being kept on a more dangerous target.

Fixed PR built from the master branch. Thank you to Zonespace for the idea of making a Wall-Penetrating Ammo subtype, with more nuance and falloff mechanics to expand on stemming from the Vulture, and all the help in messing around with the code.

The XM43E1 is an experimental anti-materiel rifle that's existed in the code for a while as a test weapon, but had several bugs and insufficient testing. This PR should hopefully allow it to be testmerged in a balanced manner as an M42A sniper sidegrade, without the need for staff to spawn it in. It can be picked as a normal option from Spec Vendors, being mutually exclusive with the M42A, as well as cryo spec kits.

Its design purpose is that of a sniper weapon that fires slower, with less direct combat utility, but more supportive utility and higher impact on shots that actually land. To this purpose, in comparison to the normal sniper rifle (M42A), it:
- Has around half the magazine of the normal sniper (8 vs 15)
- Has slightly less damage potential within the magazine (1000 vs 1050)
- Has less DPS than the M42A (125 vs 140, per 3s)
- Loses out on combining ammunition types, and their special effects. It cannot set targets on fire, hit around corners with flak, or use their respective special effects from aimed shots (blindness and slows).
- Has special damage scaling on its aimed shot, causing it to lose massive amounts of potential damage per magazine when hitting small targets. 1.8x vs Non-Xenos and Runners, 2.4x vs most Xenos, and only the normal 3x vs Big Xenos. Most T3s, the Queen, and Fortified Defenders. This comes out to as little as 1800 damage per magazine from Aimed Shots, compared to Marksman M42A mags dealing up to 3150, with a max single-target potential of 3000 vs 3150.

In return, however, it has higher single-shot damage that allows for more surprising burst, and effects of its own to enable alternative support and utility, alongside special balancing to make them more balanced and interesting to play around, explained below.

Also included in this PR are several backend fixes to actually make this work properly - without this, the weapon would deal more damage when wallbanging, and also deal multiplied damage to mobs inside of walls, like a Queen building beneath herself.

The updated XM43E1's design keeps the weapon's wall-and-mob-piercing, but adds in **_falloff_** for balancing and interactivity, **_which applies to everything it hits_.** While it retains massive damage against walls and structures, able to break even Thick Resin in a single shot, normal metal walls in 4, and reinforced in 8, it now loses a huge portion of its energy when doing so, subsequently reducing its damage and penetration power.

What this means in practice is that it can no longer theoretically multiply its potential damage by up to 8x if everything lines up. In fact, it can barely even get through 2 walls, doing very little damage after that. Mobs, it can handle a bit better, but bodyblocking for more vulnerable targets is a valid strategy to try and protect allies from the AMR. You take the brunt of the hit, and increase their technical DPS, but greatly blunt the damage to the targets behind you.

The actual numbers are 20% reduction in damage for most targets hit, and 60% for dense walls (normal and resin walls) **_and Big Xenos_.** Note that this _does not include Resin Membranes or windows_, which goes into another of the XM43E1's sidegrade utilities: the ability to **_fight Boilers_**, something Sniper Spec has been lacking for years ever since the caste was updated, and a feature this weapon had been originally been designed around.

Membranes will still reduces the AMR's effectiveness temporarily, but such defenses break under sustained focus from it, forcing the Boiler to either seek out allies to cover for it (Fortify Defender escort is back!) or shift to new positions and angles, prepared by builders to force the AMR to expend more ammo.

Additionally, this also affects the Aimed Shot's potential, stacking with the existing special modifiers. Against single targets, the XM43E1 is still far inferior in most cases, only being able to deal more damage when you can only fire off one shot before needing to run, at the cost of way less actual damage and none of the special effects compared to the M42A's special ammos, but in a crowd it gains some benefits and loses others.

The main benefit is that the technical damage it's dealing is increasing. More targets, more hits, more efficiency. **_However, the Aimed Shot target can be covered by others, potentially reducing that multiplier down, or even blocking the shot completely_.** A few big xenos can entirely eat an Aimed Shot for a Queen, **_and they don't take the increased damage_.**

Wallbreaking and wallbanging potential is also heavily reduced by this new effect, as the XM43E1 should usually only be able to hit targets at most 2 walls in, and that's within pretty close ranges. At most sniping ranges it'll probably be only one. It can break a Thick Resin Wall in 1 shot, but it'll only slightly hurt the second one behind that, and have to spend another full shot to break that down, and so on. The Sniper Spec also doesn't have thermals, so while true wallbanging is still possible, it's not consistent like the PMC Sniper, nor is it nearly as effective due to the above mentioned falloff. You'll be looking at around like 50 damage after just 1 wall.

That one shot can save people, though, like the old-school masterkey supports. Gotta get someone out who got dragged behind a door? Blast away like the XM51 does these days, except from a long-range support position. Its design also emphasizes aiming for structural weakspots like Resin Membranes, glass, and doors.

Essentially, it serves as a sidegrade where you trade off superior overall stats, extremely flexible combat utility, and special Aimed Shot effects for:
- Slightly worse damage against single target, but better damage against multiple that line up, with diminished effectiveness for each target, at long ranges, and against big xenos.
- An Aimed Shot 60% as efficient as a M42A's aimed shot against runners, 80% against most xenos, and only equal efficiency against big xenos. It can potentially still hit and get that sweet multiplier through a few xenos, or if a target ducks behind a wall once the shot goes out but before it hits. However, any blocking, whether mobs or walls, will greatly reduce the final damage, and may even completely block it. No more hitting a full damage Aimed Shot through 7 xenos.
- The ability to break singular walls and obstacles with precisely aimed shots, especially resin, but at the cost of a full shot. You only get 8 per mag, and each mag costs the same as a sniper's, and it'll be 3 seconds before you can shoot something else.
- A weapon that isn't CAS/Mortar/OB to counter Boilers, by easily breaking their cover and chipping away at them, and performing better against peeking than the M42A's lower single-shot damage.
<details>

</details>

# Changelog
:cl: Kaga
add: Added the XM43E1 Anti-Materiel Rifle as a pickable sidegrade option for USCM Sniper Specs through the vendor, mutually exclusive with the normal M42A Sniper Rifle. The AMR can shred obstacles, blow a hole through most infantry and their equipment, and even hit a collateral on multiple targets or through light cover, but its lack of specialized ammunition and small magazine mean that you'll need to make every shot count. Ammo can be bought from Req, at the same cost per magazine as the M42A.
balance: Completely updated the XM43E1 AMR, adding damage falloff to its wall-piercing properties whenever it hits something alongside range loss. Most things reduce its damage by 20%. Dense walls and big xenos reduce both the damage and range 3 times as much.
/:cl:


code/datums/elements/bullet_trait/damage_boost.dm
```diff
@@ -71,7 +61,7 @@ GLOBAL_LIST_INIT(damage_boost_vehicles, typecacheof(/obj/vehicle/multitile))
 /datum/element/bullet_trait_damage_boost/proc/handle_bullet(obj/projectile/P, atom/A)
 	SIGNAL_HANDLER
 
-	atom_type = check_type(A)
+	src.atom_type = check_type(A)
```
Drulikar, line 64:
```suggestion
	atom_type = check_type(A)
```

Drulikar:
Oh also please rename the lineart.dmi state for this gun from `xm42b` to `xm43e1` so it shows up correctly in weapon stats.

Drulikar (changes requested):
Other thoughts as I observed a round with this:
- **This sniper must be mutually exclusive to the other sniper (as in if you have one, you cannot have the other)**
- I don't think the user has any idea they can fire it past a wall; their aimed shot will end when the target enters cover. Not sure if its intended to shoot through cover only if its at the very end of the shot, or if this is to only apply to manual shots (fine if its this case; just maybe I need to look at the weapon descriptions more).
- Damage multiplier w/o any obstruction as an aimed shot is probably too high overall regardless of size; but I would have to see it used more. Two shots will easily overkill a queen. A good position with this gun means no one can even attempt to engage without being above 80-90% health or risk being killed outright fully through crit. Probably the only feasible way to have this much damage would be if it took a severe penalty for any damage rolling over into crit.

One aimed shot (though close range) on each: 
image

code/datums/supply_packs/spec_ammo.dm
```diff
@@ -122,9 +120,9 @@
 	)
 	cost = 30
 	containertype = /obj/structure/closet/crate/ammo
-	containername = "XM43E1 Anti-Materiel Magazine Crate"
+	containername = "XM43E1 Marksman Magazine Crate"
 	group = "Specialist Ammo"
```
realforest2001, line 124:
```suggestion
	group = "Weapons Specialist Ammo"
```

realforest2001 (changes requested):

code/_globalvars/global_lists.dm
```diff
@@ -483,7 +483,8 @@ GLOBAL_LIST_INIT(available_specialist_sets, list(
 			"Sniper Set",
 			"Demolitionist Set",
 			"Heavy Grenadier Set",
-			"Pyro Set"
+			"Pyro Set",
+			"Anti-materiel Sniper Set"
```
realforest2001, line 487: May be worth moving it in the list to be next to main sniper.

Drulikar (changes requested):

code/datums/ammo/bullet/sniper.dm
```diff
@@ -107,19 +107,78 @@
...
+				else
+					amr.focused_fire_counter = 2
+			else
+				amr.focused_fire_counter = 0
+
+			amr_counter = amr.focused_fire_counter + 1
+			amr.focused_fire_target = WEAKREF(target_mob)
+
+		var/mob/living/living_target = target_mob
+		var/size_damage_mod = 0.8 // 1.8x vs Non-Xenos (225)
```
sl0w3st, line 129: comment and code itself contradict each other

Also, `size_damage_mod` goes to 0 when shot at T3 xenos, due -0.2 on successful xeno check and -0.6 on `MOB_SIZE_BIG` check.
So `((damage * size_damage_mod) + ((target.health + damage) * size_current_health_damage))` becomes just `(target.health + damage) * size_current_health_damage)` capping damage to `(target.health + 125) * 0.3`

GoldenDarkness55:
Why is this being considered? Wall piercing ammo was removed for a reason, same as this not being added as a non-event weapon even though this was already suggested. 

Someone shooting you through all cover from several screens away will never be fair nor fun.

sl0w3st:
So to actually use this bonus stacking ability you have to sit and read xeno tags to distinguish same looking mobs from each other or you loose your stacks by doing aimed shot at any other xeno. Or just plain aim at queen, as it's the single xeno if its type.
Is it that necessary to make it that complicated?

code/datums/ammo/bullet/sniper.dm
```diff
@@ -107,19 +107,104 @@
...
+		else
+			living_target.apply_armoured_damage((damage*size_damage_mod), ARMOR_BULLET, BRUTE, null, penetration)
+
+		// Base 1.8x damage to non-xeno targets (225), 1.6x + 10% current against Runners and Defenders (223), 1.6x + 20% current health against most non-Runner xenos, and +30% current health against Big xenos. -Kaga
+		// This applies after pen reductions. After hitting 1 other thing, it deals 80% damage, or 40% after hitting a dense wall or big xeno.
+
+		if(focused_fire_active)
+			switch(amr_counter)
+				if(1)
+					to_chat(aimed_projectile.firer, SPAN_WARNING("One Hit! You begin to carefully track the target's movements."))
```
Drulikar, line 188:
```suggestion
					to_chat(aimed_projectile.firer, SPAN_WARNING("One hit! You begin to carefully track the target's movements."))
```

code/datums/ammo/bullet/sniper.dm
```diff
@@ -107,19 +107,104 @@
...
+					if(isxeno(target_mob) && isxeno(old_target?.resolve()))
+						var/mob/living/carbon/xenomorph/old_xeno = old_target?.resolve()
+						var/mob/living/carbon/xenomorph/new_xeno = target_mob
+						if(old_xeno.hive == new_xeno.hive)
+							to_chat(old_xeno,SPAN_XENOLEADER("We sense that the far-spitter host has begun targeting another sister."))
+				if(2)
+					to_chat(aimed_projectile.firer, SPAN_WARNING("Two Hits! You're starting to get a good read on the target's patterns."))
```
Drulikar, line 195:
```suggestion
					to_chat(aimed_projectile.firer, SPAN_WARNING("Two hits! You're starting to get a good read on the target's patterns."))
```

code/datums/ammo/bullet/sniper.dm
```diff
@@ -107,19 +107,104 @@
...
+			living_target.apply_armoured_damage((damage*size_damage_mod), ARMOR_BULLET, BRUTE, null, penetration)
+
+		// Base 1.8x damage to non-xeno targets (225), 1.6x + 10% current against Runners and Defenders (223), 1.6x + 20% current health against most non-Runner xenos, and +30% current health against Big xenos. -Kaga
+		// This applies after pen reductions. After hitting 1 other thing, it deals 80% damage, or 40% after hitting a dense wall or big xeno.
+
+		if(focused_fire_active)
+			switch(amr_counter)
+				if(1)
+					to_chat(aimed_projectile.firer, SPAN_WARNING("One Hit! You begin to carefully track the target's movements."))
+					if(isxeno(target_mob) && isxeno(old_target?.resolve()))
```
Drulikar, line 189: This should also check if they are still alive; or you automatically remove the mark when your target dies/qdeleted.

code/datums/ammo/bullet/sniper.dm
```diff
@@ -107,19 +107,104 @@
...
+						var/mob/living/carbon/xenomorph/old_xeno = old_target?.resolve()
```
Drulikar, line 190:
```suggestion
						var/mob/living/carbon/xenomorph/old_xeno = old_target.resolve()
```

Drulikar (changes requested):

code/datums/ammo/bullet/sniper.dm
Drulikar: Still unresolved.

code/datums/ammo/bullet/sniper.dm
```diff
@@ -107,19 +107,104 @@
...
+
+		if(focused_fire_active && !(target_mob.is_dead()))
+			switch(amr_counter)
+				if(1)
+					to_chat(aimed_projectile.firer, SPAN_WARNING("One hit! You begin to carefully track the target's movements."))
+					if(isxeno(target_mob) && isxeno(old_target?.resolve()))
+						var/mob/living/carbon/xenomorph/old_xeno = old_target.resolve()
+						var/mob/living/carbon/xenomorph/new_xeno = target_mob
+						if((old_xeno.hive == new_xeno.hive) && !(old_xeno.stat)) // Must be in same hive and conscious
+							to_chat(old_xeno,SPAN_XENOLEADER("We sense that the far-spitter host is targeting another sister."))
```
Drulikar, line 193: This only occurs if the sniper targets a non-defender (that is alive).

code/datums/ammo/bullet/sniper.dm
```diff
@@ -107,19 +107,104 @@
...
+				size_current_health_damage += 0.1 // 1.6x Damage + 20% current health
+				focused_fire_active = 1 // Focus Fire Required. Only deals 50% bonus damage on a first Aimed Shot, then 75%, then 100%. Resets with a successful aimed shot on another target.
+
+			if(target.mob_size >= MOB_SIZE_BIG && (target.caste_type != XENO_CASTE_DEFENDER))
+				size_damage_mod -= 0.6 // Down to 1x Damage.
+				size_current_health_damage += 0.1 // 1x Damage + 30% current health.
+				focused_fire_active = 1
+				// Most T3s have around 650 to 700 HP, meaning the health modifier grants a MAXIMUM of around 195-210 damage for a total max of 320-335. This is fully focused (3 shots) and at max HP.
+				// Queen takes around 275 at max health and unfocused, 425 fully focused. Defender only takes 200 damage, even while fortified, but still causes falloff like a Big Xeno.
+				// At low health, does little more than a normal shot. Does WORSE than a normal shot if unfocused and hitting through blockers, all of which stack to reduce it.
```
Drulikar, line 169: Maybe defenders should have a different message that you can't get a read on them? Also you may want to consider if you do want the defender to get the eye marker if they aren't actually getting a stack on a defender.

Drulikar (changes requested):

code/datums/ammo/bullet/sniper.dm
```diff
@@ -107,19 +107,104 @@
...
+
+			living_target.apply_armoured_damage((final_xeno_damage), ARMOR_BULLET, BRUTE, null, penetration)
+
+		else
+			living_target.apply_armoured_damage((damage*size_damage_mod), ARMOR_BULLET, BRUTE, null, penetration)
+
+		// Base 1.8x damage to non-xeno targets (225), 1.6x + 10% current against Runners and Defenders (223), 1.6x + 20% current health against most non-Runner xenos, and +30% current health against Big xenos. -Kaga
+		// This applies after pen reductions. After hitting 1 other thing, it deals 80% damage, or 40% after hitting a dense wall or big xeno.
+
+		if((focused_fire_active || isxeno(target_mob)) && !(target_mob.is_dead()))
```
Drulikar, line 185: By allowing not `focused_fire_active` means the build up messages occur for xenos we aren't actually building up on. This looks like an oversight, but I don't know if you intended this. The issue was just that the relief message of losing target would only occur if the next target causes stacking.

Drulikar: See previous feedback if you didn't see it https://github.com/cmss13-devs/cmss13/pull/5774#pullrequestreview-1958670394 Github hiding it because of all the TM messages.

Drulikar (approved):

PR #5777, opened 2024-02-21
Relay Ping Browser

Drulikar:
# About the pull request

This PR adds a way to ping the relays and also easily connect to them. You can access this panel by clicking the ping near the top right of chat. The ping reported here will likely be inflated from what you will actually experience or see reported by TGChat (its highly dependent on how fast your computer runs the JavaScript, and is based only on a single ping per connection), but should still give an accurate relative difference between the different connections.

When the component is loaded, it starts a timer to begin the pings in 1 second. Whenever the component is unloaded, it indicates this to the ping javascript which will cause it to early return on any pending operations.

The configs are added to examples in relays.txt. You can include it in config.txt.

This should give more visibility to the relays, provide an easy way to use them, and give more of an idea the differences in connection speeds to each relay.

Random values during testing:
image

An actual test example (with last forced to time out):
image

Now with gauges:
image

Now with flex layout:
layout rework

# Changelog
:cl: Drathek
ui: Added the relay ping browser accessed by the tgchat ping to test and use alternative connections to the server
ui: Added onConfirmChange prop to Button.Confirm component.
config: Added CONNECTION_RELAY_PING and CONNECTION_RELAY_CON in the relays.txt config that is optionally included in config.txt
/:cl:

### DIFF
--- a/tgui/packages/tgui-panel/ping/PingIndicator.js
+++ b/tgui/packages/tgui-panel/ping/PingIndicator.js
@@ -6,11 +6,12 @@
 
 import { Color } from 'common/color';
 import { toFixed } from 'common/math';
-import { useSelector } from 'tgui/backend';
-import { Box } from 'tgui/components';
+import { useSelector, useBackend } from 'tgui/backend';
+import { Button, Box } from 'tgui/components';
 import { selectPing } from './selectors';
 
 export const PingIndicator = (props) => {
+  const { act } = useBackend();
   const ping = useSelector(selectPing);
   const color = Color.lookup(ping.networkQuality, [
     new Color(220, 40, 40),
@@ -19,9 +20,16 @@ export const PingIndicator = (props) => {
   ]);
   const roundtrip = ping.roundtrip ? toFixed(ping.roundtrip) : '--';
   return (
-    <div className="Ping">
+    <Button
+      lineHeight="15px"
+      width="50px"
+      className="Ping"
+      color="transparent"
+      tooltip="Ping relays"
+      tooltipPosition="bottom-start"
+      onClick={() => act('ping_relays')}>
       <Box className="Ping__indicator" backgroundColor={color} />
       {roundtrip}
-    </div>
+    </Button>
   );
 };
--- a/tgui/packages/tgui-panel/ping/PingIndicator.js
+++ b/tgui/packages/tgui-panel/ping/PingIndicator.js
@@ -25,6 +25,9 @@ export const PingIndicator = (props) => {
       width="50px"
       className="Ping"
       color="transparent"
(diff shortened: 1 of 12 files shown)


iloveloopers:
south america relay when?

tgui/packages/tgui-panel/ping/PingIndicator.js
```diff
@@ -19,9 +20,16 @@ export const PingIndicator = (props) => {
...
-    <div className="Ping">
+    <Button
+      lineHeight="15px"
+      width="50px"
+      className="Ping"
+      color="transparent"
+      tooltip="Ping relays"
+      tooltipPosition="bottom-start"
+      onClick={() => act('ping_relays')}>
       <Box className="Ping__indicator" backgroundColor={color} />
```
SabreML, line 31: The ping indicator circle seems to have been offset down a little bit:
**Before (master)/After (this PR):**
before pre-pr before
<hr>

Tweaking `Ping__indicator`'s `top` value in `tgui\packages\tgui-panel\styles\components\Ping.scss` seems to work to get it back in the centre:
```diff
.Ping__indicator {
  content: '';
  position: absolute;
-  top: 0.5em;
+  top: 0.4em;
```
**Before/After:**
before after

SabreML, line 31: Also I noticed that the whole thing was made slightly wider and that the text is greyed out until hovered over. Not sure if that's intentional or not but I thought I'd mention anyway.

Drulikar: The light theme hover state just really has no indication so I will keep the text color as is. `textColor="label"` would be the next best I can find; but I think it looks fine as is.

Drulikar: light theme changes the padding on the button; so I will just force the padding to how dark theme does it here.

harryob (changes requested):
looks like everyone's getting in on the reviewing action here

tgui/packages/tgui/interfaces/PingRelaysPanel.jsx
```diff
@@ -0,0 +1,158 @@
...
+      'Singapore',
+      'asia-se.cm-ss13.com:8998',
+      'asia-se.cm-ss13.com:1400'
+    );
+    this.startTest(
+      'Australia, Sydney',
+      'aus.cm-ss13.com:8998',
+      'aus.cm-ss13.com:1400'
+    );
+  }
```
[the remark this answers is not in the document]
Drulikar: image

harryob (approved):

PR #3546, opened 2023-06-06
ERT Custom Stations - We back

Nanu308:
# About the pull request
Custom ERT stations for shuttles to land in, most factions will get one. 

# Explain why it's good for the game
More psoul and RP

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl: Nanu, Spartanbobby, Julian56
mapadd: Added W-Y Station, UPP Station, CLF Station and...
admin: Admins can now use the custom ERT stations for events or even base of operations.
/:cl:

PR #5802, opened 2024-02-24
ports lazy templates, uses them for ert stations. distress beacons now start in transit to the ship, and return to their very own ert home

harryob:
title pretty much

actually uses the maps added back in #3546 

the lazy templates are made entirely by ZephyrTFA on /tg/, predominantly in https://github.com/tgstation/tgstation/pull/71785

:cl: ZephyrTFA, harryob
add: framework for new ert stations ported from tgstation/tgstation#71785, by ZephyrTFA
add: distress beacons now start in transit to the ship, and will return to their own home base. they can't keep going back and forth between the ship and their base
/:cl:

this pr also unfucks base lighting, because i apparently did that really bad in #4229


Nanu308 (approved):
damn this mapping so fire who did tht!!!??

lgtm!

maps/templates/lazy_templates/freelancer_ert_station.dmm
Nanu308, line 1: I'm already mapping a new ert so good for now cuz soontm changed anyways

Nanu308 (approved):
mappingegg

maps/templates/lazy_templates/freelancer_ert_station.dmm
harryob: thanks boss 😝

code/datums/emergency_calls/royal_marines.dm
```diff
@@ -2,6 +2,7 @@
 	name = "Royal Marines Commando (Squad) (Friendly)"
 	mob_max = 7
 	probability = 15
+	home_base = /datum/lazy_template/ert/twe_station
```
Drulikar, line 5:
```suggestion
	home_base = /datum/lazy_template/ert/twe_station
	shuttle_id = MOBILE_SHUTTLE_ID_ERT4
```

Without this the shuttle doesn't launch automatically somereason.

code/modules/shuttle/computer.dm
```diff
@@ -217,29 +303,16 @@
...
+		if("launch_home")
+			if(!must_launch_home)
+				return
+
+			if(ert.mode != SHUTTLE_IDLE)
+				to_chat(ui.user, SPAN_WARNING("Unable to return home."))
+				balloon_alert(ui.user, "can't go home!")
+				return
+
+			launch_home()
```
Drulikar, line 315:
```suggestion
			launch_home()
			return TRUE
```
This should still return TRUE so ideally the UI updates immediately rather than after an auto update refresh.

image

However, I also tested this and the issue still occurs somereason where the spooling UI appears up to a second later than it should.

tgui/packages/tgui/interfaces/NavigationShuttle.tsx
```diff
@@ -248,11 +250,54 @@ export const DisabledScreen = (props) => {
...
 };
 
+const LaunchHome = (props) => {
+  const { data, act } = useBackend<NavigationProps>();
+
+  return (
+    <Section title="Automatic Return Enabled">
+      <Button.Confirm
+        fluid
+        content={'Return Home'}
```
Drulikar, line 260: This for whatever reason is visually larger than the other buttons. Also worth noting that the destinations are sharp edged buttons whereas this and the door ones are rounded buttons. The only thing I think needs to change is the font/size should be the same as the destination buttons. Up to you if you also want to standardize the corners too.
image

Drulikar (changes requested):

Drulikar:
Also this has been a debug message on master some time; but I am not exactly sure it even is a problem.
`WARNING: Template shuttle Boarding Shuttle cannot dock at Response Station Landing Pad 5 2 (width too large). in code/controllers/subsystem/shuttles.dm at line 402 src: Shuttle usr: .`
Other than the TWE that needed a shuttle ID, I could not seem to replicate any issue prompting that debug message with any of the ert bases.

code/modules/shuttle/computer.dm
harryob: because it's relying on the autoupdate still, launch_home blocks until the template is loaded

tgui/packages/tgui/interfaces/NavigationShuttle.tsx
```diff
@@ -248,11 +250,54 @@ export const DisabledScreen = (props) => {
...
+  );
+};
+
+const SpoolingDimmer = (props) => {
+  const { data, act } = useBackend<NavigationProps>();
+
+  return (
+    <Dimmer>
+      <Section title="Spooling Up">
+        <ProgressBar value={0.5} />
```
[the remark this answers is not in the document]
harryob: wouldn't work as the value wouldn't update often enough, it's really only there for about a second and a half, even when the server's slower. i'll just make it a spinny icon isntead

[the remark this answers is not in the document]
Drulikar: Thats fine

code/modules/shuttle/computer.dm
Drulikar: Why not CallAsync then?

harryob: > Why not CallAsync then?

then we wouldn't end up hitting the spooling screen - instead, launch_home() just invokes a tgui refresh when spooling is set to TRUE

tgui/packages/tgui/interfaces/NavigationShuttle.tsx
harryob: resolved here https://github.com/cmss13-devs/cmss13/pull/5802/commits/9b622e6cf63d0884ac07ee42ae59f51d6cf05a2d

Drulikar (approved):

PR #5803, opened 2024-02-24
speeds up ci runs by 3 minutes

### LINKED ISSUE #5679: Unit test runs are waiting for the lobby timer to finish rather than skipping it
### Testmerges

N/A

### Round ID

N/A

### Description of the bug

With `UNIT_TESTS` defined, the server doesn't start immediately after subsystem initialization finishes, resulting in unit test runs taking 3 minutes longer than they need to.

*Random example taken from one of [#5678's workflows](https://github.com/cmss13-devs/cmss13/actions/runs/7813088107/job/21311517380?pr=5678#step:7:62):*
image

Based on some local testing, this is being caused by `/world/proc/HandleTestRun()` calling `SSticker.request_start()` which checks the status of the Nightmare subsystem, tells it to start itself up asynchronously, and then returns without actually starting the game.
(As far as I can tell, `INVOKE_ASYNC()` can't actually return anything)
https://github.com/cmss13-devs/cmss13/blob/6e5fb10de38545ec01822a110f768fc7e54e562e/code/controllers/subsystem/ticker.dm#L116-L128

The server should start, wait for subsystem initialisation to finish, then skip the lobby timer and start running any unit tests.

1. In `code\_compile_options.dm`, uncomment the `UNIT_TESTS` define on L39.
2. Build and run the server.
3. Notice that it's waiting the full 3 minutes for the lobby timer to finish.

(For comparison, try the same steps as above on another codebase with unit tests like TG.)

### Issue Bingo

harryob:
closes #5679

### DIFF
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -8,6 +8,10 @@ SUBSYSTEM_DEF(ticker)
 
 	var/current_state = GAME_STATE_STARTUP //State of current round used by process()
 	var/force_ending = FALSE //Round was ended by admin intervention
+
+	/// If TRUE, there is no lobby phase, the game starts immediately.
+	var/start_immediately = FALSE
+
 	var/bypass_checks = FALSE //Bypass mode init checks
 	var/setup_failed = FALSE //If the setup has failed at any point
 	var/setup_started = FALSE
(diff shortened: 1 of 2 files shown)


harryob:
random pr before
image

now
image

Drulikar (approved):

SabreML:
Does this still allow the Nightmare subsystem to finish loading? (`/datum/controller/subsystem/nightmare/proc/prepare_game()` specifically)
As far as I know that's part of the reason it was acting strangely in the first place.

Drulikar:
> Does this still allow the Nightmare subsystem to finish loading? (`/datum/controller/subsystem/nightmare/proc/prepare_game()` specifically) As far as I know that's part of the reason it was acting strangely in the first place.

Yes as far as I can see:
image
image

Using for example these:
image

image

SabreML (approved):

PR #5804, opened 2024-02-24
(Slightly) Less confusing failure message for built-in visors

### LINKED ISSUE #5799: Cannot remove NVG HUD insert from some helmet(s)
### Testmerges

N/A

### Round ID

21019

### Description of the bug

When attempting to screwdriver some helmets to remove the HUD, the message about the slot being reserved for headgear accessories plays. I'm guessing this is something to do with #5573. I tried putting a helmet accessory into the slot, but could not and got the same message.

dreamseeker_poeEwPQg7t

Screwdrivering a helmet should drop the HUD instead of trying to insert the screwdriver as an accessory

1. Screwdriver a helmet to remove your HUD (not sure which ones are effected)
2. You try to insert the screwdriver into the helmet instead of removing the HUD.

### Issue Bingo

SabreML:
# About the pull request

Resolves #5799 by adding a failure message when trying to remove a built-in helmet visor.

Slightly less confusing for players.

<details>

</details>

# Changelog
:cl:
qol: Added a failure message when trying to remove a built-in helmet visor.
/:cl:

### DIFF
--- a/code/modules/clothing/head/helmet.dm
+++ b/code/modules/clothing/head/helmet.dm
@@ -511,7 +511,17 @@ GLOBAL_LIST_INIT(allowed_helmet_items, list(
 			new_action.give_to(user)
 		return
 
-	if(HAS_TRAIT(attacking_item, TRAIT_TOOL_SCREWDRIVER) && length(inserted_visors))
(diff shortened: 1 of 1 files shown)


Drulikar (approved):
There might be a better way, but ultimately because of the need to not have a message when there's 0 built in it wouldn't be less lines. So fine as is.

PR #5820, opened 2024-02-26
Fixes traps triggering when shuttlecrushed

### LINKED ISSUE #5798: Burrower acid triggering through alamo landing on LZ through alamo flooring.
### Testmerges

#5007: 'Project ARES TM Holder (v4)' by realforest2001 at commit 6c41b6abac #5401: 'M540-A Armored Recon Carrier' by Zonespace27 at commit 90d7aba121 #5777: 'Relay Ping Browser' by Drulikar at commit 336ae657fc

### Round ID

_No response_

### Description of the bug

Burrower acid traps on the LZ can down marines inside through the alamo floor when the alamo lands on the LZ

Not this.

Burrower acid traps on the LZ can down marines inside through the alamo floor when the alamo lands on the LZ

### Issue Bingo

Vicacrov:
# About the pull request

Nullspaces traps when they get shuttlecrushed.

Currently, if traps get shuttlecrushed, huggers/acid appear on the dropship, instead of them simply getting deleted.

This might be a godawful way to do it, but this seemed the cleanest solution to me. Fixes #5798 

Shuttlecrushing should properly delete objects below them, it just happens that hugger/acid traps forcibly trigger upon `Destroy()`.

Hugger and acid traps getting shuttlecrushed:

<details>

https://github.com/cmss13-devs/cmss13/assets/49321394/87540656-59b1-4299-aa73-b630b72dc1b1

</details>

# Changelog

:cl:
fix: Hugger and acid traps no longer activate when they get shuttlecrushed.
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/trap.dm
+++ b/code/modules/cm_aliens/structures/trap.dm
@@ -29,7 +29,6 @@
 
 	cause_data = create_cause_data("resin trap", X)
 	set_hive_data(src, hivenumber)
-	RegisterSignal(src, COMSIG_MOVABLE_SHUTTLE_CRUSH, PROC_REF(on_shuttle_crushing))
 	if(hivenumber == XENO_HIVE_NORMAL)
 		RegisterSignal(SSdcs, COMSIG_GLOB_GROUNDSIDE_FORSAKEN_HANDLING, PROC_REF(forsaken_handling))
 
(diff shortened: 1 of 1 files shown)


code/modules/cm_aliens/structures/trap.dm
```diff
@@ -347,6 +348,11 @@
 	if(ismob(A) || isVehicleMultitile(A))
 		HasProximity(A)
 
+/obj/effect/alien/resin/trap/proc/on_shuttle_crushing()
+	SIGNAL_HANDLER
+	UnregisterSignal(src, COMSIG_MOVABLE_SHUTTLE_CRUSH)
+	loc = null
```
harryob, line 354:
```suggestion
	moveToNullspace()
```

harryob (changes requested):
alternatively, (and this would (arguably) be better) just check for the force argument in Destroy(), and if force == TRUE, don't do the side effects. or, even better better, remove the side effects from /Destroy() entirely (as it should never have side effects)

code/modules/cm_aliens/structures/trap.dm
Vicacrov: Moved the force triggering outside of Destroy() and put it into healthcheck(), which should handle qdelling. Destroy() now only handles clearing up references. (I hope??)

Tested it again, this makes traps not trigger upon getting shuttlecrushed, but they trigger when they get damaged.

harryob (approved):
i hate destroy side effects

PR #5842, opened 2024-02-28
Lowers Renegade hive number

ihatethisengine:
# About the pull request
Lowers renegade hive number to 3, so you can make renegade hive with cip3, instead of cip12.

Makes it reasonably possible to create pure renegade hive, which opens more ways to make friendly xenos as a researcher. It's still inferior to greenos because renegades cannot have a queen (nor eggs), but as an option it can be interesting.

<details>
I hope this works

</details>

# Changelog
:cl: ihatethisengine
balance: lowers renegade hive number to 3, so they can be made with cip3.
/:cl:


Segrain:
That is not what renegade hive is.

PR #5442, opened 2024-01-13
Hive collapse when less than 4 xenos is still a marine major.

TheGamerdk:
# About the pull request

Why below 4 xenos? Because 15 xenos without a Queen is still a threat, 3 xenos can easily be killed.

(I don't like chasing lurkers down for 5 minutes after everyone else is dead)

# Explain why it's good for the game

I must have my Marine Major, not this silly Marine Minor. A single lurker being alive isn't a big deal

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
add: A hive collapse (No new Queen) while at 3 xenos or below is now a marine major, instead of a minor.
/:cl:

PR #5847, opened 2024-02-29
Refactors totalXenos, tweaks some tier 0 things, adds a xeno anti-delay warning

Vicacrov:
# About the pull request

<sup>wow, expanding on xeno code surely doesn't turn into huge refactors</sup>

This PR splits the `totalXenos` list into two separate lists. Currently, the `totalXenos` list contains every tier, including facehuggers and lessers (tier 0s). A few calculations in the game have to snowflakely exclude them, while at other places, tier 0s should simply not matter, but they do (e.g. currently, if you are the last living xeno, like a drone, you won't get the warning for it, because a facehugger is AFKing somewhere under a desk).

What this PR does is that when you join as a xeno, you get added to two lists:
- `total_living_xenos`: Every single xenomorph in the hive, including tier 0s.
- `total_living_xenos_advanced`: Every xenomorph excluding tier 0s.

This PR will change a few things:
- **You can no longer join as a tier 0 if there is not a single tier 1-4 xeno.** Currently, you can join as a tier 0 if the hive had nothing but a facehugger hiding somewhere. Now, you need at least one tier 1-4 xeno to be able to do so.
- "[Hive collapse](https://github.com/cmss13-devs/cmss13/pull/5442)" (>10 minutes being queenless) happens when there are less than 4 tier 1-4 xenos. This means that if there are 3 facehuggers and a drone running around, the hive will still collapse, resulting in a marine major.
- Xenos no longer get latejoin larva if there are only tier 0s alive. (this was just silly)
- Queens can no longer hijack if more than half of the tier 1-4 xenos are not onboard. (Previously, it counted tier 0s as well.)
- Queens can no longer select tier 0s to give evolution points to.
- When there is only one tier 1-4 xeno left, they get the message of being the last one, and ARES immediately performs a scan. Again, random facehuggers hiding under tables will not stop marines from ending the round.
- **ADDED:** When a xeno dies and there are only 3 (=hive collapse number) tier 1-4 xenos left, they get notified about it. This is to avoid round delays when there are 2 runners running around, not knowing whether there are 14 xenos left or only them.

<sub>this last thing is what I wanted to add but it ended up being an entire refactor</sub>

If this refactor doesn't actually make the code better, feel free to just close it, I'm not married to the idea.

Code-wise, I believe it cleans up things a lot, including:
- Removes the `counts_for_slots` variable that was solely used for mobs that had `tier = 0`. Just use `tier`.
- Removes a few IS THIS A HUGGER OR A LESSER?? checks, including `get_real_total_xeno_count()` which was a snowflake check just for tier 0s.
- Fixes some calculations (see above), makes the game saner.

Game-wise, it is a common occurrence that once the Queen dies, xenos scatter; then we end up with a facehugger afking under a table, a drone running around aimlessly, and a runner harassing the FOB. None of these three players know the round is basically over. Let them know it.

Cannot really screenshot anything, it is mainly a refactor.

# Changelog

:cl:
refactor: Refactored living xenomorph lists (totalXenos) into two separate ones to iron out some jankiness with facehuggers and lesser drones, leading to the following changes. 
add: Xenomorphs are now notified if there are only three tier 1-4s are left (excluding tier 0s) to avoid further delays.
balance: The last xenomorph is now counted as the last one who is tier 1-4. A random facehugger afking under a table will no longer stop ARES from showing marines where the last xenomorph is hiding.
balance: You can no longer join as a tier 0 xenomorph when there is not a single tier 1-4 alive.
balance: Xenomorphs no longer gain burrowed larva from marine latejoins if there are only tier 0s alive.
balance: Hive collapse now happens when there are less than 4 tier 1-4s. Previously, tier 0s also stopped a hive collapse, leading to a possible marine minor instead of a marine major.
balance: Queens can no longer hijack if more than half of her tier 1-4s are not onboard. This was already a thing, but it counted facehuggers and lessers too, who should not matter in this calculation.
balance: Queens can no longer select tier 0s to give evolution to, in exchange for a larva.
/:cl:

### DIFF
--- a/code/modules/admin/topic/topic_teleports.dm
+++ b/code/modules/admin/topic/topic_teleports.dm
@@ -136,7 +136,7 @@
 					to_chat(owner, SPAN_ALERT("Hive choice error. Aborting."))
 					return
 				var/datum/hive_status/Hive = GLOB.hive_datum[hives[faction]]
-				var/list/targets = Hive.totalXenos
+				var/list/targets = Hive.total_living_xenos
 				for(var/mob/living/carbon/xenomorph/X in targets)
 					var/area/AR = get_area(X)
 					if(X.stat == DEAD || AR.statistic_exempt)
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -865,7 +865,7 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 	var/datum/hive_status/hive
 	for(var/hivenumber in GLOB.hive_datum)
 		hive = GLOB.hive_datum[hivenumber]
-		if(hive.total_living_xenos.len > 0)
+		if(length(hive.total_living_xenos) > 0)
 			hives += list("[hive.name]" = hive.hivenumber)
 			last_hive_checked = hive
 
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -865,7 +865,7 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 	var/datum/hive_status/hive
 	for(var/hivenumber in GLOB.hive_datum)
 		hive = GLOB.hive_datum[hivenumber]
-		if(hive.totalXenos.len > 0)
+		if(hive.total_living_xenos.len > 0)
 			hives += list("[hive.name]" = hive.hivenumber)
 			last_hive_checked = hive
 
@@ -893,7 +893,7 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 	set category = "Ghost.View"
 
 	var/datum/hive_status/hive = GLOB.hive_datum[XENO_HIVE_NORMAL]
-	if(!hive || !length(hive.totalXenos))
+	if(!hive || !length(hive.total_living_xenos))
 		to_chat(src, SPAN_ALERT("There seems to be no living normal hive at the moment"))
 		return
 
--- a/code/modules/mob/living/carbon/xenomorph/death.dm
+++ b/code/modules/mob/living/carbon/xenomorph/death.dm
@@ -119,8 +119,8 @@
 		if(SSticker.mode && SSticker.current_state != GAME_STATE_FINISHED)
 			if((GLOB.last_ares_callout + 2 MINUTES) > world.time)
 				return
-			if(hive.hivenumber == XENO_HIVE_NORMAL && (LAZYLEN(hive.totalXenos) == 1))
-				var/mob/living/carbon/xenomorph/X = LAZYACCESS(hive.totalXenos, 1)
+			if(hive.hivenumber == XENO_HIVE_NORMAL && (LAZYLEN(hive.total_living_xenos_advanced) == 1))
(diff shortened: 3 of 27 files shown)


Vicacrov:
> Xenos no longer get latejoin larva if there are only tier 0s alive. (this was just silly)

This will mess with early game latejoins, I have to iron this out. The rest should be ready to review

code/modules/admin/topic/topic_teleports.dm
```diff
@@ -136,7 +136,7 @@
 					to_chat(owner, SPAN_ALERT("Hive choice error. Aborting."))
 					return
 				var/datum/hive_status/Hive = GLOB.hive_datum[hives[faction]]
-				var/list/targets = Hive.totalXenos
+				var/list/targets = Hive.total_living_xenos
```
harryob, line 139: completely unrelated to your refactor really but this needs a .Copy() at the end, as otherwise this ends up mutating the actual list

code/modules/mob/dead/observer/observer.dm
```diff
@@ -865,7 +865,7 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 	var/datum/hive_status/hive
 	for(var/hivenumber in GLOB.hive_datum)
 		hive = GLOB.hive_datum[hivenumber]
-		if(hive.totalXenos.len > 0)
+		if(length(hive.total_living_xenos) > 0)
```
harryob, line 868:
```suggestion
		if(length(hive.total_living_xenos))
```

code/modules/mob/living/carbon/xenomorph/death.dm
```diff
@@ -116,11 +116,15 @@
...
+			// Notify the hive that we are about to perish
+			if(LAZYLEN(hive.total_living_xenos_advanced == XENO_HIVE_COLLAPSE_THRESHOLD))
+				xeno_message(SPAN_XENOANNOUNCE("We can feel our hive's power weakening - only a few of us are left."))
+				return
 			if((GLOB.last_ares_callout + 2 MINUTES) > world.time)
 				return
-			if(hive.hivenumber == XENO_HIVE_NORMAL && (LAZYLEN(hive.totalXenos) == 1))
-				var/mob/living/carbon/xenomorph/X = LAZYACCESS(hive.totalXenos, 1)
+			if(LAZYLEN(hive.total_living_xenos_advanced) == 1)
+				var/mob/living/carbon/xenomorph/X = LAZYACCESS(hive.total_living_xenos_advanced, 1)
```
harryob, line 127: this isn't generally treated as a lazy list, why here?

code/modules/mob/living/carbon/xenomorph/hive_status.dm
```diff
@@ -762,11 +765,11 @@
 		spawning_area = hive_location
 	else if(living_xeno_queen)
 		spawning_area = living_xeno_queen
-	else for(var/mob/living/carbon/xenomorpheus as anything in totalXenos)
+	else for(var/mob/living/carbon/xenomorpheus as anything in total_living_xenos)
```
harryob, line 768:
```suggestion
	else 
		for(var/mob/living/carbon/xenomorpheus as anything in total_living_xenos)
```

harryob (changes requested):

PR #5401, opened 2024-01-07
M540-B Armored Recon Carrier

Zonespace27:
# About the pull request
Design doc here: https://hackmd.io/@cmss13/arc

Adds the M540-B Armored Recon Carrier to the game.

image

The M540 is meant to act as a mobile command & intelligence post for deploying CIC officers, with all the equipment they'd need (see design doc for list). It can only be driven by SOs, XOs, COs, and synthetics.

image

The M540 can be obtained for 5 intel points at tier 1. Numbers subject to change.

The primary features of the ARC are the RE700 rotary cannon sentry and the U-56 radar antenna. The U-56 takes 8 seconds to retract or deploy. When deployed, the ARC is anchored in place but has a radar function akin to the command APC. Since the ARC is now a sitting duck, the RE700 rotary cannon is activated. Any non-USCM personnel or life forms that come within 5 tiles of the ARC will be fired upon automatically.

image

# Explain why it's good for the game
Deploying command personnel can be hard-pressed to stay alive or relevant at times (without just becoming a pointman [COs] or medic+ [SOs]). The ARC gives groundside command personnel the ability to coordinate and lead from safety, in addition to giving IOs a potential dropoff point to upload intel and decrypt disks.

# Todo

- [x] Get side-facing movement sprites for the ARC
- [ ] Get a techweb sprite for the ARC

# Changelog
:cl: Zonespace, d.1.n.a. (original sprites), esselnek (new sprites)
add: Added the M540-B Armored Recon Carrier as a t1 intel purchase.
/:cl:

PR #5851, opened 2024-02-29
Vehicle bay elevator stuff

SabreML:
# About the pull request

Makes the Vehicle Bay elevator work better.
Specifically:
- Adds some ASRS sound effects, and tweaks the timing of various elements to line up with them.
- Adds the pre-existing vehicle bay elevator shaft sprite for when the elevator has been lowered. (No animation yet though)
- Adds `\proper` to `/turf/open/floor/almayer/empty` and the elevator shaft effect to change the grammar of its examine text:
**Before/After:**
before after
- Fixes(?) the front guard rails not raising when the elevator is lowered.
- Fixes a bug where the shipside docking port would fall into the elevator shaft after the elevator left:
image
- Fixes an oversight where a human falling into the vehicle bay elevator shaft would be teleported to the requisitions elevator.
- Fixes the gear things at the corners of the elevator spinning constantly until it's called back up.

(See comparison videos below)

The vehicle bay elevator doesn't get used often, but it *does* get used. This will hopefully make it feel a bit less janky.

<details>

### Sending the elevator down and up again:
(Unmute for sound on both videos)

**Before:**

https://github.com/cmss13-devs/cmss13/assets/57483089/5fba87cd-2ef5-42ea-81f7-dc9872e4b622

**After:**

https://github.com/cmss13-devs/cmss13/assets/57483089/3f833e89-3e52-4fee-9f07-2c376ab2000b

</details>

(I've also tested this with #5401, and it seems to work without any issues.)

# Changelog
:cl:
add: Made the vehicle bay elevator look and sound better when in use.
fix: Fixed the vehicle bay elevator's front railings not closing when the elevator was lowered.
fix: Fixed the vehicle bay elevator's docking port falling down the elevator shaft when it was lowered.
fix: Fixed falling into the vehicle bay elevator shaft teleporting players to the requisitions elevator.
/:cl:

### DIFF
--- a/maps/map_files/USS_Almayer/USS_Almayer.dmm
+++ b/maps/map_files/USS_Almayer/USS_Almayer.dmm
@@ -12215,7 +12215,7 @@
 /turf/open/floor/almayer,
 /area/almayer/squads/req)
 "bpR" = (
-/turf/open/floor/almayer/empty,
+/turf/open/floor/almayer/empty/requisitions,
 /area/supply/station)
 "bpS" = (
 /obj/structure/machinery/door/poddoor/railing{
@@ -12587,7 +12587,7 @@
 /area/almayer/squads/req)
 "bsK" = (
 /obj/effect/landmark/supply_elevator,
-/turf/open/floor/almayer/empty,
+/turf/open/floor/almayer/empty/requisitions,
 /area/supply/station)
 "bsL" = (
 /obj/structure/machinery/door/poddoor/railing{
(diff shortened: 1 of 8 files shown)


maps/map_files/USS_Almayer/USS_Almayer.dmm
Nanu308, line 1: mapping lgtm

Nanu308 (approved):

Drulikar (changes requested):
Everything here appears to be working; though I think we should also fix a couple other aspects while we're at it:

`/obj/structure/machinery/door_control/railings/use_button` early returns on `if(!is_mainship_level(SSshuttle.vehicle_elevator.z))` even if the railing is down (desiredstate is 1). We should always allow railing to go up.

~~`SSshuttle.vehicle_elevator.timeLeft()` should be changed to `shuttle.has_arrive_time()` so its isn't just always "Moving" (looking for the actual reference to the shuttle atm)

Drulikar (approved):

PR #5856, opened 2024-03-01
Adds a before and after loc comparsion to picking up dropship equipment.

### LINKED ISSUE #5848: Two Powerloaders can pick up the same object.Breaks an attachment point when placed
### Testmerges

#5007,#5401,#5774,#5763,#5829,#5845

### Round ID

21091

### Description of the bug

I picked up a sentry launch point with a power loader right after another player had picked it up. I placed it on the north attachment point,right after the other placed it in the middle making the middle one unusable.

Should not be able to pick up a other sentry launch point nor be able to place it after some one else picked it up

1.Print any drop ship attachment 
2.Get Two players with power loaders
3.First player picks up the attachment with a power loader
4.Second player picks up the attachment while the first player is picking it up
5.first player places the attachment on a point
6.second player places the attachment on a point breaking the first on.

### Issue Bingo

TheGamerdk:
# About the pull request

Should fix #5848

<details>

</details>

# Changelog

:cl:
fix: Should no longer be possible to pick up the same dropship equiment with 2 different powerloaders
/:cl:

### DIFF
--- a/code/modules/cm_marines/dropship_equipment.dm
+++ b/code/modules/cm_marines/dropship_equipment.dm
@@ -112,12 +112,15 @@
 	playsound(loc, 'sound/machines/hydraulics_2.ogg', 40, 1)
 	var/duration_time = 10
 	var/point_loc
+	var/old_loc = src.loc
 	if(ship_base)
 		duration_time = 70 //uninstalling equipment takes more time
 		point_loc = ship_base.loc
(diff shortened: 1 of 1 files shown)


SabreML (commented):
Could you also add a similar check to `/obj/vehicle/multitile/proc/uninstall_hardpoint()`?
That seems to be the only other 'picking something up with a powerloader' proc that could end up having the same problem.

PR #5865, opened 2024-03-02
fixes pod cryo

### LINKED ISSUE #5863: Hypersleep/Cryopods on Escape Pods not working.
### Testmerges

#5007, #5401, #5774, #5829, #5845, #5802

### Round ID

Latest was 21119, but has occured prior.

### Description of the bug

Cryopod/Hypersleep beds on escape pods do not work and when inspected say 'It does not appear to be working.', thusly you cannot enter them to enter hypersleep and leave the game in an RP-friendly way.

Be able to enter hypersleep once away safely away from the Almayer after using an escape pod.

1. Evacuate using an escape pod (For my cases, it was the Corporate Liasons' pod, though I am quite sure it is an issue with all pods, if memory serves me right)
2. Attempt to enter hypersleep bed, it does not respond.

### Issue Bingo

harryob:
fixes #5863

:cl:
fix: you can now enter cryo on pods
fix: cryo pods no longer tell you that they're broken
/:cl:

### DIFF
--- a/code/__DEFINES/objects.dm
+++ b/code/__DEFINES/objects.dm
@@ -43,7 +43,6 @@
 #define EMPED 256 // temporary broken by EMP pulse
 #define TIPPED_OVER 512 // for tipped over vendors
 #define IN_USE 1024 // to show that it's actively used right now. Vending an object, for example.
-#define MACHINE_DO_NOT_PROCESS 32768 //Do not added these to processing queue.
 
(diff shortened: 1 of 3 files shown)


SabreML:
I don't seem to be able to reproduce the issue locally. I assume I'm missing a step somewhere?
I can see the "It does not appear to be working" message on examine, but climbing into the pod seems to work normally whether it's docked, crash-landing, or flying in space.

SabreML (commented):
LGTM other than this for the `COMSIG_PARENT_QDELETING` handler.

fira (changes requested):
Needs destroy cleanup too of the list if the dock port is deleted
LGTM otherwise

PR #5874, opened 2024-03-03
Adds Professor DUMMY to the CMO office, accessible by the CMO and the SEA

Vicacrov:
# About the pull request

Adds a new wall cabinet to the CMO's office called the "professor dummy cabinet". It contains Professor DUMMY and its tablet: it is a tool for new nurses(!!) and doctors to learn medical treatment, including surgery.

The locker can only be opened by the CMO or the SEA (not even the CO can open it). It cannot be flashbanged, melted, or slashed.

There is a hefty warning for everyone when the dummy falls out, in case someone destroys the closet in a creative way:

image

The dummy and its tablet dust upon hijack impact, to avoid either side having a godmoded mob.

It is a tool for new medical personnel to practice on. It can also ease boredom if the round is going too smoothly for marines.

Setting up the tablet and hijacking. (The dummy no longer spawns in the Morgue but in the CMO office).

<details>

https://github.com/cmss13-devs/cmss13/assets/49321394/563db21d-ebc2-41e0-a8e8-764f9b8aa61b

</details>

# Changelog

:cl:
add: Added a professor dummy cabinet to the CMO office, containing Professor DUMMY and its control tablet. It is a tool to teach new medical personnel with. The cabinet is accessible by the CMO and the SEA only.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/medical.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/medical.dm
@@ -218,7 +218,7 @@
 
 /obj/structure/closet/secure_closet/professor_dummy/emp_act(severity)
 	for(var/mob/living/carbon/human/dummy/professor_dummy/dummy in src)
-		visible_message(SPAN_DANGER("Something in the cabinet blows apart!"))
+		visible_message(SPAN_DANGER("Something in [src] blows apart!"))
 		playsound(src, 'sound/effects/metal_crash.ogg', 25, 1)
 		qdel(dummy)
 	..()
--- a/code/game/objects/structures/crates_lockers/closets/secure/medical.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/medical.dm
@@ -180,6 +180,7 @@
 	icon_locked = "surgical_wall_locked"
 	icon_opened = "surgical_wall_open"
 	icon_broken = "surgical_wall_spark"
+	health = null	// Unbreakable
 	unacidable = TRUE
 	unslashable = TRUE
 	store_mobs = TRUE
@@ -201,13 +202,23 @@
 		visible_message(SPAN_HIGHDANGER("Professor DUMMY should only be used for teaching medical personnel, exclusively done by the [JOB_CMO] or the [JOB_SEA]. Do not abuse it."))
 	..()
 
-/obj/structure/closet/secure_closet/professor_dummy/toggle(mob/living/user)
-	if(opened)
-		for(var/mob/M in loc)
-			if(!istype(M, /mob/living/carbon/human/dummy/professor_dummy))
-				visible_message(SPAN_WARNING("[src] won't budge!"))
-				return
+/obj/structure/closet/secure_closet/professor_dummy/close()
+	for(var/mob/M in loc)
+		if(!istype(M, /mob/living/carbon/human/dummy/professor_dummy))
+			visible_message(SPAN_WARNING("[src] won't budge!"))
+			return
 	..()
+
+	// Force locking upon closing it
+	locked = TRUE
+	update_icon()
 
 /obj/structure/closet/secure_closet/professor_dummy/flashbang(datum/source, obj/item/explosive/grenade/flashbang/FB)
 	return
+
+/obj/structure/closet/secure_closet/professor_dummy/emp_act(severity)
+	for(var/mob/living/carbon/human/dummy/professor_dummy/dummy in src)
+		visible_message(SPAN_DANGER("Something in the cabinet blows apart!"))
+		playsound(src, 'sound/effects/metal_crash.ogg', 25, 1)
+		qdel(dummy)
+	..()
--- a/code/game/objects/structures/crates_lockers/closets/secure/medical.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/medical.dm
@@ -187,7 +187,7 @@
 	wall_mounted = TRUE
 
 /obj/structure/closet/secure_closet/professor_dummy/Initialize()
-    . = ..()
+	. = ..()
 	var/mob/living/carbon/human/dummy/professor_dummy/spawned_dummy = new(src)
 	var/datum/equipment_preset/dummy_preset = /datum/equipment_preset/other/professor_dummy
 	arm_equipment(spawned_dummy, initial(dummy_preset.name), TRUE, FALSE)
--- a/code/game/objects/structures/crates_lockers/closets/secure/medical.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/medical.dm
@@ -192,19 +192,18 @@
 
 /obj/structure/closet/secure_closet/professor_dummy/togglelock(mob/living/user)
 	if(user.job == JOB_CMO || user.job == JOB_SEA)
-		..()
-		return
+		return ..()
 
 	to_chat(user, SPAN_WARNING("Only the [JOB_CMO] or the [JOB_SEA] can toggle this lock."))
 
 /obj/structure/closet/secure_closet/professor_dummy/dump_contents()
 	if(locate(/mob/living/carbon/human/dummy/professor_dummy) in src)
 		visible_message(SPAN_HIGHDANGER("Professor DUMMY should only be used for teaching medical personnel, exclusively done by the [JOB_CMO] or the [JOB_SEA]. Do not abuse it."))
-	..()
+	return ..()
 
 /obj/structure/closet/secure_closet/professor_dummy/close()
-	for(var/mob/M in loc)
-		if(!istype(M, /mob/living/carbon/human/dummy/professor_dummy))
+	for(var/mob/mob in loc)
+		if(!istype(mob, /mob/living/carbon/human/dummy/professor_dummy))
 			visible_message(SPAN_WARNING("[src] won't budge!"))
 			return
 	..()
@@ -216,9 +215,17 @@
 /obj/structure/closet/secure_closet/professor_dummy/flashbang(datum/source, obj/item/explosive/grenade/flashbang/FB)
 	return
 
-/obj/structure/closet/secure_closet/professor_dummy/emp_act(severity)
-	for(var/mob/living/carbon/human/dummy/professor_dummy/dummy in src)
+/obj/structure/closet/secure_closet/professor_dummy/proc/check_and_destroy_dummy()
+	var/mob/dummy = locate(/mob/living/carbon/human/dummy/professor_dummy) in src
+	if(dummy)
 		visible_message(SPAN_DANGER("Something in [src] blows apart!"))
 		playsound(src, 'sound/effects/metal_crash.ogg', 25, 1)
 		qdel(dummy)
+
+/obj/structure/closet/secure_closet/professor_dummy/emp_act(severity)
+	check_and_destroy_dummy()
+	..()
+
+/obj/structure/closet/secure_closet/professor_dummy/ex_act(severity)
+	check_and_destroy_dummy()
 	..()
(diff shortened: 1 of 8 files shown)


ihatethisengine:
maybe just make it uninteractable on hijack?

Vicacrov:
> maybe just make it uninteractable on hijack?

Dusting might be a bit jarring, but having a human-looking mob nobody can interact with would be more disruptive, I think.

Drulikar (changes requested):
This needs to instead be in the SEA vendor where its unhackable and otherwise then falls under the responsibility of the SEA that lent it to someone to use to put it away.

If you also want CMOs to access it, then they'd need a similar situation where only they (or a SEA) can vend it.

As of now, its the responsibility of the admin that spawned it to ensure it doesn't get misused.

Vicacrov:
Implemented changes:
- Made a new subtype for this dummy as I realised the character setup one was the same.
- Reverted the Morgue changes.
- The dummy now dusts upon hijack impact, not upon hijack start (code bloat on my part).
- Added a special closet to the CMO office, the "professor dummy cabinet".
  - It can be only opened with a CMO or SEA **job**, not access. (Not even a CO ID helps here.)
  - It cannot be flashbanged, slashed, or melted.
  - When the dummy falls out, everyone around the closet gets a warning how it should not be misused - this also covers the edge case where it gets destroyed by unconventional means. Ignorance is not an excuse when you get a red text this big.
  - The closet can store mobs but it fails to close if there is any other mob on its tile but the dummy. (No hiding in the indestructible closet.)

image

image

maps/map_files/USS_Almayer/USS_Almayer.dmm
Nanu308, line 1: mapping lgtm

Nanu308 (approved):

code/modules/mob/living/carbon/human/human_dummy.dm
```diff
@@ -74,11 +74,42 @@ GLOBAL_LIST_EMPTY(dummy_mob_list)
...
+	// Personalise it
+	change_real_name(src, "Professor DUMMY the Medical Mannequin")
+	age = rand(1,5)
+	gender = pick(MALE, FEMALE)
+	var/datum/preferences/A = new
+	A.randomize_appearance(src)
+	regenerate_icons()
+
+	// Ensure nurses can practice surgery on it
+	mob_flags |= EASY_SURGERY
```
Drulikar, line 98: You are missing `huggable = FALSE` like `/datum/equipment_preset/other/professor_dummy/load_race` used to do.

code/game/objects/structures/crates_lockers/closets/secure/medical.dm
```diff
@@ -171,3 +171,54 @@
...
+	..()
+
+	// Force locking upon closing it
+	locked = TRUE
+	update_icon()
+
+/obj/structure/closet/secure_closet/professor_dummy/flashbang(datum/source, obj/item/explosive/grenade/flashbang/FB)
+	return
+
+/obj/structure/closet/secure_closet/professor_dummy/emp_act(severity)
```
Drulikar, line 219: ex_act also needs to be implemented the same way. Throw a HE grenade near it and it'll dump the contents out.

code/modules/gear_presets/other.dm
```diff
@@ -845,38 +845,6 @@
 
 //*****************************************************************************************************/
 
-/datum/equipment_preset/other/professor_dummy
```
Drulikar, line 848: Removing this means it is no longer in the create humans menu; which is generally how its been spawned. It'd probably be better to just keep the preset but have your closest spawn the equipment preset into the closet.

code/game/objects/structures/crates_lockers/closets/secure/medical.dm
```diff
@@ -171,3 +171,54 @@
...
+	to_chat(user, SPAN_WARNING("Only the [JOB_CMO] or the [JOB_SEA] can toggle this lock."))
+
+/obj/structure/closet/secure_closet/professor_dummy/dump_contents()
+	if(locate(/mob/living/carbon/human/dummy/professor_dummy) in src)
+		visible_message(SPAN_HIGHDANGER("Professor DUMMY should only be used for teaching medical personnel, exclusively done by the [JOB_CMO] or the [JOB_SEA]. Do not abuse it."))
+	..()
+
+/obj/structure/closet/secure_closet/professor_dummy/close()
+	for(var/mob/M in loc)
+		if(!istype(M, /mob/living/carbon/human/dummy/professor_dummy))
```
Drulikar, line 207:
```suggestion
	for(var/mob/current_mob in loc)
		if(!istype(current_mob, /mob/living/carbon/human/dummy/professor_dummy))
```
No single letter vars please: https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#dont-use-abbreviations

code/game/objects/structures/crates_lockers/closets/secure/medical.dm
```diff
@@ -171,3 +171,54 @@
...
+	if(user.job == JOB_CMO || user.job == JOB_SEA)
+		..()
+		return
+
+	to_chat(user, SPAN_WARNING("Only the [JOB_CMO] or the [JOB_SEA] can toggle this lock."))
+
+/obj/structure/closet/secure_closet/professor_dummy/dump_contents()
+	if(locate(/mob/living/carbon/human/dummy/professor_dummy) in src)
+		visible_message(SPAN_HIGHDANGER("Professor DUMMY should only be used for teaching medical personnel, exclusively done by the [JOB_CMO] or the [JOB_SEA]. Do not abuse it."))
+	..()
```
Drulikar, line 203:
```suggestion
	return ..()
```
The other implementations are returning TRUE/FALSE/1/0

code/game/objects/structures/crates_lockers/closets/secure/medical.dm
```diff
@@ -171,3 +171,54 @@
...
+	wall_mounted = TRUE
+
+/obj/structure/closet/secure_closet/professor_dummy/Initialize()
+	. = ..()
+	new /mob/living/carbon/human/dummy/professor_dummy(src)
+
+/obj/structure/closet/secure_closet/professor_dummy/togglelock(mob/living/user)
+	if(user.job == JOB_CMO || user.job == JOB_SEA)
+		..()
+		return
```
Drulikar, line 196:
```suggestion
		return ..()
```
Although nothing normally is returned, since you're returning anyways you may as well.

code/modules/mob/living/carbon/human/human_dummy.dm
```diff
@@ -74,11 +74,42 @@ GLOBAL_LIST_EMPTY(dummy_mob_list)
...
+
+	// Spawn its control tablet
+	var/obj/item/device/professor_dummy_tablet/tablet = new (loc)
+	tablet.link_dummy(src)
+
+	if(is_mainship_level(z))
```
Drulikar, line 104: A closet's loc is not considered is_mainship_level you need to get_turf

Drulikar (changes requested):

code/game/objects/structures/crates_lockers/closets/secure/medical.dm
```diff
@@ -171,3 +171,54 @@
...
+
+	// Force locking upon closing it
+	locked = TRUE
+	update_icon()
+
+/obj/structure/closet/secure_closet/professor_dummy/flashbang(datum/source, obj/item/explosive/grenade/flashbang/FB)
+	return
+
+/obj/structure/closet/secure_closet/professor_dummy/emp_act(severity)
+	for(var/mob/living/carbon/human/dummy/professor_dummy/dummy in src)
```
Drulikar, line 220: If there only will ever be one dummy we can instead do something like
```js
var/mob/dummy = locate(/mob/living/carbon/human/dummy/professor_dummy) in src
if(dummy)
	qdel(dummy)
```

code/modules/gear_presets/other.dm
Drulikar: See https://github.com/cmss13-devs/cmss13/blob/master/code/modules/admin/topic/topic_events.dm#L151-L159 how create_humans uses these presets. You likely just need to create the mob, and then `arm_equipment` it.

Drulikar: If the concern was just not registering the signal for admin spawned; then have the closet register the signal for that spawn on the dummy's behalf (as in we aren't registering a signal on the closet, but the closet's initialize call is having the dummy register the signal)

Drulikar: Relayed on discord, but putting it here too. Something like
```js
/obj/structure/closet/secure_closet/professor_dummy/Initialize()
    . = ..()
    var/mob/living/carbon/human/dummy/professor_dummy/spawned_dummy = new(src)
    var/datum/equipment_preset/dummy_preset = /datum/equipment_preset/other/professor_dummy
    arm_equipment(spawned_dummy, initial(dummy_preset.name), TRUE, FALSE)
    spawned_dummy.RegisterSignal(SSdcs, COMSIG_GLOB_HIJACK_LANDED, TYPE_PROC_REF(/mob/living/carbon/human/dummy/professor_dummy, destroy_upon_hijack))```
should work. The signal registration here is just if we want only closet dummies to dust; but I don't mind admin spawns having the same signal either if they're on a game level z.

File: code/game/objects/structures/crates_lockers/closets/secure/medical.dm [hunk 9]
```diff
@@ -171,3 +171,64 @@
...
+	unslashable = TRUE
+	store_mobs = TRUE
+	wall_mounted = TRUE
+
+/obj/structure/closet/secure_closet/professor_dummy/Initialize()
+	. = ..()
+	var/mob/living/carbon/human/dummy/professor_dummy/spawned_dummy = new(src)
+	var/datum/equipment_preset/dummy_preset = /datum/equipment_preset/other/professor_dummy
+	arm_equipment(spawned_dummy, initial(dummy_preset.name), TRUE, FALSE)
+	spawned_dummy.RegisterSignal(SSdcs, COMSIG_GLOB_HIJACK_LANDED, TYPE_PROC_REF(/mob/living/carbon/human/dummy/professor_dummy, destroy_upon_hijack))
```
Drulikar, line 194:
```suggestion
```

code/modules/mob/living/carbon/human/human_dummy.dm
```diff
@@ -74,11 +74,21 @@ GLOBAL_LIST_EMPTY(dummy_mob_list)
...
 /mob/living/carbon/human/dummy/tutorial // Effectively an even more disabled dummy
 
 /mob/living/carbon/human/dummy/tutorial/Initialize(mapload)
 	. = ..()
 	status_flags = GODMODE
 	ADD_TRAIT(src, TRAIT_IMMOBILIZED, TRAIT_SOURCE_TUTORIAL)
 	anchored = TRUE
+
+// Professor Dummy, used by CMOs and SEAs to teach new nurses/doctors
+/mob/living/carbon/human/dummy/professor_dummy/proc/destroy_upon_hijack()
```
Drulikar, line 86:
```suggestion
/mob/living/carbon/human/dummy/professor_dummy/Initialize(mapload)
	. = ..()
	RegisterSignal(SSdcs, COMSIG_GLOB_HIJACK_LANDED, PROC_REF(destroy_upon_hijack))

/mob/living/carbon/human/dummy/professor_dummy/proc/destroy_upon_hijack()
```

Drulikar (changes requested):

Drulikar (approved):

Drulikar:
Feel free to undraft (Ready for review button near the bottom) when you're happy with it.

PR #5875, opened 2024-03-03
Battlefield executions done on the Almayer can be heard by everyone onboard

Vicacrov:
# About the pull request

If the CO battlefield executes someone on the Almayer (and does it with a gun that has no silencer on it), the shot will be audible for everyone onboard.

If the weapon is silenced or the CO is not on the Almayer, the execution's audio remains the same.

In case of a mass execution, there is a 5 seconds timer, meaning if the CO executes three people in 5 seconds, only one shot will be audible across the Almayer, the other two shots will be local.

Blaxxon Blonz brought this up in an [ideasguys thread](https://forum.cm-ss13.com/t/bes-should-be-audible-throughout-the-almayer/6811). I found it hilarious, so here it is.

His initial post aligns with my own opinion, please read it.

It would be cool to add some echo effect to it, but I do not think we are capable of doing so with this proc (since the audio is always based on the bullet being used).

1. Join a server with three ckeys, one CO, one rifleman, one doctor
2. Mute the CO and rifleman clients
3. Move the doctor to the morgue
4. Execute the rifleman in the cryostorage room
5. Listen to the audio

# Changelog

:cl:
add: Battlefield executions performed on the Almayer can be now heard by everyone onboard, unless the gun is silenced. There is a five-seconds long cooldown between ship-wide noises, pace your executions if you want everything to be heard.
/:cl:

### DIFF
--- a/code/datums/ammo/bullet/bullet.dm
+++ b/code/datums/ammo/bullet/bullet.dm
@@ -55,8 +55,9 @@
 
 	if(fired_from.flags_gun_features & GUN_SILENCED)
 		playsound(user, fired_from.fire_sound, 25, FALSE)
-	else if(is_mainship_level(user.z))
+	else if(is_mainship_level(user.z) && !TIMER_COOLDOWN_CHECK(user, COOLDOWN_SHIPWIDE_EXECUTION_ALERT))
 		playsound_z(SSmapping.levels_by_any_trait(list(ZTRAIT_MARINE_MAIN_SHIP)), fired_from.fire_sound, volume = fired_from.firesound_volume)
+		TIMER_COOLDOWN_START(user, COOLDOWN_SHIPWIDE_EXECUTION_ALERT, 5 SECONDS)
 	else
 		playsound(user, fired_from.fire_sound, fired_from.firesound_volume, FALSE)
 
--- a/code/datums/ammo/bullet/bullet.dm
+++ b/code/datums/ammo/bullet/bullet.dm
@@ -53,10 +53,12 @@
 		fired_from.delete_bullet(firing_projectile, TRUE)
 		return
 
-	if(!(fired_from.flags_gun_features & GUN_SILENCED))
-		playsound(user, fired_from.fire_sound, fired_from.firesound_volume, FALSE)
-	else
+	if(fired_from.flags_gun_features & GUN_SILENCED)
 		playsound(user, fired_from.fire_sound, 25, FALSE)
+	else if(is_mainship_level(user.z))
+		playsound_z(SSmapping.levels_by_any_trait(list(ZTRAIT_MARINE_MAIN_SHIP)), fired_from.fire_sound, volume = fired_from.firesound_volume)
+	else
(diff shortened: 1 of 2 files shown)


Vicacrov:
LTNTS suggested putting a 5s cooldown on the z-level playsound just in case of a mass execution, switching it to draft until I code that

Vicacrov:
It is ready to review, but the z-level wide sound could use some echo. If anyone can figure out how echo works (because [its documentation](https://secure.byond.com/docs/ref/index.html#/sound/var/echo) is way too confusing for my smooth brain), ping me on Discord.

Birdtalon:
Discussing this internally. Will be a decision soon.

code/datums/ammo/bullet/bullet.dm
```diff
@@ -53,10 +53,13 @@
 		fired_from.delete_bullet(firing_projectile, TRUE)
 		return
 
-	if(!(fired_from.flags_gun_features & GUN_SILENCED))
-		playsound(user, fired_from.fire_sound, fired_from.firesound_volume, FALSE)
-	else
+	if(fired_from.flags_gun_features & GUN_SILENCED)
 		playsound(user, fired_from.fire_sound, 25, FALSE)
+	else if(is_mainship_level(user.z) && !TIMER_COOLDOWN_CHECK(user, COOLDOWN_SHIPWIDE_EXECUTION_ALERT))
+		playsound_z(SSmapping.levels_by_any_trait(list(ZTRAIT_MARINE_MAIN_SHIP)), fired_from.fire_sound, volume = fired_from.firesound_volume)
```
Birdtalon, line 59: Do you reckon that you could play this as a normal sound but just with a large range and falloff so people can hear it across the other side of the Almayer but much quieter than next to it?

Birdtalon (changes requested):

code/datums/ammo/bullet/bullet.dm
Vicacrov: I have no way to measure it, but if we wanted to make it shipwide, if someone gets executed in CIC (x 100, y 150) someone in the lower deck's reactor room (x 260 y 50) is incredibly far from it. Would it not be way more expensive to send an audio file that far vs using playsound_z?

Birdtalon: I don't know how it would compare resource wise, but for a thing that happens once in a blue moon it doesn't really matter. You could proccall playsound on a turf and then ghost to various places on the almayer then call it to judge the distance and falloff.

Vicacrov: let me know if the idea itself is approved, if so, I'll start messing around with the sound range and falloff

Birdtalon:
Sadly after discussion internally we have decided not to proceed with this PR for the following reasons expressed by a couple of team members:

- Concern that it unnecessarily glorifies the process of Battlefield Execution - plus the CO is already supposed to announce it to all.
- Concern that it may incentivise its use due to having a funny effect.

PR #5838, opened 2024-02-27
Resprites - 2/2024 Edition

nauticall:
# About the pull request

Resprites various items in the game to be subjectively prettier and more up-to-date with modern standards. Among them are the following:
- First aid kits.
- Coffins.
- Material stacks.
- Mortar shells.
- Beakers, bottles and pill bottles (sprites ported and hue-shifted from Baystation/Aurora).

# Explain why it's good for the game
# Sprites
<details>
<summary>Sprites</summary>

image
image
image
image
image

</details>

# Changelog
:cl: nauticall
imageadd: Resprites first aid kits, coffins, material stacks, mortar shells, and chemistry items. Chemistry sprites ported from Baystation / Aurora.
fix: Recolors several reagent sprites to be more in-line with their established colors, such as on autoinjectors.
fix: Fixed invisible cloth tables.
fix: Fixed wrong shading on the large floodlight.
/:cl:

PR #5887, opened 2024-03-06
Pizza Station is back on the menu

Nanu308:
# About the pull request
Adds a Pizza Galaxy space diner to the game for the Pizza ERT. They deserve their own station as well!

The clothed tables will spawn incorrectly spritewise until Naut's PR #5838 is added to the game.

Once new cooking sprites are added I'll go back and tweak the kitchen area some more.

More Soulful lore and RP spots.

<details>

bilde

</details>

# Changelog

:cl:
mapadd: Added the Galaxy Pizza Space Diner ERT station, for all your outer expanse food needs!
maptweak: Placed APC's on ert stations and minor tweaks+fixes.
/:cl:


Vicacrov:
very cute station, 10/10 would visit for a pizza

Dunno if you can consider any of these issues, but here are my observations:

1. Southwest, hangar area, double closets on one tile

image

If you open one of these closets, you cannot close it, because there is another dense structure (the second closet) on the tile.

2. Southwest, hangar area, non-symmetrical offset

image

This briefcase overlaid with the vending machine looks weird because 1) the briefcase is a pixel narrower than the vending machine 2) if you pick up the briefcase, it leaves behind hot coffee machine with a 12px y-offset, while there is a cig vendor a bit to the right, in a similar corner, with 0px offset. It becomes non-symmetrical.

3. Kitchen area

The red handrails are called "sandstone handrail"s, not sure if that's what you are going for.

4. Kitchen area

image

The empty jar object is a nightmare to click on (the transparent pixels are non-clickable), it is pretty much impossible to pick it up unless you right-click it. You might want to replace it with a beaker that can be actually used.

5. East of kitchen area

image

This cig vendor, exclusively among the other vending machines, has a 3px y-offset. 😢 

6. Kitchen and cleaning

image

This fire safety closet has an x-offset, even though it is a singular closet. It would look much better (and consistent) without the offset.

7. STAFF ONLY bolted airlock

image

I do not see the point of this airlock being bolted (red) when none of the other airlocks (blue) have an ID access. Maybe give them a civilian logistics (`ACCESS_CIVILIAN_LOGISTICS`) access?

Bonus funny:

If you middle-click the handrails because you are an idiot like me who cannot aim at the boxes properly, you vault over the table and the handrail in a split second.

<details>
<summary>intended handrails mechanics</summary>

https://github.com/cmss13-devs/cmss13/assets/49321394/f18eeff4-6b87-41f8-ab87-200f7b001527

</details>

Drulikar:
CI requires fixes from #5891

Vicacrov:
> fixed le contrib review

Looks good!! It'd be probably better to make a new subtype of the sandstone railing though, rename it to "red railing", and map that in, instead of renaming every single instance on the map. It makes it more reusable (admins can spawn it in during the game), plus I think we prefer things to have its own subtype as opposed to having it varedited on the map (except for pixel-shifts).

This might be an old design rule I remember, though, because I cannot find it in the contributor guidelines. :waaa:

Drulikar (changes requested):
Layering for some of the handrails is incorrect. Also the barriers between the dock and the station probably should be named something else at least.

image

Nanu308:
> Layering for some of the handrails is incorrect. Also the barriers between the dock and the station probably should be named something else at least.
> 
> image

layers should be fixed, also named docking umbilical correctly

Drulikar (commented):
Everything else looks fine.

Minor: Layering of cig machine and stack of chairs could be elevated some so the handrails north of them aren't above them:
image
****

Drulikar (changes requested):
Mentioned on discord, but just putting it here too:

Much like the cig machine, the play has the same issue, so you should make it the same layer as the player so their layering depends on whether they are south or north of the barricade. The + 0.01 is because it tweaks the layer slightly depending on facing, so the adjustment makes it the same. It also appears that this makes the var edits for the other barricades unnecessary.

So in summary:
- Use layer = MOB_LAYER + 0.01 under /obj/structure/barricade/handrail/pizza in code
- Remove all usage of map var edited handrail overrides (should be just one prefab kind)

Drulikar (approved):
The var edited barricades still aren't perfect; but its good enough.

PR #3729, opened 2023-06-26
Ravager and Hedgehog changes

GoldenDarkness55:
# About the pull request

Nerfing base ravager by trimming the excess power without changing how the kit plays at its core. Buffing hedge rav and slightly fleshing out the management of shards unique to the strain.

# Explain why it's good for the game

Balance good.

Base rav changes pretty straightforward - maintaining current empower combo, but nerfing excess shield and stun. +1 slash per cd reset to force rav to stay longer to pull it off - should have significantly less leeway against large groups and be easier to punish.
Empower cd buff to compensate for second empower cast nerf, since the cooldown for empower starts counting on first cast and experienced players used the long timer to empower twice in rapid succession/skip a portion of the cooldown - the cooldown is effectively longer now than it was before.
Scissor cut felt rather underused and this buff should help make it more relevant, especially against static marines/sieges now that the shield is weaker.

As for Hedge - speed reduction to this extent without any dashes just doesn't work whatsoever, (for more evidence look at oppressor prae suffering the same issue) so this has to go as a prerequisite for this strain ever being good. Slash damage cut to reduce excessive power and punishment in 1v1 and put more emphasis on this being a tank/support/debuff strain (ideally it would get slash damage or slash speed during shard lockout to reward risky aggressive play and not just an escape tool, but im gonna leave that to someone who can code if someone likes the idea).
Shield duration feels a bit too low currently, but this might end up making it overtuned, unsure. Shard damage raised cause shield isn't threatening enough and marines - myself included - often happily magdump into it even PB to kill the hedge faster.
Shed and slow durations reduced because they're horrendously excessive. Slow on 3 might be too short though, considered 4-5 instead but best to test first.
Shard toss range increased to help with engagements, likely still needed despite the speed buff, rest of abilities has separate ranges.
Shard "economy" changes to put more emphasis into this unique system being a thing - currently it might aswell not exist, the shards regenerate nigh instantly so the ability costs do not matter. These changes should put more importance on managing it and reward marines for good play - for example if they didn't feed the shards back up by shooting during the shield timer, it's going to take the hedge longer to recover its armor, making it take slightly more damage. Shard per slash gain to help out with rare situations where it's doing well yet struggles to recover shards due to marines either not shooting at all or using exclusively low RoF weapons.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
[...]

PR #5890, opened 2024-03-06
buffs ravager empower

iloveloopers:
# About the pull request
empower range increased to 4

#3729 nerfed base rav too much, rav does not have great armor (only 25) which means it has to rely alot on shields gained for sustain. it also increased the max targets from 5 to 6 but didn't give any empower range, its very rare to actually be able to get the full 6 targets.

its a numbers change

# Changelog
:cl:
balance: ravager empower range is now 4 tiles
/:cl:


Drulikar:
CI requires fixes from #5891

Drulikar (changes requested):
Just increase the empower range to 4 and leave the other values as they are.

Drulikar (approved):

PR #5903, opened 2024-03-08
Disables New Varadero from the current map pool. 

joookschad:
# About the pull request

Disables New Varadero from the current map pool. 

Despite numerous fixes and improvements to the map, it is still a fundamentally flawed meme of a map(It's literally half of Old Ice, slightly revamped), it still is the universal "Vote Varadero and leave hehehehe" map.

# Changelog
:cl:

del: Removed New Varadero from the map pool

/:cl:

### DIFF
--- a/map_config/maps.txt
+++ b/map_config/maps.txt
@@ -58,6 +58,7 @@ map lv522_chances_claim
 endmap
 
 map new_varadero
+	disabled
 endmap
 
 map whiskey_outpost_v2


Drulikar:
If you don't want to play it, don't vote for it.

A rework is already in progress.

iloveloopers:
if only the voting system was actually good.
sure do love that the entire pop is forced to play varadero cause 6 people were voting for it the past 5 rounds

PR #5935, opened 2024-03-13
Loadout Point Increase

567Turtle:
# About the pull request

Changes the the loadout point limit from 7 to 8.

Honestly I don't even understand why there's a limit in the first place, but I think this will be easier to get merged then an upright removal. 90% of the stuff in the loadout menu is cosmetic and the 10% that isn't is for the most part not very useful or abusable.

<details>

testing not needed :)

</details>

# Changelog
:cl:
add: Loadout budget increased 7 > 8
/:cl:

### DIFF
--- a/code/__DEFINES/client_prefs.dm
+++ b/code/__DEFINES/client_prefs.dm
@@ -32,7 +32,7 @@
 
 #define AGE_MIN 19 //youngest a character can be
 #define AGE_MAX 90 //oldest a character can be //no. you are not allowed to be 160.
-#define MAX_GEAR_COST 7 //Used in chargen for loadout limit.
+#define MAX_GEAR_COST 8 //Used in chargen for loadout limit.
 
 ///dual_wield_pref from /datum/preferences
 ///Fire both weapons when dual wielding


fira:
The reason there is a limit is that these are meant to be special items that can be leveraged to make yoru character special or provide RP hooks. If you keep increasing the limit, or worse just remove it, it completely defeats the point because suddenly everyone can have a Type 80 Bayonet, CLF pistol, Plush, Goggles, Cigars, and a Zippo lighter. 

If for some reason this is what we want to do, we can just remove the limit. Bumping it by 1 "because this will be easier to get merged than an upright removal" pretty much admits there is no actual point to the change...

567Turtle:
> If for some reason this is what we want to do, we can just remove the limit. Bumping it by 1 "because this will be easier to get merged than an upright removal" pretty much admits there is no actual point to the change...

It appears I've worded myself incorrectly! Allow me to elaborate.

The reason for this PR is to allow a little bit more flexibility to loadout creation, allowing for more interesting / unique combos to people's character designs. Also, while I _do_ want an outright removal of the budget limit, what I meant by "because this will be easier to get merged than an upright removal" is that I do not know how to code such a thing as it _would_ require some limitations to keep people from spawning in with EVERYTHING, and I'm not very interested in delving through loadout code at this time. I agree with your point that increasing the point by one might be kicking the can down the road, so to say, and I'm fully aware of that and will probably come back to this PR at a later date, whether or not it's merged, to create a more permanent fix to the "issue". Right now, I'm just kinda busy.

morrowwolf:
Make him refactor how loadout gear is given to players for the extra point. Shit's going to be overflowing inventories at this point.

567Turtle:
> Make him refactor how loadout gear is given to players for the extra point. Shit's going to be overflowing inventories at this point.

It's an extra point, so, actually, it won't be overflowing inventories. The max this'll let you get is 4 or 5 things, more then enough for the starting satchel to carry :)

PR #5951, opened 2024-03-15
Fixes for Professor Dummy.

Segrain:
# About the pull request

Followup to #5874.

`human/dummy` is designed to be noninteractive. It has godmode (not that godmode does much), it is not processed (thus no bleeding, no medicaments etc), it is not even added to medical HUD. Medical training dummy should not be a subtype of it.
Also shuffled some code around so that, for example, somebody using a dummy in thunderdome does not suddenly lose it on hijack.

# Changelog

:cl:
fix: Professor Dummy in CMO's office now works properly.
/:cl:


Drulikar (approved):
Appears to be working in game

PR #5953, opened 2024-03-15
Pilot Officer Rework

567Turtle:
# About the pull request

Removes both the Pilot Officer and the Dropship Crew Chief. Replaces them with the Gunship Pilot and the Dropship Pilot respectfully. One mans the CAS ship, one mans the transport ship always. No more petty arguing.

Currently with two pilot officers, the round goes one of two ways. 

Either A, there's only one at round start and they get CAS.

Or B, there's two POs and they argue and bicker over who gets CAS. Eventually one of them gets CAS and the other one either cryo's or spends the next 2~ hours bored out of their mind.

Dropship PO is NOT fun. It just isn't. Anybody who says they have ever enjoyed dropship PO is a liar and a rube sent here by the CLF to spread lies and deceit AND they're paid off by the Rothschilds. The whole point of the second PO is to man the Alamo, but with how different CAS PO and Dropship PO are they might as well be two roles.

I really enjoy playing CAS, but the issue is every time I want to roll PO to do so I'm scared away by the thought that I'll get trapped playing dropship for the whole round instead of a marine roll. By separating the two roles like this, now whenever you role CAS pilot, if you get it, you will always get the experience which you want. If you roll Dropship pilot, you will always get the experience you want.

This PR will prevent the countless rounds wasted rolling PO, losing rock paper scissors, and cryoing, having to sit and wait for the next round to start to try again.

As for the DCC removal, I believe it to be an unnecessary and bloating role. This is the case before this PR, but it would especially be the case if this were to be merged without their removal. All of DCCs roles can be done for the most part by MTs, this would also give MTs something to do. Additionally for the learning aspect, I've given the Dropship Pilot the same authority and rank as the old DCCs, so the Dropship Pilot can be used for learning. Ask for tips from the more senior CAS pilot, or others who are knowledgeable and you can get associated with how to control a dropship through the much less important transport ship. Additionally, if you wish to learn the CAS aspect in particular you could play MT and ask to shadow the CAS pilot and assist them.

<details>

🐾🐾🐾🐾🐾🐾🐾🐾🐾

</details>

# Changelog
:cl:
add: Removed PO and DCC, replacing them with the Gunship Pilot who mans CAS and the Dropship Pilot who mans the transport ship.
/:cl:


GrrrKitten:
DCC doesnt get the cool pilot suit option tho :pensive:

567Turtle:
I might've missed something here or there but I think I got everything :)

Johannes2262:
On splitting up the PO role, sure, but why remove DCCs? One could argue that they are unneccessary for transport, but at least as CAS it's nice to have one to speeding up the reloading process, manage marines rappelling and whatnot.

567Turtle:
> On splitting up the PO role, sure, but why remove DCCs? One could argue that they are unneccessary for transport, but at least as CAS it's nice to have one to speeding up the reloading process, manage marines rappelling and whatnot.

I don't find DCCs necessary, I consider them bloat. Additionally if you wish to speed up reloading, ask the MTs to help, I doubt they're doing much.

nuklearcellphoneg:
DCC should be its own PR since they're separate issues tbqh.

TheManWithNoHands:
Your being very subjective here.
There is a decent ammount of players who enjoy both transport and DCC.
They need more of a reason to be enjoyd, not less.

567Turtle:
> They need more of a reason to be enjoyd, not less.

I may be being subjective but I do believe that until more of a reason for existing is made it's fairly safe to remove them. They don't really do much of anything as is and I don't really think it's in my purview to add those things, as my goal with this PR isn't to make DCC more playable, it's to make piloting as a whole more consistent and enjoyable.

code/datums/skills/uscm.dm
```diff
@@ -140,7 +140,7 @@ MILITARY NONCOMBATANT
 	)
 
 /datum/skills/crew_chief
-	name = "Dropship Crew Chief"
+	name = "Dropship Pilot"
 	skills = list(
 		SKILL_PILOT = SKILL_PILOT_TRAINED,
```
MobiusWon, line 145: I recommend giving the Dropship Pilot SKILL_PILOT_EXPERT so they can fill in for the Gunship Pilot in the event of an emergency or vacancy.

code/modules/gear_presets/uscm_ship.dm
```diff
@@ -685,7 +685,7 @@
 //*****************************************************************************************************/
 
 /datum/equipment_preset/uscm_ship/po/full
-	name = "USCM Pilot Officer (PO)"
+	name = "USCM Gunship Pilot (PO)"
```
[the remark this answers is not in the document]
567Turtle: Good catch good catch

Segrain:
> This PR will prevent the countless rounds wasted rolling PO, losing rock paper scissors, and cryoing, having to sit and wait for the next round to start to try again.

If you do not want to play transport pilot, do not sign up as a pilot. If you decide to cryo and sit the round out, it is your own choice.

Also, there is zero reason whatsoever for a pilot in charge of the transport dropship to suddenly be not an officer.

Also, there is zero reason whatsoever for anything in this PR at all.

code/game/jobs/job/command/auxiliary/crew_chief.dm
```diff
@@ -1,19 +1,19 @@
 /datum/job/command/crew_chief
-	title = JOB_DROPSHIP_CREW_CHIEF
-	total_positions = 2
-	spawn_positions = 2
+	title = JOB_DROPSHIP_PILOT
+	total_positions = 1
+	spawn_positions = 1
```
Drulikar, line 4:
```suggestion
	total_positions = 2
	spawn_positions = 2
```
Keep these two. Let there be a co-pilot that can tag team and support the other.

code/game/jobs/job/command/auxiliary/pilot.dm
```diff
@@ -1,20 +1,20 @@
 /datum/job/command/pilot
-	title = JOB_PILOT
-	total_positions = 2
-	spawn_positions = 2
+	title = JOB_CAS_PILOT
+	total_positions = 1
+	spawn_positions = 1
```
Drulikar, line 4:
```suggestion
	total_positions = 2
	spawn_positions = 2
```
Keep these two. Let there be a co-pilot that can tag team and support the other.

Drulikar (changes requested):
This needs to be done in a way that DCC hours transfer to dropship pilot, and PO hours transfer to gunship pilot. Likely its just a matter of keeping the original define value. Otherwise an implementation where we can make multiple legacy values that all contribute to a job's playtime.

code/game/jobs/job/command/auxiliary/crew_chief.dm
567Turtle: Sure, I don't see why not!

567Turtle:
> This needs to be done in a way that DCC hours transfer to dropship pilot, and PO hours transfer to gunship pilot. Likely its just a matter of keeping the original define value. Otherwise an implementation where we can make multiple legacy values that all contribute to a job's playtime.

I'm fairly certain this is already done as I didn't change any type paths, just DEFINE values. It might not've been done correctly? But I _think_ it has been.

morrowwolf:
> This needs to be done in a way that DCC hours transfer to dropship pilot, and PO hours transfer to gunship pilot. Likely its just a matter of keeping the original define value. Otherwise an implementation where we can make multiple legacy values that all contribute to a job's playtime.

That's like a real quick db update. No reason to have the code bend to old stuff when a quick sql line will do the job.

Drulikar:
> > This needs to be done in a way that DCC hours transfer to dropship pilot, and PO hours transfer to gunship pilot. Likely its just a matter of keeping the original define value. Otherwise an implementation where we can make multiple legacy values that all contribute to a job's playtime.
> 
> I'm fairly certain this is already done as I didn't change any type paths, just DEFINE values. It might not've been done correctly? But I _think_ it has been.

I can confirm the hours do not transfer; thus why I asked what I asked.

Drulikar:
> > This needs to be done in a way that DCC hours transfer to dropship pilot, and PO hours transfer to gunship pilot. Likely its just a matter of keeping the original define value. Otherwise an implementation where we can make multiple legacy values that all contribute to a job's playtime.
> 
> That's like a real quick db update. No reason to have the code bend to old stuff when a quick sql line will do the job.

It is not important what the job is internally in code if it just breaks what the DB is looking up.

567Turtle:
> It is not important what the job is internally in code if it just breaks what the DB is looking up.

I dunno changing the type paths back could cause quite a bit of issues. I feel like this is something that needs to be fixed on the DB side if at all possible. If we keep the old type paths stuff could get confusing in the future with roles not matching their type path.

Drulikar:
> > It is not important what the job is internally in code if it just breaks what the DB is looking up.
> 
> I dunno changing the type paths back could cause quite a bit of issues. I feel like this is something that needs to be fixed on the DB side if at all possible. If we keep the old type paths stuff could get confusing in the future with roles not matching their type path.

I will just have to do it myself. Its not the type path; its the string of the define I'm almost certain.

567Turtle:
> > > It is not important what the job is internally in code if it just breaks what the DB is looking up.
> > 
> > 
> > I dunno changing the type paths back could cause quite a bit of issues. I feel like this is something that needs to be fixed on the DB side if at all possible. If we keep the old type paths stuff could get confusing in the future with roles not matching their type path.
> 
> I will just have to do it myself. Its not the type path; its the string of the define I'm almost certain.

Bahh... Weird... You're more then invited to though, go ahead.

Drulikar (changes requested):
Sorry to semi-reverse a previous requested change, but Segrain feels that 2 slots of each will be problematic. But I want there to still be 4 slots for dropships in total. So what I would like instead is 1GP, 1DP, and 2DCC where the GP and DP are the same rank (MO1) and skills (effectively the old 2POs split in half), and then the DCC will be their old skillset and old rank (ME5).

Regarding the playtimes, Harry is willing to edit the DB to accommodate this, but just for future reference this would be how to do that (since I bothered to code and test it): 
image

Likely the only playtime migration will be PO to GP playtime and DP is effectively a new role to earn playtime on.

So to summarize:
- Restore DCC as it was with 2 slots, old rank (ME5), old skills (1pilot, 2 power, 1 leader, 2 medical, 1 surgery, 2 jtac, 2 engineer, 1 firearms, 1 endurance)
- Change DP to be the same rank as GP (MO1). It should have the same skills as now (2 pilot, 2 power, 1 leadership, 1 overwatch, 2 medical, 1 surgery, 2 jtac, 1 intel, 1 firearms, 1 endurance)

567Turtle:
https://github.com/cmss13-devs/cmss13/pull/5958

new pr ucs i pbroke this branch

PR #5955, opened 2024-03-15
Tactical Prybar Works For Substitute Surgical

Katskan:
# About the pull request

Adds the tactical prybar at the same level as a crowbar for improvised surgeries. Previously it wasn't included in the list probably due to an oversight.

It's bad to have types of the same tool that can't do the same thing as each other, it leads to a lot of confusion and can be frustrating. 

Tested out on a local server, worked great

# Changelog
:cl:
fix: Added the tactical prybar to the surgical tools list as a substitute tier tool, similar to the crowbar it's derived from
/:cl:


SabreML (changes requested):
Unless I'm misunderstanding the intent of the PR, the tactical crowbar already works the same as a regular one for surgery, so this isn't actually changing anything:
image
image
<hr>

Whether or not a tool can be used for surgery is decided by `/datum/surgery_step/proc/tool_check()`, which checks if the tool's `type` is in the `tools_cache` list:
https://github.com/cmss13-devs/cmss13/blob/dbf39ae2f9bfdc68c103535e3639e499232e95d3/code/modules/surgery/surgery_steps.dm#L50-L59
`tools_cache` is generated in `/datum/surgery_step/New()`, where it loops through each entry in the `tools` list (in this case `SURGERY_TOOLS_PRY_ENCASED`) and adds both the item and any subtypes to the cache:
https://github.com/cmss13-devs/cmss13/blob/dbf39ae2f9bfdc68c103535e3639e499232e95d3/code/modules/surgery/surgery_steps.dm#L32-L37

Because of this, `/obj/item/tool/crowbar/tactical` has the same behaviour as `/obj/item/tool/crowbar`.

Katskan:
Ok, I took it at face value that it had been reported to me as not working for surgery. May have been something else going on, will re-open if it becomes an issue in the future.

PR #5962, opened 2024-03-17
[s] Hellway: One of the maps of all time

Zonespace27:
TM only for the love of god


Zonespace27:
I can TM closed PRs

iloveloopers:
this appeared on the discord, you failed.

code/game/machinery/computer/computer.dm
```diff
@@ -140,3 +140,60 @@
 /obj/structure/machinery/computer/fixer/New()
 	all_configs = config
 	..()
+
+/obj/structure/red_terminal
```
Huffie56, line 144: if it's not using parent /obj/structure/machinery/computer
it should probably be in it's own file.

Zonespace27: for one-off event PRs, i don't really care

PR #5983, opened 2024-03-19
NVO no longer can be used with binoculars/optics

ihatethisengine:
# About the pull request
Original NVG wouldn't allow you to use them with binos, but when Morrow introduced helmet optics, for some reason he didn't carry over the limitation. I don't know if it was intended or simply was an oversight. Sadly my revert to NVGs (https://github.com/cmss13-devs/cmss13/pull/4807) didn't go through, but I still believe that I was right about points I brought up in the PR, so I am going to address them bit by bit now.

Using nightvision with anything that increases your field of view gives an unreasonable advantage which should be limited only to specs and SGs.

<details>
I did test it
</details>

# Changelog

:cl: ihatethisengine
balance: night vision optics no longer can be used with binoculars.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/helmet_visors.dm
+++ b/code/game/objects/items/devices/helmet_visors.dm
@@ -238,6 +238,7 @@
 		on_light = new(attached_helmet)
 		on_light.set_light_on(TRUE)
 	START_PROCESSING(SSobj, src)
+	RegisterSignal(user, COMSIG_MOB_CHANGE_VIEW, PROC_REF(change_view))
 
 /obj/item/device/helmet_visor/night_vision/deactivate_visor(obj/item/clothing/head/helmet/marine/attached_helmet, mob/living/carbon/human/user)
 	user.remove_client_color_matrix("nvg_visor", 1 SECONDS)
@@ -247,6 +248,7 @@
 	if(visor_glows)
 		qdel(on_light)
 	UnregisterSignal(user, COMSIG_HUMAN_POST_UPDATE_SIGHT)
+	UnregisterSignal(user, COMSIG_MOB_CHANGE_VIEW)
 
 	user.update_sight()
 	STOP_PROCESSING(SSobj, src)
@@ -289,6 +291,12 @@
 		user.see_in_dark = 12
 	user.lighting_alpha = lighting_alpha
 	user.sync_lighting_plane_alpha()
+
+/obj/item/device/helmet_visor/night_vision/proc/change_view(mob/user, new_size)
+	SIGNAL_HANDLER
+	if(new_size > 7) // cannot use binos with NVO
+		var/obj/item/clothing/head/helmet/marine/attached_helmet = loc
+		deactivate_visor(attached_helmet, user)
 
 #undef NVG_VISOR_USAGE
 
--- a/code/game/objects/items/devices/helmet_visors.dm
+++ b/code/game/objects/items/devices/helmet_visors.dm
@@ -301,7 +301,14 @@
 	if(new_size > 7) // cannot use binos with NVO
 		var/obj/item/clothing/head/helmet/marine/attached_helmet = loc
 		if(istype(attached_helmet))
-			attached_helmet.turn_off_visors(user)
+			return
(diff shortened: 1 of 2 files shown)


irakligabunia:
>Original NVG wouldn't allow you to use them with binos, but when Morrow introduced helmet optics, for some reason he didn't carry over the limitation

I'm pretty sure its because NVO's have way worse duration

ihatethisengine:
NVO has the same duration, but a lot more inconvenient to recharge (I may address this later)

morrowwolf:
This was not an oversight.

> NVO has the same duration, but a lot more inconvenient to recharge (I may address this later)

Incorrect.

Do with that what you will.

ihatethisengine:
What exactly is incorrect? 
NVG have 2500 max charge and spend 8 per second, 2500÷8=312.5÷60=5.20 minutes
NVO have 10000 max charge and spend 33 per second, 10000÷33=303.03÷60=5.05 minutes
The difference is literally 9.5 seconds, not that much.

morrowwolf:
Let me be more precise:
It is *dishonest* to categorize them as even remotely similar in regard to charge duration or usage time. You could take 1-2 giant cells and have your NVGs on pretty much 24/7 for a round. Walking past this by just saying it is "a lot more inconvenient to recharge" is blatantly sidestepping extremely pertinent information on past usage which was left out out of the new item on purpose.

Regarding design thoughts:
Overall I had seriously considered axing the optics as well as the goggles for standard use. SGs would also lose their night vision but would get something on their HUD for tracking targets in the dark. Having it just be an in-built MD feels lazy when you could make it so much cooler with client images and such. Darkness should be scary for marines for both gameplay and thematic reasons.

Secondary to that, giving engineers the ability to create on the fly rechargers via APCs was another thought for making power in general more fun to add to items and play with while giving engineers more utility. While possibly annoying at times consider possible recharge requirements for sentries or flashlights. Inventory and gear management is a fairly unique and important part of the overall game and I think it'd benefit from it.

*body armor light flickers*
"Your light is running low on charge..."
Oh shit...

Hunt that dread feeling.

Xeno main out 🧑‍🚀

Diegoflores31:
> Let me be more precise: It is _dishonest_ to categorize them as even remotely similar in regard to charge duration or usage time. You could take 1-2 giant cells and have your NVGs on pretty much 24/7 for a round. Walking past this by just saying it is "a lot more inconvenient to recharge" is blatantly sidestepping extremely pertinent information on past usage which was left out out of the new item on purpose.

but old NVGs , could not use binoculats , had very clear visuals , could be broken in 3 slashes , you had to use your storage to carry cells and lost durability each time the got broken.

code/game/objects/items/devices/helmet_visors.dm
```diff
@@ -271,6 +273,10 @@
 	if(!.)
 		return
 
+	if(user.client.view > 7)
+		to_chat(user, SPAN_WARNING("You cannot use \the [src] while using optics."))
```
Drulikar, line 277:
```suggestion
		to_chat(user, SPAN_WARNING("You cannot use [src] while using optics."))
```

Drulikar (changes requested):
Minor, but you should see if you can also have `/obj/item/clothing/head/helmet/marine/proc/cycle_huds(mob/user)` check whether `can_toggle` returns true with something like
```js
if(!next_visor.can_toggle(user))
	iterator++
	skipped_hud = TRUE
	continue
```

and to the second loop: 
```js
if(!new_visor.can_toggle(user))
	continue
```

Otherwise you get into this state where the visor was selected, but you had the message it won't work, and it won't work when you leave the view again.

image

Alternatively you had the visor try to activate when the view changes back to normal.

Also would be nice if the visors spit out a to_chat if there's nothing to swap to, but up to you if you want to improve that (it'd probably just be a to_chat before the first return FALSE and the final return FALSE in that same proc).

Drulikar:
This plus the other code suggestion should address all my concerns.
[patch.diff.txt](https://github.com/cmss13-devs/cmss13/files/14845473/patch.diff.txt)

Drulikar (approved):

PR #5997, opened 2024-03-24
Preferences for a darker crit overlay and a black flash overlay

Drulikar:
# About the pull request

Now these overlays have less of a shock factor for accessibility. The reasoning for also a preference for crit is because when being revived you might still be deep in crit which can also be a flashbang. These could be combined into one preference, but I feel it didn't hurt to keep them separate so they're more obvious what they affect.

- For dark crit preference, it is the original overlay but colored gray so lessen its brightness.
- For dark flash preference, the screen will turn black instead of white.

See photographs below.

It was requested that there be a preference to lessen the flash effect for people that are photosensitive.

<details>

image

normal
dark

flash_normal
flash_dark

</details>

# Changelog
:cl: Drathek
add: Added preference for a darker crit overlay and black flash overlay
/:cl:

### DIFF
--- a/code/modules/client/preferences_toggles.dm
+++ b/code/modules/client/preferences_toggles.dm
@@ -471,7 +471,7 @@
 	if(result == "White")
 		prefs.flash_overlay_pref = FLASH_OVERLAY_WHITE
 		to_chat(src, SPAN_NOTICE("If flashed your vision will now be white."))
-	if(result == "Dark")
+	else if(result == "Dark")
 		prefs.flash_overlay_pref = FLASH_OVERLAY_DARK
 		to_chat(src, SPAN_NOTICE("If flashed your vision will now be dark."))
 	prefs.save_preferences()
@@ -481,7 +481,7 @@
 	if(result == "White")
 		prefs.crit_overlay_pref = CRIT_OVERLAY_WHITE
 		to_chat(src, SPAN_NOTICE("If in critical condition your vision will now be white."))
-	if(result == "Dark")
+	else if(result == "Dark")
 		prefs.crit_overlay_pref = CRIT_OVERLAY_DARK
 		to_chat(src, SPAN_NOTICE("If in critical condition your vision will now be dark."))
 	prefs.save_preferences()
--- a/code/modules/client/preferences_toggles.dm
+++ b/code/modules/client/preferences_toggles.dm
@@ -453,6 +453,7 @@
 	else
 		CRASH("receive_random_tip() failed: null message")
 
+/// Toggle in character preferences and toggle preferences to configure what kind of blur overlay is used in game; Either blurry, impaired, or legacy.
 /client/proc/set_eye_blur_type()
 	var/result = tgui_alert(src, "What type of eye blur do you want?", "What type of eye blur do you want?", list("Blurry", "Impair", "Legacy"))
 	if(result == "Blurry")
@@ -466,6 +467,7 @@
 		to_chat(src, SPAN_NOTICE("Your vision will now have a legacy blurring effect. This is not recommended!"))
 	prefs.save_preferences()
 
+/// Toggle in character preferences and toggle preferences to configure what kind of flash overlay is used in game; Either white or black.
 /client/proc/set_flash_type()
 	var/result = tgui_alert(src, "What type of flash overlay do you want?", "What type of flash overlay do you want?", list("White", "Dark"))
 	if(result == "White")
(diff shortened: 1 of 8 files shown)


code/modules/client/preferences_toggles.dm
```diff
@@ -464,6 +466,26 @@
 		to_chat(src, SPAN_NOTICE("Your vision will now have a legacy blurring effect. This is not recommended!"))
 	prefs.save_preferences()
 
+/client/proc/set_flash_type()
+	var/result = tgui_alert(src, "What type of flash overlay do you want?", "What type of flash overlay do you want?", list("White", "Dark"))
+	if(result == "White")
+		prefs.flash_overlay_pref = FLASH_OVERLAY_WHITE
+		to_chat(src, SPAN_NOTICE("If flashed your vision will now be white."))
+	if(result == "Dark")
```
Zonespace27, line 474:
```suggestion
	else if(result == "Dark")
```

code/modules/client/preferences_toggles.dm
```diff
@@ -464,6 +466,26 @@
...
+		prefs.flash_overlay_pref = FLASH_OVERLAY_DARK
+		to_chat(src, SPAN_NOTICE("If flashed your vision will now be dark."))
+	prefs.save_preferences()
+
+/client/proc/set_crit_type()
+	var/result = tgui_alert(src, "What type of crit overlay do you want?", "What type of crit overlay do you want?", list("White", "Dark"))
+	if(result == "White")
+		prefs.crit_overlay_pref = CRIT_OVERLAY_WHITE
+		to_chat(src, SPAN_NOTICE("If in critical condition your vision will now be white."))
+	if(result == "Dark")
```
Zonespace27, line 484:
```suggestion
	else if(result == "Dark")
```

code/modules/client/preferences_toggles.dm
```diff
@@ -464,6 +466,26 @@
...
+	var/result = tgui_alert(src, "What type of flash overlay do you want?", "What type of flash overlay do you want?", list("White", "Dark"))
+	if(result == "White")
+		prefs.flash_overlay_pref = FLASH_OVERLAY_WHITE
+		to_chat(src, SPAN_NOTICE("If flashed your vision will now be white."))
+	if(result == "Dark")
+		prefs.flash_overlay_pref = FLASH_OVERLAY_DARK
+		to_chat(src, SPAN_NOTICE("If flashed your vision will now be dark."))
+	prefs.save_preferences()
+
+/client/proc/set_crit_type()
```
Zonespace27, line 479: code doc the two of these please

code/modules/mob/living/living.dm
```diff
@@ -446,8 +446,11 @@
 	return
 
 /mob/living/flash_eyes(intensity = EYE_PROTECTION_FLASH, bypass_checks, type = /atom/movable/screen/fullscreen/flash, flash_timer = 40)
-	if( bypass_checks || (get_eye_protection() < intensity && !(sdisabilities & DISABILITY_BLIND)))
-		overlay_fullscreen("flash", type)
+	if(bypass_checks || (get_eye_protection() < intensity && !(sdisabilities & DISABILITY_BLIND)))
+		if(client?.prefs?.flash_overlay_pref == FLASH_OVERLAY_DARK)
```
Zonespace27, line 450: this doesn't check that `type` is `/atom/movable/screen/fullscreen/flash`
I know that the `type` arg isn't modified anywhere but in silicon code, but i'd like to keep this working as intended for future-proofing sake

Zonespace27 (changes requested):

code/modules/mob/living/living.dm
Drulikar: A `/atom/movable/screen/fullscreen/flash` is not required only a `/atom/movable/screen/fullscreen` is, and that will cause a runtime when it tries to access screen.severity in `/mob/proc/overlay_fullscreen(category, type, severity)`. So I don't see any change necessary here.

`/mob/living/flash_eyes` ultimately just puts an `/atom/movable/screen/fullscreen` on the screen after testing eye protection and animates it out in `flash_timer` time and an animate time of 2 seconds.

SabreML (approved):

code/modules/mob/living/living.dm
SabreML: Hang on sorry, just re-read Zonespace's review.

I think what they meant was that if something in the future adds a `flash_eyes()` call with a `type` of `/atom/movable/screen/fullscreen/new_fancy_screen_effect` (for example) and the player's flash preference is set to 'dark', the `flash_overlay_pref` check will take priority and will always display a `/flash/dark`, regardless of what was passed as `type`.

If this had `istype(type, /atom/movable/screen/fullscreen/flash)` (or something to the same effect) before checking the client's preferences, then it would only do it with relevant calls.
(Maybe instead of having a default `type` argument, the proc could have an empty `type_override` argument instead?)
```
if(type_override)
	overlay_fullscreen("flash", type_override)
else
	current behaviour
```

SabreML (changes requested):

SabreML (approved):
For real this time

PR #5998, opened 2024-03-24
Fixes zombies missing claws on transformation

### LINKED ISSUE #5485: Clawless Zombies
### Testmerges

Irrelevant

### Round ID

20441

### Description of the bug

image
Claw-less zombies can interact with weapons etc

Zombies should get embedded claws in their hands that cannot be dropped etc and stop them from picking stuff up

Buckled to any bed and turn into a zombie
Spawn without claws

### Issue Bingo

vero5123:
# About the pull request
Fixes #5485, added a simple conditional to check if a mob is buckled and then unbuckle them before transformation. I didn't look into the root of the problem but this bug seems niche enough that it shouldn't really effect anything else. If I were to guess, I'm assuming it has to do with the fact that you can't hold items in your hands while buckled to a bed, but you can while you're in a chair. Hence why everything works fine while buckled to a chair.  

# Changelog
:cl:
fix: Fixes zombies missing claws on transformation
/:cl:

### DIFF
--- a/code/datums/diseases/black_goo.dm
+++ b/code/datums/diseases/black_goo.dm
@@ -141,8 +141,6 @@
 
 /datum/disease/black_goo/proc/zombie_transform(mob/living/carbon/human/human)
 	set waitfor = 0
-	if(human.buckled)
-		human.buckled.unbuckle()
 	zombie_is_transforming = TRUE
 	human.vomit_on_floor()
 	human.adjust_effect(5, STUN)
@@ -150,6 +148,8 @@
 	human.make_jittery(500)
 	sleep(30)
 	if(human && human.loc)
+		if(human.buckled)
+			human.buckled.unbuckle()
 		if(human.stat == DEAD)
 			human.revive(TRUE)
 			human.remove_language(LANGUAGE_ENGLISH) // You lose the ability to understand english. Language processing is handled in the mind not the body.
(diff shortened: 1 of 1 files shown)


code/datums/diseases/black_goo.dm
```diff
@@ -141,6 +141,8 @@
 
 /datum/disease/black_goo/proc/zombie_transform(mob/living/carbon/human/human)
 	set waitfor = 0
+	if(human.buckled)
+		human.buckled.unbuckle()
```
SabreML, line 145: This would look a bit better if it happened after the `sleep()`s, I think.

```diff
/datum/disease/black_goo/proc/zombie_transform(mob/living/carbon/human/human)
	set waitfor = 0
-	if(human.buckled)
-		human.buckled.unbuckle()
	zombie_is_transforming = TRUE
	human.vomit_on_floor()
	human.adjust_effect(5, STUN)
	sleep(20)
	human.make_jittery(500)
	sleep(30)
	if(human && human.loc)
+		if(human.buckled)
+			human.buckled.unbuckle()
		if(human.stat == DEAD)
```
<hr>

**Current:** (Unbuckles, plays a little animation, then transforms)

https://github.com/cmss13-devs/cmss13/assets/57483089/192652af-fbc1-42db-a411-546159da1aa2

**Proposed Change:** (Plays a little animation, unbuckles, then transforms)

https://github.com/cmss13-devs/cmss13/assets/57483089/4fd9a358-ca1c-4a5d-b251-2904ab7ba1ad

SabreML (commented):
Thanks for the fix! Just a small "stylistic" nitpick.

SabreML (approved):

PR #6002, opened 2024-03-24
Faster snow removal for bigger xenomorphs

cuberound:
# About the pull request

makes the snow removal dependant of xenomorph size, making it 1,5 ( lesser) to 4 times (big xenos) faster ( still takes a while even for the largest xenos to remove deap snow)

removing snow is painful job that makes maps with snow unwanted, huge xenomorphs should not have that hard time showeling some snow out of their way, makes it fair with the flames removing snow PR 

<details>

</details>

# Changelog
:cl:
balance: greatly decreses snow removal time, now it depends on xenomorph size 
/:cl:


SabreML (changes requested):
It looks like `/turf/open/snow` isn't actually used anywhere other than a nightmare variant of CORSAT. Everything else uses `/turf/open/auto_turf/snow`, so this needs to be added to that as well.
(I don't know why we have two mostly identical versions of a snow turf)

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
```diff
@@ -877,7 +877,7 @@
 
 	while(bleed_layer > 0)
 		xeno_attack_delay(M)
-		if(!do_after(M, 12, INTERRUPT_ALL, BUSY_ICON_FRIENDLY))
+		if(!do_after(M, 12/M.mob_size, INTERRUPT_ALL, BUSY_ICON_FRIENDLY))
```
Drulikar, line 880: Division by zero is technically possible.
```suggestion
		var/size = max(M.mob_size, 1)
		if(!do_after(M, 12/size, INTERRUPT_ALL, BUSY_ICON_FRIENDLY))
```

Drulikar (changes requested):

SabreML (approved):
LGTM!
I was wondering if the time should be changed from `12` to `18` so that it's unchanged for lesser drones (they now have a time of `8` due to a mob size of `MOB_SIZE_XENO_VERY_SMALL`), but I don't think it's that much of an issue.
IMO it can just be changed later on if it seems too fast.

Drulikar (approved):

RavinHobo:
crusher snow plough variant when?

PR #6003, opened 2024-03-24
Fixes carlarc's OP chocolate

### LINKED ISSUE #4695: CHUNK box breaks click delay
### Testmerges

#4575, #4539, #4337, #4653, #4475, #4679

### Description of the bug

Using the CHUNK box as a melee weapon against resin walls or weeds removes the melee click delay while acting as a bayonet knife.

It should respect click delay as most items do.

1. Grab a CHUNK food box
2. Spam click resin

### Issue Bingo

vero5123:
# About the pull request
Fixes #4695, The issue isn't inherently the attack speed, although it is quite fast, many small items also have a relatively fast attack speed. The problem is that when you combine the attack speed with a food item that somehow has a force of 35 you get a very effective weapon capable of destroying many different types of structures fairly easily, same issue with the scalpel.

My solution here was to just decrease the amount of force the items have rather than lowering the attack speed. You shouldn't be able to use a scalpel or a box of chocolate to more effectively destroy weeds rather than a machete. I doubt these are the only two items that have this problem, I guess in the future when someone comes across a similar issue just link this pr so they can easily identify the exploit. 

# Changelog

:cl: 
fix: lowers the scalpel and chunk box's ability to destroy various structures.
add: chunk box is now a melee weapon
/:cl:

### DIFF
--- a/code/game/machinery/bots/bots.dm
+++ b/code/game/machinery/bots/bots.dm
@@ -84,9 +84,9 @@
 		if(hasvar(W,"force") && hasvar(W,"damtype"))
 			switch(W.damtype)
 				if("fire")
-					src.health -= W.force * fire_dam_coeff
+					src.health -= W.force * W.demolition_mod * fire_dam_coeff
 				if("brute")
-					src.health -= W.force * brute_dam_coeff
+					src.health -= W.force * W.demolition_mod * brute_dam_coeff
 			..()
 			healthcheck()
 		else
--- a/code/game/machinery/bots/bots.dm
+++ b/code/game/machinery/bots/bots.dm
@@ -84,9 +84,9 @@
 		if(hasvar(W,"force") && hasvar(W,"damtype"))
 			switch(W.damtype)
 				if("fire")
-					src.health -= W.force * W.demolition_mod * fire_dam_coeff
+					health -= W.force * W.demolition_mod * fire_dam_coeff
 				if("brute")
-					src.health -= W.force * W.demolition_mod * brute_dam_coeff
+					health -= W.force * W.demolition_mod * brute_dam_coeff
 			..()
 			healthcheck()
 		else
--- a/code/game/machinery/deployable.dm
+++ b/code/game/machinery/deployable.dm
@@ -51,9 +51,9 @@
 	else
 		switch(W.damtype)
 			if("fire")
-				src.health -= W.force * 0.75
+				src.health -= W.force * W.demolition_mod * 0.75
 			if("brute")
-				src.health -= W.force * 0.5
+				src.health -= W.force * W.demolition_mod * 0.5
 		if (src.health <= 0)
 			src.explode()
 		..()
--- a/code/game/machinery/deployable.dm
+++ b/code/game/machinery/deployable.dm
@@ -51,11 +51,11 @@
 	else
 		switch(W.damtype)
 			if("fire")
-				src.health -= W.force * W.demolition_mod * 0.75
+				health -= W.force * W.demolition_mod * 0.75
 			if("brute")
-				src.health -= W.force * W.demolition_mod * 0.5
-		if (src.health <= 0)
-			src.explode()
+				health -= W.force * W.demolition_mod * 0.5
+		if (health <= 0)
+			explode()
 		..()
 
 /obj/structure/machinery/deployable/barrier/ex_act(severity)
--- a/code/game/objects/items/reagent_containers/food/snacks.dm
+++ b/code/game/objects/items/reagent_containers/food/snacks.dm
@@ -61,7 +61,7 @@
 		if(user == M)
 			to_chat(M, SPAN_WARNING("How do you expect to eat this with the package still on?"))
 		else
-			to_chat(M, SPAN_WARNING("[M] tried to feed you a [src] with the package still on, what a silly goose!"))
+			to_chat(M, SPAN_WARNING("[M] made an endearing attempt to force feed you a snack, albeit with the charming mishap of leaving the packaging intact."))
 		return FALSE
 
 	if(istype(M, /mob/living/carbon))
--- a/code/game/objects/items/reagent_containers/food/snacks.dm
+++ b/code/game/objects/items/reagent_containers/food/snacks.dm
@@ -54,10 +54,10 @@
 		qdel(src)
 		return FALSE
 
-	if(package && user.a_intent == INTENT_HARM)
-		return ..() // chunk box gaming
+	if(package)
+		if(user.a_intent == INTENT_HARM)
+			return ..() // chunk box gaming
 
-	if(package && user.a_intent != INTENT_HARM)
 		if(user == M)
 			to_chat(M, SPAN_WARNING("How do you expect to eat this with the package still on?"))
 		else
--- a/code/game/objects/items/reagent_containers/food/snacks.dm
+++ b/code/game/objects/items/reagent_containers/food/snacks.dm
@@ -3305,8 +3305,8 @@
 	icon_state = "hunk"
 	w_class = SIZE_MEDIUM
 	hitsound = "swing_hit"
-	force = 4
-	throwforce = 2
+	force = 35
+	throwforce = 50
 	bitesize = 20
 	wrapper = /obj/item/trash/chunk/hunk
 
--- a/code/game/objects/items/reagent_containers/food/snacks.dm
+++ b/code/game/objects/items/reagent_containers/food/snacks.dm
@@ -3289,8 +3289,8 @@
 	name = "CHUNK box"
 	desc = "A bar of \"The <b>CHUNK</b>\" brand chocolate. <i>\"The densest chocolate permitted to exist according to federal law. We are legally required to ask you not to use this blunt object for anything other than nutrition.\"</i>"
 	icon_state = "chunk"
-	force = 15 //LEGAL LIMIT OF CHOCOLATE
-	throwforce = 10
+	force = 2
+	throwforce = 1
 	bitesize = 2
 	wrapper = /obj/item/trash/chunk
 
@@ -3305,8 +3305,8 @@
 	icon_state = "hunk"
 	w_class = SIZE_MEDIUM
 	hitsound = "swing_hit"
-	force = 35 //ILLEGAL LIMIT OF CHOCOLATE
-	throwforce = 50
+	force = 2
+	throwforce = 1
 	bitesize = 20
 	wrapper = /obj/item/trash/chunk/hunk
 
--- a/code/game/objects/items/reagent_containers/food/snacks.dm
+++ b/code/game/objects/items/reagent_containers/food/snacks.dm
@@ -61,7 +61,7 @@
 		if(user == M)
 			to_chat(M, SPAN_WARNING("How do you expect to eat this with the package still on?"))
 		else
-			to_chat(M, SPAN_WARNING("[M] made an endearing attempt to force feed you a snack, albeit with the charming mishap of leaving the packaging intact."))
+			to_chat(M, SPAN_WARNING("[M] made an endearing attempt to force feed you a snack with the packaging still on."))
 		return FALSE
 
 	if(istype(M, /mob/living/carbon))
--- a/code/game/objects/items/reagent_containers/food/snacks.dm
+++ b/code/game/objects/items/reagent_containers/food/snacks.dm
@@ -3305,8 +3305,8 @@
 	icon_state = "hunk"
 	w_class = SIZE_MEDIUM
 	hitsound = "swing_hit"
-	force = 2
-	throwforce = 1
+	force = 4
+	throwforce = 2
 	bitesize = 20
 	wrapper = /obj/item/trash/chunk/hunk
 
--- a/code/game/objects/items/reagent_containers/food/snacks.dm
+++ b/code/game/objects/items/reagent_containers/food/snacks.dm
@@ -3298,6 +3298,7 @@
 	hitsound = "swing_hit"
 	force = 15
 	throwforce = 10
+	attack_speed = 10
 	demolition_mod = 0.3
 	bitesize = 2
 	wrapper = /obj/item/trash/chunk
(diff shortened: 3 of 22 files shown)


Blundir:
The whole gimmick of chunk box is it having fun amount of damage, which is a reference to its historical real life counterpart, here you just kill the whole reference joke (without even editing the item description) despite having ability to just set low attack speed, or make it so the chunk box breaks (deletes) after being used a weapon a few times, or you just could've been more smart about it and ported[ this ](https://github.com/tgstation/tgstation/pull/66967)instead, which'd solve any possible similar problems in the future forever.

vero5123:
> The whole gimmick of chunk box is it having fun amount of damage, which is a reference to its historical real life counterpart, here you just kill the whole reference joke (without even editing the item description) despite having ability to just set low attack speed, or make it so the chunk box breaks (deletes) after being used a weapon a few times, or you just could've been more smart about it and ported[ this ](https://github.com/tgstation/tgstation/pull/66967)instead, which'd solve any possible similar problems in the future forever, also your pr doesn't even fix the referenced issue to begin with.

Obviously I'll wait for maintainer input on the pr, but a few things
-The person who made the issue believed that the problem was the click delay. However, I don't believe the issue was necessarily the attack speed. The real problem imo was that you can use certain items to essentially destroy structures in a way they were not intended to do. This pr fixed that for the items listed, by modifying the force and not the attack speed, which made destroying structures take a lot longer than before. Hence now you would use a machete to destroy weeds and not a scalpel. Although other items most definitely have this problem.
-This is my personal opinion but while the linked pr you provided does solve the problem, I don't think adding complexity is necessarily the best approach (feel free to disagree). Currently as it stands any future similar issues that may arise would be from a wonky combination of force and attack_speed. This in my opinion can be mitigated by simply having a few defines for the type of attack speed we want, that way we can make everything consistent and remove the variability of the attack speeds from people just throwing in random numbers. People would then quickly identify the type of attack speed they need for their item based off the existing defined values. Force is a lot more variable in general so we can leave that as is.

vero5123:
Ok, took a closer look into the codebase and the attack speed is too variable to make it consistent between a few defines. It would essentially be a balance change which is not ideal for a bug fix. I'll go ahead and port the tgui implementation.

vero5123:
So for the maintainers to catch up, there was a discussion on discord. Turns out the whole meme of the chunk box being a weapon is in-itself a meme because you can't even hit mobs with it; the chunk box now has a demolition modifier making it pretty much useless against structures but an effective melee weapon and the attack speed was left as it was before. If it needs to be lowered then that can happen.

code/game/machinery/bots/bots.dm
```diff
@@ -84,9 +84,9 @@
 		if(hasvar(W,"force") && hasvar(W,"damtype"))
 			switch(W.damtype)
 				if("fire")
-					src.health -= W.force * fire_dam_coeff
+					src.health -= W.force * W.demolition_mod * fire_dam_coeff
```
Vicacrov, line 87: We don't use explicit src

```suggestion
					health -= W.force * W.demolition_mod * fire_dam_coeff
```

code/game/machinery/bots/bots.dm
```diff
@@ -84,9 +84,9 @@
...
 				if("brute")
-					src.health -= W.force * brute_dam_coeff
+					src.health -= W.force * W.demolition_mod * brute_dam_coeff
```
Vicacrov, line 89:
```suggestion
					health -= W.force * W.demolition_mod * brute_dam_coeff
```

code/game/objects/structures/lamarr_cage.dm
```diff
@@ -53,7 +53,7 @@
 
 
 /obj/structure/lamarr/attackby(obj/item/W as obj, mob/user as mob)
-	src.health -= W.force
+	src.health -= W.force * W.demolition_mod
```
Vicacrov, line 56:
```suggestion
	health -= W.force * W.demolition_mod
```

code/game/machinery/deployable.dm
```diff
@@ -51,9 +51,9 @@
 	else
 		switch(W.damtype)
 			if("fire")
-				src.health -= W.force * 0.75
+				src.health -= W.force * W.demolition_mod * 0.75
```
Vicacrov, line 54:
```suggestion
				health -= W.force * W.demolition_mod * 0.75
```

code/game/machinery/deployable.dm
```diff
@@ -51,9 +51,9 @@
...
 			if("brute")
-				src.health -= W.force * 0.5
+				src.health -= W.force * W.demolition_mod * 0.5
```
Vicacrov, line 56:
```suggestion
				health -= W.force * W.demolition_mod * 0.5
```

code/game/machinery/deployable.dm
```diff
@@ -51,9 +51,9 @@
...
 		if (src.health <= 0)
```
Vicacrov, line 57:
```suggestion
		if (health <= 0)
```

code/game/machinery/deployable.dm
```diff
@@ -51,9 +51,9 @@
...
 			src.explode()
```
Vicacrov, line 58:
```suggestion
			explode()
```

code/game/objects/structures/lamarr_cage.dm
```diff
@@ -53,7 +53,7 @@
...
 	src.healthcheck()
```
Vicacrov, line 57:
```suggestion
	healthcheck()
```

code/game/objects/structures/mirror.dm
```diff
@@ -124,7 +124,7 @@
 		playsound(src.loc, 'sound/effects/hit_on_shattered_glass.ogg', 25, 1)
```
Vicacrov, line 124:
```suggestion
		playsound(loc, 'sound/effects/hit_on_shattered_glass.ogg', 25, 1)
```

code/modules/vehicles/vehicle.dm
```diff
@@ -84,9 +84,9 @@
 	else if(W.force)
 		switch(W.damtype)
 			if("fire")
-				health -= W.force * fire_dam_coeff
+				health -= W.force * W.demolition_mod * fire_dam_coeff
 			if("brute")
-				health -= W.force * brute_dam_coeff
+				health -= W.force * W.demolition_mod * brute_dam_coeff
 		playsound(src.loc, "smash.ogg", 25, 1)
```
Vicacrov, line 90:
```suggestion
		playsound(loc, "smash.ogg", 25, 1)
```

code/game/objects/items/reagent_containers/food/snacks.dm
```diff
@@ -3304,8 +3313,7 @@
 	desc = "A 'crate', as the marketing called it, of \"The <b>HUNK</b>\" brand chocolate. An early version of the CHUNK box, the HUNK bar was hit by a class action lawsuit and forced to go into bankruptcy and get bought out by the Company when hundreds of customers had their teeth crack from simply attempting to eat the bar."
 	icon_state = "hunk"
 	w_class = SIZE_MEDIUM
-	hitsound = "swing_hit"
-	force = 35 //ILLEGAL LIMIT OF CHOCOLATE
+	force = 35
```
Vicacrov, line 3316: I just want to mention here that one hit on the chest from this when the person has no armour on can (and will) cause internal bleeding. (From testing) Not sure if this is the original intention here, just a heads up

code/game/objects/items/reagent_containers/food/snacks.dm
```diff
@@ -54,8 +54,14 @@
...
-	if(package)
-		to_chat(M, SPAN_WARNING("How do you expect to eat this with the package still on?"))
+	if(package && user.a_intent == INTENT_HARM)
+		return ..() // chunk box gaming
+
+	if(package && user.a_intent != INTENT_HARM)
+		if(user == M)
+			to_chat(M, SPAN_WARNING("How do you expect to eat this with the package still on?"))
+		else
+			to_chat(M, SPAN_WARNING("[M] tried to feed you a [src] with the package still on, what a silly goose!"))
```
Vicacrov, line 64: If the packaged food is improper, this will lead to "feed you a the foodname with"

image

This is a fucky situation, because most packaged foods are proper (Packaged Cheeseburger, Packaged Burrito), whereas MRE components are improper. It'd be nice if most packaged food were all improper (lowercase) to make this message uniform and then maybe add an `\a` or `\an` macro (I cannot remember which).

code/game/objects/items/reagent_containers/food/snacks.dm
vero5123: hunk is admin only.

code/game/objects/structures/mirror.dm
vero5123: this isn't part of my fix, but yeah.

code/game/objects/items/reagent_containers/food/snacks.dm
vero5123: good catch

SabreML (changes requested):

code/game/objects/items/tools/surgery_tools.dm
```diff
@@ -96,8 +96,9 @@
 	force = 10
 	sharp = IS_SHARP_ITEM_ACCURATE
 	edge = 1
+	demolition_mod = 0.1
 	w_class = SIZE_TINY
-	throwforce = 5
+	throwforce = 1
```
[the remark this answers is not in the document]
vero5123: forgot to change that back after the demolition changes.

code/game/objects/items/reagent_containers/food/snacks.dm
```diff
@@ -54,8 +54,14 @@
...
-	if(package)
-		to_chat(M, SPAN_WARNING("How do you expect to eat this with the package still on?"))
+	if(package && user.a_intent == INTENT_HARM)
+		return ..() // chunk box gaming
+
+	if(package && user.a_intent != INTENT_HARM)
+		if(user == M)
+			to_chat(M, SPAN_WARNING("How do you expect to eat this with the package still on?"))
+		else
+			to_chat(M, SPAN_WARNING("[M] made an endearing attempt to force feed you a snack, albeit with the charming mishap of leaving the packaging intact."))
```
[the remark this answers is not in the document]
vero5123: better?

code/game/objects/structures/fence.dm
```diff
@@ -170,9 +170,9 @@
 	else
 		switch(W.damtype)
 			if("fire")
-				health -= W.force
+				health -= W.force * W.demolition_mod
 			if("brute")
-				health -= W.force * 0.1
+				health -= (W.force * W.demolition_mod) * 0.1
```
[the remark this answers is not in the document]
vero5123: very true.

code/game/objects/items/reagent_containers/food/snacks.dm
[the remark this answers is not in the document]
SabreML: Yeah that works 👍

code/game/objects/items/reagent_containers/food/snacks.dm
```diff
@@ -55,7 +55,13 @@
...
 
 	if(package)
-		to_chat(M, SPAN_WARNING("How do you expect to eat this with the package still on?"))
+		if(user.a_intent == INTENT_HARM)
+			return ..() // chunk box gaming
+
+		if(user == M)
+			to_chat(M, SPAN_WARNING("How do you expect to eat this with the package still on?"))
+		else
+			to_chat(M, SPAN_WARNING("[M] made an endearing attempt to force feed you a snack with the packaging still on."))
```
SabreML, line 64: The person being clicked on here is `M`, so the message should refer to `user` instead:
```suggestion
			to_chat(M, SPAN_WARNING("[user] made an endearing attempt to force feed you a snack with the packaging still on."))
```

code/game/objects/structures/barricade/misc.dm
```diff
@@ -106,6 +106,6 @@
 /obj/structure/barricade/wooden/hit_barricade(obj/item/I)
 	switch(I.damtype)
 		if("fire")
-			take_damage( I.force * 1.5 )
+			take_damage( I.force  * I.demolition_mod * 1.5 )
```
SabreML, line 109:
```suggestion
			take_damage( I.force * I.demolition_mod * 1.5 )
```

code/game/machinery/bots/bots.dm
```diff
@@ -84,9 +84,9 @@
 		if(hasvar(W,"force") && hasvar(W,"damtype"))
 			switch(W.damtype)
 				if("fire")
-					src.health -= W.force * fire_dam_coeff
+					health -= W.force * W.demolition_mod * fire_dam_coeff
 				if("brute")
-					src.health -= W.force * brute_dam_coeff
+					health -= W.force * W.demolition_mod * brute_dam_coeff
```
SabreML, line 89: There's a few extra things that could have the `demolition_mod` modifier added to them.
Specifically:
[/obj/structure/displaycase](https://github.com/cmss13-devs/cmss13/blob/master/code/game/objects/structures/displaycase.dm#L60)
[/obj/structure/mineral_door](https://github.com/cmss13-devs/cmss13/blob/b5e180c53acbca0fcc195e9b9d13a7c42b3f4a92/code/game/objects/structures/mineral_doors.dm#L105)
[/obj/structure/alien/movable_wall](https://github.com/cmss13-devs/cmss13/blob/b5e180c53acbca0fcc195e9b9d13a7c42b3f4a92/code/game/turfs/walls/wall_types.dm#L1047) (could skip this one actually, I don't think it's used anywhere)
[/turf/closed/wall/resin](https://github.com/cmss13-devs/cmss13/blob/b5e180c53acbca0fcc195e9b9d13a7c42b3f4a92/code/game/turfs/walls/wall_types.dm#L1283)
[/obj/structure/mineral_door/resin](https://github.com/cmss13-devs/cmss13/blob/b5e180c53acbca0fcc195e9b9d13a7c42b3f4a92/code/modules/cm_aliens/XenoStructures.dm#L394)

SabreML (commented):
Couple more things

SabreML (approved):

PR #6004, opened 2024-03-24
Fixes armor light staying on after removal

### LINKED ISSUE #5672: Armor Light Not Turning Off
### Testmerges

#5007, #5392, #5552, #5628, #5669

### Round ID

20751

### Description of the bug

Armor light, when removed manually, is not turning off and can not be slashed off by xenos in the same way they can slash flashlights or guns with rail lights turned on. Curiously, the armor light still maintains the directionallity of the person when they removed said armor. 

Armor light turned on while armor is on the ground:
image

Me slashing at the armor to no success: 
image

Armor light should not be usable as a permanent, non-slashable light source. It also looks really weird being directional while it exists as a sprite on the ground.

1. Wear armor with a light on it.
2. Activate armor light.
3. Remove armor without turning the light off manually.
4. Pitch armor with light somewhere visible.
5. Profit off the inexhaustible light source.

### Issue Bingo

vero5123:
# About the pull request
Fixes #5672 , just put a conditional to check if the armor contains a light source that is currently turned on, if so we just turn it off when the user removes it from their inventory.

# Changelog
:cl:
fix: Fixes armor light staying on after removal.

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -459,8 +459,8 @@ cases. Override_icon_state should be a list.*/
 
 	if(item.flags_equip_slot & slotdefine2slotbit(slot))
 		if(is_type_in_list(item, uniform_restricted))
-			if(src.light_on)
-				src.turn_light(toggle_on = FALSE)
+			if(light_on)
+				turn_light(toggle_on = FALSE)
 			user.drop_inv_item_on_ground(src)
 			to_chat(user, SPAN_NOTICE("You drop \the [src] to the ground while unequipping \the [item]."))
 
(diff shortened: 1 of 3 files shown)


Superwelder:
Many thanks for looking at my issue!

code/game/objects/items.dm
```diff
@@ -459,6 +459,8 @@ cases. Override_icon_state should be a list.*/
 
 	if(item.flags_equip_slot & slotdefine2slotbit(slot))
 		if(is_type_in_list(item, uniform_restricted))
+			if(src.light_on)
+				src.turn_light(toggle_on = FALSE)
```
harryob, line 463:
```suggestion
			if(light_on)
				turn_light(toggle_on = FALSE)
```

harryob (changes requested):
we don't use explicit src

vero5123:
> Many thanks for looking at my issue! 

Np

harryob (approved):

PR #6006, opened 2024-03-24
Fixes chair/bed teleporting

### LINKED ISSUE #5604: Chair Teleporting
### Testmerges

#5086, 5392, 5519, 5007

### Round ID

20656

### Description of the bug

When you grab someone and click on a chair that is adjacent to you but more than 1 tile away from the person you grabbed, they will teleport to that location.

Nothing should happen.

1. Grab someone
2. Click on a chair not directly adjacent to the person grabbed, and with that grab in your active hand
3. Watch them teleport

### Issue Bingo

vero5123:
# About the pull request
Fixes #5604, prevents placing mobs onto chairs, beds and tables unless the mob is directly adjacent to the source tile. I'm hesitant to extend this functionality for the autodoc/scanner and xeno nests as I could see it getting very annoying quickly. 

# Changelog

:cl:
fix: Fixes teleporting when placing mobs onto chairs and beds. 
/:cl:

### DIFF
--- a/code/game/objects/structures/stool_bed_chair_nest/bed.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/bed.dm
@@ -166,7 +166,7 @@
 		if(ismob(G.grabbed_thing))
 			var/mob/M = G.grabbed_thing
 			var/atom/blocker = LinkBlocked(user, user.loc, loc)
-			if(!src.Adjacent(M))
+			if(!Adjacent(M))
 				visible_message(SPAN_DANGER("[M] is too far to place onto [src]."))
 				return FALSE
 			if(blocker)
--- a/code/game/objects/structures/stool_bed_chair_nest/bed.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/bed.dm
@@ -166,11 +166,14 @@
 		if(ismob(G.grabbed_thing))
 			var/mob/M = G.grabbed_thing
 			var/atom/blocker = LinkBlocked(user, user.loc, loc)
+			if(!src.Adjacent(M))
+				visible_message(SPAN_DANGER("[M] is too far to place onto the [src]"))
+				return FALSE
 			if(blocker)
 				to_chat(user, SPAN_WARNING("\The [blocker] is in the way!"))
-			else
-				to_chat(user, SPAN_NOTICE("You place [M] on [src]."))
-				M.forceMove(loc)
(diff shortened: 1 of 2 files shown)


vero5123:
Yeah, i'm not quite sure if this requires a fix tbh. I just like the feel of slamming someones face onto a table 2 tiles away.

code/game/objects/structures/stool_bed_chair_nest/bed.dm
```diff
@@ -166,11 +166,14 @@
 		if(ismob(G.grabbed_thing))
 			var/mob/M = G.grabbed_thing
 			var/atom/blocker = LinkBlocked(user, user.loc, loc)
+			if(!src.Adjacent(M))
+				visible_message(SPAN_DANGER("[M] is too far to place onto the [src]"))
```
SabreML, line 170: Referencing an atom in a string will automatically add "the" before it if its name is an improper noun, so the "the" isn't needed. (also, needs a full stop/period)
**Before:**
before
**After:**
after
```suggestion
				visible_message(SPAN_DANGER("[M] is too far to place onto [src]."))
```

code/game/objects/structures/stool_bed_chair_nest/bed.dm
```diff
@@ -166,11 +166,14 @@
 		if(ismob(G.grabbed_thing))
 			var/mob/M = G.grabbed_thing
 			var/atom/blocker = LinkBlocked(user, user.loc, loc)
+			if(!src.Adjacent(M))
```
SabreML, line 169: The `src.` is implied here:
```suggestion
			if(!Adjacent(M))
```

code/game/objects/structures/tables_racks.dm
```diff
@@ -270,6 +270,9 @@
 		var/obj/item/grab/G = W
 		if(istype(G.grabbed_thing, /mob/living))
 			var/mob/living/M = G.grabbed_thing
+			if(!src.Adjacent(M))
+				visible_message(SPAN_DANGER("[M] is too far to place onto [src]"))
+				return
```
SabreML, line 275: I think it would be fine to leave this out for tables. Like you said it's pretty satisfying to do, and it sort of makes sense intuitively that you can "throw"/"teleport" someone around you with a table slam since it's a combat action.

SabreML (approved):

PR #6012, opened 2024-03-25
Allows helmets to protect Synthetics

QuickLode:
# About the pull request

This PR allows helmets to once more protect Synthetics, undoing a hidden mechanic which was implemented almost 8 years ago.

Many people don't know this mechanic exists, including some of my own councilors. Originally I thought it was pretty unique for Synthetics to be decapitated via a stray bullet or slash, causing a talking head to appear and interact. I myself have had some interesting RP come from this(like guiding a doctor on how to repair the body, live) and so I understood the appeal. 

However, this mechanic makes no sense from any perspective other than to allow easy decaps, and this too has been alleviated. Helmets can now be removed in the same way a head can be removed - by damaging the head(helmet). Thus, this mechanic has certainly become antiquated, and I don't expect this change to have a large impact on gameplay outside of scenarios where, if things were about to hit the fan, then a Synthetic may wear it(riot events, HvH scenarios, special circumstances etc.)

As it stands, even with the shroud of this hidden mechanic, only 1-3% of Synthetics wore a helmet in the first place, as usual, it was to complement the appearance of the Synthetic. I don't predict a serious shift outside of scenarios where people would usually go out of their way to wear helmets anywho, such as a hostile human boarding team or riot scenarios, as aforementioned. 

Wearing something to protect, that offers no protection is absolutely misleading. Removing the protection mechanic from the game is unnecessary because helmets can still be removed through traditional means thanks to changes introduced since. The justification of removing the protection mechanic from Synthetic serves no purpose and it is not only so hidden and unexplained, but it genuinely is unnecessary due to decapitation being fully possible. Even under the threat of HvH, friendly fire, stray bullets or special circumstances - you will STILL be hard pressed to find a Synthetic wearing a helmet. For the times that they do, it should act as intended.

 This mechanic currently circumvents the 'popping off helmet' mechanic which in and of itself is a very unique and cool mechanic to witness. By undoing it, **we get both unique cases** within the realm of possibility: a helmet popping off with a big *THUNK* and the decapitation of a Synthetic. 

<details>

current: https://cdn.discordapp.com/attachments/1042176396711170119/1221661321595195474/image.png?ex=661363c4&is=6600eec4&hm=3815815e392d0aabcbb0728d9cc699b0c6bfe80d0bdea9676d7a86ee34953762&
(helmet or no helmet results in possible decapitation, without any notice or input)

after PR: https://cdn.discordapp.com/attachments/1042176396711170119/1221661255450755092/image.png?ex=661363b4&is=6600eeb4&hm=a0a4c3c9228cce27939139ef2f818b5df8781400bae94ede5217e90411fc021b&

(heavily wounded, helmet is popped off with a *POP* noise, then followed by decapitation if further damage occurs. no helmet results in possible decapitation as intended)
</details>

# Changelog
:cl:
balance: Helmets now properly protect Synthetics from sudden decapitation if worn, and can be knocked off just like a human.
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -1503,7 +1503,7 @@ treat_grafted var tells it to apply to grafted but unsalved wounds, for burn kit
 /obj/limb/head/limb_delimb(damage_source)
 	var/obj/item/clothing/head/helmet/owner_helmet = owner.head
 
-	if(!istype(owner_helmet) || !owner.allow_gun_usage)
+	if(!istype(owner_helmet))
 		droplimb(0, 0, damage_source)
 		return
 


morrowwolf:
It's been a core design principle for synths for quite some time that they should not be getting into serious danger in the first place and as such do not get armor bonuses in any form.

Thinking back, I'm almost certain the last ?5? game architects had wanted to bring synths more in line strength-wise. I know myself, Nanu, and Harry had some pretty long conversations about excesses we saw within the whitelist. Luckily for some of you reading this, I quit before we got around to working on it. Even those who had added things to synthetic had confided later they questioned if their additions were a good idea in the end.

Synthetics are extremely strong and have high skills in every supporting task. It's not uncommon to see even mediocre synth players capable of surviving incredible situations or even winning fights against multiple xenos when cornered or murder hobo'ing during hijack. They have built in resistances to all damages, don't go down til they legit die (and then can still be revived forever), and even have a snowflake dodge for acid blood damage.

Needless to say, synthetics are not wanting for power in game. This has only been allowed to be the case due to their special status as a whitelisted role and guarantees made to the maintainer team they will rein in excess and keep high standards for RP. I haven't been around recently much to judge whether or not this has slid anymore than it had. I have friends who are current members of the synth council and many who were former. I had gone to the synth whitelist many times for feedback on support changes and there are quite a few intelligent folk there. Even so, synths need no help when it comes to gameplay *and* I fully believe there has been a slow but sure walk towards allowing more and more anti-team and powergame-y behaviors for a while. Maybe recent events I am not privy to will make me wrong on this.

For gameplay, I believe this change would be another step in incentivizing poor behavior from some synth players. I do not have confidence the synthetic council is willing to look objectively at itself and the oversight that used to come from less biased staff members had broken down back when I was still around. But, who knows, maybe things are better.

As for roleplay/simulation (kind of combined in this context), yes, synthetics in lore are extremely strong and extremely durable and can use guns and and and and. This exception (and frankly the overall design of the role) is dumb from that perspective. On the other hand, these types of weaknesses are meant to incentivize synths to avoid combat and focus on that support and roleplay ideal. I don't see compelling reason to have the intent of the role and the actual role in game be at odds with each other. Saying all that, the helmet just magically not working for synths is dumb.

Helmets not giving a bonus aligns with the current, *and eight year old*, design. Armor stabilization that forces armor for synthetics to a specific number that cannot go up or down no matter the equipment should be a consideration under this design. It is telling how much push back there was from the synth whitelist when armored gloves/boots were removed from common spawns. It is a bummer to have people put friendships and cordiality on the line when "their" role may be changed in some form.

Bottom line, incentivization of gear based on stats for synthetics is antithetical to the current design.

But more importantly, someone (*cough* @realforest2001 *cough*) should really sit down and write down what synths should be and force everyone on the same page. The design of the role being entirely disconnected from the actual application of policy is counterintuitive, dumb, and creates tension between the maintainer team and the player base.

Drulikar:
This is probably about as good of an argument that can be argued for this, but as is no management is in favor of this change nor think it necessary to do so. So I think it best to stick with the current implementation. Sorry to waste your time with this but I wanted plenty of time for management to chime in (at least internally - none messaged here I guess), see the proposed changes, and the argument for it.

PR #6018, opened 2024-03-26
fix burrowed burrower should not be hit by gau

cuberound:
# About the pull request

adds check for being burrowed to direct GAU damage

burrowed burrower should not be hit by any effect, bugfix (untested, not good at testing stuff that reqires checking more mobs but should work"

<details>

</details>

# Changelog
:cl:
fix: GAU can not hit burrowed burrower 
/:cl:


Drulikar:
I do not think this change necessary. Burrower can already burrow out of the way, they don't need additional help. At a bare minimum they could take reduced damage for being underground, but again I don't think they need a buff.

PR #6020, opened 2024-03-27
Super Strength Grab Changes (Never skip arm day)

Ben10083:
# About the pull request

Grabs for those with super strength buffed

- passive grabs no longer guarantied resist
- target has reduced chance to resist
- if the target, _increased_ chance to resist
- if both have super strength, works as before (passive grabs guarantied for example)

Grabs are not often used, even though roles such as WJs are encouraged to do so. This is because grabs are very easy to be resisted out of. This is fine, and is necessary, but in this case, I want to have it so if someone has an advantage in strength, they have advantage in this exchange (both dealing and recieving).

I feel this was more an oversight, as logically those who are stronger will have a stronger grip than otherwise.

The balance out between the two is to prevent shenanigans between synth and pred, no grab fights there.

scenario 1)
Synth vs Human
Synth has strength, human doesn't
advantage synth

scenario 2)
human vs Pred
Pred has strength, human doesnt
advantage pred

scenario 3)
Synth vs Pred
Both have strength
No advantage (how it is now)

# Changelog
:cl:
balance: Mobs with super strength now have an advantage when resisting of a grab is done, both as the grabber and the one being grabbed (Xenos not affected, works as before).
balance: Mobs with super strength no longer have their passive grabs a guaranteed resist (unless both super).
/:cl:


Ben10083:
Quick exception for xenos added (warrior grabs wont be easier to resist)

code/modules/mob/living/living.dm
```diff
@@ -224,18 +224,32 @@
...
+		var/chance = 50
+		if(HAS_TRAIT(user, TRAIT_SUPER_STRONG) && !isxeno(pulledby)) // no extra chance to resist warrior grabs
+			chance += 30 // you are strong, you can overpower them easier
+		if(HAS_TRAIT(pulledby, TRAIT_SUPER_STRONG))
+			chance -= 30 // stronger grip
+		// above code means that if you are super strong, 80% chance to resist, otherwise, 20 percent. if both are super strong, standard 50.
+
+		if(prob(chance))
+			playsound(user.loc, 'sound/weapons/thudswoosh.ogg', 25, 1, 7)
+			visible_message(SPAN_DANGER("[user] has broken free of [pulledby]'s grip!"), null, null, 5)
```
SabreML, line 240: These were like this already, but I think they'd look a bit neater if the '5' used a named argument:
```suggestion
visible_message(SPAN_DANGER("[src] has broken free of [pulledby]'s grip!"), max_distance = 5)
```

code/modules/mob/living/living.dm
```diff
@@ -224,18 +224,32 @@
...
 			pulledby.stop_pulling()
-			return 1
+			return TRUE
 		if(moving_resist && client) //we resisted by trying to move
-			visible_message(SPAN_DANGER("[src] struggles to break free of [pulledby]'s grip!"), null, null, 5)
-			client.next_movement = world.time + (10*pulledby.grab_level) + client.move_delay
+			visible_message(SPAN_DANGER("[user] struggles to break free of [pulledby]'s grip!"), null, null, 5)
```
SabreML, line 244:
```suggestion
			visible_message(SPAN_DANGER("[src] struggles to break free of [pulledby]'s grip!"), max_distance = 5)
```

code/modules/mob/living/living.dm
```diff
@@ -224,18 +224,32 @@
...
+		/// Chance for person to break free of grip, defaults to 50.
+		var/chance = 50
+		if(HAS_TRAIT(user, TRAIT_SUPER_STRONG) && !isxeno(pulledby)) // no extra chance to resist warrior grabs
+			chance += 30 // you are strong, you can overpower them easier
+		if(HAS_TRAIT(pulledby, TRAIT_SUPER_STRONG))
+			chance -= 30 // stronger grip
+		// above code means that if you are super strong, 80% chance to resist, otherwise, 20 percent. if both are super strong, standard 50.
+
+		if(prob(chance))
+			playsound(user.loc, 'sound/weapons/thudswoosh.ogg', 25, 1, 7)
```
SabreML, line 239:
```suggestion
			playsound(loc, 'sound/weapons/thudswoosh.ogg', 25, 1, 7)
```

code/modules/mob/living/living.dm
```diff
@@ -224,18 +224,32 @@
...
+			visible_message(SPAN_DANGER("[user] has broken free of [pulledby]'s grip!"), null, null, 5)
 			pulledby.stop_pulling()
-			return 1
+			return TRUE
 		if(moving_resist && client) //we resisted by trying to move
-			visible_message(SPAN_DANGER("[src] struggles to break free of [pulledby]'s grip!"), null, null, 5)
-			client.next_movement = world.time + (10*pulledby.grab_level) + client.move_delay
+			visible_message(SPAN_DANGER("[user] struggles to break free of [pulledby]'s grip!"), null, null, 5)
+			// +1 delay if super strong, also done as passive grabs would have a modifier of 0 otherwise, causing spam
+			if(HAS_TRAIT(pulledby, TRAIT_SUPER_STRONG) && !HAS_TRAIT(user, TRAIT_SUPER_STRONG))
```
SabreML, line 246: Since `HAS_TRAIT(pulledby, TRAIT_SUPER_STRONG)` and `HAS_TRAIT(user, TRAIT_SUPER_STRONG)` are used multiple times in this proc, it would make sense to cache them as variables near the start. That way they only need to be called once rather than potentially three times.

For example:
```dm
	var/pulledby_is_strong = HAS_TRAIT(pulledby, TRAIT_SUPER_STRONG)
	var/src_is_strong = HAS_TRAIT(src, TRAIT_SUPER_STRONG)

	// Passive grabs will also be broken, UNLESS pulledby has super strength, but this is ignored if you also have super strength
	if(pulledby.grab_level || (pulledby_is_strong && !src_is_strong))
```

code/modules/mob/living/living.dm
```diff
@@ -224,18 +224,35 @@
...
-			playsound(src.loc, 'sound/weapons/thudswoosh.ogg', 25, 1, 7)
-			visible_message(SPAN_DANGER("[src] has broken free of [pulledby]'s grip!"), null, null, 5)
+	// vars for checks of strengh
+	var/pulledby_is_strong = HAS_TRAIT(pulledby, TRAIT_SUPER_STRONG)
+	var/src_is_strong = HAS_TRAIT(src, TRAIT_SUPER_STRONG)
+
+	// Passive grabs will also be broken, UNLESS pulledby has super strength, but this is ignored if you also have super strength
+	if(pulledby.grab_level || (pulledby_is_strong && !src_is_strong))
+		// Chance for person to break free of grip, defaults to 50.
+		var/chance = 50
```
SabreML, line 234: Thank you!

As a last thing, could you flip this so that the `stop_pulling()` part comes first? Since it returns after the proc call you can get an indentation level back for free.
I.e:
```dm
	// Passive grabs will also be broken, UNLESS pulledby has super strength, but this is ignored if you also have super strength
	if(!pulledby.grab_level && (!pulledby_is_strong || src_is_strong))
		pulledby.stop_pulling()
		return TRUE

	// Chance for person to break free of grip, defaults to 50.
	var/chance = 50
	if(src_is_strong && !isxeno(pulledby)) // no extra chance to resist warrior grabs
		chance += 30 // you are strong, you can overpower them easier
etc. etc.
```

SabreML (approved):
Code approved, still need a balance review. 👍

PR #6027, opened 2024-03-27
Removes OT mortar shells

iloveloopers:
# About the pull request
removes OT shells from the armylathe

image

OT shells are really just CAS fatties
they clear out xeno fortifications too easily and the big radius is unfun to play against

probably works

# Changelog
:cl:
balance: removed OT mortar shells from normal gameplay loop
/:cl:


deathrobotpunch:
classic Balance Experts removing fun mechanics from the game because they died to it.

Drulikar:
I don't think an outright removal is necessary for these, but you're welcome to make a PR to nerf them some to bring them more in line if you genuinely think they are performing too well.

PR #6031, opened 2024-03-27
reduces impact area from 9x9 to 7x7, lowers DPS slightly

cuberound:
# About the pull request

okey this is basicly copy of my suggested buff from more then a year ago, but now it is a nerf. Main complain on why the GAU is so OP and strong is its spreat ( that used to be the reason why it was so weak and shit). Therfore, lets do some math. Current situation is that the GAU fires 80 rounds per burst at area of 81 tiles( 9x9) ~0,988 round per tile, the proposed change aims the keep the damage on anyone standing in the impact area about the same ( it is a bit lower might need compensation with bringing the damage to pre double bullet values) , by bringing down the impact area to 49 tiles (7x7) and lowering the number of shots fired to 40, we get value of ~0,816 round per tile. meanwhile it keeps the rest like how long the bursts last the just as long.
(https://github.com/cmss13-devs/cmss13/pull/4302 the old "buff")

Making two bullets per bullet was not a bad idea in general, it brought the DPS on competetive level with other weapons, but the huge spread leads to too big compared to other ordinance. Bringing it in line with other ordinance should allow for easier CAS balance around smaller impact FMs.

<details>

</details>

# Changelog
:cl:
balance: reduced GAU spread from 9x9 to 7x7
balance: reduced GAU shots fired per shot back to 1 (bullets per tile from 0.99 to 0.82)
/:cl:

### DIFF
--- a/code/modules/cm_marines/dropship_ammo.dm
+++ b/code/modules/cm_marines/dropship_ammo.dm
@@ -144,7 +144,7 @@
 	ammo_used_per_firing = 40
 	point_cost = 275
 	fire_mission_delay = 2
-	var/bullet_spread_range = 4 //how far from the real impact turf can bullets land
+	var/bullet_spread_range = 3 //how far from the real impact turf can bullets land
 	var/shrapnel_type = /datum/ammo/bullet/shrapnel/gau //For siming 30mm bullet impacts.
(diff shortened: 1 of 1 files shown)


Drulikar (approved):
In general fire missions likely need to be overall buffed, but if you feel this is just superior to other options we can roll with this change and further explore changes that more globally affect firemissions.

PR #6035, opened 2024-03-28
serious incen mini rework, reduces explosion strenght and price

cuberound:
# About the pull request

reduces explosion about 1/3, justifing the incendiary payload in casing of the same size, reduces price as it might be a bit stronger then normal minirocket but not streight up stronger with no other drawback.
old explosion values based on range are
200, 156, 112, 68, 24
new one
130, 100, 70, 40, 10

the numbers are up for debate, but the concept is hopefully acceptable

(Others have said they would reduce the incen mini explosion when I have been asking long ago but welp...)

puts incen minirockets into its own position, weaker explosion the minirocket and weaker fire then laser but has both. Having super strong option behind what is basicly paywall is weard and this should allow for them to be more viable while getting it line with other ordinance.

<details>

</details>

# Changelog
:cl:
balance: reduces incen minirocket strength from 200 to 130
balance: reduces incen minirocket falloff from 44 to 30
balance: reduces incen minirocket price from 500 to 350
/:cl:

### DIFF
--- a/code/modules/cm_marines/dropship_ammo.dm
+++ b/code/modules/cm_marines/dropship_ammo.dm
@@ -382,13 +382,13 @@
 	transferable_ammo = TRUE
 	point_cost = 300
 	fire_mission_delay = 3 //high cooldown
-	explosion_strenght = 200
+	explosion_strength = 200
 	explosion_falloff = 44
 
 /obj/structure/ship_ammo/minirocket/detonate_on(turf/impact, obj/structure/dropship_equipment/weapon/fired_from)
 	impact.ceiling_debris_check(2)
 	spawn(5)
-		cell_explosion(impact, explosion_strenght, explosion_falloff, EXPLOSION_FALLOFF_SHAPE_LINEAR, null, create_cause_data(initial(name), source_mob))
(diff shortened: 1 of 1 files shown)


code/modules/cm_marines/dropship_ammo.dm
```diff
@@ -410,8 +412,9 @@
 	name = "\improper AGR-59-I 'Mini-Mike'"
 	desc = "The AGR-59-I 'Mini-Mike' incendiary minirocket is a cheap and efficient means of putting hate down range AND setting them on fire! Though rockets lack a guidance package, it makes up for it in ammunition count. Can be loaded into the LAU-229 Rocket Pod."
 	icon_state = "minirocket_inc"
-	point_cost = 500
-	fire_mission_delay = 3 //high cooldown
+	point_cost = 350
+	explosion_strength = 130
+	explosion_falloff = 30
```
Drulikar, line 417: I don't think it needs to change this much.

Maybe 400 pts, same falloff, and 150 strength?

But as I said in your other PR in general fire missions probably overall need to be more powerful. Only if you feel this is exceptionally better than the others is any change necessary.

Drulikar (changes requested):

Issue #5967, opened 2024-03-17
Huggers can miss on spriteclick pounce while adjacent if the player moves


GrrrKitten:
### Testmerges

Server revision compiled on: 2024-03-17T03:24:36+00:00 Master commit: 9b6f711d57dfc090b87670d6b18ebf81f78803bd The following pull requests are currently test merged: #5724: 'XM43E1 in regular gameplay' by Releasethesea at commit d6f45c811a #5763: 'Map: Bigred change south lambda padlock and turned it into a checkpoint.' by Huffie56 at commit 646c47711a #5829: 'Event borers' by realforest2001 at commit dee964dd25 #5950: 'Med Skill Rebalance Mk2 Demon Free Edition' by Katskan at commit 9f4b9db6bb #5006: 'Project ARES: Added Tech Unlock logs' by realforest2001 at commit c0645ab04c #5401: 'M540-A Armored Recon Carrier' by Zonespace27 at commit 9be4807644 #5925: 'Biosuit Updates' by realforest2001 at commit bf9853d96c #5946: 'Autocompressor Efficiency Increase' by Katskan at commit 93e4247706  Local commit: 9e64ad44a13f4f71e7168dc2f3c323ffb2d1432b TGS version: 6.3.1 DMAPI version: 5.9.0

### Round ID

21351 and 20807 (and a few others I dont have videos of)

### Description of the bug

Basically, if you are on an adjacent tile and ability pounce on a marine, it should hug them, but there is a hard to replicate bug where if they move the exact tick you go to hug them, the hug just doesn't go through and you go back on the ground

### What's the difference with what should have happened?

you should hug da marine!!!

### How do we reproduce this bug?

1.be adjacent to marine
2.sprite click pounce to hug on them
3.have them move the exact tick you go to hug them
4. see bug
...

https://github.com/cmss13-devs/cmss13/assets/158611449/ebb1ceef-801e-4ef7-8638-713a525f6833

https://github.com/cmss13-devs/cmss13/assets/158611449/89a1926b-348d-4fd7-9ce2-57053a792621

### Issue Bingo

- [ ] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #6040, opened 2024-03-29
Better throw logic

Git-Nivrak:
# About the pull request
Throw will now attempt to reach the target rather than the a predefined path that could change if thrown towards an object

Tested on local but this affect a lot of stuff so needs to be tested

When you try to throw something and the target slightly moves one nanosecond after you throw it sucks
<details>

</details>

# Changelog
:cl:
add: Throws will now attempt to reach their target rather than the turf they were on when thrown
/:cl:

### DIFF
--- a/code/modules/movement/launching/launching.dm
+++ b/code/modules/movement/launching/launching.dm
@@ -182,7 +182,6 @@
 
 	add_temp_pass_flags(pass_flags)
 
-	var/turf/start_turf = get_step_towards(src, LM.target)
 	var/last_loc = loc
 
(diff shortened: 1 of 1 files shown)


SabreML:
This *may* fix https://github.com/cmss13-devs/cmss13/issues/5967, but that's a hard one to reproduce.

Hopekz:
I like throwing missing because dodging people's throws is a mechanic in itself imo vs being like a homing rocket like whenever the guy from hitman throws an item.

Drulikar (approved):

PR #6042, opened 2024-03-29
Adds a white eyepatch

MOONPRINCESS17:
# About the pull request

Adds a white counterpart to the already existing eyepatch, which is black. This one appears exclusively in player loadouts and does not replace the classic eyepatch anywhere else.

For those of discerning fashion taste. That, or any possible... gimmicks such fashion might enable.

image
# Changelog
:cl:
add: Added a white eyepatch to player loadouts.
imageadd: Added sprites for a white eyepatch.
/:cl:


code/modules/clothing/glasses/glasses.dm
```diff
@@ -194,6 +194,15 @@
 	flags_armor_protection = 0
 	flags_equip_slot = SLOT_EYES|SLOT_FACE
 
+/obj/item/clothing/glasses/eyepatch_white
```
Zonespace27, line 197: make this a subtype of /obj/item/clothing/glasses/eyepatch

sleepynecrons (changes requested):
the whites are ridiculously bright. use colors from something like PMC gear if you're going to make white clothing
also is this not just you trying to add anime larp

MOONPRINCESS17:
The item is now a subtype of the normal black eyepatch and the sprite is not as luridly bright.

code/modules/clothing/glasses/glasses.dm
MOONPRINCESS17: Done

sleepynecrons:
after speaking with other maintainers i will be closing the PR. a white eyepatch isn't something we want marines wearing as it's visually inconsistent with the faction. white is something we should keep reserved to PMCs/W-Y specific groups. the item is also just a reference to https://myanimelist.net/character/65865/Rikka_Takanashi

PR #6053, opened 2024-03-30
Makes acid spray extinguish more consistent

Git-Nivrak:
# About the pull request
Acid spray used to only extinguish you if you were on the same tile when it is initialized rather than when crossing it as well, this PR makes it so it extinguishes you when crossing it as well.

Extremely weird and unintuitive when you walk over 3 tiles of acid spray and don't get extinguished.
<details>

</details>

# Changelog
:cl:
qol: Acid sprays will now extinguish fire from other xenomorphs more consistently
/:cl:

### DIFF
--- a/code/game/objects/effects/aliens.dm
+++ b/code/game/objects/effects/aliens.dm
@@ -153,6 +153,8 @@
 		var/mob/living/carbon/xenomorph/X = AM
 		if (X.hivenumber != hivenumber)
 			apply_spray(AM)
+		else
+			X.ExtinguishMob()
 	else if(isVehicleMultitile(AM))
 		var/obj/vehicle/multitile/V = AM
 		V.handle_acidic_environment(src)
--- a/code/game/objects/effects/aliens.dm
+++ b/code/game/objects/effects/aliens.dm
@@ -145,16 +145,11 @@
 /obj/effect/xenomorph/spray/Crossed(AM as mob|obj)
 	..()
 	if(ishuman(AM))
-		var/mob/living/carbon/human/H = AM
-		if(H.ally_of_hivenumber(hivenumber))
-			return
-		apply_spray(AM)
-	else if (isxeno(AM))
-		var/mob/living/carbon/xenomorph/X = AM
-		if (X.hivenumber != hivenumber)
-			apply_spray(AM)
+		var/mob/living/living_mob = AM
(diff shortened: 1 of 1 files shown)


InsaneRed:
this is also a balance change

code/game/objects/effects/aliens.dm
```diff
@@ -153,6 +153,8 @@
 		var/mob/living/carbon/xenomorph/X = AM
 		if (X.hivenumber != hivenumber)
 			apply_spray(AM)
+		else
+			X.ExtinguishMob()
```
SabreML, line 157: The acid spray extinguishes any `/living` subtype if it hits someone in `Initialize()`, so it would make sense for this to do the same. (Currently it'll only extinguish xenomorphs)
https://github.com/cmss13-devs/cmss13/blob/b5e180c53acbca0fcc195e9b9d13a7c42b3f4a92/code/game/objects/effects/aliens.dm#L107-L109

Looking into it a bit, this proc could actually be refactored and shortened quite a bit since the `ally_of_hivenumber()` proc is on `/living`, and contains a `xeno.hivenumber != hivenumber` check for xenos as well.
Something like this would work:
```dm
/obj/effect/xenomorph/spray/Crossed(AM as mob|obj)
	..()
	if(isliving(AM))
		var/mob/living/living_mob = AM
		if(living_mob.ally_of_hivenumber(hivenumber))
			living_mob.ExtinguishMob()
		else
			apply_spray(living_mob)
	else if(isVehicleMultitile(AM))
		var/obj/vehicle/multitile/V = AM
		V.handle_acidic_environment(src)
```

Git-Nivrak: Good point

code/game/objects/effects/aliens.dm
```diff
@@ -145,14 +145,11 @@
 /obj/effect/xenomorph/spray/Crossed(AM as mob|obj)
 	..()
 	if(ishuman(AM))
```
SabreML, line 147: This bit too:
```suggestion
	if(isliving(AM))
```

SabreML (changes requested):
> this is also a balance change

I'll give this a day or so once it's approved in case anyone else wants to weigh in, but I would personally say it's more on the QOL side of things than balance. At most, a burning xeno would have 1 second to step into the acid path before it dissipates.

SabreML (approved):

PR #5940, opened 2024-03-14
you can now use the mortar during hijack if you are not a coward + custom shell warning for ghosts

iloveloopers:
# About the pull request
lets you deploy the mortar shipside.
lets you fire a shell on a shipside mortar during hijack but do not expect to live.

ghosts now get a warning when an OT shell is about to land. (I did not test this and I don't know how to, I just copy pasted from CAS fire missions)

# Explain why it's good for the game
its funny and cool and soul

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

https://github.com/cmss13-devs/cmss13/assets/140007537/c7f7dcb2-98d1-438c-ba00-6a57f5ad9463

https://github.com/cmss13-devs/cmss13/assets/140007537/8862d7e1-cda9-4fda-9c2b-7b5c105fe3e0

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
balance: You can now use the mortar while shipside during hijack, but only once.
add: Ghosts now get a warning when a custom mortar shell is about to land.
/:cl:

PR #6054, opened 2024-03-30
Fix for firing mortar.

Segrain:
# About the pull request

Followup to #5940.
I understand the newfashioned crusade against old singleletter variables, but when engaging in it, can we please ensure that all the renamed variables stay effectively the same with no wires crossed? Or at least test our code?

Is fix.

# Changelog
:cl:
fix: Mortar works properly once again.
/:cl:

### DIFF
--- a/code/modules/cm_marines/equipment/mortar/mortars.dm
+++ b/code/modules/cm_marines/equipment/mortar/mortars.dm
@@ -245,10 +245,10 @@
 			else
 				to_chat(user, SPAN_RED("You realize how bad of an idea this is and quickly stop."))
 				return
-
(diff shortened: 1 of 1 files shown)


iloveloopers:
I blame zonespace for forcing me to fix the one letter vars

SabreML (approved):

SabreML:
Which bit was causing it btw?

Segrain:
> Which bit was causing it btw?

image
Used to be `T` from way back in the procedure for the target turf and `G` set in the end to make shards rain on mortar if fired under glass roof, accidentally fed that mortar turf into shell's targeting instead.

PR #6056, opened 2024-03-31
[s] Temporary client image fix 

Drulikar:
# About the pull request

Temporary measure until a more permanent solution is found.

Should be fixed in 515.1635


Drulikar:
https://www.byond.com/forum/post/2919246

Drulikar:
Confirmed fixed in 515.1635

PR #6057, opened 2024-03-31
Adds 2 new flamer fuels for OT/Chemistry to make.

iloveloopers:
# About the pull request
adds 2 new flamer fuels for OT/Chem to create and distribute to marines, sticky-napalm and high-combustion napalm
sticky napalm has 15 intensity 50 duration (max firestacks), made with welding fuel and napalm. designed for suppression and backliner fighting
high combustion has 45 intensity 10 duration, made with CLF3 and napalm. designed for high burst damage
for reference pyro's blueflame has 40 intensity 40 duration
CLF3 recipe now gives you 3 units instead of 1 to not be so straining on energy
these new flamer chems won't impact fire stats for OT casings

more flamer fuel variety good
these are sidegrades not direct upgrades, they are each good in certain situations while normal fuel is a jack of all trades

https://github.com/cmss13-devs/cmss13/assets/140007537/182093a6-2a5d-47ed-ab92-6c044fe66d0f

# Changelog
:cl: Tim Harley cookpet
add: Added 2 new flamer fuels craftable by OTs/Chemistry
balance: CLF3 recipe now gives more units per reaction
/:cl:


code/modules/projectiles/magazines/flamer.dm
```diff
@@ -17,9 +17,9 @@
 	var/flamer_chem = "utnapthal"
 	flags_magazine = AMMUNITION_HIDE_AMMO
 
-	var/max_intensity = 40
+	var/max_intensity = 50
 	var/max_range = 5
-	var/max_duration = 30
+	var/max_duration = 50
```
Drulikar, line 22: I don't think these values need to change. They contribute to the differences between all subtypes of `/obj/item/ammo_magazine/flamer_tank` and as such will affect likely more than you want to affect.

Drulikar (changes requested):

code/modules/projectiles/magazines/flamer.dm
Drulikar: Still unresolved. Welcome to make a tank subtype that has more lenient limits but it shouldn't be super easy to acquire.

iloveloopers: fair enough, I'll make something when the research rework comes out. Can it get merged as is with just the fuels?

Drulikar (approved):

PR #6058, opened 2024-03-31
Cryo Tube Rework

Drulikar:
# About the pull request

This PR reworks the cryotubes. The effect of diluting or multiplying chemicals is no longer possible, but now the use of cryoxadone or clonexadone in a cryotube can allow chemicals to process in the dead. Ultimately usage of the cryo chemicals will be much higher, and you need to be careful not to introduce too many chemicals into your patient because the dosage is 5u whenever cryoxadone or clonexadone is available and can be administered.

Cryotubes also now actually require power, update their icon if they lose power, and use the two-ping sfx for warnings (such as now warnings as the patient's revive timer is running out). For the critical messages, it will make an alert if the patient is dead, then once they hit the 50% mark (2.5 mins), then at the 1 minute mark.

See testing for a demonstration.

Cryotubes in general are very underused with the exception of exploiting how it can multiply chemicals. This should give them a unique feature to medbay.

It is likely that this may prove to be too powerful, but it puts more demand on chemistry to provide the cryoxadone or clonexadone in order to do so.

<details>

Initial testing:
https://youtu.be/PA43YvzDRVM

Critical (dead state) alert testing:
https://youtu.be/PLjQxzv28l8

Notification changes:
release

Power fixes:
power

</details>

# Changelog
:cl: Drathek
balance: Cryotubes no longer multiply chemicals.
balance: Cryotubes can now allow chemicals to process in dead if cryoxadone or clonexadone is present.
fix: Fixed cryotubes not actually updating their icons/power usage when they become unpowered.
fix: Fixed cryotubes healing despite inoperable.
fix: The cryotube Eject Occupant verb now works for people other than the occupant (useful if inoperable)
ui: Cryotubes can optionally announce again on the medical channel.
/:cl:

### DIFF
--- a/code/game/machinery/cryo.dm
+++ b/code/game/machinery/cryo.dm
@@ -299,21 +299,20 @@
 	set category = "Object"
 	set src in oview(1)
 	if(usr == occupant)//If the user is inside the tube...
-		if(usr.stat == 2)//and he's not dead....
+		if(usr.stat == DEAD)//and he's not dead....
 			return
 
-		if(alert(usr, "Would you like to activate the ejection sequence of the cryo cell? Healing may be in progress.", "Confirm", "Yes", "No") == "Yes")
+		if(tgui_alert(usr, "Would you like to activate the ejection sequence of the cryo cell? Healing may be in progress.", "Confirm", list("Yes", "No")) == "Yes")
 			to_chat(usr, SPAN_NOTICE("Cryo cell release sequence activated. This will take thirty seconds."))
-			visible_message(SPAN_WARNING ("The cryo cell's tank starts draining as its ejection lights blare!"))
+			visible_message(SPAN_WARNING("The cryo cell's tank starts draining as its ejection lights blare!"))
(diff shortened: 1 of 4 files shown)


code/game/machinery/cryo.dm
```diff
@@ -202,22 +233,39 @@
...
+
+	if(autoeject)
+		//release the patient automatically when brute and burn are handled on non-robotic limbs
+		if(!occupant.getBruteLoss(TRUE) && !occupant.getFireLoss(TRUE) && !occupant.getCloneLoss())
+			display_message("Patient's external wounds are healed.")
 			go_out(TRUE)
 			return
-		if(occupant.health >= 100 && autoeject)
-			display_message("external wounds are")
+		if(occupant.health >= 100)
```
Diegoflores31, line 263: i think it would be wiser to use occupant.health >= occupant.maxhealth JUST in case.

harryob (approved):

PR #6065, opened 2024-04-01
Toggle ghost now leaves the users mob visible

### LINKED ISSUE #5894: Ghost Vision & Self-Visibility borked
### Testmerges

Irrelevant

### Round ID

_No response_

### Description of the bug

Toggling Ghost Vision also makes your ghost invisible. Toggling Self-visibility does not work and the button sends no messages

Ghost Vision should keep your ghost and self visibility should actually work

Toggle Ghost Vision
Toggle Self Visibility
_Enlightenment_

### Issue Bingo

vero5123:
# About the pull request
Fixes #5894

bug fix

# Changelog
:cl:
fix: toggling ghost vision now keeps the user's mob visible.
/:cl:

### DIFF
--- a/code/__DEFINES/__game.dm
+++ b/code/__DEFINES/__game.dm
@@ -78,6 +78,8 @@
 #define SEE_INVISIBLE_LEVEL_TWO 45 //Used by some other stuff in code. It's really poorly organized.
 #define INVISIBILITY_LEVEL_TWO 45 //Used by some other stuff in code. It's really poorly organized.
 
+#define HIDE_INVISIBLE_OBSERVER 59 // define for when we want to hide all observer mobs.
+
 #define INVISIBILITY_OBSERVER 60
 #define SEE_INVISIBLE_OBSERVER 60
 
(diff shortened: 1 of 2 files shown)


fira:
I'm not sure removing ghost plane handling is much of a fix

Blundir:
This pr isn't exactly optimal feature-wise, seeing no ghosts at all is good for making screenshots both locally and on server itself, this pr just completely removes the ability, without providing alternatives.

vero5123:
> This pr isn't exactly optimal feature-wise, seeing no ghosts at all is good for making screenshots both locally and on server itself, this pr just completely removes the ability, without providing alternatives.

No it doesn't, you toggle ghost visibility, which hides all mobs but yourself and then you toggle self visibility which hides youe mob from your pov and others.

PowerfulBacon (commented):
Are there things on the ghost layer that would be inadvertantly hidden by changing the see_invisible flags?

vero5123:
> Are there things on the ghost layer that would be inadvertantly hidden by changing the see_invisible flags?

Afik, No. Just the ghosts.

Drulikar (approved):

PR #6070, opened 2024-04-02
'Search by key' button for the Keybind menu

SabreML:
# About the pull request

Adds a 'search by key' button to the Keybind menu, which scrolls the window to the first entry bound to the inputted key.

Ideally this would somehow loop through everything that the key is bound to, but I haven't been able to figure that out yet.

It's easy to search for the name of a keybind if you already know what you're looking for, but if for example you're trying to figure out what the 'F' key is doing, or whether or not a key you want to bind is already in use, then currently you need to look through the whole list manually.

<details open>

https://github.com/cmss13-devs/cmss13/assets/57483089/791a6855-1b23-40fd-892b-873032956ec4

</details>

# Changelog
:cl:
ui: Added a 'search by key' button to the Keybind menu.
/:cl:

### DIFF
--- a/code/modules/client/tgui_macro.dm
+++ b/code/modules/client/tgui_macro.dm
@@ -23,11 +23,12 @@ GLOBAL_LIST_EMPTY(ui_data_keybindings)
 
 /datum/tgui_macro/ui_data(mob/user)
 	. = list()
-	.["keybinds"] = prefs.key_bindings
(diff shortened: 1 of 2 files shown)


harryob (approved):
looks and works fine - but we might want to add a tooltip to the button to make it a little more intuitive for what it does

PR #5796, opened 2024-02-23
Nightmare Insert Fixes & Elevated Nightmare Errors

Drulikar:
# About the pull request

This PR elevates nightmare insert errors to a runtime so they fail checks and corrects errors in their configuration that was causing the errors in the first place.

# Explain why it's good for the game

Various nightmares are not loading because of issues with their configuration; this should restore various nightmares.

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl: Drathek
fix: Fixed various nightmare inserts that were not spawning.
code: Nightmare errors are now a runtime.
/:cl:

PR #6072, opened 2024-04-03
Small fix to Nightmare documentation example

TopHatPenguin:
# About the pull request
Fixes the documentation regarding the nightmare landmarks example. Viro_open which is presented as the example, fails to inform you that the text after the _ will not be used by the landmark and is simply additional info for the viewer. This has now been rectified with this PR.

# Changelog
:cl:
fix: Fixes Nightmare documentation example.
/:cl:

### DIFF
--- a/code/modules/nightmare/nmnodes/mapload.dm
+++ b/code/modules/nightmare/nmnodes/mapload.dm
@@ -78,7 +78,15 @@
 			context.add_task(task)
 			break
 
-
+/**
+ * Similar to variations mode, but rolls all files individually rather
+ * than picking one, using name for landmark. The prefix number is used
+ * as a percentage chance.
+ *
+ * Example:
+ *   some/folder/10.something_funny.dmm
+ * would have 10% chance to insert at the 'something_funny' landmark
+ */
 /datum/nmnode/mapload/sprinkles
 	id = "map_sprinkle"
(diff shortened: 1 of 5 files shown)


Drulikar:
Are you sure about that after https://github.com/cmss13-devs/cmss13/pull/5796 ?

code/modules/nightmare/nmnodes/mapload.dm
```diff
@@ -78,15 +78,7 @@
...
-/**
- * Similar to variations mode, but rolls all files individually rather
- * than picking one, using name for landmark. The prefix number is used
- * as a percentage chance. You can add extra text with an underscore.
- *
- * Example:
- *   some/folder/10.something_funny.dmm
- * would have 10% chance to insert at 'something' landmark
- */
+
```
Drulikar, line 81:
```suggestion
/**
 * Similar to variations mode, but rolls all files individually rather
 * than picking one, using name for landmark. The prefix number is used
 * as a percentage chance.
 *
 * Example:
 *   some/folder/10.something_funny.dmm
 * would have 10% chance to insert at the 'something_funny' landmark
 */
```

Drulikar (changes requested):

Drulikar (approved):

PR #6094, opened 2024-04-06
Deconstructing walls without repairing them

HumiliatedGoblin:
# About the pull request

You now need to have disarm or above to slice through walls, and help intent to repair walls.

Comtechs now don't need to spend 30 seconds repairing a wall, waste like 5 units of welding fuel only to deconstruct it. Now you can have choice in what you do! :D

<details>

it works trust

</details>

# Changelog

:cl:
qol: Now needs help intent to repair walls
qol: Now needs disarm, grab and harm intent to deconstruct walls
/:cl:


code/game/turfs/walls/walls.dm
```diff
@@ -482,6 +482,8 @@
 /turf/closed/wall/proc/try_weldingtool_usage(obj/item/W, mob/user)
 	if(!damage || !iswelder(W))
 		return FALSE
+	if(!(user.a_intent == INTENT_HELP))
```
ihatethisengine, line 485:
```suggestion
	if(user.a_intent != INTENT_HELP)
```

harryob (commented):
we might need some text adding to the examine strings to make this more obvious beyond a changelog entry - thoughts?

HumiliatedGoblin:
> we might need some text adding to the examine strings to make this more obvious beyond a changelog entry - thoughts?

Do you think it would be better to have info on the wall, or on the blowtorch? Or both?
My idea is to just put it underneath the weld_state string, but maybe you'd have a better idea

HumiliatedGoblin:
@harryob  HARRY PLEASE

Drulikar (approved):

PR #6104, opened 2024-04-07
Hedgehog explosion immunity rebalance

InsaneRed:
# About the pull request
Hedgehog now only has explosion stun immunity if you're above half shard rate.

Hedgehog is arguably one of the safest castes right now, always has grenade immunity, no longer has a slowdown and an increasing armor rate with its spikes and an invincibility button.

The immunity was given due to the slowdown it naturally had, but recent changes have touched that without removing or touching up its immunity, this tries to fix that.

<details>

</details>

# Changelog
:cl:
balance: Hedgehog now gains explosion immunity to nades if you have above half shard amount.
/:cl:


morrowwolf:
Hey look it's my change just missing the sprite changes that make this so you can interact with this change as a marine.

Birdtalon (dismissed):

ihatethisengine:
I don't think this will solve anything at all. Hedge is most effective when at full shards (because you get 40 armor), so this PR just gives the hedge +5 slash damage when at strongest and removes explosion resistance when at weakest. But why would anyone fight when at weakest? 

If you wanna address actual issues with hedge:
1) Look at the shield CD, it's comically low. You can use the shield twice in the same engagement.
2) Look up the shards damage. Sometimes they just one shot people (probably if all shards hit? I dunno).
3) Maybe look at shard shed, currently it's a noob trap and rarely worth the downtime. Maybe make it weaker and remove the 15 seconds no shards? You already need to spend some time to regen shards/hp, why there is even more downtime?

GoldenDarkness55:
Just apply the slowdown changes morrow pr'd a while ago and +1s on shield cd or roll back duration to 2s.
Stop adding even more convoluted mechanics, the average player still doesn't know fire interacts with zerker rav, how do you expect anyone to keep track of this. Not to mention it would make shard shed an useless noob bait ability

Touching slash damage is a bad idea because it gets too much unnecessary 1v1 power, it really does not need more damage.

code/modules/mob/living/carbon/xenomorph/strains/castes/ravager/hedgehog.dm
```diff
@@ -103,6 +102,15 @@
 	var/percentage_shards = round((shards / max_shards) * 100, 10)
 	if(percentage_shards)
 		holder.overlays += image('icons/mob/hud/hud.dmi', "xenoenergy[percentage_shards]")
+
+	if(percentage_shards >= 50)
+		bound_xeno.small_explosives_stun = FALSE
+		bound_xeno.remove_filter("hedge_unstunnable", 1, list("type" = "outline", "color" = "#421313", "size" = 1))
+	if(percentage_shards < 50)
+		bound_xeno.small_explosives_stun = TRUE
+		bound_xeno.add_filter("hedge_unstunnable", 1, list("type" = "outline", "color" = "#421313", "size" = 1))
```
Drulikar, line 111: Don't we generally apply outlines when some ability is active? I think we should probably reverse this where if you are unstunnable you have an outline, otherwise you are normal.

code/modules/mob/living/carbon/xenomorph/strains/castes/ravager/hedgehog.dm
```diff
@@ -19,9 +19,9 @@
 
 /datum/xeno_strain/hedgehog/apply_strain(mob/living/carbon/xenomorph/ravager/ravager)
 	ravager.plasma_max = 0
-	ravager.small_explosives_stun = FALSE
+	ravager.small_explosives_stun = TRUE
 	ravager.explosivearmor_modifier += XENO_EXPOSIVEARMOR_MOD_SMALL
-	ravager.damage_modifier -= XENO_DAMAGE_MOD_SMALL
+	ravager.damage_modifier -= XENO_DAMAGE_MOD_VERY_SMALL
```
Drulikar, line 24: I don't think this is necessary to change right now. But we can consider it in the future.

code/modules/mob/living/carbon/xenomorph/strains/castes/ravager/hedgehog.dm
```diff
@@ -1,6 +1,6 @@
...
+	description = "You lose your empower, charge, scissor cut and some slash damage, for a bit more explosive resistance, your explosion immunity scales with your shard count if you have more then half you get explosion immunity but if you have below half you lose it, and you gain several new abilities that allow you to become a spiky tank. You build up shards internally over time and also when taking damage that increase your armor's resilience. You can use these shards to power three new abilities: Spike Shield, which gives you a temporary shield that spits bone shards around you when damaged, Fire Spikes, which launches spikes at your target that slows them and does extra damage if they move, and finally, Spike Shed, which launches spikes all around yourself and gives you a temporary speed boost as an escape plan at the cost of all your stored shards and being unable to gain shards for thirty seconds."
```
Drulikar, line 3: This needs a lot more full stops and is particularly wordy.
```suggestion
	description = "You lose your empower, charge, scissor cut, and some slash damage in exchange for more explosive resistance. Your explosion immunity scales with your shard count and at 50% grants you resistance to some explosive stuns. You build up shards internally over time and when taking damage. You can use these shards to power three new abilities: Spike Shield which gives you a temporary shield that spits bone shards around you when damaged; Fire Spikes which launches spikes at your target to slow them and deal damage when they move; and Spike Shed which launches all your spikes, grants a temporary speed boost, and disables shard generation for thirty seconds."
```

Drulikar (changes requested):

Drulikar (approved):

PR #6108, opened 2024-04-07
Increases berserker's rage lockout is back to 10 seconds

usnpeepoo:
# About the pull request

This PR reverts the rage lockout duration of Berserker back to 10 seconds. 

Should help with punishing berserkers on their downtime, without decreasing their effectiveness

<details>

</details>

# Changelog
:cl:
balance: Berserker's rage lock-out duration increased slightly
/:cl:


Pyrodes:
I think this is just going to make it unnecessarily hard for zerk to fulfill the close-combat niche of its caste. Zerk already suffers with its health a decent amount when moving in to engage for the rage it needs for the best parts of its kit, and less armor will make it difficult to use its full kit efficiently due to having a smaller window where it can do damage before needing to disengage if it doesn't want to get shredded on the way back. I also don't see why zerk should be brought into line with base ravager's pacing without looking to see if base rav is balanced itself. Where is the point of reference upon where all the balance should be set when so many things are being tweaked one way or another so often?

GoldenDarkness55:
>still gets cucked by fire

lol
leave underpowered strains alone. NPCs with normal ammo don't do anything in general and thats an issue with the ammo system not with zerker armor

Drulikar (changes requested):
I think just the cooldown duration is fine; everything else we can discard.

usnpeepoo:
> I think just the cooldown duration is fine; everything else we can discard.

done

Drulikar (approved):

PR #6109, opened 2024-04-07
Warriors punch now slows predators, damage made a little more consistent

usnpeepoo:
# About the pull request

Warriors punch now slows predators, 1 second less than against humans. The base damage is also now the same as it is against synths, but with predator armor this means it's consistently 4-6 brute instead of less.

Completely useless ability in fights against preds, only serves to immediately let go of the pred after a lunge and avoid being hit. It's damage is usually worse then straight up slashing and it's also inconsistent that it will slow synths/humans but not predators because the if check only checks for humans (which includes synths) but not predators. The slow is a second shorter, and is a normal slow (not superslow) so it shouldn't be egregious in warrior versus pred fights.

<details>

</details>

# Changelog

:cl:
balance: Warriors punch will slow Predators, base damage increased slightly
/:cl:


Drulikar:
I don't think this change is necessary at this time.

PR #6115, opened 2024-04-08
More features and tweaks for broadcasting

ihatethisengine:
# About the pull request
Turned out that through cameras you can only see speech that you already see without cameras. So I quickly found a way to fix it with the least (for me) amount of crap code.

That's pretty much the point of broadcasting.

<details>

Screenshot_2024-04-10_01-35-26
Screenshot_2024-04-10_01-56-08
Screenshot_2024-04-10_04-14-50

</details>

# Changelog
:cl: ihatethisengine
add: Combat Correspondent can broadcast speech and emotes.
add: Speech close enough to the camera will be shown above connected TVs as abovehead messages
/:cl:


code/game/machinery/computer/camera_console.dm
```diff
@@ -219,32 +219,46 @@
...
-	if (current.c_tag == broadcastingcamera.get_broadcast_name())
+	if(current.c_tag == broadcastingcamera.get_broadcast_name())
 		current = broadcastingcamera.linked_cam
+		broadcastingcamera.viewers += resolve_viewers(concurrent_users)
 		SEND_SIGNAL(src, COMSIG_CAMERA_SET_TARGET, broadcastingcamera.linked_cam, broadcastingcamera.linked_cam.view_range, broadcastingcamera.linked_cam.view_range)
 
+/obj/structure/machinery/computer/cameras/wooden_tv/broadcast/proc/resolve_viewers(list/ref_users)
+	var/list/actual_users
+	for(var/datum/weakref/user_ref in ref_users)
+		actual_users += user_ref.resolve()
```
fira, line 259: you'll want to check this resolves into a valid ref before using it

also there's a reason this uses weakrefs - undoing it by converting these to hard refs isn't really the intended way

code/game/machinery/computer/camera_console.dm
```diff
@@ -219,32 +219,46 @@
...
 /obj/structure/machinery/computer/cameras/wooden_tv/broadcast/proc/go_back_live(obj/item/device/camera/broadcasting/broadcastingcamera)
 	SIGNAL_HANDLER
-	if (current.c_tag == broadcastingcamera.get_broadcast_name())
+	if(current.c_tag == broadcastingcamera.get_broadcast_name())
 		current = broadcastingcamera.linked_cam
+		broadcastingcamera.viewers += resolve_viewers(concurrent_users)
 		SEND_SIGNAL(src, COMSIG_CAMERA_SET_TARGET, broadcastingcamera.linked_cam, broadcastingcamera.linked_cam.view_range, broadcastingcamera.linked_cam.view_range)
 
+/obj/structure/machinery/computer/cameras/wooden_tv/broadcast/proc/resolve_viewers(list/ref_users)
+	var/list/actual_users
```
fira, line 257: are you sure that proc works? the list isn't initialized

fira (changes requested):

code/game/machinery/computer/camera_console.dm
```diff
@@ -219,32 +222,55 @@
...
+		var/mob/user = user_ref?.resolve()
+		if(istype(user) && user.client && user.client.prefs && !user.client.prefs.lang_chat_disabled && !user.ear_deaf && user.say_understands(sourcemob, language))
+			sourcemob.langchat_display_image(user)
+
+/obj/structure/machinery/computer/cameras/wooden_tv/broadcast/proc/transfer_emote(obj/item/camera, mob/living/sourcemob, emote, audible = FALSE, show_message_above_tv = FALSE)
+	SIGNAL_HANDLER
+	if(show_message_above_tv)
+		langchat_speech(emote, get_mobs_in_view(7, src), null, null, TRUE, LANGCHAT_FAST_POP, list("emote"))
+	for(var/datum/weakref/user_ref in concurrent_users)
+		var/mob/user = user_ref?.resolve()
```
Drulikar, line 270:
```suggestion
	for(var/datum/weakref/user_ref in concurrent_users)
		var/mob/user = user_ref.resolve()
```
Since you aren't using `as anything` the `?` usage is unnecessary.

code/game/machinery/computer/camera_console.dm
```diff
@@ -219,32 +222,55 @@
...
+	if(current.c_tag == broadcastingcamera.get_broadcast_name())
 		current = broadcastingcamera.linked_cam
 		SEND_SIGNAL(src, COMSIG_CAMERA_SET_TARGET, broadcastingcamera.linked_cam, broadcastingcamera.linked_cam.view_range, broadcastingcamera.linked_cam.view_range)
 
+/obj/structure/machinery/computer/cameras/wooden_tv/broadcast/proc/transfer_talk(obj/item/camera, mob/living/sourcemob, message, verb = "says", datum/language/language, italics = FALSE, show_message_above_tv = FALSE)
+	SIGNAL_HANDLER
+	if(show_message_above_tv)
+		langchat_speech(message, get_mobs_in_view(7, src), language, sourcemob.langchat_color, FALSE, LANGCHAT_FAST_POP, list(sourcemob.langchat_styles))
+	for(var/datum/weakref/user_ref in concurrent_users)
+		var/mob/user = user_ref?.resolve()
```
Drulikar, line 261:
```suggestion
	for(var/datum/weakref/user_ref in concurrent_users)
		var/mob/user = user_ref.resolve()
```
Since you aren't using `as anything` the `?` usage is unnecessary.

code/game/machinery/computer/camera_console.dm
```diff
@@ -219,32 +222,55 @@
...
+		if(istype(user) && user.client && user.client.prefs && !user.client.prefs.lang_chat_disabled && !user.ear_deaf && user.say_understands(sourcemob, language))
```
Drulikar, line 262:
```suggestion
		if(istype(user) && user.client?.prefs && !user.client.prefs.lang_chat_disabled && !user.ear_deaf && user.say_understands(sourcemob, language))
```

code/game/machinery/computer/camera_console.dm
```diff
@@ -219,32 +222,55 @@
...
+			sourcemob.langchat_display_image(user)
+
+/obj/structure/machinery/computer/cameras/wooden_tv/broadcast/proc/transfer_emote(obj/item/camera, mob/living/sourcemob, emote, audible = FALSE, show_message_above_tv = FALSE)
+	SIGNAL_HANDLER
+	if(show_message_above_tv)
+		langchat_speech(emote, get_mobs_in_view(7, src), null, null, TRUE, LANGCHAT_FAST_POP, list("emote"))
+	for(var/datum/weakref/user_ref in concurrent_users)
+		var/mob/user = user_ref?.resolve()
+		if(istype(user) && user.client && user.client.prefs && (user.client.prefs.toggles_langchat & LANGCHAT_SEE_EMOTES) && (!audible || !user.ear_deaf))
```
Drulikar, line 271:
```suggestion
		if(istype(user) && user.client?.prefs && (user.client.prefs.toggles_langchat & LANGCHAT_SEE_EMOTES) && (!audible || !user.ear_deaf))
```

Drulikar (changes requested):

code/datums/emotes.dm
```diff
@@ -132,19 +133,29 @@
...
+						seeing_obj |= object
 				if(current_mob.ear_deaf)
 					heard -= current_mob
 					continue
 				if(!(current_mob.client?.prefs.toggles_langchat & LANGCHAT_SEE_EMOTES))
 					heard -= current_mob
 			run_langchat(user, heard)
 
+	for(var/obj/object as anything in seeing_obj)
+		if(object) //It's possible that it could be deleted in the meantime.
```
harryob, line 156: in the meantime of what? this isn't a proc that utilizes sleep() or set background = TRUE, and the game is single threaded

ihatethisengine: I just copypasted this from mob/living/say :sob:

ihatethisengine:
Small update. Now instead of creating a new camera every time the translation goes live it just toggles the state of the camera. Another advantage of this is that you don't need to wait for CC to go live to tune TV to his camera, you can tune as soon as camera is spawned. It even works well with renaming.

ihatethisengine:
for some reason broadcasting wasn't working one round, this should probably let people to fix it if something goes wrong.

ihatethisengine:
So what was the bug. Deconstruction (and, well, destroying) calls QDEL_NULL(current) in the Destroy() where current is  the current camera. Usually it does nothing with regular cameras because it closes all instances tgui before qdellng, and when there are no users of tgui the current camera (on normal consoles) is set to null anyway. Although I managed to cause this (QDEL of a normal security camera) on normal consoles too, at least two times, but it wasn't reliable and there is no consistent steps of reproducing. It just can happen. Anyway, I didn't find a single use case for qdelling here, so I think it's safe to just remove it. It shouldn't happen normally anyway. 

Alternative solution would be to just set current to null in the child destroy() proc, if you think removing QDEL in the parent is a bad idea.

code/game/machinery/computer/camera_console.dm
```diff
@@ -33,7 +34,6 @@
 
 /obj/structure/machinery/computer/cameras/Destroy()
 	SStgui.close_uis(src)
-	QDEL_NULL(current)
 	UnregisterSignal(src, COMSIG_CAMERA_MAPNAME_ASSIGNED)
```
Drulikar, line 37:
```suggestion
	current = null
	UnregisterSignal(src, COMSIG_CAMERA_MAPNAME_ASSIGNED)
```
You must ensure its nulled out then.

A good rule of thumb to know if you need to qdel something on destroy is if this thing is the creator of said thing; but past that things need to not hold on to references of other things else you hold up SSgarbage forcing a del() to eventually occur which will cause lag.

Drulikar (changes requested):

harryob (approved):

PR #6122, opened 2024-04-08
Fix additional cryo reagent duplication via TOUCH reaction

Drulikar:
# About the pull request

This PR is a follow up to #6058 where it was brought to my attention that chemicals were still duplicating from cryotube beakers. This is because the cryotube was trying to simulate the patient being immersed in the reagents, but with a 50% permeability that doesn't even consume the reacted reagents, it easily gets out of control if large volumes are involved and or the ratio of cryo meds is small in the mixture causing frequent reapplications of reagents.

It was intended to remove chemical duplication via cyropods. This shores up another source of chemical duplication.

<details>

cryo

After rework 200 bicaradine and 20 cryoxadone:

image

</details>

# Changelog
:cl: Drathek
fix: Fixed cryotubes still duplicating chemicals because of touch reactions
/:cl:

### DIFF
--- a/code/__DEFINES/objects.dm
+++ b/code/__DEFINES/objects.dm
@@ -157,6 +157,8 @@ GLOBAL_LIST_INIT(RESTRICTED_CAMERA_NETWORKS, list( //Those networks can only be 
 
 #define TOUCH 1
 #define INGEST 2
+/// Will convert to TOUCH after skipping permeability code in /datum/reagent/proc/reaction_mob
+#define TOUCH_NO_DUPLICATION 3
 
 /// Marks an object as organic. Used for alien structures and any other organic material
 #define OBJ_ORGANIC (1<<0)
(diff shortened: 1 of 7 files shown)


harryob (approved):

fira:
if the issue is the duplication of reagents, couldn't this be disabled in reaction() instead to maintain per-chemical touch interactions?

Drulikar:
> if the issue is the duplication of reagents, couldn't this be disabled in reaction() instead to maintain per-chemical touch interactions?

See https://github.com/cmss13-devs/cmss13/blob/master/code/modules/reagents/Chemistry-Reagents.dm#L95-L116

If the reaction is touch, and user has no biosuit, 50% is duplicated into the mob. The reaction itself probably should change such that whatever enters the mob is deleted from holder, but I have no idea the implications what all that may do; it also just makes rationing what is applied to the mob difficult.

Drulikar:
Let me know if you prefer that.

fira (approved):

PR #6137, opened 2024-04-12
white phosphorus smoke can no longer override stronger flames

iloveloopers:
# About the pull request
white phosphorus smoke will no longer forcefully set the reagent of the mob

when hit by a WP rocket, a mob is supposed to be set with the blueflame reagent but sometimes the phosphorus smoke can override it just as you leave the fire and the mob's fire reagent ends up being normal UT napalm, which is signficantly weaker than blueflame.

yes I tested it

# Changelog
:cl:
fix: White phosphorus will no longer forcefully set a mob's fire_reagent to be UT napthal
/:cl:


ihatethisengine:
just make it a regular smoke? Otherwise it's a balance change because it makes a lot easier to follow up on WP.

iloveloopers:
> just make it a regular smoke? Otherwise it's a balance change because it makes a lot easier to follow up on WP.

done

PhantomEpicness:
how about you fix the overriding instead of nerfing WP rockets

ihatethisengine:
> how about you fix the overriding instead of nerfing WP rockets

how is this a nerf

fira:
why not just fix the phosphorous smoke?

iloveloopers:
> why not just fix the phosphorous smoke?

what do you mean by fix?

iloveloopers:
> why not just fix the phosphorous smoke?

is this good enough?

fira:
Oh yeah that works i think

Drulikar (approved):
Counter-intuitive that the proc works that way but works.

PR #6140, opened 2024-04-14
Lesser drone nerfs

Git-Nivrak:
# About the pull request

Lesser drones can go for extremely high risk high reward strategies without any risk as there is basically an infinite amount of respawns and no punishment for dying.

I don't see why they should get a different treatment than huggers as they can just run in and apply facehuggers manually or for going off weeds to sneak into backline and place huggers on AFK marines they saw from ghost or just running around being an annoyance.

And this is coming from a xeno main.
<details>

</details>

# Changelog
:cl:
fix: You can no longer infinitely extend  facehugger's lifetime by REDACTED
balance: Lessers now lose 5 hp every 2 seconds off weeds
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
@@ -38,6 +38,9 @@
 	/// the nearest human before dying
 	var/jumps_left = 2
 
+	var/time_to_live = 30 SECONDS
+	var/death_timer
+
 	var/icon_xeno = 'icons/mob/xenos/effects.dmi'
 	var/icon_xenonid = 'icons/mob/xenonids/xenonid_crab.dmi'
 
@@ -54,6 +57,9 @@
 	icon = new_icon
 	set_hive_data(src, hivenumber)
 	go_active()
+
+	if (hivenumber != XENO_HIVE_TUTORIAL)
+		death_timer = addtimer(CALLBACK(src, PROC_REF(end_lifecycle)), time_to_live, TIMER_OVERRIDE|TIMER_STOPPABLE|TIMER_UNIQUE)
 
 
 /obj/item/clothing/mask/facehugger/Destroy()
@@ -75,6 +81,10 @@
 	if(QDESTROYING(src))
 		return
 	addtimer(CALLBACK(src, PROC_REF(check_turf)), 0.2 SECONDS)
+
+	if(!death_timer && hivenumber != XENO_HIVE_TUTORIAL)
+		death_timer = addtimer(CALLBACK(src, PROC_REF(end_lifecycle)), time_to_live, TIMER_OVERRIDE|TIMER_STOPPABLE|TIMER_UNIQUE)
+
 	if(stat == CONSCIOUS && loc) //Make sure we're conscious and not idle or dead.
 		go_idle()
 	if(attached)
@@ -173,9 +183,18 @@
 	if(exposed_temperature > 300)
 		die()
 
-/obj/item/clothing/mask/facehugger/equipped(mob/M)
+/obj/item/clothing/mask/facehugger/equipped(mob/holder)
 	SHOULD_CALL_PARENT(FALSE) // ugh equip sounds
 	// So picking up a hugger does not prematurely kill it
+	if (!isxeno(holder))
+		return
+
+	var/mob/living/carbon/xenomorph/xeno = holder
+
+	if ((xeno.caste.hugger_nurturing || hivenumber == XENO_HIVE_TUTORIAL) && death_timer)
+		deltimer(death_timer)
+		death_timer = null
+
 	go_idle()
 
 /obj/item/clothing/mask/facehugger/Crossed(atom/target)
@@ -409,6 +428,10 @@
 	if(jump_timer)
 		deltimer(jump_timer)
 	jump_timer = null
+
+	if(death_timer)
+		deltimer(death_timer)
+	death_timer = null
 
 	if(!impregnated)
 		icon_state = "[initial(icon_state)]_dead"
--- a/code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
@@ -354,8 +354,11 @@
 
 /obj/item/clothing/mask/facehugger/proc/try_jump()
 	jump_timer = addtimer(CALLBACK(src, PROC_REF(try_jump)), time_between_jumps, TIMER_OVERRIDE|TIMER_STOPPABLE|TIMER_UNIQUE)
+	if(stat != CONSCIOUS || isnull(loc)) //Make sure we're conscious and not idle or dead.
+		jumps_left-- // Huggers should still lose hp, even when idle
 
-	if(isnull(loc))
+		if(!jumps_left)
+			end_lifecycle()
 		return
 
 	if(isxeno(loc))
@@ -363,9 +366,7 @@
 		if(X.caste.hugger_nurturing) // caste can prevent hugger death
 			return
 
-	if(stat == CONSCIOUS) //Make sure we're conscious and not idle or dead.
-		leap_at_nearest_target()
-	
+	leap_at_nearest_target()
 	jumps_left--
 	if(!jumps_left)
 		end_lifecycle()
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/castes/lesser_drone.dm
```diff
@@ -118,3 +119,21 @@
...
+	return null
+
+/mob/living/carbon/xenomorph/lesser_drone/handle_queen_screech(mob/living/carbon/xenomorph/queen/queen)
+	to_chat(src, SPAN_DANGER("The mighty roar of the queen makes you tremble and fall over!"))
+	adjust_effect(6, STUN)
+	apply_effect(6, WEAKEN)
+
+/mob/living/carbon/xenomorph/lesser_drone/get_status_tab_items()
+	. = ..()
+
```
Drulikar, line 133:
```suggestion
```
Does nothing.

code/modules/mob/living/carbon/xenomorph/castes/lesser_drone.dm
```diff
@@ -16,9 +16,10 @@
 	can_be_revived = FALSE
 
 	build_time_mult = BUILD_TIME_MULT_LESSER_DRONE
+	behavior_delegate_type = /datum/behavior_delegate/lesser_drone_base
 
 	caste_desc = "A builder of hives."
-	can_hold_facehuggers = 1
+	can_hold_facehuggers = 0
```
Drulikar, line 22:
```suggestion
	can_hold_facehuggers = TRUE
```
Its fine they can hold huggers. Make it instead that huggers transfer their time on the ground to their in hand variant so you cannot reset the timer and they expire in hands. Carriers however should reset the timer in their hand and not expire in their hands.

Drulikar (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/lesser_drone.dm
Drulikar: Unresolved. Likely just need a check for carrier in the code that was removed to make it only occur for carriers.

Git-Nivrak: Should be fixed
Carriers now prevent them from losing hp while in their hands

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
Drulikar, line 1: The changes to this file just overall make huggers very unreliable to use (they die when idle even on the ground) and kinda breaks the xenomorph tutorial if you aren't quick. Though the tutorial likely should hook into whether a hugger dies and to allow it to respawn; but I wouldn't stress about that if tutorial huggers just have the old hold behavior.

Consider  reverting all changes to this file and instead determine whether we allow go_idle on pickup depending on nurturing/tutorial.

Drulikar (changes requested):

morrowwolf:
None of the explanation here covers what you believe lesser drones *should be*. In a vacuum, this PR feels reasonable. 

But it feels reasonable because there's zero context or actual design considerations here. This is a consistent theme across dozens of PRs that slowly but surely creep their way along to making something either useless, no longer interesting, or completely and totally different. The ancient open source trap of incremental changes where each step feels reasonable but then you look back and go, "How did we even get here?"

Moreover, *why* should facehuggers and lesser drones be considered under the same umbrella? An apple and an orange are both fruits but that doesn't mean an apple should have the same characteristics of an orange. And sure, facehuggers were stunned to avoid screech hugs right? And yes, lesser drones can go and do similar by just carrying a hugger. *But*, that doesn't encompass all that a lesser drone *is*. Extending this behavior means you should justify the behavior in the new total context.

https://en.wikipedia.org/wiki/Blind_men_and_an_elephant

Iunno, maybe this PR is good. Hard to say with how little justification is given and it genuinely reeks of salt PR.

code/modules/mob/living/carbon/xenomorph/castes/lesser_drone.dm
```diff
@@ -118,3 +119,18 @@
...
+/mob/living/carbon/xenomorph/lesser_drone/handle_screech_act(mob/self, mob/living/carbon/xenomorph/queen/queen)
+	return null
+
+/mob/living/carbon/xenomorph/lesser_drone/handle_queen_screech(mob/living/carbon/xenomorph/queen/queen)
+	to_chat(src, SPAN_DANGER("The mighty roar of the queen makes you tremble and fall over!"))
+	adjust_effect(6, STUN)
+	apply_effect(6, WEAKEN)
+
+/datum/behavior_delegate/lesser_drone_base
+	name = "Base Lesser Drone Behavior Delegate"
```
Drulikar, line 132: Also not worth having these changes once the hugger changes are resolved. If your fear was lessers face hugging people that gets screeched, it should no longer be a problem. In general I want to promote lessers working in groups, and them being discouraged from hanging around the queen only pushes that goal further away.

Drulikar (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/lesser_drone.dm
Drulikar: Health drain when off weeds I will note though I believe does help encourage lessers sticking around other xenos more.

Git-Nivrak:
> None of the explanation here covers what you believe lesser drones _should be_. In a vacuum, this PR feels reasonable.
> 
> But it feels reasonable because there's zero context or actual design considerations here. This is a consistent theme across dozens of PRs that slowly but surely creep their way along to making something either useless, no longer interesting, or completely and totally different. The ancient open source trap of incremental changes where each step feels reasonable but then you look back and go, "How did we even get here?"
> 
> Moreover, _why_ should facehuggers and lesser drones be considered under the same umbrella? An apple and an orange are both fruits but that doesn't mean an apple should have the same characteristics of an orange. And sure, facehuggers were stunned to avoid screech hugs right? And yes, lesser drones can go and do similar by just carrying a hugger. _But_, that doesn't encompass all that a lesser drone _is_. Extending this behavior means you should justify the behavior in the new total context.
> 
> https://en.wikipedia.org/wiki/Blind_men_and_an_elephant
> 
> Iunno, maybe this PR is good. Hard to say with how little justification is given and it genuinely reeks of salt PR.

Yea I am going to have to say that's one of the few times I actually agree with you however the reason I think they should get the same treatment as facehuggers (which is also stated in the PR description) is there being no cost \ punishment for them dying while having a lot of impact if they do succeed to facehug someone perhaps their cost \ cooldown can be modified in a way that killing them becomes significant and they can be allowed to go along with screeches but that's out of scope for this PR

code/modules/mob/living/carbon/xenomorph/castes/lesser_drone.dm
Git-Nivrak: My worry here is they would just sit at an egg morpher waiting for screech \ drag people into eggs etc etc

Drulikar: If a screech is occuring that close to a Morpher I'm not concerned. Every other xeno ought to be trying to do the same thing in that situation - it's nothing special to lessers.

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
```diff
@@ -343,30 +350,32 @@
...
 
 	stat = UNCONSCIOUS
 	icon_state = "[initial(icon_state)]_inactive"
-	if(jump_timer)
-		deltimer(jump_timer)
-	jump_timer = null
+	if(delete_timer)
+		if(jump_timer)
+			deltimer(jump_timer)
+		jump_timer = null
```
Drulikar, line 362: To be idle, we should not have a jump timer OR try jump should do nothing if we are in an idle state.

The current problem is that now the hugger is dying while idle when the only behavior we want to change is for them to expire in hands (and ideally not to have a way to prolong).

So my recommendation is to revert the changes to this proc and to restore the CONSCIOUS check in try_jump. The stat being UNCONSCIOUS is basically your indicator for whether we are idle or not.

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
```diff
@@ -76,7 +76,9 @@
 		return
 	addtimer(CALLBACK(src, PROC_REF(check_turf)), 0.2 SECONDS)
 	if(stat == CONSCIOUS && loc) //Make sure we're conscious and not idle or dead.
-		go_idle()
+		var/jumps_before = jumps_left
+		go_idle(FALSE)
+		jumps_left = jumps_before
```
Drulikar, line 81: If you prefer to do this a different way, have the argument for the proc call be what determines whether jumps_left gets reset back to initial or not.

Drulikar (changes requested):

Drulikar (approved):

Skatuk:
They could still bypass this since they can weed.

Drulikar:
> They could still bypass this since they can weed.

They require weeds nearby to weed and have very low plasma. But yes we are completely aware the fact that weeds bypass the health decrease when not on weeds.

PR #6151, opened 2024-04-15
Small xeno tailstab nerf

Git-Nivrak:
# About the pull request

Completely removes the risk from some castes (Runner, Acider)
While giving an extra attack in a combo while completely safe for others (Lurker)
Or just being used as an annoying hit an run mechanic with low risk and high damage (Lurker, Lessers, Drones etc.)
Seems just unhealthy, On large xenos with less mobility it seems more sensible to have a 2 tiles range as well as it being needed for some large castes (e.g zerker using tail stab with apprehend to get the initial slow on)
<details>

</details>

# Changelog
:cl:
balance: Reduces tailstab range on small xenos from 2 to 1.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/general_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/general_abilities.dm
@@ -473,6 +473,9 @@
 	/// Used for defender's tail 'stab'.
 	var/blunt_stab = FALSE
 
+	var/small_xeno_range = 1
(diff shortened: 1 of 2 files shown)


Drulikar:
We can maybe consider this in the future, but for now I don't think it a change that is necessary. As just a 1-tile range this would just become a clunky ability you need to work into your rotation to be optimal. With 2-tile range its clunky, but gives you a fair bit more versatility in how you can fight.

PR #6155, opened 2024-04-16
Corrupted Queen can set up personal alliances with humans

ihatethisengine:
# About the pull request

Corrupted queen can mark people of her choice (within her view) as allies. Corrupted xenos won't be able to harm them. Just as regular alliances, personal alliances break on queen's death. Synths are human subtypes so you can ally them too.

Credits to @SubjectD9341 for the sprites.

More roleplay :)

https://github.com/cmss13-devs/cmss13/assets/115417687/2bd55b2c-c4c5-4859-ab59-1112be9f4951

# Changelog
:cl: ihatethisengine
add: Corrupted Queen can set up personal alliances with humans
/:cl:


code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
```diff
@@ -383,9 +390,82 @@
...
+		return
+	if(!user_xeno.check_state())
+		return
+	var/datum/hive_status/corrupted/hive = user_xeno.hive
+	hive.sanitize_allies()
+	if(!length(hive.personal_allies))
+		to_chat(user_xeno, SPAN_WARNING("We don't have personal allies."))
+		return
+	var/mob/living/carbon/target_mob = tgui_input_list(usr, "Target", "Break the Personal Alliance With...", hive.personal_allies, theme="hive_status")
+	if(!target_mob) return
```
fira, line 448: mobs could have been deleted etc in the meantime of tgui_input_list. you'll want to redo sanity checking after calling it.

code/modules/mob/living/carbon/xenomorph/hive_status.dm
```diff
@@ -979,6 +979,7 @@
 	need_round_end_check = TRUE
 
 	var/list/defectors = list()
+	var/list/personal_allies = list()
```
fira, line 982: Please explicit list type. In general, are allies specifically humans or any carbons or even any mob? Different code features point to all 3 - i'm arguing internally this should be /mob/living everywhere even if you decide to only allow adding human allies

code/modules/mob/living/carbon/xenomorph/hive_status.dm
```diff
@@ -1313,20 +1314,65 @@
...
 		xeno.iff_tag = null
 	if(!length(defectors))
 		return
 
-	xeno_message(SPAN_XENOANNOUNCE("You sense that [english_list(defectors)] turned their backs against their sisters and the Queen in favor of their slavemasters!"), 3, hivenumber)
+	xeno_message(SPAN_XENOANNOUNCE("We sense that [english_list(defectors)] turned their backs against their sisters and the Queen in favor of their slavemasters!"), 3, hivenumber)
 	defectors.Cut()
 
+/datum/hive_status/corrupted/proc/add_personal_ally(mob/living/carbon/human/human)
+	personal_allies += human
```
fira, line 1328: that's not delete safe due to dangling references, even if you call sanitize proc once in a while. sounds like you want to store weakrefs here

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
```diff
@@ -383,9 +390,82 @@
...
+	if(!user_xeno.check_state())
+		return
+	var/datum/hive_status/corrupted/hive = user_xeno.hive
+	hive.sanitize_allies()
+	if(!length(hive.personal_allies))
+		to_chat(user_xeno, SPAN_WARNING("We don't have personal allies."))
+		return
+
+	if(tgui_alert(user_xeno, "Are you sure you want to clear personal allies?", "Clear Personal Allies", list("No", "Yes"), 10 SECONDS) == "Yes")
+		hive.clear_personal_allies()
```
fira, line 468: similarly, checks above could be rendered invalid during the tgui_alert. i can see this happening for example if you click the button several times, letting you clear several times when the list is now already empty

fira (dismissed):
To sum up some things mentioned in other comments:
 * Typing consistency across code - this could also be generalized to /mob or /mob/living
 * Missing post alert/input checks in adding/removing/clearing allies, things can change during the dialogs
 * You can't just sanitize deleted mobs once in a while, they need to be cleared immediately by signal or just using a weakref

As extra points I'd like to bring up, hardcoding this to Corrupted hive is bit of a shame. This could be made into a feature toggleable for all hives from the get-go. You could also two-birds-one-stone it by using a component hooked to is_ally instead of bothering with CORRUPTED_ALLY at all. But both of these are secondary points...

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
```diff
@@ -383,9 +390,82 @@
...
+		return
+	var/mob/living/carbon/human/target_mob = tgui_input_list(usr, "Target", "Set Up a Personal Alliance With...", target_list, theme="hive_status")
+
+	if(!target_mob) return
+
+	if(!user_xeno.check_state(TRUE))
+		return
+
+	if(target_mob.hivenumber)
+		to_chat(user_xeno, SPAN_WARNING("We cannot set up a personal alliance with a hive cultist."))
```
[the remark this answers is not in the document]
ihatethisengine: They see the HUD icon

Drulikar:
> As extra points I'd like to bring up, hardcoding this to Corrupted hive is bit of a shame. This could be made into a feature toggleable for all hives from the get-go.
That was a requirement set by me. But if you'd rather it be a variable that only corrupted has by default set so an admin can toggle it on for other hives, sure.

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
ihatethisengine: There is a check? Or you need more?

code/modules/mob/living/carbon/xenomorph/hive_status.dm
ihatethisengine: yeah it meant to be human only, can you explain why it should be /mob/living everywhere though?

code/modules/mob/living/carbon/xenomorph/hive_status.dm
ihatethisengine: I'll try to refactor this to weakrefs

ihatethisengine:
HUDs is another reason I went for hardcode because making personal HUDs for every hive just for a feature that meant to be corrupted only sounds like a big overkill.

> Also of note, other research hives this is fine for

As far as I now, only corrupted (and renegades, but they don't have queen so it doesn't matter) can act friendly towards humans. I.e. Alpha hive should always be hostile even if research make them. 

Anyway, tell me if you need this feature for all hives and I will look what can I do. If not, I prefer to keep it simple.

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
ihatethisengine: Doesn't sound like a problem to me but okay, I'll throw more checks in.

ihatethisengine:
> * You can't just sanitize deleted mobs once in a while, they need to be cleared immediately by signal or just using a weakref

Won't I need to sanitize weekrefs once in a while anyway?

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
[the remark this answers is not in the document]
ihatethisengine: And ideally the cultist should not even appear in the list input, it's just an extra check.

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
ihatethisengine: Oh, it will spam the message, that is a problem, you're right

fira:
> That was a requirement set by me. But if you'd rather it be a variable that only corrupted has by default set so an admin can toggle it on for other hives, sure. Also of note, other research hives this is fine for; but I don't think it appropriate for normal hive.

Yeah I meant this purely codewise, it can be incorporated into global hive code and disabled when not appropriate rather than hardcoded and require a refactor if used elsewhere. But it might just be wishful thinking, if we really have no use case for it, it's not worth bothering 

> Won't I need to sanitize weekrefs once in a while anyway?

You can pop invalid ones out of the list if you want, but i'd say it's not a requirement for something low traffic like this

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
fira: I meant a check regarding validity of removal of the ally but that's already handled in the *other* remove_personal_ally

code/modules/mob/living/carbon/xenomorph/hive_status.dm
fira: Well my reasoning is that CORRUPTED_ALLY is a status_flags so relevant to /mob,  and is_ally will get called for xenos too. I don't see a reason to fence the whole thing to only work with humans rather than living or carbon. You can always just only allow humans to be added when selecting allies

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
```diff
@@ -437,35 +440,66 @@
...
+
 	var/datum/hive_status/corrupted/hive = user_xeno.hive
-	hive.sanitize_allies()
 	if(!length(hive.personal_allies))
 		to_chat(user_xeno, SPAN_WARNING("We don't have personal allies."))
 		return
 
-	if(tgui_alert(user_xeno, "Are you sure you want to clear personal allies?", "Clear Personal Allies", list("No", "Yes"), 10 SECONDS) == "Yes")
-		hive.clear_personal_allies()
+	if(tgui_alert(user_xeno, "Are you sure you want to clear personal allies?", "Clear Personal Allies", list("No", "Yes"), 10 SECONDS) == "No")
```
Drulikar, line 496:
```suggestion
	if(tgui_alert(user_xeno, "Are you sure you want to clear personal allies?", "Clear Personal Allies", list("No", "Yes"), 10 SECONDS) != "Yes")
```
So there's typically 3 situations for alerts to resolve: Your first option, your first option, or no option (X button & time out). It is indeed the norm that the failure case is often the inside look because we have the norm of early returns.

As written, this makes it so allies are cleared on timeout & cancel, so instead test not the accept case.

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
```diff
@@ -437,35 +440,66 @@
...
+
 	if(!length(hive.personal_allies))
 		to_chat(user_xeno, SPAN_WARNING("We don't have personal allies."))
 		return
-	var/mob/living/carbon/target_mob = tgui_input_list(usr, "Target", "Break the Personal Alliance With...", hive.personal_allies, theme="hive_status")
-	if(!target_mob) return
+
+	var/list/mob/living/allies = list()
+	var/list/datum/weakref/dead_refs = list()
+	for(var/datum/weakref/ally_ref in hive.personal_allies)
```
Drulikar, line 455:
```suggestion
	for(var/datum/weakref/ally_ref as anything in hive.personal_allies)
```
So if we know this can only contain weakrefs, we might as well not type check every entry to be a weakref.

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
```diff
@@ -437,35 +440,66 @@
...
+		var/mob/living/ally = ally_ref.resolve()
+		if(istype(ally))
```
Drulikar, line 457:
```suggestion
		if(ally)
```
`/datum/weakref/proc/resolve` both checks if its qdeleted (which involves a null check) and if the weak_reference circular reference still points back to this weakref, so type checking it should not be necessary. But do try it out.

code/modules/mob/living/carbon/xenomorph/hive_status.dm
```diff
@@ -1324,40 +1324,33 @@
...
-	if(!(human.status_flags & CORRUPTED_ALLY))
-		return
-	human.status_flags &= ~CORRUPTED_ALLY
-	human.med_hud_set_status()
-	personal_allies -= human
-	xeno_message(SPAN_XENOANNOUNCE("Our Queen has decided that [human] is no longer our ally!"), 3, hivenumber)
+/datum/hive_status/corrupted/proc/remove_personal_ally(datum/weakref/ally_ref)
+	personal_allies -= ally_ref
+	var/mob/living/ally = ally_ref.resolve()
+	if(istype(ally))
```
Drulikar, line 1336:
```suggestion
	if(ally)
```
Should be the same case here, but try it out and let me know if I am wrong.

code/modules/mob/living/carbon/xenomorph/hive_status.dm
```diff
@@ -1324,40 +1324,33 @@
...
+		ally.med_hud_set_status()
+	xeno_message(SPAN_XENOANNOUNCE("Our Queen has decided that [ally] is no longer our ally!"), 3, hivenumber)
 
 /datum/hive_status/corrupted/proc/clear_personal_allies(death = FALSE)
-	for(var/mob/living/carbon/human/human in personal_allies)
-		human.status_flags &= ~CORRUPTED_ALLY
-		human.med_hud_set_status()
+	for(var/datum/weakref/ally_ref in personal_allies)
+		var/mob/living/ally = ally_ref.resolve()
+		if(!istype(ally))
```
Drulikar, line 1344:
```suggestion
		if(!ally)
```
"

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
```diff
@@ -383,9 +390,116 @@
...
+	var/mob/living/carbon/xenomorph/queen/user_xeno = owner
+	if(user_xeno.hive.hivenumber != XENO_HIVE_CORRUPTED)
+		return
+
+	if(!user_xeno.check_state())
+		return
+
+	var/datum/hive_status/corrupted/hive = user_xeno.hive
+	var/list/target_list = list()
+	for(var/mob/living/carbon/human/possible_target in view(7, user_xeno))
```
Drulikar, line 413: Is user_xeno here always the queen even on ovi and in eye view - or is it where she is looking?  Ideally this should be in an area the queen is looking.

code/modules/mob/living/carbon/xenomorph/hive_status.dm
```diff
@@ -1313,20 +1314,58 @@
...
+	ally.med_hud_set_status()
+	xeno_message(SPAN_XENOANNOUNCE("Our Queen proclaimed [ally] our ally! We must not harm them."), 3, hivenumber)
+
+/datum/hive_status/corrupted/proc/remove_personal_ally(datum/weakref/ally_ref)
+	personal_allies -= ally_ref
+	var/mob/living/ally = ally_ref.resolve()
+	if(istype(ally))
+		ally.status_flags &= ~CORRUPTED_ALLY
+		ally.med_hud_set_status()
+	xeno_message(SPAN_XENOANNOUNCE("Our Queen has decided that [ally] is no longer our ally!"), 3, hivenumber)
```
Drulikar, line 1339:
```suggestion
	xeno_message(SPAN_XENOANNOUNCE("Our Queen has declared that [ally] is no longer our ally!"), 3, hivenumber)
```
Since you used proclaimed for allying; merely deciding seems weak.

Drulikar (changes requested):

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
ihatethisengine: Does work now (without ovi, on ovi and in eye form away from the queen itself). I wonder if I need some additional checks here (like if eye is actually a queen eye), not sure if I am missing anything.

Also I didn't know that view checks for walls even if you have wallhack so I use range now.

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_powers.dm
ihatethisengine: Well, it's working without issues. Tested this proc, deleted the mob and it didn't appear.

nauticall (approved):
sprite 👍 

the health hud bit being modified is a bit weird but it looks the same so, alright

Drulikar (approved):
If nothing has been discovered at this point, we can fix it ever something comes up. But as far as I can tell its all working locally.

PR #6163, opened 2024-04-18
Heavy Sniper Part 2 - The Heavying

Kaga-404:
# About the pull request

New Updates:
- AMR Aimed Shot base cast time increased by 60% from the M42A (1.25s to 2s), to emphasize working with spotter and normal shots, and solidly setting the Aimed Shot DPS under the M42A even under optimal conditions (decreased by ~40%). If this is too crippling, may increase Spotter Bino bonus to the AMR or slightly roll back cast time in the future.
- Interrupt no longer applies to Queen and Crushers unless they are the primary target of an Aimed Shot. Damage threshold requirements still apply, and any piercing will still prevent it from interrupting big xenos at all.
- Charger Strain no longer gets interrupted or slowed while charging.
- Aimed Shot slow no longer applies to Runners, as they didn't have the protection of Focused Fire to reduce the duration by half initially.
- Screenshake maximum against big xenos decreased by 1 step. Now applies at 90 and above to them, instead of 60.
- Changed non-xeno (human) stun to stamina damage instead of directly stunning, to match slugs and high impact.

XM43E1 AMR shots now have stopping power, a scaling system that functions similarly to shotgun slugs and heavy impact ammo, both in role and visible effect.

**SCREENSHAKE:** Hitting a target will shake their screen so they know they got hit by something heavy. This shaking scales with damage dealt, with the maximum shake being equal to slugs and impact ammo. At 60 or less damage, this does not shake the screen.

**INTERRUPTION:** Hitting a target will apply a mini-stun, scaling with damage dealt. At 60 or less damage, this does not interrupt targets at all.

- Big Xenos need a base damage of 90 or more damage, will only ever by stunned for 0.1s max, and **cannot be knocked back.**
- Other xenos can be stunned for up to 0.3s if the base damage is more than 120, losing 0.1s per 30 base damage, and will be knocked back at 90 damage or more.
- Non-xenos can be stunned for up to 3s, losing 1s per 30 base damage.

For comparison, impact rounds stun for 1s, and slugs stun for 1.4s as well as buckshot within range. **The stun on this is much lower as it is only meant to interrupt, and can affect big xenos if they are the only target hit, which other heavy rounds cannot.**

**AIMED SHOT:** Can now potentially apply a slow scaling with stopping power, similar but much weaker than M42A Flak. At 60 or less base damage, no slows are applied. 

- Lasts for at most 5s of slow, 2.5s of superslow against runners and non-xenos, at 120 or more base damage.
- Slows are reduced by 1s and 0.5s respectively per size tier, and for every 30 base damage lost.
- If this slows any amount, it will always give 0.2s of reduced vision as a visible warning.

Other Changes:
- Defenders are no longer counted as an exception for Aimed Shot, only Runners. Defenders should take 200+20% current health damage on an aimed shot now, but are affected by Focus modifiers, so only 10% current health on first, then 15%, then 20% current health. Being Fortified reduces the flat damage to 125 but increases current health damage to 15/22.5/30%.
- Increased base armor penetration for XM43E1 and M42C, as in rare cases target armor could go above 50.

The heavy sniper currently has almost no incentive to use its standard fire due to how little ammo it has, and being less damage efficient overall, as well as no utility on its normal shots. Consequently, it's unable to actually support allies in most cases. These mechanics should hopefully encourage standard fire more, especially against non T3s, allowing heavy sniper users to juggle their objectives of focusing aimed shots against a priority target while also using their standard shots at critical points to save allies from getting dragged off. This leans further into the weapon's identity as a support, as playtesting so far has shown that it's generally much worse at securing kills due to massively reduced DPS and aimed shot damage against T3s in the majority of situations, with its current health scaling.

Because the stopping power mechanic has very low stun numbers, max 0.3s against xenos, and the firerate is incredibly low at 3s between shots, there's effectively no kill potential with this support feature. Instead, it's a way to save allies from a far distance like slug shotguns, but with IFF and much higher range, allowing it to be done from safety with no risk of slugging your teammates in the back because they resisted out of a grab.

Additionally, high-caliber rounds should feel heavy, which is why ammo like Scout Impact and shotgun slugs give both mechanical and visual cues when they hit a target. Briefly reduced vision, stun, knockback.

Both from the user's and target's perspectives, getting hit by a heavy sniper round doesn't have any visual cue different from a pistol bullet, and the only noticeable impact is health loss unless you're staring at chat to see the 'hit by' message. There's a distinctive firing sound, but in an active front it doesn't really let the target know they got hit by the AMR, in the midst of all the other guns firing in their direction.

With this update, the screenshake and small knockdown should make it much more likely for targets to realize they've been hit by something dangerous.

<details>

</details>

# Changelog

:cl: Kaga
balance: The Heavy Sniper's XM43E1 rounds are now actually heavy. Targets hit will get screenshake, like slugs and impact rounds, scaling with damage dealt.
balance: XM43E1 shots can now potentially interrupt targets if it deals enough damage. The heavier the target, the more force is needed to interrupt them. Big xenos require a full-damage shot with no piercing, while smaller targets can also be knocked back with enough damage.
balance: XM43E1 Aimed Shots can now potentially apply a slow to the main target only, scaling with base damage and target size. If a slow is applied, vision range is reduced for a moment as a warning.
balance: XM43E1 and M42C ammo AP increased to 75 (from 50)
balance: Removed the Defender exception from the AMR Aimed Shot calculations.
/:cl:


Drulikar:
In the future I recommend basing off of a current master so you don't have all this ancillary commit history.

I'm skeptical of 3s stun potential ontop of slows/superslows/dazes/knockbacks. I'm also skeptical of the AP increase. I'm also skeptical of the increased damage to defenders - likely you want to test if they are fortified or not and do no status effects nor damage increase to them if fortified.

Diegoflores31:
i feel this series of small buffs will end up making AMR a sniper 2.0 instead of an alternative as it is suposed to be .
it has wall penetration , stun , screen shake .    where is the cost or flaw to justify all of that ?

sl0w3st:
Love to see scout's impact ammo being nerfed with distance due to offscreen scope chain-stun combat but this one being added

Why chosing "ammo variety" and "bigger mag" over more damage, damage stacking, stuns and pushback?

PhantomEpicness:
The flicker stun is too powerful on charger crushers as it allows an AMR sniper to stop a charger crusher dead in it's tracks, even in the middle of a full momentum charge, allowing a very easy kill. 

As charging is it's primary ability for movement and damage (along with the fact that it was only designed with AP rockets to completely shut down a charge), this effectively hard counters the entire strain.

I suggest a unique interaction with charger crusher:

If a charger crusher is charging (using TRAIT_CHARGING), lower it's charge momentum instead. This makes AMR able to deter charger crushers and make their life harder without sentencing them to death.

ihatethisengine:
If you want interrupting make interrupting, why make it a stun? Or don't misrepresent it as "interrupting". You can easily code in "stop pulling" on shot or whatever you want, instead of sneaking in a stun.

Diegoflores31:
Just saw some AMR in action . i can say its insanely strong . it can secure kills in T1s T2s T3s super easy with the stun + the damage + IFF + range. and they dont even need to aim shot to get the stun which makes all of this more hilarious they can just normal shot you as T2 and bye bye with a single bullet.

sl0w3st:
> If you're going to make a claim, I'd ask that you first ensure that it's accurate. At max HP and max stacks, a matured xeno Queen has 1000 HP, and takes 125 + 30% = 425 damage, leaving her with 575. I'm not sure where you're getting '1/3' HP from.

True, my bad. I was talking from in-game perspective, not test enviroment where you are alone sniper vs alone queen - charging queen always would get shot at by some folks around
Full stacked shot deals 425 damage to queen, true. 425 of total 1000 HP. Quite close to 1/2 HP, huh
That's two aimed shots with regular marksman in one. With microstun and slowdown applied. Microstun could be ignored due its only meaningful impact being breaking grabs

> You pick the normal M42A to be able to flexibly hit any xeno, at any point, with the highly effective and long-lasting debuffs like fire, blindness, superslow, or raw execute damage that doesn't need to ramp up. This also comes with higher base DPS and higher aimed shot DPS, cheaper ammo, and not being helpless against your focused target leaving that front and forcing you to use another 2 inefficient shots....

2 aimed shots take around 4-6 seconds and being shot at queen causing:
420 damage by regular marksman aimed shot, resulting in 70+ DPS
~720 damage by AMR aimed shot, resulting in 120+ DPS

and let's add this one...

> AMR Scaling Damage: First shot on any T3 does less than 210 damage at 566 or less HP. This means if you're switching targets, and they're not at around max health, you are doing less damage to a T3 with the AMR aimed shot than a normal sniper rifle. At the 'execute range' of a marksman aimed shot (210 HP), the first aimed shot on a big xeno with the AMR will deal 156.5 damage with a longer aimed shot cooldown and less ammo.

I understand idea about losing damage stacks, but why would player even consider switching if he can just wait? Xenos are forced to frontline, only boiler and resin whisperer may avoid open ground for whole round. You stack you damage, wait when xeno heals fully and goes fighting again and do your aimed shot. Repeat. Oh, maybe if xeno went to another flank.
But how Queen would avoid being aimed at? Never charge out? 
Also to mention T2 castes being shot down more thah half HP from full by single aimed shot with full damage stacking

> You can pre-chamber ammunition and stack debuffs on top of one another easily, like Incendiary in chamber. Normal shot first, then aimed shot with flak to have a xeno on fire and superslowed for 8-14s.

Do you really try to "combat" or, guess, "get on par with" this one exact mechanic here? 70 damage from regular marksman plink followed by aimed shot with incendiary stacks or flak slowdown? Apply one fire stack and cause slowdown? I have hard time to see it being similar with current AMR and read it as "to make M42A as good as AMR you need to juggle your mags, marine". On other words, be very skilled sniper with M42A but with AMR just be patient to win

Question still stands: why players would choose overall weaker kit if damage means more? Was it really needed to compile weaker, yet a flak and twice as marksman into one thing under sauce of more utility?
Why removing defender blocking? What do you propose to counterplay this specialist as xenos?
Marines are able to get second AMR sniper with intel and Foxtrot. One proper spec kit becomes game ender now?

ihatethisengine:
You really should try to look at your project from a different angle. Sniper is by design a very low risk spec, you have IFF, you shoot from several screens away, darkness is not a problem for you, in your version you have bonus stacks you **never** lose unless you choose to. And it deals damage equals up to 50% of T3 HP with full AP, and on top of that you decided to put microstun (which is probably actually longer than 0.1 sec because of how status processing works), screenshake, and even slowdown. I literally cannot understand what is going on here. How can you look at 425/1000 dmg on Queen, 321/650 dmg on base rav and think "well, that's not enough". Queen can just disengage? Okay, you cannot always just disengage depending on the situation, but even if you can, forcing Queen to disengage with one single shot is already insanely strong power. Losing 50% of your HP because of spec you can do literally nothing about, and also getting a "microstun" (that gives people the edge to start shooting you a bit sooner), and even slow. 

This is extremely broken. It already was broken if you ask me, dealing 50% of HP in one shot from several screens away should have never been a thing. You compare it to regular sniper for whatever reason, which is funny, because this is the definition of powercreep, this is how it happens every time: we add an op gun, say it underperforms in comparison to another very strong gun (which has a lot of the same balancing problems, I am fairly certain that sniper spec is the most broken spec in the game, and just like your gun now, got significantly powercreep'd. You used to be able to use aimed shot only when camouflaged, now you can do it anytime, which significantly reduced preparation time, so you can just run around and instantly use aim shot, so the only downside which was the lack of mobility was removed), so we buff the first gun to make it even stronger. Next step is to put up a PR that buffs M42A, because now it underperforms compared to the new AMR.

But all of your comparisons are wrong, because you will never deal 321-425 damage in one shot with M42A. And this is all what matters. I literally cannot understand how you came to the conclusion that your AMR is somehow 100% weaker. The fact M42A can instakill some xenos with regshot + aimed shot combo is a reason to nerf M42A, not to buff AMR.

Also you say Queen should change her strategy if she's getting targeted by sniper. I wonder how. To use your main ability as Queen you gotta leave cover. What can you do about spec plinking you for 425 hp? Change flanks? Sniper can do the same, thankfully command will announce Queen's movement. Guess you should just go on ovi. 

But even that wouldn't be a fail state for the sniper, as he can just choose a new target. Stacking bonuses is not hard, you never lose them (even if you miss or don't use aimed shot, which is just baffling, where are the risks, trade-offs or downsides???), and you are never in rush, because you are in complete safety, with IFF, several screens of view, etc. Compare it to SADAR, for example, who has to try to utilize every opportunity, who is always at risk, who is a priority target, etc. etc.

code/datums/ammo/bullet/sniper.dm
```diff
@@ -106,11 +106,61 @@
 	accuracy = HIT_ACCURACY_TIER_8
 	damage = 125
 	shell_speed = AMMO_SPEED_TIER_6
+	penetration = ARMOR_PENETRATION_TIER_10 + ARMOR_PENETRATION_TIER_5
+
+/datum/ammo/bullet/sniper/anti_materiel/proc/stopping_power_knockback(mob/living/living_mob, obj/projectile/fired_projectile)
+	var/stopping_power = min(CEILING((fired_projectile.damage/30), 1), 5) // This is from bullet damage, and does not take Aimed Shot into account.
+
+	. = stopping_power
+
```
Drulikar, line 115:
```suggestion
```
No real point unless you think a runtime is possible.

code/modules/cm_marines/equipment/kit_boxes.dm
```diff
@@ -71,7 +71,7 @@
 
 /obj/item/storage/box/spec/sniper/anti_materiel/fill_preset_inventory()
 	name = "\improper AMR equipment case"
-	desc = "A large case containing an experimental XM43E1, a set of M45 ghillie armor and helmet, a M42 scout sight, ammunition, spotter equipment, and additional pieces of equipment.\nDrag this sprite onto yourself to open it up! NOTE: You cannot put items back inside this case."
+	desc = "A large case containing an experimental XM43E1, a set of M45 ghillie armor and helmet, an M42 scout sight, ammunition, a set of spotter gear, and additional pieces of equipment.\nDrag this sprite onto yourself to open it up! NOTE: You cannot put items back inside this case."
```
Drulikar, line 74:
```suggestion
	desc = "A large case containing an experimental XM43E1, a set of M45 ghillie armor and helmet, a M42 scout sight, ammunition, a set of spotter gear, and additional pieces of equipment.\nDrag this sprite onto yourself to open it up! NOTE: You cannot put items back inside this case."
```

Drulikar (approved):

PR #6170, opened 2024-04-19
[Balance] OT Damage Falloff Rebalance Attempt

TeDGamer:
# About the pull request

Makes falloff modifier never negative. Falloff modifiers being negative meant that the explosion was still pretty strong a few tiles away.

image

https://github.com/cmss13-devs/cmss13/commit/006988ef0ac5d76e98e85f20997f157cc79aabf5

Looking through the git history and commit for this particular balance change, I'm unsure why some of these values were negative, especially welding fuel which is so easily accessible. The philosophy seemed to be the more power, the less the falloff but either way it seemed that the falloff would decrease instead of increase.

Thermite and ammonium nitrate had lower power, but also higher falloff. 

image
image

Given the above, I'm unsure how the falloff modifiers for the other chemicals were negative since while they increased power, they decreased falloff.

Before:
Welding fuel: Decreased falloff by 0.1 per volume
ANFO: Decreased falloff by 0.6 per volume
Nitroglycerin: Decreased falloff by 0.5 per volume
Cyclonite: Decreased falloff by 0.4 per volume
Octogen: Decreased falloff by 0.2 per volume

After:
Welding fuel: Increased falloff by 0.1 per volume
ANFO: Increased falloff by 0.2 per volume
Nitroglycerin: Increased falloff by 0.4 per volume
Cyclonite: Increased falloff by 0.5 per volume
Octogen: Increased  falloff by 0.6  per volume

This change should keep with the philosophy of trading between ANFO and Octogen.

image

image

Didn't touch this formula or anything but it's also a bit confusing.

So the higher the level, the stronger the power, but the smaller the decrease to falloff.

Level 1 = 1 explosion power, -3 falloff modifier
Level 10 = 10 explosion power, -0.3 falloff modifier

Rockets shouldn't somehow be critting anyone a huge distance away.

Chemicals should not be somehow decreasing the falloff.

# Changelog

:cl:
balance: Falloffs modifiers for chemicals are now positive
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry_reagents/other.dm
+++ b/code/modules/reagents/chemistry_reagents/other.dm
@@ -491,7 +491,7 @@
 	//------------------//
 	explosive = TRUE
 	power = 0.12
-	falloff_modifier = -0.1
+	falloff_modifier = 0.1
 	burncolor = "#ff9900"
 	chemclass = CHEM_CLASS_RARE
 	properties = list(PROPERTY_FUELING = 5, PROPERTY_OXIDIZING = 3, PROPERTY_VISCOUS = 4, PROPERTY_TOXIC = 1)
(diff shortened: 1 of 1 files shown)


iloveloopers:
I don't think it OT works that way

PR #6171, opened 2024-04-20
fixes high combustion napalm and sticky napalm working for OT assemblies

iloveloopers:
# About the pull request
I didn't intend for these to work in OT assemblies, they're too good
apparently chemfiresupp doesn't work if you have fire properties cause they set the reagent's chemfiresupp to true forcefully
so I manually set the durationmod, intensitymod and radiusmod to negative to exatcly counter how much they affect

bugfix good
unintended behaviour

it works, trust me

# Changelog
:cl:
fix: custom flamer fuels no longer work for increasing OT assembly's caps
/:cl:


Drulikar (approved):

BygoneDays236:
so fun fact Sticky napalm (which IRL normally ignites on impact) does not create fire in-game due to this change

PR #6103, opened 2024-04-07
Med Vendor Rebalance

Drulikar:
# About the pull request

This PR is a follow up to #5677 and does several things:
- Disables splitting of medical items
- In the case of splints, to avoid splitting them via applying and removing them, recovered splints are considered contaminated. However, you can get them to not be deemed contaminated if you form a full stack with them.
- Increases the starting stock of several medical vendors
- Adds periodic automatic restocking to med linked medical vendors
- Initializes groundside medical vendors with randomly reduced stock amounts

This PR also fixes a some miscellaneous things:
- Using a fully stacked item on a medvendor will now try to fill the vendor with that item (rather than only performing this action with a mouse drop).
- Fixes some of the grammar issues regarding restocking
- Fixes some unused squad prep vendors that were not properly adding headsets to their list of items
- Fixes removing nano splints creating splints with 0 amount

# Explain why it's good for the game

The ultimate goal of this PR is to remove the infinite quantity of medical items groundside so there is more pressure on requisitions to keep groundside supplied.

It also is a little goofy that a colony that got decimated by a xenomorph infestation would have fully stocked medical supplies.

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

https://youtu.be/6ES88Zre6Gg

</details>

# Changelog

:cl: Drathek
balance: Increased the starting stock of most medical vendors
balance: Med linked medical vendors will now automatically restock periodically
balance: Groundside medical vendors will now have random stock missing
balance: Medical items can no longer be split into separate stacks
qol: Restocking a vendor manually can be done with just a click if it is a full stack rather than only via mouse drop
fix: Fixed some currently unused squad prep vendors
fix: Removing nano splints on a person no longer creates a 0 amount normal splint
spellcheck: Tweaked some restock messages
/:cl:

PR #5677, opened 2024-02-07
Requisition Centralization [Part 1.5] AKA medvendor infinite restock outside medbay cheese removal redux

ItsVyzo:
# About the pull request

This PR adds a new structure: the medical supply link port. This is a structure that allows medical vendors to link to the Almayer's internal medical supplies and have larger restocking capabilities. They can be found around the Almayer where existing medical vendors are.

This PR makes it impossible to restock/refill non full stacks at medical vendors which are not linked to a medical supply link port

This PR makes it impossible to restock/refill non full autoinjectors and bottles at medical vendors which are not linked to a medical supply link port

revives #2155

# Explain why it's good for the game

This PR is good for the game because it will get req more involved in the medical supply process, as well as limiting the capability of medics to instantly resupply from a vendor at the fob or in hydro. This makes it so that medics have to more carefully pick their supplies, and plan ahead for resupplying, as well as consider other medics' use of supplies, as nothing is infinite groundside anymore.

Overall, it will make the medic supply gameplay loop more engaging and interesting, as well as reducing marine capability to infinitely heal and resupply themselves groundside.

Makes #5490 more useable

# Testing Photographs and Procedure

<details>

https://gyazo.com/d9e8e3b63daab1241e046c23bb5dd140

</details>

# Changelog

🆑 Stan_albatross,kugamo,LTNTS,Drathek
balance: medical vendors, excepting nanomeds, can now only refill/restock stacks and autoinjectors/bottles when on special medical supply link ports that can only be found shipside. Sprite by kugamo.
fix: supply pads will no longer generate ambient occlusion.
/:cl:

PR #6180, opened 2024-04-22
Adds medlinks to WO and ground maps

Drulikar:
# About the pull request

This PR is a follow up to #5677 assuming changes from #6103 adding a medlink to each ground map as well as medlinks for the WO main base.

The other map changes are just the result of StrongDMM's sanitize variables setting where it deletes var edits that match what the type already has initially.

Helps establish colony medbays as a temporary objective that can help the groundside operation with medical supplies. Some maps this is more practical than others, but the medlinks need to not be too close to a landing zone else it is too hard to contest them.

As far as WO, there isn't a shipside medical so the main base instead should serve that purpose.

<details>

</details>

# Changelog
:cl: Drathek
maptweak: Added medlinks to WO
maptweak: Added a medlink to each ground map
/:cl:

### DIFF
--- a/maps/map_files/BigRed/BigRed.dmm
+++ b/maps/map_files/BigRed/BigRed.dmm
@@ -27889,6 +27889,14 @@
 	icon_state = "wood"
 	},
 /area/bigredv2/outside/library)
+"efh" = (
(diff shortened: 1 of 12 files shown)


Nanu308 (approved):
lgtm

PR #6192, opened 2024-04-25
Command announcements for survivors

Git-Nivrak:
# About the pull request
Allows survivors to view tacmap and hear command announcements once they get IFF and marine headsets.

When you redeploy as a survivor (which is what you do 90% of the time) it is extremely painful to play without knowing what's going on, no announcements, no tacmap no anything. It is also makes sense that you will have access to those once temporarily enlisting with the marines and having the proper equipment.
<details>

</details>

# Changelog
:cl:
add: Survivors can now receive command announcements with IFF and a marine headset.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -115,7 +115,7 @@
 			. += "Primary Objective: [html_decode(assigned_squad.primary_objective)]"
 		if(assigned_squad.secondary_objective)
 			. += "Secondary Objective: [html_decode(assigned_squad.secondary_objective)]"
-	if(faction == FACTION_MARINE || (FACTION_MARINE in wear_id.faction_group))
+	if(faction == FACTION_MARINE)
 		. += ""
 		. += "<a href='?MapView=1'>View Tactical Map</a>"
 	if(mobility_aura)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -115,7 +115,7 @@
 			. += "Primary Objective: [html_decode(assigned_squad.primary_objective)]"
 		if(assigned_squad.secondary_objective)
 			. += "Secondary Objective: [html_decode(assigned_squad.secondary_objective)]"
-	if(faction == FACTION_MARINE)
+	if(faction == FACTION_MARINE || (FACTION_MARINE in wear_id.faction_group))
```
Drulikar, line 118:
```suggestion
	if(faction == FACTION_MARINE || (FACTION_MARINE in wear_id?.faction_group))
```
Ever you take off your id, this runtimes: 
image

Drulikar (changes requested):
Currently only announcements work.

Tactical map update messages and accessing the marine tacmap do not work.

image
image

Various changes would have to be made for all faction checks throughout minimap.dm. It was kinda intentionally hardcoded to marine/normal hive. But look at uses of `FACTION_MARINE` in that file and consider if they aren't either marine or normal hive to check their IFF tag is marine, and if so consider their faction marine in that case.

The act for `if("selectAnnouncement")` is also where that tactical map update in progress message would get sent.

Git-Nivrak:
Going to atomize this to only announcements
There appears to be more to the tacmap than just changing the faction checks and so I will rather deal with at a different time

Drulikar (approved):

PR #6195, opened 2024-04-26
Consistent OOC UX

blackdragonTOW:
# About the pull request

Unifies OOC UX so that a players name color can actually be used to understand their position within the community. 

Removes color self OOC that way name colors in chat actually mean something other than "I have perms."   Good UX means that those colors mean something rather than snowflake randomness.

<details>

</details>

# Changelog

OOC name colors are now unified
:cl:
config: Made OOC name colors consistent with ranks within the community.
/:cl:


fira:
This doesn't change anything on live server, because it has its own config file. The file you're editing is just an example for contributors

If you want this to change you need to reach out to Host/Dev Managers/Sysadmins

PR #6199, opened 2024-04-28
Adds an animation when missing melee attacks

Git-Nivrak:
# About the pull request
Adds an animation (similar to xeno swipe when missing slashes) when missing melee attacks with items.

More indicative to the player, Rather than thinking their clicks get "eaten" they will be able to see when they miss and when the attack delay is in effect

https://github.com/cmss13-devs/cmss13/assets/59925169/2e788aae-03c0-4b1a-b19e-6936467e8d13

# Changelog

:cl:
add: Added an animation when missing melee attacks with items
/:cl:

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -149,13 +149,26 @@
 		if(W.attack_speed && !src.contains(A)) //Not being worn or carried in the user's inventory somewhere, including internal storages.
 			next_move += W.attack_speed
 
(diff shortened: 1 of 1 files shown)


fira:
I dunno... I think this could be misleading in its own way too

Maybe the animation could be changed so you can tell both apart?

Drulikar (changes requested):
As I mentioned on discord, try doing what Fira suggested and tweak the animation for a miss. Even just a different tween may be enough.

PR #6211, opened 2024-04-29
validates sql in unit testing

harryob:
no player facing etc


fira:
Awesome idea!
That's all there is to it, RUST_BAKCTRACE provides neccessary output?

PR #6233, opened 2024-05-02
Repaths `old_research.dm`

Zonespace27:
# About the pull request

Converts some paths into snake case from camel case.

### DIFF
--- a/maps/predship/regular.dmm
+++ b/maps/predship/regular.dmm
@@ -1177,8 +1177,8 @@
 /area/yautja)
 "dp" = (
 /obj/structure/surface/table/reinforced,
-/obj/item/XenoBio/Blood,
-/obj/item/XenoBio/Blood,
+/obj/item/xeno_piece/blood,
+/obj/item/xeno_piece/blood,
 /turf/open/floor/holofloor{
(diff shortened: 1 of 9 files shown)


kiVts:
this will conflict so bad with the organs pr :sob: + this is already included in organs pr

maps/predship/regular.dmm
```diff
@@ -1177,16 +1177,16 @@
...
+/obj/item/xeno_piece/blood,
 /turf/open/floor/holofloor{
 	dir = 2;
 	icon_state = "cult"
 	},
 /area/yautja)
 "dq" = (
 /obj/structure/surface/table/reinforced,
-/obj/item/XenoItem,
+/obj/item/xeno_item,
```
fira, line 1189: The desc makes it sound like an abstract one (prompting opening a bug report) - i think it's worth either making it so and changing this one to another, or clearing that up from the description

fira (changes requested):

Zonespace27:
> this will conflict so bad with the organs pr 😭 + this is already included in organs pr

so it was

funny-sad-emoji-getting-disintegrated-into-dust

PR #6237, opened 2024-05-03
Watcher nerf

Git-Nivrak:
# About the pull request
Watcher facehugger strain now takes damage off weeds the same as the base variant 

It's a straight up upgrade to the base variant + Metagaming afk marines \ open cades from ghost is cancer
<details>

</details>

# Changelog
:cl:
balance: Watcher is now immobilized when zoomed out and loses health off weeds when not zoomed out
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/ability_helper_procs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/ability_helper_procs.dm
@@ -195,6 +195,10 @@
 			client.pixel_x = -viewoffset
 			client.pixel_y = 0
 
+	for (var/datum/action/xeno_action/onclick/toggle_long_range/action in actions)
+		action.on_zoom_in()
+		return
+
 /mob/living/carbon/xenomorph/proc/zoom_out()
 	if(!client)
 		return
--- a/code/modules/mob/living/carbon/xenomorph/abilities/facehugger/facehugger_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/facehugger/facehugger_powers.dm
@@ -25,14 +25,23 @@
 		if(current_airlock.density) //if its CLOSED YOU'RE SCUTTLING AND CANNOT POUNCE!!!
 			to_chat(owner, SPAN_WARNING("We cannot do that while squeezing and scuttling!"))
 			return FALSE
+
+	if(HAS_TRAIT(owner, TRAIT_IMMOBILIZED))
+		to_chat(owner, SPAN_WARNING("We cannot do that while immobilized!"))
+		return FALSE
+	
 	return ..()
 
-/datum/action/xeno_action/onclick/toggle_long_range/facehugger/use_ability()
+/datum/action/xeno_action/onclick/toggle_long_range/facehugger/on_zoom_out()
 	. = ..()
 
 	var/mob/living/carbon/xenomorph/facehugger/facehugger = owner
 
-	if (facehugger.is_zoomed)
-		ADD_TRAIT(facehugger, TRAIT_IMMOBILIZED, TRAIT_SOURCE_ABILITY("Long-Range Sight"))
-	else
-		REMOVE_TRAIT(facehugger, TRAIT_IMMOBILIZED, TRAIT_SOURCE_ABILITY("Long-Range Sight"))
+	REMOVE_TRAIT(facehugger, TRAIT_IMMOBILIZED, TRAIT_SOURCE_ABILITY("Long-Range Sight"))
+
+/datum/action/xeno_action/onclick/toggle_long_range/facehugger/on_zoom_in()
+	. = ..()
+
+	var/mob/living/carbon/xenomorph/facehugger/facehugger = owner
+	ADD_TRAIT(facehugger, TRAIT_IMMOBILIZED, TRAIT_SOURCE_ABILITY("Long-Range Sight"))
+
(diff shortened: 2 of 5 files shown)


HeresKozmos:
Watcher's entire gimmick is losing the ability to hide while gaining the ability to move off weeds without taking damage, this PR literally just makes the strain entirely pointless since it removes the main effect of the strain.

InsaneRed:
Theres no good argument as to why a free role that can permanetly change the round of somebody, which cost no resources basically to be moveable off weeds without any consequences, this strain is literally just an upgrade and this is required. hiding or not doesnt matter when youre ghosted and just go watch the afk marine.

Butlerblock:
Horrible design idea, but might cut down on metagaming: What if huggers and lessers had a 20-30 second minimum incubation time inwhich they're forced to sit in hive with no dchat or ghosting. There would be no hive-limit and no ghost-time requirement to start incubation, but would automatically spawn the hugger/lesser once the players ghost-cooldown and the hive's limits become available.

Drulikar:
> Horrible design idea, but might cut down on metagaming: What if huggers and lessers had a 20-30 second minimum incubation time inwhich they're forced to sit in hive with no dchat or ghosting. There would be no hive-limit and no ghost-time requirement to start incubation, but would automatically spawn the hugger/lesser once the players ghost-cooldown and the hive's limits become available.

Not a discussion for here, but if you've read some of Harry and my messages on discord & forums, you'll know that something along the lines of this will be implemented for the entire hive. As in to be in the xenomorph queue, you will become a ghost of some kind tethered to that hive's weeds/xenos. This will include lessers & huggers.

Drulikar (changes requested):
Please instead make this so the watcher hugger has the toggle zoom ability. When that ability is active, health decrease off of weeds halts (but you do not heal either) and you are anchored in place. Otherwise without the toggle_zoom ability you will lose health off weeds just like regular hugger.

code/modules/mob/living/carbon/xenomorph/abilities/facehugger/facehugger_powers.dm
```diff
@@ -26,3 +26,13 @@
...
+
+/datum/action/xeno_action/onclick/toggle_long_range/facehugger/use_ability()
+	. = ..()
+
+	var/mob/living/carbon/xenomorph/facehugger/facehugger = owner
+
+	if (facehugger.is_zoomed)
+		ADD_TRAIT(facehugger, TRAIT_IMMOBILIZED, TRAIT_SOURCE_ABILITY("Long-Range Sight"))
+	else
+		REMOVE_TRAIT(facehugger, TRAIT_IMMOBILIZED, TRAIT_SOURCE_ABILITY("Long-Range Sight"))
```
Drulikar, line 38: Two things:
- There are more ways for the ability to be toggled off than clicking the button (e.g. resting). You need to instead hook this logic instead into `/mob/living/carbon/xenomorph/proc/zoom_in()` and `/mob/living/carbon/xenomorph/proc/zoom_out()`
- TRAIT_IMMOBILIZED does not currently prevent pouncing - so you can just pouce to move around.

code/modules/mob/living/carbon/xenomorph/strains/castes/facehugger/watcher.dm
```diff
@@ -21,5 +21,5 @@
 	name = "Watcher Facehugger Behavior Delegate"
 
 /datum/behavior_delegate/facehugger_watcher/on_life()
-	if(bound_xeno.body_position == STANDING_UP && !(locate(/obj/effect/alien/weeds) in get_turf(bound_xeno)))
+	if(bound_xeno.body_position == STANDING_UP && !(locate(/obj/effect/alien/weeds) in get_turf(bound_xeno)) && !bound_xeno.is_zoomed)
 		bound_xeno.adjustBruteLoss(1)
```
Drulikar, line 25:
```suggestion
	// Sap health if we're standing, not on weeds, and not zoomed out
	if(bound_xeno.body_position != STANDING_UP)
		return
	if(bound_xeno.is_zoomed)
		return
	if(locate(/obj/effect/alien/weeds) in get_turf(bound_xeno))
		return
	bound_xeno.adjustBruteLoss(1)
```
This check has become unwieldly. https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#use-early-returns Also its slightly better to check easy checks before more complicated check, so do the is_zoomed check earlier.

Drulikar (changes requested):

code/modules/mob/living/carbon/xenomorph/abilities/facehugger/facehugger_powers.dm
```diff
@@ -25,4 +25,23 @@
...
+		to_chat(owner, SPAN_WARNING("We cannot do that while immobilized!"))
+		return FALSE
+	
 	return ..()
+
+/datum/action/xeno_action/onclick/toggle_long_range/facehugger/on_zoom_out()
+	. = ..()
+
+	var/mob/living/carbon/xenomorph/facehugger/facehugger = owner
+
```
Drulikar, line 39:
```suggestion
```

code/modules/mob/living/carbon/xenomorph/abilities/ability_helper_procs.dm
```diff
@@ -195,6 +195,10 @@
 			client.pixel_x = -viewoffset
 			client.pixel_y = 0
 
+	for (var/datum/action/xeno_action/onclick/toggle_long_range/action in actions)
```
Drulikar, line 198: This is a weird way to do this, but using `get_xeno_action_by_type` isn't much better so its whatever.

Drulikar (approved):

PR #6240, opened 2024-05-05
Instant delimb nerf

Git-Nivrak:
# About the pull request
Limbs can now only be delimbed if they are fractured

Instant delimbs based on nothing but rng from one slash are unfun to play against
<details>

</details>

# Changelog
:cl:
balance: Limbs will now only be delimb-able if they are fractured
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -367,7 +367,7 @@
 	var/no_perma_damage = owner.status_flags & NO_PERMANENT_DAMAGE
 	var/no_bone_break = owner.chem_effect_flags & CHEM_EFFECT_RESIST_FRACTURE
 	if(previous_brute > 0 && !is_ff && body_part != BODY_FLAG_CHEST && body_part != BODY_FLAG_GROIN && !no_limb_loss && !no_perma_damage && !no_bone_break)
-		if(CONFIG_GET(flag/limbs_can_break) && brute_dam >= max_damage * CONFIG_GET(number/organ_health_multiplier))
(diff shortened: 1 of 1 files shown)


fira:
But can't it still be broken then removed in the same hit with this.....? I'm not sure that really help address the issue

Git-Nivrak:
> But can't it still be broken then removed in the same hit with this.....? I'm not sure that really help address the issue

The way the code is written you need the limb to be at full damage to be fractured and it can't happen at the same instance of damage as the one where it reaches max meaning you will need 2 to get delimbed rather than one

Steelpoint:
Currently it is possible for someone to lose a limb in a minimum of two slashes, with no fracture having occurred on the first slash, this change would increase this to at least three slashes before a delimb is possible.

Drulikar (approved):

TheKillfish:
Post merge comment; experience after this change is that delimbing happens _a lot_ more frequently, and can indeed happen in as few as two hits. More common when being hit by high-damage xenos.

Git-Nivrak:
Hey! Thanks for the feedback however this PR does not increase delimb chance in any way.
This is called the "placebo" nerf effect, one notable example of that was in the first seasom of LoL where a champion was nerfed and everyone kept saying it is unplayable and its winrate dropped, however, it was later found out that the change never went through. Anyhow my point is because of this PR you are probably noticing delimb a lot more when they DO happen. Also note this PR does not nerf delimb chance it just prevents you from being delimbed on a limb with no damage instantly.

PR #6241, opened 2024-05-05
Xenos cannot pylon working comms anymore

ihatethisengine:
# About the pull request
First of all, it looks like a simple oversight. You cannot turn on pyloned comms tower, but for some reason it works the other way around — you can pylon working tower. Don't think this was intended. 

Secondly, I think it's a little too boring that you can get both pylon and working comms if you have corrupted xenos. It would be more interesting if you needed to choose between these options.

Consistency, balance, important choices.

<details>

</details>

# Changelog
:cl: ihatethisengine
balance: only disabled comms can be pyloned
/:cl:

### DIFF
--- a/code/game/machinery/telecomms/presets.dm
+++ b/code/game/machinery/telecomms/presets.dm
@@ -306,14 +306,16 @@ GLOBAL_LIST_EMPTY(all_static_telecomms_towers)
 	else
 		update_icon()
 
+/obj/structure/machinery/telecomms/relay/preset/tower/mapcomms/update_state()
+	..()
+	if(inoperable())
+		handle_xeno_acquisition(get_turf(src))
+
 /// Handles xenos corrupting the tower when weeds touch the turf it is located on
 /obj/structure/machinery/telecomms/relay/preset/tower/mapcomms/proc/handle_xeno_acquisition(turf/weeded_turf)
 	SIGNAL_HANDLER
 
 	if(corrupted)
-		return
-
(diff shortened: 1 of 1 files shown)


HeresKozmos:
Seems pretty cool.

code/game/machinery/telecomms/presets.dm
```diff
@@ -313,6 +313,9 @@ GLOBAL_LIST_EMPTY(all_static_telecomms_towers)
 	if(corrupted)
 		return
 
+	if(!inoperable() || on)
```
Drulikar, line 316:
```suggestion
	if(operable())
```
Not sure I see the point of testing `on` we only care about whether its operable.

Drulikar (changes requested):
I could be wrong, but this is not a repeating call to `handle_xeno_acquisition` (except if those timers are used below). This is the proc called for the signal `COMSIG_WEEDNODE_GROWTH` or a re-call using one of the addtimers a little further down. You likely need to in this situation either add a signal for the tower going inoperable, or for you to add a timer to recheck again later.

ihatethisengine:
What about this?

Drulikar (changes requested):
This works, but means that this proc is now very spammable (e.g. toggling the power on and off at the apc). The timers should then be considered unique and allowed to override one another else you can create lots of timers that are redundant: 
image

Drulikar (approved):

Issue #6246, opened 2024-05-06
Humans can't buckle to vehicle gunner seats


Doubleumc:
### Testmerges

N/A

### Description of the bug

A human attempting to buckle to a vehicle's gunner seat (click-drag) causes the user to move to the seat's tile but does not buckle and does not grant the exterior vision or control over the hardpoints.

Believe the issue is here:
https://github.com/cmss13-devs/cmss13/blob/ff95ae946a2a653dc81d2bd3dccde715e6063f3c/code/modules/vehicles/interior/interactable/seats.dm#L111

Humans do not inherently have `TRAIT_OPPOSABLE_THUMBS`, it is assumed either through a `ishuman()` check or a `IsAdvancedToolUser()` check.

### What's the difference with what should have happened?

Humans should be able to buckle to the gunner seat, allowing control over the hardpoints.

### How do we reproduce this bug?

1. Spawn vehicle with gunner seat (APC, tank)
2. Spawn & inhabit a human that should be able to control gunnery (vehicle crewman)
3. Click-drag to buckle self to gunner seat

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #6034, opened 2024-03-28
Xenothumbs Minor Event Button Comeback

PhantomEpicness:
# About the pull request
Working version that is TMable for april fools. 

Revived and fixed up https://gitlab.com/cmdevs/colonial-warfare/-/merge_requests/1915/diffs#9ec310cca070259d82839a4a70e89dd858ded438_958_958

https://github.com/cmss13-devs/cmss13/pull/2997

Additions from  Vanagandr's original MR
- xenos can use health items on marines
- vendors are usable now, all checks are simply skipped so introducing xenos with thumbs to snowflake vendors is a horrible idea (as if enabling this in the first place wasn't bad enough) [needs further testing]
- xenos can drive
- M2C usable

I think this is pretty much done, it's already exceeding the scope of the original and we're just making xenos human past this point

Original description
> TL;DR: https://streamable.com/rlvs86
> 
> Mainly this allows xenos to pick up items. They can then use most of them.
> 
> Ex. a xeno can't flip a plasteel cade open - but if she has a toolbelt and time to work, she can unwrench it.
> 
> They can fumble about but a lot of QoL and objects (middleclick container open, click-dragging containers, tables, etc.) were never designed to be interacted with by xenos so item interactions are a bit clunky.
> 
> They can't strip clothing and gear from people.
> 
> They *can* strip and attach attachments but can't aim down scopes.
> 
> They *can* FF.
> 
> They *can't* use MGs.
> 
> IFF works, roughly, not picking up on specific hives but allowing them to shoot past other xenos and hit humans.
> 
> They pass spec checks and can clumsily handle medic gear.
> 
> needless to say this is all colossally imbalanced and has potential for fuckery limited only by the xeno's imagination and whether the thing they're messing with makes an explicit ishuman() check. Or is a structure. They can't hand-interact with structures, so they can shove shells into a mortar but can't change its coords.
> 
> Cosmetically, gun inhands don't line up with xenos. Tweaked it a bit to center it horizontally but lining inhands up with hands for every xeno and xenonid sprite is a bit more hassle than seems worthwhile ATM.
> 

# Explain why it's good for the game
TM funny
# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:  Vanagandr, totalepicness
[...]

PR #6247, opened 2024-05-06
Fixes humans unable to use vehicle gunner seat

Doubleumc:
# About the pull request

Fixes https://github.com/cmss13-devs/cmss13/issues/6246

Vehicle gunnery is meant to be used by humans.
<details>

</details>

# Changelog
:cl:
fix: humans can use vehicle gunner seats again
/:cl:


code/modules/vehicles/interior/interactable/seats.dm
```diff
@@ -108,7 +108,7 @@
 				to_chat(user, SPAN_WARNING("You are unable to use heavy weaponry."))
 			return
 
-		if(!HAS_TRAIT(user, TRAIT_OPPOSABLE_THUMBS))
+		if(!user.IsAdvancedToolUser())
```
Drulikar, line 111: Though this likely fixes the problem, you likely want to test both, or just give humans the trait (likely better if there's still other places that PR caused problems). The source of it is because of this toggle that's been a reoccurring april fools PR that was decided to just be incorporated: https://github.com/cmss13-devs/cmss13/pull/6034/files#diff-5d524b4f8747a9837378e330b2ec7dc3be0b3866c42591d0f84d735f4aec52f7

Drulikar (changes requested):

code/modules/vehicles/interior/interactable/seats.dm
Doubleumc: Reverted that logic change and granted all humanoid species `TRAIT_OPPOSABLE_THUMBS`.

Doubleumc:
Somewhere in the mess of `ishuman()`, `IsAdvancedToolUser()`, `allow_gun_usage`, `has_fine_manipulation`, `TRAIT_OPPOSABLE_THUMBS`, `KNOWS_TECHNOLOGY` there's a good solution.

Drulikar (changes requested):
So looking at this closer, there currently appear to be side effects of us giving this trait to humans that we need to rectify. For example, https://github.com/cmss13-devs/cmss13/blob/1f79350ddb991151b55efc83dcd9c831d60cb51a/code/datums/action.dm#L189-L195 now makes it possible for humans to use actions while not standing, when that was not the case before. So this example either should refactor the other case to an else, or return FALSE if not standing. Though this proc could in general be refactored; e.g. both conditions both want is_mob_incapacitated, and `body_position` doesn't need to be for a human, its a living mob variable.

To me it seems like we could refactor this proc to below because I don't see much reason why xenos should be able to do this not standing up either (though I don't even know what actions this allows them to do):
```js
	if(!HAS_TRAIT(owner, TRAIT_OPPOSABLE_THUMBS))
		return FALSE
	if(!isliving(owner))
		return FALSE
	var/mob/living/living_mob = owner
	if(living_mob.body_position != STANDING_UP)
		return FALSE
	return TRUE
```

So please take a look at the uses of `TRAIT_OPPOSABLE_THUMBS` and mitigate situations where a human would then fall into that xeno w/ thumbs case and have less restrictions, or maybe just consider fixing it the opposite way instead.

But maybe this then ought to just be a xeno exclusive trait and we just fix the opposite problems.

Also of note `/mob/living/carbon/xenomorph/IsAdvancedToolUser()` will currently return true for xenos w/ thumbs, but that also makes https://github.com/cmss13-devs/cmss13/blob/1f79350ddb991151b55efc83dcd9c831d60cb51a/code/modules/projectiles/gun.dm#L1528 redundant...

To me it looks like these are currently problematic with the current implementaiton:
- /datum/action/item_action/can_use_action()
- /obj/structure/machinery/cm_vending/ui_act(action, list/params, datum/tgui/ui, datum/ui_state/state) // drops squad restriction
- /obj/structure/machinery/cm_vending/proc/can_access_to_vend(mob/user, display = TRUE, ignore_hack = FALSE) // drops all restrictions

Doubleumc:
I agree there are likely more issues stemming from the April Fools' PR and the tangle of `ishuman()`/`IsAdvancedToolUser()`/`has_fine_manipulation`/`TRAIT_OPPOSABLE_THUMBS` cries out for a refactor, but that is way out of scope of this PR and beyond what I want to tackle at the moment.

What I have right now is:
- add the missing `TRAIT_OPPOSABLE_THUMBS` to the global trait list
- fix missing return in xeno buckling logic, no longer buckle twice
- remove the thumbs check entirely from vehicle seats, thumbless xenos are already prevented from using the seat earlier in the proc chain

So currently humans can use the gunner seat, thumbless xenos can't use the gunner seat, thumbed xenos can use the gunner seat.

Drulikar (approved):

Issue #1764, opened 2022-11-29
Inconsistancies with the Shotgun Sentry.


NNJTM:
## Testmerges:
-[#1546](https://github.com/cmss13-devs/cmss13/pull/1546)
-[#1562](https://github.com/cmss13-devs/cmss13/pull/1562)
-[#750](https://github.com/cmss13-devs/cmss13/pull/750)
-[#1760](https://github.com/cmss13-devs/cmss13/pull/1760)
-[#1224](https://github.com/cmss13-devs/cmss13/pull/1224)

## Description
Watching the Shotgun Sentry from a ghosted position revealed to me that pellet count is Inconsistent from what it should be.  

## Difference between expected and actual behavior
The Shotgun Sentry should shoot regular old Buckshot. 4 pellets, 65 damage each pellet. It seems inconsistent, hitting for less pellets even at point blank range.

## Steps to reproduce
1.) Place shotgun sentry. 
2.) Place enemy in range.
3.) Check logs and witness the inconsistent pellet count.

## Issue bingo (not required to be all checked)

- [x ] Issue could be reproduced at least once
- [x ] Issue happened in a recent (less than 7 days ago) round
- [x ] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Segrain:
Tested this on live server when it was reported. Sentry on thunderdome was shooting all the additional pellets as intended, but logs show sentries in the field proper only ever hitting with the main shell. Needs investigation.

NNJTM:
> Tested this on live server when it was reported. Sentry on thunderdome was shooting all the additional pellets as intended, but logs show sentries in the field proper only ever hitting with the main shell. Needs investigation.

To Clarify, this was on the map LV-642. In case that matters.

usnpeepoo:
On a different note, shotgun sentries are kind of wacky. On a very laggy, high-pop round of LV I was playing runner and triggered the shotgun sentry but did not get immediately shot by it. Almost a minute later, I was carrying a capture and suddenly was hit by the triggered shotgun sentries pellets far away from it and went into crit instantly. I don't know if its a code related issue or a server side issue, but just wanted to say.

Ansekishokuu:
Fixed with #7406

PR #6249, opened 2024-05-06
Fixes bonus projectiles not hitting

### LINKED ISSUE #3053: Buckshot Projectiles Not Hitting Targets Simultaneously
### Testmerges

#2944 , #2964, #3031, #3039 , #3046 , #3040 , #3048 at time of report but this issue has existed for months.

### Description of the bug

When firing a buckshot round from a shotgun, the round has two parts to it: the buckshot slug which causes the stun/slowdown/damage and the additional buckshot that just does basic damage but a lot of it. The issue is that when firing, the buckshot slug hits the target well before the additional buckshot. In most cases this does not matter because it all hits eventually. However, if a person shoots a warrior that has lunged a marine outside more than 1 tile away(not a point blank hit) the warrior is stunned by the slug, releases their target, and the additional buckshot then 'catches up' and hits the warrior's target due to the target instantly standing up. 

The projectiles should hit at the same instant. It makes zero sense that, if you shoot a warrior that has lunged a marine, you are going to hit the marine on the ground with your damage rather than the warrior. I asked in the coding section of the Discord if this was intended and every single person said it was absolutely not intended projectile behavior.

1.See marine lunged by a warrior.
2. Shoot warrior with a buckshot round while not PBing but also so that the slug portion of the round will stun said warrior.
3. Observe the warrior fall stunned while the marine that was lunged soaks all the damage.
4. Seethe.

### Issue Bingo

Git-Nivrak:
# About the pull request
#1608 added random speed variance to projectiles to make them look less beamy however some projectiles and most notably buckshot requires all projectiles to be of equal speed because the main projectile causes a stun which will make the rest of the projectiles miss if they don't hit first.

Potentially fixes #3053 and #1764 

Fixes a bug
<details>

</details>

# Changelog
:cl:
fix: Fixes bonus projectiles not hitting sometimes
/:cl:

### DIFF
--- a/code/datums/ammo/ammo.dm
+++ b/code/datums/ammo/ammo.dm
@@ -236,7 +236,7 @@
 		final_angle += rand(-total_scatter_angle, total_scatter_angle)
 		var/turf/new_target = get_angle_target_turf(curloc, final_angle, 30)
 
-		P.fire_at(new_target, original_P.firer, original_P.shot_from, P.ammo.max_range, P.ammo.shell_speed, original_P.original) //Fire!
+		P.fire_at(new_target, original_P.firer, original_P.shot_from, P.ammo.max_range, P.ammo.shell_speed, original_P.original, FALSE) //Fire!
 
 /datum/ammo/proc/drop_flame(turf/turf, datum/cause_data/cause_data) // ~Art updated fire 20JAN17
 	if(!istype(turf))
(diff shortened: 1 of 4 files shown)


fira:
Well this should work, but I think a better way to go about this would be to add a delay to dodging by lying down, it'd also address similar issues with explosion sharpnel

Drulikar (changes requested):
Try what is proposed by Fira, or indicate that you would still prefer this method.

Drulikar:
I'm not really finding any change with this PR, but maybe I'm not understanding the issue.

PR #6250, opened 2024-05-06
ARC features

Diegoflores31:
# About the pull request
- Adds support for custom access requirements for the driving seat ( something like a key )
- increases the ARC passeger amount to 4
- Adds Command Access lock for the ARC driver.
- Adds a pamphlet crate with driving skill on requisitions.

-Adds more flexibility to the ARC in cases where there may not be 2 SOs available or they preffer a different driver. to focus more on Commanding inside of the ARC.
-Prevents future funnsies from happening . you now require a Access to drive the ARC meaning . only people granted access by command may drive the ARC. (( can be changed to ACCESS_CREWMAN in the future or even create custom "keys"))
-adds one more passenger to the ARC  so marines can retrieve the 3 corpses inside or someone can enter to talk .. etc . the extra slot is meant for a passerby therefore no seat will be added.

<details>

</details>

# Changelog
:cl:
add: ARC now has a Generic Command access lock for the driver seat
add: Adds a a pamplet crate containing 1 pamplet for the ARC for 5000$
balance: increased ARC capacity to 4
code: Added support for Access locks on Vehicles
/:cl:

### DIFF
--- a/code/modules/vehicles/interior/interactable/seats.dm
+++ b/code/modules/vehicles/interior/interactable/seats.dm
@@ -18,7 +18,6 @@
 
 	// Which vehicle skill level required to use this
 	var/required_skill = SKILL_VEHICLE_SMALL
-	req_access = list()
 
 /obj/structure/bed/chair/comfy/vehicle/ex_act()
 	return
@@ -80,8 +79,8 @@
 		if(target == user)
 			to_chat(user, SPAN_WARNING("You have no idea how to drive this thing!"))
 		return FALSE
-	if(!allowed(target))
-		to_chat(user, SPAN_WARNING("The driving console blinks as it refuses your ID, Access Denied"))
+	if(!allowed(target) && !isxeno(target)) // xenos cant hold IDs DUH
+		to_chat(user, SPAN_WARNING("The driving console blinks as it refuses your ID"))
 		return FALSE
 	if(vehicle)
 		vehicle.vehicle_faction = target.faction
--- a/code/modules/vehicles/interior/interactable/seats.dm
+++ b/code/modules/vehicles/interior/interactable/seats.dm
@@ -18,6 +18,7 @@
 
 	// Which vehicle skill level required to use this
 	var/required_skill = SKILL_VEHICLE_SMALL
+	req_access = list()
 
 /obj/structure/bed/chair/comfy/vehicle/ex_act()
 	return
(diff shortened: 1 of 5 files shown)


code/modules/vehicles/interior/interactable/seats.dm
```diff
@@ -79,7 +80,9 @@
 		if(target == user)
 			to_chat(user, SPAN_WARNING("You have no idea how to drive this thing!"))
 		return FALSE
-
+	if(!allowed(target))
+		to_chat(user, SPAN_WARNING("The driving console blinks as it refuses your ID, Access Denied"))
```
Zonespace27, line 84:
```suggestion
		to_chat(user, SPAN_WARNING("The driving console blinks as it refuses your ID."))
```

code/modules/vehicles/interior/interactable/seats.dm
```diff
@@ -18,6 +18,7 @@
 
 	// Which vehicle skill level required to use this
 	var/required_skill = SKILL_VEHICLE_SMALL
+	req_access = list()
```
Zonespace27, line 21: why define the empty list here?

Zonespace27 (changes requested):
> add: Adds a a pamplet crate containing 1 pamplet for the ARC for 5000$

I am entirely against this.

Diegoflores31:
> > add: Adds a a pamplet crate containing 1 pamplet for the ARC for 5000$
> 
> I am entirely against this.

That was added specifically to go along the ID lock on the ARC . there wont be any ARC hijack or that kind of stuff unless command specifically allows it . also opens grounds for other command/support to drive the ARC under specific circumstances winout having to give every command member a VC skill

code/modules/vehicles/multitile/multitile.dm
```diff
@@ -65,7 +65,7 @@
 
 	// The amount of skill required to drive the vehicle
 	var/required_skill = SKILL_VEHICLE_SMALL
-
+	var/driving_access = null
```
Zonespace27, line 68: code doc

Zonespace27:
> That was added specifically to go along the ID lock on the ARC . there wont be any ARC hijack or that kind of stuff unless command specifically allows it . also opens grounds for other command/support to drive the ARC under specific circumstances winout having to give every command member a VC skill

Here are the issues I see with this:
- It's inevitably going to lead to people buying this to fuck around with the van / ARC, which is a lot of wasted cash.
- Having someone who isn't an SO drive the ARC isn't worth 5k req points.
- Outside of the (usually dead) IOs[^1], the only marines who have `ACCESS_MARINE_COMMAND` are the ones not allowed to deploy (MP, CMP, SEA, etc.).
- I still believe this goes against my core design principle of it being "the command vehicle". I've used the ARC a handful of times in an actual round, and I haven't had a problem driving the ARC and still being a member of command at the same time.

[^1]: I also don't want IOs to become the "oh command buy an arc and req spend 5k on a pamphlet so i can drive pleasepleaseplease" role.

PR #6251, opened 2024-05-06
Fixes light attachments setting and add pref for ammo counter

### LINKED ISSUE #6238: Light attachments turning itself back on after being disabled.
### Testmerges

#5401: 'M540-B Armored Recon Carrier' by Zonespace27 at commit ab0e59f1cc #6103: 'Med Vendor Rebalance' by Drulikar at commit 219a08fc20 #5957: 'OT CAS rocket' by iloveloopers at commit 2484c7b012 #6150: 'Falling Falcons: Paratroopers Edition' by ihatethisengine at commit 8fead532be #6146: 'Smoke Flamer' by iloveloopers at commit ca1fe8460d #6178: 'fences can be shot through' by cuberound at commit 5f3e4afc6b #5265: 'WIP - Paperwork Prefabs' by realforest2001 at commit 831b82144d

### Round ID

22307

### Description of the bug

Rail flashlight doesn't stay off, if you toggle it.

It should stay off if you disable it, not turning itself back on

1. Equip your gun, turn on the rail light or any light attachment that gets disabled when its in suit storage
2. Turn the rail light off, holster the gun back in suit storage
3. take it out again, its enabled.
...

### Issue Bingo

vero5123:
# About the pull request
Closes #6238, also made the ammo display less spammy. 

:cl:
qol: Toggleable ammo counter, can now cycle between displaying every fifth or single bullet. 
fix: Light attachments will no longer automatically turn on when the weapon is equipped from suit storage
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -632,7 +632,7 @@ GLOBAL_LIST_INIT(bgstate_options, list(
 			dat += "<b>Toggle Vendors Vending to Hands: \
 					</b> <a href='?_src_=prefs;preference=toggle_prefs;flag=[TOGGLE_VEND_ITEM_TO_HAND]'><b>[toggle_prefs & TOGGLE_VEND_ITEM_TO_HAND ? "On" : "Off"]</b></a><br>"
 			dat += "<b>Toggle Semi-Auto Ammo Display: \
-					</b> <a href='?_src_=prefs;preference=toggle_prefs;flag=[TOGGLE_AMMO_DISPLAY_TYPE]'><b>[toggle_prefs &= TOGGLE_AMMO_DISPLAY_TYPE ? "On" : "Off"]</b></a><br>"
+					</b> <a href='?_src_=prefs;preference=toggle_prefs;flag=[TOGGLE_AMMO_DISPLAY_TYPE]'><b>[toggle_prefs & TOGGLE_AMMO_DISPLAY_TYPE ? "On" : "Off"]</b></a><br>"
 			dat += "<a href='?src=\ref[src];action=proccall;procpath=/client/proc/switch_item_animations'>Toggle Item Animations Detail Level</a><br>"
 			dat += "<a href='?src=\ref[src];action=proccall;procpath=/client/proc/toggle_dualwield'>Toggle Dual Wield Functionality</a><br>"
 		if(MENU_SPECIAL) //wart
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -631,7 +631,7 @@ GLOBAL_LIST_INIT(bgstate_options, list(
 					</b> <a href='?_src_=prefs;preference=toggle_prefs;flag=[TOGGLE_MIDDLE_MOUSE_SWAP_HANDS]'><b>[toggle_prefs & TOGGLE_MIDDLE_MOUSE_SWAP_HANDS ? "On" : "Off"]</b></a><br>"
 			dat += "<b>Toggle Vendors Vending to Hands: \
 					</b> <a href='?_src_=prefs;preference=toggle_prefs;flag=[TOGGLE_VEND_ITEM_TO_HAND]'><b>[toggle_prefs & TOGGLE_VEND_ITEM_TO_HAND ? "On" : "Off"]</b></a><br>"
-			dat += "<b>Toggle Semi-Auto Ammo Display: \
+			dat += "<b>Toggle Semi-Auto Ammo Display Limiter: \
 					</b> <a href='?_src_=prefs;preference=toggle_prefs;flag=[TOGGLE_AMMO_DISPLAY_TYPE]'><b>[toggle_prefs & TOGGLE_AMMO_DISPLAY_TYPE ? "On" : "Off"]</b></a><br>"
 			dat += "<a href='?src=\ref[src];action=proccall;procpath=/client/proc/switch_item_animations'>Toggle Item Animations Detail Level</a><br>"
 			dat += "<a href='?src=\ref[src];action=proccall;procpath=/client/proc/toggle_dualwield'>Toggle Dual Wield Functionality</a><br>"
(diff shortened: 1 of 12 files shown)


ihatethisengine:
I don't think ammo counter change is needed unless you make it a setting.

vero5123:
> I don't think ammo counter change is needed unless you make it a setting.

can do.

vero5123:
Can toggle it now.

code/modules/projectiles/gun.dm
```diff
@@ -1608,6 +1610,9 @@ not all weapons use normal magazines etc. load_into_chamber() itself is designed
 		user = gun_user
 
 	if(flags_gun_features & GUN_AMMO_COUNTER && current_mag)
+		// toggleable spam control.
+		if(flags_gun_features & GUN_LIMIT_AMMO_DISPLAY && current_mag.current_rounds % 5 != 0 && current_mag.current_rounds > 15)
```
Drulikar, line 1614: This is wonky with burst fire if the burst doesn't end on a multiple of 5, and though understandable you then start displaying per round when the mag is starting to go dry - it just feels like its randomly bugged out.

You should have this either send a message ever we *pass* the threshold, or just always on a burst. You could also convey that the preference only works when the mag count is above 15.

Drulikar (changes requested):
In the future fill out the PR template; don't just omit it.

Drulikar (changes requested):
As mentioned on discord:
- This would be better as a player preference rather than something you need to set per weapon IMO
- I have no idea where verb desc even gets displayed for us, so conveying the < 15 bullet information needs to also be somewhere else; or just make it a different option for the pref (e.g. every bullet, every 5 bullets unless low, every 5 bullets)
- Additionally, you should mention this only applies to semi fire (e.g. make the preference called semi-auto ammo counter)

code/modules/client/preferences_toggles.dm
```diff
@@ -264,6 +264,7 @@
 		"<a href='?src=\ref[src];action=proccall;procpath=/client/proc/toggle_help_intent_safety'>Toggle Help Intent Safety</a><br>",
 		"<a href='?src=\ref[src];action=proccall;procpath=/client/proc/toggle_auto_eject'>Toggle Guns Auto-Ejecting Magazines</a><br>",
 		"<a href='?src=\ref[src];action=proccall;procpath=/client/proc/toggle_auto_eject_to_hand'>Toggle Guns Auto-Ejecting Magazines to Your Hands</a><br>",
+		"<a href='?src=\ref[src];action=proccall;procpath=/client/proc/toggle_ammo_display_type'>Semi-auto ammo counter</a><br>",
```
Drulikar, line 267: These toggles normally also exist in https://github.com/cmss13-devs/cmss13/blob/5da327a71f6da77ba0d82e71d4aba80f9a42166a/code/modules/client/preferences.dm#L621

code/modules/projectiles/gun.dm
```diff
@@ -1608,6 +1610,9 @@ not all weapons use normal magazines etc. load_into_chamber() itself is designed
 		user = gun_user
 
 	if(flags_gun_features & GUN_AMMO_COUNTER && current_mag)
+		// toggleable spam control.
+		if(user.client.prefs.toggle_prefs & TOGGLE_AMMO_DISPLAY_TYPE && current_mag.current_rounds % 5 != 0 && current_mag.current_rounds > 15 && gun_firemode != GUN_FIREMODE_BURSTFIRE)
```
Drulikar, line 1614:
```suggestion
		if(user.client.prefs.toggle_prefs & TOGGLE_AMMO_DISPLAY_TYPE && gun_firemode == GUN_FIREMODE_SEMIAUTO && current_mag.current_rounds % 5 != 0 && current_mag.current_rounds > 15)
```
Equal to because by not working for full auto I mean that you would have to change the logic to be when you *pass* the threshold to work properly. At that point it could be used for all fire modes, but done this way it can only work for semi auto.

Drulikar (changes requested):

code/modules/projectiles/gun.dm
vero5123: Sorry, I was going on the previous suggestion
> You should have this either send a message ever we pass the threshold, or just always on a burst. You could also convey that the preference only works when the mag count is above 15.

code/modules/client/preferences.dm
```diff
@@ -631,6 +631,8 @@ GLOBAL_LIST_INIT(bgstate_options, list(
 					</b> <a href='?_src_=prefs;preference=toggle_prefs;flag=[TOGGLE_MIDDLE_MOUSE_SWAP_HANDS]'><b>[toggle_prefs & TOGGLE_MIDDLE_MOUSE_SWAP_HANDS ? "On" : "Off"]</b></a><br>"
 			dat += "<b>Toggle Vendors Vending to Hands: \
 					</b> <a href='?_src_=prefs;preference=toggle_prefs;flag=[TOGGLE_VEND_ITEM_TO_HAND]'><b>[toggle_prefs & TOGGLE_VEND_ITEM_TO_HAND ? "On" : "Off"]</b></a><br>"
+			dat += "<b>Toggle Semi-Auto Ammo Display: \
+					</b> <a href='?_src_=prefs;preference=toggle_prefs;flag=[TOGGLE_AMMO_DISPLAY_TYPE]'><b>[toggle_prefs &= TOGGLE_AMMO_DISPLAY_TYPE ? "On" : "Off"]</b></a><br>"
```
Drulikar, line 635:
```suggestion
					</b> <a href='?_src_=prefs;preference=toggle_prefs;flag=[TOGGLE_AMMO_DISPLAY_TYPE]'><b>[toggle_prefs & TOGGLE_AMMO_DISPLAY_TYPE ? "On" : "Off"]</b></a><br>"
```
Runtimes otherwise.

code/modules/client/preferences.dm
```diff
@@ -631,6 +631,8 @@ GLOBAL_LIST_INIT(bgstate_options, list(
 					</b> <a href='?_src_=prefs;preference=toggle_prefs;flag=[TOGGLE_MIDDLE_MOUSE_SWAP_HANDS]'><b>[toggle_prefs & TOGGLE_MIDDLE_MOUSE_SWAP_HANDS ? "On" : "Off"]</b></a><br>"
 			dat += "<b>Toggle Vendors Vending to Hands: \
 					</b> <a href='?_src_=prefs;preference=toggle_prefs;flag=[TOGGLE_VEND_ITEM_TO_HAND]'><b>[toggle_prefs & TOGGLE_VEND_ITEM_TO_HAND ? "On" : "Off"]</b></a><br>"
+			dat += "<b>Toggle Semi-Auto Ammo Display: \
```
Drulikar, line 634:
```suggestion
			dat += "<b>Toggle Semi-Auto Ammo Display Limiter: \
```
Or `Type` its a bit ambiguous at the moment.

image

Drulikar (changes requested):

Drulikar (approved):

code/modules/client/preferences.dm
vero5123: Yeah, should have tested.

PR #6259, opened 2024-05-07
lets you dig out shrapnel with the throwing knife (+ screwdriver)

neeshacark:
# About the pull request

does what it says on the tin, found this to be mildly irritating that you couldn't do it

mostly a convenience thing, lets marines eschew the normal bayonet boot knife for a throwable if they prefer without having to carry the bayonet in their webbing or rip the one off their gun every time. 

<details>

</details>

# Changelog

:cl:
add: you can now remove shrapnel from your body with the M11 Throwing Knife and screwdrivers
/:cl:

### DIFF
--- a/code/game/objects/items/tools/maintenance_tools.dm
+++ b/code/game/objects/items/tools/maintenance_tools.dm
@@ -50,6 +50,7 @@
 	throw_range = 5
 	matter = list("metal" = 75)
 	attack_verb = list("stabbed")
+	flags_item = CAN_DIG_SHRAPNEL
 	inherent_traits = list(TRAIT_TOOL_SCREWDRIVER)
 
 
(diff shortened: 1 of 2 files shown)


HumiliatedGoblin:
Can we also use screwdrivers with a bigger chance to cause pain

realforest2001:
I would be fine with adding screwdrivers / scalpels to this. Might give CLF a chance to survive.

neeshacark:
> I would be fine with adding screwdrivers / scalpels to this. Might give CLF a chance to survive.

if i recall scalpels can already dig shrapnel

i could add screwdrivers but im not a coder, one line changes and mapping related code changes are about all i can do - so the whole more pain thing i could not do

realforest2001:
I took a glance through the code and it wouldn't be too difficult to amplify pain for the screwdriver itself, however digging out shrapnel doesn't actually cause real pain, it just makes an emote happen so no need to worry about it.

neeshacark:
> I took a glance through the code and it wouldn't be too difficult to amplify pain for the screwdriver itself, however digging out shrapnel doesn't actually cause real pain, it just makes an emote happen so no need to worry about it.

alright, you can do it with screwdrivers now too

Drulikar (approved):

PR #6267, opened 2024-05-09
VP78 Range + ROF Buff

Steelpoint:
# About the pull request

This PR increases the rate of fire of the VP-78 to something comparable, but slightly slower, than the Mod88 Pistol. Primarily this is significantly buffing the weapon's single-shot mode to be comparable to the rate of fire of the burst firing mode, it also very slightly increases the burst fire's rate of fire. The idea here is to ensure the single-fire is viable to use alongside the burst-fire, wherein currently the single-fire is objectively inferior to the burst fire. 

This PR also increases the minimum damage falloff range from 0 tiles to 3 tiles, meaning the weapon will now experience falloff from 3 tiles onwards instead of immediately upon exiting the barrel. This is to make this weapon function on-par with almost every other firearm in the game. 

Born out of this discussion thread: https://forum.cm-ss13.com/t/make-vp78-not-dogshit/8028

In universe the VP78 is billed as the imminent replacement sidearm of the USCMC, and it is also the standard issue sidearm of the Three World Empire's Commando forces, and for the highly placed Corporate PMC forces. 

In practice the weapon is completely unusable and is widely seen as being useless as a firearm. It suffers from a myriad of issues, such as noted above how the firearm experiences damage falloff immediately as soon as the bullet leaves the gun, meaning you will only deal the full, advertised, damage to a target if you are sprite-click point-blank shooting a target.

By the time your bullet is 2 to 3 tiles away from you, you are already dealing less than half your expected damage. 

Its accuracy is suboptimal as it struggles to strike long range targets (even if you are already dealing 20 to 10 damage per bullet at this range) and its burst fire mode is prone to wild inaccuracy.

Furthermore, its rate of fire is so slow that assuming you are in a white room perfect testing conditions environment, you can spend at minimum of 20 seconds, often far longer, to kill a single, T2 offensive, xeno after expending multiple magazines wherein any other firearm has done so within 5 seconds and with only one magazine. 

Because the weapon seems billed as a very powerful pistol, its resupply is constrained to represent this, meaning most Marine roles need to buy ammo from their personal vendor, and requisitions will not purchase, or even bother to send a, resupply crate or spare ammo. 

The summation is that this weapons in-universe predecessor, the Mod88, is heads above a superior firearm in virtually every single possible respect to the VP78. Furthermore, the M4A3 pistol, a decidedly even more ancient firearm, is also superior of a firearm to this .  

This change is aimed at making the VP78 an actually viable firearm. With damage output similar to some primary weapons, and at least a similar damage output to the other good pistols, but with a harsh ammo supply limit.

I am 100% open to making needed balance changes to make this merge ready, but I assessed this is a good starting spot as the weapon needs something to make it more than a fluff piece. 

<details>

</details>

# Changelog
:cl:
balance: The VP78 pistol has received a rate of fire boost, and it will deal full damage up to 3 tiles from the shooter's position before suffering damage falloff. 
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -715,9 +715,9 @@
 
 /obj/item/weapon/gun/pistol/vp78/set_gun_config_values()
 	..()
-	set_fire_delay(FIRE_DELAY_TIER_4)
+	set_fire_delay(FIRE_DELAY_TIER_11)
 	set_burst_amount(BURST_AMOUNT_TIER_3)
-	set_burst_delay(FIRE_DELAY_TIER_11)
+	set_burst_delay(FIRE_DELAY_TIER_10)
 	accuracy_mult = BASE_ACCURACY_MULT
 	accuracy_mult_unwielded = BASE_ACCURACY_MULT
 	scatter = SCATTER_AMOUNT_TIER_6
@@ -726,7 +726,6 @@
 	damage_mult = BASE_BULLET_DAMAGE_MULT
 	recoil = RECOIL_AMOUNT_TIER_5
 	recoil_unwielded = RECOIL_AMOUNT_TIER_4
-
 
 //-------------------------------------------------------
 /*
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -717,7 +717,7 @@
 	..()
 	set_fire_delay(FIRE_DELAY_TIER_7)
 	set_burst_amount(BURST_AMOUNT_TIER_3)
-	set_burst_delay(FIRE_DELAY_TIER_10)
+	set_burst_delay(FIRE_DELAY_TIER_11)
 	accuracy_mult = BASE_ACCURACY_MULT
 	accuracy_mult_unwielded = BASE_ACCURACY_MULT
 	scatter = SCATTER_AMOUNT_TIER_6
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -727,6 +727,7 @@
 	recoil = RECOIL_AMOUNT_TIER_5
 	recoil_unwielded = RECOIL_AMOUNT_TIER_4
 
+
 //-------------------------------------------------------
 /*
 Auto 9 The gun RoboCop uses. A better version of the VP78, with more rounds per magazine. Probably the best pistol around, but takes no attachments.
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -715,7 +715,7 @@
 
 /obj/item/weapon/gun/pistol/vp78/set_gun_config_values()
 	..()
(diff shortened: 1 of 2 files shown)


nerocavalier:
The rate of fire buff isn't needed, the 3 tiles before falloff is enough to make it into a solid pistol.

Steelpoint:
Burst ROF is almost identical to prior
 Single shot ROF is higher so its closer in fore rate to brust

HumiliatedGoblin:
> Burst ROF is almost identical to prior Single shot ROF is higher so its closer in fore rate to brust

@Steelpoint what did you mean by this

Steelpoint:
I said the PR provides a large buff to the single shot ROF and only a small one to the burst ROF. Burst always had a higher ROF but fairly poor accuracy. The single shot buff makes single shot actually viable instead of a bad choice

Steelpoint:
For transparency I'll repost some stat testing I did on the forums

----

My personal design philosophy is I’d rather over-buff something to start with and work down during the review process, than be timid to start with and effect a change that has only a minor effect.

In either case, I can give you more raw stats about how the weapons functions.

Basically TL;DR, my proposed fire rate achieves a fairly balanced kill time, in white room conditions, compared to other mainline secondary and primary weapons. The VP78 will still suffer from the fact it has the harshest damage falloff of most guns in the game (even mod88 or M4A3 still deals most of its damage up to 7 tiles whereas the buffed VP78 will still lost half its damage by then) and that the VP78 is much harder to keep resupplied in a long fight due to its ammo being non-standard issue.

----

(all tests conducted against Warrior at 3 tiles distance, unless stated otherwise these all took a single magazine)

VP78 (With No Damage Falloff to 3 Tiles but with current Rate of Fire)
- Time To Kill Single Shot -: 13 Seconds
- Time To Kill: Burst Fire -: 8 Seconds

VP78 (No Falloff plus increased Rate of Fire)
- Time to Kill Single Shot -: 5 Seconds
- Time To Kill Burst Fire -: 6 Seconds (Sometimes 8 due to accuracy debuff)
- 
For Comparison to other weapons

M41A Mk2
- Time To Kill Burst or Single Fire -: 5-6 Seconds

Mod88
- Time To Kill Single Shot -: 7-8 Seconds
- Time To Kill Burst Fire -: 6-7 Seconds

M4A3 (takes 1.5 magazines, speed heavily based on how fast you can reload, if you use macros you could achieve sub-3 second kill times)
- Time To Kill Single Shot -: 4-5 Seconds

code/modules/projectiles/guns/pistols.dm
```diff
@@ -715,9 +715,9 @@
 
 /obj/item/weapon/gun/pistol/vp78/set_gun_config_values()
 	..()
-	set_fire_delay(FIRE_DELAY_TIER_4)
+	set_fire_delay(FIRE_DELAY_TIER_11)
```
Drulikar, line 718:
```suggestion
	set_fire_delay(FIRE_DELAY_TIER_7)
```
Try something more like this which should achieve ~9s based on your numbers.

Drulikar (changes requested):

Drulikar (approved):

PR #6272, opened 2024-05-10
Additional Labeler Workflow Triggers

Drulikar:
# About the pull request

This PR simply expands the trigger for the labeler workflow. I would like the workflow to be required, but as is it simply blocked the merge queue when the `label` was made into a required check.

# Changelog

No player facing changes.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,7 +1,7 @@
 name: Labeling and Verification
 on:
   pull_request_target:
-    types: [opened, edited]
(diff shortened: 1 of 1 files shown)


harryob (approved):
one best way to test action prs right

Drulikar:
Went back to sync method via https://github.com/cmss13-devs/cmss13/commit/b7722a36826211e6e756ae60e301a8a92203b496 this won't allow me to make it required, but atleast it should always be listed in checks rather than poofing.

PR #6274, opened 2024-05-10
accessory removal menu is now radial

AndroBetel:
# About the pull request

title

easier to use and prettier, visual = good

<details>

Screenshot_43
Screenshot_44

</details>

# Changelog

:cl:
qol: Added radial menu to accessory removal
/:cl:

### DIFF
--- a/code/modules/clothing/clothing_accessories.dm
+++ b/code/modules/clothing/clothing_accessories.dm
@@ -132,13 +132,20 @@
 		return
 	var/obj/item/clothing/accessory/A
 	var/list/removables = list()
+	var/list/choice_to_accessory = list()
 	for(var/obj/item/clothing/accessory/ass in accessories)
-		if(ass.removable)
-			removables |= ass
-	if(LAZYLEN(accessories) > 1)
-		A = tgui_input_list(usr, "Select an accessory to remove from [src]", "Remove accessory", removables)
+		if(!ass.removable)
+			continue
+		var/capitalized_name = capitalize_first_letters(ass.name)
+		removables[capitalized_name] = image(icon = ass.icon, icon_state = ass.icon_state)
+		choice_to_accessory[capitalized_name] = ass
+
+	if(LAZYLEN(removables) > 1)
+		var/use_radials = usr.client.prefs?.no_radials_preference ? FALSE : TRUE
+		var/choice = use_radials ? show_radial_menu(usr, src, removables, require_near = TRUE) : tgui_input_list(usr, "Select an accessory to remove from [src]", "Remove accessory", removables)
+		A = choice_to_accessory[choice]
 	else
-		A = LAZYACCESS(accessories, 1)
+		A = choice_to_accessory[removables[1]]
 	if(!usr.Adjacent(src))
 		to_chat(usr, SPAN_WARNING("You're too far away!"))
 		return
--- a/code/modules/mob/living/carbon/human/human_stripping.dm
+++ b/code/modules/mob/living/carbon/human/human_stripping.dm
@@ -131,9 +131,25 @@ GLOBAL_LIST_INIT(strippable_human_items, create_strippable_list(list(
 	var/obj/item/clothing/under/uniform = sourcemob.w_uniform
 	if(!LAZYLEN(uniform.accessories))
 		return FALSE
-	var/obj/item/clothing/accessory/accessory = LAZYACCESS(uniform.accessories, 1)
+
+	var/obj/item/clothing/accessory/accessory
+	var/list/removables = list()
+	var/list/choice_to_accessory = list()
+
+	for(var/obj/item/clothing/accessory/acc in uniform.accessories)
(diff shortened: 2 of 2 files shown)


code/modules/mob/living/carbon/human/human_stripping.dm
```diff
@@ -131,9 +131,25 @@ GLOBAL_LIST_INIT(strippable_human_items, create_strippable_list(list(
...
+		choice_to_accessory[capitalized_name] = acc
+
 	if(LAZYLEN(uniform.accessories) > 1)
-		accessory = tgui_input_list(user, "Select an accessory to remove from [uniform]", "Remove accessory", uniform.accessories)
+		var/use_radials = usr.client.prefs?.no_radials_preference ? FALSE : TRUE
+		var/choice = use_radials ? show_radial_menu(user, source, removables, require_near = TRUE) : tgui_input_list(user, "Select an accessory to remove from [src]", "Remove accessory", removables)
+		accessory = choice_to_accessory[choice]
+	else
+		accessory = choice_to_accessory[removables[1]]
+
```
ihatethisengine, line 152: Add a check here checking for if the target adjacent. Yeah radials have require_near but with tgui you can abuse this by calling the menu when adjacent and then choosing an accessory when not

code/modules/clothing/clothing_accessories.dm
```diff
@@ -132,13 +132,20 @@
 		return
 	var/obj/item/clothing/accessory/A
```
Drulikar, line 133: Please rename this to `picked_accessory` or `accessory` or something like that like you have in human_stripping. Also consider just making most of this a dedicated proc like `pick_remove_accessory` since you have the same code in two places.

Drulikar (changes requested):

code/modules/mob/living/carbon/human/human_stripping.dm
```diff
@@ -129,13 +129,9 @@ GLOBAL_LIST_INIT(strippable_human_items, create_strippable_list(list(
...
 	var/obj/item/clothing/under/uniform = sourcemob.w_uniform
-	if(!LAZYLEN(uniform.accessories))
-		return FALSE
-	var/obj/item/clothing/accessory/accessory = LAZYACCESS(uniform.accessories, 1)
-	if(LAZYLEN(uniform.accessories) > 1)
-		accessory = tgui_input_list(user, "Select an accessory to remove from [uniform]", "Remove accessory", uniform.accessories)
-	if(!istype(accessory))
-		return
+
+	var/obj/item/clothing/accessory/accessory = uniform.pick_accessory_to_remove(user, sourcemob)
```
Drulikar, line 133: It is not necessarily guaranteed `accessory` will be set to anything here (normally the UI won't offer this button if there isn't one - but better to have the handling), so you need to return here if nothing gets picked.

code/modules/clothing/clothing_accessories.dm
```diff
@@ -124,28 +124,39 @@
 	set name = "Remove Accessory"
 	set category = "Object"
 	set src in usr
-	if(!isliving(usr))
+
+	remove_accessory(usr, pick_accessory_to_remove(usr, usr))
+	if(!LAZYLEN(accessories))
+		verbs -= /obj/item/clothing/proc/removetie_verb
```
Drulikar, line 130: The verb removal probably ought to just be `remove_accessory`; but it also has a flaw as is.

So not only are you not performing this if we are stripping an accessory via `/datum/strippable_item/mob_item_slot/jumpsuit/alternate_action(atom/source, mob/user)` but the verb is only added if the accessory is removeable. So it would be more correct to remove the verb in `remove_accessory` if `accessories` is empty, or has no accessories that have `removable` true (I don't think we currently have any that aren't removable though but we might as well support the functionality since it was bothered to be implemented at least partially).

image

Drulikar (changes requested):
Otherwise appears to be working.

AndroBetel:
stale go away

GrrrKitten:
my hero

code/modules/clothing/clothing_accessories.dm
```diff
@@ -118,34 +118,51 @@
 
 	A.on_removed(user, src)
 	LAZYREMOVE(accessories, A)
+	var/list/removables = list()
+	for(var/obj/item/clothing/accessory/accessory in accessories)
+		if(accessory.removable)
+			removables += accessory
+
+	if(!LAZYLEN(removables))
+		verbs -= /obj/item/clothing/proc/removetie_verb
```
Drulikar, line 127:
```suggestion

	var/any_removable = FALSE
	for(var/obj/item/clothing/accessory/accessory in accessories)
		if(accessory.removable)
			any_removable = TRUE
			break
	if(!any_removable)
		verbs -= /obj/item/clothing/proc/removetie_verb
```

Drulikar (approved):

PR #6275, opened 2024-05-10
Allows maintenance jack to be quick drawn

Git-Nivrak:
# About the pull request
Allows maintenance jack to be quick drawn as well as adding a flag that allows to replicate this behavior for other items.

Quality of life, Requested by a player
<details>

</details>

# Changelog
:cl:
qol: Maintenance jack can now be quick drawn
code: Added a flag that allows items to be quick draw-able despite not being weapons.
/:cl:


code/modules/projectiles/gun_helpers.dm
```diff
@@ -518,10 +518,10 @@ DEFINES in setup.dm, referenced here.
 
 	if(istype(slot) && (slot.storage_flags & STORAGE_ALLOW_QUICKDRAW))
 		for(var/obj/cycled_weapon in slot.return_inv())
-			if(isweapon(cycled_weapon))
+			if(isweapon(cycled_weapon) || cycled_weapon.flags_atom & QUICK_DRAWABLE)
```
Zonespace27, line 521: Honestly just remove the weapon check and slap the quickdraw flag on all weapons

Zonespace27 (changes requested):

code/modules/projectiles/gun_helpers.dm
Git-Nivrak: yea just thought of that

code/game/objects/items/tools/maintenance_tools.dm
```diff
@@ -511,6 +511,7 @@
 
 /obj/item/maintenance_jack/Initialize()
 	. = ..()
+	flags_atom |= QUICK_DRAWABLE
```
Zonespace27, line 514: why not just put this in the definition itself? Same can be argued for doing the same for `/obj/item/weapon` but i'm slightly less pushy on that one

Zonespace27 (changes requested):

code/game/objects/items/tools/maintenance_tools.dm
Git-Nivrak: To not override other flags by accident

Drulikar (changes requested):
All overrides to flags_atom need to be looked at.

For example `/obj/item/weapon/gun/launcher/rocket` can normally go into your suit slot and you can press E to draw it, but since `/obj/item/weapon/gun` overrides flags_atom it now no longer does so since you need the `QUICK_DRAWABLE` flag, and it doesn't have it.

You may consider using [`::`](https://www.byond.com/docs/ref/#/operator/::) to |= existing flags in definitions; but honestly there aren't that many to fix. The regex to find all weapon definitions in this scenario is probably something like `^/obj/item/weapon/[\w/]*?$((.|\n)(?!^/))*?flags_atom ?=(?!^/)` 
image

Drulikar (approved):

PR #6280, opened 2024-05-13
Removes Generic Tailstab for Vamp lurker

Diegoflores31:
# About the pull request

Removes vampire lurket tail stab ability as it was intended in the original design.

Tail stab was not supposed to be on the kit of Vamp lurker . the reason being is that Tail jab is pretty much the same ( tail stab did not exist when vampo lurker was designed).
it provides vamp with an insane burst of damage if they movestack both abilities which no good.

<details>

</details>

# Changelog
:cl:

balance: Removes Generic tail stab from Vampire lurker
spellcheck: adjusted Vampire lurker description
/:cl:


Drulikar (approved):

Drulikar:
This may sway it into the too weak category compared to base strain but I guess we'll see. No need to do further adjustments based on this - just giving my thoughts.

PR #6291, opened 2024-05-16
Roundstart xenomorph count changes

Git-Nivrak:
# About the pull request
Changes roundstart xeno amounts to try to keep a 1:3 to 1:4 xeno to combat role ratio

https://imgur.com/5dSXlmX
This is just an example but this image is taken pre-drop and there are 23 (25 if you count IOs who do not take part on the front for the most part) compared to 12 xenos (one died to pred), the goal was always to keep a 1:3 or 1:4 ratio and it is currently not working and causing lowpop rounds to be onesided steamrolls.
Every time marines win in lowpop you hear after the round from xenos "we had 7 burrowed" so yeah something here is wrong.
<details>

</details>

# Changelog
:cl:
balance: Latejoin marines to larva ratio changed to 1:4 for the first 15 minutes, going back to 1:2.5 afterwards.
/:cl:

### DIFF
--- a/code/game/jobs/role_authority.dm
+++ b/code/game/jobs/role_authority.dm
@@ -259,7 +259,7 @@ I hope it's easier to tell what the heck this proc is even doing, unlike previou
 	var/list/roles_left = list()
 	var/assigned = 0
 	for(var/priority in HIGH_PRIORITY to LOW_PRIORITY)
-		// Assigning xenos first
+		// Assigning xenos first.
 		assigned += assign_initial_roles(priority, roles_for_mode & GLOB.ROLES_XENO, unassigned_players)
 		// Assigning special roles second. (survivor, predator)
 		assigned += assign_initial_roles(priority, roles_for_mode & (GLOB.ROLES_WHITELISTED|GLOB.ROLES_SPECIAL), unassigned_players)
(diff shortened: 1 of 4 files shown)


code/game/jobs/role_authority.dm
```diff
@@ -259,8 +259,8 @@ I hope it's easier to tell what the heck this proc is even doing, unlike previou
 	var/list/roles_left = list()
 	var/assigned = 0
 	for(var/priority in HIGH_PRIORITY to LOW_PRIORITY)
-		// Assigning xenos first.
-		assigned += assign_initial_roles(priority, roles_for_mode & GLOB.ROLES_XENO, unassigned_players)
+		// Assigning xenos first, should not be counted
+		assign_initial_roles(priority, roles_for_mode & GLOB.ROLES_XENO, unassigned_players)
```
Drulikar, line 263: What makes you think this? This contributes to the `GLOB.players_preassigned` when then is used to `set_spawn_positions` where the quantity of xenos is set to be 1:4. If say there were 100 players ready and it could assign all of them, `assign_initial_roles` should return the value 100. Now when you call `set_spawn_positions` for xeno jobs, that count gets multiplied against .25 meaning we ought to have 25 xenos. 25/100 is 1:4. If you instead don't count them, and lets say 25 would get assigned to xeno for `assign_initial_roles` then the count will be 75. 75 * .25 will be 18.75. So then `set_spawn_positions` will make 18/100 positions which is not 1:4.

Drulikar (changes requested):

Drulikar (approved):

PR #6292, opened 2024-05-16
DS equipment is unslashable and unacidable when installed

cuberound:
# About the pull request

Truns on unslashable and unacidable flags when DS equipment is installed on DS.

Makes split drops less punishing, prevents first drop equipment camping a sneaky melting, while keeping the main reason why they are slashable and meltable in the first place, that being to prevent them being used as baricades. If you think that those are risky for xenos, they are not as long as lesser drones are in the game.

<details>

</details>

# Changelog
:cl:
balance: dropship equipment can not be slashed or melted while it is installed on the DS
/:cl:

### DIFF
--- a/code/modules/cm_marines/dropship_equipment.dm
+++ b/code/modules/cm_marines/dropship_equipment.dm
@@ -20,6 +20,8 @@
 	var/point_cost = 0 //how many points it costs to build this with the fabricator, set to 0 if unbuildable.
 	var/skill_required = SKILL_PILOT_TRAINED
 	var/combat_equipment = TRUE
+	unslashable = FALSE
+	unacidable = FALSE
 
 
 /obj/structure/dropship_equipment/Destroy()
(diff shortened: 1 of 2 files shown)


cuberound:
> I feel the equipment shouldn’t be acid proof/invincible, but rather tougher and takes more time to acid. Because the DS should be punished should it land in the wrong spot, or if marines leave FOB open imo.
> 
> giving it 4x health and acid time should do it imo.

if marines leave FOB you can punish them in other ways then this, melt supplies kill doctors... more rewarding and engaging for both sides that clicking stationary stuff that on come maps is in place where none has reason to be

Drulikar (changes requested):
I really do not understand the reasons given. How does this make split drops *more* punishing. Do you mean less?

I also do not think it necessary to make them immune to damage if equipped. Equipment can be damaged on vehicles - dropships should be no different. Just make them healthier/higher tier so its more of a commitment/require more/stronger acid.

`/obj/proc/get_applying_acid_time()` affects how long acid takes to apply
`/turf/proc/can_be_dissolved()` affects the tier of acid to use but is for turfs, so you'd need to implement a similar thing for objects too in `/mob/living/carbon/xenomorph/proc/corrosive_acid(atom/O, acid_type, plasma_cost)` if you want to restrict it to stronger acid types.

I recommend doing something similar like in that above proc does for `if(istype(O, /obj/vehicle/multitile))` where the acid gets temporarily applied. Just being under the effect of acid at all shouldn't mean it will be destroyed, but if its health drops to 0 while under that effect then it is destroyed, so more health would require multiple uses if you implement the temporary effect like for vehicles.

PR #6300, opened 2024-05-17
Maintenance Jack pry time decreased on unpowered doors

Nomoresolvalou:
# About the pull request

This changes the maintenance jack crowbar mode to now open and close unpowered doors just as fast as a normal crowbar, and allows it to pry open doors closed. powered doors still take three seconds to pry open.

As it stands, the maintenance jack is a hard sell for synthetics. the draw of the jack is that it is two tools bundled into one, with benefits for each mode not found on the individual tools (wrench can open bolts, crowbar can pry powered doors). The problem with this is that the crowbar takes forever to open closed and unpowered doors (one of the most common uses of the crowbar) which makes it practically obsolete compared to the normal crowbar. Additionally, the jack can't pry doors closed, a trait exclusive to the jack. This means that not only does it take forever to pry an unpowered door open, but you also can't pry it back closed again. This change should lead to more diversity in synthetic loadouts, as now you aren't using your precious experimental vendor points on a wrench and a straight up worse crowbar.

<details>

</details>

# Changelog

:cl: Nomoresolvalou
balance: unpowered doors are now opened instantly in maintenance jack crowbar mode instead of in three seconds.
balance: crowbar mode can now pry doors closed when unpowered.
/:cl:

### DIFF
--- a/code/game/objects/items/tools/maintenance_tools.dm
+++ b/code/game/objects/items/tools/maintenance_tools.dm
@@ -568,8 +568,8 @@
 		if(attacked_door.density && !attacked_door.arePowerSystemsOn()) // if its closed and unpowered
 			attacked_door.open(TRUE)
 			return
-		if(!attacked_door.density) //If its open
-		return
+		if(!attacked_door.density && attacked_door.arePowerSystemsOn()) //If its open and powered
+			return
 
 		user.visible_message(SPAN_DANGER("[user] jams [src] into [attacked_door] and starts to pry it open."),
 		SPAN_DANGER("You jam [src] into [attacked_door] and start to pry it open."))
--- a/code/game/objects/items/tools/maintenance_tools.dm
+++ b/code/game/objects/items/tools/maintenance_tools.dm
@@ -568,7 +568,7 @@
 		if(attacked_door.density && !attacked_door.arePowerSystemsOn()) // if its closed and unpowered
 			attacked_door.open(TRUE)
 			return
-		if(!attacked_door.density && attacked_door.arePowerSystemsOn()) //If its open and powered
+		if(!attacked_door.density) //If its open
 			return
 
 		user.visible_message(SPAN_DANGER("[user] jams [src] into [attacked_door] and starts to pry it open."),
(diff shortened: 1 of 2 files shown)


code/game/objects/items/tools/maintenance_tools.dm
```diff
@@ -557,13 +557,17 @@
 		if(requires_superstrength_pry)
 			if(!HAS_TRAIT(user, TRAIT_SUPER_STRONG)) //basically IS_PRY_CAPABLE_CROWBAR
 				return
-		if(!attacked_door.density) //If its open
-			return
```
Drulikar, line 561: This should probably just be moved after your other checks (line 571) because it now allows you to force open a powered door that is already open (doing nothing but wasting your time); whereas before with this check in place it would make nothing happen.

Drulikar (changes requested):

code/game/objects/items/tools/maintenance_tools.dm
```diff
@@ -557,13 +557,19 @@
...
 			return
 		if(user.action_busy)
 			return
+		if(!attacked_door.density && !attacked_door.arePowerSystemsOn()) //If its open and unpowered
+			attacked_door.close(TRUE)
+			return
+		if(attacked_door.density && !attacked_door.arePowerSystemsOn()) // if its closed and unpowered
+			attacked_door.open(TRUE)
+			return
+		if(!attacked_door.density && attacked_door.arePowerSystemsOn()) //If its open and powered
```
Drulikar, line 571:
```suggestion
		if(!attacked_door.density) //If its open and powered
```
We already know the power systems are on because otherwise it would have entered the earlier check.

Drulikar (changes requested):

code/game/objects/items/tools/maintenance_tools.dm
Nomoresolvalou: I tried doing it without power check but it clocked me an error for some reason

Nomoresolvalou: ok I tried it again and it worked for some reason

Drulikar: The comment was still correct, but its fine. As I said we know it is powered.

Drulikar (approved):

PR #6306, opened 2024-05-18
Cats now have the correct name and description when held and in your bag

Nomoresolvalou:
# About the pull request

changes the get_scooped proc so that it now uses the correct name and description when creating a new mob_holder object. This means that mobs will have the correct name and description when held or in a container.

saying, "Here's the colony cat, Orion!" and then using the show held item verb, only to show "cat" with the default cat description makes no sense and breaks my immersion.

<details>

</details>

# Changelog

:cl: Nomoresolvalou

fix: fixed cats not specifically defined in code displaying wrong info when held or stored in a container. This should also fix any other mob capable of being picked up that is not defined in code as well.

/:cl:

### DIFF
--- a/code/modules/mob/holder.dm
+++ b/code/modules/mob/holder.dm
@@ -34,6 +34,15 @@
 	for(var/mob/living/M in contents)
 		M.show_message(message,m_type)
 
+/obj/item/holder/get_examine_text(mob/user)
+	. = list()
+	. += "[icon2html(src, user)] That's \a [src]."
+	if(desc)
+		. += desc
+	if(desc_lore)
+		. += SPAN_NOTICE("This has an <a href='byond://?src=\ref[src];desc_lore=1'>extended lore description</a>.")
+
+
 //Mob procs and vars for scooping up
 /mob/living/var/holder_type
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/holder.dm
```diff
@@ -49,7 +49,8 @@
 
 	var/obj/item/holder/mob_holder = new holder_type(loc)
 	src.forceMove(mob_holder)
```
Drulikar, line 51:
```suggestion
	forceMove(mob_holder)
```

code/modules/mob/holder.dm
```diff
@@ -49,7 +49,8 @@
...
-	mob_holder.name = loc.name
+	mob_holder.name = src.name
+	mob_holder.desc = src.desc
```
Drulikar, line 53:
```suggestion
	mob_holder.name = name
	mob_holder.desc = desc
```
You might also want to copy over the gender: https://www.byond.com/docs/ref/#/atom/var/gender
e.g. `mob_holder.gender = gender`

Drulikar (changes requested):
You might also want to implement `get_examine_text` for a `/obj/item/holder` If its holding an item, you can just `return ..()` to use the item implementation. But if a mob, you ought to replicate I think `/atom/proc/get_examine_text(mob/user)`

Currently the mob holder uses the generic item implementation resulting in the difference here: 
image

Drulikar (approved):
Btw you can undraft when ready here: 
image

PR #6310, opened 2024-05-19
Fixes trapper fire damage

Git-Nivrak:
# About the pull request
Globber PR featured an increased fire damage multiplier but forgot to account for the trapper strain

Oversights \ bugs are bad, fix good
<details>

</details>

# Changelog
:cl:
fix: Trapper no longer takes triple fire damage
/:cl:

### DIFF
--- a/code/__DEFINES/xeno.dm
+++ b/code/__DEFINES/xeno.dm
@@ -522,6 +522,9 @@
 // Hive Pylon protection range
 #define XENO_HIVE_PYLON_PROTECTION_RANGE 8
 
+// Fire damage multiplier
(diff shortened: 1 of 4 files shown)


Zonespace27 (changes requested):
This doesn't work as intended. This extends the lifespan of an on-fire trapper about 2x that of a regular boiler. Trapper and regular strain should both last the same amount of time on fire.

Drulikar:
I'm not sure what Zonespace saw needing changes; but I don't think Trapper strain is intended to have a perk of fire resistance (at least compared to the caste the strain modified). Trapper probably needs something to make it more interesting, but I don't think this is it, nor do I think it needs this at least right now.

PR #6311, opened 2024-05-19
AMR Anti-Focusing nerfs

Git-Nivrak:
# About the pull request
Makes the AMR counter reset to 0 after two fully charged shots

Being rewarded for focusing down a single xeno is toxic and bad for the game. Instead, after hitting it with 2 heavy shots you will have a choice between continuing to focus that xeno or focusing a different xeno with no loss of efficiency.
According to the guy who added AMR focusing only one target is terribly inefficient so this is infact a buff.
<details>

</details>

# Changelog
:cl:
balance: AMR will now reset its focus after two fully charged aimed shots on the same target.
/:cl:

### DIFF
--- a/code/datums/ammo/bullet/sniper.dm
+++ b/code/datums/ammo/bullet/sniper.dm
@@ -135,7 +135,7 @@
 				if(amr.focused_fire_counter < 2) // Can stack up to twice.
 					amr.focused_fire_counter += 1
 				else
-					amr.focused_fire_counter = 0
+					amr.focused_fire_counter = -1
 			else // If it's a new target
 				amr.focused_fire_counter = 0 // Stacks to 0
 				if(human_firer && !(target_mob.is_dead()))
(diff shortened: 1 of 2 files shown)


code/datums/ammo/bullet/sniper.dm
```diff
@@ -135,7 +135,7 @@
 				if(amr.focused_fire_counter < 2) // Can stack up to twice.
 					amr.focused_fire_counter += 1
 				else
-					amr.focused_fire_counter = 2
+					amr.focused_fire_counter = -1
```
Drulikar, line 138: This should be 0 else you won't get the one hit message; or you need to change the other handling of this counter. Just changing it to 0 though doesn't fix it incorrectly having the blue laser for the first hit. So do check that implementation too & test.

Drulikar (dismissed):

code/datums/ammo/bullet/sniper.dm
Git-Nivrak: yea I didn't realize the laser and counter were different things for some reason

harryob (approved):

PR #6325, opened 2024-05-23
modernises the readme and makes it considerably prettier

harryob:
infinite thanks to https://github.com/ParadiseSS13/Paradise/pull/25259 for a lot of the nice formatting here

the icons here (aside from the logo) are made by me, using font awesome (free) icons

preview it here! https://github.com/harryob/cmss13/tree/new-readme


realforest2001 (approved):

Drulikar (approved):
I sort of wanted some better indication for the headers to be links since somereason github doesn't underline them. But doing `<ins></ins>` on the link then would look funky with the subheading line. Past that was considering https://fontawesome.com/icons/arrow-up-right-from-square?f=classic&s=solid at the end of them, but then it'd seem like you should use that for all links, so I dunno its fine as is.

For me the first time looking at the page I was going to suggest you should add Guide to Git, because my attention was drawn more to the tip for the Guide to Contributing instead of the header's link. But I probably should just edit to better highlight guide to git on that page since all contributors need to follow those steps. It in a little git section doesn't make it seem relevant or important to contributing.

PR #6327, opened 2024-05-23
Adds WY & UPP Sentries

Comxy:
# About the pull request
This adds four versions of defenses WY branded for the PMC engineer. All WY defenses take longer to deploy. One of three can get picked by a PMC engineer:
Normal Sentry: shoots omni directional and slow.
Flame Sentry: shoots omni directional but with sticky napalm which is a worse fuel than blueflame.
WY Flag: same buffs as warbanner with more range but you cannot walk over it.
The other sentry a PMC engineer starts with by default.
Mini Sentry: like USCM but shoots faster with less damage.

Also adds an old type WY turret for future maps which has less dps but more health than other gauss turrets. I also made UPP versions of them.

I have tested the PR and everything works and is tested. I think everything is very well balanced, tested by me.

I am tired of seeing USCM sentries on a WY facility. It doesn't make sense to me why they would not have their own branded sentries. This fixes that issue.
<details>

Promo

</details>

# Changelog
:cl:
add: Adds W-Y branded sentries.
imageadd: Adds W-Y defenses sprites.
/:cl:


Zenith00000:
This is cool. You think you could add a "sentry_holder" version as well for WY specific colonies? It would be a cool addition to a map I'm working on, although the current sentry_holder sentries are already pretty OP, I wouldn't recommend changing their values to make them even more overpowered. Either way good stuff.

Comxy:
Sure, sounds simple enough

code/modules/defenses/sentry_flamer.dm
```diff
@@ -99,3 +99,37 @@
...
+	)
+
+/obj/structure/machinery/defenses/sentry/flamer/wy/destroyed_action()
+	visible_message("[icon2html(src, viewers(src))] [SPAN_WARNING("The [name] starts spitting out sparks and smoke!")]")
+	playsound(loc, 'sound/mecha/critdestrsyndi.ogg', 25, 1)
+	for(var/i = 1 to 6)
+		setDir(pick(NORTH, EAST, SOUTH, WEST))
+		sleep(2)
+
+	if(ammo.current_rounds != 0)
```
vero5123, line 130:
```suggestion
	if(ammo.current_rounds)
```

sleepynecrons (changes requested):
the pure white sentries are too much. tone down the white into something more subtle. alternatively you could give them an alternate (and more accurate) design entirely https://avp.fandom.com/wiki/Remote_Sentry_Turret

the logo on the flags should be W-Y yellow not orange

code/modules/defenses/sentry_flamer.dm
Comxy: It cannot read if you var has a value for some reason when I do this

vero5123: Then just revert it back as it was then, not a big deal.

Comxy:
Toned down the white and also made UPP versions. I thought that sentry from the wiki looked more like a UPP sentry.

cm13-github:

cm13-github:

code/modules/defenses/defenses.dm
```diff
@@ -144,6 +144,8 @@
 			faction_group = FACTION_LIST_HUMANOID
 		if(SENTRY_FACTION_COLONY)
 			faction_group = list(FACTION_MARINE, FACTION_COLONIST)
+		if(FACTION_UPP)
```
realforest2001, line 147: I would suggest also adding an option in here, however you see fit, to set IFF exclusively to WY & PMCs, without including Marines.

Aside from that, all code looks good to me. Sprites are great, though I can't review those.

nauticall (approved):
looks good now. sprites approved

PR #6328, opened 2024-05-24
Containers can be dumped into seed extractor

doganesi:
# About the pull request

Allows you to dump storage containers into the seed extractor.

Some nice QOL so you don't have to individually place in 40 different plants one by one into the seed extractor.

<details>

https://streamable.com/6ter1w

</details>

# Changelog
:cl: Asmocard
qol: Can now dump containers into the seed extractor
/:cl:


code/game/machinery/seed_extractor.dm
```diff
@@ -8,6 +8,43 @@
...
+							var/obj/item/grown/F = item
+							new_seed_type = GLOB.seed_types[F.plantname]
+						else
+							var/obj/item/reagent_container/food/snacks/grown/F = item
+							new_seed_type = GLOB.seed_types[F.plantname]
+
+						if(new_seed_type)
+							to_chat(user, SPAN_NOTICE("You extract some seeds from [item]."))
+							var/produce = rand(1,4)
+							for(var/i = 0;i<=produce;i++)
```
kiVts, line 35: for(var/i in 0 to produce)

kiVts: faster

code/game/machinery/seed_extractor.dm
```diff
@@ -8,6 +8,43 @@
...
 /obj/structure/machinery/seed_extractor/attackby(obj/item/O as obj, mob/user as mob)
 
+	// Plant bag and other storage containers.
+	if(istype(O,/obj/item/storage))
+		var/obj/item/storage/container = O
+		if(container.contents.len > 0)
+			to_chat(user, SPAN_NOTICE("You start dumping the contents of [container] into [src]."))
+			if(!do_after(user, 1.5 SECONDS, INTERRUPT_ALL, BUSY_ICON_GENERIC))
+				return
+			for(var/obj/item/item in container)
```
kiVts, line 18: since you're typechecking below, you can add "as anything" to scik the typecheck for loop does.

so essentially this 
for(var/obj/item/item as anything in container)

I could be wrong but it seems very logical no?

code/game/machinery/seed_extractor.dm
```diff
@@ -8,6 +8,43 @@
...
+				if(!istype(item, /obj/item/reagent_container/food/snacks/grown) && !istype(item, /obj/item/grown))
+					to_chat(user, SPAN_WARNING("Cannot get seeds from [item]."))
+					break
+				else
+					if(user.temp_drop_inv_item(item))
+						var/datum/seed/new_seed_type
+						if(istype(item, /obj/item/grown))
+							var/obj/item/grown/F = item
```
kiVts, line 26: maints will get mad at one letter var

doganesi: kill the person who wrote the original seed extractor code, but I'll fix it

code/game/machinery/seed_extractor.dm
```diff
@@ -8,16 +8,53 @@
 
 /obj/structure/machinery/seed_extractor/attackby(obj/item/O as obj, mob/user as mob)
 
+	// Plant bag and other storage containers.
+	if(istype(O,/obj/item/storage))
+		var/obj/item/storage/container = O
+		if(container.contents.len > 0)
```
kiVts, line 14: length(container.contents) over .len since its nullsafe (and tiny bit faster too, at least I was told so.)

code/game/machinery/seed_extractor.dm
```diff
@@ -8,16 +8,53 @@
 
 /obj/structure/machinery/seed_extractor/attackby(obj/item/O as obj, mob/user as mob)
 
+	// Plant bag and other storage containers.
+	if(istype(O,/obj/item/storage))
+		var/obj/item/storage/container = O
+		if(length(container.contents) > 0)
```
kiVts, line 14: vero got a point, this code nests with if statements that could be early returned pretty deep, you can early return this. Essentially make the if statement opposite and then return in it with a message that you failed for whatever reason. 

so essenially this 

```dm
if(!correct)
     to_world("you failed")
     return
code here
```
instead of
```dm
if(correct)
   code here
else
    to_world("you failed")
```
or in this case swap the > with = or wrap them in brackets() and put ! infront of them to flip the result. 

`if(!(result == 0))`

code/game/machinery/seed_extractor.dm
```diff
@@ -8,16 +8,53 @@
 
 /obj/structure/machinery/seed_extractor/attackby(obj/item/O as obj, mob/user as mob)
 
+	// Plant bag and other storage containers.
+	if(istype(O,/obj/item/storage))
```
kiVts, line 12: same here, just put ! infront of istype to flip the result and return it

code/game/machinery/seed_extractor.dm
```diff
@@ -8,16 +8,53 @@
...
+					if(user.temp_drop_inv_item(item))
+						var/datum/seed/new_seed_type
+						if(istype(item, /obj/item/grown))
+							var/obj/item/grown/plant = item
+							new_seed_type = GLOB.seed_types[plant.plantname]
+						else
+							var/obj/item/reagent_container/food/snacks/grown/plant = item
+							new_seed_type = GLOB.seed_types[plant.plantname]
+
+						if(new_seed_type)
```
kiVts, line 32: same here, just put ! infront like the usual and sort that stuff there. also remove the else clause from this if and put that there.

doganesi:
Did a pretty stupid workaround with the /storage proc I added to fix item boxes in the container after processing, let me know if there's a better way of doing that please.

code/game/machinery/seed_extractor.dm
```diff
@@ -8,16 +8,38 @@
 
 /obj/structure/machinery/seed_extractor/attackby(obj/item/O as obj, mob/user as mob)
 
+		// Plant bag and other storage containers.
+	if(istype(O,/obj/item/storage))
```
kiVts, line 12: you can probably typecheck for plant bag here instead of storage in general, I doubt someone would use backpack for this

doganesi: I mean eh I could but why not leave it open for storage, chance like a mess tech dumps a box of stuff they get from req into their extractor

PR #6145, opened 2024-04-14
Synthetics X Gibber Fanfiction (Reworks how Synthetics and the meat gibber interact)

Ben10083:
# About the pull request

- Synthetic is delimbed and spat out of gibber instead of a true gib, making them not die in the process
- Synthetics now spawn their proper meat as desired

# Explain why it's good for the game

image
Desired by lore contributor, makes it so synthetic snowflake situation where they are shoved in gibber is handled properly.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

https://github.com/cmss13-devs/cmss13/assets/91219575/96d92eaf-a60e-45f3-ac80-6d0275e82428

</details>

# Changelog
:cl:
add: Synthetics when shoved into meat gibber will not be delimbed and spawn their own flesh, but would not die (they are spat out as a torso and head)
admin: Adjustment to meatgibber code to notify admins when a synthetic is being shoved into gibber
/:cl:

PR #6332, opened 2024-05-25
Fixes gibber not killing mobs

### LINKED ISSUE #6331: Gibber don't work
### Testmerges

N/A

### Round ID

N/A

### Description of the bug

Gibber machine broke. When trying to gib human corpses. Probably for the best since gibbing people is against the rules.

Gibber should maybe work.

1. Try to gib a corpse, it never finishes gibbing.
2. Trying to empty the gibber does not work. The body is now stuck.
3. You can put in a second corpse, which can be removed, but also not gibbed.
...

### Issue Bingo

vero5123:
# About the pull request
Fixes #6331, gibber was essentially only partially gibbing synths leaving all other mobs unaffected.

The gibber should be able to gib mobs that aren't just synths.

# Changelog
:cl:
fix: Fixes non-synthetic mobs getting stuck in the gibber
/:cl:

### DIFF
--- a/code/game/machinery/kitchen/gibber.dm
+++ b/code/game/machinery/kitchen/gibber.dm
@@ -262,13 +262,12 @@
 		occupant.death(create_cause_data("gibber", user), TRUE)
 		occupant.ghostize()
 
+	addtimer(CALLBACK(src, PROC_REF(create_gibs), totalslabs, allmeat), gibtime)
 	if(synthetic)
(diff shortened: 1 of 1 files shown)


Drulikar (approved):
This looks correct re: https://github.com/cmss13-devs/cmss13/pull/6145/files#diff-985f572033f1b0b8c2e02b09a2d80cb94763734073b668b4a4f0630c2c4c69c4L207

PR #6343, opened 2024-05-27
Makes it possible for all Officers and all NCO's (E4+) to recommend people for medals.

TheGamerdk:
# About the pull request

I've been meaning to do this.

Corporal and above + all officers can recommend medals.

I tested it.

Tried to give a medal recommendation as a PFC: Failed
Tried to give a medal recommendation as a FTL: Success
Tried to give a medal recommendation as an SO: Success
Tried to give a medal recommendation as XO: Success

<details>

</details>

# Changelog
:cl:
add: All officers and NCO's can now recommend people for medals.
/:cl:

### DIFF
--- a/code/datums/medal_awards.dm
+++ b/code/datums/medal_awards.dm
@@ -479,6 +479,10 @@ GLOBAL_LIST_INIT(xeno_medals, list(XENO_SLAUGHTER_MEDAL, XENO_RESILIENCE_MEDAL, 
 			continue
 		recipient_ranks[recipient_name] = record.fields["rank"]
 		possible_recipients += recipient_name
+	if(length(possible_recipients) == 0)
+		to_chat(recommendation_giver, SPAN_WARNING("It's not possible to give medals when the ship is empty. Tough luck, partner..."))
+		return FALSE
+
 	var/chosen_recipient = tgui_input_list(recommendation_giver, "Who do you want to recommend a medal for?", "Medal Recommendation", possible_recipients)
 	if(!chosen_recipient)
 		return FALSE
--- a/code/game/objects/items/devices/radio/headset.dm
+++ b/code/game/objects/items/devices/radio/headset.dm
@@ -417,7 +417,7 @@
 	var/obj/item/card/id/id_card = wearer.wear_id?.GetID()
 	if(!istype(id_card))
 		return
-	var/paygrade = wearer.get_paygrade()
+	var/paygrade = get_paygrade_id_by_name(wearer.get_paygrade(FALSE))
 	var/last_number = text2num(copytext(paygrade, length(paygrade), length(paygrade)+1))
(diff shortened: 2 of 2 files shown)


code/game/objects/items/devices/radio/headset.dm
```diff
@@ -417,8 +417,10 @@
 	var/obj/item/card/id/id_card = wearer.wear_id?.GetID()
 	if(!istype(id_card))
 		return
-	if(!(id_card.rank in list(JOB_SO, JOB_XO, JOB_SQUAD_LEADER)))
-		to_chat(wearer, SPAN_WARNING("Only Staff Officers, Executive Officers and Squad Leaders are permitted to give medal recommendations!"))
+	var/paygrade = get_paygrade_id_by_name(wearer.get_paygrade(FALSE))
+	var/last_number = text2num(copytext(paygrade, length(paygrade), length(paygrade)+1))
+	if(!(findtext(paygrade, "O") || last_number >= 4))
+		to_chat(wearer, SPAN_WARNING("Only officers or NCO's (ME4+) can recommend medals!"))
```
Drulikar, line 423: String manipulation is not the way to go about this. Either just explicitly list out the ranks that are allowed, or consider a more appropriate value e.g. `officer_grade` on the `/datum/paygrade`

Drulikar (changes requested):

code/game/objects/items/devices/radio/headset.dm
TheGamerdk: You’ll be having a really long list, and officer_grade doesn’t account for NCOs.
There’s nothing wrong with string manipulation.

TheGamerdk:
Not going to spend time coming up with an inferior solution

ProfanityTM:
Does this apply to ASO? They couldn't do medal recomendations last I recall.

TheGamerdk:
> Does this apply to ASO? They couldn't do medal recomendations last I recall.

They’re an officer, so yes :)

code/datums/medal_awards.dm
```diff
@@ -479,6 +479,10 @@ GLOBAL_LIST_INIT(xeno_medals, list(XENO_SLAUGHTER_MEDAL, XENO_RESILIENCE_MEDAL,
 			continue
 		recipient_ranks[recipient_name] = record.fields["rank"]
 		possible_recipients += recipient_name
+	if(length(possible_recipients) == 0)
```
vero5123, line 482: if u wanted u could just do if(!length(possible_recipients))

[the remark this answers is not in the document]
Drulikar: Either are fine.

Drulikar (approved):
Of note, this allows a CO to make medal recommendations, but I guess this would allow you to prepare a medal?

PR #6355, opened 2024-06-01
ID Mod computer biometrics

realforest2001:
# About the pull request

This PR adds a biometric check for LOGGING IN to an id mod computer. This will not prevent people from pressing buttons whilst another person is already logged in, but will prevent use of an invalid or stolen ID.

Primary change this will make is stopping ERTs from bypassing the deliberate no-access they have to cancel Evac and similar actions in CIC. This is not intended to prevent people sneaking access onto their cards whilst someone else legitimately logged in to the console.

<details>

</details>

# Changelog

:cl:
add: Added a biometric check to the initial authentication part of ID Modification consoles.
/:cl:

### DIFF
--- a/code/modules/cm_marines/marines_consoles.dm
+++ b/code/modules/cm_marines/marines_consoles.dm
@@ -28,6 +28,15 @@
 		visible_message("[SPAN_BOLD("[src]")] states, \"AUTH ERROR: Authority confirmation card is missing!\"")
 		return FALSE
 
+	var/bio_fail = FALSE
+	if(!istype(id_card))
+		bio_fail = TRUE
+	else if(!id_card.check_biometrics(user))
+		bio_fail = TRUE
+	if(bio_fail)
+		to_chat(user, SPAN_WARNING("Biometrics failure! You require an authenticated ID card to perform this action!"))
+		return FALSE
+
 	if(check_access(id_card))
 		authenticated = TRUE
 		visible_message("[SPAN_BOLD("[src]")] states, \"AUTH LOGIN: Welcome, [id_card.registered_name]. Access granted.\"")


code/modules/cm_marines/marines_consoles.dm
```diff
@@ -28,6 +28,15 @@
...
 		return FALSE
 
+	var/bio_fail = FALSE
+	if(!istype(id_card))
+		bio_fail = TRUE
+	else if(!id_card.check_biometrics(user))
+		bio_fail = TRUE
+	if(bio_fail)
+		to_chat(user, SPAN_WARNING("Biometrics failure! You require an authenticated ID card to perform this action!"))
+		return FALSE
```
vero5123, line 38:
```suggestion
	if(!istype(id_card) || !id_card.check_biometrics(user))
		to_chat(user, SPAN_WARNING("Biometrics failure! You require an authenticated ID card to perform this action!"))
		return FALSE
```

realforest2001: Doh! Duh!

vero5123:
Gotta fix the indentation, GitHub editor :/

JackieEstegado:
This will also stop people from getting access when only few people are left alive and want to initiate evac for pods.
Just do a faction check if you want to prevent ERTs from doing it.

Skatuk:
So your doing this to stop evac being cancelled???? When you can give it a cooldown or the requirement of USCM faction or hell even make it so they can't toggle it mechanically? I know your the vice host and all that but this is a generally a bad PR . The deliberate no access makes it so they have to scavange command ID's since it makes no sense IC how CLF have direct access to CIC (Exception being the technician) also scavenging ID's permits them the ability to gain IFF so they can actually attack defences with the limited numbers they have.

Diegoflores31:
Just for future clarification . Killing someone WITH access and using their ID to cancel evac would be then Valid game ?

Drulikar:
Stale exempt until reviewed

PR #6360, opened 2024-06-02
Set Plasteel to 40. Gave APC Board. Balance & QOL

hislittlecuzingames:
# About the pull request

Makes Com Tech a little nicer. Sets up starting equipment to be more friendly to CT. 

Just little stuff that makes CT more annoying to play. Most players don't want to sit at FOB, but if they do, this may make it a little less annoying. 

Benefits:
1. Plasteel is now equal to 5 full plasteel doors - instead of 3 doors, and 6 plasteel you'll never use. 
2. Gave APC board. Now incase you mess up as com tech, you can fix one APC like Comms or power without making an objectively incorrect decision by not taking the electronics pouch. This is so a round cannot be ruined by a com tech just not having what he needs to setup comms. Com techs already got the battery, so I'm adding the board & they get 50 metal for the frames. 

Drawbacks:
Each Com tech can now make 5 doors. I pondered only giving Com Techs 32 Plasteel so they could top up at 4 doors without having extra plasteel. 
It's possible Com Techs are less likely to take the electronics pouch to fix both Comms and power.

My purpose with this is to make it a little less annoying to play CT. With plasteel being an awkward 7 points and nearly no way to get plasteel otherwise, the plasteel thing is just annoying. 
I also added the circuit board because it's extremely easy to misclick and not be able to take the electronics pouch. 

I also ask... should the toolbelt be allowed 1 more slot so the circuit board can fit in there? Only like screwdrivers fit. It's possible a welder or two could be stuffed in there, but I think most people that hardcore metagame don't even take the light replacer, so they can already fit another welder, so it likely won't matter. 

Video attached

<details>

https://github.com/cmss13-devs/cmss13/assets/142365554/d30da0d0-6c9e-403f-ae51-cd85b24d233e

</details>

# Changelog

:cl:
balance: Set CT free plasteel to 40 and added APC Circuit board
/:cl:


hislittlecuzingames:
After talking with Drathek Briefly in Discord, I have decided to not consider increasing the toolbelt capacity size.

code/game/machinery/vending/vendor_types/squad_prep/squad_engineer.dm
```diff
@@ -205,7 +205,8 @@ GLOBAL_LIST_INIT(cm_vending_clothing_engi, list(
 		/obj/item/explosive/plastic,
 		/obj/item/stack/sandbags_empty = 25,
 		/obj/item/stack/sheet/metal/large_stack,
-		/obj/item/stack/sheet/plasteel/medium_stack,
+		/obj/item/stack/sheet/plasteel/medium_stack = 40,
```
Drulikar, line 208: Wouldn't this instead just be `/obj/item/stack/sheet/plasteel/med_large_stack`

hislittlecuzingames: That would appear to also be the same value.
I just did a built with that type. 

I did the number because I based my change on the sandbag line on line 206. 

Pushing that edit now to provide code consistency.

Drulikar (approved):

PR #6296, opened 2024-05-16
LZ fog before first drop and sentryguns to guard agains early FOB rush (they last only ten minits)

cuberound:
# About the pull request

adds a fog that is removed on DS landing and omnisentryguns ment for LZ that last for 10 minits after first touchdown, THIS IS NOT MAPPING PR, at least for now, it is not used in any map, reqires mappers to add it. Mby getting some better smoke / gas spirte if someone is able to do that would be apriciated. aka alamayer drops gasbombs and omnisentryguns on LZ to secure the landing ( welp at least some IC explonation for fog on LZ aint it?) also fog removal on flyby landing, if it happens before first drop, is intentional. sentrgun numbers NEED adjusting, I threw in some random stuff, the sentrygun is work in progress mostly

# Explain why it's good for the game
this solves two issues, survivors or xenos camping LZs for first landing. And early FOB rushes. first one is kinda meh second one is realy needed, round ending at 40 minits just is not fun for anyone, this does not make early fob pushes impossible but a lot harder, relevant voices said that mechanical solution is better then rule, sentrygun punishes you in predictable way and you do not have to argue with it. And admins hate handeling "t3s are standing on LZ first drop" tickets so it is win win 

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
add: adds option for fog that drops on first landing
add:  adds option landing zone sentryguns that last for 10 minits after the first drop
/:cl:

PR #6217, opened 2024-04-30
(WIP) CORSAT: The Shrinkening

realforest2001:
# About the pull request
Returns CORSAT to a _hopefully_ playable state.

- Removes Ice Dome
- Removes Monorail
- Removes much of the NW region of the map as it's unused.
- Moves around a few other areas of the map to condense it.
https://forum.cm-ss13.com/t/temp-corsat-map/7955

# Explain why it's good for the game

# Testing Photographs and Procedure
image

# Changelog
:cl: forest2001, Spartanbobby
mapadd: Returns CORSAT, now significantly smaller.
imageadd: Added mapping-sprites for CORSAT hull windows.
/:cl:

PR #6071, opened 2024-04-03
[V] Sorokyne Strata: Expanded Small corridors, removed 3000 unused tiles, Xenos can now weed everything except LZ's and surrounding of LZ's, allow LAZ for CAS inside structures with light blue Areas on Tacmap (glass ceiling hallways)

Venuska1117:
# About the pull request

This PR clean Sorokyne Strata file from excessive lines of code caused by APC's and small light fixture.
It changes some small rooms with have 1x1 hallways, now if possible, they are 2x2 (minimum).
Replaced 80% of indestructable walls with heavy reinforced walls, i dont see need for indestructable walls if they are not critical (for example, map edge)
Shortened river by half (south west near LZ) and placed invisible wall, i tough of preventing (in future) for burrowers to getting there and re-evolve to queen in spot where marines have no access because of invisible wall.
It "removes" walls with are 2tiles or more thick, i don't know why there are 2 tiles thick walls, but i believe they are not important.
Removes around 3000 unused tiles.

Edit: 08.08
- Allow xenos to weed everywhere (except LZ's)
- Allow marines to use LAZ for CAS inside most structures that have light blue colored areas (hallways made of glass ceiling)

Fixes: #6869

# Explain why it's good for the game

Less lines of code allow for quicker groundsite loading, even if its second faster, and also allow for more map destruction with less chockes caused by 1x1 tunnels, they don't appear much but they are "annoying", allow marines to use LAZ for CAS inside most structures (not all of them) and xenos can now weed 90% of areas except LZ's

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

img1

img2

img3

img4

img5

img6

img7

img8

img9

</details>

# Changelog

:cl: Venuska1117
maptweak: Sorokyne Strata - made most indestructable walls destructable.
maptweak: Sorokyne Strata - changed rooms/hallways with have 1x1 hallways to be wider.
maptweak: Sorokyne Strata - replaced some small light fixtures with normal light fixtures.
maptweak: Sorokyne Strata - reduced 2 tiles or more thick walls to 1 tile thick walls.
balance: Made most areas weedable, except LZ's zones and surrounding LZ's sub-variants.
balance: Allow LAZ inside Outposts (mining and normal ones) as long as they aim at light blue areas (glass ceiling hallways).
code: Changed pixel offset for "tube" lights to be closer to "floor".
/:cl:

PR #6363, opened 2024-06-02
First Drop Protections: Gas and Static Turrets

Drulikar:
# About the pull request

This PR is a more fleshed out implementation of #6296 that can likely be expanded more into Zonespace's plans for generators powering turrets and more (additional help for low pop).

Smoke takes 6 seconds from its spawn before it will deal burn damage and smoke in vehicles takes up to 7 seconds to dissipate.

The sequence of events are as follows:
1. After ~~two~~ three minutes from round start, a "fake" OB will be launched to each landing zone that then causes a miasma (loosely considered CN20-X miasma since it affects xenos too) in all landing zone areas.
2. When the first dropship departs, the smoke will clear
3. When the first dropship lands, four ~~10~~ 20 minute gauss sentries are automatically deployed randomly in the NW, NE, SW, and SE regions of the LZ immediately around the dropship:
dropship
4. As the sentry runs out of power it warns the battery is running low (battery still depletes if undeployed & cannot be moved or manipulated any other than deployment/undeployment)
5. When unpowered, it can only be undeployed so it is no longer blocking movement.

Since I am primarily using the `is_landing_zone` area var to determine where to make inhospitable, all maps need to be verified:
- [X] LV-624
- [X] Shivas Snowball
- [X] Solaris Ridge
- [X] Fiorina Science Annex
- [X] Trijent Dam
- [ ] Sorokyne Strata (Would need mapping fixes to expand LZs, but would conflict with https://github.com/cmss13-devs/cmss13/pull/6071)
- [X] Kutjevo Refinery
- [X] Chances Claim
- [X] New Varadero
- [X] Corsat* (Current Corsat is fine, https://github.com/cmss13-devs/cmss13/pull/6217 would need some changes)
- [X] Fiorina Cellblocks*
- [X] LV-759 Hybrisa Prospera*

Ultimately this came about because of problems where marines are not able to get a foothold quickly enough first drop before being overwhelmed. This especially affects low pop.

The goals of this change are to make survivors holding LZs basically impossible, to deter xenos from wanting to go near LZs, and to give marines at least ~~10~~20 minutes after first drop time to create a FOB.

<details>

https://youtu.be/o8kE13Oua6o

image

vehicle smoke

</details>

# Changelog
:cl: Drathek cuberound
balance: After three minutes from round start, landing zones will be covered in a deadly gas
balance: First dropship deployment will clear gas, and first dropship arrival will deploy 4 sentries that last 20 minutes
refactor: Refactored smoke code
maptweak: Expanded LZ areas on Trijent Dam and Kutjevo Refinery
maptweak: Shrunk LZ1 on Shivas Snowball and LZ2 on LV624 slightly
maptweak: Removed fog on Chances Claim since it will be replaced by smoke and tweaked southern tunnel area to not be considered a LZ
fix: Fixed weird LZ lighting on some maps
spellcheck: Replaced/removed synth coughing emotes in smoke
/:cl:

### DIFF
--- a/code/game/objects/effects/effect_system/smoke.dm
+++ b/code/game/objects/effects/effect_system/smoke.dm
@@ -160,6 +160,8 @@
 	opacity = FALSE
 	alpha = 75
 	color = "#301934"
+	var/burn_damage = 4
+	var/xeno_yautja_multiplier = 3
 
 /obj/effect/particle_effect/smoke/miasma/Initialize(mapload, oldamount, new_cause_data)
 	. = ..()
@@ -176,8 +178,17 @@
 /obj/effect/particle_effect/smoke/miasma/affect(mob/living/carbon/affected_mob)
 	..()
 
-	affected_mob.apply_damage(10, TOX)
-	affected_mob.SetEyeBlind(1)
+	var/damage = burn_damage
+	if(isxeno(affected_mob))
+		damage *= xeno_yautja_multiplier
+	else if(isyautja(affected_mob))
+		if(prob(75))
+			return FALSE
+		damage *= xeno_yautja_multiplier
+
+	affected_mob.apply_damage(damage, BURN)
+	affected_mob.AdjustEyeBlur(0.75)
+
 	if(affected_mob.coughedtime < world.time && !affected_mob.stat)
 		affected_mob.coughedtime = world.time + 2 SECONDS
 		if(ishuman(affected_mob)) //Humans only to avoid issues
--- a/code/game/objects/effects/effect_system/smoke.dm
+++ b/code/game/objects/effects/effect_system/smoke.dm
@@ -177,9 +177,15 @@
 
 /obj/effect/particle_effect/smoke/miasma/process()
 	. = ..()
-	for(var/obj/structure/closet/container in get_turf(src))
+	var/turf/turf = get_turf(src)
+	for(var/obj/structure/closet/container in turf)
 		for(var/mob/living/carbon/mob in container)
 			affect(mob)
+	var/obj/vehicle/multitile/car = locate() in turf
+	var/datum/interior/car_interior = car?.interior
+	var/list/passengers = car_interior?.get_passengers()
+	for(var/mob/living/mob as anything in passengers)
+		affect(mob)
 
 /obj/effect/particle_effect/smoke/miasma/affect(mob/living/carbon/affected_mob)
 	. = ..()
--- a/code/game/objects/effects/effect_system/smoke.dm
+++ b/code/game/objects/effects/effect_system/smoke.dm
@@ -118,7 +118,7 @@
 
 
 /obj/effect/particle_effect/smoke/proc/affect(mob/living/carbon/affected_mob)
-	if(istype(affected_mob))
+	if(!istype(affected_mob))
 		return FALSE
 	return TRUE
 
@@ -136,17 +136,23 @@
 		affect(affected_mob)
 
 /obj/effect/particle_effect/smoke/bad/affect(mob/living/carbon/affected_mob)
-	..()
-	if (affected_mob.internal != null && affected_mob.wear_mask && (affected_mob.wear_mask.flags_inventory & ALLOWINTERNALS))
-		return
-	else
-		if(prob(20))
-			affected_mob.drop_held_item()
-		affected_mob.apply_damage(1, OXY)
-		if(affected_mob.coughedtime < world.time && !affected_mob.stat)
-			affected_mob.coughedtime = world.time + 2 SECONDS
-			if(ishuman(affected_mob)) //Humans only to avoid issues
-				affected_mob.emote("cough")
+	. = ..()
+	if(!.)
+		return FALSE
+	if(affected_mob.internal != null && affected_mob.wear_mask && (affected_mob.wear_mask.flags_inventory & ALLOWINTERNALS))
+		return FALSE
+	if(issynth(affected_mob))
+		return FALSE
+
+	if(prob(20))
+		affected_mob.drop_held_item()
+	affected_mob.apply_damage(1, OXY)
+
+	if(affected_mob.coughedtime < world.time && !affected_mob.stat)
+		affected_mob.coughedtime = world.time + 2 SECONDS
+		if(ishuman(affected_mob)) //Humans only to avoid issues
+			affected_mob.emote("cough")
+	return TRUE
 
 /////////////////////////////////////////////
 // Miasma smoke (for LZs)
@@ -176,7 +182,11 @@
 			affect(mob)
 
 /obj/effect/particle_effect/smoke/miasma/affect(mob/living/carbon/affected_mob)
-	..()
+	. = ..()
+	if(!.)
+		return FALSE
+	if(affected_mob.stat == DEAD)
+		return FALSE
 
 	var/damage = burn_damage
 	if(isxeno(affected_mob))
@@ -188,17 +198,23 @@
 
 	affected_mob.apply_damage(damage, BURN)
 	affected_mob.AdjustEyeBlur(0.75)
+	affected_mob.last_damage_data = cause_data
 
 	if(affected_mob.coughedtime < world.time && !affected_mob.stat)
 		affected_mob.coughedtime = world.time + 2 SECONDS
 		if(ishuman(affected_mob)) //Humans only to avoid issues
-			if(prob(50))
-				affected_mob.emote("cough")
+			if(issynth(affected_mob))
+				affected_mob.visible_message(SPAN_DANGER("[affected_mob]'s skin is sloughing off!"),\
+				SPAN_DANGER("Your skin is sloughing off!"))
 			else
-				affected_mob.emote("gasp")
+				if(prob(50))
+					affected_mob.emote("cough")
+				else
+					affected_mob.emote("gasp")
 			if(prob(20))
 				affected_mob.drop_held_item()
 		to_chat(affected_mob, SPAN_DANGER("Something is not right here..."))
+	return TRUE
 
 /////////////////////////////////////////////
 // Sleep smoke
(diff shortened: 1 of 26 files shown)


code/modules/defenses/sentry.dm
```diff
@@ -546,6 +546,45 @@
...
+/obj/structure/machinery/defenses/sentry/premade/deployable/colony/landing_zone/get_examine_text(mob/user)
+	. = ..()
+	switch(battery_state)
+		if(TURRET_BATTERY_STATE_LOW)
+			. += "Its battery indictor is flashing yellow."
+		if(TURRET_BATTERY_STATE_CRITICAL)
+			. += "Its battery indictor is flashing red."
+		if(TURRET_BATTERY_STATE_DEAD)
+			. += "It appears to be offline."
+
```
vero5123, line 573: maybe?
```suggestion
/obj/structure/machinery/defenses/sentry/premade/deployable/colony/landing_zone/get_examine_text(mob/user)
	. = ..()
	switch(battery_state)
		if(TURRET_BATTERY_STATE_OK)
			. += "Its battery indictor is green, fully charged"
		if(TURRET_BATTERY_STATE_LOW)
			. += "Its battery indictor is flashing yellow."
		if(TURRET_BATTERY_STATE_CRITICAL)
			. += "Its battery indictor is flashing red."
		if(TURRET_BATTERY_STATE_DEAD)
			. += "It appears to be offline."
```

code/modules/defenses/sentry.dm
```diff
@@ -546,6 +546,45 @@
...
+/obj/structure/machinery/defenses/sentry/premade/deployable/colony/landing_zone/get_examine_text(mob/user)
+	. = ..()
+	switch(battery_state)
+		if(TURRET_BATTERY_STATE_LOW)
+			. += "Its battery indictor is flashing yellow."
+		if(TURRET_BATTERY_STATE_CRITICAL)
+			. += "Its battery indictor is flashing red."
+		if(TURRET_BATTERY_STATE_DEAD)
+			. += "It appears to be offline."
```
vero5123, line 572: Why should it be omitted? 
```suggestion
/obj/structure/machinery/defenses/sentry/premade/deployable/colony/landing_zone/get_examine_text(mob/user)
	. = ..()
	switch(battery_state)
		if(TURRET_BATTERY_STATE_OK)
			. += "Its battery indictor is green, fully charged."
		if(TURRET_BATTERY_STATE_LOW)
			. += "Its battery indictor is flashing yellow."
		if(TURRET_BATTERY_STATE_CRITICAL)
			. += "Its battery indictor is flashing red."
		if(TURRET_BATTERY_STATE_DEAD)
			. += "It appears to be offline."
```

code/modules/defenses/sentry.dm
```diff
@@ -546,6 +546,45 @@
 	handheld_type = /obj/item/defenses/handheld/sentry/mini
 	composite_icon = FALSE
 
+/obj/structure/machinery/defenses/sentry/premade/deployable/colony/landing_zone
+	name = "\improper UA-577 Spaceborn Gauss Turret"
+	fire_delay = 2
+	omni_directional = TRUE
+	var/battery_low = 7 MINUTES
+	var/battery_critical = 9 MINUTES
+	var/battery_dead = 10 MINUTES
```
vero5123, line 555: these vars don't change could maybe make it a define instead

code/modules/defenses/sentry.dm
```diff
@@ -546,6 +546,45 @@
...
+			playsound(loc, 'sound/weapons/smg_empty_alarm.ogg', 15, 1)
+			visible_message(SPAN_WARNING("[name] beeps steadily as its battery is getting low."))
+		if(TURRET_BATTERY_STATE_CRITICAL)
+			playsound(loc, 'sound/weapons/smg_empty_alarm.ogg', 20, 1)
+			visible_message(SPAN_WARNING("[name] beeps steadily as its battery gets critically low."))
+		if(TURRET_BATTERY_STATE_DEAD)
+			playsound(loc, 'sound/machines/terminal_shutdown.ogg', 35, 1)
+			turned_on = FALSE
+			power_off_action()
+			update_icon()
```
vero5123, line 587:
```suggestion
		if(TURRET_BATTERY_STATE_DEAD)
			playsound(loc, 'sound/machines/terminal_shutdown.ogg', 35, 1)
			visible_message(SPAN_WARNING("[name] beeps faintly as it begins to power off"))
			turned_on = FALSE
			power_off_action()
			update_icon()
```

code/game/objects/effects/effect_system/smoke.dm
```diff
@@ -128,22 +132,73 @@
...
+	if(affected_mob.coughedtime < world.time && !affected_mob.stat)
+		affected_mob.coughedtime = world.time + 2 SECONDS
+		if(ishuman(affected_mob)) //Humans only to avoid issues
+			if(prob(50))
+				affected_mob.emote("cough")
+			else
+				affected_mob.emote("gasp")
+			if(prob(20))
+				affected_mob.drop_held_item()
+		to_chat(affected_mob, SPAN_DANGER("Something is not right here..."))
```
vero5123, line 201: is it lore accurate that synths can cough?

code/game/objects/effects/effect_system/smoke.dm
```diff
@@ -62,61 +62,65 @@
...
 				continue
-		var/obj/effect/particle_effect/smoke/S = new type(T, amount, cause_data)
-		S.setDir(pick(GLOB.cardinals))
-		S.time_to_live = time_to_live
-		if(S.amount>0)
-			S.spread_smoke()
+		var/obj/effect/particle_effect/smoke/smoke = new type(cur_turf, amount, cause_data)
+		smoke.setDir(pick(GLOB.cardinals))
+		smoke.time_to_live = time_to_live
+		if(smoke.amount>0)
```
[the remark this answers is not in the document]
Drulikar: So negative values are also truthy, and we explicitly only want positive numbers to call this recursive proc else you will have an infinite loop.

code/game/objects/effects/effect_system/smoke.dm
```diff
@@ -543,34 +592,36 @@
...
+			if(istype(cur_atom, /obj/flamer_fire))
+				qdel(cur_atom)
 
-		S.setDir(pick(GLOB.cardinals))
-		S.time_to_live = time_to_live
-		if(S.amount>0)
-			S.spread_smoke()
+		smoke.setDir(pick(GLOB.cardinals))
+		smoke.time_to_live = time_to_live
+		if(smoke.amount>0)
```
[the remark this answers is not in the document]
Drulikar: So negative values are also truthy, and we explicitly only want positive numbers to call this recursive proc else you will have an infinite loop.

code/modules/defenses/sentry.dm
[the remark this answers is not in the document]
Drulikar: It doesn't make sense to me to globally define that a battery goes low in 7 minutes, critical in 9 minutes, and dead in 10 minutes. These are configs exclusive to this sentry and and other implementations of this sentry can override these values to have a different configuration.

code/modules/defenses/sentry.dm
Drulikar: power_off_action already includes a visible message: https://github.com/cmss13-devs/cmss13/blob/a521b0fbb6459e87ce3feddfdf1f8e63fe9850ef/code/modules/defenses/sentry.dm#L192

code/modules/defenses/sentry.dm
Drulikar: Sure though its because its already a giant examine block so I didn't deem it necessary to include when its not a concern yet. 
image

code/game/objects/effects/effect_system/smoke.dm
Drulikar: ¯\\_(ツ)_/¯ Nothing stopping them from faking or programming to just do it. But regardless synths will not be immune to this smoke for gameplay reasons. If someone cares, they can fix all the other situations that don't synth check in this file & there is an emote.

- /obj/effect/particle_effect/smoke/xeno_weak_fire
- /obj/effect/particle_effect/smoke/xeno_burn
- /obj/effect/particle_effect/smoke/phosphorus
- /obj/effect/particle_effect/smoke/sleepy
- /obj/effect/particle_effect/smoke/bad

code/game/objects/effects/effect_system/smoke.dm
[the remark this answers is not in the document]
vero5123: I feel like that should be handled elsewhere to make sure it never even gets set below 0.

code/modules/defenses/sentry.dm
vero5123: my b.

code/game/objects/effects/effect_system/smoke.dm
Steelpoint: No reason it can't be lore friendly, synthetics are not immune to several environmental factors, so it can be just fluffed up this upgraded chemical weapon also works on Synthetic tissue.

vero5123: > ¯\_(ツ)_/¯ Nothing stopping them from faking or programming to just do it. But regardless synths will not be immune to this smoke for gameplay reasons. If someone cares, they can fix all the other situations that don't synth check in this file & there is an emote.
> 
> * /obj/effect/particle_effect/smoke/xeno_weak_fire
> * /obj/effect/particle_effect/smoke/xeno_burn
> * /obj/effect/particle_effect/smoke/phosphorus
> * /obj/effect/particle_effect/smoke/sleepy
> * /obj/effect/particle_effect/smoke/bad

I ain' gonna do it

Drulikar:
Gas now takes 3 mins to occur, gas itself takes 6 seconds before it will apply damage, and how vehicles handle gas is reworked (now actually spawns gas in the vehicle which both means it now has a visual but also means you don't get damage multiplied depending on how many turfs the vehicle occupies)

harryob (approved):

PR #6379, opened 2024-06-05
Vulture Touchups & Fixes

Zonespace27:
# About the pull request

- Removes RFN requirement for pamphlet, now all squad roles can do it.
- Fixes the skillless spotting scope breaking on the 2nd redeployment
- Fixes the gun's magazine overlay disappearing when the gun ran out of ammo
- Fixes the gun being usable without pamphlet

- The pamphlet change is happening because this gun is inherently an event one. As such, I don't think it needs to be locked to _only_ riflemen.
- Fixes are good

# Changelog
:cl:
balance: All squad roles can now use M707 pamphlets
fix: Skillless M707 spotting scopes now consistently work.
fix: M707 magazine overlay no longer disappears when the gun is empty with an inserted magazine
fix: Fixed M707 being usable without the correct trait
/:cl:

### DIFF
--- a/code/game/objects/items/pamphlets.dm
+++ b/code/game/objects/items/pamphlets.dm
@@ -200,6 +200,10 @@
 		to_chat(user, SPAN_WARNING("You know this already!"))
 		return FALSE
 
+	if(!(user.job in JOB_SQUAD_ROLES_LIST))
+		to_chat(user, SPAN_WARNING("Only squad riflemen can use this."))
+		return FALSE
+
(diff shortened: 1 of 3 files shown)


deathrobotpunch:
I remember once an admin gave me a vulture to field test as a CL. field trials proved a success though the medic that helped with it had every bone broken in their body.

vero5123:
Glad to see zone got around to fixing some of the issues with this. Love the weapon overall.

harryob (approved):

PR #6391, opened 2024-06-07
fixes observe menu bug

zzzmike:
# About the pull request

Fixes observe menu bug, PMC engineer was being listed as Human category and not Survivor round 22539. Looking at the code it seems like the same thing should happen for PMC medic too, so made the same changes there.

My immersion.

<details>

</details>

# Changelog
:cl:
fix: more survivors now listed in survivor category on observe menu
/:cl:


realforest2001 (dismissed):
You don't need to change their assignments, only remove the 'rank' var override from the presets.

zzzmike:
> You don't need to change their assignments, only remove the 'rank' var override from the presets.

ah okay thanks, and thanks for giving the preset in the round as well

Drulikar (approved):
I'm sure there's still more offenders like this, but tackle them as you find them.

PR #6401, opened 2024-06-08
Personal Civilian Clothing Storage Unit For Marines

Steelpoint:
# About the pull request

Adds a set of civilian clothing vendors on the Almayer in the Gym., these vendors can be used by anyone to vend civilian style clothing, the contents are almost identical to the Synthetic clothing vendor, minus the military themed clothing and unique backpack/similar options. These vendors are locked behind a set of shutters with the same high-level access requirements as the Dress uniform vendors, so CIC can determine when access is authorized. 

Anyone can access the vendor, each person has a 120 point limit to what they can buy. 

Grants regular Marines the ability to vend their own personal civilian clothing, for events or whenever CIC authorizes it (Maybe a event where the marines are going on shore leave before something terrible happened?).

The advantage here is to further allow regular players to craft their own characters apperance and roleplay what they might wear off-duty, in circumstances that allow for it. 

The area is locked thus CIC has control on when Marines can access it, and there is also the fact it is the job of the MPs to ensure Marines are wearing proper clothing. 

<details>

</details>

# Changelog
:cl:
mapadd: Civilian Clothing Storage Vendors have been added to the Gym on the USS Almayer. These allow anyone to vend civilian clothes. These vendors require CIC access to unlock however. 
/:cl:

### DIFF
--- a/code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
+++ b/code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
@@ -412,3 +412,105 @@
 		/obj/item/clothing/suit/storage/marine/heavy/skull,
 		/obj/item/clothing/suit/storage/marine/heavy/smooth,
 	)
+
+//------------MARINE CIVILIAN CLOTHING---------------
+
+GLOBAL_LIST_INIT(cm_vending_clothing_marine_snowflake, list(
+	list("SHIRTS AND UNIFORMS", 0, null, null, null),
+	list("White T-Shirt and Brown Jeans", 12, /obj/item/clothing/under/tshirt/w_br, null, VENDOR_ITEM_REGULAR),
+	list("Gray T-Shirt and Blue Jeans", 12, /obj/item/clothing/under/tshirt/gray_blu, null, VENDOR_ITEM_REGULAR),
+	list("Red T-Shirt and Black Jeans", 12, /obj/item/clothing/under/tshirt/r_bla, null, VENDOR_ITEM_REGULAR),
+	list("Frontier Jumpsuit", 12, /obj/item/clothing/under/rank/synthetic/frontier, null, VENDOR_ITEM_REGULAR),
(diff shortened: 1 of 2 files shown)


CeUvi:
What is the reasoning behind locking the vendor under CIC access? It seems like either CIC will open them every round anyway or forget about them entirely and in both those scenarios not having it locked under CIC access sounds a lot better IMO.

hislittlecuzingames:
> What is the reasoning behind locking the vendor under CIC access? It seems like either CIC will open them every round anyway or forget about them entirely and in both those scenarios not having it locked under CIC access sounds a lot better IMO.

Probably beta test or prevent front lines being full of booty jorts and crop tops.

Steelpoint:
> What is the reasoning behind locking the vendor under CIC access? It seems like either CIC will open them every round anyway or forget about them entirely and in both those scenarios not having it locked under CIC access sounds a lot better IMO.

This PR has the most minimal impact on the average game. The MPs, CIC and the Commander are still bound by the SOP's to make Marines follow a proper dress code, and whilst it is the duty of the MPs to ensure Marines follow proper dress code, I nonetheless felt that this implementation of restricting the civilian clothing so that only CIC or admins can allow it to occur, would be the best way to implement it in such a way that this PR would be accepted.

Perhaps there will be room to allow more free access to these items in the future, or even to allow free access in this very PR. But I chose to start conservatively.

Lagomorphica:
> Perhaps there will be room to allow more free access to these items in the future, or even to allow free access in this very PR

if service jackets and service uniforms are locked behind CO access (not just CiC - this PR also locks it behind the CO, which is good and it should be kept as that if it is even merged) then this should also be locked behind the same thing rather than being free access

harryob (approved):
what's the worst that could happen

nuklearcellphoneg:
Look,

Put the Robustness Award Plaque back on the wall. I don't care about the uniform vendor nonsense, let marines die in a jacket if they want to, but put the plaque back.

PR #6407, opened 2024-06-08
Makes Anti-parasitic slightly more powerful to match larva gestation increases

wshuwshuwshi:
# About the pull request

Larva gestation was sped up at one point or another, and that made Anti-parasitic too slow to actually prevent gestation or remove parasites at all
Applies a 1.5 multiplier to the potency removed from the larva timer. 

Larva gestation was sped up at one point or another, and that made Anti-parasitic too slow to actually prevent gestation or remove parasites at all. In its current state, Anti-Parasitic is incapable of removing larvae at higher levels, due to it being too slow acting, with a slight increase, it may be viable (of note, I'm not sure if the 1.5 multiplier is actually high enough to be able to counter-act and remove larvae, and may need to be increased further)
Of note, it still does potency burn damage, meaning that if it is at potency 8, it will do 8 burn damage to whatever the chemical is being metabolized in.

Gives the Antiparasitic chemical property its niche back. At lower levels it's supposed to delay the inevitable and buy time for the marine to get to surgeons, and at higher levels it's supposed to cure larvae from the marine. Currently, it's not powerful enough to do either to a noticeable degree.

N/A. All that was done was add a 1.5 multiplier to the potency, when it is removed from the larva timer

# Changelog
:cl:
fix: Adds a 1.5x modifier to the amount taken from larva timer based on chem property level
fix: swaps the 1.5 multiplier for a delta_time multiplier
fix: improved the formatting of the code to have proper spacing
/:cl:


kiVts:
You need to have a changelog for this, check other PRs.

code/modules/reagents/chemistry_properties/prop_positive.dm
```diff
@@ -492,7 +492,7 @@
 		var/obj/item/alien_embryo/embryo = content
 		if(embryo && istype(embryo))
 			if(embryo.counter > 0)
-				embryo.counter = embryo.counter - potency
+				embryo.counter = embryo.counter - potency*1.5
```
Diegoflores31, line 495: remove the 1.5 modifier and replace it with delta_time this has been caused by beagle adding "speed" to the larva gestation and then fixing APS incorrectly. label this as a fix since his PR was likely not suposed to make APS work worse 
#3844

wshuwshuwshi: Will do. I presume you mean to add delta_time and not multiply by it? Or do you genuinely mean to multiply by delta_time

code/modules/reagents/chemistry_properties/prop_positive.dm
```diff
@@ -492,7 +492,7 @@
 		var/obj/item/alien_embryo/embryo = content
 		if(embryo && istype(embryo))
 			if(embryo.counter > 0)
-				embryo.counter = embryo.counter - potency*1.5
+				embryo.counter = embryo.counter - potency*delta_time
```
Diegoflores31, line 495: dont forget spacing . (( potency * delta_time)) instead of potency*delta_time

wshuwshuwshi: Gotcha, saw a similar format to what I did. I'll correct it now.

Drulikar (approved):

PR #6415, opened 2024-06-09
adds upgraded fuel tanks to the research biomass machine

iloveloopers:
# About the pull request
adds new upgraded smoke and normal fuel tanks printable by the biomass machine
upgraded smoke tank has a 200 unit capacity instead of 100
upgraded custom tank has 60 max intensity and 50 duration, 7 tile range and 200 unit capacity (you still have to make a flamer fuel that reaches these caps)
i will be severely impressed if anyone manages to make a chem that even goes above 50 intensity

lets research reach the full potential of custom flamer chems while still putting it behind a reasonable paywall

it works

# Changelog
:cl:
add: Added new upgraded incinerator fuel tanks printable by the biomass analyzer.
/:cl:


code/modules/projectiles/magazines/flamer.dm
```diff
@@ -249,3 +249,24 @@
...
+//tanks printable by the research biomass machine
+/obj/item/ammo_magazine/flamer_tank/custom/upgraded
+	name = "Upgraded custom incinerator tank"
+	desc = "A fuel tank used to store fuel for use in the M240 incinerator unit. This one has been modified with a larger and more sophisticated internal propellant tank, allowing for bigger capacity and stronger fuels."
+	matter = list("metal" = 50) // no free metal
+	flamer_chem = null
+	max_rounds = 200
+	max_range = 7
+	fuel_pressure = 1
+	max_duration = 100 //if you manage to reach these caps I will be impressed
```
Drulikar, line 262:
```suggestion
	max_duration = 50
```
Should not exceed the power of a large flamer tank.

code/modules/projectiles/magazines/flamer.dm
```diff
@@ -249,3 +249,24 @@
...
+	max_intensity = 100
+	custom = TRUE
+
+/obj/item/ammo_magazine/flamer_tank/smoke/upgraded
+	name = "Large custom incinerator smoke tank"
+	desc = "A tank holding powdered smoke that expands when exposed to an open flame and carries any chemicals along with it. This one has been outfitted with an upgraded internal compressor, allowing for bigger capacity."
+	matter = list("metal" = 50) //no free metal
+	flamer_chem = null
+	custom = TRUE
+	max_rounds = 200
```
Drulikar, line 272:
```suggestion
	max_rounds = 150
```
I'm almost more inclined to just make `/obj/item/ammo_magazine/flamer_tank/smoke` have max rounds at 50 and this be then 100.

code/modules/projectiles/magazines/flamer.dm
```diff
@@ -249,3 +249,24 @@
...
+/obj/item/ammo_magazine/flamer_tank/custom/upgraded
+	name = "Upgraded custom incinerator tank"
+	desc = "A fuel tank used to store fuel for use in the M240 incinerator unit. This one has been modified with a larger and more sophisticated internal propellant tank, allowing for bigger capacity and stronger fuels."
+	matter = list("metal" = 50) // no free metal
+	flamer_chem = null
+	max_rounds = 200
+	max_range = 7
+	fuel_pressure = 1
+	max_duration = 100 //if you manage to reach these caps I will be impressed
+	max_intensity = 100
```
Drulikar, line 263:
```suggestion
	max_intensity = 60
```
Should not exceed the power of a large flamer tank.

code/datums/research_upgrade_datum.dm
```diff
@@ -111,6 +112,26 @@
...
+	behavior = null
+	upgrade_type = ITEM_ACCESSORY_UPGRADE
+
+/datum/research_upgrades/item/flamer_tank/smoke
+	name = "Upgraded Incinerator Smoke Tank"
+	desc = "An upgraded incinerator smoke tank with a larger capacity."
+	value_upgrade = 100 //not useful enough to be expensive
+	clearance_req = 1
+	item_reference = /obj/item/ammo_magazine/flamer_tank/smoke/upgraded
+	change_purchase = 0
```
Drulikar, line 132:
```suggestion
	change_purchase = -100
```

code/datums/research_upgrade_datum.dm
```diff
@@ -111,6 +112,26 @@
...
+	minimum_price = 0
```
Drulikar, line 133:
```suggestion
	minimum_price = 100
```

code/datums/research_upgrade_datum.dm
```diff
@@ -111,6 +112,26 @@
...
 
+/datum/research_upgrades/item/flamer_tank
+	name = "Upgraded Incinerator Tank"
+	desc = "An upgraded incinerator tank, with larger capacity and able to handle stronger fuels."
+	value_upgrade = 500
+	clearance_req = 1
+	change_purchase = -100
+	minimum_price = 100
+	item_reference = /obj/item/ammo_magazine/flamer_tank/custom/upgraded
+	behavior = null
```
Drulikar, line 123: Why null behavior? Shouldn't this be `RESEARCH_UPGRADE_TIER_1` or something?

Drulikar (changes requested):

code/datums/research_upgrade_datum.dm
iloveloopers: if I do use RESEARCH_UPGRADE_TIER_1 the icon_state when the item gets printed is changed for some reason. Hundson told me to just set it to null

Drulikar: So again I ask. Why should this be the first to have a null behavior?

code/datums/research_upgrade_datum.dm
Drulikar: Unresolved - you need to explain why you insist on this being different than any of the other repeatables.

code/datums/research_upgrade_datum.dm
Drulikar: Also still unresolved because you opted to just remove it instead.

code/datums/research_upgrade_datum.dm
kiVts: because behavior is passed to init of an item, a null _should_ just not pass anything, 


Nevermind Im wrong, null gets passed into init, and in this case, right here:
image
And since its one(aka true) it does that.

harmless here, but I really should have added a check for null  earlier when pr was open now that I think about it

iloveloopers:
null define added

code/datums/research_upgrade_datum.dm
Drulikar: As a define now this is acceptable; but in a different PR `behavior` really needs to be handled differently. Its a neat idea to allow specifying an argument into the Initialize; but its too unintuitive and restrictive to assume everything should accept a number 1-5.

PR #6438, opened 2024-06-11
UNTIL_OR_TIMEOUT macro

Zonespace27:
# About the pull request
UNTIL() but you can add a timeout. This shouldn't be used except with rustg HTTP requests. Complete bandaid, but it's what needs to be done

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1501,6 +1501,17 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 
 #define UNTIL(X) while(!(X)) stoplag()
 
+#define UNTIL_OR_TIMEOUT(X, __time) \
+	do {\
+		var/__start_time = world.time;\
+		while(!(X)) {;\
+			stoplag();\
+			if(__start_time + __time <= world.time) {;\
+				CRASH("UNTIL_OR_TIMEOUT hit timeout limit of [__time]");\
+			};\
+		};\
+	} while(FALSE)
+
 //Repopulates sortedAreas list
 /proc/repopulate_sorted_areas()
 	GLOB.sorted_areas = list()


code/__HELPERS/unsorted.dm
```diff
@@ -1501,6 +1501,17 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 
 #define UNTIL(X) while(!(X)) stoplag()
 
+#define UNTIL_OR_TIMEOUT(X, __time) \
+	do {\
+		var/__start_time = world.time;\
+		while(!(X)) {;\
+			stoplag();\
+			if(__start_time + __time <= world.time) {;\
+				CRASH("UNTIL_OR_TIMEOUT hit timeout limit of [__time]");\
```
Drulikar, line 1510: A crash will end the proc call, and as a macro, that means that the proc that uses this macro will be the one ended. Are you sure you want a timeout to not fail gracefully? Consider `stack_trace` if your goal was only to have a verbose error.

Drulikar (changes requested):

code/__HELPERS/unsorted.dm
```diff
@@ -1501,6 +1501,17 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 
 #define UNTIL(X) while(!(X)) stoplag()
 
+#define UNTIL_OR_TIMEOUT(X, __time) \
```
Drulikar, line 1504: __time also might better if it was forced to be >= 0 (e.g. with max it and 0) else the timeout will never possibly occur.

code/__HELPERS/unsorted.dm
Zonespace27: That's intended, yes. If the UNTIL_OR_TIMEOUT() of a `/datum/http_request` goes on until the specified `__time` timeout, then we know that the query won't work and should abort to prevent any further lag to the server.

code/__HELPERS/unsorted.dm
Zonespace27: done

Drulikar (approved):

harryob (approved):
lets see

PR #6470, opened 2024-06-15
Fixes a bug that would jam guns entirely

GrrrKitten:
# About the pull request

I couldn't figure out what caused this bug but it was getting kinda bad, something about alternating akimbo would cause guns to be entirely disabled, with the current_mag having been ejected, but still marked as in the gun, but not in the contents, meaning the gun was entirely broke as you couldnt clear out the mag inside of it cause there was no mag.

this pr just adds a failsafe to prevent this from ever happening

Having your gun becoming entirely unusable and having to just toss it out cause a bug broke it isnt fun

<details>

</details>

# Changelog

:cl:
fix: fixes a bug that fully jammed guns being used in akimbo
/:cl:


harryob (approved):
as much as i hate failsafes to catch conditions that we should just not allow to happen, it's fine

PR #6498, opened 2024-06-17
Removes delay when failing to move

Git-Nivrak:
# About the pull request
Remove the delay when you fail to move (aka move into a dense object)
(there is still a minimal delay to reduce spam on code)

Makes moving a lot smoother, currently it just feels like your clicks are getting eaten which gets worse with lower movement speeds and higher ping (what is happening is you get full move delay if you move into a dense object so if you mistime it looks like your clicks get eaten)

Both videos are me holding the W key and pressing the D key rapidly

Current:

https://github.com/cmss13-devs/cmss13/assets/59925169/6c8051d9-5433-4540-8d42-7d1171acfaeb

 

After the PR:

https://github.com/cmss13-devs/cmss13/assets/59925169/fbbeb784-f7ba-4c83-8a03-59bea51d347f

# Changelog

:cl:
qol: Removed delay when failing to move
fix: You longer take damage from shrapnel when failing to move
/:cl:

### DIFF
--- a/code/__DEFINES/dcs/signals/atom/mob/signals_mob.dm
+++ b/code/__DEFINES/dcs/signals/atom/mob/signals_mob.dm
@@ -185,3 +185,6 @@
 #define COMSIG_MOB_END_TUTORIAL "mob_end_tutorial"
 
 #define COMSIG_MOB_NESTED "mob_nested"
+
+/// When a mob successfully moves to a different tile (wasn't blocked) : (client/Move(n, direct))
+#define COMSIG_MOB_MOVE_SUCCESS "mob_move_success"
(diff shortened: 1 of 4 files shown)


iloveloopers:
how does this affect bodyblocking?

Git-Nivrak:
> how does this affect bodyblocking?

Needs to be TMed to find out

Git-Nivrak:
From some wacky testing it looks like it makes it impossible to bodyblock mobs that a bigger than you (mobs that you can't push)
This can probably be fixed but I am not sure how much of a bad thing it is

Git-Nivrak:
Actually it is probably a bad thing since it allows you to push things at full speed

Drulikar:
I am not inclined this should change. Nor does it eliminate the possibility of inputs being dropped in the situation its explicitly trying to improve. Waiting on other maintainer feedback before just closing however.

Butlerblock:
In a similar question to body blocking, does this rapidly speed-up the rate at which non-anchored objects are pushed? This could inadvertently be much more powerful for xenos. If I had a suggestion, it might be be more appropriate to only remove the delay if the user pushes against a non-anchored object.

cm13-github:

PR #6500, opened 2024-06-17
Adds complete vent system to Fiorina

HeresKozmos:
# About the pull request

Adds a complete vent system to Fiorina.

Unlike most other maps, Fiorina doesn't have a vent system at all; which this fixes and brings it in line with the other maps.

To stop you from going from one edge of the map to the other, each major area has its own closed loop vent system; for example you can only enter a vent inside of Engineering and exit from somewhere else inside Engineering. I've added grates where possible to help with identifying and locating pipes so you don't have to rip up the entire floor.

The main point of this PR is to bring it up to the same level that large maps like Chance's Claim or New Varadero have when it comes to vent availability.

This is an image of Fiorina with no turf, you can zoom in and follow the pipes to look at where they go to and from:
Fiorina Vent System - Turfless

# Changelog

:cl:
maptweak: Adds closed loop vent systems to each major area in Fiorina.
maptweak: Changes some floors to include grates to make pipes visible/easier to locate.
maptweak: Moved a few minor map props to make space for vents a short distance from their original spots.
/:cl:

### DIFF
--- a/maps/map_files/FOP_v3_Sciannex/Fiorina_SciAnnex.dmm
+++ b/maps/map_files/FOP_v3_Sciannex/Fiorina_SciAnnex.dmm
@@ -102,6 +102,25 @@
 	icon_state = "darkyellow2"
 	},
 /area/fiorina/station/telecomm/lz1_cargo)
+"adh" = (
+/obj/structure/pipes/standard/simple/hidden/supply{
+	dir = 4;
+	icon_state = "exposed01-supply"
+	},
+/turf/open/floor/prison{
+	dir = 5;
+	icon_state = "darkpurple2"
+	},
+/area/fiorina/tumor/servers)
+"adl" = (
+/obj/structure/pipes/standard/simple/hidden/supply{
+	dir = 10
+	},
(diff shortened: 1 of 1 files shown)


Drulikar:
Stale exempt until mapping reviewed.

cm13-github:

maps/map_files/FOP_v3_Sciannex/Fiorina_SciAnnex.dmm
Nanu308, line 1: image
image
image
image
image
image

Only one entrance per vent. You're using two threeway's for no raisin!!!

image
Missing pipe

image
Don't run pipes under walls

image
You also has a lot of redundancy in your pipe system, tldr. it looping into itself and having several way to reach the same spot. It's fine to a certain degree, but I'd say cut a few of these pathways and in some other areas.

One scrubber and one vent pump per room is the standard, look at the Almayer as an example. Of course, is the room massive you can add in another set of two and so forth.

Nanu308 (changes requested):

PR #6505, opened 2024-06-18
Makes worn_id a private var

Zonespace27:
# About the pull request

Title

Stops https://github.com/cmss13-devs/cmss13/pull/6411 from happening again


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -369,6 +369,7 @@
 
 //gets ID card object from special clothes slot or null.
 /mob/living/carbon/human/proc/get_idcard()
+	RETURN_TYPE(/obj/item)
```
Drulikar, line 372: Wouldn't this be more correct as `/obj/item/card/id`

code/_onclick/hud/human.dm
Drulikar, line 1: Is it not correct that the changes here are returning the ID card. It just makes it so if say you have your ID in a wallet, the wallet will disappear because somereason you're having it directly search for the ID card: 
id

Think of `wear_id` merely as the inventory SLOT. It was just merely incorrect before that things would assume the only thing that could be slotted was an ID card.

Drulikar (changes requested):

Drulikar:
In general I just really don't think this is worth doing. There's nothing special about why this inventory slot needs to be treated differently than any other inventory slot. It was just an incorrect assumption that code made before that the only thing that could be slotted was of the type `/obj/item/card/id` and the definition being typed to that is what furthered that wrong assumption.

PR #6511, opened 2024-06-19
Nuclear Authentication Disks reduce Nuclear Weapon decryption time.

blackdragonTOW:
# About the pull request

Adds a simple 30% reduction in what ever decryption time is left when we use a Nuclear Authentication Disk on a nuke.

Note: Has no effect on the admin spawned nuke which does not have a decryption timer.

Presently our maps have Nuclear Authentication Disks, and when found it's treated with some semblance of gravity, but in actuality they do nothing.  This rewards finding such an important object and in the occasion that the disk is found and the End Game for the round is also a decryption nuke.

Tested with all listed fail states contained within /obj/structure/machinery/nuclearbomb/tech/attackby()

<details>

</details>

# Changelog
:cl:
add: Nuclear Authentication Disks now reduce remaining decryption time by 30%.
/:cl:

### DIFF
--- a/code/game/machinery/nuclearbomb.dm
+++ b/code/game/machinery/nuclearbomb.dm
@@ -632,22 +632,18 @@ GLOBAL_VAR_INIT(bomb_set, FALSE)
 
 /obj/structure/machinery/nuclearbomb/tech/attackby(obj/item/nuke_disk as obj, mob/user as mob)
 	var/reduction_multiplier = 0.7 //30% reduction
-	if(istype(nuke_disk, /obj/item/disk/nuclear))
+	if(!istype(nuke_disk, /obj/item/disk/nuclear))
 		attack_hand(user)
-		if(decrypting)
-			if (decryption_time > 0)
-				decryption_end_time = LERP(world.time, decryption_end_time, reduction_multiplier)
-				decryption_time = decryption_end_time - world.time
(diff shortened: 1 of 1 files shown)


vero5123:
lord, you need to do early returns man.

blackdragonTOW:
> lord, you need to do early returns man.

Yeahhhh, still learning, tried it another way that should be easier to read.

Drulikar (changes requested):
As mentioned on the discord; maybe tie this instead into self destruction. E.g. 50% faster or just instant SD or something. Already xenos struggle to get to a nuke fast enough. If marines are able to manage to get objectives to get a SD online, then this would encourage people to actually go for that; whereas now SD is really rare.

Steelpoint:
Maybe the disk allows you to set a custom detonation timer? Instead of a flat 60 seconds, you can set it to detonate either instantly or up to several minutes slower than normal?

blackdragonTOW:
> Maybe the disk allows you to set a custom detonation timer? Instead of a flat 60 seconds, you can set it to detonate either instantly or up to several minutes slower than normal?

Likely would get shotdown faster than just the decrypt timer.  Marines wont set it longer than 60s and making it shorter than 60s would be "unfair" according to our resident xeno experts. 

Regardless changing this to making the reactors behave differently during SD is way off Theme for what this sets out to do so I'll just kill it here.

PR #6527, opened 2024-06-22
Vomitting now empties your stomach

blackdragonTOW:
# About the pull request

After much fun talking to a medic about inducing vomiting to remove toxins/pills I figured this might be a fun addition.

- Vomiting now removes 30% of whatever is in your stomach
- PROPERTY_EMETIC now has use
- Emetic's vomit chance was very low for a chemically induced trigger
- Created a pill packet, pill, and reagent for Ipicac
- Added the pill packet to WeyYu Med vendors

This allows players to deal with ingested chems rapidly at the cost of getting rid of ALL of the chems ingested and the imbiber not having a great time as they eject whatever is in them.

<details>

</details>

# Changelog
:cl:
add: Ipicac Pills
code: Vomiting now removes 30% of everything in the users system
imageadd: added Ipicac Pills
/:cl:


blackdragonTOW:
I'm stupid

blackdragonTOW:
Still learning source control. :C

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -894,9 +894,14 @@
 	if(istype(location, /turf))
 		location.add_vomit_floor(src, 1)
 
+	if(reagents.reagent_list.len)
```
Zonespace27, line 897: length(), not .len

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -894,9 +894,14 @@
...
+		var/purge_percent = 0.7 //30%
```
Zonespace27, line 898: what's the point of making this a local var if you're using it nowhere else but line 900

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -894,9 +894,14 @@
...
+		for(var/datum/reagent/ingested_chem in reagents.reagent_list)
```
Zonespace27, line 899: as anything

code/modules/reagents/chemistry_reagents/medical.dm
```diff
@@ -411,3 +411,12 @@
 	custom_metabolism = AMOUNT_PER_TIME(1, 200 SECONDS)
 	data = 0
 	properties = list(PROPERTY_CURING = 2)
+
+/datum/reagent/medical/ipicac
```
Zonespace27, line 415: Isn't it Ipecac

Zonespace27 (changes requested):

code/modules/mob/living/carbon/human/human.dm
blackdragonTOW: if(reagents.reagent_list.length())

code/modules/mob/living/carbon/human/human.dm
blackdragonTOW: Originally I was just exposing vars for ease of seeing what value was being used (I thought I needed three, but over time managed to whittle it down to one) I'll just set line 900 to * 0.7

code/modules/reagents/chemistry_reagents/medical.dm
blackdragonTOW: ...........yes

code/modules/mob/living/carbon/human/human.dm
blackdragonTOW: for(var/datum/reagent/ingested_chem as anything in reagents.reagent_list)

code/modules/mob/living/carbon/human/human.dm
blackdragonTOW: Sorry, 	if(length(reagents.reagent_list))

My javascript background comes back to haunt me.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -894,9 +894,13 @@
 	if(istype(location, /turf))
 		location.add_vomit_floor(src, 1)
 
+	if(length(reagents.reagent_list))
+		for(var/datum/reagent/ingested_chem as anything in reagents.reagent_list)
+			ingested_chem.volume = (ingested_chem.volume * 0.7)
```
Drulikar, line 899: This behavior of removing chems in your stomach needs to set a cooldown of something like I dunno 15 minutes or so. Doesn't need to restrict the act of vomiting nor the stun associated, but just "You heave but your stomach is empty" or something along those lines.

Effectively what this is imitating is that chems that haven't entered your bloodstream yet you can eject from your stomach without going so far as tracking the age of each reagent in the body.

Drulikar (changes requested):

code/modules/mob/living/carbon/human/human.dm
blackdragonTOW: > This behavior of removing chems in your stomach needs to set a cooldown of something like I dunno 15 minutes or so. Doesn't need to restrict the act of vomiting nor the stun associated, but just "You heave but your stomach is empty" or something along those lines.
> 
> Effectively what this is imitating is that chems that haven't entered your bloodstream yet you can eject from your stomach without going so far as tracking the age of each reagent in the body.

That's what the percentage does in this case.  It's impossible to vomit the "entirety" of whatever is in you.  The larger the volume (and thus, technically a more "recently ingested" chem) the more you'll purge, the smaller values however will remove even less as say if you had 16u of a chem, even 6 vomit cycles (a very long time in this case) would reduce that number down to 2u.  (200u of something though very clearly loses a lot more than that on the first purge)

Regardless, the percentage based system specifically works to target higher volume (and again: thus more recent) items in the system rather than long sitting, smaller value chems.   I feel the curve that comes with 30% to be smartly applied in this case and would ask a reconsideration on a cooldown as I don't think anyone is willing to wait 15 minutes to actually use this potential avenue for the sake of vomitting twice (and there would be no reasonable way to express to the player that the 15min interval was up and their next puke will be another "good" one)

Drulikar: Those are the requested changes. Merely having a percentage means nothing. Simply ingest lots of some harmless reagent to work around it. The alternative is to have a more in depth tracking of whether reagents would be new enough to still reside in the stomach.

blackdragonTOW: Aight well, that's way off theme and a 15 minute timer makes the entire feature useless.  GG I guess.

Drulikar: All this ultimately is conveying stomach emptiness so this can't be abused. Behaves the same in real life. You can only vomit so much before there's nothing to vomit.

blackdragonTOW: I guess what you mean but the timer inherent to vomit() and the random draw already means that you're naturally going to be spending 30s-50s between purges.   Asking for a 900s cooldown just means that people will naturally metabolize 90u (rate @ .1 per sec) of whatever is in their system which baring a laughably horrible OD where you've ingested 200u of Bica somehow, this would never affect them.   Subjecting yourself to uncontrollable stuns for a couple minutes should have a tangible reward on the other end of it, especially since the best case for this is for OD control.  I don't see a world where OD control can really be abused in any capacity as being OD'd never conveys any benefits, and thus you undergo a slightly less bad outcome to not have to deal with a worse outcome.  It's an emergency stopgap.  The alternative is getting connected to the blood cleaner device which has zero downsides and takes seconds to purge -everything- from your system.  The bonus here is that you don't have to go back to FOB, the trade off being that you're going to be screwed for two minutes while you wait (time that could have been spent going to FOB)

Regardless, I'll close this out as it's not going to be considered.

Drulikar: I don't know how you got the notion that I was just outright denying what you were intended to do. I just requested that using this substance to evoke vomiting needs to be something that can only be done sparingly, so to simulate how you actually need contents in your stomach but we don't actually simulate stomach contents getting absorbed, to just put the positive effect of it on a cooldown.

I see this as a in-field way to deal with an overdose. I would want it to be an inferior solution to hemodialysis. But this shouldn't be a tool that can just be used freely. It should only be used in an emergency like with an overdose.

Drulikar: If your concern was that with a cooldown like this then the 30% is too small of an effect, sure then make it 50% or something. I wasn't even hard on how long the time needed to be but IMO I think 15 mins is probably too short. Someone shouldn't be having to deal with ODs that quickly.

PR #6534, opened 2024-06-22
Swaps MOTD and TM load order in chatbox, nukes no admin event message

GrrrKitten:
# About the pull request

Removes the 
image
message from showing up

Also swaps MOTD and TM message so people actually see the MOTD instead of 10 TM's

pt 1) if there isnt a message set for admin events, you can easily intuit that admins arent running an event or they didnt think it was necessary

pt 2) people werent seeing the MOTD cause it would get pushed up by 10 tm's, meaning people weren't seeing potentially important info

<details>

</details>

# Changelog

:cl:
ui: Swaps TM and MOTD order so that people actually see the MOTD
admin: Removes the message for when no event info is input
/:cl:


vero5123:
debby

harryob (changes requested):
this doesn't do what you want, you need to change the order they're sent in /mob/new_player/proc/lobby()

harryob (approved):

PR #6537, opened 2024-06-23
Only researchers can now use the biomass machine

iloveloopers:
# About the pull request
only USCM researchers can now use the biomass analyzer

# Explain why it's good for the game
The biomass machine is there for researchers, CMO or survs shouldn't be able to barge in and waste all of your points instantly because "I want to"

# Testing Photographs and Procedure
it works

# Changelog
:cl:
add: Only researchers can now use the biomass analyzer.
/:cl:

PR #6538, opened 2024-06-23
CMO no longer has research skill

iloveloopers:
# About the pull request
CMOs no longer have research skill

you should roll for researcher if you want to do research.
no more CMOs barging in research and wasting the team's credits and biomass while they can't do anything about it.

yes

# Changelog
:cl:
del: CMO no longer has research skill
/:cl:


ihatethisengine:
CMO is your boss, you are not independent. It's like being mad at QM as a CT. Play CMO if you don't want someone to spend your credits.

Even Fira's (CT main) PR to remove ASO req access was denied: https://github.com/cmss13-devs/cmss13/pull/5038

iloveloopers:
multikeying is truly the only option

Drulikar (changes requested):
This is a worse solution than #6537 because we don't have a mechanic that allows someone to train up their skills. Whereas access can be ID modded. Regardless as I mentioned there, really only should apply to survivors but you should be able to ID mod them if command wants them to help out.

PR #6564, opened 2024-06-25
"Upgraded custom incinerator tank" nerf

kalazus:
# About the pull request

Downgrades flamer_tank/custom/upgraded to flamer_tank/large/X (almost, napalmx has 40 duration instead of 50). Pyro still has even better custom tank.

It's too op for non specialists to have 1.5(+fire penetration, +1 range) better fire than roundstart pyro, and same as custom pyro. It is not ok for non specialist to deal about 500 damage (initial setting on fire and 10 second before extinguish) in 1 click of flamer. Also, it makes pyro with custom fuel fully replaceable.

<details>

</details>

# Changelog

:cl:
balance: reduces max intensity and max range of "Upgraded custom incinerator tank".
/:cl:

### DIFF
--- a/code/modules/projectiles/magazines/flamer.dm
+++ b/code/modules/projectiles/magazines/flamer.dm
@@ -257,10 +257,9 @@
 	matter = list("metal" = 50) // no free metal
 	flamer_chem = null
 	max_rounds = 200
-	max_range = 7
+	max_range = 6
 	fuel_pressure = 1
 	max_duration = 50
-	max_intensity = 60
 	custom = TRUE
 
 /obj/item/ammo_magazine/flamer_tank/smoke/upgraded


iloveloopers:
At least lower it's price in the biomass machine to be dirt cheap since this nerfs it's main advantage over the free custom tanks printable at OT. The range buff is minimal and duration barely matters at the front where xenos are together and will vomit/pat each other.

code/modules/projectiles/magazines/flamer.dm
```diff
@@ -257,10 +257,9 @@
 	matter = list("metal" = 50) // no free metal
 	flamer_chem = null
 	max_rounds = 200
-	max_range = 7
+	max_range = 6
 	fuel_pressure = 1
 	max_duration = 50
```
Drulikar, line 262:
```suggestion
	max_duration = 50
	max_intensity = 50
```

Drulikar (changes requested):
Alternatively, if you feel they're too plentiful/powerful, increase the cost of them in research and or have them scale faster.

Of note they are still empty tanks so they *also* need the chemicals in them too to even utilize their increased limits.

TheManWithNoHands:
Ok, ignoring the fact for a moment that these tanks need to be earnd through killed xenos and arent free.

You do know, that the values your changing are only a cap. They dont change the power of for exampl: napalm in them.
They still need custom fuel, research needs to spend points on, to reach said cap. Or fuel that takes unresonable ammoutn.
Something they where always able to do. And often use for Ot nades.

Like, its very much earned to effectivly used those tanks.

code/modules/projectiles/magazines/flamer.dm
kalazus: They are too powerful, not plentiful (at least haven't seen 3 with it at the same time). It replaces Pyro with research which is no ok. I'm not ok even with napalmx level fire for non specialist (replaces pyro without custom tank), but flamer_tank/custom was added 3 years ago, and wasn't such big of a deal.
What do you mean by "have them scale faster"?

kalazus:
If it will be too garbage with these changes, what about full revert than?
Pyro still has flamer_tank/large/custom to use "much earned" chemicals.

iloveloopers:
We should probably take this balance discussion into discord but why shouldn't marines get stronger stuff as the round progresses? you're acting as if marines get their hands on this stuff 5 minutes after drop. Theres been about 5 rounds total where strong research flame was at play and they always arrived at the frontlines after 1:20 - 1:30~ hours or after xenos had already hijacked.

TheManWithNoHands:
> If it will be too garbage with these changes, what about full revert then? Pyro still has flamer_tank/large/custom to use "much earned" chemicals.

No, your change is just salt driven. Because, you wanna know a fun thing?
Base m240 OT tanks are capable of using flamer fuel with intensity 40. The intensity of Xfuel. The very thing you said shouldnt be allowed.

Besides that, Research tanks are the fairest thing research can even make. Its the only thing with a combat impact that is finite (cant make infinite more of them once you unlcoked it like chems) are destroyable, rquire a constant influx of new rescources, and cant be automaticly unlocked via stalling. Unlick stims and OT chems, wich can.

code/modules/projectiles/magazines/flamer.dm
Drulikar: See the parameters for https://github.com/cmss13-devs/cmss13/blob/31e1ffbacbed8194287286d1092e16456592539b/code/datums/research_upgrade_datum.dm#L115-L135

It is currently 300 biomass to produce one, requires the first clearance level, and each print of one increases the cost by 100 biomass, to a max of 1000 biomass.

As far as outclassing pyro, OT/researchers can also give pyro better chems in a /obj/item/ammo_magazine/flamer_tank/custom/large which doesn't need any unlock or has any scaling cost.

iloveloopers: I'd like to note that the clearance level is mostly irrelevant as you need clearance 5 anyways to produce something capable of reaching the tank's caps

kalazus: If you have chems you can use tank infinitely (scaling and initial value doesnt matter). Also, 300 biomass is nothing for brand new pyro specialist.

Diegoflores31:
The PR has not been UP for more than 2 weeks and you are already asking for a nerf ? I literally never ever seen it on a round since it was merged.

You are basically nerfing a recently approved balance PR with no justification at all.

Diegoflores31:
> We should probably take this balance discussion into discord but why shouldn't marines get stronger stuff as the round progresses?

Bro this dude is just nerfing it for no reason . Upgraded flamer has not even been on the game for 2 weeks and literally never ever seen it used ingame.

PR #6571, opened 2024-06-26
Removes rank requirement from prosthetic limbs

wshuwshuwshi:
# About the pull request

Removes the rank requirements and the section of code which checks for rank requirements on prosthetic limbs

From a balance perspective, robotic limbs don't provide any benefit to those who have them and often are more of a detriment
From a fluff perspective, marines are given robotic limbs in round whenever they lose a limb, so it can be annoying if you want to add some more fluff to a character when they lost a limb last round and you aren't able to, even though you had one ten minutes ago.

Tested on a local host. All works as it should
<details>

image

</details>

# Changelog
:cl:
del: Removed rank requirements for robotic limbs
/:cl:


ihatethisengine:
I think it was removed because no bonebreaks/bleeding/pain is a bit too strong and does in fact provide a benefit.

wshuwshuwshi:
They don't have literal fractures, but they do break
The limb 'malfunctions' and becomes unusable, meaning you drop what you're holding, or fall over. Likewise it causes slowdown per malfunctioning limb
You also cannot splint them, nor do they heal over time or with chemicals
That being said, they do not bleed, and do not suffer internal bleeding

Diegoflores31:
> I think it was removed because no bonebreaks/bleeding/pain is a bit too strong and does in fact provide a benefit.

pretty sure it was removed because normal marines are supposed to be "ready for battle" with no disabilities nor mental problems when they are woken up for action its on discord somewhere

ihatethisengine:
> The limb 'malfunctions' and becomes unusable, meaning you drop what you're holding, or fall over. Likewise it causes slowdown per malfunctioning limb

Yes, but it's a lot easier to fix than a bonebreak that gives you debuffs even if splinted.

BonniePandora:
> > I think it was removed because no bonebreaks/bleeding/pain is a bit too strong and does in fact provide a benefit.
> 
> pretty sure it was removed because normal marines are supposed to be "ready for battle" with no disabilities nor mental problems when they are woken up for action its on discord somewhere

Yet any rank can seemingly take the bad leg, which looks like a much more delibating sort of trait than having a robotic leg.

Honestly, if it was made available for any non-squaddie role it'd be quite nice, as most of those don't go off running into combat (With IO's being a potential exception, though they're already able to take the robotic leg traits as 2nd LT's currently)

Diegoflores31:
> Yet any rank can seemingly take the bad leg, which looks like a much more delibating sort of trait than ha
> Honestly, if it was made available for any non-squaddie role it'd be quite nice, as most of those don't go off running into combat (With IO's being a potential exception, though they're already able to take the robotic leg traits as 2nd LT's currently)

Bad leg is a trait that was recently added , it's an oversight that You can take it as a combat role

wshuwshuwshi:
> I think it was removed because no bonebreaks/bleeding/pain is a bit too strong and does in fact provide a benefit.

Strangely enough? Robotic limbs DO feel pain when you get hurt. My presumption is that pain is handled through total damage to the person rather than individual limbs.
That being said, robotic limbs don't get internal bleeding, and their fractures are easier fixed
Trade-off being that you cannot heal them through chemicals (unless research slaps the one property in them which they rarely will), nor with standard medical equipment, you cannot get a buddy to splint them, instead requiring someone with surgical skill to reset them.

wshuwshuwshi:
> Bad leg is a trait that was recently added , it's an oversight that You can take it as a combat role

As a genuine statement do you want me to just make a quick PR to make it so that you can't take Bad Leg as a combat role? Shouldn't be too hard to just put the same rank restriction that robotic limbs currently have on it.

Drulikar:
It'd be okay to have this for non-combat roles, but I don't want this for combat roles. I doesn't matter if they lose a limb mid-round and get a prosthetic during the round. But they shouldn't be trying to use traits for any balance reason.

And yes if bad leg should get the same treatment where it can only be applied to non-combat roles.

PR #6573, opened 2024-06-26
M4RA Burstifre

joaopslima:
# About the pull request
Adds burstfire mode to the M4RA

Makes the M4RA a bit nicer to use. 

<details>

</details>

# Changelog
:cl:
add: The M4RA battle rifle now has a two-shot burst-fire mode.
/:cl:


Red-byte3D:
weapon named dmr

HeresKozmos:
I don't see why the M4RA should get burst fire or why it even needs it in the first place, especially since the bullet damage got buffed a while back.

GrrrKitten:
If it's a 2 shot burst fire, why not have it be via the burst fire augmentor attachment

joaopslima:
The scout M4RA had already got burst-fire, so I thought it made sense to add it to the normal gun and to make it a little nicer to use too. Does the code work as intended though? I did it by intuition and might look more into DM in the future

Zonespace27:
Thank you for the PR, but I don't think there's interest in this for the moment. I do think the M4RA needs some work, but this isn't it. (Also in the future, I'd recommend having more than 1 sentence in "why this is good for the game")

PR #6576, opened 2024-06-27
Autocanon and Minigun sound touchup

TheManWithNoHands:
# About the pull request

* Added a new SFX for the autocannon as a hitsound. The sound was created using already existing sfx in the code and aditional royalty free sfx.
* Swapped the miniguns firing sound with the sound thats specificly for it, instead of the upp gatling gun sound. Sfx already existed. This is meerly a code change.

The autocannon feels very lackluster rigth now and lacks any kind of feedback for what it can actually do. Its a big problem of feel then balance.
The fact that it causes AOE damage is not communicated in any way for the enemy besides red damaged flashes, that could also come from any other damage source. The new sound should give people more feedback in the damage they have taken.
From feedback on discord its gonna stay sound only for now. If the feedback is not enough, migth need to ad a visual indicator in the future as well.

The mingun for the tank has three sound files speccificly for it that wherent used. Instead it used the gatling gun sound file the code was likly based on.
I didnt manage to figure out how to properly implement rev up sound and rev down, so it only replaces the loop sound.
If anyone knows how to implement the start end sounds ping me, i got plans with that for more cool things for this.
The sounds will also make the minigun feel more beefy, as the main weapon of a tank should feel. Should make it a much more fun experince to fire the gun.

Autocannon hit sound:
https://jmp.sh/s/F6J3BMTpnVsjy9Twci4g

Minigun sound:
https://jmp.sh/s/dYtaYUmRblRBothUmCbR

# Changelog
:cl: TheManWithNoHands

soundadd: added a new sound for autocannon

/:cl:


Doubleumc:
I'm the one that messed with the minigun sound. I'll explain why.

Originally the minigun (and all vehicle weapons) did not have the autofire system implemented. That's the thing that lets them keep firing when you hold the trigger. The minigun was effectively semi-auto with each firing being a short burst. It would play the minigun firing sound at the start of each burst. The sound was of several shots being fired, which made sense for the burst.

Now the minigun *does* use the autofire system, so as you hold down the trigger it keeps firing shots at a faster and faster rate, with a firing sound played per shot. The original sound was long enough (2 seconds!), loud enough, and the gun fired fast enough that the sounds started stacking on top of each other which became obnoxious. I switched it to the closest equivalent, the UPP minigun sound, which was shorter and softer.

For your goal I suggest removing the firing sound entirely and instead use a looping sound. Looping sounds support a start, end, and middle looping sound: https://github.com/cmss13-devs/cmss13/blob/29d93a39ce7dfac85b33198bd9486688e85d72da/code/datums/looping_sounds/_looping_sound.dm

TheManWithNoHands:
Thats actually really helpfull.
I been trying to make it work but didnt know how to.
I will look into fixing it.

TheManWithNoHands:
need more time for fixing

PR #6582, opened 2024-06-28
Moves click_adjacent logic to MouseDown

Git-Nivrak:
# About the pull request
Makes it so adjacent clicks are handled on when you press down the button instead of when you press it up
This affects things like pointblanks, melee attacks, picking up items etc etc

Did some testing and fixed things like clicks not working or no click delay melee attacks but there are probably more so it needs to be tested

Pointblank wise all of the gun code was already using MouseDown so no reason for pointblanks to stay behind, Other than that this should make adjacent clicks get "eaten" less (since if you too slow to mouse up the click won't be counted) and overall faster reaction time to your actions
<details>

</details>

# Changelog
:cl:
balance: Adjacent clicks are now registered on mouse down instead of mouse up, this affects pointblanks, melee attacks etc.
/:cl:

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -17,12 +17,13 @@
 	~ BMC777
 */
 
-/client/Click(atom/A, location, control, params)
+/client/Click(atom/A, location, control, params, mouse_down=FALSE)
+	to_world("Mouse Down: [mouse_down] Ignore Next Click: [ignore_next_click]")
 	if (control && !ignore_next_click) // No .click macros allowed, and only one click per mousedown.
 		ignore_next_click = TRUE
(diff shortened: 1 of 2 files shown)


code/_onclick/click.dm
```diff
@@ -17,12 +17,13 @@
 	~ BMC777
 */
 
-/client/Click(atom/A, location, control, params)
+/client/Click(atom/A, location, control, params, mouse_down=FALSE)
+	to_world("Mouse Down: [mouse_down] Ignore Next Click: [ignore_next_click]")
```
Doubleumc, line 21:
```suggestion
```

PR #6584, opened 2024-06-28
Some vehicle work around

blackcrystall:
# About the pull request

Just some fix of code mess and creating new mess plus downgrade, solution to polution

The only thing that affect game, vehicle arsc can normaly handle multiuse and auto update modules storage to right one without *sticks and rocks* making to work it for one case

:cl: BlackCrystalic
code: Work around with tank/apc, vehicle asrs
/:cl:


Nanu308 (commented):
Mapping lgtm

PR #6585, opened 2024-06-28
Queen Maturity Timer

blackcrystall:
# About the pull request

I don't remember very good, but in game still exist forever at 1m timer of queen maturity, now it's viewing seconds too and stop showing it'self when you reach mature stage on queen

# Changelog

:cl: BlackCrystalic
fix: Queen maturity timer
/:cl:


code/modules/mob/living/carbon/xenomorph/castes/Queen.dm
```diff
@@ -545,9 +545,8 @@
 
 	. += "Pooled Larvae: [stored_larvae]"
 	. += "Leaders: [xeno_leader_num] / [hive?.queen_leader_limit]"
-	if(queen_age_timer_id != TIMER_ID_NULL)
-		var/time_left = time2text(timeleft(queen_age_timer_id) + 1 MINUTES, "mm") // We add a minute so that it basically ceilings the value.
-		. += "Maturity: [time_left == 1? "[time_left] minute" : "[time_left] minutes"] remaining"
+	if(!queen_aged && queen_age_timer_id != TIMER_ID_NULL) // Idk, bug or not, but when timer ends, it leave just "2", this is soft solution to polution
```
Drulikar, line 548: You could add this to `/mob/living/carbon/xenomorph/queen/proc/make_combat_effective()`

```js
if(queen_age_timer_id != TIMER_ID_NULL)
		deltimer(queen_age_timer_id)
		queen_age_timer_id = TIMER_ID_NULL
```

Would also mean any manual call of that proc would resolve the maturity timer too.

Drulikar (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/Queen.dm
Drulikar: This is now error prone where timeleft can now potentially be returning null to pass into time2text.

Use `queen_age_timer_id != TIMER_ID_NULL` or `!queen_aged && queen_age_timer_id != TIMER_ID_NULL)` I don't really care; but I highly recommend what I suggested.

Drulikar (approved):

Drulikar:
Btw you only need to merge with upstream master when there are changes upstream that even matter to your pr. Also trying to removal staleness just deprioritizes the PR from trackers for me since it indicates the PR is still in flux so might not actually be ready yet.

PR #6600, opened 2024-07-01
Dynamic Sound

Git-Nivrak:
# About the pull request
Makes sounds react to your and its source moving around and become louder\quieter and change direction

Splits sounds into sounds the originate from a location and sounds that originate from a source - sounds that originate from a location will stay put at the location while sounds from a source will move around with the source (For example a Queen screech is a source and being slammed into a wall is a location). Due to this some things will probably not sound right because they should be a source while being a location or vice versa and this will have to be tweaked along when\if this is tested.

You still can't hear sounds if you were completely out of range when they started.

It's really cool if it doesn't lag the game to hell and back
<details>

</details>

# Changelog
:cl:
add: Sounds will now change as you move around their source and become quieter \ louder and change direction
code: Added support for sounds having a movable source, simply put the atom as the source
/:cl:

### DIFF
--- a/code/datums/soundOutput.dm
+++ b/code/datums/soundOutput.dm
@@ -24,7 +24,26 @@
 	owner = null
 	return ..()
 
-/datum/soundOutput/proc/process_sound(datum/sound_template/T)
+#define SMOOTHING 1 // How much smoothing when moving between tiles, smaller is more smoothing
+
+/datum/soundOutput/proc/update_sounds(atom/user, direction)
+	SIGNAL_HANDLER
+	for(var/channel in current_sounds)
+		for(var/i in 0 to round(32/SMOOTHING))
+			i = i * SMOOTHING
+			switch(direction)
+				if(1)
+					process_sound(current_sounds[channel], TRUE, 0, -1+i/32)
+				if(2)
+					process_sound(current_sounds[channel], TRUE, 0, 1-i/32)
+				if(4)
+					process_sound(current_sounds[channel], TRUE, -1+i/32)
+				if(8)
+					process_sound(current_sounds[channel], TRUE, 1-i/32, 0)
+/datum/soundOutput/proc/remove_sound(channel)
+	current_sounds -= channel
+
+/datum/soundOutput/proc/process_sound(datum/sound_template/T, update=FALSE, offset_x = 0, offset_y = 0)
 	var/sound/S = sound(T.file, T.wait, T.repeat)
 	S.volume = owner.volume_preferences[T.volume_cat] * T.volume
 	if(T.channel == 0)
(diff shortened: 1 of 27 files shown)


code/modules/movement/movement.dm
```diff
@@ -92,6 +92,10 @@
 
 /atom/movable/proc/Moved(atom/oldloc, direction, Forced = FALSE)
 	SEND_SIGNAL(src, COMSIG_MOVABLE_MOVED, oldloc, direction, Forced)
+	if(ismob(src))
+		var/mob/moved = src
+		if(moved.client)
+			SEND_SIGNAL(moved.client, COMSIG_MOVABLE_MOVED, direction)
```
Zonespace27, line 98: this should be a new signal type instead of reusing COMSIG_MOVABLE_MOVED

Git-Nivrak:
OD linter should be fixed as soon as it is updated (https://github.com/OpenDreamProject/OpenDream/pull/1870)

Doubleumc:
I was looking in to a system similar to this. Some observations that may be helpful:
- When you send a `SOUND_UPDATE` to a sound channel, it applies (almost) ALL of its variables to the playing sound, such as `volume` which defaults to 100. It is important that the updating sound have the same settings as the playing sound to maintain volume/pitch/etc consistency. Your current system handles this, just noting that it is important that it keeps doing it.
- If performance is a concern the vast majority of sounds don't need the extra handling, they're short enough or soft enough that people won't notice any positional offset. For example, the M41 firing sound is less than a second and people probably won't notice it being off by less than a second's worth of movement. There's no good way to test this in code (getting a sound's duration is very inconvenient) so it would be up to the `playsound` caller to indicate whether it needs positional updates or not.

cm13-github:

cm13-github:

code/datums/soundOutput.dm
```diff
@@ -33,29 +72,54 @@
...
+				if(candidate.z != T.z)
+					return // Invalid location
+				S.falloff /= 2
+				owner_turf = candidate
+		S.x = source_turf.x - owner_turf.x + offset_x
+		S.y = 0
+		S.z = source_turf.y - owner_turf.y + offset_y
+	S.y += T.y_s_offset
+	S.x += T.x_s_offset
+	S.echo = SOUND_ECHO_REVERB_ON
```
Doubleumc, line 106: Reverb should only be enabled for a sound that exists "in-world", i.e. is a positional sound and not an interface sound like an admin ping.

Previously the entire block from `var/turf/owner_turf ...` to `S.echo ...` was wrapped in `if(T.x && T.y && T.z)` as the test for being positional, in your case something like `if(source_turf.x && source_turf.y && source_turf.z)` may be more appropriate.

code/datums/soundOutput.dm
```diff
@@ -167,18 +232,9 @@
...
 
-/datum/soundOutput/proc/on_mob_logout(datum/source)
-	SIGNAL_HANDLER //COMSIG_MOB_LOGOUT
-	UnregisterSignal(owner.mob, list(COMSIG_MOVABLE_MOVED, COMSIG_MOB_LOGOUT))
-
-/datum/soundOutput/proc/on_client_mob_logged_in(datum/source, mob/new_mob)
-	SIGNAL_HANDLER //COMSIG_CLIENT_MOB_LOGGED_IN
-	RegisterSignal(owner.mob, COMSIG_MOVABLE_MOVED, PROC_REF(on_mob_moved))
-	RegisterSignal(owner.mob, COMSIG_MOB_LOGOUT, PROC_REF(on_mob_logout))
-	update_mob_environment_override()
```
Doubleumc, line 180: Still need the `update_mob_environment_override()` part of this. When a mob's environment override changes (currently just ear-ringing when in crit) it informs the soundOutput of the change. When the client's *mob* changes, however, soundOutput doesn't know unless we tell it. So for instance someone in crit that gets ghosted will have ear-ringing unless it knows to update, such as from `COMSIG_CLIENT_MOB_LOGGED_IN`.

code/datums/soundOutput.dm
```diff
@@ -33,8 +71,26 @@
...
-	S.status = T.status
-	if(T.x && T.y && T.z)
+
+	if(!update)
+		S.params = list("on-end" = ".soundend [S.channel]")
+		S.status = T.status
+	else
+		S.status = SOUND_UPDATE 		
+	
+	var/positonal_sound = FALSE
```
Doubleumc, line 81:
```suggestion
	var/positional_sound = FALSE
```

private-tristan:
Current TM is borked and everything sounds like it's happening on your tile/adjacent

private-tristan:
Sound desyncs during lag.

https://github.com/user-attachments/assets/a7d3083d-69bb-4388-9988-4e2b8c378542

code/game/sound.dm
```diff
@@ -36,7 +38,7 @@
 //status: the regular 4 sound flags
 //falloff: max range till sound volume starts dropping as distance increases
 
-/proc/playsound(atom/source, soundin, vol = 100, vary = FALSE, sound_range, vol_cat = VOLUME_SFX, channel = 0, status , falloff = 1, y_s_offset,x_s_offset)
+/proc/playsound(atom/source, soundin, vol = 100, vary = FALSE, sound_range, vol_cat = VOLUME_SFX, channel = 0, status , falloff = 1, y_s_offset,x_s_offset, smoothing=TRUE)
```
Doubleumc, line 41:
```suggestion
/proc/playsound(atom/source, soundin, vol = 100, vary = FALSE, sound_range, vol_cat = VOLUME_SFX, channel = 0, status , falloff = 1, y_s_offset,x_s_offset, use_smoothing=TRUE)
```
You named the argument `smoothing` but are trying to call it with `use_smoothing`

Git-Nivrak:
The server can't take it
Perhaps if byond had an easy way to get sound file's length but until then gonna close it

PR #6628, opened 2024-07-06
Warrior delimbing rework

Catzonwheelz:
# About the pull request

Tweaks how warrior delimbing work, by removing the ability to target the head. As well decreasing the time it takes to delimb.

I find the ability to decap to be overall a unnecessary ability, its as it stands can only be used in cases where the Marines are easily capped. well yes it can be useful in edge cases, i think it is more detrimental to the gameplay overall as it stands we have xenos killing easily capped marines too much as it is. 

also It not really good for the gameplay loop as it, is taking away respawns for both the player who got decapped and the hive as a whole.

By removing the targeting of the head it will be a small step in correcting this poor behavior. 

Now i have also speed up the time it takes to delimb, this is to give the ability use in incapacitating Marines for the edge cases. you will now be able to delimb many times faster. This can be useful if used right for capping as well as taking away hands can make it so the marine will not be able to cut there way out of the xeno, even if they have a Magnetic Harness. this can make it much safer to cap but of course doing so risks doing to much damage to the marine risking killing the cap.

<details>

</details>

# Changelog
:cl:
balance:  Changed the time to delimb 25, 35
balance: took out Head targeting
/:cl:


joookschad:
removal for the sake of i died to this

Catzonwheelz:
Do apologies if mucked anything up i am a new to GitHub and coding in general.

Mister-moon1:
There will always be people killing caps just as a heads up

You state how it's not good because it removes spawns from both sides (???)
Would you say that dying in backlines should be removed because it permas marines, or should xenos be allowed to be revived?
The point of the CMs method of working is to commit as much permanent damage to the other side and a considerable amount of balancing goes into it
There is a reason why warrior's decap has remained, and that is because it is a feature where the marine in question needs to be in a situation of having multiple xenomorphs surrounding them considering warrior doesnt have a confirmed disarm chain, mainly reserved for where it is too risky to commit the capture, there is no suitable route to hive, it is more convenient or if the warrior along with the xenomorphs with him are required in the area in question

Catzonwheelz:
Well no the issue with Decaping is its take away spawns in cases where it is unnecessary. The vary act of getting the situation aligned with what is need to get a decap off is the same kind of situation need to get a cap.

Git-Nivrak:
I died pls nerf

TheKillfish:
Not against the limb ripping being faster, but preventing decaps is really unnecessary. It's niche enough as is, and if it happens you more than likely would have your time as marine hard end soon anyway. While it might be more an issue if the rip time is faster, just don't speed up head ripping.

Catzonwheelz:
Well for the increase of speed for Delimbing was to balance out taking decaping for few cases that decaping is sensible. after all taking off all of a marines limbs is quite powerful. also sadly i wanted to do more with this PR at first like making so warriors can causes heartbreaks and lowering the body damage took from delimbing. but unfortunately i don't have the first clue how to get it to work how i want. keep in mind this is my first PR. 

was planning on come back to it after i learn how to work with the code base better.

Steelpoint:
How quickly is this making it take for a Warrior to rip off a Marines limb? I can see its a rand (10, 20), Is this saying it could take 10 to 20 seconds to rip off a limb? Cause if true this could just lead to situations of Warriors ripping off a lot of Marines arms and legs before killing them. Which I can guarantee is going to generate far more discord than the rarer chances of removing a marines head (or removing a capped marines limbs).

Catzonwheelz:
it will most likely need adjustments as needed yes. but the thing is the PR does change the interrupt effect, as such you will still need fellow xenos helping or for them to be in pain crit. but to do the damage taked from delimbing there is only so many limbs you can take off be for they die.

Catzonwheelz:
i will do a bit of testing later to see how 20 30 does.

Catzonwheelz:
after doing some testing the warrior can get 1 limb off and a frac for marines in crit.

Red-byte3D:
this is a fucking useless nerf, youre doing this because you got owned.

when you can decap people, you have time to capture them. stop being an idiot

Drulikar:
This doesn't need to change. It requires teamwork and time to pull off. Nor am I keen on other limbs getting taken off faster in compensation. The xenos would benefit more from capturing them than decapping them too if possible, but sometimes situations where the next best would be to perma.

PR #6497, opened 2024-06-17
Old explosion rooted out, work in progress

cuberound:
# About the pull request

removing old bad code is good, only balance part is mortar shells and exploding magazines and ammo boxes (test the magazines they are changed the most I think and ammo boxes) rest is unused but plase go thrue it as I might have done some mistake. the numbers can be adjusted, especily the mortar

# Explain why it's good for the game
# Testing Photographs and Procedure
<details>
direct incen shell should do less harm then HE shell and old bad code removing is good

</details>

# Changelog
:cl:
balance: HE mortar shell is 240 power with 34 falloff incen 200 power with 33 falloff
fix: all explosions use "new" explosion code directly
/:cl:

PR #6638, opened 2024-07-06
Cooked mortar shell explosion is now fixed

Git-Nivrak:
# About the pull request
Fixed as in static, doesn't change for different mortar shell types
Makes the explosion when setting mortar shells on fire fixed and small

This was supposed to be a niche feature like ammo boxes exploding and instead it is being used to blow up inside pylons \ caves or even on shipside. This is why we can't have nice things.

<details>

</details>

# Changelog

:cl:
del: Mortar shells no longer blow up their payload, instead they will create a small fixed explosion.
/:cl:

### DIFF
--- a/code/modules/cm_marines/equipment/mortar/mortar_shells.dm
+++ b/code/modules/cm_marines/equipment/mortar/mortar_shells.dm
@@ -171,9 +171,9 @@
 		return
 	burning = TRUE
 	cause_data = create_cause_data("Burning Mortar Shell", flame_cause_data.resolve_mob(), src)
-	handle_fire()
+	handle_fire(cause_data)
 
-/obj/item/mortar_shell/proc/handle_fire()
+/obj/item/mortar_shell/proc/handle_fire(cause_data)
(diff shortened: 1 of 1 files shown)


ihatethisengine:
Maybe just limit the throw range to like 4 (you can also do this just to OT shells, or even nerf OT shells even further) and just make the explosions weaker, like 0.75 strength of what you would get from launching it via mortar

cuberound:
do NOT add old explosion code it does not behave like you would expect, I think this just spawns cell_explosion 120 power with fallof 40

Pyrodes:
Removing intuitive gameplay one feature removal at a time

Drulikar (changes requested):
Please check whether there even is ordinance that can explode (namely customs). Up to you if you want a different kind of handling for liquids (napalm). Also please use the other explosion proc since https://github.com/cmss13-devs/cmss13/pull/6497 intends to refactor out what you're using.

Drulikar (approved):

PR #6652, opened 2024-07-07
autoclimbable handrails on harm intent

cuberound:
# About the pull request

you do climb when you collide with handrail, rejoice.  this makes handrails lot less of an obsticle for both huge xenos and highly rained and agile marines. also there is autoclimbable var that can be set to FALSE on req railing so you can not throw others out of the line

handrails can be major pain in ass for both marines and xenos, especily as they are used near ledges so you auto climb from one side to other but have to manualy climb back. they are also sometimes hard to spot and can get you stuck and killed, generly getting stoped by railing that you did not see feels like shit as both sides can shoot or slash it down in seconds. I am gona put this as QOL but if you think it is balance then so be it

<details>

</details>

# Changelog
:cl:
qol: handrails are autoclimbable when on harm intent, you take some harm when doing so
balance: crusher and ravager destroy handrails when trying to pass them similar to tables
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade/handrail.dm
+++ b/code/game/objects/structures/barricade/handrail.dm
@@ -18,7 +18,18 @@
 	var/build_state = BARRICADE_BSTATE_SECURED
 	var/reinforced = FALSE //Reinforced to be a cade or not
 	var/can_be_reinforced = TRUE //can we even reinforce this handrail or not?
+	///Whether a ground z-level handrail allows auto-climbing on harm intent
 	var/autoclimb = TRUE
+
+/obj/structure/barricade/handrail/Initialize(mapload, ...)
+	. = ..()
+	if(!is_ground_level(z))
+		if(autoclimb && is_mainship_level(z))
+			RegisterSignal(SSdcs, COMSIG_GLOB_HIJACK_LANDED, PROC_REF(reset_autoclimb))
+		autoclimb = FALSE
+
+/obj/structure/barricade/handrail/proc/reset_autoclimb()
+	autoclimb = initial(autoclimb)
 
 /obj/structure/barricade/handrail/update_icon()
 	overlays.Cut()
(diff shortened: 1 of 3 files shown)


code/game/objects/structures/barricade/handrail.dm
```diff
@@ -40,6 +41,22 @@
...
+		if(!xenomorph.stat) //No dead xenos jumpin on the bed~
+			visible_message(SPAN_DANGER("[xenomorph] plows straight through [src]!"))
+			deconstruct(FALSE)
+	else
+		if(ismob(movable) && autoclimb)
+			var/mob/living/climber = movable
+			if(climber.a_intent == INTENT_HARM)
+				var/climbed = do_climb(climber)
+				if(climbed)
+					climber.apply_damage(15,BRUTE)
```
Drulikar, line 56: This is too much damage, probably doesn't need to always happen (prob 25 or something), should ensure `no_limb_loss` is set, and needs to induce movement delay (really goofy how fast you just vault any rails in sequence).

Req is going to get really bloody with this PR. I highly recommend you consider downloading strongdmm to edit req's railing to not have this behavior either with a var edit or a new subtype.

the worst vaulter

code/game/objects/structures/barricade/handrail.dm
```diff
@@ -40,6 +41,22 @@
 		if(E.icon_path && E.obj_icon_state_path)
 			overlays += image(E.icon_path, icon_state = E.obj_icon_state_path)
 
+/obj/structure/barricade/handrail/Collided(atom/movable/movable)
+	if(istype(movable,/mob/living/carbon/xenomorph/ravager) || istype(movable,/mob/living/carbon/xenomorph/crusher))
+		var/mob/living/carbon/xenomorph/xenomorph = movable
+		if(!xenomorph.stat) //No dead xenos jumpin on the bed~
```
Drulikar, line 47:
```suggestion
		if(xenomorph.stat == CONSCIOUS) //No dead xenos jumpin on the bed~
```

code/game/objects/structures/barricade/handrail.dm
```diff
@@ -40,6 +41,22 @@
...
+			deconstruct(FALSE)
+	else
+		if(ismob(movable) && autoclimb)
+			var/mob/living/climber = movable
+			if(climber.a_intent == INTENT_HARM)
+				var/climbed = do_climb(climber)
+				if(climbed)
+					climber.apply_damage(15,BRUTE)
+					climber.visible_message(SPAN_WARNING("You hit yourself as you vault over the [src]"))
+	..()
```
Drulikar, line 58: A normal crusher using the charge ability will double deconstruct this handrail. Consider just returning after the earlier `deconstruct(FALSE)`

Drulikar (changes requested):

code/game/objects/structures/barricade/handrail.dm
```diff
@@ -40,6 +41,27 @@
...
+	else
+		if(ismob(movable) && autoclimb)
+			var/mob/living/climber = movable
+			if(climber.a_intent == INTENT_HARM)
+				var/climbed = do_climb(climber)
+				if(climbed)
+					climber.client.move_delay += 3 DECISECONDS
+					if(prob(25))
+						if(ishuman(climber))
+							climber.apply_damage(15,BRUTE,no_limb_loss = TRUE)
```
Drulikar, line 59: See https://github.com/cmss13-devs/cmss13/actions/runs/10111981060/job/27965048466?pr=6652#step:8:20

Basically either you need to type `climber` as a `/mob/living/carbon/human` or you add the missing arguments to `/mob/living/proc/apply_damage`

Drulikar (changes requested):

code/game/objects/structures/barricade/handrail.dm
Drulikar: Application of this suggestion was reverted so is still outstanding.

code/game/objects/structures/barricade/handrail.dm
```diff
@@ -40,6 +41,27 @@
...
+		if(!xenomorph.stat)
+			visible_message(SPAN_DANGER("[xenomorph] plows straight through [src]!"))
+			deconstruct(FALSE)
+	else
+		if(ismob(movable) && autoclimb)
+			var/mob/living/climber = movable
+			if(climber.a_intent == INTENT_HARM)
+				var/climbed = do_climb(climber)
+				if(climbed)
+					climber.client.move_delay += 3 DECISECONDS
```
Drulikar, line 56:
```suggestion
					climber.client?.move_delay += 3 DECISECONDS
```
Mobs won't always have clients.

Drulikar (changes requested):

code/game/objects/structures/barricade/handrail.dm
Drulikar: 15 damage is still too much even with 25 prob. Try something more like 5.

code/game/objects/structures/barricade/handrail.dm
```diff
@@ -40,6 +41,29 @@
...
+			visible_message(SPAN_DANGER("[xenomorph] plows straight through [src]!"))
+			deconstruct(FALSE)
+			return
+	else
+		if(ismob(movable) && autoclimb)
+			var/mob/living/climber = movable
+			if(climber.a_intent == INTENT_HARM)
+				var/climbed = do_climb(climber)
+				if(climbed)
+					climber.client?.move_delay += 3 DECISECONDS
```
Drulikar, line 57: This needs to be moved before the `do_climb` else it will never take effect until after all the do_afters are complete (what we're trying to delay)

Drulikar (changes requested):

Git-Nivrak:
This makes it so you auto climb even when pushed which shouldn't be a thing

Drulikar (changes requested):
- if someone else can push you over a railing, the collide should just behave as normal preventing you from moving that direction
- if you don't want to make railing exceptions via mapping edits, consider instead making a z level check for the Almayer to prevent the auto vaulting of railings on that z level (or maybe only if it's hijack?)

harryob:
pr seems to be stale, let us know if you want it re-opening, or just make a fresh pr for it

Drulikar (approved):

PR #6664, opened 2024-07-08
Change runechat font from Small fonts to MS Serif

Blundir:
# About the pull request

Swaps fonts for runechat

Before:
dreamseeker_1HqQxjyTtY

dreamseeker_SyCSzBL0hJ

dreamseeker_F5uOv0blMR

After:

dreamseeker_uwE0JWu48m

dreamseeker_jftdTaUyNm

dreamseeker_dUQTn3nfCb

More visually pleasing font, that is also less big, which would support longer messages, as well as this font supporting wider range of letters.

# Changelog
:cl:
ui: changed runechat font from Small fonts to MS Serif
/:cl:

### DIFF
--- a/interface/skin.dmf
+++ b/interface/skin.dmf
@@ -215,7 +215,7 @@ window "mapwindow"
 		font-size = 7
 		is-default = true
 		saved-params = "zoom;letterbox;zoom-mode"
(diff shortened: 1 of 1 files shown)


Drulikar:
Its more varied at the cost of less readability; I think we're fine with the existing fonts for default usage.

If you really want some of these, make it where its a markup option or tied to knowing some language to act as an accent or something. See existing values: https://cm-ss13.com/wiki/Chat_Markup

PR #6671, opened 2024-07-10
Successful AA interceptions now destroy the hijacked shuttle and end the round

blackdragonTOW:
# About the pull request

Many people love to complain about hijack and how much they don't like it.  What if we just removed 1/6th of hijacks?

This makes it so that if the hijack area lines up with the AA protection area, the weapon system does its job rather than politely asking the ship to turn around.  The shuttle is deleted and the round ends with a Xeno Minor.

An added benefit to this might make hijack location selection something other than defaulting to upper aft to pre-cog the evac attempt.  And might even turn into some 4D chess with AA location setting and hijack selection trying to out maneuver what the other placed.  Otherwise statistically this is a 1 in 6 chance for an intercept, assuming marines actually set AA (my god so many marines forget to set AA)

<details>
A silly amount of hijackings to make sure the correct messages and timings are used.  Really kinda learned how difficult it is to iterate at a reasonable speed.

</details>

# Changelog
:cl:
balance: AA intercepts during hijack now destroy the shuttle and end the round
/:cl:


ihatethisengine:
I mean, unironically adding a better effect of successful AA is a neat idea. But round end and complete wipe of a side just over a roll of a dice is way too funny. At least maybe send the shuttle back to the ground so it's not that sad for xenos.

hislittlecuzingames:
Finally some strategy beyond "Set AA for lifeboats."

blackdragonTOW:
> Finally some strategy beyond "Set AA for lifeboats."

Exactly.  And if the meta shifts do hitting lower instead, that then becomes a meta chase and there's actual strategy involved for placing AA in a rock paper scissors sense of strategy.  Dare I say that putting thought into what button you're pressing is a good thing rather than slamming the same thing over and over.

code/modules/shuttle/dropship_hijack.dm
```diff
@@ -126,9 +127,14 @@
 					to_chat(M, SPAN_DANGER("You feel the ship turning sharply as it adjusts its course!"))
 					shake_camera(M, 60, 2)
 			playsound_area(internal_area, 'sound/effects/antiair_explosions.ogg')
+		addtimer(CALLBACK(src, PROC_REF(aa_interception), shuttle), 5 SECONDS)
 
 	hijacked_bypass_aa = TRUE
 
+/datum/dropship_hijack/almayer/proc/aa_interception(/obj/docking_port/mobile/shuttle)
+	shuttle.intoTheSunset()
```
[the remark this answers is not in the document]
zzzmike: Running a check to see if there's a living queen before deciding on xeno minor or marine major would be good. Sometimes rounds have a queen bumrushing the dropship as a last resort, getting the launch off, but all xenos dying before the crash. It doesn't result in a xeno minor in that case, and shouldn't do so regardless of the new AA.

code/modules/shuttle/dropship_hijack.dm
```diff
@@ -126,9 +127,14 @@
...
+	SSticker.mode.round_finished = MODE_INFESTATION_X_MINOR
```
zzzmike, line 136:
```suggestion

```
(if my other changes go through this would also need to be taken out)

code/modules/shuttle/dropship_hijack.dm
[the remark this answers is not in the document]
blackdragonTOW: Hmm, I see your point.  I'll wait until a maint says they feel good about it as I don't want to prematurely fire a M_MAJOR, but I'm not wholly unopposed to something that slaps people who try to cockpit dive last second.

code/modules/shuttle/dropship_hijack.dm
```diff
@@ -116,7 +116,8 @@
 		if(!offset_target)
 			offset_target = target // Welp the offsetting failed so...
 		crash_site.forceMove(offset_target)
-		marine_announcement("A hostile aircraft on course for the [target_ship_section] has been successfully deterred.", "IX-50 MGAD System", logging = ARES_LOG_SECURITY)
+		marine_announcement("A hostile aircraft on course for the [target_ship_section] has been successfully engaged and destroyed.", "IX-50 MGAD System", logging = ARES_LOG_SECURITY)
+		xeno_announcement(SPAN_XENOANNOUNCE("The Almayer's AA weapons are spinning up! OH NO! LOOK OUT!"), "everything", XENO_HIJACK_ANNOUNCE)
```
Zonespace27, line 120: This isn't a very well worded xeno announcement

Zonespace27 (changes requested):

Steelpoint:
If the idea of killing every Xenomorph is too excessive, you could rework it that the Almayer's AA simply shoots down the Dropship and causes it to land/crach on the planet, while also ending the round.

blackdragonTOW:
> If the idea of killing every Xenomorph is too excessive, you could rework it that the Almayer's AA simply shoots down the Dropship and causes it to land/crash on the planet, while also ending the round.

What does this mechanically accomplish?

Steelpoint:
> What does this mechanically accomplish?

As I wrote it accomplishes the same end-state this PR envisions, but it allows the xenomorphs to react to the events that have transpired from an in-character perspective instead of deceased ghosts. It may also make this outcome more palatable to the xenomorph players as whilst the outcome is the same they at least they may not feel as negatively impacted as they were not killed.

Drulikar:
IMO correctly selecting an AA location is too brainless of a thing for marines to turn a losing situation around. And is also pretty lame for those who would be preparing for a fight to miss out on it from both sides. For example, xenos managing to grow a King successfully only for marines to turn tail, attempt to hijack, and just blow up without any action.

Additionally, this is far too prone to metacommunication from both sides. Either:
A) Queen can be informed where the AA location is set to.
B) Marines can be informed where to set the AA location to.
If you lock it out so it can only be set once, A then becomes likely. If you allow it to be freely changed, B is more likely. If you lock it out only after hijack, A is more likely. I'm not sure what to suggest here that would ever resolve this problem outside just removing the selection entirely and just making it RNG entirely which makes the player agency issue that much more of an issue - xenos just sometimes get robbed of a successful hijack by chance.

I'm not totally against the idea but AA up to this point has only been a means to ensure the location you intend to hold doesn't get obliterated. Maybe instead successful AA could actually shoot at the dropship, causing an explosion in the traveling shuttle spacing/gibbing those unlucky where it happens to land, but otherwise still crash lands in some other location in the ship.

Until at the very least the metacommunication issue is resolved, this is DNM.

blackdragonTOW:
I've updated the PR with what I feel is an elegant solution. 

Firstly, allow me to remind our players that meta-gaming is a ban-able offense and can result in you being removed from the community!

This solution follows good security on how best to go about keeping any sort of meta to a minimum.   Firstly, AA cannot be set until hijack has started.  This means two things: 1: Only the Queen knows where the DS is going.  and 2: Marines have ~60s until we check to see if AA lines up with the hijack target.  (As a note: setting the AA to even the correct location after that period is up still results in a hijack impact, so late attempts will not work).  

Given how there is no hive based messaging for where hijack is pointed after the Queen selects it, that means the Queen is in charge of that information and keeps it secret for the narrow time frame that it is relevant.  That then means once 60-70s have passed, the Queen can share that info (or honestly, she can keep it to herself until the hijack zoom sound starts, because it's not like there's much prep to do while on the shuttle ride) and we play like nothing had changed.  

And should meta-gaming happen by the Queen, it's an easy find on their ckey as we have the hijack start logged and the Queen is literally the only xeno who can start hijack.

blackdragonTOW:
Also substantially more soul in setting the AA once there's an actual incoming air threat, as well as the officer rushing to the foredeck to get there in time.  

Really anything to keep people from slowly gravitating to upper-aft after FOB evac.

Steelpoint:
Changelog will need updating regarding the change to how setting AA is being changed. 

Perhaps Marines might need a IC message alerting them to set up the AA.

code/modules/shuttle/dropship_hijack.dm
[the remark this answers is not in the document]
Drulikar: Suggested change would just cause `shuttle.intoTheSunset()` to be called multiple times; was no point moving that into the for loop if it always occurs in every possibility. Regardless I don't think this necessary. If anything it'd be a marine minor for failing to do any good to save the colony, leaving it infested, but xenos having no means to get themselves off the planet for now.

Drulikar:
If you're certain you want to go forward with this, making a limited time to select where to put AA and it relying on the queen to keep that secret is workable as far as my DNM concerns once there's a bit more messaging indicating AA is available like Steelpoint was mentioning above.

I personally still think this is worse for the game however. I would much rather xenos taking hits from AA fire and them being in worse condition on crash land than just the entire hijack part of the game to abruptly end. Hijack doesn't have to be in their favor always; but I'd rather the sequence of events to not make players feel like they got robbed of gameplay they expected.

I will be considering reverting this PR or reworking what effect AA has when https://hackmd.io/@Drathek/BywFsOfH0 is implemented.

code/modules/shuttle/dropship_hijack.dm
Drulikar, line 136: This should be handled by the gamemode `check_win` proc. Consider setting maybe `SShijack.hijack_status` to a hijack_averted or something, and then `/datum/game_mode/colonialmarines/check_win()` can check that value and end the round.

As is, this means every game mode that shoots down the dropship successfully will abruptly end in distress signal xeno minor.

code/modules/shuttle/dropship_hijack.dm
```diff
@@ -126,9 +127,14 @@
 					to_chat(M, SPAN_DANGER("You feel the ship turning sharply as it adjusts its course!"))
```
Drulikar, line 127: Technically this message is unnecessary, but its okay to leave if you like.

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -297,6 +297,8 @@
 	dropship.crashing = TRUE
 	dropship.is_hijacked = TRUE
 
+	SShijack.hijack_status = HIJACK_OBJECTIVES_SHIP_INBOUND
```
Drulikar, line 300:
```suggestion
	if(SShijack.hijack_status == HIJACK_OBJECTIVES_NOT_STARTED)
		SShijack.hijack_status = HIJACK_OBJECTIVES_SHIP_INBOUND
```

Drulikar (changes requested):

Drulikar (changes requested):
I am fine with the change making AA only able to be set when there is a hostile dropship enroute. But merely performing `intoTheSunset` on the shuttle and setting xeno minor I am not comfortable with.

Please make it so the shuttle takes damage and potentially spaces any of the hive that happens to be where the AA fired onto the shuttle. I am not concerned about the hive being at full strength for a hijack, but I would like the hijack to play out, even if it by chance happened to space the entire hive.

Ultimately hijack should not be aborted, the shuttle should still crash, and it should be a spectacle while giving AA more meaning and realism.

code/modules/shuttle/dropship_hijack.dm
Drulikar: No longer relevant

code/modules/shuttle/dropship_hijack.dm
Drulikar: No longer relevant.

blackdragonTOW:
The intent of this PR was to skip a small number of hijacks, given that the end goal of that here would no longer be achieved, I think changing this to something else would deviate from the desired outcome of what I set out to do.  Rather than turn this into something it's not I'll close it out and consider this little experiment concluded. 

Also lmao 35 thumbs up, never seen it ever get that high on a PR before.

PR #6674, opened 2024-07-10
roofs added

cuberound:
# About the pull request

feels strange to add this only as hybrasia behavior, made it into proper structure, it still NEEDS some work. Give me feedback regarding how (and if) it should handle explosions, xeno attack, accid bullets and any other stuff you come up with ( I used catwalk as default sprite as I have no roof sprite, if someone has default roof sprite they could give, go ahead)

gives another tool to mappers, it is not exactly ment for covering huge areas with roofs (even tho it works too, might be a bit server heavy tho no idea how it would behave with dozens of mobs constantly moving under roof and being searched for on each mvoement) was made for hybrasia lattices and billboards

<details>

</details>

# Changelog
:cl:
add: roof structures added, can be also used for lattices and billboards, go transparent when you are near them
/:cl:

### DIFF
--- a/code/game/objects/structures/roof.dm
+++ b/code/game/objects/structures/roof.dm
@@ -51,7 +51,7 @@
 	SIGNAL_HANDLER
 	mob.client.images += normal_image
 
-/obj/structure/roof/Destroy()
+/obj/structure/roof/Destroy(force, ...)
 	linked_master.remove_roof(src)
 	for(var/icon in GLOB.player_list)
 		var/mob/mob = icon
@@ -91,7 +91,7 @@
 		for(var/obj/effect/roof_node/node in get_step(src,direction))
 			node.link_master(master)
 
-/obj/effect/roof_node/Destroy(force)
+/obj/effect/roof_node/Destroy(force, ...)
 	linked_master.connected_nodes -= src
 	. = ..()
 
@@ -156,10 +156,11 @@
 
 
 
-/datum/roof_master_node/Destroy()
+/datum/roof_master_node/Destroy(force, ...)
 	for(var/obj/effect/roof_node/roof_node in connected_nodes)
 		roof_node.Destroy()
 	for(var/obj/structure/roof/roof in location)
 		roof.Destroy()
 	. = ..()
+
 
--- a/code/game/objects/structures/roof.dm
+++ b/code/game/objects/structures/roof.dm
@@ -159,10 +159,12 @@
 
 
 /datum/roof_master_node/Destroy(force, ...)
-	for(var/obj/effect/roof_node/roof_node in connected_nodes)
-		roof_node.Destroy(1)
-	for(var/obj/structure/roof/roof in location)
-		roof.Destroy(1)
+	if(connected_nodes)
+		for(var/obj/effect/roof_node/roof_node in connected_nodes)
+			roof_node.Destroy(1)
+	if(connected_nodes)
+		for(var/obj/structure/roof/roof in connected_roof)
+			roof.Destroy(1)
 	. = ..()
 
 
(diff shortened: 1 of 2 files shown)


code/game/objects/structures/roof.dm
```diff
@@ -0,0 +1,164 @@
...
+/datum/roof_master_node/proc/connect(location)
+	for(var/obj/effect/roof_node/node in location)
+		node.link_master(src)
+	for(var/obj/structure/roof/roof in location)
+		roof.link_master(src)
+
+/datum/roof_master_node/proc/remove_roof(obj/structure/roof/roof) //roof tile got removed
+	connected_roof -= roof
+	if(!length(connected_roof))
+		Destroy(1)
```
Drulikar, line 164: Don't call Destroy directly; thats on SSgarbage to call. use qdel instead.

code/game/objects/structures/roof.dm
```diff
@@ -0,0 +1,164 @@
...
+	var/location
+
+/datum/roof_master_node/Destroy(force, ...)
+	if(connected_nodes)
+		for(var/obj/effect/roof_node/roof_node in connected_nodes)
+			roof_node.Destroy(1)
+	if(connected_nodes)
+		for(var/obj/structure/roof/roof in connected_roof)
+			roof.Destroy(1)
+	. = ..()
```
Drulikar, line 116:
```suggestion
	return ..()
```

code/game/objects/structures/roof.dm
```diff
@@ -0,0 +1,164 @@
...
+		return
+	if(isliving(mover))
+		var/mob/living/mob = mover
+		linked_master.add_under_roof(mob)
+
+/obj/effect/roof_node/Destroy(force, ...)
+	if(linked_master)
+		if(linked_master.connected_nodes)
+			linked_master.connected_nodes -= src
+	.=..()
```
Drulikar, line 91:
```suggestion
	return ..()
```

code/game/objects/structures/roof.dm
```diff
@@ -0,0 +1,164 @@
...
+	icon_state = null
+
+	RegisterSignal(SSdcs, COMSIG_GLOB_MOB_LOGGED_IN, PROC_REF(add_default_image))
+
+	for(var/icon in GLOB.player_list)
+		add_default_image(SSdcs, icon)
+	if(lazy_nodes) //creates new node on each surounding tile if there is not one already
+		var/obj/effect/roof_node/neighbor = locate() in loc
+		if(!neighbor)
+			neighbor = new(src.loc)
```
Drulikar, line 31:
```suggestion
			neighbor = new(loc)
```

code/game/objects/structures/roof.dm
```diff
@@ -0,0 +1,164 @@
...
+		for(var/direction in CARDINAL_ALL_DIRS)
+			var/adjacent_loc = get_step(src, direction)
+			neighbor = locate() in adjacent_loc
+			if(!neighbor)
+				neighbor = new(adjacent_loc)
+	return INITIALIZE_HINT_LATELOAD
+
+/obj/structure/roof/LateInitialize() //we use late init to allow for lazy nodes to spawn first on mapload
+	. = ..()
+	if(!linked_master)
```
Drulikar, line 41: Early return instead.

As in

```js
if(linked_master)
	return
```

Drulikar (changes requested):
There should also be a way to designate other kinds of objects that go transparent with this group too; e.g. any `/obj/structure/prop/hybrisa/billboardsandsigns` mounted on the lattices. As far as I can tell they will have the same loc as a lattice.

Drulikar (approved):
As far as I can tell everything is working

Issue #6312, opened 2024-05-19
a apc that cant be fixed


tzTZtzTZqwq:
### Testmerges

-

### Description of the bug

o fix the apc the wire needs to be removed
to remove the wire the floor need to be removed
but theres no floor
so the wire cant be removed
so the apc cant be fixed
![b3decc0248f2d65429e8ec2edafca889cc3de1ed](https://github.com/cmss13-devs/cmss13/assets/65105227/be33a49d-7097-47f9-9b8e-113191da6794)
![6ee068431df6f8598f8a1203f42143ad34bd4251](https://github.com/cmss13-devs/cmss13/assets/65105227/a7eac2ea-da88-4087-afdb-ca0040bd2ae0)

### What's the difference with what should have happened?

there should be a floor and wire on that tile so the wire can be removed

### How do we reproduce this bug?

the location of the comms is marked on the img

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

hislittlecuzingames:
Good first issue

Diegoflores31:
Mapping Issue Tag @Huffie56

hislittlecuzingames:
https://github.com/cmss13-devs/cmss13/pull/6499

PR #6686, opened 2024-07-11
Re: Add Directional APC's sprites, reduces lines of code by adding and using Updatepaths.txt file.

Venuska1117:
# About the pull request

This PR add directional APC's and Updatepath.txt

standardize APC's by creating 2 separate types, normal and upgraded variant, replace APC's with cell types: high/super/hyper with upgraded variant with own HIGH cell type. (10000 capacity)

Edit: Add directional APC sptires made by Blundir.

Fixes: #6792
Fixes: #6312

Reduces lines of codes in mapping, i know there is already code in apc.dm to make them have pixel_y or pixel_x offset.

Mainly issue is: in map editor they have var edited offsets, with mean it not only override code that already exist, but also occupy more space, with new code it should do same in exchange for less lines of code. (i didnt remove directional code for other stuff like broken APC's etc.)

Edit: Add directional APC's sprites for better visuals

New directional APC's sprites by @Blundir.
image

Image from Blundir:
image

# Changelog
:cl: Venuska1117, Blundir
imageadd: add directional APC's Sprites.
balance: Now there are only 3 APC's types, APC's with call_type high/super/hyper are replaced with upgraded APC variant containing high capacity cell, weak variant got replaced with normal APC variant.
fix: Fixes APC's with wrong offsets (strata and Prison)
/:cl:


blackdragonTOW:
Hello! Please remember to attach screenshots of your altered sprites in the PR description!!

Nanu308 (changes requested):
From looking at the map diff bot, you seem to have removed some APC's entirely. Were these apc's doubled in an area?

Venuska1117:
> From looking at the map diff bot, you seem to have removed some APC's entirely. Were these apc's doubled in an area?

yeah, the new checking system did not allow me to commit, beacause there were duplicated APC's, nanomeds etc.

nauticall (approved):
sproites

Drulikar:
Wires are still incorrect. E.g. 
image

And no, this is not the case on master:
image

The reasoning is because in the DMI you reversed all directions for the apcs and wires. So if you are certain thats how you want them in the dmi, you must ensure every apc accounts for the change.

tools/UpdatePaths/Scripts/6686-apc-replacement.txt
Drulikar, line 1: Why did you opt to eliminate all start_charge values?

Drulikar (changes requested):

tools/UpdatePaths/Scripts/6686-apc-replacement.txt
Venuska1117: by default apc.dm have start_charge 90 (with mean 90%)
image

so i don't need to edit them, only if i need to add apc's with 0% of power (like no_power ones)

code/modules/power/apc.dm
Drulikar, line 1: Is there any reason to keep `directional` as part of the path? You altered the behavior of `/obj/structure/machinery/power/apc` anyways so not like there's any unique behavior for the directional subtypes.

Also `/obj/structure/machinery/power/apc/nocharge` has no use now so should just be removed.

maps/map_files/LV522_Chances_Claim/LV522_Chances_Claim.dmm
Drulikar, line 1: Chances claim used to use `/obj/structure/machinery/power/apc/weak` which has a capacity of 1000. They are now using apcs with cells with 5000 capacity. Since there are 56 APCs that have a 90% charge here, consider either keeping the `weak` variant, or make a variant or var edit all the chances claims apcs that were weak to have 1/5 the starting charge.

Drulikar (changes requested):

maps/map_files/LV522_Chances_Claim/LV522_Chances_Claim.dmm
Venuska1117: i changed in .txt file, now they are replaced with start_charge = 20 (20% from 5000) so its back to 1000.

code/modules/power/apc.dm
Venuska1117: the `directional` part of path is for people who will intend to map in future, most of that have edit of pix_x and pix_y (depending with direction you want them) to make more easy to understand as mapper without need to check code every time you want to do some inserts/mapping. _(they are visible in mapping to be moved in direction, so giving them `directional` in path name make it more readable)_

maps/map_files/LV522_Chances_Claim/LV522_Chances_Claim.dmm
Venuska1117: if hyper ones for CORSAT need be start_charge = 300 (so it go back to 30000 if used on apc/upgraded)

Drulikar: CORSAT doesn't need to change; as I said before I think it was more a lore reason for them to have higher tech stuff. My feedback here was on Chances Claim's weakones since basically every APC is powered now and powered with significantly more power than they used to.

code/modules/power/apc.dm
Drulikar: Directional in the path suggests that there's *non* directional apcs which is not that case. I would rather the unnecessary path removed. APCs have always had a concept of needing to face a certain direction; they'd just pixel shift before and now they actually have a sprite.

maps/map_files/Sorokyne_Strata/Sorokyne_Strata.dmm
Drulikar, line 1: image
image

I believe this is fixed in #6071 but in the interim can you fix this one and maybe consider adding that non-dirt turf for the other tcoms?

maps/map_files/FOP_v2_Cellblocks/Prison_Station_FOP.dmm
Drulikar, line 1: Near showers in the north high security checkpoint there is an east facing APC that should be south facing or moved:
image
image

Drulikar, line 1: The Medium-Security Cellblock East APC near the ship is south facing and floating. It should be moved somewhere else.
image
image

Drulikar (changes requested):
Otherwise everything else looks fine

maps/map_files/Sorokyne_Strata/Sorokyne_Strata.dmm
Venuska1117: yes it was fixed in #6071 i tough it was something on my side (but with this i am sure its not that) as for non-dirt turf, i will see what i can do.

Drulikar: For the non-dirt turf you just need to do the same as you did in 6071. I was just saying for both of those fixes to also be in this pr

Drulikar: Specifically this fix: 
IMG_1426

Drulikar (approved):

PR #6692, opened 2024-07-12
Use RANGE_TURFS

Doubleumc:
# About the pull request

Switches several uses of `for(var/turf/... in range())` over to `RANGE_TURFS()` instead. Doesn't switch when the expected radius is less than 5. Where possible, removes the loop entirely.

For finding turfs, `RANGE_TURFS` shows a slight performance benefit over `range`. This performance benefit becomes more pronounced as the radius increases.
<details>

Performance testing. All procs in question return a list of turfs.
The procs of interest:
image

Radius 1:
image

Radius 5:
image

Radius 10:
image

Radius 20:
image

Results show that when able to convert away from a loop (`turftest_range_turf_direct`) that is always the best choice. For results under a radius of 5 the standard `range` (`turftest_range`) was notably faster, for a radius of 5 or more `RANGE_TURFS` (`turftest_range_turf`) was slightly faster.

</details>

# Changelog
No player facing changes.

### DIFF
--- a/code/__DEFINES/turfs.dm
+++ b/code/__DEFINES/turfs.dm
@@ -1,14 +1,15 @@
-#define RANGE_TURFS(RADIUS, CENTER) \
-	block( \
-		(CENTER).x-(RADIUS), (CENTER).y-(RADIUS), (CENTER).z, \
-		(CENTER).x+(RADIUS), (CENTER).y+(RADIUS), (CENTER).z \
-	)
-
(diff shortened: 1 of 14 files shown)


Drulikar (approved):
As far as I can tell code is fine. It has been TM'd for a while without any issues seemingly. I also tested some of the more obscure things changed such as teleporter, bluespace tomatoes, and doors bordering space (or not) without issue.

Issue #6536, opened 2024-06-22
Secondary Flamer Hardpoint (Tank) sometimes doesn't fire properly.


MPhonks:
### Testmerges

#6071: '[V] Sorokyne Strata: fixes and changes .' by Venuska1117 at commit 77b1ffb107 #6388: 'gayons' by 567Turtle at commit 4f9151f1ba #6334: 'Xenomorph Endgame' by Git-Nivrak at commit 5de79d5ca1 #6400: 'Adds Sunsets to Roundstart, adds Sunrises to Marine Victories, and Survs can hear Evo Announcement Screeches now' by GrrrKitten at commit 2927345720 #6308: 'Hostile Survivor Rework' by Steelpoint at commit bf51f9e551 #6039: 'Project ARES TM Holder (v5)' by realforest2001 at commit 29745fcac6

### Round ID

22759

### Description of the bug

Firing with the secondary flamer hardpoint of the tank sometimes at specific angles sometimes doesn't work and results in a single tile of flames.

### What's the difference with what should have happened?

The secondary flamer hardpoint has a few more tiles of range attributed to it than just... one.

### How do we reproduce this bug?

1. Get a tank.
2. Get a secondary flamer.
3. Try to shoot a few angles.
...
Profit.

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Doubleumc:
Am unable to replicate. Do you recall if there were any obstacles between the tank and the target? Was the tank actively moving during firing?

If you could describe the specific angles or get a screenshot that would be helpful.

PR #6699, opened 2024-07-13
Fix tank secondary flamer abnormal stoppage

Doubleumc:
# About the pull request

Fixes https://github.com/cmss13-devs/cmss13/issues/6536 , the tank's secondary flamer stopping after one tile.

As a side effect, flamers can now fire streams over a barricade that's immediately adjacent to them. Stream is still stopped if it hits the "wrong" side of a barricade after that first tile.

`get_cardinal_dir()` now returns a consistent direction.

The tank flamer shouldn't get hung up on random objects.
<details>

Before:
image
The stream is being blocked by the air pipe below. It is getting stopped because the flameshape tries to trace from the user (firer) on the first tile, and in this case the user is on an entirely different z-level (the tank).

Flameshape's `user` removed, using current `get_cardinal_dir`:
image
image
Removing reliance on the user makes the tank flamer work, but replacing the user with `get_cardinal_dir` causes issues with the triangular shape of B-gel. Namely `get_cardinal_dir` *randomly* picks which cardinal to get from a diagonal, potentially causing an obliquely fired stream to orient on top of itself.

After, with fixed `get_cardinal_dir`:
image

</details>

# Changelog
:cl:
fix: fixed tank secondary flamer stopping after one tile
balance: flamers can now fire streams over the "wrong" side of a barricade, when adjacent
/:cl:

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1100,25 +1100,17 @@ GLOBAL_DATUM(action_purple_power_up, /image)
 			else
 				air_master.tiles_to_update += T2*/
 
-/// Returns the cardinal dir between two atoms. Favors clockwise on perfect diagonals. Consistent and reversible.
-/proc/get_cardinal_dir(atom/start, atom/end)
+/// Returns the nearest cardinal dir between two atoms. Favors NORTH/SOUTH on perfect diagonals. Consistent and reversible.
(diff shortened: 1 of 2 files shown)


Drulikar:
Stale exempt until reviewed

Drulikar (approved):

PR #6712, opened 2024-07-15
Use DVIEW

Doubleumc:
# About the pull request

Switches several uses of `view` over to `dview` and related procs/defines instead. Only swaps over where it would be inappropriate for native BYOND darkness to effect the results, i.e. it cares about a lack of obstructions rather than actual visibility.

Created `doview()` and `USE_DOVIEW` as the `oview` version of `dview`.

Tweaked `dview` and related procs to work slightly more like `view` would. `view` is from the perspective of the provided atom, seeing itself, its loc (usually a turf), and if on a turf, any surrounding turfs. `dview` set its loc to the provided atom, meaning it was in the *contents* of that atom, and would see itself and its loc (the atom)... which, as an atom, means it wouldn't see surrounding turfs. Existing uses of `dview` got around this by using `get_turf` on the provided atom. I changed the behavior so this wasn't necessary, allowing you to `dview` from the perspective of a atom and have it work like `view` would.
Perhaps that should just be a built-in `get_turf` instead, but this more closely matches how `view` works.

Used `HIDE_INVISIBLE_OBSERVER` as the see invis for `dview` as all the uses I saw were for in-character things, and thus should only effect things that were also in-character. Anything as invisible as an observer or more is assumed to be out-of-character.

Should (currently) have no effect on the game as native BYOND darkness isn't used (all areas have luminosity), but is one step towards allowing that system to work appropriately if enabled.
<details>

Performance testing. All procs in question return a list of atoms.
The procs of interest:
image

Current, with darkness disabled (area luminosity = 1):
image
No performance difference between `view` and `dview`.

As a test, darkness enabled (area luminosity = 0):
image
`dview` is notably slower than `view`, likely because `view` has smaller lists of things it can see due to darkness and thus shorter loops.

</details>

# Changelog
No player facing changes.

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1337,29 +1337,34 @@ GLOBAL_LIST_INIT(WALLITEMS, list(
 GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 
 /// Version of view() which ignores darkness, because BYOND doesn't have it (I actually suggested it but it was tagged redundant, BUT HEARERS IS A T- /rant).
-/proc/dview(range = world.view, center, invis_flags = 0)
+/proc/dview(range = world.view, atom/center, invis_flags = 0)
 	if(!center)
 		return
 
-	GLOB.dview_mob.loc = center
-
+	GLOB.dview_mob.loc = isturf(center) ? center : center.loc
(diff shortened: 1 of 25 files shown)


Drulikar (approved):
Code appears fine, it has been tested without issue seemingly for a while, and I also tested a few more obscure scenarios without problem.

PR #6727, opened 2024-07-16
Buffs cades against projectiles

Zonespace27:
# About the pull request

Barricades and tables are now better against projectiles.
Assuming a standard rifleman with an unmodified m41a vs. a barricade, the probabilities of hitting are:

| Dist | Old | New | New (Xeno)
|---|---|---|---|
|  1 |  0 | 0   | 0|
|  2 |  -0.1833 |  0.0045 |0|
|  3 | -0.0466  |  0.52 |0|
|  4 |  0.0550 |  0.85 |0|
|  5 |  0.2367  |  0.85 |0.0045|
|  [6, ∞) | 0.3383  | 0.85  |0.52|

Cades are completely worthless in hvh situations, because they fail to block the overwhelming majority of bullets (especially in cqc, where most gunfights happen). This makes them better while still allowing people who get within 2 tiles to attack easier.

# Changelog
:cl:
balance: Barricades are now far better at blocking bullets from the front. They will not block most bullets if the shooter is within 2 tiles, however.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -660,7 +660,7 @@
 
 //Used by machines and structures to calculate shooting past cover
 /obj/proc/calculate_cover_hit_boolean(obj/projectile/P, distance = 0, cade_direction_correct = FALSE)
-	if(istype(P.shot_from, /obj/item/hardpoint)) //anything shot from a tank gets a bonus to bypassing cover
+	if(istype(P.shot_from, /obj/item/hardpoint) || istype(P.ammo, /datum/ammo/xeno)) //anything shot from a tank or a xeno gets a bonus to bypassing cover
 		distance -= 3
 
 	if(distance < 1 || (distance > 3 && cade_direction_correct))
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -668,10 +668,9 @@
 
 	//an object's "projectile_coverage" var indicates the maximum probability of blocking a projectile
 	var/effective_accuracy = P.get_effective_accuracy()
-	var/distance_limit = 6 //number of tiles needed to max out block probability
 	var/accuracy_factor = 50 //degree to which accuracy affects probability   (if accuracy is 100, probability is unaffected. Lower accuracies will increase block chance)
 
-	var/hitchance = min(projectile_coverage, (projectile_coverage * distance/distance_limit) + accuracy_factor * (1 - effective_accuracy/100))
+	var/hitchance = min(projectile_coverage, (projectile_coverage * distance / projectile_coverage_distance_limit) + accuracy_factor * (1 - effective_accuracy/100))
 
 	#if DEBUG_HIT_CHANCE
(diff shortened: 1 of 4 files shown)


Git-Nivrak:
Bullets being blocked from 2 and 3 tiles is *very* very bad considering an FOB defense situation
You are basically forcing a lot of weapons to be adjacent to cades now (not talking about hvh)

Perhaps make it so this new chance is only applied for bullets coming from the front side of the cade?

ihatethisengine:
Probably should have higher change to block bullets only from the front?

Zonespace27:
> Bullets being blocked from 2 and 3 tiles is _very_ very bad considering an FOB defense situation You are basically forcing a lot of weapons to be adjacent to cades now (not talking about hvh)
> 
> Perhaps make it so this new chance is only applied for bullets coming from the front side of the cade?

Rectified

> Probably should have higher change to block bullets only from the front?

It now does

RainbowStalin:
Don't know if this should be another PR, but i still think the Anti-FF cade and sandbags in particular could be buffed in terms of bullet resistance. Because you can still very easily shoot them down with a single mag (UPP that is).

Zonespace27:
> Don't know if this should be another PR, but i still think the Anti-FF cade and sandbags in particular could be buffed in terms of bullet resistance. Because you can still very easily shoot them down with a single mag (UPP that is).

Out of scope for this PR

Git-Nivrak:
Does this also affect xeno projectiles? Spits etc? If it does it probably shouldn't

code/modules/projectiles/projectile.dm
```diff
@@ -668,10 +668,9 @@
 
 	//an object's "projectile_coverage" var indicates the maximum probability of blocking a projectile
 	var/effective_accuracy = P.get_effective_accuracy()
-	var/distance_limit = 6 //number of tiles needed to max out block probability
 	var/accuracy_factor = 50 //degree to which accuracy affects probability   (if accuracy is 100, probability is unaffected. Lower accuracies will increase block chance)
 
-	var/hitchance = min(projectile_coverage, (projectile_coverage * distance/distance_limit) + accuracy_factor * (1 - effective_accuracy/100))
+	var/hitchance = min(projectile_coverage, (projectile_coverage * distance / (projectile_coverage_distance_limit * (cade_direction_correct ? 3 : 1))) + accuracy_factor * (1 - effective_accuracy/100))
```
Drulikar, line 673: xeno projectiles are affected by this change. At a minimum they should have a a similar mitigation like hardpoints do.

Drulikar (changes requested):

code/modules/projectiles/projectile.dm
Zonespace27: Do those numbers (see PR desc) look good?

Drulikar: They seem fine as long as that's not the numbers for xenos.

Drulikar:
Undraft when you're happy with it.

Drulikar (approved):

PR #6731, opened 2024-07-17
Buffs L42A damage to the same number as M4RA

Blundir:
# About the pull request

L42A and M4RA now have the same damage

L42A is a reskin and shouldn't differ stats wise, when M4RA damage was buffed, L42A was left behind because it wasn't easily obtainable at that moment, that changed.

Now L42A and M4RA have identical stats.

# Changelog
:cl:

balance: L42A and M4RA now have the same damage (and all other stats)
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -1642,7 +1642,7 @@
 	set_burst_amount(0)
 	accuracy_mult = BASE_ACCURACY_MULT + HIT_ACCURACY_MULT_TIER_5
 	accuracy_mult_unwielded = BASE_ACCURACY_MULT - HIT_ACCURACY_MULT_TIER_4
-	damage_mult = BASE_BULLET_DAMAGE_MULT + BULLET_DAMAGE_MULT_TIER_6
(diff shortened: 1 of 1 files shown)


Diegoflores31:
L42A is supposed to be discontinued from USCM , it should not have the same stats than M4RA

Blundir:
> L42A is supposed to be discontinued from USCM , it should not have the same stats than M4RA

It did have them for a long time

Diegoflores31:
> > L42A is supposed to be discontinued from USCM , it should not have the same stats than M4RA
> 
> It did have them for a long time

Because it was feared it was going to be too OP. It was buffed and nerfed . Currently is a completely different weapon because of the attachment variety it supports 

You could give L42A its own traits  . probably something like Mk1 - Mk2 kind of thing

Blundir:
> > > L42A is supposed to be discontinued from USCM , it should not have the same stats than M4RA
> > 
> > 
> > It did have them for a long time
> 
> Because it was feared it was going to be too OP. It was buffed and nerfed . Currently is a completely different weapon because of the attachment variety it supports
> 
> You could give L42A its own traits . probably something like Mk1 - Mk2 kind of thing

It is almost identical in terms of attachments, it does not support vgrip and agrip, but apart from it the list is identical

Drulikar:
stale exempt until reviewed

PR #6737, opened 2024-07-17
Planted flag bugfix and warcry removal

Zonespace27:
# About the pull request

- Fixes the flag being able to be duplicated
- The flag does a small amount of damage when hitting someone
- Item flag is indestructable
- Removes forced warcries

- Bugs are bad etc. etc.
- You are hitting someone with a big flagpole, it's fair that it does a bit of damage.
- Stops the flag from getting annihilated by a stray nade or etc
- I talked with another maintainer about this, and I was given some fair arguments about why forcing emotes is bad. In summary, on an RP server, the player in charge of a character should ideally have as much control over what the character does as possible. While this isn't the case for things like brain damage causing forced movement, those are consequences of negative things occurring to a character that serves as a hindrance to being able to play. As such, you should try to minimize the amount of times that forced emotes occur when not as a punishment, such as here. 
- What hasn't changed:
- The flag planter still warcries
- Having >= 14 people still makes a custom warcry sound for the flag planter.

<details>
image

</details>

# Changelog
:cl: Zonespace, Vile Beggar
add: Handheld planted flags are now indestructible and do a small amount of damage on hit.
fix: Plantable flags can no longer be duplicated.
del: Planting a UA flag no longer forces anyone but the planter to warcry.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/flags.dm
+++ b/code/game/objects/items/stacks/flags.dm
@@ -170,7 +170,11 @@
 	icon = 'icons/obj/structures/plantable_flag.dmi'
 	inhand_x_dimension = 64
 	inhand_y_dimension = 64
+	force = 15
+	throwforce = 5
+	hitsound = "swing_hit"
 	unacidable = TRUE
+	indestructible = TRUE
 	item_icons = list(
 		WEAR_L_HAND = 'icons/mob/humans/onmob/items_lefthand_64.dmi',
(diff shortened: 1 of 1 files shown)


Drulikar (approved):

Skatuk:
Unneeded change to make it so people aren't forced to emote everything else is fine just this change. There is a requirement for the warcry, and theres a cooldown. Personally I'm against this and just removing a cool feature like killed like 50% of why the flag is fun. You drop the flag and everyone warcry's and it looks cool.

PR #6763, opened 2024-07-19
Fixes lifeboat launch console being slashable / acidable

zzzmike:
# About the pull request

Lifeboat launch console is clearly meant to not be slashable / acidable (especially since it already has breakable=false flag and the lifepods are already properly flagged unacidable / unslashable. This fixes the problem.

Fixes a bug.

<details>

</details>

# Changelog

:cl:
fix: lifeboat launch console is no longer slashable / acidable
/:cl:

### DIFF
--- a/code/modules/shuttle/computer.dm
+++ b/code/modules/shuttle/computer.dm
@@ -364,6 +364,8 @@
 	icon_state = "terminal"
 	req_access = list()
 	breakable = FALSE
+	unslashable = TRUE
+	unacidable = TRUE
 	///If true, the lifeboat is in the process of launching, and so the code will not allow another launch.
 	var/launch_initiated = FALSE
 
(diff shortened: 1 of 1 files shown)


zzzmike:
(I didn't mean for this to be flagged as feature)

Drulikar (changes requested):
You need to also implement the proc `ex_act` and return (do nothing) since it currently can be destroyed that way too.

zzzmike:
thanks :)

Drulikar (approved):

HaultyAnonie:
Why doesn't `breakable = FALSE` account for all types of damage? What's it used for then?

Drulikar:
> Why doesn't `breakable = FALSE` account for all types of damage? What's it used for then?

`breakable` only applies to some structure's implementation of `attack_animal` and `attack_alien` But in general there isn't much standardization to defines like `breakable`, `unslashable`, `unacidable`, etc. You just need to look at their references in code to see what they do.

PR #6768, opened 2024-07-20
Department Heads now start with department-specific radio keys in their Office

efzapa:
# About the pull request

Gives all Department Heads 2 Encryption Keys either in their locker or in their office if they dont have one. These keys are programmed with the department specific channel and nothing else.

Allows roles to give out Radio Keys to people they see fit without also giving them channels they don't need (ie. giving someone MP comms, but you don't also want them to have squad/command comms)

<details>

</details>

# Changelog
:cl:
add: All Department Heads now start with department-specific radio keys in their Office.
/:cl:

### DIFF
--- a/html/changelogs/AutoChangeLog-pr-6667.yml
+++ b/html/changelogs/AutoChangeLog-pr-6667.yml
@@ -1,0 +1,7 @@
+author: "TotalEpicness5, Triiodine, monkeyfist, Blundir"
+delete-after: True
+changes:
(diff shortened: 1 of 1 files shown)


realforest2001 (changes requested):
Can you change this so it's a vendor item rather than loose in a closet. Most roles don't use closets at all, and I do not wish to encourage more break-ins to steal keys.

efzapa:
> Can you change this so it's a vendor item rather than loose in a closet. Most roles don't use closets at all, and I do not wish to encourage more break-ins to steal keys.

probably not doable unless i make *every* Head Equipment Vendor a separate vendor

which is doable, just not sure if that's what your intention in saying that was

efzapa:
why does this force close when resetting to earlier commits on VSC, bruh

PR #6762, opened 2024-07-19
Brig Detainment Cameras now have their own Console

efzapa:
# About the pull request

All cameras in the "detainment" area of the Brig now get their own console
Also adds cameras to each of the Perma Brig cells, one each respectively.

# Explain why it's good for the game

Generally a good QOL Update so that MPs dont have to type in or scroll *every single time* they want to check in on inmates, now they have a dedicated console for it.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

image
image

</details>

# Changelog

:cl:
add: All Brig Detainment Cameras are now linked to a new Brig Camera Console Computer, available in the Brig Lobby, Warden's Office, CIC, and CMP Office.
maptweak: Removed one of the Security Records Consoles in the Brig Lobby in place for a Brig Camera Console.
mapadd: Both Brig Perma Cells now have brig cameras in them.
/:cl:

PR #6783, opened 2024-07-23
New "Bulky" computers, small bug fixes also related to computers, addition of altitude controller to the CIC

LC4492:
# About the pull request

Adds new "Bulky" Computers to the CIC with their own unique designs, replacing the older and the added later computers (APOLLO and ARES consoles). Finally, physically adds the Altitude Controller to the CIC (With its own icon), giving players the opportunity of changing the Ship's altitude. Does a lot of fixes to the altitude changing code and makes small changes to the tactical map's sprite to make it look better. See the changelog for the explanation on how altitude changing works.

Attaching this here for good measure:
https://github.com/cmss13-devs/cmss13/pull/2760

The Altitude Control Console have been integrated and designed to have two main altitudes: The lowest the Almayer can go and the most optimal, with the highest altitude being a non-selectable and mainly detrimental altitude. Staying on low altitude gives you considerable boosts to OB cooldown and Dropship Transport flights, but also requires the Ship's engines themselves to cooldown, given that remaining in low altitude will cause overheat. When the Ship hits 100% of overheat, ARES will automatically set the Almayer to the non-selectable high orbit, of which will give a 50% detriment to both OB cooldown and Dropship Transport flights, with ARES returning the  ship to "most optimal" when the overheat percentage reach 0% again. That process can be manually jumped by a supervising and experienced Navigator. When not used, will not impact the round at all, given the Almayer will always remain in the "Most optimal" altitude when untouched.

I mean, I think the computers look good enough, the altitude controller being available to the players is neat, and that bugfixes usually don't hurt, so... do you have any explanation to why it could be bad?

<details>

image

</details>

# Changelog
:cl:
fix: Changes which icon the "Intel Database Computer" uses to fix a long-lasting bug. Alternate crew monitor icon fixes.
imageadd: Added new "IBM series 10" computers based on the "sensor computer" props that already exist in the code. Adds new tech-tree icon. Makes the tactical-map look a little better.
maptweak: Tweaks the CIC a little to accommodate the new computers. Adds the Altitude Controller to the ship's CIC. Changes the position of the medical belts to inside the armory. Makes the new "independent brig's console" on CIC to share with the bubble's camera console.
balance: Low altitude reduces OB cooldown and dropship flight times, but can only be maintained until overheated. Overheating will automatically enter high orbit for a time penalizing the OB cooldowns and dropship flight times until cooled over to overridden. Hijack flight time is unaffected.
/:cl:


code/modules/cm_marines/altitude_control_console.dm
```diff
@@ -20,9 +20,10 @@ GLOBAL_VAR_INIT(ship_temp, 0)
 GLOBAL_VAR_INIT(ship_alt, SHIP_ALT_MED)
 
 /obj/structure/machinery/computer/altitude_control_console
```
Drulikar, line 22: You will want to implement ex_act to ignore explosions, and make it unslashable/unacidable/unbreakable unless there is an avenue to repair/build the console again. Also `deconstructible` too depending?

code/game/machinery/computer/camera_console.dm
```diff
@@ -315,12 +315,19 @@
...
 	circuit = null
 
-
 /obj/structure/machinery/computer/cameras/almayer
 	density = FALSE
 	icon_state = "security_cam"
 	network = list(CAMERA_NET_ALMAYER)
 
+/obj/structure/machinery/computer/cameras/almayer/bulky
+	name = "Security Cameras"
```
Drulikar, line 324:
```suggestion
	name = "security cameras"
```
Else you need `\improper`

code/modules/cm_marines/altitude_control_console.dm
```diff
@@ -54,7 +55,7 @@ GLOBAL_VAR_INIT(ship_alt, SHIP_ALT_MED)
 				continue
 			current_mob.apply_effect(3, WEAKEN)
 			shake_camera(current_mob, 10, 2)
-		ai_silent_announcement("Attention performing high-G maneuverer", ";", TRUE)
+		ai_silent_announcement("Attention performing high-G maneuver", ";", TRUE)
```
Drulikar, line 58: Just remove this one, and fix the capitalization of `Orbital` above.

Drulikar (changes requested):

code/modules/cm_marines/altitude_control_console.dm
```diff
@@ -100,35 +111,38 @@ GLOBAL_VAR_INIT(ship_alt, SHIP_ALT_MED)
...
+				. = TRUE
+			if("med_alt")
+				change_altitude(user, SHIP_ALT_MED)
+				. = TRUE
+			if("high_alt")
+				change_altitude(user, SHIP_ALT_HIGH)
+				. = TRUE
+		message_admins("[key_name(user)] has changed the ship's altitude to [action].")
+
+	add_fingerprint(ui.user)
```
Drulikar, line 131:
```suggestion
	add_fingerprint(user)
```
Doesn't really matter, but if you have a the var, might as well use it rather than arbitrarily do it different for this.

code/modules/cm_marines/altitude_control_console.dm
```diff
@@ -100,35 +111,38 @@ GLOBAL_VAR_INIT(ship_alt, SHIP_ALT_MED)
...
+		switch(action)
+			if("low_alt")
+				change_altitude(user, SHIP_ALT_LOW)
+				. = TRUE
+			if("med_alt")
+				change_altitude(user, SHIP_ALT_MED)
+				. = TRUE
+			if("high_alt")
+				change_altitude(user, SHIP_ALT_HIGH)
+				. = TRUE
```
Drulikar, line 128:
```suggestion
```
Doesn't need to be selectable; there's zero benefit so you're just griefing your team doing so. And yes even if there is no front end button, that doesn't stop anyone if this string is accepted here.

code/modules/cm_marines/altitude_control_console.dm
```diff
@@ -100,35 +111,38 @@ GLOBAL_VAR_INIT(ship_alt, SHIP_ALT_MED)
...
 		return
-	GLOB.ship_alt = new_altitude
-	TIMER_COOLDOWN_START(src, COOLDOWN_ALTITUDE_CHANGE, 90 SECONDS)
-	for(var/mob/living/carbon/current_mob in GLOB.living_mob_list)
-		if(!is_mainship_level(current_mob.z))
-			continue
-		current_mob.apply_effect(3, WEAKEN)
-		shake_camera(current_mob, 10, 2)
-	ai_silent_announcement("Attention: Performing high-G manoeuvre", ";", TRUE)
+	else
```
Drulikar, line 136: This else can be removed and everything under it tabbed over because of the early return above.

code/modules/cm_marines/altitude_control_console.dm
Drulikar: This is still outstanding.

code/modules/cm_marines/altitude_control_console.dm
Drulikar, line 1: - It is weird that overheat can be at 100% and overheat will only occur after it ticks past 100%: I would recommend having the increments be 10% (or if you really want random between 10-20%), and for reaching 100% to be when overheating occurs and the ship is forced to high orbit
- I would recommend maybe making a note of what altitude the ship currently is in the tgui for the flight computers (both remote and in dropship) or estimated flight time +/- some randomness (but not actually result in randomness)

tgui/packages/tgui/interfaces/AltitudeControlConsole.jsx
Drulikar, line 1: Consider adding a line saying current altitude; though you can intuit by what buttons are disabled (or not).

Drulikar (changes requested):
Changelog and optionally description should mention what altitude does, especially since we've eliminated the hijack effects and only allow low/medium manually. Add it as a balance tag.

Otherwise appears working.

Drulikar:
Will very likely conflict with https://github.com/cmss13-devs/cmss13/pull/6762 due to CIC brig console computer (everything else should be handled automatically by hooks)

cm13-github:

cm13-github:

tgui/packages/tgui/interfaces/AltitudeControlConsole.jsx
```diff
@@ -45,21 +65,17 @@ export const AltitudeControlConsole = () => {
...
             </Button>
           }
         </Section>
+        <Section title="System Operational Warning">
+          <Box textAlign="center" fontSize="20px">
+            The Ship will always be set according to the disabled option, if
+            none of the two options available to you are disabled, that means
+            the ship is on high altitude
+          </Box>
+        </Section>
```
Drulikar, line 78: The information contained here is not a warning, nor is relevant. If you insist on keeping an information block here (that doesn't even display on the interface without scrolling), it should contain information relevant such as when overheating occurs and or what happens when overheating. Or the effects that even changing altitude even does.

code/modules/cm_marines/altitude_control_console.dm
```diff
@@ -44,34 +55,50 @@ GLOBAL_VAR_INIT(ship_alt, SHIP_ALT_MED)
...
+		GLOB.ship_alt = SHIP_ALT_MED
+		for(var/mob/living/carbon/current_mob in GLOB.living_mob_list)
+			if(!is_mainship_level(current_mob.z))
+				continue
+			current_mob.apply_effect(3, SLOW)
+			shake_camera(current_mob, 10, 2)
+		return
+	if(GLOB.ship_alt == SHIP_ALT_LOW && GLOB.ship_temp >= OVERHEAT)
+		ai_silent_announcement("Attention: Low altitude orbital maneuver no longer sustainable, moving to furthest geo-synchronous orbit until engine cooloff.", ";", TRUE)
 		GLOB.ship_alt = SHIP_ALT_HIGH
```
Drulikar, line 87: No cooldown is currently applied when the ship automatically makes a change. It needs to.

Drulikar (changes requested):
Changelog is too wordy. Remember this is what is displayed in game. Keep the information succinct. The current block for `balance` would be fine to contain within the body for about the PR though.

LC4492:
List of problems encountered:
1. Ares' messages were being repeated to each and every crewmember, with the spamming of over 30+ exact messages occuring often (fixed)
2. Dropship flight times are not properly tied to the ship's altitude, not working at all (Need to look further on why "GLOB.ship_alt" is not being enough)
3. Same above applies to the Orbital Cannon
4. Too much shaking from constant altitude changes were giving players motion-sickness (fixed)

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -58,15 +58,15 @@
...
+	var/flight_duration =  is_flyby ? DROPSHIP_TRANSIT_DURATION / GLOB.ship_alt : DROPSHIP_TRANSIT_DURATION * GLOB.ship_alt
 	if(optimised)
 		if(is_flyby)
 			flight_duration = DROPSHIP_TRANSIT_DURATION * 1.5
 		else
 			flight_duration = DROPSHIP_TRANSIT_DURATION * SHUTTLE_OPTIMIZE_FACTOR_TRAVEL
 
 	// recharge time before dropship can take off
-	var/recharge_duration = SHUTTLE_RECHARGE
+	var/recharge_duration = SHUTTLE_RECHARGE * GLOB.ship_alt
```
Drulikar, line 69:
```suggestion
	var/recharge_duration = SHUTTLE_RECHARGE
```
Recharge time does not need to be affected by altitude.

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -58,15 +58,15 @@
 		return
 
 	// initial flight time
-	var/flight_duration =  is_flyby ? DROPSHIP_TRANSIT_DURATION : DROPSHIP_TRANSIT_DURATION * GLOB.ship_alt
+	var/flight_duration =  is_flyby ? DROPSHIP_TRANSIT_DURATION / GLOB.ship_alt : DROPSHIP_TRANSIT_DURATION * GLOB.ship_alt
 	if(optimised)
 		if(is_flyby)
 			flight_duration = DROPSHIP_TRANSIT_DURATION * 1.5
 		else
 			flight_duration = DROPSHIP_TRANSIT_DURATION * SHUTTLE_OPTIMIZE_FACTOR_TRAVEL
```
Drulikar, line 66:
```suggestion
			flight_duration *= 1.5
		else
			flight_duration *= SHUTTLE_OPTIMIZE_FACTOR_TRAVEL
```
This is why you are seeing `GLOB.ship_alt` have no effect because if the ship is optimized; this code completely ignores the existing value of `flight_duration` where you applied `GLOB.ship_alt`

Drulikar (changes requested):
I will be committing below suggestions, but do look at them.

code/modules/shuttle/computers/dropship_computer.dm
Drulikar, line 1: For consistency:

- `1.5` should instead be turned into a define like `SHUTTLE_OPTIMIZE_FACTOR_FLYBY` here https://github.com/cmss13-devs/cmss13/blob/master/code/__DEFINES/objects.dm#L95-L100 
- `flight_duration = flight_duration / SHUTTLE_FUEL_ENHANCE_FACTOR_TRAVEL` lower down should be replaced with `flight_duration /= SHUTTLE_FUEL_ENHANCE_FACTOR_TRAVEL`
- `flight_duration = flight_duration * SHUTTLE_FUEL_ENHANCE_FACTOR_TRAVEL` lower down should be replaced with `flight_duration *= SHUTTLE_FUEL_ENHANCE_FACTOR_TRAVEL`

But this doesn't affect anything, just better for future contributors like yourself to better understand what's being done in this file.

Drulikar:
Changelog has been amended for spelling and clarity.

rythenx:
I'd really like an option to completely disable screen shake as a client side preference.  This is a graphical option that majority of games that have screen shake let you disable.  I'm really sensitive to screen shaking in games and will generally avoid games that make heavy use of it if they don't provide an option to disable it.  I really don't think it should be relied on as a balancing feature of using low orbit, there should be other drawbacks that don't just try to mess with your perception.  Slowdowns for movement or having a chance to fall over or things like that are all fine cause they don't cause this OOC problem.

Drulikar (changes requested):
Please just remove the altitude mechanic and UI and repurpose the computer made for it for something else. The altitude mechanic I am just no longer comfortable with nor see any avenue for its use.

cm13-github:

Drulikar:
Feel free to open again or request a reopen if altitude computer mechanic is removed.

PR #6583, opened 2024-06-28
Port WHO to TGUI

blackcrystall:
# About the pull request

Just QoL

# Explain why it's good for the game

Argghhh. just TGUI new panel? No benefits on round, but for visual and maybe a little bit server, because it don't generate every time for you who list

# Changelog

:cl: Blackcrystalic
qol: port who/staffwho to TGUI
/:cl:

PR #6786, opened 2024-07-24
Hold my Beer (Fix of hi intellligent data lookup in TGUI Who/Staff Who

blackcrystall:
:cl: BlackCrystalic
fix: No more admin data sended to normal players in who/staff who
/:cl:


Drulikar (changes requested):
- Both implementations listing out clients need to skip invisible admins being listed unless its the for admins list. See https://github.com/cmss13-devs/cmss13/pull/6583/files#diff-4da1f7388c33ae5a42dc442cf07f0a6830e02130bfe77927fc9406787d159fb3L178-L179 (you'll have to expand the hidden removed file to see the lines): 
```js
			if(!CLIENT_HAS_RIGHTS(src, R_STEALTH) && (CLIENT_IS_STEALTHED(C)))
				continue
```
- `usr` should never be used in a `ui_act`. Use `ui.user` instead.
- mobs_ckey should not be holding hard refs to mobs; and should instead hold weakrefs
- Consider a color key for https://github.com/cmss13-devs/cmss13/issues/6778

Drulikar (changes requested):
invisim & stealthed (these are two different things, thus the two different checks I previously mentioned) staff still appear in who for non-admins, and are still counted in the category in staff who. The only effect is invisim removes the entry under the dropdown (stealthed does nothing):
image

Drulikar (changes requested):

Drulikar (changes requested):
I again implore you to review and or tinker with the old who code you removed to properly maintain the same behavior it had before.

---

Please test your changes:
- Counts are now NotANumber for anyone non-staff
image
- Invisim and or stealthed staff still show up in Who to regular players: 
image
- Stealthed ought to note STEALTHED (again only those who have stealth permission would ever see it): 
image
- Logic in tgui code doesn't change the fact that the data is still sent to the client. Namely in this case you hid stealthed admins using tgui logic, but their data packet still contains who is stealthed: 
image

Drulikar (changes requested):
- Invismin staff still show up in Who for non-staff (however stealthed work correctly):
image
- When staff goes invisimin, they are hidden to other staff incorrectly (but because of above issue the Who is correct): 
image

- It is a bit weird that invismin retains assumingly the playing color, but stealthed then is red colored. If the text is going to include stealthed, then its unnecessary to change the color. 
image
image

To reiterate again:
- Invisim should hide that staff member to non-staff in StaffWho and Who.
- Stealth should hide that staff member to staff who don't have stealth perms in StaffWho and Who and non-staff in StaffWho and Who.
- The state of the person in both Who and StaffWho (e.g. Playing) should be displayed for all staff (if they can see that person due to above restrictions), but not for non-staff.

Drulikar (changes requested):
Additionally, why for the host is a lobby moderator colored green:
image
 But the moderator doesn't see them as green (and has no tooltips)?
image
Their color does not change when playing either:
image
 Isn't the color supposed to indicate the state that the tooltip describes?

Drulikar (changes requested):
- Currently a moderator cannot view the playing status of players or staff like a host can:
image
Only non-staff should have the limited view.
- Because of above, that staff member then can't even see themself when they go invisim. 
invis
Staff should still always have a staff view. The only escalated permission is that only management can see stealthed and utuilize stealth. Inivismin should only restrict non-staff from seeing them.

tgui/packages/tgui/interfaces/Who.jsx
```diff
@@ -142,48 +124,83 @@ const GetAddInfo = (props, context) => {
...
-        'border-color': admin ? color : '#2185d0',
-        'border-style': 'solid',
-        'border-width': '1px',
-        color: admin ? color : ckey_color,
+        borderColor: color || '#2185d0',
+        borderStyle: 'solid',
+        borderWidth: '1px',
+        color: color || ckey_color || '#2185d0',
       }}
       onClick={() => act('get_player_panel', { ckey: ckey })}
```
Drulikar, line 148: Clicking on any player button throws 
image
image
because `useBackend` is absent from this function.

tgui/packages/tgui/interfaces/Who.jsx
```diff
@@ -13,18 +13,22 @@ import { Window } from '../layouts';
...
+    factions_additional,
   } = data;
 
+  const total_players = mergeArrays(
+    base_data?.total_players,
+    player_additional?.total_players,
+    player_stealthed_additional?.total_players,
+  );
+
   const [searchQuery, setSearchQuery] = useLocalState('searchQuery', '');
```
Drulikar, line 28: useLocalState is depreciated. Use `useState` instead. https://react.dev/reference/react/useState

tgui/packages/tgui/interfaces/Who.jsx
```diff
@@ -13,18 +13,22 @@ import { Window } from '../layouts';
 export const Who = (props, context) => {
   const { act, data } = useBackend(context);
```
Drulikar, line 14: useBackend no longer takes an argument.
```suggestion
  const { act, data } = useBackend();
```

tgui/packages/tgui/interfaces/StaffWho.jsx
```diff
@@ -4,31 +4,24 @@ import { Window } from '../layouts';
 
 export const StaffWho = (props, context) => {
   const { data } = useBackend(context);
```
Drulikar, line 6: useBackend no longer takes an argument.
```suggestion
  const { data } = useBackend();
```

tgui/packages/tgui/interfaces/StaffWho.jsx
Drulikar, line 1: It would be ideal for this file to be converted to a tsx file, but it is not required.

tgui/packages/tgui/interfaces/Who.jsx
Drulikar, line 1: It would be ideal for this file to be converted to a tsx file, but it is not required.

tgui/packages/tgui/interfaces/Who.jsx
```diff
@@ -44,7 +48,7 @@ export const Who = (props, context) => {
                     fluid
                     onEnter={(e, value) =>
```
Drulikar, line 49: Nothing happens if you enter a valid ckey and then press enter (as in no player panel opens).

Additionally, if you enter an invalid ckey you get:
image
image

Drulikar (changes requested):
All above issues appear to be resolved.

Some of the tgui changes mentioned below are as a result of https://github.com/cmss13-devs/cmss13/pull/6244
image

tgui/packages/tgui/interfaces/Who.jsx
```diff
@@ -11,20 +13,24 @@ import {
...
   } = data;
 
-  const [searchQuery, setSearchQuery] = useLocalState('searchQuery', '');
+  const total_players = mergeArrays(
+    base_data?.total_players,
+    player_additional?.total_players,
+    player_stealthed_additional?.total_players,
+  );
+
+  const [searchQuery, setSearchQuery] = useState('', '');
```
Drulikar, line 30:
```suggestion
  const [searchQuery, setSearchQuery] = useState('');
```
useState does not have an implementation taking two arguments.

Drulikar (approved):

Drulikar (changes requested):
Resolve above since you disallow maintainer modification.

Drulikar (approved):

PR #6807, opened 2024-07-27
stops people from buckling others when they shouldn't be able to

zzzmike:
# About the pull request

adds checks to stop people from buckling others when they are knocked out etc.

fixes bug

<details>

</details>

# Changelog
:cl:
fix: stops people from buckling others when they are knocked out etc.
/:cl:

### DIFF
--- a/code/game/objects/objs.dm
+++ b/code/game/objects/objs.dm
@@ -277,7 +277,7 @@
 
 //trying to buckle a mob
 /obj/proc/buckle_mob(mob/M, mob/user)
-	if (!ismob(M) || (get_dist(src, user) > 1) || user.is_mob_restrained() || HAS_TRAIT(src, (TRAIT_IMMOBILIZED || TRAIT_INCAPACITATED || TRAIT_FLOORED)) || user.stat || buckled_mob || M.buckled || !isturf(user.loc))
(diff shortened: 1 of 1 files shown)


code/game/objects/objs.dm
```diff
@@ -277,7 +277,11 @@
 
 //trying to buckle a mob
 /obj/proc/buckle_mob(mob/M, mob/user)
-	if (!ismob(M) || (get_dist(src, user) > 1) || user.is_mob_restrained() || user.stat || buckled_mob || M.buckled || !isturf(user.loc))
+	if (!ismob(M) || (get_dist(src, user) > 1) || user.is_mob_incapacitated() || user.stat || buckled_mob || M.buckled || !isturf(user.loc))
+		return
```
Drulikar, line 281: Consider splitting this up to have `user.is_mob_incapacitated()` in the if block below so it has a message.

code/game/objects/objs.dm
Drulikar, line 1: Further down in this file is `// mobs that become immobilized should not be able to buckle themselves.` that should then be removed since you're checking TRAIT_IMMOBILIZED more broadly.

Drulikar (changes requested):

Drulikar (approved):

PR #6816, opened 2024-07-29
You can use russian warcry if you know russian

Blundir:
# About the pull request

Warcry now checks for the current default language selected instead of faction, so knowing russian you can yell russian warcries, can probably be expanded to other languages later.

https://github.com/user-attachments/assets/0fd15eca-a0f8-42ed-ab70-0cf126d78d5d

Basically copypaste of chunk from this https://github.com/RU-CMSS13/RU-CMSS13/pull/225

It just logically makes sense, also solves cases where UPP survivors or UPP background roles couldn't use the proper warcries which was very awkward.

# Changelog

:cl:
add: russian warcries can be used if russian language is set on default
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/emote.dm
+++ b/code/modules/mob/living/carbon/human/emote.dm
@@ -359,14 +359,15 @@
 	user.show_speech_bubble("warcry")
 
 /datum/emote/living/carbon/human/warcry/get_sound(mob/living/user)
+	var/default_lang = user.get_default_language()
 	if(ishumansynth_strict(user))
 		if(user.gender == MALE)
-			if(user.faction == FACTION_UPP)
+			if(default_lang == GLOB.all_languages[LANGUAGE_RUSSIAN])
 				return get_sfx("male_upp_warcry")
 			else
 				return get_sfx("male_warcry")
 		else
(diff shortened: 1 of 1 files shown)


ihatethisengine:
For survs it makes sense, for marines I am not so sure.

Blundir:
> For survs it makes sense, for marines I am not so sure.

Majority of people have english on default all of the time, not that many people even have custom languages, also there is tons of ethnically justified characters that should be able to do it, so i see no problem with it whatsoever

Skatuk:
Speedmerge pls

Noname995:
Cool feature :)

Zonespace27:
While I do think it makes sense, I'm against this as it takes away some of the uniqueness of the UPP (plus I know it'll cause marines to take russian to be dumb with warcries)

Blundir:
> While I do think it makes sense, I'm against this as it takes away some of the uniqueness of the UPP (plus I know it'll cause marines to take russian to be dumb with warcries)

Then at least make sure that all upp roles (espcially survs) have the upp faction for them

Zonespace27:
> Then at least make sure that all upp roles (especially survs) have the upp faction for them

I'd be fine with this, yeah

PR #6820, opened 2024-07-30
Updates close backpack icon

Blundir:
# About the pull request

Replaces red cross made in 2007 with an actual icon.

Before:

dreamseeker_ibtZ94zQ86

After:

dreamseeker_fB1ekGWiK7

Placeholder bad, but not only that, the cross itself is a hitbox for click detection, and cross isn't really ideal shape for such icon since most of the people would try to hit the centre of it (~81 pixels area), and new icon has bigger centre area overall (~144 pixels) and is easier to hit because it's a lot harder to miss on a big square without blind spots and which is actually connected to the main UI body.

Sprite itself from [TGMC](https://github.com/tgstation/TerraGov-Marine-Corps).

# Changelog
:cl:
imageadd: replaces ancient close backpack red cross with a modern icon
/:cl:


Git-Nivrak:
Make it wider, it is extremely hard to click

TopHatPenguin:
cm shouldn't accept tgmc altered sprites

Mister-moon1:
Has already been said but the sprite should be as wide as the original one, right now its inconvenient to click compared to the previous one

realforest2001 (approved):

PR #6825, opened 2024-07-30
autowiki 2: generates individual pages for guns

harryob:
just lets us be slightly more flexible with the autowiki


Drulikar (approved):
It runs locally, though I don't know how to fully test this pipeline so we can just see what it does in the end.

PR #6847, opened 2024-08-01
Fixes Healer drone Overheal exploit

Diegoflores31:
# About the pull request

Fixes Healer drone being able to give you "overhealing". by adding a 1 second cooldown to their healing and making the healing process faster.

You can still get a little over heal after this PR but is not as bad now.

The overhealing was insane . healer drone should not be able to preemptively heal you so you can keep healing mid-combat.

<details>

</details>

# Changelog
:cl:
balance: Healer drone apply salve now has a 1 second cooldown.
balance: Salve wound now heals slightly quicker.
/:cl:


GrrrKitten:
Metafriends weeping

ihatethisengine:
You kinda need to elaborate what exactly is exploit here. That healing over time heals over time? That you can transfer more HP than they have so they will keep healing even after hitting max hp? That you can stack several healing over time so they will heal quite fast? That you can fight while still healing? 

Anyway, you basically can do literally the same with the fruits. You should probably change the way healing over time works, maybe instead of stacking it should just kinda prolong the effect of healing. This way healing mid-combat will not be that strong. 

Also 1 second CD is pretty long, make it at least 0.5 IMO.

code/modules/mob/living/carbon/xenomorph/strains/castes/drone/healer.dm
```diff
@@ -124,9 +128,9 @@
 	adjustBruteLoss(amount * damage_taken_mod)
 	use_plasma(amount * 2)
 	updatehealth()
-	new /datum/effects/heal_over_time(target_xeno, amount, 10, 1)
-	target_xeno.xeno_jitter(1 SECONDS)
-	target_xeno.flick_heal_overlay(10 SECONDS, "#00be6f")
+	new /datum/effects/heal_over_time(target_xeno, heal_amount = amount)
+	target_xeno.xeno_jitter(5 SECONDS)
```
ihatethisengine, line 132: You clearly wanted to make jitter last 1 second, and flick 5, not the other way around

Diegoflores31: chat is real

Diegoflores31:
> You kinda need to elaborate what exactly is exploit here. That healing over time heals over time? That you can transfer more HP than they have so they will keep healing even after hitting max hp? That you can stack several healing over time so they will heal quite fast? That you can fight while still healing?
> 
> Anyway, you basically can do literally the same with the fruits. You should probably change the way healing over time works, maybe instead of stacking it should just kinda prolong the effect of healing. This way healing mid-combat will not be that strong.
> 
> Or maybe make hitting full HP remove the healing over time effect.
> 
> Also 1 second CD is pretty long, make it at least 0.5 IMO.
> 
> tl;dr you should just fix the thing you deem exploitable instead of making the caste less convenient to play, because with this PR you punish players who didn't mean to exploit

1.- you can stack healing on a xeno that can go beyond their maximun health . so for example if a drone uses all its stacks on a lurker when a lurker is almost full . the lurker will keep healing , basically he can go out in combat and receive insane amounts of damage . 
                example: you heal 400hp lurker with your 5-6 stacks of health . lurker goes into battle and somehow can tank up to 800  HP because the still has about 400 HP in healing for the next 8 seconds .  usually with other healings like queen is not a problem because it cannot be stacked and heals slow but with drone it is a problem because he can stack it.

2.- while fruits Can be "stacked" they cant be instantly eaten nor  you can stack 5 fruits in less than 1 second , unless you have 5 xenos feeding you simultaneusly of course but that is a crazy hypothetical compared to a drone just pressing a heal multiple times.

3.-  i have tested this and the healing is almost the same. I originally made it 0,5 but realized it made no difference unless it was 1 second.

code/modules/mob/living/carbon/xenomorph/strains/castes/drone/healer.dm
```diff
@@ -58,10 +58,14 @@
 	macro_path = /datum/action/xeno_action/verb/verb_apply_salve
 	action_type = XENO_ACTION_CLICK
 	ability_primacy = XENO_PRIMARY_ACTION_3
+	xeno_cooldown = 1 SECONDS
```
Drulikar, line 61:
```suggestion
	xeno_cooldown = 0.5 SECONDS
```

Drulikar (changes requested):
Otherwise feels fine

Drulikar (approved):

PR #6850, opened 2024-08-02
Fullscreen improvements

Doubleumc:
# About the pull request

Added a toggle fullscreen key, default F11.

Switched to `is-fullscreen` implementation, which preserves window size/maximized from before fullscreening. The previous `is-maximized` disable/enable dance was to avoid a BYOND issue with fullscreening a maximized window. That issue was resolved in 515.1639: https://www.byond.com/forum/post/2924219

Fullscreen now hides the menu bar.

Toggling fullscreen now triggers adaptive zoom and/or auto viewport resize if those are enabled.

Renamed the proc to `update_fullscreen` since that part wasn't actually toggling anything.

Boots without issue. Preference is toggled and preserved. Fullscreening while maximized works properly *on version 1639 and above*. Non-fullscreened screen state is preserved.

# Changelog
:cl:
add: Added toggle fullscreen hotkey, default F11
qol: fullscreening hides the menu bar
fix: toggling fullscreen preserves your non-fullscreened window options
/:cl:

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -730,7 +730,7 @@ GLOBAL_LIST_INIT(whitelisted_client_procs, list(
 					winset(src, "srvkeybinds-[REF(key)]", "parent=default;name=[key];command=whisper")
 
 /client/proc/update_fullscreen()
-	if(prefs.toggle_prefs & TOGGLE_FULLSCREEN))
+	if(prefs.toggle_prefs & TOGGLE_FULLSCREEN)
 		winset(src, "mainwindow", "is-fullscreen=true;menu=")
 	else
 		winset(src, "mainwindow", "is-fullscreen=false;menu=menu")
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -419,11 +419,7 @@ GLOBAL_LIST_INIT(whitelisted_client_procs, list(
 	if(prefs.lastchangelog != GLOB.changelog_hash) //bolds the changelog button on the interface so we know there are updates.
 		winset(src, "infowindow.changelog", "background-color=#ED9F9B;font-style=bold")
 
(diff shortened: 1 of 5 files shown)


code/modules/client/client_procs.dm
```diff
@@ -733,12 +729,16 @@ GLOBAL_LIST_INIT(whitelisted_client_procs, list(
 				if(WHISPER_CHANNEL)
 					winset(src, "srvkeybinds-[REF(key)]", "parent=default;name=[key];command=whisper")
 
-/client/proc/toggle_fullscreen(new_value)
-	if(new_value)
-		winset(src, "mainwindow", "is-maximized=false;can-resize=false;titlebar=false;menu=menu")
+/client/proc/update_fullscreen()
+	if(CHECK_BITFIELD(prefs.toggle_prefs, TOGGLE_FULLSCREEN))
```
harryob, line 733:
```suggestion
	if(prefs.toggle_prefs & TOGGLE_FULLSCREEN))
```

code/modules/client/preferences_toggles.dm
```diff
@@ -224,9 +224,9 @@
 	set category = "Preferences"
 	set desc = "Toggles whether the game window will be true fullscreen or normal."
 
-	prefs.toggle_prefs ^= TOGGLE_FULLSCREEN
+	TOGGLE_BITFIELD(prefs.toggle_prefs, TOGGLE_FULLSCREEN)
```
harryob, line 227:
```suggestion
	prefs.toggle_prefs ^= TOGGLE_FULLSCREEN
```

harryob (changes requested):
we don't really like these helpers - they were incidentally ported at some point before tg removed them (https://github.com/tgstation/tgstation/pull/49106) and we should do the same

harryob (approved):

PR #6864, opened 2024-08-03
Fixes and Readjusts Larva Surgery 

MPhonks:
# About the pull request

Currently, the way the larva surgery was written, using the hemostat for the final larva removal step was **SLOWER** than using ghetto tools such as the wirecutters and fork. This PR aims to fix the issue by making the hemostat the best tool, whilst keeping the original time it took to do so, and readjusts the niche "using one's hands" in an emergency to be the fastest way at the cost of both the doctor's and patient's health.

People used to ghetto surgery on larva will notice it being much slower.

It makes no sense for wirecutters and the fork to be faster than using the intended tool. Having the option to make an emergency procedure with one's hands gives doctors a satisfying way of saving their patients from impending chestburst death.

N/A

# Changelog

:cl: MarpleJones
fix: Using the hemostat on the final larva surgery step is now better than using wirecutters or the fork.
add: The final larva surgery step will now apply organ damage when done barehanded. Barehanded is faster than using tools by a couple of seconds, at the cost of the doctor's and the patient's health.
add: Adds an acidsplash sound effect to doing the barehanded step.
/:cl:


code/modules/surgery/chestburster.dm
```diff
@@ -111,12 +111,12 @@
...
+	Unlike before, the hemostat is now the best tool for removing removing the larva, as opposed to wirecutters and the fork.*/
 	tools = list(
-		/obj/item/tool/surgery/hemostat = 1.5,
-		/obj/item/tool/wirecutters = SURGERY_TOOL_MULT_SUBOPTIMAL,
-		/obj/item/tool/kitchen/utensil/fork = SURGERY_TOOL_MULT_SUBSTITUTE
+		/obj/item/tool/surgery/hemostat = 1.5, //Same as optimal (1) for 9s
+		/obj/item/tool/wirecutters = 1.8, //Same as suboptimal (1.2) for 9s
+		/obj/item/tool/kitchen/utensil/fork = 2.1 //Same as substitute (1.4) for 9s
 		)
 	time = 6 SECONDS
```
Contrabang, line 121: Is there any reason to not just increase the timer by 1.5 times and then use the normal defines?
```suggestion
		/obj/item/tool/surgery/hemostat = SURGERY_TOOL_MULT_IDEAL, //Same as optimal (1) for 9s
		/obj/item/tool/wirecutters = SURGERY_TOOL_MULT_SUBOPTIMAL, //Same as suboptimal (1.2) for 9s
		/obj/item/tool/kitchen/utensil/fork = SURGERY_TOOL_MULT_SUBSTITUTE //Same as substitute (1.4) for 9s
		)
	time = 9 SECONDS
```

MPhonks: It's a bit sloppy, but the reason I didn't use the normal defines was to keep the interaction speed we previously had when doing the surgery step without tools (IE: 9 seconds with hemostat, 6 seconds without any tools).

I also don't quite know enough about ss13 coding, so I opted for the solution which looks 🤨. I would have used the defines if I knew how to check whether or not a tool is being used, but that seemed like too much of an undertaking for my beginner self.

Contrabang: That makes sense, doing something like this so we can still have it match the defines is probably good then.
```suggestion
		/obj/item/tool/surgery/hemostat = 1.5 * SURGERY_TOOL_MULT_IDEAL,
		/obj/item/tool/wirecutters = 1.5 * SURGERY_TOOL_MULT_SUBOPTIMAL,
		/obj/item/tool/kitchen/utensil/fork = 1.5 * SURGERY_TOOL_MULT_SUBSTITUTE
		)
	time = 6 SECONDS
```

code/modules/surgery/chestburster.dm
```diff
@@ -148,10 +148,16 @@
...
-				SPAN_WARNING("[user]'s hands are burned by acid as \he pulls a wriggling parasite out of [target]'s ribcage!"))
-
+				SPAN_WARNING("Your hands and your patient's insides are burned by acid as you forcefully rip a wriggling parasite out of [target]'s ribcage!"),
+				SPAN_WARNING("[user]'s hands are burned by acid as \he rips a wriggling parasite out of your ribcage!"),
+				SPAN_WARNING("[user]'s hands are burned by acid as \he rips a wriggling parasite out of [target]'s ribcage!"))
+			var/datum/internal_organ/I = pick(surgery.affected_limb.internal_organs)
+			I.take_damage(5,0)
+			if(target.stat == CONSCIOUS)
+				target.emote("scream")
+			target.apply_damage(15, BURN, target_zone)
```
harryob, line 158:
```suggestion
			var/datum/internal_organ/impacted_organ = pick(surgery.affected_limb.internal_organs)
			impacted_organ.take_damage(5, FALSE)
			if(target.stat == CONSCIOUS)
				target.emote("scream")
			target.apply_damage(15, BURN, target_zone)
```
no single letter vars, no using 0 instead of FALSE (i know this is copypaste from elsewhere, but!)

harryob (changes requested):

harryob:
by the way, when maintainers have reviewed something and requested changes, we convert the pr to a draft:
CleanShot 2024-08-05 at 13 54 21@2x

you then have to mark your pr as ready for review again:
CleanShot 2024-08-05 at 13 54 50@2x

otherwise it'll just get lost! obviously no worries this time, though

harryob (approved):

PR #6874, opened 2024-08-04
Normalized Queen's build range to match every other building xeno

TheKillfish:
# About the pull request
Changes the Queen's build range when not on ovi from 0 to 1 and updated a very old note regarding it.

Consistency with other building castes plus 0 range building is very awkward. This doesn't affect building when on ovi because ovi building ignores build range.

Added single line of code that works with other xenos.

# Changelog

:cl: Killfish
qol: Queen now has a build range of 1 instead of 0. Does not affect building on ovi.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Powers.dm
@@ -24,7 +24,7 @@
 
 	var/turf/current_turf = get_turf(A)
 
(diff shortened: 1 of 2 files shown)


Drulikar:
It is an intentional design decision that the queen building range is what it is off Ovi. In general most things encourage the queen to be on Ovi rather than off. You should generally expect a queen to go off Ovi if the hive is capturing poorly (so the queen's cc is needed), or if a fob is being sieged. Although this seems like a harmless change, I don't think it is needed.

PR #6884, opened 2024-08-05
Revert "Buffs cades against projectiles"

Zonespace27:
Reverts cmss13-devs/cmss13#6727

This has had a few too many knock-on effects (see below) and I'm a bit too burned out to properly fix this; will re-introduce this later when properly adjusted.

- Not happy with xeno cade hit % adjustments
- I've heard there's issues with cades getting shredded on diagonals
- You can now do what's seen in the image without issue, and you couldn't previously 
image


Drulikar (approved):
Though I am interested in what this was trying to do.

Steelpoint:
Beyond the scope of this PR but this would have been a great thing for HvH events. Perhaps if the concept can't be made to work properly for the base game, it could be integrated as a toggle for the HvH game mode? (WYCI I suppose).

PR #6885, opened 2024-08-05
improvements to research datum

kiVts:
# About the pull request
Ever since the init argument pr I had common sense take over and understand how goofy that was, so we are now here. This makes it *much* easier to add technology that doesnt spawn something. eg techtree upgrades while also stopping the init var nonsense, THIS is how it should have been.

Merge after anti caustic armor so I can adapt it easily

If someone adds upgrade that does something that requires non physical items, they can now.

<details>

</details>

# Changelog
:cl:
code: Research datum now supports more broad ideas in the future.
/:cl:

### DIFF
--- a/code/datums/research_upgrade_datum.dm
+++ b/code/datums/research_upgrade_datum.dm
@@ -9,6 +9,8 @@
 	var/value_upgrade = 1000
 	///In which tab the upgrade should be.
 	var/upgrade_type
+	///Path to the item, upgrade, if any.
+	var/item_reference
 	///Clearance requirment to buy this upgrade. 5x is level 6. Why is it not that way? no one knows.
 	var/clearance_req = 5
 	///The change of price for item per purchase, recommended for mass producing stuff or limited upgrade.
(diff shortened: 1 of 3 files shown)


harryob (approved):

harryob (changes requested):
actually, why not just have the default implementation new the held typepath, and just override it for the ones you give extra new() args to, so the autodoc ones?

kiVts:
Good catch

code/datums/research_upgrade_datum.dm
```diff
@@ -22,53 +20,66 @@
 	///the maximum price which we cant go any more expensive, usually dont need to set this if change price is 0 or negative
 	var/maximum_price = INFINITY
 
+///gets called once the product is purchased, override if you need to pass any special arguments or have special behavior on purchase.
+/datum/research_upgrades/proc/on_purchase(turf/machine_loc)
+	if(isnull(item_reference))
+		return
+	new item_reference(machine_loc)
+	return
```
harryob, line 28:
```suggestion
```

harryob (changes requested):

harryob (approved):

PR #6901, opened 2024-08-08
Chat Filter

Git-Nivrak:
# About the pull request
Adds a chat filter which automatically prevents messages in-character, OOC, LOOC or dchat from being sent if they contain certain words, this will notify the user and warn them that they will be banned should they try to circumvent it.

This is obviously just a config option so if this is merged the config will have to be edited accordingly.

Instead of instantly banning people for saying certain words, simply prevents them from saying them in the first place - this way no harm is done to anyone who might be offended as well as the player which (possibly mistakenly) said the word not being banned and instead let know the word is not allowed.
<details>

</details>

# Changelog
:cl:
config: Added config options for filtering out certain words
/:cl:

### DIFF
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -62,12 +62,12 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
 /proc/prefix_to_channel(prefix)
 	return GLOB.department_radio_keys[prefix]
 
-/proc/filter_message(user, message)
+/proc/filter_message(client/user, message)
 	var/filter = CONFIG_GET(string/word_filter)
 	var/message_lowercase = lowertext(message)
 	if(!filter)
 		return TRUE
-		
+
 	if(findtext(message_lowercase, regex(filter)))
 		to_chat(user,
 			html = "\n<font color='red' size='4'><b>-- Word Filter Message --</b></font>",
@@ -77,9 +77,9 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
 			html = "\n<font color='red' size='4'><b>Your message has been automatically filtered due to its contents. Trying to circumvent this filter will get you banned.</b></font>",
 			)
 		SEND_SOUND(user, sound('sound/effects/adminhelp_new.ogg'))
-		log_admin("[user.client] triggered the chat filter with the following message: [message].")
+		log_admin("[user.ckey] triggered the chat filter with the following message: [message].")
 		return FALSE
-	
+
 	return TRUE
 
 ///Shows custom speech bubbles for screaming, *warcry etc.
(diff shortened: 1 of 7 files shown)


code/modules/mob/living/say.dm
```diff
@@ -62,6 +62,24 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
 /proc/prefix_to_channel(prefix)
 	return GLOB.department_radio_keys[prefix]
 
+/proc/filter_message(user, message)
+	var/list/banned_words = CONFIG_GET(str_list/word_filter)
+	var/message_lowercase = lowertext(message)
+
+	for(var/word in banned_words)
+		if(findtext(message_lowercase, regex(lowertext(word))))
```
Zonespace27, line 70: For the sake of performance, you should be doing this differently.
The config should be a single regex instead of a list of banned words. As such, the code should be changed to
```
    if(findtext(message_lowercase, regex(config_entry_here)))
```
And the config entry should be a single string stead of a string list. Additionally, don't lowertext() the naughty word regex, that can be handled in the config.

The performance gains are noticeable. At a list length of two words, your code can run ~46560 times in a tick. Mine can run ~96342 times in a tick. The difference in perf becomes even larger as the word list grows, with a list length of seven words being ~19670 and ~79061 times in a tick, respectively.

code/modules/mob/living/say.dm
```diff
@@ -62,6 +62,24 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
...
+			to_chat(user,
+				html = "\n<font color='red' size='4'><b>-- Word Filter Message --</b></font>",
+				)
+			to_chat(user,
+				type = MESSAGE_TYPE_ADMINPM,
+				html = "\n<font color='red' size='4'><b>Your message has been automatically filtered due to its contents. Trying to circumvent this filter will get you banned.</b></font>",
+				)
+			SEND_SOUND(user, sound('sound/effects/adminhelp_new.ogg'))
+			return FALSE
```
Zonespace27, line 79: Have it send a message to admins maybe?

config/example/word_filter.txt
```diff
@@ -0,0 +1,6 @@
+## Word filter
+## WORD_FILTER <word>
+## Case insensitive, Uses regex
```
Zonespace27, line 3: case sensitive, with my changes

config/example/config.txt
```diff
@@ -5,6 +5,7 @@
 # $include resources.txt
 # $include icon_source.txt
 # $include relays.txt
+$include word_filter.txt
```
Zonespace27, line 8:
```suggestion
# $include word_filter.txt
```

Zonespace27 (changes requested):

code/modules/mob/living/say.dm
Git-Nivrak: The entire point is this is a preliminary measure to admin intervention

PR #6931, opened 2024-08-11
"See who is boss of the GYM" (Global playtimes section)

blackcrystall:
# About the pull request

Just fun around, now you can see global playtimes (the best numbers for each role)

(This can be optimized, with just some mess around request order editing and selecting greater value, then just ordering every role, but... NUH UH! I'm lazy)

# Changelog

:cl:BlackCrystalic
add: Global playtimes, now every player can see the best playtime for each role
/:cl:


zzzmike:
I could see this being used in negative ways ("wow you have 500 hours as nurse what's wrong with you, just play doctor" etc.). If you added an option to opt-in/opt-out (with the default being opt-out) of being included in the stats, I would support this.

zzzmike:
I agree, that's a great idea, hide ckey is perfect, no reason to exclude from stats completely.

HeresKozmos:
Not particularly a fan of this; even if it's just the name of the person (with no CKey or other identifiable information*) then even that is enough to tell who that character belongs to as some people openly list their names/usernames - zzzmike also makes a good point of "Wow you have (*insert number*) of hours on (*role*), you should play (*role*) instead." being used to annoy people.

TheManWithNoHands:
Yeah, best have as option to fully opt out. Otherwise this will just be a place the sweaty meta huggers go to find their next victim to bully.

Drulikar:
We do not need more ways for players to bully other players.

PR #6932, opened 2024-08-11
Lv maxpoplock

cuberound:
# About the pull request

caps LV to not be votable for on highpop ( not shure how to test this on local but the var exists there for it)

There is big fight about updating LV by "making it bigger" "making chokes wider" agains "keep it the same" "original CM" to solve this situation by adressing the main issue of LV being too small with tight chokes, the least invasive and agreed for on discord maxpop is added, works same as minpop on highpop maps preventing it from being voted for and being selected above 129 players. The issues with LV of size and chokes are due to it being made for low amount of players and thus best course of action is to keep it in its role. And getting LV on highpop is kidna pain everyone as it ends in cave stalemate majority of times and is not generly fun to play on highpop.

<details>

</details>

# Changelog

:cl:
add: changes LV 624 into lowpop map, it can not be voted for above 129 players
/:cl:


Drulikar:
From discord:
> I'm not interested in the reverse restriction of if there's too many players the map isn't available.
Either make the last played map unavailable for another round, or weight votes differently.
What I mean by weigh votes differently: if there's many players, it takes twice the votes to select a map without pop limit, whereas a pop limited map is normal weight.
The carry over votes would be weighted whatever is the current population.

CABSHEP:
It is super toxic to just outright block a map from being playable because there is too many people.

No, LV is not too small for even 250 pop CM.  

This LV hate has to stop. 

One idea would be to adress the issues directly, so if highpop is detected, chokepoints should get wider and Central Cave Network should have it's southern cave wall should get additional opening in the middle, just north-west from t-fort.

PR #6933, opened 2024-08-12
Datumizes Specialist Sets

Zonespace27:
# About the pull request
Converts specialist sets into datums instead of list/string hell, improves a fair bit of surrounding code

Fixes a few things:
- If a heavy sniper went to cryo, it would reopen the sniper spot instead
- Demo/Scout had a chance to not have the skill to use their c4 if they got a specialist set from their vendor instead of a spec kit

Our old system sucked, this sucks significantly less

<details>

Tested:
- Vending spec sets
- Cryoing as spec
- Using spec kits
- Cryoing as spec kit spec

</details>

# Changelog
:cl:
fix: Fixed heavy sniper spec opening up the wrong specialist slot on admin cryo
/:cl:


code/game/machinery/vending/cm_vending.dm
```diff
@@ -565,48 +565,28 @@ GLOBAL_LIST_EMPTY(vending_products)
...
+								return
+
+							if(GLOB.specialist_set_name_dict[p_name].get_available_vendor_num() <= 0)
 								to_chat(user, SPAN_WARNING("That set is already taken."))
 								vend_fail()
 								return FALSE
-							var/obj/item/card/id/card = human_user.get_idcard()
-							if(!card?.check_biometrics(user))
+
+							var/obj/item/card/id/card = human_user.wear_id
```
kiVts, line 583:
```suggestion
							var/obj/item/card/id/card = human_user.get_idcard()
```
Didnt drathek spent bunch of time to move *all* of these to get_idcard?

code/modules/cm_marines/specialist.dm
```diff
@@ -38,3 +38,151 @@
...
+		redeemer.put_in_any_hand_if_possible(new kit_typepath, FALSE)
+		for(var/path in incompatible_sets)
+			GLOB.specialist_set_datums[path].available_kit_num--
+		ADD_TRAIT(redeemer, TRAIT_SPEC_KIT, TRAIT_SOURCE_INHERENT)
+	else
+		for(var/path in incompatible_sets)
+			GLOB.specialist_set_datums[path].available_vendor_num--
+		ADD_TRAIT(redeemer, TRAIT_SPEC_VENDOR, TRAIT_SOURCE_INHERENT)
+	ADD_TRAIT(redeemer, TRAIT_SPEC(trait_to_give), TRAIT_SOURCE_INHERENT)
+	var/obj/item/card/id/idcard = redeemer.wear_id
```
kiVts, line 103:
```suggestion
	var/obj/item/card/id/idcard = redeemer.get_idcard()
```

kiVts (changes requested):
two things that I saw

code/modules/cm_marines/specialist.dm
kiVts: Otherwise I think it looks good? not doing anything incase its *somehow* intentional

code/game/machinery/vending/cm_vending.dm
Zonespace27: My bad yeah

Zonespace27:
> Is it not currently reopening both? It removes both from the list when picking either, so the re-opening was just piggybacking off of it (old cryopod.dm line 345)

Looks like it was only done in when cryoed by pod, not when [cryoed by admin](https://github.com/cmss13-devs/cmss13/pull/6933/files#diff-e4907929eb30ca4a557b4a0cf23f01f2c39a8bbd26479cfb63d44e8e3684583e)

harryob (approved):

PR #6978, opened 2024-08-19
Bursting+facehugger stats no longer count NPCs

zzzmike:
# About the pull request

Bursting+facehugger stats no longer count NPCs

Right now the stats seem to be half and half on whether they count NPCs or not, it should be one way or the other.

<details>

</details>

# Changelog
:cl:
fix: bursting+facehugger stats disregard NPCs
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Embryo.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Embryo.dm
@@ -341,7 +341,7 @@
 		if(burstcount)
 			step(larva_embryo, pick(GLOB.cardinals))
 
-		if(GLOB.round_statistics && (ishuman(victim)) && (SSticker.current_state == GAME_STATE_PLAYING) && (world.time - SSticker.round_start_time > 1 MINUTES))
+		if(GLOB.round_statistics && (ishuman(victim)) && (SSticker.current_state == GAME_STATE_PLAYING) && (ROUND_TIME > 1 MINUTES))
 			GLOB.round_statistics.total_larva_burst++
 		GLOB.larva_burst_by_hive[hive] = (GLOB.larva_burst_by_hive[hive] || 0) + 1
 		burstcount++
--- a/code/modules/mob/living/carbon/xenomorph/Embryo.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Embryo.dm
@@ -341,7 +341,7 @@
 		if(burstcount)
 			step(larva_embryo, pick(GLOB.cardinals))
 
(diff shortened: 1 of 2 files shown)


harryob (changes requested):
ishuman() || isyautja() is redundant - all yautja are human, so the first would pass, and the second would never be evaluated for yautja anyway

zzzmike:
thanks harry - now that I think about it I think I need to also add something that stops it from counting burst stats in the first minute or so from spawning at roundstart, because it doesn't really make sense that the initial instant bursts from roundstart xenos should count in the stats either (if anyone knows the best way to write that please lemme know)

zzzmike:
thank you @Doubleumc for the help with the round time code on discord 👍

code/modules/mob/living/carbon/xenomorph/Embryo.dm
```diff
@@ -341,7 +341,7 @@
 		if(burstcount)
 			step(larva_embryo, pick(GLOB.cardinals))
 
-		if(GLOB.round_statistics)
+		if(GLOB.round_statistics && (ishuman(victim)) && (SSticker.current_state == GAME_STATE_PLAYING) && (world.time - SSticker.round_start_time > 1 MINUTES))
```
harryob, line 344:
```suggestion
		if(GLOB.round_statistics && (ishuman(victim)) && (SSticker.current_state == GAME_STATE_PLAYING) && (ROUND_TIME > 1 MINUTES))
```

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
```diff
@@ -350,7 +350,7 @@
 	else
 		target.visible_message(SPAN_DANGER("[src] violates [target]'s face!"))
 
-	if(GLOB.round_statistics && ishuman(target))
+	if(GLOB.round_statistics && (ishuman(target)))
```
harryob, line 353:
```suggestion
	if(GLOB.round_statistics && ishuman(target))
```
this define has parenthesis around it anyway - is this necessary?

harryob (changes requested):

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
zzzmike: good catch, it was a leftover from the yautja thing woops

zzzmike:
thanks again harry :)

harryob (approved):

PR #6903, opened 2024-08-08
Keycard authentication device now reqires two users to activate events

cuberound:
# About the pull request

it is  writen down that two are reqired to activate red and so on but one can swipe twice fast and it work. this fixes it

# Explain why it's good for the game

bugfix and has rp incentive to get second man to start redalert

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
fix: two persons are needed to activate red, ert and emergency acces using cards
/:cl:

PR #6995, opened 2024-08-21
Revert #6903

Mister-moon1:
Reverts cmss13-devs/cmss13#6903

On lowpop red alert and other actions requiring keycards can become impossible due to lack of CIC staff, along with the fact of if you are not a commander you cannot use ares to activate red alert and there will likely not be a joe

# Changelog
:cl:
balance: Keycard swipers no longer require two unique keycards to function.
fix: fixed the description text for the card swipe to reduce confusion
/:cl:

### DIFF
--- a/code/modules/security_levels/keycard_authentication.dm
+++ b/code/modules/security_levels/keycard_authentication.dm
@@ -1,6 +1,6 @@
 /obj/structure/machinery/keycard_auth
 	name = "Keycard Authentication Device"
-	desc = "This device is used to trigger station functions, which require more than one ID card to authenticate."
+	desc = "This device is used to trigger station functions, which require multiple swipes of an ID card to authenticate."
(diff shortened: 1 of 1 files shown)


code/modules/security_levels/keycard_authentication.dm
```diff
@@ -36,11 +36,8 @@
 			if(active == 1)
 				//This is not the device that made the initial request. It is the device confirming the request.
 				if(event_source)
-					if(event_source.event_triggered_by == user)
-						user.visible_message(SPAN_DANGER("Your ID is rejected, as it is the one that triggered the event!"))
-						return
 					event_source.confirmed = 1
-					event_source.event_confirmed_by = user
+					event_source.event_confirmed_by = usr
```
Zonespace27, line 40:
```suggestion
					event_source.event_confirmed_by = user
```

harryob (approved):

PR #7009, opened 2024-08-23
Fix for broken nightmare inserts

FOFFLES105:
# About the pull request

All but one nightmare insert for Fiorina was broken due to some strange naming bug. In addition, one nightmare insert for Kutjevo was broken.  This PR fixes those nightmare inserts and adds them back.  Also cleaned up several unused sprinkles and moved them to an appropriate folder.

More nightmare inserts are cool. 

<details>

</details>

# Changelog

:cl:

fix: fixed issue where nightmare inserts would not spawn
fix: moved unused nightmare inserts to "unused" folder so that they don't cause errors.

/:cl:


harryob (changes requested):
i don't think this is the best way to do it, we can fix this in code

in code/modules/nightmare/nmnodes/mapload.dm, on line 58, with the var/regex/matcher, we can replace the new() statement with `new(@"^([0-9]+)([\.\+]).*?\.dmm$", "i")` instead

harryob (approved):

PR #7025, opened 2024-08-24
Fixes giant lizard code breaking when disarmed, and adds some minor features to the mob

VileBeggar:
# About the pull request

stops the code from breaking when getting disarmed
they'll no longer move a tile before resting
fixes factions being constantly added by feeding them even if they were an ally beforehand (also stops the message spam)
fixes (sloppily) a weird bug where the mob would start moving even though its sprite shows it as incapacitated.
fixes (hopefully) a bug where they would never lose their target_mob when they get stuck in alert stance somehow, which would lead them to becoming inert rocks.
fixes preds taming lizards not working (their faction_group was set to NETURAL, so their actual faction was never added as an ally)
what did i learn? NEVER TRUST BYOND'S INBUILT PATHFINDING PROCS TO WORK 

also adds code for breaking structures, both for the lizard AI and a player-controller one. touches up some minor things, like retreating.
they will now switch targets if they get damaged and their original one is far away
they can now vault over window frames and tables
the range for losing targets is now increased which lets skirmishes last longer and better (hopefully?).
lizards will now respawn once after a random amount of time if the current number of alive lizards is small
they'll now eat anything that has meat protein in it 
damage against xenos is increased by a 1/3rd more or less
adds tongue flickering
adds wound overlays

firstly, bug is bad.
secondly the other changes are something i've noticed to be needed while observing a few rounds:

1. structure destruction is a good touch since they had no way of breaking things before.
2. considering it's a mob similar to a runner, and is fast and nimble, i don't see any reason why it wouldn't be able to vault over tables and window frames. it helps it fight in urban areas.
3. retreat times are lowered because if they get stuck retreating, the long length of the retreat would end in their death due to no valid escapes.
4. target switching stops the mob from ignoring an entire group of enemies just to attack someone who sneezed on it and is far away
5. lizards respawning will help keep the outer areas a bit more dangerous even later in the round, and it allows for marines to interact with them even if they got kidnapped/killed by survivors or xenos early into the round
6. checking for meat protein is better than only comparing it to some list. the list is still there due to some MRE snacks not having protein and resin fruit not having any at all
7. even though xenos might be the perfect lifeform, i'd like it if these guys proved more of a threat than just a slight annoyance to xenos. right now the only threat they pose is if two of them swarm a runner.
8. tongue flickering is a neat little addition, and it replaces a boring text emote
9. wound overlays are a very good way of roughly displaying the HP of the mob instead of being forced to either guess or examine it midcombat

<details>

</details>

# Changelog
:cl: VileBeggar, SubjectD9341
add: Giant lizards are now capable of breaking structures and their combat capabilities are slightly enhanced. Several other tweaks and minor additions have been made, such as wound overlays and tongue flickering.
add: After a random amount of time, a 2nd wave of lizards might pop up in the jungles if their numbers are too low.
fix: Giant lizards no longer become paralyzed after getting disarmed, nor will they randomly become paralyzed due to weird combat logic.
/:cl:


cm13-github:

cm13-github:

code/modules/mob/living/simple_animal/hostile/retaliate/giant_lizard.dm
```diff
@@ -345,15 +488,20 @@
...
-			food_target = food
-			stance = HOSTILE_STANCE_ALERT
-			stop_automated_movement = TRUE
-			MoveTo(food_target)
-			break
+			var/is_meat = FALSE
+			for(var/datum/reagent/nutriment/meat/meat in food.reagents.reagent_list)
+				if(istype(meat))
+					is_meat = TRUE
+					break
```
harryob, line 497: iterating through a list like this already typechecks, so is redundant. but you can also just do this

```suggestion
			var/is_meat = locate(/datum/reagent/nutriment/meat) in food.reagents.reagent_list
```

code/modules/mob/living/simple_animal/hostile/retaliate/giant_lizard.dm
```diff
@@ -444,9 +623,30 @@
...
+	AttackingTarget(target)
+	//turf attacks are missed attacks so it should just be a minor penalty to attack delay
+	next_move = isturf(target) ? world.time + 4 : world.time + 8
+
+/mob/living/simple_animal/hostile/retaliate/giant_lizard/DestroySurroundings()
+	if(prob(break_stuff_probability))
+		for(var/obj/structure/obstacle in view(1, src))
+			if(is_type_in_list(obstacle, destruction_targets))
+				AttackingTarget(obstacle)
+				return
```
harryob, line 635:
```suggestion
	if(!prob(break_stuff_probability))
		return

	for(var/obj/structure/obstacle in view(1, src))
		if(is_type_in_list(obstacle, destruction_targets))
			AttackingTarget(obstacle)
			return
```

harryob (changes requested):
so sorry did forget about this one

harryob (approved):

Issue #6842, opened 2024-08-01
Crawling whilst being Fireman carried


MistChristmas:
### Testmerges

#6334 #6071 #5829 #6191 #6650 #5265

### Description of the bug

Humans can crawl whilst being fire-man carried.

### What's the difference with what should have happened?

They shouldn't be able to crawl on the ground if they're lifted off of it.

### How do we reproduce this bug?

1.Fire-man carry a marine
2. Have the marine try to crawl
3. The crawl works for some reason
...

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #7039, opened 2024-08-25
Stops Crawling whilst being grabbed. Adds a notice as well.

MistChristmas:
# About the pull request
Addresses #6842
Stops crawling if you're being aggressive grabbed or grabbed by a xeno.

Meaning people being carried or aggressively grabbed can't crawl and look silly whilst you're carrying them or be mildly annoying whilst you're not.

This previously also let you crawl if a xeno grabbed you. Which seems silly but actually useful because they'd just shove you down and that would stop you moving anyway.

Fixes an oversight on crawling. Doesn't let you crawl away whilst a xeno has you. Doesn't let you crawl whilst you're being fireman carried.
<details>

</details>

# Changelog
:cl: MistChristmas
fix: Stops crawling whilst grabbed
/:cl:

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -148,7 +148,7 @@
 		return
 	if(living_mob.body_position == LYING_DOWN && !living_mob.can_crawl)
 		return
-	if(living_mob.body_position == LYING_DOWN && mob.pulledby && mob.pulledby.grab_level >= GRAB_AGGRESSIVE || living_mob.body_position == LYING_DOWN && isxeno(mob.pulledby))
+	if(living_mob.body_position == LYING_DOWN && (mob.pulledby?.grab_level >= GRAB_AGGRESSIVE || isxeno(mob.pulledby)))
 		next_movement = world.time + 20 //Good Idea
 		to_chat(src, SPAN_NOTICE("You can not crawl right now."))
 		return
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -146,7 +146,9 @@
 
 	if(!(living_mob.mobility_flags & MOBILITY_MOVE))
 		return
-	if(living_mob.body_position == LYING_DOWN && !living_mob.can_crawl || living_mob.body_position == LYING_DOWN && mob.pulledby)
+	if(living_mob.body_position == LYING_DOWN && !living_mob.can_crawl)
+		return
+	if(living_mob.body_position == LYING_DOWN && mob.pulledby)
 		next_movement = world.time + 20 //Good Idea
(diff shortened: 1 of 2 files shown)


code/modules/mob/mob_movement.dm
```diff
@@ -148,6 +148,10 @@
 		return
 	if(living_mob.body_position == LYING_DOWN && !living_mob.can_crawl)
 		return
+	if(living_mob.body_position == LYING_DOWN && mob.pulledby)
+		next_movement = world.time + 20 //Good Idea
+		to_chat(usr, SPAN_NOTICE("You can not crawl right now."))
```
kiVts, line 153:
```suggestion
		to_chat(src, SPAN_NOTICE("You can not crawl right now."))
```
dont use usr, user is the answer. but you can also just use src.

kiVts (changes requested):

MistChristmas:
Changed it from any grab to aggressive or above grab, or a xeno grab.

kiVts (changes requested):
in that case

kiVts (approved):

zzzmike:
the code as it stands now doesn't fix anything... you already can't move while in an aggressive grab, it prompts you to do a moving_resist as per code here 
image

zzzmike:
the bug report that this is based off of is not a bug from my testing, the person being carried and trying to move in any capacity including crawl does the moving_resist as normal, they only move if the resist rng passes, which is the right behavior

image

MistChristmas:
> the bug report that this is based off of is not a bug from my testing, the person being carried and trying to move in any capacity including crawl does the moving_resist as normal, they only move if the resist rng passes, which is the right behavior, because you're resisting out of the grab, you're no longer grabbed if the rng passes (you're on the ground), and at that point you should be allowed to move however you want to
> 

Weird I remember making the report because specifically one person was crawling whilst being in my carry without breaking my grip. Causing it to look like they're lagging behind whilst moving and being carried. Maybe there's more to it to pull it off. But yeah my testing concurs that you're right and I'm dumb.

MistChristmas:
Binning this for now unless I can recreate the initial issue that triggered this. Probably not.

zzzmike:
> > the bug report that this is based off of is not a bug from my testing, the person being carried and trying to move in any capacity including crawl does the moving_resist as normal, they only move if the resist rng passes, which is the right behavior, because you're resisting out of the grab, you're no longer grabbed if the rng passes (you're on the ground), and at that point you should be allowed to move however you want to
> 
> Weird I remember making the report because specifically one person was crawling whilst being in my carry without breaking my grip. Causing it to look like they're lagging behind whilst moving and being carried. Maybe there's more to it to pull it off. But yeah my testing concurs that you're right and I'm dumb.

No worries, maybe something in the code changed since the report. I just tested this because I fireman carry a lot and I was curious.

PR #7065, opened 2024-08-30
SG falloff rebalance and far sight removal

Git-Nivrak:
# About the pull request
This PR:
1. Removes far-sight
2. Increases falloff to 4 (from 1)
3. Increases effective range to 7 (from 0) - This means falloff starts counting at the 7th tile (edge of the screen)
4. Needs testing

Getting hit by bullets offscreen as well as dealing no damage on your own screen is rather annoying.
On a more serious note SGs are currently encouraged to go as close as possible in order to actually deal damage properly which is in contrast to the purpose of the role of providing suppressing fire from the back (AP bullets for example losing around 50% of their damage at the edge of the screen)
As for far-sight that was a rubber-band to try to fix the above (SGs going melee range) but just ended up being annoying to play against.
<details>

</details>

# Changelog
:cl:
balance: Removed far-sight from SG sights
balance: Increased smartgun bullet falloff to 4 from 1
balance: Increased smartgun bullet effective range to 7 from 0
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/smartgun.dm
+++ b/code/modules/projectiles/guns/smartgun.dm
@@ -18,7 +18,6 @@
 	force = 20
 	wield_delay = WIELD_DELAY_FAST
 	aim_slowdown = SLOWDOWN_ADS_SPECIALIST
-	effective_range_max = 7
 	var/obj/item/smartgun_battery/battery = null
 	/// Whether the smartgun drains the battery (Ignored if requires_battery is false)
 	var/requires_power = TRUE
--- a/code/modules/projectiles/guns/smartgun.dm
+++ b/code/modules/projectiles/guns/smartgun.dm
@@ -18,6 +18,7 @@
 	force = 20
 	wield_delay = WIELD_DELAY_FAST
 	aim_slowdown = SLOWDOWN_ADS_SPECIALIST
+	effective_range_max = 7
 	var/obj/item/smartgun_battery/battery = null
 	/// Whether the smartgun drains the battery (Ignored if requires_battery is false)
 	var/requires_power = TRUE
--- a/code/datums/ammo/bullet/special_ammo.dm
+++ b/code/datums/ammo/bullet/special_ammo.dm
@@ -9,6 +9,7 @@
 	icon_state = "redbullet"
 	flags_ammo_behavior = AMMO_BALLISTIC
 
+	damage_falloff = DAMAGE_FALLOFF_TIER_7
 	max_range = 12
 	accuracy = HIT_ACCURACY_TIER_4
 	damage = 30
@@ -31,6 +32,7 @@
 	shrapnel_chance = SHRAPNEL_CHANCE_TIER_7
 	accurate_range = 32
 	accuracy = HIT_ACCURACY_TIER_3
+	damage_falloff = 0
 	damage = 40
 	penetration = 0
 
--- a/code/datums/ammo/bullet/special_ammo.dm
+++ b/code/datums/ammo/bullet/special_ammo.dm
@@ -14,6 +14,7 @@
 	accuracy = HIT_ACCURACY_TIER_4
 	damage = 30
 	penetration = 0
+	effective_range_max = 7
 
 /datum/ammo/bullet/smartgun/armor_piercing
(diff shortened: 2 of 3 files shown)


TheManWithNoHands:
Cant hurt to TM.
But losing farsigth is a giant downside.
The intel from that is far to valubel if your behind marine lines.

Changing falloff definitly is a good idea, but getting rid of farsigth will cause SG to go closer to the first firing line. So two steps forward, one back.
Rather adjust the fall off so bullets are useless in the extra range farsigth gives, so its mainly for combat intel.

cuberound:
code wise you can do a better job getting rid of /datum/action/item_action/m56_goggles/far_sight, as it is only used there so getting rid of it properly would be nice if it is to be remvoed

CABSHEP:
>But losing farsigth is a giant downside.
>The intel from that is far to valubel if your behind marine lines.

Just get binocs, lol.

Choolemal:
With the decrease of vision and the emphases on closer combat I'd increase the fall slightly further but increase damage.

BasilHerb:
Farsight being removed isn't that big a debuff... As CAB said, literally just use binoculars. One tile of sight is miniscule compared to the SG finally not being useless at-range. Thinking that if you lose one tile of sight, it means SGs will 'frontline' more is farcical. If anything, the falloff was the biggest contributor to frontlining by far.  

Knowing that 20 AP damage -> 12 AP damage (almost HALF) at 8 tiles means that any SG with a brain will move closer to actually deal damage, which increases the risk of frontlining. (This system is what we have right now.)

With this PR, there is NO incentive for SGs to frontline, besides hellpushing. This negates frontlining completely- It does not increase the risk, it completely erases any incentives for it. This PR does not emphasis CQC. This PR emphasises the OPPOSITE of CQC. What emphasises CQC is the strange falloff system we currently have, where you do peanuts damage unless you literally PB xenos in the face.

The biggest risk I suppose, would be IFF stacking being more of an issue now (4 AP SG IFF Stacks, focusing fire on one xeno = 2.6 AP MK2s worth of firepower with this PR) But that's what TMing is for. I suspect that's taking it to the extreme, given in reality SGs usually kill themselves like lemmings instead of sticking together.
I'm probably guessing damage will be nerfed by a bit, or whatever- But overall, I think this PR is good for SGs.

hislittlecuzingames:
If sg are currently being encouraged to get close, and you believe it to be antithetical to the role, why did you remove far sight capability? It's only like 2 tiles in each direction if I'm not mistaken. 

With the way Marines seem to be getting rolled recently, I'm not sure this makes complete sense. 
But you are increasing the range of the smart gun?

TheManWithNoHands:
> With the way Marines seem to be getting rolled recently, I'm not sure this makes complete sense. But you are increasing the range of the smart gun?

Smartgun range is kinda fucked because falloff starts on the first tile Iirc. Making the gun pretty shit at long range.



> Just get binocs, lol.

Your comparing a omni directional sigth increase, that is a pssive toggle, during wich you can operate your gun fully.
With a one directional sigth increase thats an active ability during wich you cant do anything but watch and strave, normal movment breaks said ability.
Those are very much not the same, and binos do not replace it at all.
They are an passive ability that has a use in the middle of combat. Binos are not. Dont pretend like they because you got no other argument.

realforest2001 (changes requested):
As said by @cuberound need to delete the definition of `/datum/action/item_action/m56_goggles/far_sight` as it's used nowhere else.

Git-Nivrak:
Originally planned to leave it in for whiteout but forgot which is why it was not deleted - Fixed now

code/modules/projectiles/guns/smartgun.dm
```diff
@@ -18,6 +18,7 @@
 	force = 20
 	wield_delay = WIELD_DELAY_FAST
 	aim_slowdown = SLOWDOWN_ADS_SPECIALIST
+	effective_range_max = 7
```
kiVts, line 21: add this to the projectile not the gun

kiVts (changes requested):

kiVts (approved):

kiVts:
Overall it addressed the issues SG had pretty well

PR #7070, opened 2024-08-31
Removes the option to produce OT rockets

Mister-moon1:
# About the pull request
As the title says, removes the option to create custom rockets utilising OT

# Explain why it's good for the game
For the longest time there has been an issue with both OT and research due to the power they withhold, the main issue with OT though that many have addressed is the fact of the ability to create custom rockets. It has almost become standard to have an infinite supply of insta gib rockets funnelled to FOB to be used until sadar kicks the bucket.

In no sense of the game is it fun to see a sadar, begin to act to avoid a punish yet have yourself and anyone near you gibbed after a rocket hit 3 tiles away from you, mind you normal rockets would have a quarter second stun and you're back on your feet.

Attempts to balance have been made with very little success, and so I feel until a proper solution is formed, OT rockets should be removed from standard play for the time being.

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl:
balance: removed the ability to create Custom rockets in standard play
/:cl:

PR #7072, opened 2024-08-31
Option 2 for OT rockets

Mister-moon1:
# About the pull request
Reduces overall effectiveness of OT rockets to be more in line with what is already available to the demo spec

reasons stated in https://github.com/cmss13-devs/cmss13/pull/7070

<details>

</details>

# Changelog

:cl:
balance: reduced overall effectiveness of OT rockets to be a minor amount stronger than standard rockets available
balance: reduced metal cost by 3750 for both components of the rocket, now needing 3 metal sheets total to make 1 set
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/warhead.dm
+++ b/code/game/objects/items/explosives/warhead.dm
@@ -9,11 +9,11 @@
 	name = "84mm rocket warhead"
 	desc = "A custom warhead meant for 84mm rocket shells."
 	icon_state = "warhead_rocket"
-	max_container_volume = 210
+	max_container_volume = 180
 	allow_star_shape = FALSE
 	matter = list("metal" = 11250) //3 sheets
-	reaction_limits = list( "max_ex_power" = 240, "base_ex_falloff" = 90,"max_ex_shards" = 64,
-							"max_fire_rad" = 6, "max_fire_int" = 40, "max_fire_dur" = 48,
+	reaction_limits = list( "max_ex_power" = 180, "base_ex_falloff" = 90,"max_ex_shards" = 64,
+							"max_fire_rad" = 3, "max_fire_int" = 30, "max_fire_dur" = 36,
 							"min_fire_rad" = 2, "min_fire_int" = 4, "min_fire_dur" = 5
 	)
 	has_blast_wave_dampener = TRUE
--- a/code/modules/projectiles/magazines/specialist.dm
+++ b/code/modules/projectiles/magazines/specialist.dm
@@ -238,7 +238,7 @@
 	desc = "An 84mm custom rocket."
 	icon_state = "custom_rocket"
 	default_ammo = /datum/ammo/rocket/custom
-	matter = list("metal" = 7500) //2 sheets
+	matter = list("metal" = 3750) //1 sheet
 	var/obj/item/explosive/warhead/rocket/warhead
 	var/obj/item/reagent_container/glass/fuel
 	var/fuel_requirement = 60


Skatuk:
So this from what I can see makes it worse than a maxcap grenade the m15 grenade variant, better than a m40 grenade variant by 5 units. Rocket power is nerfed by almost 50%, this doesn't seem like a solid change but I guess we'll see if it gets merged.

I comparted the information from the wiki which is pretty damn outdated considering on the wiki it states max power for OT Rockets being 300 but in the original code it being 240, and with the PR nerf it now stands at 180.

Mister-moon1:
This is still a direct upgrade to HE rockets, because of the low thresholds, the stats should be able to be maxed out and it will make up for the lack of telegraphing that other forms of OT have

TemptedGhost:
this would mean that M15 grenades are more superior, I would even reach with enough effort. M40's can be as same balance power as the " standard " if you want this the norm. i would recommend nerfing the incen range of OT rocket by 2 tiles so it's 1 tile bigger of a WP rocket

Mister-moon1:
If the explosive power is increased then no change would have been done at all, these rockets capable of making opening and dealing heavy damage to some t3s and most t2s are mass produce-able and almost infinite depending how patient the medical chemline is, this is NOT weak compared to other rockets available, the grenades have their own flaws to why they are not seen as an issue compared to rockets

Rockets have 0 telegraphing, you only find out its OT once you have been hit, thus it is bad design to have it be considerably stronger to be able to singlehandedly kill targets, a reminder that sadar is supposed to stun priority targets for a counter, not kill them outright

TemptedGhost:
But OT is supposed to be an upgrade for Sadars and every category. If you are just nerfing OT rockets down so much that it becomes the same result as regular supply. What's the point of upgrading a regular old rocket if the results gonna be the same?

Mister-moon1:
If you think that it is only the fire range then you are unaware of the ability to quite literally 1 shot praes, boilers, all t2s and all t1s with explosive power

Diegoflores31:
making 2 PRs one "nerfing" by just doing damage changes and the other outright removing a feature that has been on the game for more than 3 YEARS with very little reasoning . ngl this PR looks like just salt

Mister-moon1:
Or you could look around what people have to say with OT rockets and their experiences with them, how they are unfun and unfair to fight against, how they can be mass produced before first drop, how there is quite literally 0 counter play because you do not know what they have loaded or whats coming at you

right now you either get gibbed by a maxcap, or hit with a lake of fire that is impossible to escape from, this isnt even just t2s, this is everything except rav, crusher and queen

Mister-moon1:
if you do not believe said sentence then watch this clip https://cdn.discordapp.com/attachments/1266712926140629022/1266714771001053225/CM-SS13_-_USS_Almayer_2024-07-27_03-19-23.mp4?ex=66d44c0f&is=66d2fa8f&hm=d8da20d56543129d40cd4a98dee0c580ea459da68e11de4e8f306a35a1b79091&

Diegoflores31:
> Or you could look around what people have to say with OT rockets and their experiences with them, how they are unfun and unfair to fight against, how they can be mass produced before first drop, how there is quite literally 0 counter play because you do not know what they have loaded or whats coming at you
> 
> right now you either get gibbed by a maxcap, or hit with a lake of fire that is impossible to escape from, this isnt even just t2s, this is everything except rav, crusher and queen
> 
> If you do not believe this, then feel free to watch this clip https://cdn.discordapp.com/attachments/1266712926140629022/1266714771001053225/CM-SS13_-_USS_Almayer_2024-07-27_03-19-23.mp4?ex=66d44c0f&is=66d2fa8f&hm=d8da20d56543129d40cd4a98dee0c580ea459da68e11de4e8f306a35a1b79091&

It looks more like you are complaining about the concept of SADAR , OT rockets do have a counterplay , either get away from sadar aim or get closer and let the blast kill him too .
You cant mass produce "maxcap" rockets . You can only make 4 , 8 if you are willing to make them non maxxed

Your PR is making it so literally a SADAR rocket does less than a burst of Mk2 or a sniper flak

TheManWithNoHands:
Dont fall for the trap of just reducing dmg. Thats the lazy devs balance aproach.
Rather reduce falloff or force enable falloff dampening (it cuts falloff hard off Iirc. Migth be dead wrong though).
In general we see falloff dampening to little of.

Ot SADAR biggest problem is how big the explosion is. Compare the size with all the other SADAR rockets, even SADARs HE is half that size, if not less. 
SADARs role is about causing incredibly high singel target damage. Yes, He or Incen can do AoE, but they are still better for Single target then AoE. They can just tecnicly do that. Also nobody picks incen SADAR, dont pretend.
OT rockets are very very very good for multiple targets and break that role. They arent for single target dmg.

If you keep that in mind while rebalancing the rockets it should be fine, and OTs can still make strong rockets. Just that they dont fuck over the hole hive in one go.

Also, i feel like you havent tested this on a private server.

Mister-moon1:
the stats of the OT rocket now is pretty similar to HE although definitely stronger, it has an increased falloff so it will be confined area rather than a screenwide stun ideally, in simple terms if maxed then it should be a stronger HE without the capabilities of wiping everything nearby

Diegoflores31:
current changes makes OT rockets unable to even kill a warrior on direct hit at maxcap

Mister-moon1:
That is very much the intended effect, not insta killing everything, you may disagree but as it turns out getting insta killed from a non telegraphed attack is not fun

TheManWithNoHands:
> it has an increased falloff so it will be confined area rather than a screenwide stun ideally

No, it doesnt. You changed of the blast only the dmg cap, but thats its. Falloff is unchanged.
grafik

Keep damage and duration from before, but increase falloff

Mister-moon1:
the falloff is faster because its a lower number, it doesnt need to be changed because its already got less range

you are also proposing that on direct hit it do the same damage, which is a gib, which is what im trying to avoid, HE rockets are more than able to make opportunities for killing a possible 13/16 xenos, without strains taken into effect, in the traditional sadar way of having followup, sadar is the 1 OT item with 0 telegraphing and so should under no circumstanced remove someone from the round due to it alone.

If you believe that it is not worth making rockets anymore, good on you, you are diversifying the OT items distributed to marines

TheManWithNoHands:
> the falloff is faster because its a lower number, it doesnt need to be changed because its already got less range

I dont think you know how falloff works. The lower the number the larger is the AoE.
Falloff is subtracted from the dmg for every tile the explosion travels. Low falloff, large explosion. High falloff, small explosion.

Diegoflores31:
image
this is also making any kind of fire rocket completely unviable

cuberound:
also if you want ot rockets to stun not kill then reduce metal price alongside the caps

foxymacko-real:
man if we are removing OT rockets from this game cause they kill xenos why dont we remove guns and slashing and instead just make everyone roleplay it, *shoots you. smh.
dum nothing burger PR
mikhail out.

Nanu308:
Going to be looking into these two PR's this or next week, as i do agree OT rockets are too strong in particular. #7217 #7072

Nanu308:
As the other PR has been merged i will close this. If you feel OT has been adversly affected, or is still too strong, you can make a new PR in a weeks time.

PR #7077, opened 2024-09-01
Sets up structs (pseudo-structs) as a replacement for datums for si…

boskoramen:
# About the pull request

- Instead of creating and using datums for storing data, can just initialize simple lists
- Syntax is pretty clunky/verbose, open to suggestions for how to improve it for readability and elegance
- Controller logic mostly pulled from the global vars controller

Pros:
- Do not need to create and delete datums
- Straightforward interface

Cons:
- Verbose syntax
- Debugging can be annoying since there is a lot of C macro stuff under the hood

This was a fun project to do over the weekend, wanted to see if this was doable after found a mini use case for unit tests I was planning to write. Can't really think of anything that can directly use this except for launch metadata at the moment, curious to see if people can think of other stuff.
<details>

Tested locally with warrior throw, human throw, and explosions. Explosion lag caused some issues with indexing the list but looking at variables in debugger not sure if this is an issue with the structs or with throw logic and lag.

</details>

# Changelog
:cl: TheDonkified
code: Added pseudo structs construct for storage of simple data
/:cl:


Zonespace27 (commented):
Interesting concept, but I have some concerns.

> Do not need to create and delete datums

This isn't especially an issue. The cost of creating or deleting a datum is insignificant, and in cases like `launch_metadata`, the datum would be automatically GCed once it goes out of scope.

> Straightforward interface

Somewhat. However, there lie two issues in this. First, this is a code feature that would be unique to CM, making it more complicated for coders from other codebases to understand. Secondly, it seems you lose the ability to cast member variables in a struct, as opposed to the more traditional datum.

> Debugging can be annoying since there is a lot of C macro stuff under the hood

This is absolutely a massive concern. This level of macro abuse makes it much harder to debug things.

-----

Overall, the concept is interesting but I don't see what it brings to the table over regular datums. I'd like to hear more justification for this PR.

boskoramen:
Think main benefit is memory usage, though doing some profiling don't think our datum count will scale enough for the memory benefit to matter.

Gonna close this PR, was a fun exercise but my original understanding of datum init/delete cost was misguided.

PR #7095, opened 2024-09-03
Adds text to speech

Watermelon914:
# About the pull request

As the title says. Adds text to speech. Will be testmerged for a bit to gauge player feedback and see what people think.

TTS can be disabled individually if you don't like it via your preferences menu within the Preferences -> Settings section. There is an option to switch to blips only (don't starve voices) and an option to completely turn it off for yourself. Additionally, TTS volume can be configured alongside ambience and sfx volume.

You can change your voice and pitch via the preferences menu individually for your human character, synth character, and xenomorph character.

Yautja do not have TTS at this time, pending proper filter implementation to make them sound yautja-like.

Can significantly help with communication. For now, it's unavailable over radio due to complications but it'll be available for hivemind xenomorphs, limited to the queen and leaders only for now.

This can be extended to radio messages in some capacity sometime later if people like it enough to keep it but this is mostly just testing the waters.

I've tested it locally with a local key and it seems to work all fine.
<details>

</details>

# Changelog
:cl:
add: Adds text to speech for most characters. Modify your voice within your preference menu. Can be turned off via preferences menu or by setting TTS volume to 0.
/:cl:


Kitsunemitsu:
Small bug, multicast says everything 4 times.

TheGamerdk:
Working Joes say their voice-lines twice.

Also, what's the performance impact of this?

Watermelon914:
> Working Joes say their voice-lines twice.
> 
> Also, what's the performance impact of this?

Performance impact should generally be minimal, it's basically just playing sounds around people, though they are dynamically generated sounds.

Some things need to be examined tho, and you might experience clientside lag if you have a weak connection since it basically sends you over the sound file to play.

Watermelon914:
after testing and feedback, this doesn't seem to be worthwhile to pursue

ThatGuyThe:
Personally, i quite enjoyed playing with it, id say it just needs a few tweaks and the ability to opt out for the 500 people complaining that turning down the volume to 0 does not disable it

Skatuk:
It doesn’t fit into the game because it doesn’t captivate the tone that your trying to portray and it just makes you look goofy and pretty much a joke when being listened to.

PR #7107, opened 2024-09-04
Removes hugger strain: Watcher 

TheManWithNoHands:
# About the pull request

* Removes the watcher strain

In #6237 watcher was nerfed. And they even snuck in some nerf to the watcher they only listed in the changelog and nowhere else. Being that watcher could no longer move while being zoomed out.. for some reason.
Not like the nerf that they die off weeds wasnt enough.

Watcher strain was nerfed because of meta level shit. As a temporary solution until some fixes are implemented to prevent meta lv abuse.
To bad that temporary fix sucks balls.

Watcher strain is barley a strain rigth now. And because its nerfs cant be reverted currently, its best removed for now.
Its sad that the huggers that sneak around in the almayer are a thing of the past, but some things just has to end it seems.

<details>

</details>

# Changelog
:cl: TheManWithNoHands
del: Removes Watcher as strain choice for Hugger
/:cl:


Unknownity:
"If watcher doesnt work, remove it." so delete the code, not just disable it and leave the strain rot in the code.

TheManWithNoHands:
> "If watcher doesnt work, remove it." so delete the code, not just disable it and leave the strain rot in the code.

Ok. I like code preservation but if you say this is better.
But then again, easy to revert PRs. So it was a bit silly of me.

Diegoflores31:
change tittle

Hopekz:
As the original author of the watcher I support this PR because this codebase seems to have ruined the watcher experience while I was gone I don't even know how it happened the way it did, but it did. 

I actually brought this up to the project lead a while ago but they said that this new varient fit their idea of where the game should be. I don't agree but we are allowed to have our different opinions.
image


Was really fun to experience the watcher before it was changed and I know a lot of people that I've talked to that agree but what it is at this current moment should not exist in my opinion. Mainly just sad to see that many others in the future won't be able to experience the original mechanic because of the amount of fun it brought to me and others.

It makes perfect sense to me that this new experience for the watcher that was brought on by changes is something that people do not want to keep around and I do not take offense to this removal because this hasn't been the watcher that I added to the codebase for a while now.

Thank you!

TheManWithNoHands:
> would be better to have it disabled as a choice vs code removed imo for a potential future rework

Other comment said fully remove it, i planned to just disable it and keep the code in. 👋😩🫱
But to revert a pr is literally done by pressing a single button. So not the biggest problem.

But yeah. Watcher is barley a strain rigth now. Old watcher can be implemented again when the ghost reatrictions are implementet. Then the meta abuse wont be a problem.

Hopekz:
What really urks me in this situation is that is if you search "watcher" on the CM discord and read the feedback to the new watcher  you will just encounter streams of people giving feedback and it is all just disappointment on what it has become with people saying how they had a lot of fun with the feature in the past.

It is easy to get "something isn't right with these changes" feedback in this case but nothing seems to be done about it. 

If I had any say in this project and I decided that this mechanic needed to change I would bare minimum try to at least approach changes in a different way that fits the mechanic more after seeing the reception these changes have gotten.

TheManWithNoHands:
> It is easy to get "something isn't right with these changes" feedback in this case but nothing seems to be done about it.

The main problem is just that huggers are already a ghost role and vunriable to meta abuse. Watcher strain just exegrated that.
It is possible to creat a new hugger strain, but i highly doubt its currently possible to make watcher work.
Legit, if someone wants to make watcher work, dont. Rather make a new strain at this point. 
Watcher core mecanics are just too abusable. Fun, but because of the assholes that cant behave, we cant have said fun.

Wintermote:
First time I played watcher hugger, I assumed I'd be able to run around off weeds and see further so I could size up attacking someone.

But no, it's just a regular hugger except you can't hide. It's completely pointless.

Drulikar:
If you don't like the strain, then don't use it. Removal is unnecessary and my previous statement that is conveniently reposted here are still my thoughts. I am open to suggestions on how you can make it more interesting w/o the original issues however.

PR #7121, opened 2024-09-05
Automatic emotes are no longer are affected by emote cooldown (baton edition for now)

zzzmike:
# About the pull request

The intent behind this is to not have automatic emotes trigger, and be affected by, emote cooldown. Starting with automatic baton emote for now. Might change this if someone points out a better way to fix this for all automatic emotes.

Emote cooldown is designed to stop players spamming them and being LRP, it should never affect automatic emotes. It doesn't make sense that automatic emotes can't happen because you did a manual one a few seconds ago.

<details>

</details>

# Changelog
:cl:
sounddel: automatic emotes are not affected by emote cooldown (baton edition for now)
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/stunbaton.dm
+++ b/code/game/objects/items/weapons/stunbaton.dm
@@ -150,7 +150,9 @@
 	//stun effects
 
 	if(!isyautja(human_target) && !isxeno(human_target)) //Xenos and Predators are IMMUNE to all baton stuns.
+		ADD_TRAIT(human_target, TRAIT_EMOTE_CD_EXEMPT, "Batoned")
 		human_target.emote("pain")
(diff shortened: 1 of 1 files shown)


harryob (changes requested):
non-intentional emotes shouldn't impacted by cooldown anyway, we shouldn't have to do funny trait stuff like this, check `/mob/proc/emote`

zzzmike:
> non-intentional emotes shouldn't impacted by cooldown anyway, we shouldn't have to do funny trait stuff like this, check `/mob/proc/emote`

it's bugged, throws the 'you just did an emote' error when getting hit by a baton multiple times quickly @harryob

zzzmike:
it's bugged then, throws a 'you just did an emote' error when getting hit by baton multiple times quickly

harryob:
seems to be stale, feel free to make an issue report if you're not interested in this and i can try to take a look more thoroughly at some point

PR #7087, opened 2024-09-02
Psychic Radiance for humans

realforest2001:
# About the pull request

Well, they can talk 1:1. Why not this.

# Explain why it's good for the game
# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
add: Added psychic radiance for humans.
fix: Humans cannot psychic speak while dead.
add: Psychic whisper and radiance now only shows to ghosts who are in the target radii.
/:cl:

PR #7123, opened 2024-09-06
Adds the preference to trigger xeno/human abilities with right click instead of middle click or shift click

Watermelon914:
# About the pull request

As the title says. Adds a new preference option that can be toggled between middle clicking and shift clicking. When enabled, if you have a xeno ability active, the right mouse button will trigger the ability where you click.

This'll disable the context menu when pressing right click if the ability is active, but you can hold shift before right clicking to re-enable the context menu if you need it for some specific purpose without having to deactivate the ability or switch your preferences.

Right mouse button is just a lot more convenient for me and I'm sure this'll be useful for other players too. This is primarily a quality of life change, the preference to use the middle mouse button or the left mouse button with shift click will still exist.

It's a big problem for me since the middle mouse button on my mouse is somewhat faulty and doesn't always register clicks correctly. Shiftclick is also really awkward for me to use because I don't want to have to hold down multiple buttons to use my abilities. Adding a new option on the rmb makes sense here.

Tested on my local and seems to work fine. You can rightclick normally when the ability is disabled or if you have the preference off, but rightclick to proc ability works fine.

<details>

</details>

# Changelog

:cl:
qol: Adds a new preference option for activating xeno abilities: Right click. Prevents the context menu from opening whenever you have an ability active, and the ability can then be triggered with right click. Context menu can still be forced open by holding shift before right clicking, even if the ability is active.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_abilities.dm
+++ b/code/modules/mob/living/carbon/human/human_abilities.dm
@@ -144,12 +144,12 @@
 
 	var/list/target_list = list()
 	for(var/mob/living/carbon/possible_target in view(7, human_owner))
-		if(possible_target == human_owner || !possible_target.client)
+		if(possible_target == human_owner || !possible_target.client) 
 			continue
 		target_list += possible_target
 
 	var/mob/living/carbon/target_mob = tgui_input_list(human_owner, "Target", "Send a Psychic Whisper to whom?", target_list, theme = "hive_status")
-	if(!target_mob)
+	if(!target_mob) 
 		return
 
 	human_owner.psychic_whisper(target_mob)
--- a/code/modules/mob/living/carbon/human/human_abilities.dm
+++ b/code/modules/mob/living/carbon/human/human_abilities.dm
@@ -1,3 +1,12 @@
+/mob/living/carbon/human/proc/set_selected_ability(datum/action/human_action/activable/ability)
+	if(!ability)
+		selected_ability = null
+		client?.set_right_click_menu_mode(shift_only = FALSE)
+		return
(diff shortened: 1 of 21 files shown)


Mister-moon1:
can no longer point using the default shift + MMB with an ability active

Watermelon914:
fixed

code/modules/mob/living/carbon/human/human_abilities.dm
```diff
@@ -135,12 +144,12 @@
 
 	var/list/target_list = list()
 	for(var/mob/living/carbon/possible_target in view(7, human_owner))
-		if(possible_target == human_owner || !possible_target.client) 
+		if(possible_target == human_owner || !possible_target.client)
```
realforest2001, line 147:
```suggestion
		if(possible_target == human_owner || !possible_target.client) 
```

code/modules/mob/living/carbon/human/human_abilities.dm
```diff
@@ -135,12 +144,12 @@
...
 			continue
 		target_list += possible_target
 
 	var/mob/living/carbon/target_mob = tgui_input_list(human_owner, "Target", "Send a Psychic Whisper to whom?", target_list, theme = "hive_status")
-	if(!target_mob) 
+	if(!target_mob)
```
realforest2001, line 152:
```suggestion
	if(!target_mob) 
```

realforest2001 (commented):
Don't want to take TM off, these two spaces are causing a conflict for #7087

harryob (approved):

PR #7135, opened 2024-09-09
Fire stops Xeno regen (Main - Vamp)

ghostsheet:
# About the pull request

Fires now stops all forms of xenos regen, don't expect to gain health while on fire

This was supposedly an intended effect but was poorly implemented.
Now it properly checks and prevents all form of xeno regen.
Don't waste your heals on a burning xeno.

As requested seperated the changes for easier decisions:

Beserker Ravs no longer heals while on fire #7137

Weed and Pheromones regeneration no longer heals while on fire #7136

# Explain why it's good for the game

Gives good and interesting counter-play to Zerker ravs and Vamp lurkers and some other healing caste
Increases the effectiveness of the underperforming flamers without it being a number change.

Giving fire slightly more importance promotes good counterplay from both sides.
Allowing marines to counter healing xenos
While giving more importance to other xenos who can extinguish fires, like acid pillar, spitter, prae and boiler's spit and generally working together.

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

https://streamable.com/idpec6

ezgif-7-be831f7f30

</details>

# Changelog

:cl: ghostsheet
balance: Vamp Lurkers no longer heals while on fire.
balance: Healer Drone sacrifice abilities now also extinguishes the target before healing.
/:cl:

PR #7136, opened 2024-09-09
Fire stops Xeno regen (Weed&Phero)

ghostsheet:
# About the pull request

Optional changes for #7135

removes weed and phero regen while on fire

# Changelog

:cl: ghostsheet
balance: Weed and Pheromones regeneration no longer heals while on fire.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -338,7 +338,7 @@ Make sure their actual health updates immediately.*/
 
 	if(caste)
 		if(caste.innate_healing || check_weeds_for_healing())
-			if(!hive) return // can't heal if you have no hive, sorry bud
+			if(!hive || on_fire) return // can't heal if you have no hive, sorry bud
 			plasma_stored += plasma_gain * plasma_max / 100
 			if(recovery_aura)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -338,7 +338,7 @@ Make sure their actual health updates immediately.*/
 
 	if(caste)
 		if(caste.innate_healing || check_weeds_for_healing())
-			if(!hive) return // can't heal if you have no hive, sorry bud
+			if(!hive || on_fire) return // can't heal if you have no hive, sorry bud
```
Watermelon914, line 341: this will completely cut off the rest of the proc from running, which does not seem ideal - you'll prevent regular things such as ability icons updating, actions running the `life_tick`, as well as preventing pheros from being cancelled if you are out of plasma.

You're going to want to remove this early return, it makes little sense here anyways. Just add this into the above checks

harryob (changes requested):
address walter's review

Nanu308:
Making fire more in line with itself and how it acts.

If fire prove too strong as support weapon later on we can look at it more.

harryob:
pr remains stale, please reopen pr when you're able to address outstanding review

PR #7148, opened 2024-09-11
Grants the ASO the senior command headset (Same as XO)

LordJike:
# About the pull request

This pull request grants the ASO the senior command headset similarly to the XO and also allows him to dispense one from his locker, granting him access to the MP channel.

The ASO is the third in command in the ship, and carries substantial weight not only in ship-side affairs, but also in taking command of the operation when the XO is absent in CiC.

It is clear enough that the ASO carries substantial authority, to the point that it is one of the non-MP roles that has access to the brig, and a high playtime ASO gains the rank of Captain, same as the XO.

Additionally, the ASO tends to be the prime subject for delegation of appeals when the XO is too busy with their head in the op.

Despite all of this, the ASO has no access to the MP channel on their radio. I believe this should change (and perhaps, we should also change the default SOP to allow the ASO to take initiative to step in for appeals instead of the XO if the XO is unable to attend the appeal without requiring delegation. I've seen enough appeals time out and XOs be hit with DoD because they were too distracted processing the flood of information that goes on constantly in the CiC.)

Shouldn't be necessary, but feel free to look at the code in case I may have done something wrong.

# Changelog

:cl:
add: ASO now starts with a senior command headset (and thus access to the MP radio channel.) and can also get a new one from their locker.
/:cl:


HumiliatedGoblin:
It would be better to give them the IO headset, where they have all channels but only what is needed enabled on round start. 
If ASO ever needs to take command, they can simply turn the channels on their headset.

LordJike:
ASO already has access to basically every other channel, with the squad channels turned off at round start. Pretty sure it's the same headset as SOs

harryob (approved):

PR #2766, opened 2023-03-04
Firearms skills rework

carlarctg:
# About the pull request

Reworks the firearms skill.

unskilled is now 'SKILL_FIREARMS_CIVILIAN'
default is now 'SKILL_FIREARMS_TRAINED'
trained is now 'SKILL_FIREARMS_EXPERT'

- Civilian skill will allow you to use pistols, SMGs, and certain weapons that have their civilian override variable set to TRUE, such as bolt-action rifles, the ABR-40, the HG-37, or the double barrel shotguns.
- Trained skill is the same as always.
- Same with expert skill. The renames are for readability.

# Explain why it's good for the game

Civilian gun usability is horribly updated. It shouldn't slow your firerate as that makes no sense and makes guns feel terrible to use, and it shouldn't be applied in such a way that it makes pistols and SMGs unusable and the best option running around with a one handed shotgun.

Pistols and SMGs are reasonably newbie-friendly guns for civvies to know how to use, and the civilian shotguns are, well, built for them.

This paves the way for survivors to not be on the same level as marines. If this is an approved idea I can include it here.

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl:
balance: Reworks the firearms skill. Civilians can now fire pistols, SMGs, and certain other civilian weapons without penalties. Civilian gun penalties have had their firerate reduction remove and scatter increased.
/:cl:

PR #7154, opened 2024-09-13
increases bullet hit accuracy to  CL’s ES-4

deathrobotpunch:
# About the pull request
this PR increases the hit accuracy of the CL’s Sidearm, with this change the bullets it fires will have hit accuracy tier 4.

Due to CL being a pleb and having 0 firearms skill, the accuracy of the ES-4 was severely decreased causing most of your shots missing due to RNG making it almost useless when compared to pepper spray. additionally the ES-4 was advertised as being very accurate in its description when that wasn’t the case.

By the way I am completely open to change so if you have any complaints I’ll take them to heart ❤️.
<details>

</details>

# Changelog
:cl:
balance: buffed the stun subtype of rubber bullets to have hit accuracy tier 4
/:cl:

### DIFF
--- a/code/datums/ammo/bullet/pistol.dm
+++ b/code/datums/ammo/bullet/pistol.dm
@@ -94,6 +94,8 @@
 	name = "stun pistol bullet"
 	sound_override = null
 
+	accuracy = HIT_ACCURACY_TIER_4
+
 // Used by M1911, Deagle and KT-42
 /datum/ammo/bullet/pistol/heavy
 	name = "heavy pistol bullet"


deathrobotpunch:
why did it add the feature label?

Diegoflores31:
> why did it add the feature label?

after CL weapon was added , support was added to make some weapons need no weapons training to be shot . you could add that

deathrobotpunch:
> after CL weapon was added , support was added to make some weapons need no weapons training to be shot . you could add that

really? where could I find this

BOBAMAx:
> 
> really? where could I find this

civilian_usable_override from #2766, however I do not think that it has an effect on accuracy, only scatter.

zzzmike:
> why did it add the feature label?

because the first version of your PR didn't have changelog done properly - it didn't change the label after you edited it for whatever reason

deathrobotpunch:
> civilian_usable_override from #2766, however I do not think that it has an effect on accuracy, only scatter.

read the code and this is already applied to all pistols and SMG’s

Nanu308:
this is fine

kiVts (approved):

PR #7160, opened 2024-09-13
Fixes duplicate badges on marine recruiter surv

deathrobotpunch:
# About the pull request

fixes issue #7011
removes extra rank pin

<details>

</details>

# Changelog
:cl:
fix: fixes duplicate rank pins on marine recruiter survivor
/:cl:


harryob (approved):

harryob (changes requested):
wrong button, this would just cause for them to only spawn in with the incorrect pin. you can make a new proc on the parent type `/datum/eqipment_preset/survivor/uscm` which would usually just give the e2 pin, and override it here to give the e5 pin instead

PR #7178, opened 2024-09-17
Disables New Varadero from the current map pool. 

joookschad:
# About the pull request
Removes varadero from the voting pool

In it's current state, Varadero is a fundamentally broken map seemingly made from half of a old map. It suffers from instant FOB sieges, leading to just incredibly boring rounds for both sides and there are seemingly very few people who like the map. 

My last PR was denied on grounds of a rework being in progress and "just don't vote for it", but seeing as the map vote system forces people to play it when there's 2-6 trolls voting for it every round, and the person making the rework has abandoned the rework, I think it's fair to remove it from the map pool, especially now that Chances Claim is a lowpop map.

# Changelog
:cl:

del: New Varadero can no longer be voted for 

/:cl:

### DIFF
--- a/map_config/maps.txt
+++ b/map_config/maps.txt
@@ -58,6 +58,7 @@ map lv522_chances_claim
 endmap
 
 map new_varadero
+	disabled
 endmap
 
 map whiskey_outpost_v2


RainbowStalin:
I personally like the map, and its the last part of nostalgic ice we have that isn't terrible.

vero5123:
No.

Drulikar:
Situation has not changed nor has my opinion changed since what I mentioned last time. Regardless I am currently waiting on response from Nanu their thoughts atm otherwise I would just outright close this PR.

As with any PR that can't be bothered to fix anything other than removal/disable I highly encourage you to instead make an attempt at addressing whatever concerns you have with it.

PR #7185, opened 2024-09-20
Revert "fixes undroppable wrist blades"

Mister-moon1:
Reverts cmss13-devs/cmss13#7159

Causes bugs with the wristblades being dropped if knocked down

# Changelog
:cl:
code: reverts #7159
/:cl:


harryob (approved):
though i don't think this pr actually broke anything more, i also don't think it fixed anything particularly. so

PR #7195, opened 2024-09-22
predator buffs

irakligabunia:
# About the pull request

It's a nudge in the right direction for the predator WL, as they've been severely underpowered ever since the "glass cannon" idea. The glass cannon idea is simply too weird for us, as we were supposed to give more damage but the rework got abandoned and we were stuck with low armor. In it's current state, any above average marine can very easily kill a predator. This change will hopefully turn them into something more than a loot pinata.

As apart of the glass cannon predator rework, predators had their armor reduced; although the rework got abandoned and we were stuck with low survivability and low damage. This PR is good for the game as it'll make the fights last longer, the predators will cornerslime (and slime in general) less and make predators pose an actual threat instead of being easier to kill than xenomorphs
I also increased the armor of thralls because they've been completely abandoned as one marine with 2 bullets could easily ruin their round. Now it takes around 3-4 🗡️ 
I fully intend to revisit this when the xeno buff (to preds) comes along, @Git-Nivrak get on ittttt
I'll be closely maintaining an eye in case of a testmerge.

its a numberchange

# Changelog

:cl: stalkerino
balance: Buffed predator armor, thrall armor,  bullet resistance has been raised. Heavy clan armor has slightly less slowdown (by 1 tier)
/:cl:

### DIFF
--- a/code/modules/cm_preds/yaut_items.dm
+++ b/code/modules/cm_preds/yaut_items.dm
@@ -123,7 +123,7 @@
 	armor_bio = CLOTHING_ARMOR_HIGH
 	armor_rad = CLOTHING_ARMOR_HIGH
 	armor_internaldamage = CLOTHING_ARMOR_HIGH
-	slowdown = 1
+	slowdown = 0.75
 	var/speed_timer = 0
 	item_state_slots = list(WEAR_JACKET = "fullarmor")
 	allowed = list(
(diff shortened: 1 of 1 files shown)


Diegoflores31:
Guys i am buffing yautja because i got killed by a marine and this is not ok because yautja should not be able to die 🤓

irakligabunia:
> Guys i am buffing yautja because i got killed by a marine and this is not ok because yautja should not be able to die 🤓

be quiet chud flores rule #7 of github says only good players can comment on PRs

Gregory123211 (commented):
I would enjoy this

harryob (approved):

PR #7197, opened 2024-09-22
Sentinel buffs

GoldenDarkness55:
# About the pull request

Revives https://github.com/cmss13-devs/cmss13/pull/4023 - was never told what the problem with it was and the sentinel issues are undeniably still present. Would rather not leave a rework that I started unfinished.

Follows-up on my old sentinel rework to buff its ability to 1v1 and address a bunch of other issues.
The initial design goal was to turn the annoying bodyblocker into an offensive support caste with strong abilities for a t1 - but unable to make good use of them by itself, thus heavily reliant on teamwork.
I initially worried about them being too strong and overdid the necessity for perfect play in 1v1 scenarios, leading to them struggling too much to the point where capturing even a lone out of position marine is way too difficult.

- Hibernate ability (see videos section)

Their entire kit is about supporting and working with other caste, which requires them walking out with bulkier t2s and t3s and leads to them eating a ton of bullets that they’re ill-suited to tank.
Compounded with said teamwork often being unreliable (either because the damage was too high for sent to follow, the players just didn’t mesh very well or the spit was blocked by xenos) and their super slow healing it can be frustrating to play when on such a bad streak of achieving nothing and having to rest.
Therefore the "hibernate" ability, to mitigate the downtime somewhat and better allow it to actually support the tankier frontline xenos.
Also considering sentinels are supposed to guard the hive I think it’s cool and fitting to have an ability centered on hibernating somewhere.

- Speed increase

To let and encourage them to support backliners aswell as make it easier to work near the front, getting into a position from which you can spit without getting blocked and avoiding damage since sent has has 0 armor.

- Tackle and paralyzing slash buffs 

Currently the window to tackle successfully off the slash stun is extremely tight, you’d have to start tackling the moment the marine falls down which in practice just almost never happens. Current proc timer also gives too much leeway in getting away before getting downed. Might still be too much

- Neuro

Less spammy, more impactful and easier to chain with other abilities.

- Scatter spit

The spread cone was too wide and more unreliable at hitting than intended. Technically a nerf since it won't hit as wide of an area anymore. Cooldown lowered to give more impact when things are going well and reward taking the risk of staying in combat.

<details>

https://github.com/user-attachments/assets/12946cd7-12a8-4d4d-b74f-cb4fcd396cfc

</details>

# Changelog

:cl:
add: Sentinel new ability - "Hibernate" - causes them to enter a 10 second sleep that restores their health and plasma to full. Requires weeds. 50 second cd and 150 plasma cost.
balance: Sentinel neuro spit cooldown raised 1.5 to 2s, superslow 2.5 to 4s
balance: Sentinel scatter spit scatter lowered 60 to 45, cooldown lowered 8s to 6s
balance: Sentinel paralyzing slash stuns after 3 seconds instead of 4 and lasts 2.5 seconds from 2.
balance: Sentinel speed raised from -0.8 to -1
balance: Sentinel tackle now requires 3 attempts instead of 4.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/Sentinel.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Sentinel.dm
@@ -11,15 +11,15 @@
 	xeno_explosion_resistance = XENO_EXPLOSIVE_ARMOR_TIER_1
 	armor_deflection = XENO_NO_ARMOR
 	evasion = XENO_EVASION_NONE
-	speed = XENO_SPEED_TIER_7
+	speed = XENO_SPEED_HELLHOUND
 
 	caste_desc = "A weak ranged combat alien."
 	evolves_to = list(XENO_CASTE_SPITTER)
 	deevolves_to = list("Larva")
 	acid_level = 1
 
-	tackle_min = 4
-	tackle_max = 4
+	tackle_min = 3
+	tackle_max = 3
 	tackle_chance = 50
 	tacklestrength_min = 4
 	tacklestrength_max = 4
@@ -49,6 +49,7 @@
 		/datum/action/xeno_action/activable/slowing_spit, //first macro
 		/datum/action/xeno_action/activable/scattered_spit, //second macro
 		/datum/action/xeno_action/onclick/paralyzing_slash, //third macro
+		/datum/action/xeno_action/activable/hibernate, //fourth macro
 		/datum/action/xeno_action/onclick/tacmap,
 	)
 	inherent_verbs = list(
@@ -68,7 +69,7 @@
 	// State
 	var/next_slash_buffed = FALSE
 
-#define NEURO_TOUCH_DELAY 4 SECONDS
+#define NEURO_TOUCH_DELAY 3 SECONDS
 
 /datum/behavior_delegate/sentinel_base/melee_attack_modify_damage(original_damage, mob/living/carbon/carbon_target)
 	if (!next_slash_buffed)
@@ -103,6 +104,6 @@
 #undef NEURO_TOUCH_DELAY
 
 /datum/behavior_delegate/sentinel_base/proc/paralyzing_slash(mob/living/carbon/human/human_target)
-	human_target.KnockDown(2)
-	human_target.Stun(2)
+	human_target.KnockDown(2.5)
+	human_target.Stun(2.5)
 	to_chat(human_target, SPAN_XENOHIGHDANGER("You fall over, paralyzed by the toxin!"))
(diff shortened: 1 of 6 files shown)


Yellows91:
Oh boy here comes everyone from the queen nerf PR comment section

NexusStorm1:
Sentinel has needed some buffs for a while now, and I think the speed, paralyzing slash and scatterspit changes are much needed for the caste. Hibernate seems like a very niche ability, not too sure how useful it will be since the caste is more restrained by plasma than health. Tackle buffs seems excessive though, especially with the buffed paralyzing slash stun duration, but only playtesting will tell. 

Overall, should make the caste easier to play and more fun. +1 TM this right now admins!!!

TheManWithNoHands:
Hell yeah. A Senti buff that leaves them as mainly a support?
Count me in.

Senti is the forgotten cast for a long while now. And they need a LOT of love from the years of neglect.

Drulikar (changes requested):
- Please remove hibernate (I think this ability has merit for backliners in some form, but I don't really like that it has a simple counter-play that will only result in your death. Consider replacing the mechanic of recovery nodes healing you faster with the hibernation action instead in another PR)
- Please remove the change to tackle chance. Sentinels already have a good advantage with the paralyzing slash to get someone knocked down and can follow up with tackles to keep them down.

GoldenDarkness55:
> * Please remove hibernate (I think this ability has merit for backliners in some form, but I don't really like that it has a simple counter-play that will only result in your death. Consider replacing the mechanic of recovery nodes healing you faster with the hibernation action instead in another PR)

To clarify, it wasn't proposed with backlining in mind, since as you said it is too easy/too punishing to counter if used in non xeno controlled areas.

Drulikar (approved):

chefoxara:
wow yeah buff the OP caste that can easily solo cap!!! lemme give him more speed and also easier tackling!!
like???????????

Drulikar (changes requested):
Please drop the speed change; but otherwise good to go.

Drulikar (approved):

code/modules/mob/living/carbon/xenomorph/castes/Sentinel.dm
```diff
@@ -103,6 +103,6 @@
 #undef NEURO_TOUCH_DELAY
 
 /datum/behavior_delegate/sentinel_base/proc/paralyzing_slash(mob/living/carbon/human/human_target)
-	human_target.KnockDown(2)
-	human_target.Stun(2)
+	human_target.KnockDown(2.5)
+	human_target.Stun(2.5)
```
Diegoflores31, line 107: i recomend using apply_effect , also Weaken should work better than knockdown+ stun

PR #5957, opened 2024-03-16
OT CAS rocket

iloveloopers:
# About the pull request
adds AIM-224B-C 'Widowmaker Custom', basically an OT missile 
works the same as other OT assemblies except you crowbar to remove items
doesn't require any skills to tinker with the rocket so that research can also possibly make something (cool smoke?)
does not require a detonator
does a small explosion (pathetic) if you leave it unlocked or without containers
costs 350 points
4 second travel time
direct bombard only
explosion location is telegraphed by a blinking dot for the entire travel time

# Explain why it's good for the game
more CAS options are good
requires teamwork for maximum rocket efficiency so inter-department cooperation or whatever nanu likes to say
explosion limits up for debate

# Testing Photographs and Procedure

https://github.com/cmss13-devs/cmss13/assets/140007537/cf9b4e5f-d675-4198-9feb-e33455ba140d

image

image

# Changelog
:cl:
add: new missile AIM-224B-C 'Widowmaker Custom' which allows for custom payload.
imageadd: added AIM-224B-C 'Widowmaker Custom' sprites.
/:cl:

PR #7216, opened 2024-09-25
Adds a proper warning dot to thermobaric missile direct fire

GoldenDarkness55:
# About the pull request

Stole code from https://github.com/cmss13-devs/cmss13/pull/5957 to add a proper warning dot.

Thermobaric is currently the only CAS direct fire missile that cannot be dodged through the warning intended to make it possible because you cannot clear the AoE before it pulls you back in. This fixes that.
Essentially the same kinda issue that caused the FATTY rocket to be removed in the past. No warning = bad. 

<details>

https://github.com/user-attachments/assets/a736acb2-9413-436e-932f-f908d881bee1

</details>

# Changelog
:cl:
add: Added a warning to thermobaric missile direct fire.
/:cl:

### DIFF
--- a/code/modules/cm_marines/dropship_equipment.dm
+++ b/code/modules/cm_marines/dropship_equipment.dm
@@ -701,7 +701,7 @@
 
 	if(ammo_travelling_time && istype(SA, /obj/structure/ship_ammo/rocket/thermobaric))
 		playsound(impact, ammo_warn_sound, ammo_warn_sound_volume, 1, 15)
-		var/total_seconds = max(floor(ammo_travelling_time/10),1)
+		var/total_seconds = max(floor(ammo_travelling_time / 10), 1)
 		for(var/i in 0 to total_seconds)
 			sleep(1 SECONDS)
 			new /obj/effect/overlay/temp/blinking_laser (impact) //no decreased accuracy if laser dissapears, it will land where it is telegraphed to land
--- a/code/modules/cm_marines/dropship_equipment.dm
+++ b/code/modules/cm_marines/dropship_equipment.dm
@@ -686,9 +686,9 @@
 			break
 
 	msg_admin_niche("[key_name(user)] is direct-firing [SA] onto [selected_target] at ([target_turf.x],[target_turf.y],[target_turf.z]) [ADMIN_JMP(target_turf)]")
-	if(ammo_travelling_time)
+	if(ammo_travelling_time && !istype(SA, /obj/structure/ship_ammo/rocket/thermobaric))
 		var/total_seconds = max(floor(ammo_travelling_time/10),1)
-		for(var/i = 0 to total_seconds)
+		for(var/i in 0 to total_seconds)
 			sleep(10)
 			if(!selected_target || !selected_target.loc)//if laser disappeared before we reached the target,
 				ammo_accuracy_range++ //accuracy decreases
@@ -698,6 +698,14 @@
 
 	var/list/possible_turfs = RANGE_TURFS(ammo_accuracy_range, target_turf)
 	var/turf/impact = pick(possible_turfs)
+
+	if(ammo_travelling_time && istype(SA, /obj/structure/ship_ammo/rocket/custom_missile))
+		playsound(impact, ammo_warn_sound, ammo_warn_sound_volume, 1,15)
+		var/total_seconds = max(floor(ammo_travelling_time/10),1)
+		for(var/i in 0 to total_seconds)
+			sleep(10)
+			new /obj/effect/overlay/temp/blinking_laser (impact) //no decreased accuracy if laser dissapears, it will land where it is telegraphed to land
+
 	if(ammo_warn_sound)
 		playsound(impact, ammo_warn_sound, ammo_warn_sound_volume, 1,15)
 	new /obj/effect/overlay/temp/blinking_laser (impact)
--- a/code/modules/cm_marines/dropship_equipment.dm
+++ b/code/modules/cm_marines/dropship_equipment.dm
@@ -703,7 +703,7 @@
 		playsound(impact, ammo_warn_sound, ammo_warn_sound_volume, 1,15)
 		var/total_seconds = max(floor(ammo_travelling_time/10),1)
 		for(var/i in 0 to total_seconds)
-			sleep(10)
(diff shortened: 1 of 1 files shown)


iloveloopers:
code stealer!!!

cuberound:
imagine if there was some warning about the possibility of hitting there, or wait, there is, the bloody target for it need to be there for at least three seconds before it hits there (realisticly 5 or more with flare as it does now show up for few secs after placed and laser has several seconds of preaim for the target to show up), when you willingly ignore the target and die to it, that is up to you. Instead of giving 4 seconds of warning to missile with impact range that you can clear under two seconds. even tho you take it from missile that easily had two times wider impact area...

tallfission:
I think this is a balance PR more than anything, as there /is/ a warning for the vacuum missile already, from what I can tell you're simply extending it. 

Also, take it with a grain of salt as I play CAS a lot and might be biased, but imho there's already a lot of passive warning for a missile to land, if you don't want to get hit by CAS you shouldn't be standing next to CAS flares (flashing green, should be easy to spot) or lasers (also pretty obvious),

GoldenDarkness55:
Its an addition, if devs wanna pop a balance tag onto it they can and probably should do so, but cl's correct.

>code stealer!!! thief!!!

I steal the code, they steal old FATTY arguments, the cm cycle.

> next to CAS flares (flashing green, should be easy to spot) or lasers (also pretty obvious),

The issue is that by the time it flashes green you're already dead to this, as I said in the PR it's the only missile that's undodgeable from the warning. As for lazes you can just stop lazing the moment PO fires - that way there's no laze before the (insufficient for this missile) direct fire warning appears - same as fatty.
> would make more sense as a removal

Maybe. I don't think there's any balance issue with using it in firemissions though so I don't wanna remove it entirely if the direct fire issue can be fixed.

>  addition wise: imagine if there was some warning about the possibility of hitting there, or wait, there is, the bloody target for it need to be there for at least three seconds before it hits there (realisticly 5 or more with flare as it does now show up for few secs after placed and laser has several seconds of preaim for the target to show up), when you willingly ignore the target and die to it, that is up to you. Instead of giving 4 seconds of warning to missile with impact range that you can clear under two seconds and can not even kill a drone on its own. even tho you take it from missile that easily had two times wider impact area... if you want to give it more warnining make it extra 1 second not 5 (3 with targeting)

It doesn't have a proper warning unlike other missiles, see above. And no, you cannot clear the area because you're pulled back into it. It also doesn't matter how little damage it does on its own, because you fire a second missile alongside it that kills anything pulled into it including t3s.

Red-byte3D:
all the people against the nerf are people who are abusing the items, curious

VileBeggar:
https://github.com/user-attachments/assets/4ff049ed-5fcd-499c-8f2f-93ebd68921f4

they're extremely fast and don't give you much of window of opportunity to escape. before someone says "why are you hanging around a red dot", i could have been trying to melt a CAS flare and i would have suffered the same exact consequences. 
you don't have much of a choice except giving every single CAS signal a wide berth which is pretty atypical (most loadouts don't incur this amount of fear) and it's sometimes nigh impossible in the heat of battle

cuberound:
> all the people against the nerf are people who are abusing the items, curious

curious when you do it with warriors and and prae that you personaly buff hmm. I am for extra warning but 4 or 6 seconds of warning is a BIT too much considering you can cross easily corss 15 tiles in that time

tallfission:
I do agree the BLU probably needs a nerf, but as I've said, and should probably stop saying (this will be my last comment likely) this is just going to kill it and turn it to never use or you're bad at CAS status.

Venuska1117:
I like idea of giving it actually proper warning, because normally you hear it 2 seconds before it hit (-1second to realize what is happening), so giving it long warning *before* it hits its very good change, however, the actuall change is little unbalanced for CAS side, because it mean you have very expensive missles that can be easy dodged, it show for 8 seconds (the moment you fire it) on open space will be too easy to dodge, so to make people happy it would need one of this changes:

1. Decrease cost of missle in exchange for that 8 second warning.

2. Decrease warning time to 4 seconds before hit, but it can keep its high cost.

Nanu308:
CAS Is very strong right now in my opinion, and i agree that Xenos should have proper warnings as with how it currently sucks xenos in can be destructive to say the least. Balance Approved.

code/modules/cm_marines/dropship_equipment.dm
```diff
@@ -698,6 +698,14 @@
 
 	var/list/possible_turfs = RANGE_TURFS(ammo_accuracy_range, target_turf)
 	var/turf/impact = pick(possible_turfs)
+
+	if(ammo_travelling_time && istype(SA, /obj/structure/ship_ammo/rocket/thermobaric))
+		playsound(impact, ammo_warn_sound, ammo_warn_sound_volume, 1,15)
+		var/total_seconds = max(floor(ammo_travelling_time/10),1)
+		for(var/i in 0 to total_seconds)
+			sleep(10)
```
harryob, line 706:
```suggestion
			sleep(1 SECONDS)
```

code/modules/cm_marines/dropship_equipment.dm
```diff
@@ -698,6 +698,14 @@
 
 	var/list/possible_turfs = RANGE_TURFS(ammo_accuracy_range, target_turf)
 	var/turf/impact = pick(possible_turfs)
+
+	if(ammo_travelling_time && istype(SA, /obj/structure/ship_ammo/rocket/thermobaric))
+		playsound(impact, ammo_warn_sound, ammo_warn_sound_volume, 1,15)
```
harryob, line 703:
```suggestion
		playsound(impact, ammo_warn_sound, ammo_warn_sound_volume, 1, 15)
```

code/modules/cm_marines/dropship_equipment.dm
```diff
@@ -698,6 +698,14 @@
...
+		var/total_seconds = max(floor(ammo_travelling_time/10),1)
```
harryob, line 704:
```suggestion
		var/total_seconds = max(floor(ammo_travelling_time / 10), 1)
```

harryob (changes requested):
just some nits

harryob (approved):
code approved

PR #7224, opened 2024-09-26
Introduces new shrapnel types to OT

GoldenDarkness55:
# About the pull request

Allows OT to create the same shrapnel type as used by hornet grenades (damage taken increase debuff) by adding sulfuric acid and adds a new slowing neurotoxin (coated) shrapnel that requires neurotoxin plasma from research/xeno bodies. 
Doesn't seem like the effects stack even if you separate containers - it will apply only one special effect to all shrapnel.

Doubled amount of holostacks per shrapnel on hornets and reduced max shard divide from 4 to 2 for OT incendiary since I believe it was excessive.

Effects in the video section, though values used was 10 stacks on hornets and 0.4 slow per neuro which have been raised since.

More options good.

- Hornet

No reason not to give the OTs access to effects that exist in the code already, it's not particularly strong by itself so won't affect balance.
Doubled the stacks from 10 to 20 to make it more reliable for OT usage aswell as buffing the grenade itself - it's so bad literally no one uses them, raising the 5% to 10% dmg debuff on direct hit might not fix that but at least they'll be more worthwhile.

- Neurotoxin shrapnel

Research and OT having more in common = good. Doubt it'll make a big balance impact but should be situationally useful. Basically the kinda fun, not gamebreaking additions shipside should be about.

<details>

https://github.com/user-attachments/assets/c7c6ad86-185b-47f8-8fa8-594eef271700

https://github.com/user-attachments/assets/41ad27a6-2d91-4796-9325-7c3950ede7f4

</details>

# Changelog
:cl:
add: Slowing neuro shrapnel effect to OT achieved by adding 30 neurotoxin plasma to the beaker with iron. 0.8s slow per shard hit up to 6s max
add: Hornet (holo) shrapnel effect to OT achieved by adding 10 sulfuric acid to the beaker with iron
balance: Reduced the divide on incendiary OT shards from 4 to 2
balance: Raised holo stacks per hornet shrapnel hit from 10 to 20.
/:cl:


cuberound:
just to correct the math , 10 holosack are 1% extra dmg and the extra dmg caps at 10% (100 holostack), ot shrapnels need better accuracy in general

GoldenDarkness55:
> just to correct the math , 10 holosack are 1% extra dmg and the extra dmg caps at 10% (100 holostack), ot shrapnels need better accuracy in general

Ye it was 5% on direct hit with the hornet nade, 10% now.
As for shrapnel accuracy im pretty sure OT doesn't use any separate one and it's the same shrapnel most other explosives use, the default one at top of the shrapnel file. I somewhat agree it should be higher, but don't really wanna make that big a change here.
You can try opening a separate PR for it

iloveloopers:
So you wait for marines to bring up a xeno body, pray its a sent/spitter/prae, bother research to give you neurotoxin plasma, sacrifice precious container space that could have been used for a bigger boom on the grenade, to slow a xeno for 2 whole seconds.

Why would an OT go through all this trouble when they can just fill the grenade with cyclonite for an actual explosion with damage?

chefoxara:
2 sec slowdown is nice, one thrown into a xeno push could make marines get more kills cause xenos will not have the speed to run away.

L0pz:
More ways for OT to kill marines, I approve

GoldenDarkness55:
> to slow a xeno for 2 whole seconds.

its 0.8 per shard hit up to 6 seconds, actually. As I said the video value was lower (0.4), I just didn't bother recording again.

code/datums/ammo/shrapnel.dm
```diff
@@ -70,6 +70,19 @@
...
+/datum/ammo/bullet/shrapnel/neuro
+	name = "neurotoxin coated shrapnel"
+	icon_state = "neurotoxin"
+	flags_ammo_behavior = AMMO_STOPPED_BY_COVER
+
+	shell_speed = AMMO_SPEED_TIER_1
+	damage = 30
+	penetration = ARMOR_PENETRATION_TIER_4
+
+/datum/ammo/bullet/shrapnel/neuro/on_hit_mob(mob/living/M, obj/projectile/P)
```
kiVts, line 82: Single letter var

kiVts (changes requested):

Nanu308:
Will take a look at after OT nerf pr is fully merged.

kiVts (approved):
I doubt it will be powerfull if at all. Main cmplaints about OT it their maxcaps not the anti xeno smoke/etc

PR #7230, opened 2024-09-28
Sentinel's Slowing Spit can now drain Medical and Research chems on hit

TheKillfish:
# About the pull request

Adds options for the apply_neuro effect used by neurospit to drain Medical Reagents and Generated Reagents on hit. Primarily intended for Sentinel, however Queen can be also be made to use it. Drain, **_for now_**, is based off effect power, meaning 1u for each applicable chem per hit in Sent's case and 2u if Queen is allowed to.

The draining effects have been ordered such that Med draining is blocked by Neuroblocking traits while Stims will always be drained.

Stims are currently very unfair for Xenos to deal with as they have no counter save for getting a King, and so far they have not been nerfed particularly much. Sentinel meanwhile has been suffering from lack of content for a while. This sorts both problems, giving Xenos a more attainable counter to stim usage and giving Sentinel a new purpose; draining stimmed Marines.

I added medchem draining as an option since most things that purge reagents apply to all reagents good and bad, and not just a specific subtype like what I've done here. This gives backed-in modularity, and modularity es gud.

# Changelog
:cl: Killfish
balance: Added Stim and Medical Chem draining effects to Sentinel's Slowing Spit.
/:cl:


harryob (approved):
interested to see how this plays out

PR #7235, opened 2024-09-28
Staff default toggles

realforest2001:
# About the pull request

Allows staff to disable their AFK protections.
Also updates default chat settings so new staff aren't spammed by a million attack logs from legacy defaults which were never updated on savefiles.

# Changelog
:cl:
admin: Staff can now toggle their AFK protections on and off.
code: Default chat settings now include FF Attack logs for staff.
/:cl:


harryob (changes requested):
this won't be applied retroactively, make a savefile migration so people without it have it toggled on

PR #7247, opened 2024-09-30
Adds Scandinavian as a language option in the traits menu

WazLach:
# About the pull request

I have added Scandinavian as a language in the traits option. The syllables used are mostly the same as the German ones after being translated into Norwegian, I chose Norwegian since that's the one I'm most familiar with. Scandinavian languages are lumped together because:
A: They are mutually intelligible.
B: They are hard to distinguish if you don't speak it.
C: I assume with time the languages have only gotten more similar.
D: It gives an option for characters from multiple countries to have a language.

NOTE: The key for the language is currently set at 0, as all other numbers are taken. I do not like this, I would rather it be set to something else, however I do not know all the keys, if there is a better suggestion feel free to change it.

Allows characters more roleplay options, while German exists and is similar to Scandinavian languages, they are distinct and it does restrict roleplay options. I generally believe that having more options to flesh out a character is good, especially when it doesn't impact the game itself majorly.

<details>

https://github.com/user-attachments/assets/6c771fd0-6fcf-40aa-85a9-f597b9f2abdb

</details>

# Changelog

:cl: lokideslimee
add: Added Scandinavian as a language
code: Coded in the Scandinavian
code: Added Scandinavian as a language for mobs that should know the language
/:cl:


WazLach:
Note that this is my first actualy PR so if there's any feedback I'd love to hear it.

TheManWithNoHands:
Just and fyi. You can make the language key use letters too, not just numbers. Should give more options for you to use then 0

code/modules/mob/language/languages.dm
```diff
@@ -57,6 +57,17 @@
 
 	syllables = list("die", "das", "wein", "mir", "und", "wir", "ein", "nein", "gen", "en", "sauen", "bin", "nein", "rhein", "deut", "der", "lieb", "en", "stein", "nein", "ja", "wolle", "sil", "bei", "der", "sie", "sch", "kein", "nur", "ach", "kann", "volk", "vau", "gelb", "grun", "macht", "zwei", "vier", "nacht", "tag")
 
+/datum/language/scandinavian
+	name = LANGUAGE_SCANDINAVIAN
+	desc = "While not technically one language, Scandinavian languages have grown similar and are nearly indistinguishable from one another unless you actually know the languages."
+	speech_verb = "blubbers"
+	ask_verb = "queries"
+	exclaim_verb = "yelps"
+	color = "blue"
```
TheManWithNoHands, line 66: Use a hexcode for color.
And if you want to keep it similar to how the other languages do color, add the color ID and hexcode for it here.
I say ad it, dont know TGUI myself well enough to know if its needed or not.

cmss13\tgui\packages\tgui-panel\styles\goon\chat-dark.scss

WazLach:
I've added the colour properly now, I found it readable, though if others have trouble I'd like to know. Also changed some of the syllables in the list, since some of them weren't great, the result should pass as Scandinavian mumbo jumbo better.

WazLach:
A final note the keys, I'm leaving it as "0" since there's not many other options, other than symbols such as "!" which I want to avoid using for consistency, though I've tested it and it does seem to work.

code/modules/mob/language/languages.dm
```diff
@@ -57,6 +57,17 @@
 
 	syllables = list("die", "das", "wein", "mir", "und", "wir", "ein", "nein", "gen", "en", "sauen", "bin", "nein", "rhein", "deut", "der", "lieb", "en", "stein", "nein", "ja", "wolle", "sil", "bei", "der", "sie", "sch", "kein", "nur", "ach", "kann", "volk", "vau", "gelb", "grun", "macht", "zwei", "vier", "nacht", "tag")
 
+/datum/language/scandinavian
+	name = LANGUAGE_SCANDINAVIAN
+	desc = "While not technically one language, Scandinavian languages have grown similar and are nearly indistinguishable from one another unless you actually know the languages."
+	speech_verb = "blubbers"
```
harryob, line 63: uh oh

harryob (approved):

PR #7249, opened 2024-09-30
Fixes neuraline being in research chem recipes 

Blundir:
# About the pull request

Neuraline is a special chem only found in CL's implant, and it cannot be extracted or obtained by other means (except pure gamble of botany mutations), which makes recipes containing it being impossible to fulfill.

D5gWUYCMSz

# Changelog
:cl:
fix: fixed neuraline being a potential part of a chem recipe
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry_reagents/medical.dm
+++ b/code/modules/reagents/chemistry_reagents/medical.dm
@@ -186,7 +186,7 @@
 	custom_metabolism = AMOUNT_PER_TIME(1, 5 SECONDS)
 	overdose = 2
 	overdose_critical = 3
-	chemclass = CHEM_CLASS_RARE
+	chemclass = CHEM_CLASS_SPECIAL
 	properties = list(PROPERTY_NERVESTIMULATING = 5)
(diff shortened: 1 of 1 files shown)


private-tristan:
Neuraline is actually obtainable from mutagen plants.
However, you CAN'T obtain is thwei or green blood, due to both being exclusively from preds (crystals), vial boxes, and 1 or 2 nightmare spawns on LV.

Blundir:
> Neuraline is actually obtainable from mutagen plants. However, you CAN'T obtain is thwei or green blood, due to both being exclusively from preds (crystals), vial boxes, and 1 or 2 nightmare spawns on LV.

That is not a good way to get one, also it's supposed to be implant only considering the comments

iloveloopers:
Pretty sure it is intended to be found in plants because unlike ultrazine (removed from mutagen plant loot), it is useless without the implant due to the low OD (2) and fast metabolization. You only use it to get NST for a stim though it should absolutely be removed from recipes due to how much of a pain it is to get.

harryob (approved):

PR #7259, opened 2024-10-01
Slightly increases the amount of time it takes to exit ventcrawling from disconnected pipes

Tyranicranger4:
# About the pull request

This PR slightly increases the time it takes to exit ventcrawling from a disconnected pipe segment (doesn't affect exiting via vents/scrubbers), from 2 seconds to 3.

Unlike vents and scrubbers, pipe segments have the possbility of being completely hidden (more often than not this is the case, due to how explosions work); in which case they have no visible cue for when something is emerging from them, and the potential threat isn't something you can easily identify in advance. Thus, I think it's fair for there to be slightly more leeway in the amount of time you get to react when something *does* try to emerge from it. This primarily concerns player controlled huggers, which can begin channeling a pounce *immediately* upon exiting a vent.
<details>

https://streamable.com/q8inq2

</details>

# Changelog
:cl:
balance: Slightly increased the time it takes to exit ventcrawling from disconnected pipe segments (this doesn't affect exiting through a vent/scrubber)
/:cl:

### DIFF
--- a/code/game/objects/structures/pipes/pipes.dm
+++ b/code/game/objects/structures/pipes/pipes.dm
@@ -136,7 +136,10 @@
 		to_chat(user, SPAN_NOTICE("You begin to climb out of [src]"))
 		animate_ventcrawl()
 		user.remove_specific_pipe_image(src)
-		if(!do_after(user, 20, INTERRUPT_NO_NEEDHAND))
+		var/exit_time = 20 //So that exiting through a destroyed pipe segment takes longer than a proper vent
+		if(!istype(src, /obj/structure/pipes/vents))
+			exit_time += 10
+		if(!do_after(user, exit_time, INTERRUPT_NO_NEEDHAND))
 			animate_ventcrawl_reset()
 			return
 


Unknownity:
FYI, exiting from pipes has a visual clue, the pipe shakes.
"What about hidden pipes" Could simply solve that by removing the tile if someone exits the pipe to prevent repeat attempts.

Tyranicranger4:
> FYI, exiting from pipes has a visual clue, the pipe shakes.

This is mostly directed towards hidden pipes, which are more often than not where the pipe segments break at due to how explosions work. I forgot to specify this, so I've edited the description slightly for clarity.

> "What about hidden pipes" Could simply solve that by removing the tile if someone exits the pipe to prevent repeat attempts.

This doesn't have quite the effect I'm looking for. It would also make hidden pipes less viable for sneakily bypassing frontline fortifications, which isn't a change I agree with. My issue here isn't that you can exit from hidden pipes, it's that you can do so directly on top of someone and with little warning. This just gives you a slightly longer grace period in that situation.

PR #7265, opened 2024-10-02
Allows using empty verb on storage box items

rythenx:
# About the pull request
This makes it so that the empty verb can be used on storage box items to empty them as a quality of life thing.  This impacts all types of objects that use STORAGE_FLAGS_BOX, specifically:

- cardboard boxes including pre-filled ones
- pill bottles
- medical kits
- pill packets
- engineering kits

Being able to use the empty verb rather than taking items out one at a time is just a QOL thing and shouldn't impact balance in any way.  If I time myself taking items out of the box 1 by 1 and dropping it on the floor it takes nearly the same amount of time as the delay that is imposed by the empty verb.  It just makes it so you don't have to click rapidly to accomplish it.

The fact that XOR was being used on this flag is also a bit strange to me.  Both operands (default storage flags) and allow storage drawing were 1s which resulted in a 0 in that position.  With the way this was written, if the DEFAULT_STORAGE_FLAGS were ever to be set so that emptying was disabled, this would then ENABLE it for boxes but disable it for everything else which doesn't make a lot of sense.

<details>

</details>

# Changelog
:cl:
qol: boxes can be emptied using empty verb
/:cl:


harryob (approved):
sure, but i'll maintain that verbs are bad ux

PR #7271, opened 2024-10-05
Fix Xeno huds Alpha Invisibility (Lurkers and Burrowers)

Venuska1117:
# About the pull request

Makes xeno huds visible for xenos that use alpha to become invisible.

Makes xenos like lurkers more "visible" to other sisters around, now you can see their HUD's and help them when they are in critical, i seen few times that lurkers died near sisters in critical because they were invisible and hard to spot.

<details>
<summary>Click HERE to see Videos.</summary>

Before:

https://github.com/user-attachments/assets/b9dc2ca3-53c3-417f-8929-8a637af90acb

After: 

https://github.com/user-attachments/assets/cb9e8f06-e4fc-42d3-bd77-f1b864790cdf

</details>

# Changelog
:cl: Harry, Venuska1117
code: Makes all xeno huds ignore alpha, it makes them visible during invisibility.
/:cl:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -104,6 +104,8 @@
 				I = image('icons/mob/hud/hud_yautja.dmi', src, "")
 			if(HOLOCARD_HUD)
 				I = image('icons/mob/hud/marine_hud.dmi', src, "")
+		if(hud in /datum/mob_hud/xeno::hud_icons)
+			I.appearance_flags |= RESET_ALPHA
 		I.appearance_flags |= NO_CLIENT_COLOR|KEEP_APART|RESET_COLOR
 		hud_list[hud] = I
 


RainbowStalin:
Does this also apply for enemy hives?

Venuska1117:
> Does this also apply for enemy hives?

image

Yes, because they all share same HEALTH_HUD_XENO, from what i remember it apply on enemy burrowers too.

image

There is photo taken from master branch.

ihatethisengine:
I am pretty sure there was a PR like this years ago (maybe on gitlab) and it was rejected

Unknownity:
> I am pretty sure there was a PR like this years ago (maybe on gitlab) and it was rejected

https://gitlab.com/cmdevs/colonial-warfare/-/merge_requests/1810

closed on the basis of XvX and HvH issues

Drulikar (changes requested):
I'm not terribly opposed to this, but I still think the current behavior is preferred over marginally nerfing stealth for xenos. If you are certain you want to move forward with this, consider exploring how you might get the hud unique per hive and ally status to add/remove that hud from that hive. It'll be a moderately sized endeavor to accomplish I think.

Alternatively, make a new hud for all xenos that indicates they are low health but is otherwise hidden. This hud can be the one that uses RESET_ALPHA and it can be as simple as maybe a border around the xeno health bar or something (or even a duplicate of the health HUD, but I think it would be more interesting the other way to help all xenos not just stealthed ones).

Venuska1117:
i will re-open this PR after i find a way to make it better.

PR #7084, opened 2024-09-02
Add visual sprite display of ammo count for dropships, add toggle to auto deploy to sentries/MGs on landing, dropship fabricator build queue

Sindorman:
# About the pull request

1.  Added ability to set Dropship Sentries and MG modules on dropship to auto deploy after each landing. Default set to off, can be enabled/disable in the equipment menu of the sentry via dropship equipment computer.

1. Added visual representation for gunship ammo to display how much is left in sprite

1. Added visual representation of how many rockets loaded into mini rocket launcher when mounted on wings

1. Added a build queue system for dropship build fabricator with ability to cancel items in the queue too.

# Explain why it's good for the game

1. This will help dropship pilot by removing tedious slow process of deploying every external Sentry/MG after landing

1. This will help gunship pilot or DCC be able to tell ammo count left for used ammo that is sitting on the ground

1. This will help gunship pilot or DCC be able to tell how many rockets is loaded into the mini rocket launcher when they are mounted on the wings

1. This will help people to queue up things to build to allow multi tasking.

# Testing Photographs and Procedure

1. Installed Sentry and MG on dropship and flew it to LZ
3. Flew Gunship loaded with every type of weapon, fired them and then pulled them out to see. Also tested the wing module, also tested by changing VV.
4. Same thing as 2
5. Tested on the fabricator by queuing up things to be built and cancelling some.

<details>
<summary>Screenshots & Videos</summary>

Sentries and MG deploying after landing:
https://github.com/user-attachments/assets/d36529b1-d94b-44af-907a-9a6706cd62ec

New gunship ammo visual representation:
dreamseeker_GtK0jZV2VQ

Auto deploy Toggle:
xJkTiBTI1b

Auto deploy preference in action:
dreamseeker_8W8C9LEt98

New Dropship lights
dreamseeker_gSFIfKyLZj

Dropship fabricator queue:
https://github.com/user-attachments/assets/254b80cd-159e-4991-bae7-6d58409c6e9f

</details>

# Changelog

:cl: PoZe
add:  Added ability to set Dropship Sentries and MG modules on dropship to auto deploy after each landing. Default set to off, can be enabled/disable in the equipment menu of the sentry via dropship equipment computer.
add: Added visual display in Gunship ammo for all ammo types so that you can tell how much is left by looking at it on the ground.
qol: Added ability to queue up items to be fabricated in dropship fabricator as well as ability to remove items from the queue too. Crew will no longer need to babysit the fabricator to produce multiple things as soon as possible.
[...]

PR #7282, opened 2024-10-07
stops pilots breaking the bank

harryob:
:cl:
fix: pilots can no longer take out enormous loans to buy their equipment
/:cl:

### DIFF
--- a/code/modules/cm_marines/vehicle_part_fabricator.dm
+++ b/code/modules/cm_marines/vehicle_part_fabricator.dm
@@ -92,6 +92,7 @@
 		if(get_point_store() < entry.cost)
 			if(!TIMER_COOLDOWN_CHECK(src, COOLDOWN_PRINTER_ERROR))
 				balloon_alert_to_viewers("out of points - printing paused!")
+				visible_message(SPAN_WARNING("[src] flashes a warning light."))
 				TIMER_COOLDOWN_START(src, COOLDOWN_PRINTER_ERROR, 20 SECONDS)
 			busy = FALSE
 			return
--- a/code/modules/cm_marines/vehicle_part_fabricator.dm
+++ b/code/modules/cm_marines/vehicle_part_fabricator.dm
@@ -142,7 +142,7 @@
 				return
 			cost = initial(produce.point_cost)
 			build_part(produce, cost, user)
-			return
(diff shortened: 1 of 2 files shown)


code/modules/cm_marines/vehicle_part_fabricator.dm
```diff
@@ -149,7 +157,7 @@
 		var/index = params["index"]
```
Drulikar, line 157: I doubt this is bug caused by this PR (more likely #7084 ) ; but in testing this I am able to get the index == null if I queue up items and cancel them: 
image

Seems fairly easy to reproduce by queueing say 5 things, and then spam clicking cancel on whatever is item 2 until you get an exception.

Drulikar (dismissed):

code/modules/cm_marines/vehicle_part_fabricator.dm
```diff
@@ -149,7 +157,7 @@
...
 
 		if(length(build_queue))
-			if(index == 1)
+			if(busy && index == 1)
 				to_chat(user, SPAN_WARNING("Cannot cancel currently produced item."))
 				return
```
Sindorman, line 163: This to fix issue @Drulikar was seeing
```suggestion
				if(index == null || index > length(build_queue))
					return
```

realforest2001 (approved):

code/modules/cm_marines/vehicle_part_fabricator.dm
```diff
@@ -143,21 +151,24 @@
...
 			build_part(produce, cost, user)
-			return
+			return TRUE
 
 	if(action == "cancel")
 		var/index = params["index"]
 
 		if(length(build_queue))
-			if(index == 1)
+			if(index > length(build_queue))
```
Sindorman, line 160: This should still have a `null` check since @Drulikar was able to get `index = null` edge case
```suggestion
			if(index == null || index > length(build_queue))
```

harryob: the error in drathek's screenshot is not null, it's going oob because the tgui value had not updated in the act. but sure

Nanu308 (approved):
awesum

PR #7284, opened 2024-10-07
automatically connect people via selected relay

harryob:
prompts people to connect to a relay instead of directly if available. can opt out of the warning by intentionally selecting no relay in settings

:cl:
server: if relays have been set up, the server will prompt people to select a preferred relay, and connect them over the relay when possible
/:cl:

### DIFF
--- a/code/__DEFINES/client_prefs.dm
+++ b/code/__DEFINES/client_prefs.dm
@@ -53,3 +53,6 @@
 #define DUAL_WIELD_SWAP 1
 ///Do nothing when dual wielding
 #define DUAL_WIELD_NONE 2
+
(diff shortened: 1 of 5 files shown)


Drulikar (changes requested):
Also - though not a result of this PR - the relay connection doesn't have the confirm override text anymore somereason so you wouldn't know to click to confirm again: 
image

But I guess take a look if you don't beat me to it.

*Edit* What do you think of this change? 
image
Should fix the issue here.

Drulikar:
Rectified what I was talking about w/ the ping browser in #7474

cm13-github:

PR #7286, opened 2024-10-07
Buffs contractor ABR40

CapCamIII:
# About the pull request

Contractor ABR40 is a bit better, firing faster and doing a bit more damage, as well as having an integrated magharn, in addition the ABR40 loadout has been rearranged to have more ammo.

ABR40 sucked, as it was balanced mainly as a survivor rifle, so having this be more equal to marine guns is good, more ammo is important for its fire-rate as well.

# Changelog
:cl:
balance: The contractor ABR40 now fires slightly faster and does a bit more damage, also having an integrated magharn.
add: Shuffles around contractor ABR40 loadout, sacrificing the smoke grenades and holotargeting rounds for a bit more normal ammo.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -1714,31 +1714,34 @@
 		/obj/item/attachable/bayonet,
 		/obj/item/attachable/bayonet/upp,
 		/obj/item/attachable/bayonet/co2,
-		//Rail,
-		/obj/item/attachable/reddot,
-		/obj/item/attachable/reflex,
-		/obj/item/attachable/flashlight,
-		/obj/item/attachable/magnetic_harness,
-		/obj/item/attachable/scope,
-		/obj/item/attachable/scope/mini,
-		/obj/item/attachable/scope/mini/hunting,
 		//Under,
 		/obj/item/attachable/flashlight/grip,
 		//Stock,
-		/obj/item/attachable/stock/carbine,
 		/obj/item/attachable/stock/carbine/wood,
 		/obj/item/attachable/stock/carbine/wood/tactical,
 	)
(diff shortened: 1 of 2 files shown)


Diegoflores31:
cant you just buff the VAIPO version ? they have their own subtype , its not the same that the surv one which means you can tweak it winout affecting surv balance

cuberound:
funny enough the tactical is curently the same as surv one plust attachments perhaps, but yes buffing their subtype instead of removing it might be better decision

Blundir (dismissed):
I agree, we have a whole unique version sitting in a closet, just make it better instead of straight removal

code/modules/projectiles/guns/rifles.dm
```diff
@@ -1714,32 +1714,35 @@
...
+	random_under_chance = 100
 	random_spawn_under = list(/obj/item/attachable/flashlight/grip)
 
+/obj/item/weapon/gun/rifle/l42a/abr40/tactical/handle_starting_attachment()
+	..()
+	var/obj/item/attachable/magnetic_harness/Integrated = new(src)
+	Integrated.flags_attach_features &= ~ATTACH_REMOVABLE
+	Integrated.hidden = TRUE
+	Integrated.Attach(src)
+	update_attachable(Integrated.slot)
```
harryob, line 1733:
```suggestion
	var/obj/item/attachable/magnetic_harness/integrated = new(src)
	integrated.flags_attach_features &= ~ATTACH_REMOVABLE
	integrated.hidden = TRUE
	integrated.Attach(src)
	update_attachable(integrated.slot)
```

harryob (changes requested):

harryob (approved):

PR #7294, opened 2024-10-09
Halves the time required to suicide via mouth target.

efzapa:
# About the pull request

Mouth target suicide now takes 2 seconds instead of 4.

In a hijack scenario where you might consider it is worth it to end your own life to prevent your own capture, it might be appealing to use the built in suicide mechanic, until you realize it takes 4 seconds, which coincidentally is long enough for a Queen, off weeds to run right at you and shuffle you even coming from off screen, the mechanic, and self harm shooting should be (roughly) the same time. See the video attached

<details>

https://github.com/user-attachments/assets/74461abd-1210-40cb-8851-003f6949ae65

</details>

# Changelog
:cl:
qol: Mouth Target Suicide now takes 2 seconds instead of 4.
/:cl:


Sindorman (commented):
This might be out of scope of PR. But something to consider adding:
Lot of issues I find in CM UX is some actions should be locked behind intents. For example its possible to stab yourself on help intent and also stab other people, which lots of other codebases lock behind harm intents and such to avoid accidents or unintentional actions. I think this specific action should be locked behind harm intent or at least non-help intent, because its possible to almost accidentally do it.

VileBeggar:
> Lot of issues I find in CM UX is some actions should be locked behind intents. For example its possible to stab yourself on help intent and also stab other people, which lots of other codebases lock behind harm intents and such to avoid accidents or unintentional actions.
CM has a "don't harm on help intent" option, it's just hidden in one of the preferences menu (game or combat preferences, i'm not sure)

efzapa:
> This might be out of scope of PR. But something to consider adding:
> Lot of issues I find in CM UX is some actions should be locked behind intents. For example its possible to stab yourself on help intent and also stab other people, which lots of other codebases lock behind harm intents and such to avoid accidents or unintentional actions. I think this specific action should be locked behind harm intent or at least non-help intent, because its possible to almost accidentally do it.

basically what Vile said but harming yourself, atleast normally is locked behind a thing in your prefs tab, not sure if that also affects mouth target suicide but it shouldn't really be a problem given almost nobody should be targeting mouth UNLESS they expect to be shooting themselves anyways

deathrobotpunch:
> Well, it's kinda weird that "toggle ability to hurt yourself" also prevents you from suiciding through this mechanic. So if you want to suicide using this mechanic, you need to disable the pref first, which can take a lot of time. Out of the scope, but would be cool if you make this an exception for the pref.

TGMC has a dedicated keybind to toggle self harm which is set to the 0 key

Nanu308 (approved):
lgtm

PR #7306, opened 2024-10-13
Changes to Donator Kits and custom_items.txt

BeagleGaming1:
# About the pull request

Changes the config/custom_items.txt to fill the donator kit of each donator kit having its own typepath

Being able to grab the donator kit instead of spawning with it makes inventories not be full of gear right out of spawning
Also just better code in general

# Changelog

:cl:
add: Adds the personal gear vendor, where people can get their donator items or a random loadout item
add: Custom Items can be gotten from Requisitions instead of spawning with them
config: Changes to config/custom_items.txt
/:cl:


Nanu308 (approved):
mapping lgtm

deathrobotpunch:
not appreciative of the stealth loadout item nerf

Drulikar (changes requested):
Need a working config before I can approve/merge.

Drulikar (changes requested):
Only remaining issue appears to be that the shake verb is possible to get items out w/o ckey lock. Can either disable the `storage_flags` of `STORAGE_ALLOW_EMPTY`, re-implement shake to do nothing, or implement its `allowed` proc.

Drulikar (changes requested):
Some more polish and implementation of `allowed` to disable shaking (empty would also respect `allowed` but kept your implementation for the message.

Moving around the parent calls is to handle issues where you smack the console with a box rather than just return it.

Use `git apply path/to/the/patch.txt` and try out my suggestions: 
[patch.diff.txt](https://github.com/user-attachments/files/17694914/patch.diff.txt)

image

Drulikar (approved):

SASoperative:
I just saw that this got accepted, and as one of the only donator's left in existence, I don't really care for it.. since now, I can only have one kit at a time, instead of all of my kits. which I used to mix and match from sometimes.

PR #7313, opened 2024-10-14
Adds the ability for dual-tube shotguns to rechamber from the other tube

VileBeggar:
# About the pull request

adds an item ability to dual tube shotguns (which are just the hg and it's military variant) which allows rechambering. it functions like this:
1. assume the first tube is filled with buckshot rounds, and the second tube with slugs. the shotgun currently has buckshot chambered.
2. when we activate the ability, every time the user switches tubes the ammo type from the other tube is chambered, and the currently chambered round is stored back into its own tube or is unloaded if there isn't room for it.
3. so in practice, this means when we switch from the buckshot tube to the slug one, the chambered buckshot is stored if there's space, and a slug round from the other tube is immediately chambered. works vice versa, with slug switching to buck.

in short, it lets you switch from buck to slug and fire that slug immediately without being forced to waste that buckshot shell
i suck at explaining, so the video in the testing section would probably explain it better

the dual tube mechanic of the shotguns is extremely gimmicky and they're rarely utilized for their intended purpose. most of the time they're being used as second-hand shotgun with the dual tube mechanic being ignored, or in the case of the military version, being utilized for it's flat damage bonus (even then, xenos don't even bother aciding these things and they're flat out ignored by marines for some reason).

being able to switch the chambered round on demand is a very big boon and allows the user to immediately respond to changes in combat as the situation demands.
 
the issue with the dual tubes is that you have to pay a lot of attention to what you chamber and when. you can have flechette currently chambered, but if a situation crops up where you need to fire a slug you have no way of rechambering it without wasting that flechette.

i don't even know if this is possible to do in real life but it's a video game and i think fun and usability is more important

<details>

https://github.com/user-attachments/assets/aa1ece78-f273-43ac-90c6-640705271c21

</details>

# Changelog
:cl:
add: Dual-tube shotguns (like the HG-37) can now rechamber their currently loaded shell with the other tube. This allows you to instantly switch from, e.g. a tube filled with buck to one filled with slugs, and fire the slug immediately just by switching shotgun tubes. 

/:cl:


chefoxara:
peak

Blundir (approved):

Blundir:
doesn't intervene with sprite freeze

harryob (approved):

PR #7325, opened 2024-10-16
No more multiple patient surgeries on one tile

### LINKED ISSUE #5516: Multiple people can lie on the same operating table, allowing surgery for every one of them
### Testmerges

irrelevant

### Round ID

20498

### Description of the bug

You can pile 2+ marines on the same tile with an operating table on it, and perform alien embryo removal on both of them.

One table should fit one marine. If there is already a marine on the tile, lying down, the second one should either a) flat out refuse surgery b) count the surface as a floor, aka awful surgery surface.

It looks silly when two or more doctors do surgery on the same operating table, on multiple different patients.

1. Spawn a `/obj/structure/machinery/optable`
2. Put two human mobs on them, lying down (not buckled)
3. Do surgery on them

### Issue Bingo

blackcrystall:
# About the pull request
When you do surgery on somebody who shares space with someone else (likely living being), you can't do as expected surgery, exception only you and patient on same turf

I have no clue what wrong with that, but okay, if it's marked as bug, because you can't do multiple surgeries steps at ONCE, so you need finish one action, then do another in another surgery... but yea, you can have multiple surgeries ongoing at same time...

# Changelog

:cl: BlackCrystalic
fix: No more multiple surgeries on same operation table for 2x and more patients at same time
/:cl:

Closes #5516

### DIFF
--- a/code/modules/surgery/surgery_procedure.dm
+++ b/code/modules/surgery/surgery_procedure.dm
@@ -98,7 +98,7 @@
 	for(var/mob/living/potential_blocker in get_turf(target))
 		if(potential_blocker == user || potential_blocker == target)
 			continue
-		to_chat(user, SPAN_WARNING("You can't operate, when there's small amount of space! Remove everybody else."))
+		to_chat(user, SPAN_WARNING("You can't operate when you don't have enough space! Remove everybody else."))
 		return FALSE
 
 	if(user == target)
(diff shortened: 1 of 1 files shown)


ihatethisengine:
Probably someone was doing was doing multiple operations on the same table with other doctors. The bug can be abused this way if you have a doc and a synth for example, both working on different patients but on the same table.

harryob (approved):

PR #7326, opened 2024-10-16
Xeno grabs now block crawling

private-tristan:
# About the pull request
see title, code is copied from  #7039 

players (mostly synths and predators) were able to easily escape xeno bodyblocks by just crawling away, and there was nothing that xeno players could do to counter it. Now preds/synths can still try to crawl to escape a bodyblock (over attempting to crit a xeno blocking them), at the cost of it failing if a xeno grabs them.
Tested it, works.

# Changelog
:cl: MistChristmas, Private-Tristan
balance: You can no longer crawl while being grabbed by a xenomorph.
/:cl:

### DIFF
--- a/code/modules/mob/living/living_verbs.dm
+++ b/code/modules/mob/living/living_verbs.dm
@@ -24,6 +24,10 @@
 			to_chat(src, SPAN_WARNING("You can't resist in your current state."))
 			return
 
+	if(pulledby && isxeno(pulledby))
+		to_chat(src, SPAN_WARNING("You can't resist while a xeno is grabbing you."))
+		return
+
 	resisting = TRUE
 
 	next_move = world.time + 20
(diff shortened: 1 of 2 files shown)


JustLat:
Hi! As a player with one of the aforementioned whitelists listed here, the reason this would not be a good implementation is:

It’s really easy to trap preds/synths in a corner or with 4 xenos in each cardinal direction! Mainly due to constant body-blocking, as well as queen screech!
Crawling exists as a means of escape so you aren’t screwed over as a player when a horde of xenos are surrounding you! This has barely any correlation to normal marines because if there is 2+ xenos on them, bodyblocking, they are very quick to die or be capped!

This seems like a significant middle finger to both synthetics and yautja!

TL;DR: This is an unneeded change!

GoldenDarkness55:
Crawling under xenos should be prevented by default - grabbing a small humanoid sprite from under a xeno isn't very likely to happen.

Nanu308:
Reminder for myself to look at this

ihatethisengine:
Can't you remove the grab by simply resisting?

harryob (approved):
code

PR #7342, opened 2024-10-18
AMR focus fire removal, spotter and sniper tweaks

GoldenDarkness55:
# About the pull request

Removed AMR focus fire stacking, rebalanced the damage.
Raised base sniper aimed shot time, halved flak aimed shot slow duration and raised its aimed shot damage and base armor pen.
Lowered the bonus from spotters and slightly raised aim time without them, removed ghillies and smock from spotter kit.
Changed both sniper rifles to use a blue targeting reticle to differentiate from spotter lazing
Inspired by all the complaints in https://forum.cm-ss13.com/t/amr-begone/10830

- Spotters

The bonus made aimed shot way too fast, messing up the number balance entirely and leaving next to no time to react for the xeno. Ghillie nor smock isn't necessary for the role and introduced confusion as to who is the sniper when multiple marines were wearing it.

**How improved**
Straightforward, lowered bonuses and removed offending item. Put more importance on the role by raising the timers without a spotter to encourage teamwork further as otherwise new values would be too negligible.

- AMR focus fire

Current stuck between a degrade to the main rifle and dealing exorbitant damage at full stacks, discuraging people from playing entirely once the sniper has reached max stacks, until he changes the target.
Now, sure it could be fixed by adding a timer that people would just wait out, but that both requires sitting there paying attention to chat messages till its done (sitting in forced afk corner bad) and would make the rifle too weak when compared to the base sniper, since people would just choose to wait it out. 

**How improved**
With rebalanced stats it deals more damage than the initial shot but less than max stacks. With slower RoF, less ammo and additional effects per hit (knockdown, knockback, slow) remaining the same, it is a proper sidegrade with a niche against high hp targets, rather than overreliant on the stacking gimmick which would need to be fixed anyway.
At proposed stats TTK for warrior is lower than base (15s to 10s), same for rav (16s), both cases AMR leaves them crit, base kills them. Considering higher damage at higher current hp and effects I think it balances out.
- Flak ammo and aimed shot timer

Flak is currently easily the strongest sniper ammo, because of the ability to slow t3s. Such an insanely long duration is overpowered and leads to t3s taking far more damage than they would from the damaging basic ammo. On the other hand, non-aimed shot flak is so bad no one bothers firing it.
Aimed shot duration was overtuned since the unnecessary buff which made finding cover in time way too difficult and raised sniper dps too high.

**How improved**
Pretty straightforward, moved stats around so flak's no longer incredibly OP, but remains a good ammo type and reverted the aimed shot duration nerf.
<details>

</details>

# Changelog
:cl:
del: Removed smock, ghillie suit and helmet from spotter kit
del: Removed AMR stacking shot mechanics
qol: Sniper rifles use a blue marker for aimed shot to differentiate from spotters
balance: Sniper flak ammo armor penetration raised from 0 to 20, aimed shot slow halved to 2 for t3s and 3.5 for other caste and aimed shot damage raised from 55 to 110
balance: AMR aimed shot damage lowered for super low health targets but buffed against higher health (around 2nd shot with old focus)
balance: Raised M42A aimed shot time from 1.25 to 1.5 and lowered aimed shot aiming time bonuses from laser (0.6 to 0.7), spotter (0.5 to 0.7) and both (0.35 to 0.5).
/:cl:


Drulikar:
Find something other than removing teamwork from the game. DNM until you can find a solution with spotters.

Drulikar:
Past that I'm very unlikely going to accept focus fire removal too atleast not without some other core mechanic differentiating a sniper's weapon choice. As of right now you're just establishing one weapon will be superior to the other.

GoldenDarkness55:
> Find something other than removing teamwork from the game. DNM until you can find a solution with spotters.

Well they cannot buff the sniper above balanced values since we go back to why they need to be removed in the first place.
I could make aimed shot worse without a spotter, so it's balanced around the assumption the spotter's there if you want?

There's a reason the other specs don't have any spotter equivalent nor specific teamwork besides RPG reload - there is simply no room for this, never was and never will be without reworking specialists themselves or introducing balance issues.
As much as I like teamwork some features are just plain bad and ought to be removed: see carrier's shaman strain which operated on teamwork aswell.


> Past that I'm very unlikely going to accept focus fire removal too atleast not without some other core mechanic differentiating a sniper's weapon choice. As of right now you're just establishing one weapon will be superior to the other.

And which one would the superior be in this case? I don't think I can buff AMR any further dmg-wise nor do I think it needs any buffs.
In current state base sniper is way better, with this PR they're way closer to each other strength-wise to the point where I can't really pin one as better. The core mechanic is wall piercing and the ammo debuffs+stats are very different, one would have to replace aimed shot itself with something else entirely to go further.

TheManWithNoHands:
> Well they cannot buff the sniper above balanced values since we go back to why they need to be removed in the first place. I could make aimed shot worse without a spotter, so it's balanced around the assumption the spotter's there if that's fine??

Dont do that, that would only fully break spotter. Spotter is currently a feature that doesnt cause a negative for anyone. If the spec doesnt have a spotter, that doesnt impact them much.
All it does, is reward players for actuall teamwork. Wich is honestly the best way we can introduce this stuff.

You cant hidge the base performance of player A on the performance of player B. One player can improve the other, but never ever should you be required another player to function properly at all. Dont go even sligthly into that direction. It will only end with sniper&spotter turning into a toxic mess.

ThiccFucc:
So what you're saying is take out the whole AMR; can't disrupt teamwork if the team isn't there to begin with

GoldenDarkness55:
@Drulikar 
Alright, restored spotter but removed ghillie suit + helmet (NVGs remain), changed both rifle beams to blue (same as amr max used to be) to differentiate from spotter laser.
Changed the spotter aim time bonus to be the same as sniper laser alone (0.5 to 0.7 and raised laser 0.6 to 0.7) and when both are active the aim time is reduced to 50% rather than to 35%. 
**I.e normal sniper with a spotter went from 0.4375 to 0.75 and without 0.75 to 1.05. AMR spotter from 0.7 to 1, without 1.2 to 1.4.**

Role remains, there's more pressure on actually working together to make the 2 team gimmick more apparent and the aiming speed isn't as ridiculous. I think that's an improvement.

As for AMR vs normal sniper balance without focus fire, as I said I think they're fine right now with neither being obviously better - DPS vs dmg up front and weighing effects to preference.

Drulikar:
Why not at least keep some stacking functionality? E.g. first stack is worse damage wise than the M42A; but second stack is better damage wise than the M42A? The two snipers already have a solid identity I think in how they mechanically work and removing that mechanic will just make the AMR a lot more boring.

TheManWithNoHands:
> Why not at least keep some stacking functionality? E.g. first stack is worse damage wise than the M42A; but second stack is better damage wise than the M42A? The two snipers already have a solid identity I think in how they mechanically work and removing that mechanic will just make the AMR a lot more boring.

There is also the rework from backsea, wich had a more support focused aproach then just give it higher dmg.
I honestly think it be better to try the diffrent rework we have for this instead of trying to fix this rework attempt that just did not work out at all. The % dmg proofed to be too strong at its core, being that it bypasses the normal dmg calucations and armor system. It cant be worse then how the AMR is currently on the live buid.

Worst case scenario it doesnt work out just like this one, best case the other rework ends up working out better then this one.
It cant go worse.

GoldenDarkness55:
> Why not at least keep some stacking functionality? E.g. first stack is worse damage wise than the M42A; but second stack is better damage wise than the M42A? The two snipers already have a solid identity I think in how they mechanically work and removing that mechanic will just make the AMR a lot more boring.

I think the need for stacking makes it worse than M42A unless overcompensated into dmg as high as current live and even then it's more annoying/enforcing afk than likely to actually kill. I also don't like how that mechanic encourages focusing the same player, even if he pushes out with others it'll always be him that gets targeted and forced to heal (it's 2 shots at max stack before it resets). Just frustrating with little actual use.
Besides, I think it's a boring mechanic and doesn't really fit a powerful rifle like this. Rather than rifle specific I could see sniper in general requiring that kinda "warmup shot" that's weaker than normal if he hasn't fired for X  time to prevent offscreen surprise but I don't think it's necessary.

If it had to be differentiated I'd rather if it lost aimed shot and used the techweb railgun damange in a line gimmick with oppressor prae warning (or the thick wavy laser beam PMC sniper uses). Would fit in with the rifle's power without being too broken due to telegraphing. Though that's beyond me to code in.

chefoxara:
all this PR does is just nerf sniper specs and discourage team work, boring. + makes the spotter more killable by backliners (since you decided to remove the ghille suit for some reason) which means even less teamwork since a lurker can decide to just ambush them while they're busy spotting, or they'll get pushed since theyre not cloaked. >>> that will result in even less people going spotter/less snipers wanting spotters
just stop with these PRs, ugh. add something into the game instead of nerfing and removing things.

TheManWithNoHands:
Btw, the blue laser is used to highligth that the sniper is an AMR sniper. Specificly that the AMr snipers aimed shot is charged. Just FYI before you slap it on the M42A.

Red-byte3D:
fix changelog

GoldenDarkness55:
> fix changelog

Whats wrong with it? Adjusted the AMR aimed shot line to be clearer.

Nanu308:
Straight removal is by itself bad, besides this takesaway from teamwork aspect of one of the few spec roles that has it.

I'd rather have you look into nerfing the AMR which was on my list already, as i agree it is too strong.

GoldenDarkness55:
> Straight removal is by itself bad, besides this takesaway from teamwork aspect of one of the few spec roles that has it.
> 
> I'd rather have you look into nerfing the AMR which was on my list already, as i agree it is too strong.

@Nanu308 I think you misread since this no longer removes spotters. Only AMR focus fire mechanic which is whats causing the gun to be OP through damage stacking.

TheManWithNoHands:
> @Nanu308 I think you misread since this no longer removes spotters as a role. Only thing that's removed is the extra ghillie/helmet/backpack to avoid imitating sniper and AMR focus fire mechanic which is whats causing the gun to be OP through damage stacking.

... my guy, your still removing things.

code/datums/ammo/bullet/sniper.dm
```diff
@@ -63,18 +63,18 @@
...
+		var/slow_duration = 3.5
 		var/mob/living/L = M
 		if(isxeno(M))
 			var/mob/living/carbon/xenomorph/target = M
 			if(target.mob_size >= MOB_SIZE_BIG)
-				slow_duration = 4
+				slow_duration = 2
 		M.adjust_effect(slow_duration, SUPERSLOW)
-		L.apply_armoured_damage(damage, ARMOR_BULLET, BRUTE, null, penetration)
+		L.apply_armoured_damage(damage*1, ARMOR_BULLET, BRUTE, null, penetration)
```
Diegoflores31, line 77: why *1 ???

code/datums/ammo/bullet/sniper.dm
```diff
@@ -161,74 +161,32 @@
...
-					human_firer.client?.images -= human_firer.focused_fire_marker // Remove old marker
-					qdel(human_firer.focused_fire_marker)
-					human_firer.focused_fire_marker = focused_fire_marker_temp // Store new marker ref
-					human_firer.client?.images += focused_fire_marker_temp // Add new marker
-
-			amr_counter = min(amr.focused_fire_counter + 1, 3)
-			amr.focused_fire_target = WEAKREF(target_mob)
-
-		var/size_damage_mod = 0.8 // 1.8x vs Non-Xenos (225)
+		var/size_damage_mod = 0.6 // 1.6x vs Non-Xenos (200)
```
Diegoflores31, line 164: undocumented change this is for non xenos

code/datums/ammo/bullet/sniper.dm
```diff
@@ -161,74 +161,32 @@
...
 		var/size_current_health_damage = 0 // % Current Health calculation, only used for Xeno calculations at the moment.
-		var/focused_fire_active = 0 // Whether to try and use focused fire calculations or not, for that kind of target.
 		var/slow_duration = stopping_power // Based on damage dealt.
 
 		if(slow_duration <= 2) // Must be over 60 base damage.
 			slow_duration = 0
 
 		if(isxeno(target_mob))
 			var/mob/living/carbon/xenomorph/target = target_mob
-			size_damage_mod -= 0.2 // Down to 1.6x damage, 200.
```
Diegoflores31, line 208: readd this VAR and revert the old one if you want to lower it for xenos

Diegoflores31:
i think this PR should be splitted individually for AMR and M4A2 . its scope is too broad and it may end up in overnerfing

harryob:
pr has been on dnm for like a month with no movement, reach out to nanu to remove this if you want to work on it

PR #7347, opened 2024-10-20
Revert "Extracts some specific effects properties into an element, adds afterimage effect"

Strobia:
Reverts cmss13-devs/cmss13#7220
This heavily breaks the PB mechanic due to marines being able to attack the blood spray off of their targets https://streamable.com/y9x9f4 

🆑
fix: REMOVES BROKEN PR
/🆑


irakligabunia:
YOU SUCK AT GITHUB DELETE YOUR ACCOUNT 👎

Drulikar (changes requested):
Reversion looks unnecessary.

Consider instead adding `mouse_opacity = MOUSE_OPACITY_TRANSPARENT` to the effects that were converted from `code\game\objects\effects\temporary_visuals` by that PR.

PR #7357, opened 2024-10-22
Pill bottle emptying bugfix

SpypigDev:
# About the pull request

Fixes an annoying bug introduced in #7265 which caused pill bottles to run through two simultaneous emptying processes whenever used in the hand. First trying to empty their contents, then moving to their own pill drawing code, and trying to draw a single pill from the empty bottle.

This PR bypasses the typical emptying system, for pill bottles and packets, allowing a single pill to be drawn from a pill bottle in hand with Z, but still leaves the empty verb functional, so bottles can be emptied by selecting the option after right clicking.

While I recognize I am in the vast minority, some HM players still prefer not to use the belt system of drawing directly from pill bottles when treating patients.

Drawing a pill from a pill bottle held in the hand allows HMs to keep track of which medications have been given, and in what order.

#7265 broke this system by forcing HMs to start emptying pill bottles in the hand on a click, rather than drawing a single pill, making it exceptionally annoying to medicate wounded without belt mode, and near impossible to do so without a belt.

This PR allows both systems to co-exist, letting HMs draw a single pill from bottles held in the hand if they choose to do so, and still leaving the empty verb on pill bottles, so that their contents can be dumped through the right click menu

<details>

</details>

# Changelog
:cl:
fix: no longer forces pill bottles to begin emptying on a click
/:cl:


Drulikar (changes requested):
This would be better as a flag that you can just have pill bottles in their definition say they don't allow rather than having every single item check if its a pill bottle.

E.g. 

in storage.dm
```js
	if(length(contents))
		if(!(storage_flags & STORAGE_DISABLE_USE_EMPTY))
			empty(user)
		return
```

in equipment.dm
```js
/// Whether using this item will try not to empty it if possible
#define STORAGE_DISABLE_USE_EMPTY (1<<12)
```
and in bitfields.dm:
```js
	"STORAGE_DISABLE_USE_EMPTY" = STORAGE_DISABLE_USE_EMPTY,
```

and in firstaid.dm for `/obj/item/storage/pill_bottle`:
```js
	storage_flags = STORAGE_FLAGS_BOX|STORAGE_CLICK_GATHER|STORAGE_QUICK_GATHER|STORAGE_DISABLE_USE_EMPTY
```

Please also move all the `//` comments on those storage flags to be above them as autodoc comments `///`

PR #7359, opened 2024-10-22
Fixes Funny Wall Bug

Comxy:
# About the pull request

This PR stops wall turfs from reflecting bonus damage to mobs. This PR is tested and makes sure the damage multiplier isn't accounted for when damage gets reflected. 

Yesterday I got one shot by a XM51 as hivelord at close if not full health. I was wondering what happened and remembered that XM51 does bonus damage on walls and I build one under myself as hivelord. It turns out you can kill a queen in one shot with XM51 if they are in a resin wall.

I would make a video if I knew how to but you can test it for yourself pretty easily.

# Changelog
:cl:
fix: Fixes wall damage multipliers from reflecting to mobs.
/:cl:


cuberound:
it looks like your fix means that when there is the mob instead of damaging both for wall dmg you hit both for mob dmg?

Comxy:
> it looks like your fix means that when there is the mob instead of hitting both for wall dmg you hit both for mob dmg?

Yeah, it removes the wall multiplier, the damage goes only to the mob with the normal amount now

harryob (approved):

PR #7373, opened 2024-10-24
base rav slight buff

Red-byte3D:
# About the pull request

increases the stun from the base ravager dash

Base ravager is very single target, the only target you knockdown infront of suppressing fire getting up too fast isnt really a fun time for an alleged tier 3 supposed to take on multiple targets,

This should allow you to have an easier time to kill the target you have atleast dashed.

<details>

</details>

# Changelog
:cl:
balance: base ravager empowered charge stun is increased from 1.3 to 1.6
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/Ravager.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Ravager.dm
@@ -74,7 +74,7 @@
 /datum/behavior_delegate/ravager_base
 	var/shield_decay_time = 15 SECONDS // Time in deciseconds before our shield decays
 	var/slash_charge_cdr = 3 SECONDS // Amount to reduce charge cooldown by per slash
-	var/knockdown_amount = 2
+	var/knockdown_amount = 1.6
 	var/fling_distance = 3
 	var/empower_targets = 0
 	var/super_empower_threshold = 3
--- a/code/modules/mob/living/carbon/xenomorph/castes/Ravager.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Ravager.dm
@@ -74,7 +74,7 @@
 /datum/behavior_delegate/ravager_base
 	var/shield_decay_time = 15 SECONDS // Time in deciseconds before our shield decays
 	var/slash_charge_cdr = 3 SECONDS // Amount to reduce charge cooldown by per slash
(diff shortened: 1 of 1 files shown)


GoldenDarkness55:
1.3 is enough to get your charge reset on him before the marine gets up. The timer is a bit tight though so you should just set it to 1.5.
We were at 2 and it was way excessive because it's enough to 100-0 a marine before he has a chance to get up. Single click round removal bad.

Red-byte3D:
balance discussion is not for github

ihatethisengine:
Not sure if this change is the right way to buff base rav. Scissor cut you usually use against the dashed target already superslows your target for 3 seconds, which with a ton of damage from dash makes it fairly easy to finish them.

If I wanted to buff base rav I'd either buff its shield (you need to overextend to get empowerment, but the shield you get is not reliable enough to survive the overextention), or maybe increase CDR from slashes a bit (you already can do two dashes from a single empowerment, but only if you didn't waste a single second, which is fair, but I think it should be a bit easier to pull off. Right now if you attempt this you are risking to use your second dash at the very moment you lose your empowerment, both not getting the stun and wasting your ability, leaving you with no means to escape).

> enforcing murderballs as the only counter bad too.
Local prime rav doesn't know that the counter to base rav is to be in groups of less than 3 people.

GoldenDarkness55:
> balance discussion is not for github

im bringing it up cause you're trying tp reverse https://github.com/cmss13-devs/cmss13/pull/3729

Red-byte3D:
> Not sure if this change is the right way to buff base rav. Scissor cut you usually use against the dashed target already superslows your target for 3 seconds, which with a ton of damage from dash makes it fairly easy to finish them.
> 
> > enforcing murderballs as the only counter bad too.
> 
> Local prime rav doesn't know that the counter to base rav is to be in a group of less than 3 people.

Super slow doesn't necessarily mean you're not getting damaged or shot at, understand you and i can do this but the average player might panick when 4+ people are already shooting at him and the guy gets up too, if you can dm me on discord i can discuss this with you further since i know you play rav

Nanu308 (changes requested):
Increase the timer to 1.6, live server lag makes thing feel different than local.

Drulikar (approved):

PR #7376, opened 2024-10-25
Remote building nerfs

GoldenDarkness55:
# About the pull request

See video section for changed values in practice.

Adjusted queen remote building cooldown because it had none (it began the moment building started and finished when building was done) and raised the build time modifier by half to give players more time to react.
Slightly raised resin whisperer build time for the same reason.
Raised the cooldown multiplier for interrupting remote building so interrupting it is finally a valid counterplay that stops it for a time.

Initial 30 minute building buff is unchanged. Notable that hivelord and queen's building modifiers seem to function differently as they're around the same time in practice despite different values.

Queen's remote building was designed to help with building up defenses, but ended up as an incredibly cancerous ability constantly used on the frontline. Most of the time this is enough to almost by itself push the entire marine force back, because cleaning these weeds and fighting at the same time isn't possible.

Even if people dedicated themselves to breaking weeds to prevent it, both the build duration and cooldown made this incredibly difficult/impossible, turning their game into a 24/7 whack-a-mole until they got inevitably killed by one of the nearby xenos cause they were too preoccupied trying to halt the wall onslaught.

The very quick built time also sometimes leads to xenos dying because by the time they realize a wall's being built it's already finished.
It is simply not fun for either side and im confident there isn't a single good reason to keep it in current state. I'd rather deal with screech having a 14 tile range than something this anti-fun.

With proposed changes there is slightly more time to react, the wall spam itself is slower and interrupting it does put it on a reasonable cooldown. Should make this possible to play against if queens choose to use it offensively, without making the ability useless for its intended purposes.

<details>

https://github.com/user-attachments/assets/485f8e4f-ee52-4fdc-82f0-4712fce01b97

</details>

# Changelog
:cl:
balance: Cooldown multiplier for interrupting remote building raised from 2 to 3
balance: Queen remote building modifier raised from 1 to 1.5 (slower, around 2.3-2.5s), cooldown raised from 2s to 4s
balance: Resin Whisperer remote building modifier raised from 2 to 2.5 (slower, around 2.5s), cooldown raised from 1s to 2.5s
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_abilities.dm
@@ -50,11 +50,11 @@
 	action_icon_state = "secrete_resin"
 	ability_name = "projected resin"
 	plasma_cost = 100
-	xeno_cooldown = 2 SECONDS
+	xeno_cooldown = 4 SECONDS
 	ability_primacy = XENO_PRIMARY_ACTION_5
 
 	care_about_adjacency = FALSE
-	build_speed_mod = 1
+	build_speed_mod = 1.5
 
 	var/boosted = FALSE
 
@@ -75,11 +75,12 @@
 		boosted = TRUE
 		xeno_cooldown = 0
 		plasma_cost = 0
+		build_speed_mod = 1
 		RegisterSignal(owner, COMSIG_XENO_THICK_RESIN_BYPASS, PROC_REF(override_secrete_thick_resin))
 		addtimer(CALLBACK(src, PROC_REF(disable_boost)), boost_duration)
 
 /datum/action/xeno_action/activable/secrete_resin/remote/queen/proc/disable_boost()
-	xeno_cooldown = 2 SECONDS
+	xeno_cooldown = 4 SECONDS
 	plasma_cost = 100
 	boosted = FALSE
 	UnregisterSignal(owner, COMSIG_XENO_THICK_RESIN_BYPASS)
--- a/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
@@ -241,7 +241,7 @@
 	switch(X.build_resin(A, thick, make_message, plasma_cost != 0, build_speed_mod))
 		if(SECRETE_RESIN_INTERRUPT)
 			if(xeno_cooldown)
-				apply_cooldown_override(xeno_cooldown * 2)
+				apply_cooldown_override(xeno_cooldown * 3)
 			return FALSE
 		if(SECRETE_RESIN_FAIL)
 			if(xeno_cooldown)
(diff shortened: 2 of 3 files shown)


chefoxara:
im saying this as a MARINE MAIN
this is a team work game, marines will need to use their >TOOLS< and >TEAMWORK< to prevent queen remote building (which is rarely used anyways)
if we are going for 70/30 win ration xeno to marine or whatever, this doesnt make sense as a PR since the xenos are meant to OUTMATCH the marines
if marines PERFORM poorly and what they should properly, they should get outremote builded and killed/capped, they have breaching shotguns, they got flamers which prevent remote building, they got UBs, knifes, machetes, nades (TOOLS)

GoldenDarkness55:
> if marines PERFORM poorly in what they should do properly, they should get outremote builded and killed/capped, they have breaching shotguns, they got flamers which prevent remote building, they got UBs, knifes, machetes, nades (TOOLS)

it's not a matter of performing poorly, but the ability being overtuned - even if you counter it properly by constantly canceling the building, it simply doesn't matter due to how low the cooldowns and timers are.
i.e the intended counterplay is ineffective and this PR fixes that.

chefoxara:
it literally is a matter of performing poorly, i played queen, i spammed remote building, marines did good, they knifed everything i built, i could not outbuild them since knifing is faster than building
current remote building is FINE

 git gud

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_abilities.dm
```diff
@@ -50,11 +50,11 @@
 	action_icon_state = "secrete_resin"
 	ability_name = "projected resin"
 	plasma_cost = 100
-	xeno_cooldown = 2 SECONDS
+	xeno_cooldown = 4 SECONDS
```
Drulikar, line 53: Shouldn't this remain 2 seconds so that when build boost ends it then is 4 seconds?

Drulikar (changes requested):

code/modules/mob/living/carbon/xenomorph/abilities/queen/queen_abilities.dm
GoldenDarkness55: Build boost overrides to 0 cooldown slightly below, haven't changed that.
Didn't check if I can just set this to 0 and remove override without breaking something, but I think it'd hurt readability so would rather not.

Drulikar (approved):

Veglok:
Queen's building already gets nerfed 35 mins into the round, (which I guess can be lowered by 5-10 min.) making it slower to build and I believe if something cancels your remote building it will add a cooldown for your next structure, not to mention the plasma cost is quite high for remote building so you won't be able to use heals as a queen, which is 90% of the time more preferable so in short, if you're building past 35min into the round I'm not sure about the necessity of this nerf.

One thing that may be addressed is queens being able to build around survs super fast with almost no restrictions, but that's about it in my opinion.

GoldenDarkness55:
> Queen's building already gets nerfed 35 mins into the round, (which I guess can be lowered by 5-10 min.) making it slower to build and I believe if something cancels your remote building it will add a cooldown for your next structure, not to mention the plasma cost is quite high for remote building so you won't be able to use heals as a queen, which is 90% of the time more preferable so in short, if you're building past 35min into the round I'm not sure about the necessity of this nerf.
> 
> One thing that may be addressed is queens being able to build around survs super fast with almost no restrictions, but that's about it in my opinion.

There actually was no difference in build speed or cooldown regardless of the boost - the 2s cooldown was over by the time you finish building becoming 0 in practice and the cancel penalty was too low since it bases off said cooldown being multiplied.
Plasma costs somewhat solve themselves since it's slower

PR #7400, opened 2024-10-27
Instant delimb nerf (for real this time)

ihatethisengine:
# About the pull request
1) I got two tap delimbed :sob: 
2) https://github.com/cmss13-devs/cmss13/pull/6240 failed to solve the issue as Nivrak failed to read the code properly. As the delimb check happens AFTER the frac check, you just can get frac & delimb with the same hit. While this indeed reduced the chance of getting delimbed (as now you need to roll both for frac and delimb), it is still pure RNG which that PR wanted to avoid. (Also the only thing that KIND OF saves you from first tap delimb is previous_brute > 0 check, but as long as you have 0.000001 brute on the limb it's irrelevant)

1) I got two tap delimbed :sob: 
2) Instant delimbs based on nothing but rng from one slash (yeah technically it's not true as you need at least 0.000001 brute on the limb, but it was so when Nivrak made his PR too) are unfun to play against

<details>
I did test it by spawning myself as rav and slashing 100 naked humans.

</details>

# Changelog

:cl: ihatethisengine
balance: limbs are now only be delimb-able if they were fractured BEFORE the hit
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -277,6 +277,7 @@
 
 	var/previous_brute = brute_dam
 	var/previous_burn = burn_dam
+	var/previous_bonebreak = (status & LIMB_BROKEN)
 
 	var/is_ff = FALSE
 	if(istype(attack_source) && attack_source.faction == owner.faction)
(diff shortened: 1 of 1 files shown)


Git-Nivrak:
I didn't fail to read the code - the intention was that you cannot get delimbed in 1 hit unless your limb is at max damage (which is something you can avoid to an extent) which raised the minimum amount of slashes required to delimb a "healthy" non-fractured limb from 1 to 2

ihatethisengine:
> the intention was that you cannot get delimbed in 1 hit unless your limb is at max damage or fractured

But you can? If you have some damage on your limb, you can get delimbed in 1 hit if this hit also fracs your limb. And if you had no damage (at all), your limb already couldn't be delimbed with one hit even before your PR.

harryob (approved):

Git-Nivrak:
That is just not true though, every limb has a min_break_damage, if you are below that your limb will never break in one slash because this is checked before applying damage, if you don't believe me go to local and set a rav slash damage to 2000 and check if it gets delimbed.
If your limbs are above the damage threshold that is a risk you take and can interact with.

ihatethisengine:
> if you don't believe me go to local and set a rav slash damage to 2000 and check if it gets delimbed.

It won't be delimbed because of this line first of all: https://github.com/cmss13-devs/cmss13/pull/7400/files#diff-887b725f2d3588dfa6fce3c02f9a3539f118dda10340c170cd7d7e0306b78b94R373 (again, you already need and needed at least SOME (more than 0) damage on your limb for it to become delimbable)

For all delimbable limbs (except head) min_broken_damage is about 20 dmg, which is... really low.

Nanu308:
I will balance approve this based on the fact the prior pr specifically says fracture is needed prior, and that I do agree with the points I hate this engine is making.

PR #7406, opened 2024-10-27
Fixes additional projectiles

Ansekishokuu:
# About the pull request

Fixes Issue #3053 (and potentially #1764)
This PR aims to fix a bug that has been around for a long time which is explained below.
Changes tested.

The reason why buckshots and any other multi-projectile weapons (or explosions) are not working correctly is because of the PR that added random speed variance for projectiles (#1608). This PR caused the buckshot's additional projectiles to either lag behind or go faster than the main projectile which stuns the target, causing the additional projectiles to not hit the target most of the time.
However there is also another factor which causes the additional projectiles to not hit the target. That is the fact that when the main projectile hits and stuns the target (makes them fall down to the floor), the additional projectiles miss the target because they are on the floor and go over the target.
This causes the shotgun to not deal the intended damage because only the main projectile hits and barely does any damage.

**This is a bugfix. It doesnt buff the projectile damage in any way. It technically also only affects stunnable targets because the problem is targets being stunned and the projectiles missing them.
For example; Hivelords dont get stunned, therefore they dont have an issue with the additional projectiles.**

A buckshot shouldn't miss a target that you are adjacent to.
The bug causes the buckshot to be completely useless at times, because it doesnt work as intended.
Even though you are adjacent to your target, the projectiles miss the target which clearly shouldn't happen.
If you are still confused as to what this PR fixes exactly, videos below should show the problem.

**These are before / after footages. First video is before the fix, second video is after the fix.**
<details>

Before

https://github.com/user-attachments/assets/364bb19e-25a5-4f6f-9dd3-ce4beea0fcfa

After

https://github.com/user-attachments/assets/b797c3b3-97d2-476b-bac6-f450b976efe6

</details>

# Changelog
:cl: Ansekishoku
fix: Buckshot additional projectiles no more miss the target when they get stunned by the main projectile.
/:cl:


code/modules/projectiles/projectile.dm
```diff
@@ -241,8 +242,12 @@
 	src.speed = speed
 	// Randomize speed by a small factor to help bullet animations look okay
 	// Otherwise you get a   s   t   r   e   a   m of warping bullets in same positions
-	src.speed *= (1 + (rand()-0.5) * 0.30) // 15.0% variance either way
-	src.speed = clamp(src.speed, 0.1, 100) // Safety to avoid loop hazards
+	if (randomize_speed)
+		src.speed *= (1 + (rand()-0.5) * 0.30) // 15.0% variance either way
+		src.speed = clamp(src.speed, 0.1, 100) // Safety to avoid loop hazards
```
Doubleumc, line 247:
```suggestion
```
Clamps it again 3 lines down outside the `if` block.

Ansekishokuu: Thanks for the correction, missed that.

harryob (approved):

PR #7433, opened 2024-10-29
Forecon SMG ammo is moved to their belt

private-tristan:
# About the pull request
Forecon now gets a rig filled with their m39 and mags, instead of having just the m39 rig itself. This touches on both the survivor and adminspawn ERT variants.

being forced to toss away all your sidearm mags just so that you can have space to hold stuff kinda sucks, especially when every other sidearm spawn that forecon has keeps its' mags in the belt

Inventory before, note how there's no storage unless you toss out mags (or move your radio)

image

Inventory after:

image

# Changelog
:cl:
balance: Forecon M39 spawn now has its spawn ammo in belt, instead of in satchel.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -1307,6 +1307,16 @@
 			"icon_x" = -11,
 			"icon_y" = -5))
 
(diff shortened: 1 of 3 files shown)


Nanu308 (approved):
lgtm

Drulikar (approved):

PR #7438, opened 2024-10-30
Tacmap Polish

Drulikar:
# About the pull request

This PR reworks how the minimap system crops tacmaps to allow maps larger than the screen. For now the canvas size was increased from 480 to 600 but it is trimmed down to 512. It is also much more strict about trimming the edge of the map. This allows the Almayer to be fully displayed in the observer on screen tacmap even though it is larger than the screenspace by scrolling the map left and right automatically. (I originally wanted to just scale it down enough to fit on the screen, but this resulted in inconsistent wall thickness).

The tacmap TGUI is also more polished meaning there is less variation between each panel and more of the window space is used to accommodate the maps.

The only sprite change is changing a blank canvas image to be larger.

This also turns out to fix an issue where predators and observers were using the same tacmap because now the shifting value is part of the hash making observer maps unique. They cannot share the same map because they both need unique screen_loc values. I am not completely convinced this is the only issue that might prevent ghost tacmaps from opening though; but will need more investigation.

Allows us to more easily support large maps (such as now the Almayer and its AI Core) so they can still be displayed on the minimap and polishes the tacmap TGUI panels to be more consistent and utilize space more efficiently. 

<details>

shift
shift2
minimap
image
image
image
image
image
image
image

</details>

# Changelog
:cl: Drathek
ui: The ghost observer tacmap will now scroll both horizontally and vertically as needed to display a large map.
ui: The TGUI panels for tacmaps and drawings have been polished to utilize more minimap space fully.
refactor: Refactored minimap generation code to calculate min and max faster, to more aggressively trim minimaps, and to support larger minimaps (was 480px now 512px).
fix: Fixes an issue where predator tacmap would take over the ghost tacmap preventing it from opening for observers
/:cl:


Blundir (approved):

Wintermote:
Could just copy google maps: Click on tacmap, drag to pan around, mousewheel to zoom in and out.

harryob (approved):

PR #7440, opened 2024-10-30
doubles facehugger generation rate

Detective-Google:
# About the pull request

doubles the facehugger automatic generation rate

every xeno I've heard feedback from since the change has said it's too low

<details>

</details>

# Changelog
:cl: detectivegoogle
balance: morphers generate huggers twice as fast
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/special/egg_morpher.dm
+++ b/code/modules/cm_aliens/structures/special/egg_morpher.dm
@@ -20,7 +20,7 @@
 	///Datum used for mob detection.
 	var/datum/shape/range_bounds
 	///How long it takes to generate one facehugger.
-	var/spawn_cooldown_length = 120 SECONDS
+	var/spawn_cooldown_length = 60 SECONDS
 	COOLDOWN_DECLARE(spawn_cooldown)
 
 


HeresKozmos:
lgtm

Red-byte3D:
You could just make them generate huggers faster if you put bodies in them, this way you can afford a bigger buff by also adding a way of putting work in to get better rewards

Nanu308:
Xeno captures goes

Nanu308 (approved):
Live check will commence later date and see if it has issues with carriers etc. being unneeded

Drulikar (changes requested):
Do you think you could tie this to whether the queen is on ovi or not? As in buff is only active when she is ovi?

Detective-Google:
> Do you think you could tie this to whether the queen is on ovi or not? As in buff is only active when she is ovi?

I think I could probably get something to that effect to work but making sure it only affects the correct hive, etc is bug fixing out of the scope for this PR, which was really just a webedit number tweak to shush the people decrying the PR adding this system as "ruining" carrier. 

in less words, I don't have any interest in devoting a lot of time to this and you're welcome to close it for someone else to change instead.

Drulikar:
See above then

PR #7441, opened 2024-10-30
Prevents headsets from accepting duplicate keys

Ansekishokuu:
# About the pull request

Fixes issue #7375 
Prevents headsets from accepting multiple encryption keys of the same kind.

Changes tested.

Fixed headsets accepting the same type of encryption key.

<details>

</details>

# Changelog

:cl: Ansekishoku

fix: Headsets no longer accept multiple encryption keys of the same kind.

/:cl:


code/game/objects/items/devices/radio/headset.dm
```diff
@@ -165,6 +165,11 @@
 			to_chat(user, SPAN_NOTICE("This headset doesn't have any encryption keys!  How useless..."))
 
 	if(istype(W, /obj/item/device/encryptionkey/))
+		for (var/obj/item/device/encryptionkey/key in keys)
+			if (key.type == W.type)
+				to_chat(user, SPAN_NOTICE("There is already the same type of encryption key in the headset!"))
```
vero5123, line 170: seems off, feel like this would sound better, untested btw. 
```suggestion
				to_chat(user, SPAN_NOTICE("\The [W.name] is already installed on this device."))
```

code/game/objects/items/devices/radio/headset.dm
```diff
@@ -165,6 +165,11 @@
 			to_chat(user, SPAN_NOTICE("This headset doesn't have any encryption keys!  How useless..."))
 
 	if(istype(W, /obj/item/device/encryptionkey/))
+		for (var/obj/item/device/encryptionkey/key in keys)
```
vero5123, line 168:
```suggestion
		for (var/obj/item/device/encryptionkey/key as anything in keys)
```
I haven't looked at the code but if the only type in keys are of type /obj/item/device/encryptionkey you can use as anything, it's just a micro opt.

vero5123: It's been a minute since I touched dm but irrc there is probably a macro to check if types are equal instead of x.type == y.type like is_type.

code/game/objects/items/devices/radio/headset.dm
Ansekishokuu: Yeah, thanks for the correction! I tested this change and its better this way.

code/game/objects/items/devices/radio/headset.dm
Ansekishokuu: Also approving this change. 

Thanks for the macro check tip, changing that section too.

harryob (approved):

PR #7456, opened 2024-10-31
updates barricade costs (sorry kids)

thwompa:
# About the pull request
increases cade costs. 

metal, 4->8
folding metal, 6->10
plasteel, 8->10
folding plasteel, 6->12

makes it far more difficult to create fobs with six+ layers of cadelines. this is bad for marines to play with, and bad for xenos to play against. it is pretty unfun to have a drawn out siege for 30+ minutes. i like a good last stand as much as the next guy, but every other round shouldn't look like whiskey outpost.

if you have a complaint or want to send me death threats, my dms are open

<details>

image

</details>

# Changelog
:cl:
balance: barricade construction costs increased.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/sheets/sheet_types.dm
+++ b/code/game/objects/items/stacks/sheets/sheet_types.dm
@@ -12,8 +12,8 @@
  */
 GLOBAL_LIST_INIT_TYPED(metal_recipes, /datum/stack_recipe, list ( \
 	new/datum/stack_recipe("barbed wire", /obj/item/stack/barbed_wire, 1, 1, 20, time = 1 SECONDS, skill_req = SKILL_CONSTRUCTION, skill_lvl = SKILL_CONSTRUCTION_TRAINED), \
-	new/datum/stack_recipe("metal barricade", /obj/structure/barricade/metal, 4, time = 2 SECONDS, one_per_turf = ONE_TYPE_PER_BORDER, on_floor = 1, skill_req = SKILL_CONSTRUCTION, skill_lvl = SKILL_CONSTRUCTION_TRAINED, min_time = 1 SECONDS), \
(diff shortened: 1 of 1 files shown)


Wintermote:
We've slowly crept into sprawling cadelines due to a combination of things making marines ineffective without them. 3 years ago the fobs we have today would be impenetrable fortresses that xenos of the time would never be able to get through. Today, xenos have so many tools that 5 layers of cades protects as well as 1 layer used to.

You can view this as a sort of balance minimum; where the game mechanics continually 'roll downhill' to the minimal tolerable conditions. I don't think you can presently cut fob size in half and still have balance; there need to be additional changes that make cadelines more effective or xenos less able to destroy them.

Wintermote:
Also important to note this is going to have a huge impact on making cades anywhere outside fob, as engineers cannot carry large stacks of metal on them.

thwompa:
image

efzapa:
basically what Winter said but 6 cades per stack (not even including the metal also needed for razor wire probably bringing it down to like 4 or 5) effectively makes it so you cannot cade and hold the frontline unless you bring 4 stacks of metal, which is just not feasible for almost any CT

cuberound:
off topic but, sandbag rejoice, I quess you are the only option thanks to flying under radar and deployable baricades as now you are cheaper then normal cades or right and CODEWISE there is an issue a big one same as when the plasteel cade was added you updated COSTS but no RETURNS when taken apart

Drulikar:
FWIW this has merit at least for high pop, but this currently can't be changed until at least first drop fob set up can be faster and more streamlined. Put a pin in it.

cuberound:
this PR makes it so that baricades do not give back the materials used for creating them, adjust stack_amount and destroyed_stack_amount to adress the issue or state in your PR that you intend on certain cades to not give back the materials that were used to build them when taken apart using tools as it is standard now and past instances of such mistakes were marked as bug and fixed

PR #7458, opened 2024-10-31
CMB Spam Update - CL can now print Blackmarket Scanners from their Briefcase

efzapa:
# About the pull request

CL can now print Blackmarket Scanners from their Briefcase for 50 points each

Adds more CL fluff, since the role is in desperate need of mechanical things to do. This should give the CL the option to spice up the round for both them and req should they both choose.

<details>

</details>

# Changelog
:cl:
add: CL can now print Blackmarket Scanners from their Briefcase for 50 points each.
/:cl:


deathrobotpunch:
I think buying one should have some sort of penalty like locking your briefcase or something

efzapa:
> The blackmarket scanner used to always be available, then it was made rare so it wouldn't happen every round. This would just cause it to happen every round again.

"rare" as in praying for RNG to be on your side
if there was a definitive way to get a scannner, even if it was difficult it would be fine
but banking a fun feature on pure RNG is a bad way to do it

in any case we can TM it, if it is abused then I can scrap this PR and make another rebalancing BM

efzapa:
actually gonna work on another PR that should fix the abundance problem then just PR that instead of this

PR #7460, opened 2024-10-31
Fixes being able to use smartpack abilities while dead

### LINKED ISSUE #6956: Smart pack abilities can be used while dead
### Testmerges

N/A

### Round ID

Multiple

### Description of the bug

You can use synth smart pack abilities (like repair and protection mode) while dead, via the action buttons top left 

You shouldn’t be able to use them 

1. Be synth with charger smart pack
2. Die
3. Use repair mode and witness yourself be repaired (not revived)

### Issue Bingo

Ansekishokuu:
# About the pull request

Fixes issue #6956
(Synths can repair themselves while dead with the smartpack.)

Changes tested.

Smartpack should not be able to function if the user is dead.

<details>

</details>

# Changelog

:cl: Ansekishoku
fix: Smartpack abilities no longer function if the user is dead.
/:cl:

Closes #6956


code/game/objects/items/storage/smartpack.dm
```diff
@@ -263,7 +265,8 @@
 
 
 /obj/item/storage/backpack/marine/smartpack/proc/repair_form(mob/user)
-	if(!ishuman(user) || activated_form || repairing)
+	if(!istype(user) || activated_form || repairing || user.stat == DEAD)
```
Drulikar, line 268: You ought to either to type out the user var more, or restore the old ishuman check here.

Drulikar (changes requested):

code/game/objects/items/storage/smartpack.dm
Ansekishokuu: The others are istype, i thought it would cause an issue if it was inshuman since synths arent considered humans.
But i will change it back to ishuman, maybe i am overthinking it!

Drulikar: Just for clarity the istype check here since `repair_form(mob/user)` user is typed to be a `/mob` it will ensure that user is a mob type. Whereas for `protective_form(mob/living/carbon/human/user)` user it typed to be a `/mob/living/carbon/human` so an istype check will ensure it is a human.

Drulikar (approved):

code/game/objects/items/storage/smartpack.dm
```diff
@@ -176,7 +176,8 @@
 	update_icon(user)
 
 /obj/item/storage/backpack/marine/smartpack/proc/protective_form(mob/living/carbon/human/user)
-	if(!istype(user) || activated_form || immobile_form)
+	if(!istype(user) || activated_form || immobile_form || user.stat == DEAD)
```
vero5123, line 179: mob.is_dead() exists, not that u have to use it.

Ansekishokuu: There wouldn't be a difference between `mob.is_dead()` and `user.stat == DEAD` if we are talking about synths since synths dont fall unconscious.

If there is however, please let me know and thanks for your attention.

PR #7466, opened 2024-11-01
UPP Dropship Morana & Devana | USCM DS3 Saipan Fix

Steelpoint:
# About the pull request

This PR adds two new UPP dropships to the game, the Morana & Devana. They are fully functional dropships intended for use by the UPP during admin events. They are linked to only be able to use UPP-aligned landing zones by default, this ensures only UPP personnel can control the craft. This can be changed by admins for events.

Both dropships are almost identical outside of some sprite differences. 

This PR also fixes the Saipan to make it useable, as previously the ship was impossible to spawn.

Many thanks to @Zenith00000 for the sprites. 

This grants additional options for event runners to employ the use of the UPP without needing to use a USCM dropship that has been modified for use. 

(Old rear-hatch sprite)
Screenshot 2024-11-02 00 48 27

# Changelog
:cl: Steelpoint, Zenith
mapadd: Two UPP dropships have been added for event use, the Morana and the Devana.
maptweak: Fixes a issue preventing the Saipan, a USCM dropship, from being spawned by admins for events. 
/:cl:


Blundir (changes requested):
Firstly, we've discussed the names already, choose something else, secondly, south door sprites are a placeholder, make or find the ones that aren't

Wintermote:
I think this would be a good opportunity to cut down on the dropship size by reducing the walls to something like glass windoors where you can stand in the same tile as the wall.

Steelpoint:
I have updated the rear-hatch sprites as requested. 

I am not sure what alternative names you would prefer. I initially presented the idea of explorer themed ship names of a Soviet theme as a alternative to the USCM dropship convention of picking names of famous US battles. But I'm open to alternative if the names could be different.

Steelpoint:
Updated the names to the Morana and Devana, as per the discussion I've gleaned. More in-line with Slavic mythical entities.

ihatethisengine:
> I initially presented the idea of explorer themed ship names of a Soviet theme as a alternative to the USCM dropship convention of picking names of famous US battles. But I'm open to alternative if the names could be different.

I mean you already used Rostock  for the main ship name, which is a German city. Can go further just with UPP cities name. According to wiki UPP is Russia, Germany and Spain. In one of the events we called the ship we had "Moscow". The second ship can be called "Madrid" by the same logic. Or if you want to go with less known cities then "Irkutsk" and "Granada", for example. I think that's better than Slavic myth things no one knows about in the faction where half of the countries are not in fact Slavic.

irakligabunia:
The names are cool, but still want to throw my 2c in

When I was drafting something alongside this I had the names picked out to be Bagration and Borodino; 
I had an idea to put livery on them regarding these two, both homages to the most important parts of Russia's (relatively) recent history.

Steelpoint:
Using city names for capital ships feels appropriate.

I'm happy with the name I've picked. Changing the names is a time consuming process so I'd rather not make another change.

Nanu308 (approved):
Mapping lgtm, besides lack of cameras, but that would maybe require code to split them from the USCM Camera network etc.

Regardless, 👍

harryob (approved):

Blundir (approved):

cuberound:
better design then USCM DS in regards to attachment spots placement

PR #7475, opened 2024-11-02
tgui body color/type/size picker 

harryob:
vido

https://github.com/user-attachments/assets/be1b699a-4bda-4343-8d13-90f979ff19ac

:cl:
add: a new tgui with better previews for the different skin colors / body types / body sizes
/:cl:

### DIFF
--- a/code/modules/client/color_picker.dm
+++ b/code/modules/client/color_picker.dm
@@ -66,6 +66,9 @@
 	if(!ui)
 		ui = new(user, src, "BodyPicker", "Body Picker")
 		ui.open()
+		ui.set_autoupdate(FALSE)
+
+	winset(user, ui.window.id, "focus=true")
 
 /datum/body_picker/ui_state(mob/user)
 	return GLOB.always_state
--- a/code/modules/client/color_picker.dm
+++ b/code/modules/client/color_picker.dm
@@ -1,0 +1,71 @@
+/datum/body_picker/ui_static_data(mob/user)
+	. = ..()
+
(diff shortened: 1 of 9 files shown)


code/modules/client/color_picker.dm
```diff
@@ -0,0 +1,71 @@
...
+	return TRUE
+
+/datum/body_picker/tgui_interact(mob/user, datum/tgui/ui)
+	. = ..()
+
+	ui = SStgui.try_update_ui(user, src, ui)
+
+	if(!ui)
+		ui = new(user, src, "BodyPicker", "Body Picker")
+		ui.open()
```
Drulikar, line 68: You may also want to force focus here too because if its already open but behind the window it will feel like to the user that its broken when they need to move the window around. As far as I can tell autoupdate is unnecessary unless you want it to be annoying about staying ontop of everything :P

```suggestion
		ui.open()
		ui.set_autoupdate(FALSE)

	winset(user, ui.window.id, "focus=true")
```

Technically though if they reload the slot it will have out of date data in the UI, but it will update if they change anything so I'm not concerned.

Drulikar (changes requested):
Otherwise nothing else sticks out to me. Great job with this o/

code/modules/client/color_picker.dm
harryob: autoupdate is all good yeah, slots are a pain. ty

Drulikar (approved):

PR #7487, opened 2024-11-04
CC Ovi Fix

BOBAMAx:
# About the pull request

You can now ovi/build hives anywhere in the atmospheric processor on CC

being able to ovi and build the core in caves is good

# Changelog

:cl: BOBAMA
balance: You can now go on Ovi and build Hive Cores anywhere in the Atmospheric Processor on Chance's Claim
/:cl:

### DIFF
--- a/code/game/area/LV522_Chances_Claim.dm
+++ b/code/game/area/LV522_Chances_Claim.dm
@@ -403,6 +403,7 @@
 	ceiling = CEILING_REINFORCED_METAL
 	ambience_exterior = AMBIENCE_SHIP
 	minimap_color = MINIMAP_AREA_ENGI
+	unoviable_timer = FALSE
 
 /area/lv522/atmos/outdoor
 	name = "Atmospheric Processor - Outdoors"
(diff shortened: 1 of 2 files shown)


Drulikar (changes requested):
These are the areas you altered (including by inheritance): 
image

You will likely want to include `north_east_street` as well if you're going to have an island out there that can be ovi (`atmos/cargo_intake`).

Of note, if a map does not specify any unoviable_timer, it will at a minimum include places w/ queen spawn landmarks so thats why there were any places.

BOBAMAx:
just made it it's own container area that's a child of outdoors/colony_streets. both those yellow areas are actually the same area, so that island is probably just another oversight from the rework as it used to be directly connected to the reactor. Doesn't make any sense to be a child of atmos as it's not in atmos. Also I feel bad I've autorequested nanu like 4 times this week lol.

cc

Drulikar (approved):

Drulikar:
Double checked: 
image

PR #7493, opened 2024-11-04
hair/facial hair character setup picker

harryob:
video

https://github.com/user-attachments/assets/36c68a58-2ecf-4847-85db-ccf519ff42d4

:cl: harryob, drathek
ui: replaces the lists for hair/facial hair with images in a separate picker
ui: replaces the eye color picker with a tgui color picker
code: added unit test for duplicate sprite accessories (hair)
/:cl:

color picker modal is from https://github.com/BeeStation/BeeStation-Hornet/pull/9417 thank you


Drulikar (changes requested):
Been tinkering with this for a bit:
- Using Enter/Esc on the hex field didn't do anything, it now would
- Entering any value for the color fields would result in NaN numbers because of our different callback argument count
- crt blue scrollbars are now not as dark
- confirm button is moved from the top to the bottom right of the modal and is instead part of the modal optionally if onConfirm is set.
- removed logger from ColorPickerModal too

I was unsuccessful getting tab order to work to go from one input field to the next (it really wants to tab focus the sliders)

I did experiment with eye color using tgui color picker and that seems to work too. Though you might want to tell this to use the large buttons by default maybe?

image
image
[patch.diff.txt](https://github.com/user-attachments/files/17624839/patch.diff.txt)

harryob:
addressed, also added gradient style/color to the picker menu

Drulikar (changes requested):
- Some hair seems to be multi-selecting in the UI e.g.:
image
image

- Minimum height is not enough for color picker for females: 
image

- Facial hair seems to still be displayed (in the slot UI not here) despite changing genders even though only shaved is gender neutral?
- No hair dye results in a 0: 
image
- First open of dropdown menu freaks out the scrolling. I've encountered this before but I don't know any solution except for the dropdown to not be at the bottom of the window; but its likely a bug w/ Dropdown itself: 
jump

harryob:
all addressed, bar
> Facial hair seems to still be displayed (in the slot UI not here) despite changing genders even though only shaved is gender neutral?

the picker doesn't autoupdate, so if you switch genders in the main ui it won't force a refresh of the picker to hide the menu. you can't pick any invalid hair styles, though

tgui/packages/tgui/interfaces/HairPicker.tsx
```diff
@@ -0,0 +1,226 @@
...
+            name="Facial Hair"
+            icon={facial_hair_icon}
+            hair={facial_hair_styles}
+            active={facial_hair_style}
+            color={facial_hair_color}
+            setColor={setColorPicker}
+            action="facial_hair"
+          />
+        )}
+        {gradient_available && (
```
Drulikar, line 107:
```suggestion
        {!!gradient_available && (
```
This also was making 0s

tgui/packages/tgui/components/Dropdown.tsx
```diff
@@ -143,7 +143,9 @@ export function Dropdown(props: Props) {
       scrollTo(selectedIndex);
     }
 
-    innerRef.current?.focus();
+    innerRef.current?.addEventListener('load', () => {
+      innerRef.current?.focus();
+    });
```
Drulikar, line 148: I'm not seeing any benefit from this change. Probably not necessary unless you saw an improvement. I should probably just explore what TG has done to Popper and Dropdown since the parity port. E.g. https://github.com/tgstation/tgstation/pull/80556

tgui/packages/tgui/interfaces/HairPicker.tsx
```diff
@@ -0,0 +1,226 @@
...
+    gradient_color,
+  } = data;
+
+  const [colorPicker, setColorPicker] = useState<
+    'hair' | 'facial_hair' | 'gradient' | false
+  >(false);
+
+  let height = 340;
+  if (facial_hair_styles.length > 1) {
+    height = height + 300;
```
Drulikar, line 61:
```suggestion
    height = height + 310;
```
image
instead of
image

code/modules/client/preferences.dm
Drulikar, line 1: For getting the pickers to refresh when backend changes: 
[patch.diff.txt](https://github.com/user-attachments/files/17656080/patch.diff.txt)

Drulikar (changes requested):

tgui/packages/tgui/components/Dropdown.tsx
harryob, line 148:
```suggestion
    innerRef.current?.focus();
```

harryob:
> I'm not seeing any benefit from this change (it still spazzes first open).

coulda sworn, but that's a failure of my debugging

Drulikar (approved):

PR #7505, opened 2024-11-05
Revert "Sentinel buffs"

irakligabunia:
### **About the Pull Request**

Reverts the Sentinel Buffs made by GoldenDarkness, no changes made by me.

### **Explain why it's good for the game**

Due to the lack of testing and review on the "sentinel buffs" PR, the caste Sentinel has went from being a support T1 caste which was relatively strong due to its stun and long range slow to a one man machine.

As it currently stands: 
1. One Paralyzing slash guarantees you a capture 100% of the time
2. Due to the buffs, you can superslow somebody for even longer despite it previously requiring you to keep attacking the target; making it very, very strong at least in my opinion
3. You get stunned faster due to "Sentinel paralyzing slash stuns after 3 seconds instead of 4 and lasts 2.5 seconds from 2."

TL;DR It has made sentinel into an unbeatable machine that with one ability can win an engagement.

### **Changelog**

:cl: stalkerino
balance: reverts sentinel buffs
/:cl:

### DIFF
--- a/code/__DEFINES/weapon_stats.dm
+++ b/code/__DEFINES/weapon_stats.dm
@@ -66,7 +66,7 @@ It DOES NOT control where your bullets go, that's scatter and projectile varianc
 ////SCATTER////
 */
 
-#define SCATTER_AMOUNT_NEURO 45
+#define SCATTER_AMOUNT_NEURO 60
 #define SCATTER_AMOUNT_TIER_1 15
 #define SCATTER_AMOUNT_TIER_2 10
 #define SCATTER_AMOUNT_TIER_3 8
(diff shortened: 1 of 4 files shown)


irakligabunia:
cool PR dude

TheManWithNoHands:
The Pr your reverting touched more then just paralizing slash.
If you only have problems with that, only adjust the slash.

Sentinel did get a ton PRs in a bunch though, so adjusting sentinal it a bit is a good idea.
Also keep in mind that Paralizing slash recently got a major QoL change making it much easier to use. Something the buff to it, didnt account for.

Drulikar (changes requested):
A reversion is dead in the water. But if you want to tweak the values have at it. But no matter what a buff is necessary over old.

Drulikar:
Issue w/ CL btw was that you used emoji for one and not for the other. Both must match.

irakligabunia:
fuck my contributor life

PR #7524, opened 2024-11-07
implements a new loadout picker tgui menu

harryob:
MORE tgui slop yum yum more `<DmIcon />` please!

vido

https://github.com/user-attachments/assets/385f3aab-0df5-478b-b2e1-b7d0d85be67c

:cl:
ui: adds a prettier custom loadout menu with images
/:cl:

### DIFF
--- a/code/modules/client/loadout_picker.dm
+++ b/code/modules/client/loadout_picker.dm
@@ -1,0 +1,87 @@
+/datum/loadout_picker/ui_static_data(mob/user)
+	. = ..()
+
+	.["categories"] = list()
+	for(var/category in GLOB.gear_datums_by_category)
+		var/list/datum/gear/gears = GLOB.gear_datums_by_category[category]
+
+		var/items = list()
+
+		for(var/gear_key as anything in gears)
+			var/datum/gear/gear = gears[gear_key]
+			items += list(
+				list("name" = gear.display_name, "cost" = gear.cost, "icon" = gear.path::icon, "icon_state" = gear.path::icon_state)
+			)
+
+		.["categories"] += list(
+			list("name" = category, "items" = items)
+		)
+
+	.["max_points"] = MAX_GEAR_COST
+
+/datum/loadout_picker/ui_data(mob/user)
+	. = ..()
+
+	var/datum/preferences/prefs = user.client.prefs
+	if(!prefs)
+		return
+
+	var/points = 0
+
+	.["loadout"] = list()
+
+	for(var/item in prefs.gear)
+		var/datum/gear/gear = GLOB.gear_datums_by_name[item]
+		points += gear.cost
+
+		.["loadout"] += list(
+			list("name" = gear.display_name, "cost" = gear.cost, "icon" = gear.path::icon, "icon_state" = gear.path::icon_state)
+		)
+
+	.["points"] = points
+
+/datum/loadout_picker/ui_act(action, list/params, datum/tgui/ui, datum/ui_state/state)
+	. = ..()
+
+	var/datum/preferences/prefs = ui.user.client.prefs
+
+	switch(action)
+		if("add")
+			var/datum/gear/gear = GLOB.gear_datums_by_name[params["name"]]
+			if(!istype(gear))
+				return
+
+			var/total_cost = 0
+			for(var/gear_name in prefs.gear)
+				total_cost += GLOB.gear_datums_by_name[gear_name].cost
+
+			total_cost += gear.cost
+			if(total_cost > MAX_GEAR_COST)
+				return
+
+			prefs.gear += gear.display_name
+
+		if("remove")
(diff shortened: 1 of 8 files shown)


code/modules/client/preferences.dm
Drulikar, line 1: Where we have `// Refresh pickers` in this file, we should also do this new picker. Likely at this point you should just make a proc so shove them all into that proc.

code/modules/client/loadout_picker.dm
```diff
@@ -0,0 +1,87 @@
...
+		.["categories"] += list(
+			list("name" = category, "items" = items)
+		)
+
+	.["max_points"] = MAX_GEAR_COST
+
+/datum/loadout_picker/ui_data(mob/user)
+	. = ..()
+
+	var/datum/preferences/prefs = user.client.prefs
```
Drulikar, line 25:
```suggestion
	var/datum/preferences/prefs = user.client?.prefs
```
Just in case I guess?

tgui/packages/tgui/interfaces/LoadoutPicker.tsx
```diff
@@ -0,0 +1,125 @@
...
+}) => {
+  const { item, loadout } = props;
+
+  const { icon, icon_state, name, cost } = item;
+
+  const { data, act } = useBackend<LoadoutPickerData>();
+
+  const { points, max_points } = data;
+
+  const atLimit = points + cost >= max_points;
```
Drulikar, line 101:
```suggestion
  const atLimit = points + cost > max_points;
```
else you cannot buy therapy plush ever

Drulikar (changes requested):

code/modules/client/loadout_picker.dm
```diff
@@ -0,0 +1,87 @@
...
+		.["loadout"] += list(
+			list("name" = gear.display_name, "cost" = gear.cost, "icon" = gear.path::icon, "icon_state" = gear.path::icon_state)
+		)
+
+	.["points"] = points
+
+/datum/loadout_picker/ui_act(action, list/params, datum/tgui/ui, datum/ui_state/state)
+	. = ..()
+
+	var/datum/preferences/prefs = ui.user.client.prefs
```
Drulikar, line 46:
```suggestion
	var/datum/preferences/prefs = ui.user.client?.prefs
```

Also any particular reason why you null tested prefs in ui_data but not here?

harryob: the whole tgui chain should prevent a null client from slipping through, but...

code/modules/client/preferences.dm
```diff
@@ -1031,39 +1029,7 @@ GLOBAL_LIST_INIT(bgstate_options, list(
...
-						to_chat(user, SPAN_WARNING("Adding \the '[choice]' will exceed the maximum loadout cost of [MAX_GEAR_COST] points."))
-
-				if("remove")
-					var/i_remove = text2num(href_list["gear"])
-					if(i_remove < 1 || i_remove > length(gear)) return
-					gear.Cut(i_remove, i_remove + 1)
-
-				if("clear")
-					gear.Cut()
+			loadout_picker.tgui_interact(user)
```
Drulikar, line 1032:
```suggestion
			loadout_picker.tgui_interact(user)
			return
```
To address issue of slot UI always focusing when opening the panel.

tgui/packages/tgui/interfaces/LoadoutPicker.tsx
```diff
@@ -0,0 +1,125 @@
...
+        <Button
+          tooltip={name}
+          onClick={() => act(loadout ? 'remove' : 'add', { name: name })}
+          disabled={!loadout && atLimit}
+        >
+          <DmIcon
+            icon={icon}
+            icon_state={icon_state}
+            height="64px"
+            width="64px"
```
Drulikar, line 115:
```suggestion
            width="64px"
            mb={1}
            mt={1}
```
To address issue of micromovement when DmIccons loads.

tgui/packages/tgui/interfaces/LoadoutPicker.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const atLimit = points + cost > max_points;
+
+  return (
+    <Stack>
+      <Stack.Item>
+        <Button
+          tooltip={name}
+          onClick={() => act(loadout ? 'remove' : 'add', { name: name })}
+          disabled={!loadout && atLimit}
```
Drulikar, line 109:
```suggestion
          disabled={!loadout && atLimit}
          width="78px"
          height="74px"
          mb="3px"
          mt="3px"
```
To address issue of tiny boxes appearing for 1 frame while DmIcon loads.

Drulikar (changes requested):

Drulikar (approved):

PR #7526, opened 2024-11-07
implements a picker for character traits

harryob:
we're almost there! only maybe 1-2 more individual prs and then i can make ANOTHER pr JUST to tgui the human screen. kino! kino?

https://github.com/user-attachments/assets/014faf38-1c9a-4bbe-8cd4-d5dea28d1282

after this video i made it so the buttons appeared disappeared when at the cap but i couldn't be bothered to rerecord

:cl:
ui: added a new tgui to select character traits. it's quite similar to the old one
/:cl:


cuberound:
OMG IT IS TGUI LOBBY TIME

irakligabunia:
kino called they said movie

TheManWithNoHands:
kino indeed.
Finally character set up wont look like its from 1997.

code/modules/client/preferences.dm
Drulikar, line 1: Where we have `// Refresh pickers` in this file, we should also do this new picker. Likely at this point you should just make a proc so shove them all into that proc.

code/modules/client/traits_picker.dm
```diff
@@ -0,0 +1,96 @@
...
+			)
+
+		.["categories"] += list(
+			list("name" = group.trait_group_name, "traits" = traits)
+		)
+
+/datum/traits_picker/ui_data(mob/user)
+	. = ..()
+
+	var/datum/preferences/prefs = user.client.prefs
```
Drulikar, line 27:
```suggestion
	var/datum/preferences/prefs = user.client?.prefs
```
Also do you want a prefs null test here?

code/modules/client/traits_picker.dm
```diff
@@ -0,0 +1,96 @@
...
+	for(var/trait_type as anything in prefs.traits)
+		var/datum/character_trait/trait = GLOB.character_traits[trait_type]
+		.["traits"] += list(
+			list("name" = trait.trait_name, "desc" = trait.trait_desc, "cost" = trait.cost, "type" = trait_type)
+		)
+
+/datum/traits_picker/ui_act(action, list/params, datum/tgui/ui, datum/ui_state/state)
+	. = ..()
+
+	var/datum/preferences/prefs = ui.user.client.prefs
```
Drulikar, line 43:
```suggestion
	var/datum/preferences/prefs = ui.user.client?.prefs
```

Also do you want a prefs null test here?

code/modules/client/preferences.dm
```diff
@@ -2257,52 +2223,6 @@ GLOBAL_LIST_INIT(bgstate_options, list(
 	alert("The key sequence is [key_buf].")
 	return key_buf
 
-/datum/preferences/proc/open_character_traits(mob/user, character_trait_group)
-	if(!read_traits)
-		read_traits = TRUE
-		for(var/trait in traits)
-			var/datum/character_trait/character_trait = GLOB.character_traits[trait]
-			trait_points -= character_trait.cost
```
Drulikar, line 2265: This is likely the logic not retained causing the issue w/ more trait points each launch.

Drulikar (changes requested):
- Traits that are incompatible with other traits ought to be conveyed in some way. E.g. all these buttons are enabled, but nothing happens if you click them since you already have a prosthetic: 
image
- Every launch of the game seems you're given 2 points so you can keep adding more (Does not seem to be an issue on master).
image
And it will go negative :p but this likely won't be possible once above is fixed: 
image

deathrobotpunch:
Forrest, can you make it so character impairments don’t cost points?

harryob:
> Forrest, can you make it so character impairments don’t cost points?

who are you talking to?

deathrobotpunch:
> who are you talking to?

¯\_(ツ)_/¯ I imagine things

cm13-github:

cm13-github:

Drulikar (approved):

PR #7536, opened 2024-11-08
Adds crosshairs to all ranged weapons

VileBeggar:
# About the pull request

all weapons now have custom crosshairs depending on what category they are. rifles, dmrs/snipers, shotguns, pistols/smgs, flamer, explosive launchers and the smartgun have unique crosshairs. if i missed a weapon they'll use a default cross marker like rifles do. the XM88 HAS A UNIQUE crosshair (it existed before but it was scope-only) that is now implemented to work without a scope. these crosshairs will appear only when the weapon is wielded

image

preds have a special cursor for the plasma caster: 
image

they're optional, the preference for custom crosshairs already exists but it's on by default.

most servers already have a crosshair in some form or another. while these might not be the best looking nor the best implementation (if this gets merged i'll probably try to get dynamic crosshairs working), i feel like they're an upgrade over shooting with the default mouse cursor. 

i dunno how to describe it but it just feels better. it's definitely a placebo but even using the dinky windows crosshair made me feel more precise

<details>
video quality is pr bad due to the size limit, but it looks better in-game

https://github.com/user-attachments/assets/916b9433-2901-40cf-a57a-ff5174ba51c1

</details>

# Changelog
:cl:
add: Weapons now have custom crosshairs. They're optional and can be turned off in the preferences menu.
/:cl:


Detective-Google:
another banger from the soultainer

VileBeggar:
i haven't gone through the pred weapons but i know exactly what you're talking about. good idea

Osmos123:
My crosshair disappears after I vault over an object (wooden barricade for me)
Wield gun and hit middle mourse or drag over, the crosshair disappears and turns into regular mouse cursor.

nauticall (approved):
in the future we should probably make the color of the crosshairs a preference, since blue might be hard to see for some people; but this looks good otherwise for now

PR #7573, opened 2024-11-12
Nerfs exploitable room on Chances Claim.

HeresKozmos:
# About the pull request

Nerfs pointlessly fortified room/micro-area on Chances Claim to make it less exploitable.

Stops ridiculous cheese like this video below; there's no reason why a room with seemingly no purpose has an un-acidable, un-explodable and un-pryable double shutter, hull walls and a ladder that people can hide behind, barricade and camp inside of. 

They literally just sat there and camped while calling down a crate and camping near the sensor tower - a player shouldn't be forced to babysit this one area should a marine/group just set up camp inside of this incredibly safe kill-box that is impossible for Xenomorph players to break into.

There is nothing actually inside the room to justify such tough walls and shutters as the total contents of the room amount to a bedroll, some bloody gauze, some wrappers, some Souto, a single (normal) blowtorch and exactly two small crates. There is zero reason why this warrants hull walls and an invincible shutter which can be opened and closed from inside.

https://github.com/user-attachments/assets/ebbc75d2-8649-4204-98c5-92a92d42f3d0

image

# Changelog

:cl:
maptweak: Lowered overall toughness (removed immunities*) of two shutters on Chance's Claim.
maptweak: Adjusts adjacent walls to be normal, non-hull walls which can actually be broken.
/:cl:

### DIFF
--- a/maps/map_files/LV522_Chances_Claim/LV522_Chances_Claim.dmm
+++ b/maps/map_files/LV522_Chances_Claim/LV522_Chances_Claim.dmm
@@ -45325,18 +45325,14 @@
 "xLr" = (
 /obj/structure/machinery/door/poddoor/almayer{
 	dir = 2;
-	id = "map_corpo";
-	explo_proof = 1;
+	id = "map_corpo2";
 	name = "Emergency Blast Door";
-	unacidable = 1;
 	use_power = 0
 	},
 /obj/structure/machinery/door/poddoor/almayer{
 	dir = 2;
-	id = "map_corpo2";
-	explo_proof = 1;
+	id = "map_corpo";
 	name = "Emergency Blast Door";
-	unacidable = 1;
 	use_power = 0
 	},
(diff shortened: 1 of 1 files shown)


deathrobotpunch:
The souto vault needs to be well defended pal, you can’t have commies stealing it and making their own souto do you?

BOBAMAx:
This part of the map is supposed to be out of bounds as flavor for ghosts, not an actual playable part of the map. People being there is an unintended bug, you shouldn't be making the area more accessible.

spartanbobby:
This room is considered out of bounds if people are getting in here they are using a bug to

Nanu308 (changes requested):
Change it such that people are unable to enter, not allow it to be more accessible. As the map creator Spartanbobby has clarified.

harryob:
pr appears to be stale with requested changes - please address the changes and ask us to reopen, or make a new pr

PR #7581, opened 2024-11-13
Standard Shrapnell Buff + Shrapnel rocket now shoots directional shrapnell

Diegoflores31:
# About the pull request

Increases shrapnell damage accuracy and damage . 
Rockets will now shoot their shrapnell (if they have any) in a 90 angle facing the objective.

Currently Shrapnell is near useless in combat against both xenos and humans . this PR aims to make shrapnell more usefull by having greater accuracy. 

The objective of this is having shrapnel have a somewhat consistent output on dmg . instead of being fully RNG . specially on xenomorph . currently ingame an entire 200 shrap mortar can land next to you and you only get about  100 dmg as most of them will miss.

rocket now shoots its shrapnell directionally so the operator doesnt kill itself and others with it .

<details>

</details>

# Changelog
:cl:
balance: Shrapnell accuracy and damage increased
balance: Shrapnell now does 60% bonus damage to xenomorphs
balance: Shrapnell slows xenomorphs on hit 
balance: Custom rockets will now shoot their shrapnell on a 90 degree angle
/:cl:


GoldenDarkness55:
Can you make something like the blast dampener except with the effect of angling shrapnel instead? Would be better to keep it optional (wculd it also work on grenades then?)

As for shrap you'l shouldnt be removing embed chance and lowering pen. Accurate_range_min has to be kept at 1 to prevent single mobs eating insane shrap amounts.
Likely better to make a custom shrap type for OT with higher accuracy and damage but capped to 6 range with heavy falloff around 4 so you take 0-10 DMG on max tile to prevent ff, since as long as the range ff potential remains they won't work for HvX. Would have to make a mortar subtype with proper range though - frag shells already need buffs anyway

Diegoflores31:
> Can you make something like the blast dampener except with the effect of angling shrapnel instead? Would be better to keep it optional (wculd it also work on grenades then?)

on grenades it kinda makes no sense unless we make a new AGM for OT or smth.

the values you see i "deleted" are still there its just that they come from inheritance

min_accurate_range is necesary . its normal to eat big amounts of dmg if you are point blank of an hefa . remember the shrapnel amount is divided by 4 per tile traveled . on xenos is really hard to die from even with new values .

i want it on main shrap because ngl all shrap is weak as hell currently . (hail hefa)

Diegoflores31:
UPDATE : 
Shrapnell damage now slows(normal slow)  xenos on hit for 4 seconds ( it doesnt stack ) ((AGM superslow will be removed))
Shrapnell does 1.6 of its normal dmg to xenomorphs ( about 45 on paper , tested ingame is about 39-40 with no armor )

This changes have been tested ingame . HEFA is now about the same as HE but winout the stun which is compensated by the slow

harryob:
> UPDATE : Shrapnell damage now slows(normal slow) xenos on hit for 4 seconds ( it doesnt stack ) ((AGM superslow will be removed)) Shrapnell does 1.6 of its normal dmg to xenomorphs ( about 45 on paper , tested ingame is about 39-40 with no armor )
> 
> This changes have been tested ingame . HEFA is now about the same as HE (for targets with no armor, HE does more dmg to armored targets ) but winout the stun which is compensated by the shrapnell slow

changelog?

kiVts:
No idea how it will perform.

kiVts (approved):

PR #7596, opened 2024-11-15
Notifies when intel points are maxed for non-xeno non-yautja corpses.

zzzmike:
# About the pull request

Notifies when intel points are maxed for non-xeno non-yautja corpses.

If we're gonna limit intel points for corpses to 50 bodies, then there should at least be a notice sent so that people aren't trying to recover bodies for points when it's not possible.

<details>

</details>

# Changelog
:cl:
qol: USCM notice sent when intel points for non-xeno non-yautja corpses are maxed out
/:cl:

### DIFF
--- a/code/modules/objectives/mob_objectives.dm
+++ b/code/modules/objectives/mob_objectives.dm
@@ -122,7 +122,7 @@
 		return OBJECTIVE_LOW_VALUE
 
 	else if(ishumansynth_strict(target) && length(scored_humansynth_corpses) == 49)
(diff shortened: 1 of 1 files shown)


Wintermote:
It would be nice to have this displayed in the intel screen too, something like (Max reached) next to the amount of intel displayed for gathering bodies.

zzzmike:
> It would be nice to have this displayed in the intel screen too, something like (Max reached) next to the amount of intel displayed for gathering bodies.
> 
> A single announcement when the limit is reached means latejoiners or people who miss the message will never actually know - at least if it shows up in the intel window you can check and see it whenever.

I never interact with the intel screen so I'm not sure offhand how to do this - I'm happy to add it though in any case if someone does a suggested change or a PR to my branch or something.

harryob (approved):

PR #7597, opened 2024-11-15
Throwing&Flaming mortar shells rebalance.

Sunless2:
# About the pull request

This rebalances the mortar shells explosions to be more dangerous. while nerfing throw range to compensate against people using them as budget OT nades on the frontline.

Cooking mortar shells by accident or on purpose is interesting and soul, but should not be a standard strategy for frontliners, this attempts to create such a space.

The 1 tile throwing range means anyone trying to use these shells to destroy walls (at 175 explosion it's not good at that) they are opening themselves up to counterplay, while it would've been much safer to have just launched a grenade.

_**It's worthwhile to note the explosion is still relatively weak.**_ 
Even at 1 tile range it cannot kill a drone or defender that were set on fire, see video.

<details>

Throwing
https://github.com/user-attachments/assets/54463748-f053-494d-8802-2ec483598eaa

Explosion
https://github.com/user-attachments/assets/74226d58-ac9b-4afa-8ca7-e6ccc5b1398e

</details>

# Changelog

:cl:

balance: Increases explosive strength of cooked mortar shells
balance: Reduces throwing range of mortar shells

/:cl:

### DIFF
--- a/code/modules/cm_marines/equipment/mortar/mortar_shells.dm
+++ b/code/modules/cm_marines/equipment/mortar/mortar_shells.dm
@@ -29,6 +29,8 @@
 	name = "\improper 80mm high explosive mortar shell"
 	desc = "An 80mm mortar shell, loaded with a high explosive charge."
 	icon_state = "mortar_ammo_he"
+	throw_speed = SPEED_SLOW
+	throw_range = 1
 
 /obj/item/mortar_shell/he/detonate(turf/T)
 	explosion(T, 0, 3, 5, 7, explosion_cause_data = cause_data)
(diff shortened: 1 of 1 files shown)


Blundir:
Keep in mind, this idea was already tried before in the past, and wasn't well received https://github.com/cmss13-devs/cmss13/pull/6707
We had a nerf to that not that long ago recently https://github.com/cmss13-devs/cmss13/pull/6638, nerfing it even harder while making job of OT and req harder isn't a good idea.

Sunless2:
> Keep in mind, this idea was already tried before in the past, and wasn't well received #6707 We had a nerf to that not that long ago recently #6638, nerfing it even harder while making job of OT and req harder isn't a good idea.

I forgot to mention that I made this in reaction to post 10-11 for this thread:

https://forum.cm-ss13.com/t/viability-of-flamer-and-mortar-shells/11292/10

DorfGuy1:
Someone once mentioned the improvised use of explosives on CM as a great example of creativity in ss13, and I think that to decrease that is not fun/good/worthwhile to do.

Blundir:
Thank you for your pr, but i'll have to close it, after some talks, better alternative for this would've been cooked shells having unpredictable direction of fire/shrapnel, it'd solve the issue, make the thing more interesting and wouldn't hurt any other roles, but this pr is very one sided in that regard, it cannot be approved especially considering that one nerf for it went in already.

PR #7599, opened 2024-11-16
Allow donators with more than 1 kit to retrieve additional kits from the personal gear vendor

SASoperative:
# About the pull request

Adds code to the Personal Gear Vendor to check if a user has additional kits and allow them to retrieve additional kits until they have retrieved all kits.

Allows donators who had multiple kits to still keep the kits they would have normally spawned with in their backpacks, by getting them from the personal gear vendor.
# Changelog
:cl:
add: Personal Gear Vendor allows retrieval of additional kits for multi-kit donators
/:cl:

### DIFF
--- a/code/modules/cm_marines/custom_items.dm
+++ b/code/modules/cm_marines/custom_items.dm
@@ -11,7 +11,7 @@ GLOBAL_LIST_EMPTY(donator_items)
 	unslashable = TRUE
 	///a random item that can be claimed if there is no valid kit
 	var/static/list/random_personal_possessions = list()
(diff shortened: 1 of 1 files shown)


Drulikar (changes requested):
This allows the donator to retrieve multiple of the same kit.

Try out something more like this (`git apply path/to/my/file.patch`):
[patch.diff.txt](https://github.com/user-attachments/files/17783809/patch.diff.txt)

SASoperative:
Applied the changes and tested them, appreciate the assistance with that!

Drulikar (approved):

PR #7428, opened 2024-10-29
sg nerf

Red-byte3D:
# About the pull request
nefs sg so they are a bit less effective at having full screen considering they have IFF and are a constant suppressing fire elemetn

# Explain why it's good for the game

Xenos have too much stupid firerate stuff to deal with already, sg was buffed way too much, having full control over your entire screen adds way too much dps for xenos to survive with the recent marine changes either fire rate needs to be moderated (hpr, m2c) or xenos health need to be buffed, and i dont want to make xenos even spongier

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
balance: sg effective range nerfed to 5
balance: sg falloff increased 
/:cl:

PR #3990, opened 2023-07-25
Xeno Alliance Announcements + Greeno Civil War

ihatethisengine:
# About the pull request

Xenos now get messages when their queen set/break alliance to another faction and when other queen set/break alliance with their hive.

Corrupted Xenos with implanted IFF tag now has a choice to defect from the hive and become Renegade (hive allied to USCM) when Queen decides to break alliance with USCM. Xenos that stay loyal to Queen rip the IFF tags out. You have only 10 seconds to make a decision, so think quick. By default xenos stay loyal to Queen.

Renegade Hive is allied to all human factions, but it mostly affects structures and weeds. Renegade ability to attack someone fully depends on its IFF tag settings.

Please check my messages because I'm not very good at writing.

# Explain why it's good for the game

More announcements are good because it's less confusing. You know when someone set ally to you and you know when someone is betraying you. It makes sense because allied xenos share announcements anyway. And sudden betrayals are always a bit cheesy.

I think the Renegade addition makes research a little more fun and rewarding. Now, if you implant corrupted xeno with an IFF tag, the xeno player can choose to remain loyal to you if/when the Queen decides to betray. Plus corrupted xenos with the IFF tag are no longer forced to betray, so it's also good. Not sure if that makes sense lore-wise, but since corrupted are artificially created and their DNA is decrypted, it makes at least some sense. Plus we kinda have tamed xenos which work really similarly. 

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl: ihatethisengine
add: Added announcements for xenos about forming and breaking alliances.
add: Xenos with IFF tag now have a choice to stay loyal to USCM when Queen decides to betray.
/:cl:

PR #7600, opened 2024-11-16
SG Frontline Mode & Friends - Yay!!!

BasilHerb:
# About the pull request

Before I explain this PR, I'd like to give a massive shoutout to ihatethisengine, who basically wrote all the code for this PR, and may as well be the author. He's a hero for sticking with my dumbass for 7 straight hours in code chat, grinding out this PR in less than a day. Also massive thanks to Drathek, Blundir, HarryOB, kiVts, detectivegoogle, cuberound and Tristan.

This PR introduces a new firemode to the USCM, PMC and RMC smartguns, as well as some other minor changes.

**FRONTLINE MODE:**
The smartgun can now toggle on 'frontline mode', which adds Alt IFF to the gun (so you can't shoot over friendlies), and in exchange buffs your DPS and effective range to almost that of a Mk2 Pulse Rifle.
You can toggle frontline mode OFF whenever you want with a press of a button (hotkeys are supported!!!) to return to conventional IFF, with the same damage output as we have now.

26/11/2024 EDIT: When frontline mode is disabled, the SG will now fire blue tracers, this lets both xenos and marines know if an SG has frontline mode enabled or disabled. Frontline damage nerfed by 2 across the board for better balance, after testmerge observations & feedback.

**AMMO SWITCHING WITH IFF OFF:**
It always made me really sad that you couldn't use AP with IFF off, especially since it meant anyone wearing armour during a mutiny was basically invincible to you. 

The purpose of this PR is to satisfy both sides of the aisle. IFF has historically been a controversial topic, with many advocating for the 'skilled, pointmannification' of SG, and others preferring SG retain it's 'unique identity as a support gunner'. Because both the current status quo (SG nerf) and Alt IFF are both unpopular on their own, I believe that making Alt IFF an optional firemode is the way to go. SGs who wish to be rewarded for 'skill' and 'gamesense' can do so, while those who prefer to 'support from the back' can also do so. It also makes SG way more interesting and unique than it otherwise would be. 

I have extensively tested this PR with RMC, PMC and USCM SGs, and (after a LOT of testing), have encountered no bugs in regular play. 

The SG is currently in a pretty bad spot. Most people I've talked to who actually play SG have said it feels like the worst it's ever been. Buffing the SG's IFF deathstacking ability is out of the question due to the decades long powercreep of IFF, and SG itself has also felt stale. It's stagnated, and there hasn't been a mechanical rework since full auto got added. 

I've talked with many people, both normal players and management staff, and they've mostly told me that the optimal solution would be for the SG to have a seperate firemode for frontlining/pointmanning they can turn on at will. 

As such, I believe that this is an SG rework that the majority of players would like to see happen, and as a long time SG player, I'm happy to see SG finally get a breath of fresh air, with a far more tactical dimension added, rewarding skilled experience with honour and glory! (I will get lungecapped by prime warriors 0.5 seconds after toggling frontline mode)

On the non-soul side, this PR also pragmatically adds Alt-IFF components, as well as a basic framework for future PRs to expand upon. (Imagine frontline mode conferring other buffs and debuffs to make it more engaging! hint hint, cuberound?)

https://github.com/user-attachments/assets/024695fa-9c6d-4231-b721-3a772630dd8d

soul old 64p video

https://github.com/user-attachments/assets/ef00208c-4f05-470f-ad27-1ceaa81638e8

unsoul new HD video

# Changelog

:cl:
add: Added a new SG mode, 'Frontline mode', which uses Alt IFF for the USCM, PMC & RMC SGs.
add: Ability to switch ammo types with IFF off.
add: Added balloon alerts to all SG toggles (the floaty text)
add: AP bullets on all SG variants (M56D, M56T, etc.) should now be yellow, just like on the M56B for consistency.
add: SG now fires blue tracers when not on frontline mode (variants for AP and precision), along with blue muzzle flash, etc. as a visual cue for when frontline mode is off/on.
add: sprites both for when frontline mode is enabled and disabled.
balance: SG on frontline mode adjusted to have moderately weaker than Mk2 DPS, and 7 tiles of effective range.
code: Added Alt IFF component
soundadd: Added Alt IFF sound
refactor: Refactored datum/action to have cooldown capabilities on base type
fix: Fixed listen_signal never getting unregistered for actions
/:cl:


RainbowStalin:
The only problem with this I can see, as a that SGs, are gonna be absolutely useless against zombies, and more prominently, greenos who betray marines. As both of those have IFF, and you have no way to turn off your IFF.

ihatethisengine:
> The only problem with this I can see, is that SGs, are gonna be absolutely useless against zombies, and more prominently, greenos who betray marines. As both of those have IFF, and you have no way to turn off your IFF.

1) per this PR https://github.com/cmss13-devs/cmss13/pull/3990 greenos rip their IFF rag out on the betrayal.
2) You still able to disable IFF and even actually do increased DMG.

BasilHerb:
> The only problem with this I can see, is that SGs, are gonna be absolutely useless against zombies, and more prominently, greenos who betray marines. As both of those have IFF, and you have no way to turn off your IFF.

Yeah, don't worry- IFF mode is still a thing, you can turn IFF on to kill teammates for the fun fact. Frontline mode is a seperate thing. Apologies for the confusion- I've made a new video which shows the IFF button in action, which should be uploaded soon (in better quality)

Wintermote:
Since you're reworking the smartgun entirely it would be a good idea to look at autofire.

Personally I think autofire mode should be removed outright. At the moment it does almost nothing; firing 1 bullet when it detects something hostile, but the time between detection and firing is ~2-3 seconds and it only ever seems to fire 1 bullet. Most of the time xenos wont even trigger it before they're off your screen. You also have to click it back on every time you re-wield the smartgun too. 

Removal at least gets rid of a non-functional utility and clears up UI space. Adjusting it to be useful... not sure what you could really do there.

BasilHerb:
> Since you're reworking the smartgun significantly it would be a good idea to look at autofire.
> 
> Personally I think autofire mode should be removed outright. At the moment it does almost nothing; firing 1 bullet when it detects something hostile, but the time between detection and firing is ~2-3 seconds and it only ever seems to fire 1 bullet. Most of the time xenos wont even trigger it before they're off your screen. You also have to click it back on every time you re-wield the smartgun too.
> 
> Removal at least gets rid of a non-functional utility and clears up UI space. Adjusting it to be useful... not sure what you could really do there.

Hmmm, I think I'll probably just remove it in a separate PR in the future instead. Totally agree though, autofire is literally a noob trap that does absolutely nothing of value. I would be against a rework that makes autofire useful, since well... then you'd just be AFKing and shooting automatically, which is terrible game design imo.

Butlerblock:
Im of the personal opinion this just puts sg even closer to pfc+. nobody is gonna want to turn their frontline mode off, and theyre just gonna become a pfc with night vision and a bigger mag, not to mention their gimmick of longer vision was removed.

BasilHerb:
> Im of the personal opinion this just puts sg even closer to pfc+. nobody is gonna want to turn their frontline mode off, and theyre just gonna become a pfc with night vision and a bigger mag, not to mention their gimmick of longer vision was removed.

Hmmm, I'll have to disagree on this point. SGs who refuse to turn off frontline mode will be actively detrimental to themselves, because there are many, many situations where you can't frontline (e.g. hellchokes, extremely risky pushes), where frontlining isn't worth the risk. In those cases, you'd switch frontline mode off, and play 'normally'. 
The most optimal way to play in this PR is to constantly, actively switch between frontline and non-frontline mode, exactly the same as how good SGs switch between AP and precision depending on which xeno strain they're shooting at.

There's a situation for frontline mode, and there's a situation for not using it. A good, experienced SG will actively and seamlessly change between the two for the best efficiency possible- It rewards gamesense.
As such, only inexperienced or 'lazy' players will leave frontline on perpetually, similar to how some people stay on AP or precision mode and never change, even if there's a fortified defender or a hivelord.

Given that this is a purely optional firemode, I'm not very concerned about whether it puts SG 'closer to PFC'. SGs who lean towards frontlining already frontline. SGs who want to stay back will still stay back. SGs who want to play optimally will play the same, because optimal SG playstyle already involves frontlining every so often.

Osmos123:
I wonder if you can add a slight cooldown after toggling frontline mode, seems very strong to be able to switch freely to maximize dps. Maybe can add a visual component for xenos to know a smartgunner has frontline mode enabled to better understand the higher damage dealt.

BasilHerb:
> I wonder if you can add a slight cooldown after toggling frontline mode, seems very strong to be able to switch freely to maximize dps. Maybe can add a visual component for xenos to know a smartgunner has frontline mode enabled to better understand the higher damage dealt.

Hmmm, yeah I’ve been thinking about this too. 
As an SG who switches ammo modes frequently, I’ve found that I’m able to switch frontline mode fast enough to basically fire with minimal pauses, even when moving behind and in front of friendlies constantly. ihatethisengine also raised similar concerns, so I think you’re right. I’ll worry and work on a cool-down in between frontline mode- I’m not really sure on how long the delay should be though. I’ll ask around on discord to try and get an idea for that.

Regarding visual cues, I also agree. Cuberound was going to have visual cues for his SG rework, and I also thought it would be good to telegraph that an SG was on frontline mode, both for friendlies and xenomorphs.
I’m thinking of using those highlights e.g. Ravager red glow empower, holo-target green glow effect. 

Thank you for the feedback- I’ll get to work on it!

Wintermote:
For frontline mode I'd borrow HPR sound (at least until a better one is found), edit yellow projectile sprite with blue/orange gradient, and maybe increase projectile velocity. I think that'd work very well as indicators while maintaining the game's atmosphere.

I don't think having the smartgunner glow in the dark with an outline in frontline mode is a good idea. It's very un-aesthetic and gamey - like painting rocks yellow to indicate a climbable area.

BasilHerb:
> I don't think having the smartgunner glow in the dark with an outline in frontline mode is a good idea. It's very un-aesthetic and gamey - like painting rocks yellow to indicate a climbable area.

Yeah that got scrapped after other people said the same.

code/modules/projectiles/gun.dm
```diff
@@ -1626,12 +1664,32 @@ not all weapons use normal magazines etc. load_into_chamber() itself is designed
...
+	// Apply effective range and falloffs/buildups
+	projectile_to_fire.damage_falloff = damage_falloff_mult * projectile_to_fire.ammo.damage_falloff
+	projectile_to_fire.damage_buildup = damage_buildup_mult * projectile_to_fire.ammo.damage_buildup
+
+	projectile_to_fire.effective_range_min = effective_range_min + projectile_to_fire.ammo.effective_range_min //Add on ammo-level value, if specified.
+	projectile_to_fire.effective_range_max = effective_range_max + projectile_to_fire.ammo.effective_range_max //Add on ammo-level value, if specified.
+
+	projectile_to_fire.shot_from = src
+
+	return 1
```
harryob, line 1689:
```suggestion
	return TRUE
```

code/modules/projectiles/guns/smartgun.dm
```diff
@@ -30,13 +30,19 @@
...
-	var/datum/ammo/ammo_primary = /datum/ammo/bullet/smartgun //Toggled ammo type
-	var/datum/ammo/ammo_secondary = /datum/ammo/bullet/smartgun/armor_piercing //Toggled ammo type
+	var/datum/ammo/ammo_primary //Toggled ammo type
+	var/datum/ammo/ammo_secondary //Toggled ammo type
+	var/datum/ammo/ammo_primary_def = /datum/ammo/bullet/smartgun
+	var/datum/ammo/ammo_secondary_def = /datum/ammo/bullet/smartgun/armor_piercing
+	var/datum/ammo/ammo_primary_alt = /datum/ammo/bullet/smartgun/alt
+	var/datum/ammo/ammo_secondary_alt = /datum/ammo/bullet/smartgun/armor_piercing/alt
 	var/iff_enabled = TRUE //Begin with the safety on.
+	var/frontline_enabled = FALSE //Begins with 'normal IFF' enabled (shoots through friendly marines, instead of Alt-IFF)
```
harryob, line 45: add a code doc as to what this variable does, which is like
```
/// This controls ... 
var/whatever
```

code/modules/projectiles/guns/smartgun.dm
```diff
@@ -270,10 +289,30 @@
...
+	action_icon_state = "iff_toggle_on"
+	listen_signal = COMSIG_KB_HUMAN_WEAPON_TOGGLE_FRONTLINE_MODE
+	button.name = name
+	button.overlays.Cut()
+	button.overlays += image ('icons/mob/hud/actions.dmi', button, action_icon_state)
+
+/datum/action/item_action/smartgun/toggle_frontline_mode/action_activate()
+	. = ..()
+	var/obj/item/weapon/gun/smartgun/G = holder_item
+	G.toggle_frontline_mode(usr)
```
harryob, line 304:
```suggestion
	G.toggle_frontline_mode(owner)
```
usr is bad

code/modules/projectiles/guns/smartgun.dm
```diff
@@ -313,6 +352,14 @@
 
 //more general procs
 
+/obj/item/weapon/gun/smartgun/proc/toggle_frontline_mode(mob/user)
+	to_chat(user, "[icon2html(src, usr)] You [frontline_enabled? "<B>disable</b>" : "<B>enable</b>"] \the [src]'s frontline mode. You will now [frontline_enabled ? "be able to shoot through friendlies" : "deal increased damage but be unable to shoot through friendlies"].")
```
harryob, line 356:
```suggestion
	to_chat(user, "[icon2html(src, user)] You [frontline_enabled? "<B>disable</b>" : "<B>enable</b>"] [src]'s frontline mode. You will now [frontline_enabled ? "be able to shoot through friendlies" : "deal increased damage but be unable to shoot through friendlies"].")
```
src is bad, \the isn't needed

code/modules/projectiles/guns/smartgun.dm
```diff
@@ -337,12 +384,9 @@
 	toggle_ammo_type(usr)
 
 /obj/item/weapon/gun/smartgun/proc/toggle_ammo_type(mob/user)
-	if(!iff_enabled)
-		to_chat(user, "[icon2html(src, usr)] Can't switch ammunition type when \the [src]'s fire restriction is disabled.")
-		return
 	secondary_toggled = !secondary_toggled
-	to_chat(user, "[icon2html(src, usr)] You changed \the [src]'s ammo preparation procedures. You now fire [secondary_toggled ? "armor shredding rounds" : "highly precise rounds"].")
-	balloon_alert(user, "firing [secondary_toggled ? "armor shredding" : "highly precise"]")
+	to_chat(user, "[icon2html(src, usr)] You changed \the [src]'s ammo preparation procedures. You now fire [secondary_toggled ? "armor piercing rounds" : "highly precise rounds"].")
```
harryob, line 388:
```suggestion
	to_chat(user, "[icon2html(src, user)] You changed [src]'s ammo preparation procedures. You now fire [secondary_toggled ? "armor piercing rounds" : "highly precise rounds"].")
```

harryob (dismissed):
just some nit picks

code/modules/projectiles/gun.dm
BasilHerb: done & dusted!

code/modules/projectiles/guns/smartgun.dm
BasilHerb: bon'd and appetite'd

code/modules/projectiles/guns/smartgun.dm
BasilHerb: ground & girlbossed

code/modules/projectiles/guns/smartgun.dm
BasilHerb: swept & swaddled

code/modules/projectiles/guns/smartgun.dm
BasilHerb: clapped & crippled

code/modules/projectiles/guns/smartgun.dm
Drulikar: Theres a difference between `//` and `///`
See https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/AUTODOC.md

You only did `//` when you needed to do `///`

Red-byte3D:
This does too much damage, especaially ofr something that has nvgs. 34 damage per bullet at a 7 screenwide with nvgs is already busted when a spotter sight + mk2 combo happens, we dont need this much damage being done with the firerate + ammo capacity of an sg, look into lowering it

Red-byte3D:
For some more feedback, i dont think sg is balanceable, thats my take, it either needs to be staying very low but constant damage yet always present with other utiltiy (iff, extra vision, nvgs, no need for different ammo mags for ap / normal)

I think you could try just splitting this into two different kits, making it so the sg choice is more valuable rather then being able to change in the battlefield, this makes sg too powerful at everything.

 Making it so its a kit allows you to also balance it better by for example, if its the alternative iff mode, lets you make the mags 150 rounds (for example) so its not as strong in the terms of its constant fire.

Xenomorphs already have to deal with an abysmally large amount of suppressing fire and damage in 2-5 seconds that this addition just breaks that even more, sgs have all of the other utilities to make up for its lack of damage, by the fact that you're not doing alot yes, but you're always there and youre also good at scaring things away. 


From my own experience i have seen alot of castes get literally instantly melted by the alternative iff, simply because this has screenwide range and the faloff starts at the edge of the screen (if i remember reading the code correctly) and in the current state of things its just going to make being a xenomorph even more annoying.

BasilHerb:
> 34 damage per bullet at a 7 screenwide with nvgs is already busted when a spotter sight + mk2 combo happens

Hmmm, I don't believe this to be *that* accurate, as precision mode lacks any armour penetration whatsoever. The damage output is realistically 23 damage per bullet because you'd be using AP, as precision suffers from about -50% damage against 20 armour (warriors+). I suppose we could just remove the pure damage buff altogether (effective range buff remains). I'd rather wait and hear more xeno opinions before doing anything though.

> lets you make the mags 150 rounds (for example) so its not as strong in the terms of its constant fire.

I think an ammo nerf could be interesting- Forcing people to have to pick their fights instead of being able to mostly unload hundreds of rounds at everything would at least partially solve concerns about the 'constant presence' of SGs, given they might not decide to shoot even if you come on screen. Not really sure what exactly the 'sweet spot' for balance would be, but that's what the TM is for anyways.

> I think you could try just splitting this into two different kits, making it so the sg choice is more valuable rather then being able to change in the battlefield, this makes sg too powerful at everything.

> i would personally just make it work like an actual weapon instead of an iff at all

Hmmm, I get what you're saying, but honestly I feel like splitting it into two kits would effectively just create... well, a MK2 but worse. From what I've observed, juggling frontline and ammo modes is actually quite a challenge for most SG players, it's especially noticeable given SGs now fire tracers if frontline mode is disabled (though given half of them unload on hedgehog shields, I'm not really surprised). It's kind of like how specialists work- In the hands of an experienced player, SG should be a versatile role with decent round impact. In the hands of an inexperienced person however, SG is essentially as effective as it was pre-nerf, or woyer food.

Overall, thanks for the feedback InsaneRed! It's very useful to get opinions from people playing xenomorph, instead of people trying out the TM. 

I don't think the 2-kit idea will be all that fun to play, but I could certainly do more about damage and ammo availability. I think I'll wait and see if more xeno players voice the same concerns before deciding to commit to anything, if that's alright.

Drulikar (changes requested):
As is, this does far too much damage and you need to be careful with what falloff values you use per ammo because its a flat amount reduced per tile not percentage.

Frontline mode should be the harder hitting mode compared to the mode that shoots over people, but the values you should shoot for should more be the pre-nerf values before everyone started futzing with SG.

What I mean by pre-nerfs:

- https://github.com/cmss13-devs/cmss13/pull/7065 (set effective range to 7 and reduced falloff for dirty sg ammo)
- https://github.com/cmss13-devs/cmss13/pull/7428 (reduced range more and made falloff 1 tier worse)
- https://github.com/cmss13-devs/cmss13/pull/7473 (buff - was me correcting the falloff not being considered for AP ammo)

So that's to say, damage values should be what they are now for regular mode, but frontline mode can be more pre-nerf values (I would still recommend not having crazy range though). Additionally, non-frontline mode should probably be the one with better range not frontline mode. Keep both modes relevant and consider how they are wanting you the SG to position themselves. If they're in the back in non-frontline mode, them doing less damage and having less range is pointless to use.

cm13-github:

cm13-github:

code/modules/projectiles/guns/smartgun.dm
```diff
@@ -279,6 +309,26 @@
...
+	listen_signal = COMSIG_KB_HUMAN_WEAPON_TOGGLE_FRONTLINE_MODE
+	button.name = name
+	button.overlays.Cut()
+	button.overlays += image ('icons/mob/hud/actions.dmi', button, action_icon_state)
+
+/datum/action/item_action/smartgun/toggle_frontline_mode/action_activate()
+	. = ..()
+	var/obj/item/weapon/gun/smartgun/G = holder_item
+	G.toggle_frontline_mode(owner)
+	if(G.frontline_enabled)
```
Drulikar, line 325:
```suggestion
	var/obj/item/weapon/gun/smartgun/gun = holder_item
	gun.toggle_frontline_mode(owner)
	if(gun.frontline_enabled)
```

code/modules/projectiles/guns/smartgun.dm
```diff
@@ -762,8 +842,12 @@
 	desc = "The actual firearm in the 2-piece L56A2 Smartgun System. This Variant is used by the Three World Empires Royal Marines Commando units.\nYou may toggle firing restrictions by using a special action.\nAlt-click it to open the feed cover and allow for reloading."
 	current_mag = /obj/item/ammo_magazine/smartgun/holo_targetting
 	ammo = /obj/item/ammo_magazine/smartgun/holo_targetting
-	ammo_primary = /datum/ammo/bullet/smartgun/holo_target //Toggled ammo type
-	ammo_secondary = /datum/ammo/bullet/smartgun/holo_target/ap ///Toggled ammo type
+	ammo_primary //Toggled ammo type
+	ammo_secondary //Toggled ammo type
```
Drulikar, line 846:
```suggestion
```

Drulikar (changes requested):

Drulikar (approved):

PR #7604, opened 2024-11-17
Burrower and Carrier max plasma buff

Osmos123:
# About the pull request

Increases the max plasma of burrower and carrier castes to match drone. 

80% of the gameplay involves resting and waiting for plasma pool to regenerate. Burrower only has 600 and carrier 500. It burns through very fast and is not that very engaging. You should not be punished from evolving into a T2 with limited slots (technically 1 reserved slot) while drone enjoys playing around with 1000 plasma. It is also a pain making hive structures, I believe they cost more than the current max plasma pools of either of these castes. 

I expect as a consequence of this that there will be more resin traps, but I think that is manageable coming from a someone that plays marine mainly.
<details>

</details>

# Changelog
:cl:
balance: Carrier max plasma increased 500>1000, Burrrower max plasma increased 600>1000
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/general_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/general_abilities.dm
@@ -418,9 +418,9 @@
 		button.icon_state = "template_active"
 
 /datum/action/xeno_action/onclick/place_trap
-	name = "Place resin hole (200)"
+	name = "Place resin hole (300)"
 	action_icon_state = "place_trap"
-	plasma_cost = 200
+	plasma_cost = 300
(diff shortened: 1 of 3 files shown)


ihatethisengine:
This is the only thing that balances the amount of traps you can lay. Either find an alternative limiter, or don't just mindlessly buff it without thinking of consequences. 

It's still pretty easy to accomplish a lot with the current numbers, ask the queen for leadership/plasma, have a pocket drone or, well, just do something else while you regain plasma (resting doesn't speed up plasma regen in case you didn't know).

Drulikar:
This is not needed. Thank you. Work with a second xeno to help you fill traps and or try to get the most out of recovery pheros.

PR #7609, opened 2024-11-17
Makes you unable to pixelshift structures (barricades/sentries/others)

TheGamerdk:
# About the pull request

MOST of these cases are covered by the anchored check. But some structures (like barricades) can be unanchored and re-anchored. 
This looks kinda silly, and could trick people into misjudging what the barricade is actually protecting.

Prevents hiding barricades and sentries, makes stuff look less silly.

<details>

</details>

# Changelog
:cl:
add: Barricades have been covered in petroleum jelly. They can no longer be pixelshifted.
/:cl:

### DIFF
--- a/code/datums/keybinding/human.dm
+++ b/code/datums/keybinding/human.dm
@@ -154,6 +154,8 @@
 		return FALSE
 	if(grabbed_atom.anchored)
 		return FALSE
+	if(istype(grabbed_atom, /obj/structure))
+		return FALSE
 	return TRUE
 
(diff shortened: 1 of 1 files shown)


thevandie:
It's really cool to be able to move vendors and such for aesthetic reasons. For example, moving vendors into or out from being inset to walls.
I'd like to humbly suggest you exclude vendors and other shipside structures for that reason. It increases creative expression.

harryob (approved):

PR #7612, opened 2024-11-18
Update donator example config & Refactoring

Drulikar:
# About the pull request

This PR is a follow up to #7306 updating the example config, makes the ckey entry in the config a bit more resilient to stuff like case sensitivity by using the [ckey proc](https://www.byond.com/docs/ref/#/proc/ckey), and refactors the two master lists to GLOBs that automatically initialize. Also makes it possible to reload the custom_items config at runtime via advanced proc call.

More flexible config, and an accurate example.

<details>

image
image

</details>

# Changelog
:cl:
code: Refactored donator lists to GLOBs that automatically initialize and made ckeys in custom_items.txt more flexible
config: Updated example config for custom_items.txt
/:cl:

### DIFF
--- a/code/modules/cm_marines/custom_items.dm
+++ b/code/modules/cm_marines/custom_items.dm
@@ -1,30 +1,9 @@
 GLOBAL_LIST_FILE_LOAD(custom_items, "config/custom_items.txt")
-GLOBAL_LIST_EMPTY(donator_items)
+GLOBAL_LIST_INIT(donator_items, generate_donor_kits(FALSE))
+GLOBAL_LIST_INIT(random_personal_possessions, generate_random_possessions(FALSE))
 
-/obj/structure/machinery/personal_gear_vendor
-	name = "personal gear vendor"
-	desc = "A console that allows the user to retrieve their personal possessions from the ASRS."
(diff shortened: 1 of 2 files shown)


code/modules/cm_marines/custom_items.dm
```diff
@@ -53,20 +32,40 @@ GLOBAL_LIST_EMPTY(donator_items)
...
+
+	if(assign_to_glob)
+		GLOB.donator_items = .
 
-/obj/structure/machinery/personal_gear_vendor/proc/generate_random_possessions(regenerate_kits = FALSE)
-	if(regenerate_kits)
-		random_personal_possessions = list()
+	return .
+
+/proc/generate_random_possessions(assign_to_glob = TRUE)
```
harryob, line 46: what's the point of the argument, here? it seems to only be called with FALSE, anyway

Drulikar: Was me attempting to maintain a semblance of beagle's regenerate kits argument. Pointless for `generate_random_possessions` probably; but *could* have use for `generate_donor_kits` if there was functionality to re-read `"config/custom_items.txt"` but currently there isn't that functionality.

harryob (approved):

PR #7633, opened 2024-11-20
Add de-evolve cooldown per caste & De-evolve fixes

Drulikar:
# About the pull request

This PR adds a 5 minute cooldown to go to the same caste again after a de-evolution.

It also fixes various oversights between the self de-evolve vs the queen de-evolve (they really ought to be unified, but there's also various subtle differences between them). Also fixes the Mature Larva name never getting applied.

Normally I would not expect a player needing to de-evolve and re-evolve to the same caste, but there are edge cases where this can be used to reset conditions so it should be discouraged.

<details>

https://youtu.be/ScnsdwFa4yo

After fixes/refactoring:
https://youtu.be/LxlmKzN9QkU

</details>

# Changelog
:cl: Drathek
add: Added 5 minute cooldown when de-evolving to evolve into the same caste
fix: Fixed queen de-evolve deleting her own organ rather than target xeno's
fix: Fixed queen de-evolve not transferring built_structures list
fix: Fixed regular de-evolve not stopping stat tracking on old xeno
fix: Fixed de-evolve verb getting restored if it was taken away (Wouldn't without admin intervention currently)
fix: Fixed mature larva name never getting applied when full evo threshold
del: De-evolving to larva is now never bloodied (bloody used to only depend on evolution points to determine bloody status)
/:cl:


ihatethisengine:
> Normally I would not expect a player needing to de-evolve and re-evolve to the same caste

Sometimes people do this to reset their strain (if their strain became useless in the circumstances, or if they took freed xeno), although we probably should add another mechanic for this.

Drulikar:
> > Normally I would not expect a player needing to de-evolve and re-evolve to the same caste
> 
> Sometimes people do this to reset their strain (if their strain became useless in the circumstances, or if they took freed xeno), although we probably should add another mechanic for this.

I still want it discouraged.

squidqueen:
Would it be possible to use a full evo meter to reset your strain? That'd be a nice change that doesn't hamper drones as much when they devo to unstrain from non builders after other builders die

harryob (approved):

PR #7643, opened 2024-11-21
reinforces the pipes in evac, again

realforest2001:
# About the pull request

I did this quite a while ago but it got overwritten in a merge conflict so here's round two. This makes all pipes in the main evac area (from the right of the cameras in picture below) not explode during hijack. It also applies the same to the maintenance access corridors for the lifeboats.

Everyone takes all the pipes out of these areas anyway, and I've made this change previously it just got overwritten. It does make sense that the evacuation area would have some degree of protection from disastrous overpressure.

I could apply the same logic to CIC and reactor room, but I'm choosing not to.

image
image

<details>

</details>

# Changelog
:cl:
maptweak: Lifeboat air pipes are now reinforced (again) and no longer explode.
/:cl:


Nanu308 (approved):
ja

PR #7650, opened 2024-11-22
Yet Another Attempt To Fix Vamp

ihatethisengine:
# About the pull request
Tail jab loses extra damage from sprite click and now needs to be charged like tail stab, tail jab applies and respects slash delay like tail stab. In exchange tail jab regains the short stun it had.
Vamp can no longer heal while on fire (balance approved: https://github.com/cmss13-devs/cmss13/pull/7135).
~~Vamp loses 5 armor (20 > 15).~~
Flurry CD increased by 1 sec (3 > 4).
Rush distance is increased by 1 (4 > 5), but now the ability can only target mobs and is more consistent at hitting them.

Healer drone's sacrifice now extinguishes the target (balance approved: https://github.com/cmss13-devs/cmss13/pull/7135).

To understand the reasoning behind this PR you must read all 300 messages in this thread: https://forum.cm-ss13.com/t/please-nerf-vampire-lurkers-or-do-something-holy-shit/8658

But tl;dr:
**TOO MUCH DPS**
The main issue with vamp is that this caste has 3 attacking abilities, all of them deal damage. Thus you can use them all in an instance and land a slash on top, as there is no delay between using abilities. This leads to huge burst damage output in an instance: tail jab + rush + flurry + slash deal about 70 damage to armored target. Again, in an instance. To fix this we remove extra damage from direct tail stab (was a dumb idea in the first place, and powercreep), and turn tail jab into something more like tail stab by adding charging and making it apply and respect slash delay. Thus you won't be able to chain it with slash as easily, which lowers the burst damage against armored target to about 50. To land that extra slash you need to overstay your welcome which is a fair risk if we are to apply other changes.

Now to the stun. It was removed in an attempt to solve the issue with good vamp players killing the entire frontline. Truth is, it wasn't the reason, the burst damage was. If you have good ping and/or good at chaining abilities, you don't need to rely on the tiny stun to be able to offload your combo. With current armor values you don't even need to care as much about PB. So I truly believe that stun removal only made inexperienced players lives worse, who try to learn the caste, while barely affecting good players. Now in combination with all the other changes, short stun should be an important part of the kit, as you are now more vulnerable to damage. It shouldn't turn the caste into another base lurker, because the micro stun mostly only gives you some leeway against PB, temporary lowers incoming damage (by making your enemy drop the weapon) and makes it easier for new players to finish their combo.

**TOO MUCH SURVIVABILITY**
Rush being a "get out of jail" card on the fastest caste in the game seems not right. You really have enough speed to run away, you don't need to rely on dash to play the game (You maybe needed it against surprise PBs, but now you can control incoming PBs with the stun). This is why I make it target only mobs. I include friendly mobs, so rush keeps some escaping utility, but now it's more limited. The distance is now increased, but the effective leap distance is pretty much stays the same as 5th tile is usually the target. It's also a bit easier to land now, as you basically cannot missclick, which is another nod to inexperienced players.

~~20 armor is honestly a lot. It made vamp way tankier than base lurker. Even with 15 armor it's still tankier, but now it's at least close. I think having both high speed and high effective health is not something that the caste is supposed to have. Now getting shot and getting PBed is actually scary, but with stun coming back you can control PBs and incoming damage to a degree.~~

Flurry CD is just comically short, so increasing it is the only option. It goes both in DPS and survivability sections, as flurry is yet another attacking ability you can use to chain with slash and increase your DPS, but also gives you 30 hp for every target you hit. 1 second is not much, but I believe it is needed.

Everything regarding fire is carried over from ghostsheet PR.

<details>

</details>

# Changelog
:cl: ihatethisengine
balance: Tail jab loses extra damage from sprite click and now needs to be charged like tail stab, tail jab applies and respects slash delay like tail stab.
balance: Tail jab regains short (0.5 seconds) stun.
balance: Vamp can no longer heal while on fire.
balance: Flurry CD increased by 1 sec (3 > 4).
balance: Rush distance is increased by 1 (4 > 5), but now the ability can only target mobs and is more consistent at hitting them.
balance: Healer drone's sacrifice now extinguishes the target.
/:cl:


Red-byte3D:
the entire reason i removed the stun was that because you do not need to aim it at all, i dont think it needs ot come back

Drulikar (changes requested):
I think this is going too far with the nerfs. I will not approve this without the tailstab doing some damage.

Revert all changes to tail jab and instead just remove the extra damage on sprite click.

DorfGuy1:
For the love of god, dont' give them stuns. They don't need it. Having to charge your tailstrike and a little damage is worth way less than a stun.

ihatethisengine:
Rush putting you on top of your target just like pounce felt clunky (and made combo execution annoying), so I made sure you land next to the resting target instead of on top of them.

code/modules/mob/living/carbon/xenomorph/abilities/lurker/lurker_powers.dm
```diff
@@ -144,9 +144,66 @@
...
+	var/mob/living/carbon/xenomorph/xeno = owner
+	var/datum/behavior_delegate/lurker_vampire/vamp = xeno.behavior_delegate
+	if(!vamp.rush_target_ref)
+		return
+
+	var/mob/living/target = vamp.rush_target_ref.resolve()
+	if(!target)
+		return
+
+	if(target in new_loc) //don't wanna jump on top of them
```
ihatethisengine, line 199:
```suggestion
	if(target.Adjacent(xeno)) //don't wanna jump on top of them, strike when within the reach
```

ihatethisengine (commented):
a bit more consistent.

Mister-moon1:
After testing the changes, i cannot say the rush change feels good, i think it would be better staying as not needing to target a mob as the marine only sees guaranteed hits while the vamp player experiences every missed ability usage, this just makes it more frustrating in the eyes of both sides rather than achieving it's goal of reducing mobility with rush as an ability.

I would recommend maybe a cooldown increase instead and reverting it to 4 tiles or maybe a short windup before the rush happens

BlitzArde:
I've played a lot of vamp lurker thus far and have no trouble using the rush to hit Marines consistently, however, I do find it a bit disappointing that I can no longer use rush to break windows quicker or quickly vault over window frames/barricades. Please do not add a windup to the rush I beg...

cm13-github:

code/modules/mob/living/carbon/xenomorph/abilities/general_abilities.dm
```diff
@@ -223,6 +223,7 @@
 
 	var/list/pounce_callbacks = null // Specific callbacks to invoke when a pounce lands on an atom of a specific type
 										// (note that if a collided atom does not match any of the key types, defaults to the appropriate X_launch_collision proc)
+	var/list/pounce_end_callbacks = null // Callbacks to call when the pounce (throw) ends
```
hry-gh, line 226:
```suggestion
	/// Callbacks to call when the pounce (throw) ends
	var/list/pounce_end_callbacks = null
```

hry-gh, line 226: when is this ever used

hry-gh (changes requested):

code/modules/mob/living/carbon/xenomorph/abilities/general_abilities.dm
ihatethisengine: https://github.com/cmss13-devs/cmss13/pull/7650/files#diff-731c63cbde0bacf61511200742fbd91dbd38d0d3b4336b54f4cb6542e3869a42R444 here

ihatethisengine:
I'll resolve the conflicts in a few days

Drulikar:
Let us know when you come back with the milk.

PR #7670, opened 2024-11-24
Fax Responders - late spawn faxes

realforest2001:
# About the pull request

Thanks to Harry, spawns any faxes sent before the machines spawned in when the platform loads up.

Makes responding to faxes that were sent before any responders woke up easier.

Tested and confirmed loading correctly.

</details>

# Changelog
:cl:
add: Fax Responder platform now includes faxes sent before it loaded.
/:cl:

### DIFF
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -237,6 +237,8 @@ SUBSYSTEM_DEF(ticker)
 
 	// We need stats to track roundstart role distribution.
 	mode.setup_round_stats()
+	// Loads up the fax responder platform.
+	mode.load_fax_base()
 
 	//Configure mode and assign player to special mode stuff
 	if (!(mode.flags_round_type & MODE_NO_SPAWN))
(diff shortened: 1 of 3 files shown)


harryob (changes requested):
this kind of runs contrary to lazy templates (as they are intended to load lazily when required), and i don't like how this hooks into SSticker

instead, lets make the fax machines just print out what was sent to them when they load (just apply this patch)

code/controllers/subsystem/ticker.dm
```diff
@@ -237,6 +237,8 @@ SUBSYSTEM_DEF(ticker)
 
 	// We need stats to track roundstart role distribution.
 	mode.setup_round_stats()
+	// Loads up the fax responder platform.
+	mode.load_fax_base()
```
harryob, line 241:
```suggestion
```

harryob (changes requested):
forgot to actually remove this in the patch but its redundant now

harryob (approved):

PR #7681, opened 2024-11-26
Langchat is now actually readable through cameras

ihatethisengine:
# About the pull request
Makes tgui able to read langchat styles.

Readable!

before:

after:

# Changelog
:cl: ihatethisengine
add: makes langchat readable through tgui menus.
/:cl:

### DIFF
--- a/code/controllers/subsystem/tgui.dm
+++ b/code/controllers/subsystem/tgui.dm
@@ -32,6 +32,7 @@ SUBSYSTEM_DEF(tgui)
 	var/polyfill = file2text('tgui/public/tgui-polyfill.min.js')
 	polyfill = "<script>\n[polyfill]\n</script>"
 	basehtml = replacetextEx(basehtml, "<!-- tgui:inline-polyfill -->", polyfill)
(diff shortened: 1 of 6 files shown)


harryob:
this is because the off screen maps that are used for rendering this don't have the stylesheet property that is applied via the skin.dmf

image

this also resolves the issue, but has its own issue of having another copy of the same stylesheet contained in skin.dmf

we can resolve both issues by putting the stylesheet into a define, using `winset(src, "map", "style=\"[MAP_STYLESHEET]\"")` somewhere in client new and putting the map stylesheet into tgui config

https://github.com/user-attachments/assets/32d1fdb5-7107-4427-8e26-1e3cf9b45cfa

this is a patch of that  [diff.patch](https://github.com/user-attachments/files/17921750/diff.patch)

ihatethisengine:
Thank you harry :pray: :pray: :pray: :pray: :pray: :pray: :pray: :pray: :pray: :pray: :pray:

harryob (approved):

PR #7695, opened 2024-11-28
Synthetic Maintenance Station is now constructable, Nerfs max charge

tallfission:
# About the pull request

Synthetic maintenance stations are now constructable with components (1 Micro manipulator, 1 Micro laser, 1 Scanning module and 2 Capacitors), 2 of the boards for these machines also spawn in circuit board vendors.

Synthetic maintenance stations on some maps are basically inaccessible, and while this would usually be a mapping issue, xenos also tend to rush to all of the spots of maintenance stations on the colony and destroy them either way. It's kind of hit-or-miss. With internal components being feasibly impossible to repair otherwise, this leaves the only option for internal damage repairs to be returning to the ship. which is a lengthy process and kind of annoying. This PR will allow these machines to be constructed in the FOB or elsewhere. 

I think it's a valid change because now in order for these machines to really work, the area it's in has to be powered - I've reduced the original base charge from 15000 to 12500, and testing locally a synthetic at around 100~(?) damage will take almost half the charge of the station in total. Without power, they'll be quickly drained and become useless. Additionally, burrowers or a cheeky runner can still get into the FOB and destroy them while someone isn't looking, and with there being only 2 boards on the Almayer, it could be something else for Bravo to keep watch over. Amount of base power might need tweaking at some point, will do that if necessary. 

All changes tested locally. Station is constructable, all part requirements show up correctly, nothing broke as far as I could tell. 

<details>
</details>

# Changelog

:cl:
balance: Synthetic maintenance stations are now constructable 
balance: Synthetic maintenance stations now have 12500 base charge instead of 15000 base charge. 
/:cl:

### DIFF
--- a/code/game/machinery/rechargestation.dm
+++ b/code/game/machinery/rechargestation.dm
@@ -12,9 +12,9 @@
 	/// the borg inside
 	var/mob/living/occupant = null
 	/// Two charged borgs in a row with default cell
-	var/max_internal_charge = 7500
+	var/max_internal_charge = 12500
 	/// Starts charged, to prevent power surges on round start
-	var/current_internal_charge = 7500
+	var/current_internal_charge = 12500
 	/// Active Cap - When cyborg is inside
 	var/charging_cap_active = 25000
 	/// Passive Cap - Recharging internal capacitor when no cyborg is inside
@@ -33,6 +33,7 @@
 		/obj/item/stock_parts/capacitor,
 		/obj/item/stock_parts/capacitor,
 	)
+
 
 
 /obj/structure/machinery/recharge_station/Initialize(mapload, ...)
--- a/code/game/machinery/rechargestation.dm
+++ b/code/game/machinery/rechargestation.dm
@@ -12,9 +12,9 @@
 	/// the borg inside
 	var/mob/living/occupant = null
 	/// Two charged borgs in a row with default cell
-	var/max_internal_charge = 15000
+	var/max_internal_charge = 7500
 	/// Starts charged, to prevent power surges on round start
-	var/current_internal_charge = 15000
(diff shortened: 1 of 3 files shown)


irakligabunia:
Synthetic maintenance stations on some maps are basically inaccessible, and while this would usually be a mapping issue, xenos also tend to rush to all of the spots of maintenance stations on the colony and destroy them either way. It's kind of hit-or-miss. With internal components being feasibly impossible to repair otherwise, this leaves the only option for internal damage repairs to be returning to the ship. which is a lengthy process and kind of annoying. This PR will allow these machines to be constructed in the FOB or elsewhere.

They're rare for exactly that reason, it does internal organ repair, heals all kinds of damage within 1-2 minutes
It's handholding to the extreme, synths are already unkillable beasts; they don't need more buffs in terms of endurance and or protection

tallfission:
It still repairs slower than the smartpack can repair, and those are basically indestructible unless the synth takes it off _(and_ it's rechargeable from any APC with a little power), and I think hand repairing can also be faster unless you've got like <10 damage on a lot of body parts. Don't quote me on that, though. The major thing is really the organ damage.

You can't move it, you still have to return to the FOB or another safe area where you've built it. All I really see this doing to make them stronger is making it less annoying and a bit quicker for the synthetic, who can full repair anyways if they take the 10-15~ minutes to get to the ship and then get back. Now it's only the few minutes travel time to the FOB, plus the few minutes it takes to fully repair. 

And you still need to have colony power operational, which isn't always a guarantee, which is _also_ another thing for backlining xenos to target.

kiVts:
This feels like the same area of giving marines an autodoc board roundstart to construct at fob and thus is not exactly acceptable in todays standards.

PR #7728, opened 2024-12-03
Embryo 1s Spawn Protection

Drulikar:
# About the pull request

This PR adds a 1s spawn protection window when larva bursts from an embryo. The mob has godmode for this duration and will signal no ignite. The `RECENTSPAWN` flag helps identify a temporary godmode, and is what also makes all bullets miss temporarily.

This also can be a useful admin tool since the time is variable callable on any living mob.

Unnested bursts are already near suicide for someone to take. This should hopefully make a narrow escape possible. Marines can still trap bursts, but merely shooting the burst immediately will no longer work for 1s.

<details>

https://youtu.be/u1PXcbKuUWc

</details>

# Changelog
:cl: Drathek
add: Chestbursters now have spawn protections for 1s (godmode, evasion, and ignores fire)
/:cl:


harryob (approved):
we should really use traits more often

PR #7729, opened 2024-12-03
Fix bitwise ops (including sentry flamer ammo)

Drulikar:
# About the pull request

This PR corrects a few bitwise ops I found using `&[^\|\(]*\|[^\|]` because of operation order. Basically `flag & ALPHA|BRAVO` is going to always be a truthy result (depends on the actual value for `BRAVO`) since its resolved as `(flag & ALPHA)|BRAVO`. The fixes to matrix editor would have already been caught by checking the admin holder. I don't know about dropship runway. But a potentially concerning one is w/ flame ammo.

Adds a grep lint to find ambiguous bitwise ORs like this using `^(?:[^\/\n]|\/[^\/\n])*(&[ \t]*\w+[ \t]*\|[ \t]*\w+)` (complication came from supporting `//` though block quotes will still trip it which IMO is fine)

These are clear mistakes, but this may make sentry flamer ammo more effective on non-immune xenos.

Fixes: 
image

<details>

fire

Linter: https://github.com/cmss13-devs/cmss13/actions/runs/12153015470/job/33890398213?pr=7729#step:7:36

</details>

# Changelog
:cl: Drathek
fix: Fix code mistakes involving bitwise OR
balance: Sentry flamer ammo now actually applies its initial damage (~30) to non-immune xenos
balance: AMMO_ANTISTRUCT|AMMO_ANTIVEHICLE are now actually required for a vehicle damage multiplier
code: Bitwise OR mistakes like this are now linted against
/:cl:

### DIFF
--- a/code/datums/matrix_editor.dm
+++ b/code/datums/matrix_editor.dm
@@ -7,7 +7,7 @@
 	set name = "Matrix Editor"
 	set category = "Debug"
 
-	if(!usr.client || !usr.client.admin_holder || !(usr.client.admin_holder.rights & R_DEBUG|R_ADMIN))
+	if(!usr.client || !usr.client.admin_holder || !(usr.client.admin_holder.rights & (R_DEBUG|R_ADMIN)))
 		to_chat(usr, SPAN_DANGER("develop man only >:("))
 		return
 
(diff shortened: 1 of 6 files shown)


Git-Nivrak:
Make this a linter as well maybe

harryob (approved):

PR #7739, opened 2024-12-04
Adds a fax machine to requisitions

L0pz:
# About the pull request

Adds a fax machine to requisitions

I've always found it weird how an entire department doesn't have a singular fax machine. Especially since req can use them to ask for unique equipment, which they currently have to do via the public fax machine In the meeting hall.

<details>
<summary>

image

</summary>

</details>

# Changelog
:cl:
add: Added a fax machine to requisitions. It's a girl, by the way.
/:cl:

### DIFF
--- a/maps/map_files/USS_Almayer/USS_Almayer.dmm
+++ b/maps/map_files/USS_Almayer/USS_Almayer.dmm
@@ -48140,11 +48140,6 @@
 	dir = 1;
 	icon_state = "pipe-c"
 	},
-/obj/structure/machinery/faxmachine/uscm{
-	desc = "The beloved fax machine of the requisitions department, for all your equipment-requesting needs";
-	gender = "female";
-	department = "Requisitions"
-	},
 /turf/open/floor/almayer/plating/northeast,
 /area/almayer/squads/req)
 "pXZ" = (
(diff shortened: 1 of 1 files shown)


efzapa (commented):
the `desc` vars need quotes around them, that's what's causing it to fail

maps/map_files/USS_Almayer/USS_Almayer.dmm
```diff
@@ -9046,7 +9046,11 @@
 "blw" = (
 /obj/structure/surface/table/almayer,
 /obj/item/tool/lighter,
-/obj/structure/machinery/faxmachine/uscm,
+/obj/structure/machinery/faxmachine/uscm{
+	desc = The beloved fax machine of the Requisitions department. Be sure to make a convincing argument when you ask high command for a crate of lunge mines.;
```
Nanu308, line 9050: quotation marks required for text strings. Repeat for other errors. Also no need to reference lunge mines.

Nanu308 (changes requested):

DexterDude:
I don't think you removing the SEA fax machine was intentional, you should probably fix that.

L0pz:
> I don't think you removing the SEA fax machine was intentional, you should probably fix that.

Wait shit I thought I had fixed that

Nanu308 (approved):
Le fine

PR #7750, opened 2024-12-06
Makes tackle messaging less confusing

zzzmike:
# About the pull request

Removes the "but they are already down" part of the "tries to tackle but they are already down" messaging for xeno tackles.

The current messaging implies that the xeno tackling can fail because the target is lying down. This just isn't true, and it's confusing for people who aren't familiar with the code. Since being horizontal or not has zero bearing on xeno tackles, the messaging should stay the same in both states.

<details>

</details>

# Changelog
:cl:
del: Xeno tackle messaging no longer cares if the target is horizontal or not (because being horizontal has no bearing on tackles)
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
+++ b/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
@@ -210,8 +210,8 @@
 				knocked_down = TRUE
 			else
 				playsound(loc, 'sound/weapons/alien_claw_swipe.ogg', 25, 1)
-				attacking_xeno.visible_message(SPAN_DANGER("[attacking_xeno] tries to tackle [src]"), \
-				SPAN_DANGER("We try to tackle [src]"), null, 5, CHAT_TYPE_XENO_COMBAT)
+				attacking_xeno.visible_message(SPAN_DANGER("[attacking_xeno] tries to tackle [src]!"), \
+				SPAN_DANGER("We try to tackle [src]!"), null, 5, CHAT_TYPE_XENO_COMBAT)
 				knocked_down = FALSE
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/attack_alien.dm
```diff
@@ -210,12 +210,8 @@
...
 			else
 				playsound(loc, 'sound/weapons/alien_claw_swipe.ogg', 25, 1)
-				if (body_position == LYING_DOWN)
-					attacking_xeno.visible_message(SPAN_DANGER("[attacking_xeno] tries to tackle [src], but they are already down!"), \
-					SPAN_DANGER("We try to tackle [src], but they are already down!"), null, 5, CHAT_TYPE_XENO_COMBAT)
-				else
-					attacking_xeno.visible_message(SPAN_DANGER("[attacking_xeno] tries to tackle [src]"), \
-					SPAN_DANGER("We try to tackle [src]"), null, 5, CHAT_TYPE_XENO_COMBAT)
+				attacking_xeno.visible_message(SPAN_DANGER("[attacking_xeno] tries to tackle [src]"), \
+				SPAN_DANGER("We try to tackle [src]"), null, 5, CHAT_TYPE_XENO_COMBAT)
```
harryob, line 214:
```suggestion
				attacking_xeno.visible_message(SPAN_DANGER("[attacking_xeno] tries to tackle [src]!"), \
				SPAN_DANGER("We try to tackle [src]!"), null, 5, CHAT_TYPE_XENO_COMBAT)
```

harryob (changes requested):

harryob (approved):

PR #7754, opened 2024-12-06
Adds Watchtowers

Git-Nivrak:
# About the pull request
Adds watchtowers

Construction:
zqxqUHb (1)
1. Make 2x2 of dislodged girders
2. Welder the bottom left dislodged girder
3. 60 metal rods
4. Blowtorch, Wrench
5. Screwdriver + 50 metal in offhand
6. Crowbar + 25 plasteel in offhand
7. Wrench + 60 metal rods in offhand
8. Blowtorch + 25 plasteel in offhand

Hitboxes:
Orange - Removed in later commit, ignore
Red - Always blocks movement (two 2x1s)
Green - Allows movement if you are ontop of the watchtower (1x2, this is where you after climbing into the watchtower)
image

Mechanics:
1. Only marines and non-big xenomorphs can climb the watchtower, from the tile infront of the ladder
2. Being ontop of the tower gives you 2 extra tiles of view and makes your shots arc so they only hit what you click, skipping all other mobs
3. Xenos can damage the tower by slashing it
4. Marines can repair the tower by patching it with a welder and metal in offhand
5. The tower does not protect you from most xeno abilities

Besides the fact it's cool I cannot tell if it is actually good for the game without some testing and adjusting

Some videos are from different commits so they might be outdated but they show the general concept

https://github.com/user-attachments/assets/de764cb7-f3b5-437f-8e18-18e07b1299ad

https://github.com/user-attachments/assets/902cccac-3295-432b-87d8-075bb6a04be1

Uploading 2024-12-05 21-38-03.mp4…

# Changelog

:cl: Nivrak, Thwomper
add: Added watchtowers, sprites by thwomper
/:cl:


code/modules/watchtower/watchtower.dm
```diff
@@ -0,0 +1,283 @@
...
+		if(!do_after(user,30, INTERRUPT_ALL|BEHAVIOR_IMMOBILE, BUSY_ICON_BUILD))
+			return
+
+		REMOVE_TRAIT(user, TRAIT_ON_WATCHTOWER, "watchtower")
+		var/turf/actual_turf = locate(x, y-1, z)
+		user.forceMove(actual_turf)
+		user.client.change_view(user.client.view - 2)
+		var/atom/movable/screen/plane_master/roof/roof_plane = user.hud_used.plane_masters["[ROOF_PLANE]"]
+		roof_plane?.invisibility = 0
+		for(var/obj/item/weapon/gun/gun in user)
```
Contrabang, line 234: if someone has a gun in their bag, or throws their gun off the watchtower, this trait wont be removed I think

Git-Nivrak: Updated the bullet trait to automatically delete itself if the user doesn't have the watchtower trait, also made it so any items picked up will get the bullet traits as well

neeshacark:
you can pb and punch people on the ground whilst on towers

putting down stretchers functions weirdly on towers

you cant throw things off towers, either to other towers or just down

troytroy400:
I messed around with mounting placeable machine guns to the towers. It seems to work fine with the M56D but the M2C seems to have some issues with its rounds colliding with the tower. May be something to look into, would be cool if maybe they were mountable to the barrier or something.

cuberound:
define ex_act, common mistake of new structures, default ex_act just destroys the structure on any explosion

Catzonwheelz:
BUG report here.
Seems that when you lay down and get dragged off the tower you will retain the extra vision. you can start to stack up the range yet again, by repeating having some one dragging you off the tower.
here's a screenshot of where i did it few times.

scrnshot1

nuklearcellphoneg:
When manning an m56d, when you let go of the gun voluntary or otherwise, it resets your vision to regular distance so you have to climb the tower again.

nuklearcellphoneg:
Also firing the M56D from the watchtower hits cades, not sure if that's intentional

code/game/objects/structures/girders.dm
```diff
@@ -176,6 +176,33 @@
...
+					var/found_girder = FALSE
+					for(var/obj/structure/girder/girder in current_turf)
+						if(girder.state == STATE_DISPLACED)
+							found_girder = TRUE
+							girders += girder
+					if(!found_girder)
+						return
+
+
+				if(!do_after(user,30, INTERRUPT_ALL|BEHAVIOR_IMMOBILE, BUSY_ICON_BUILD))
```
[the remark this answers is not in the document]
Drulikar: 3 SECONDS

Drulikar (changes requested):

code/game/objects/structures/girders.dm
Drulikar, line 1: image
`/obj/structure/girder/get_examine_text(mob/user)`
should have a welding hint now that you've added that as an option.

I don't think you have construction hints for any other steps either for `/obj/structure/watchtower/get_examine_text(mob/user)`.

code/modules/watchtower/watchtower.dm
Drulikar, line 1: Either weeds need to be destroyed on watchtower construction and weeds prevented from growing under watchtowers (otherwise no construction on weeds but I don't recommend this route since they are moveable), or weed_food needs to have handling to know to not interact with bodies in watchtower turfs. Its wonky to have a body on a watchtower, and the body just randomly disappears (because it then gets layered under the watchtower).

code/modules/watchtower/watchtower.dm
```diff
@@ -0,0 +1,381 @@
...
+
+			if(!do_after(user, 4 SECONDS * user.get_skill_duration_multiplier(SKILL_CONSTRUCTION), INTERRUPT_NO_NEEDHAND|BEHAVIOR_IMMOBILE, BUSY_ICON_FRIENDLY, src))
+				return
+
+			if(rods.use(60))
+				to_chat(user, SPAN_NOTICE("You add connection rods to the watchtower."))
+				stage = WATCHTOWER_STAGE_COLUMNS
+				update_icon()
+			else
+				to_chat(user, SPAN_NOTICE("You failed to construct the connection rods. You need more rods."))
```
Drulikar, line 95: You either need to accumulate rods, or just immediately tell them before a do_after that they need all 60.

code/modules/watchtower/watchtower.dm
```diff
@@ -0,0 +1,381 @@
...
+
+/obj/structure/watchtower/attackby(obj/item/item, mob/user)
+	if(user.action_busy)
+		return
+	
+	if(istool(item) && !skillcheck(user, SKILL_CONSTRUCTION, SKILL_CONSTRUCTION_ENGI))
+		to_chat(user, SPAN_WARNING("You are not trained to configure [src]..."))
+		return TRUE
+
+	switch(stage)
```
Drulikar, line 80: Construction steps *generally* have a message prior to the do_after and then a message after. I would prefer adding messages before e.g. `"You start adding connection rods to [src]"`

There's also no playsound usage in your steps.

code/modules/watchtower/watchtower.dm
Drulikar, line 1: You should be able to replace string instances of `the watchtower` with `[src]` for better future proofing if there are different kinds of watchtowers added.

code/game/objects/structures/girders.dm
```diff
@@ -176,6 +176,33 @@
...
+						return
+
+
+				if(!do_after(user, 3 SECONDS, INTERRUPT_ALL|BEHAVIOR_IMMOBILE, BUSY_ICON_BUILD))
+					return
+
+				new /obj/structure/watchtower(loc)
+				
+				for(var/list/obj/structure/girder as anything in girders)
+					qdel(girder)
```
Drulikar, line 204: There is zero to_chat feedback on this action both before or after the do_after. Nor any sound e.g. `playsound(loc, 'sound/items/Welder2.ogg', 25, 1)`

Drulikar (changes requested):

Blundir (changes requested):
spriter should redo the ladder sprite, elaborated on discord

Blundir (approved):

Git-Nivrak:
Will probably revisit this with multiz

PR #7766, opened 2024-12-08
lobby screen update

harryob:
https://github.com/user-attachments/assets/fa36dad9-d2cd-4742-8dc2-14bc7f1ec59e

:cl: harryob, tyeagg (uscm logo), blundir (upp logo)
add: added a new lobby screen
/:cl:


thevandie:
https://github.com/user-attachments/assets/7411b9b4-9cab-4ea9-ba7c-25f7aa546654

zzzmike:
The old observe confirmation prompt has a little countdown bar before it automatically declines, this one doesn't, what gives? Do we have INFINITE time to decide now? 🤓

harryob:
> The old observe confirmation prompt has a little countdown bar before it automatically declines, this one doesn't, what gives? Do we have INFINITE time to decide now? 🤓

yes. it seems to be a non-issue either way

zzzmike:
> yes. it seems to be a non-issue either way

huge buff for those with decision paralysis 👍

Nanu308 (approved):
mapping approved

harryob:
> i'm not really a fan of the way it obscures the lobby screen art, what with it being opaque & immovable from center.

it's slightly transparent actually

harryob:
> > i'm not really a fan of the way it obscures the lobby screen art, what with it being opaque & immovable from center.
> 
> it's slightly transparent actually

and because im really nice i actually let you toggle it, you can just click on the uscm logo to hide the buttons

cm13-github:

cm13-github:

realforest2001 (approved):

code/modules/asset_cache/assets/lobby.dm
Drulikar, line 1: Not really a concern, but this really does not gracefully handle misconfiguration of lobby assets (in this case I renamed my `lobby_art` config folder to `lobby_artz`: 
image
Mentioning it if you cared to convey bad configuration a different way.

tgui/packages/tgui/interfaces/LobbyMenu.tsx
```diff
@@ -0,0 +1,608 @@
...
+            </Stack.Item>
+          </Stack>
+        </Stack.Item>
+
+        <TimedDivider />
+
+        <LobbyButton
+          index={1}
+          onClick={() => act('tutorial')}
+          disabled={!tutorials_ready}
```
Drulikar, line 318: Consider a tooltip when disabled mentioning wait for game start or something.

Drulikar (changes requested):

code/modules/mob/new_player/login.dm
```diff
@@ -11,27 +18,223 @@
...
+	.["tutorials_ready"] = SSticker?.current_state == GAME_STATE_PLAYING
+	.["round_start"] = !SSticker || !SSticker.mode || SSticker.current_state <= GAME_STATE_PREGAME
+	.["readied"] = ready
+
+	.["upp_enabled"] = GLOB.master_mode == /datum/game_mode/extended/faction_clash/cm_vs_upp::name
+	.["xenomorph_enabled"] = client.prefs && (client.prefs.get_job_priority(JOB_XENOMORPH))
+	.["predator_enabled"] = SSticker.mode?.flags_round_type & MODE_PREDATOR && SSticker.mode.check_predator_late_join(src, FALSE)
+	.["fax_responder_enabled"] = SSticker.mode?.check_fax_responder_late_join(src, FALSE)
+
+	.["lobby_author"] = SSlobby_art.author
```
Watermelon914, line 78: data that doesn't change should probably go into ui_static_data

code/modules/mob/new_player/login.dm
```diff
@@ -11,27 +18,223 @@
...
+			return TRUE
+
+		if("late_join")
+			if(SSticker.current_state != GAME_STATE_PLAYING || !SSticker.mode)
+				to_chat(src, SPAN_WARNING("The round is either not ready, or has already finished..."))
+				return
+
+			if(SSticker.mode.flags_round_type & MODE_NO_LATEJOIN)
+				to_chat(src, SPAN_WARNING("Sorry, you cannot late join during [SSticker.mode.name]. You have to start at the beginning of the round. You may observe or try to join as an alien, if possible."))
+				return
```
Watermelon914, line 132: somewhat inconsistent return values here, should return FALSE if you don't intend to create an update.

same with the entire proc

code/modules/tgui/tgui.dm
```diff
@@ -81,20 +91,21 @@
  *
  * return bool - TRUE if a new pooled window is opened, FALSE in all other situations including if a new pooled window didn't open because one already exists.
  */
-/datum/tgui/proc/open()
+/datum/tgui/proc/open(preinitialized = FALSE)
```
Watermelon914, line 94: Add to the documentation up top what this preinitialized variable is doing

tgui/packages/tgui/interfaces/LobbyMenu.tsx
```diff
@@ -0,0 +1,608 @@
...
+
+  const { lobby_author } = data;
+
+  const onLoadPlayer = useRef<HTMLAudioElement>(null);
+
+  const [modal, setModal] = useState<ReactNode | false>(false);
+
+  const [disableAnimations, setDisableAnimations] = useState(false);
+  const [filterDisabled, setFilterDisabled] = useState(false);
+  const [themeDisabled, setThemeDisabled] = useState<boolean | undefined>();
```
Watermelon914, line 67: just use a default value here instead of using undefined (i.e. false)

tgui/packages/tgui/interfaces/LobbyMenu.tsx
```diff
@@ -0,0 +1,608 @@
...
+  return (
+    <Window
+      theme={themeDisabled ? 'weyland_yutani' : 'crtgreen'}
+      fitted
+      scrollbars={false}
+    >
+      <audio src={resolveAsset('load.mp3')} ref={onLoadPlayer} />
+      <Window.Content
+        className={`LobbyScreen ${
+          filterDisabled ? '' : 'filterEnabled'
```
Watermelon914, line 106: use `classes` function

tgui/packages/tgui/interfaces/LobbyMenu.tsx
```diff
@@ -0,0 +1,608 @@
...
+    predator_enabled,
+    fax_responder_enabled,
+    upp_enabled,
+    tutorials_ready,
+    xenomorph_enabled,
+  } = data;
+
+  const [xenoName] = useState(
+    `${xeno_prefix}-${randomInteger(0, 999)}${xeno_postfix}`,
+  );
```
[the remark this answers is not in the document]
Watermelon914: Use useEffect instead, not useState

tgui/packages/tgui/interfaces/LobbyMenu.tsx
```diff
@@ -0,0 +1,608 @@
...
+          Observe
+        </LobbyButton>
+
+        {round_start ? (
+          <Stack.Item>
+            <Stack>
+              <Stack.Item grow>
+                <LobbyButton
+                  index={5}
+                  disabled={!!readied}
```
Watermelon914, line 372: readied should be a boolean value so this should be redundant
If it has a chance of being undefined, you can declare a default by just doing something like
```tsx
  const {
    // ...
    readied = false,
    // ...
  } = data;

File: tgui/packages/tgui/interfaces/LobbyMenu.tsx [hunk 9]
```diff
@@ -0,0 +1,608 @@
...
+                  onClick={() => act('ready')}
+                  icon={xenomorph_enabled ? 'virus' : 'check'}
+                  color={xenomorph_enabled ? 'purple' : undefined}
```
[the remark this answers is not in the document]
Watermelon914: this should really be handled by the theme. If you want a CRT-themed version, there should exist a CRT-themed xenomorph version instead

tgui/packages/tgui/interfaces/LobbyMenu.tsx
```diff
@@ -0,0 +1,608 @@
...
+                  <Flex>
+                    <Flex.Item>
+                      <Box className="gnarp" inline />
+                    </Flex.Item>
+                    <Flex.Item>Join the Hunt</Flex.Item>
+                  </Flex>
+                </LobbyButton>
+              </Stack.Item>
+            )}
+            {!!fax_responder_enabled && (
```
Watermelon914, line 503: Same thing here, this should just be a boolean value at all times.
If it can be undefined, declare a default value

tgui/packages/tgui/interfaces/LobbyMenu.tsx
```diff
@@ -0,0 +1,608 @@
...
+
+  const { themeDisable } = context;
+
+  useEffect(() => {
+    if (!themeDisable) {
+      setTimeout(() => {
+        ref.current!.style.display = 'block';
+      }, 1500);
+    }
+  });
```
Watermelon914, line 554: I assume you only want this running if themeDisable's value is changed. Otherwise, it will run every re-render.
```suggestion
  useEffect(() => {
    if (!themeDisable) {
      setTimeout(() => {
        ref.current!.style.display = 'block';
      }, 1500);
    }
  }, [themeDisable]);
```

tgui/packages/tgui/styles/interfaces/LobbyMenu.scss
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    .loadEffect {
+      height: 0;
+      animation: loading 1s steps(20, end) forwards;
+      animation-delay: 1.3s;
+    }
+
+    .sectionLoad {
+      animation: section-loading 1s forwards;
```
Watermelon914, line 29: Might be a good idea to allow players to disable animations if they want a more snappy experience. Complex animations can generally be slow and I've noticed the UI lagging when I hover over buttons, etc

tgui/packages/tgui/styles/interfaces/LobbyMenu.scss
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    @keyframes loading {
+      from {
+        height: 0;
+      }
+      to {
+        height: 100%;
+      }
+    }
```
Watermelon914, line 70: Width and height animations are slow in CSS, so it could be a contributing factor to the laggy UI experience.

Watermelon914: Check https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Performance/CSS for more details

Watermelon914 (changes requested):

code/modules/mob/new_player/login.dm
Watermelon914: if it does end up changing but only very occasionally, you can manually trigger a full update on clients (i.e. when the lobby art is loaded)

tgui/packages/tgui/interfaces/LobbyMenu.tsx
Watermelon914: ah wait, this might be a byond boolean (1,0), so this is probably fine

tgui/packages/tgui/interfaces/LobbyMenu.tsx
Watermelon914: nvm this is a byond boolean so this is fine

tgui/packages/tgui/interfaces/LobbyMenu.tsx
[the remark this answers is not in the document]
harryob: if we were using `useEffect`, we'd still need state to persist over re-renders

Blundir (changes requested):
please have an actual pred icon

PR #7772, opened 2024-12-10
Cigarette Vendors

thevandie:
# About the pull request

Adds two new cigarette vendors, whilst replacing the old one. That being the WY vendor and the Koorlander vendors. Koorlander made the lowest bid for the UA's exclusive cigarette contract, and thus on more UA-leaning maps I have replaced the vendors with primarily Koorlanders, and for more WY maps I have replaced them with WY vendors (if the map's alignment is uncertain, or more independent, I used Koorlanders). The Alaymer is now fully stocked with Koorlanders, so you have 3 (soon to be 5) options for smokes now instead of just luckies or cigars. Luckies, Koorlander Golds, Lady Fingers, and eventually both colors of American Spirits.

Executive Selects are 'available' from the WY vendor, but they are always out of stock as a gag. This makes them more unique and exotic, and it's a fun RP thing.

I also made an entire suite of new slogans for the WY vendor specifically, since the koorlander one was basically already covered.

I have Balaji Imperial and American Spirit cigs planned, but I'm not adding them until after the sprite re-org is done.

This does have sprites, but Blundir said it wouldn't effect the sprite reorg n said I had the go-ahead. 
<details>
<summary></summary>

image

</details>
I changed all of the vendors on every map with either a colony or luckystrikes vendor to WY or Koorlander ones. All of them. 
I ran Mapmerge before committing, so we should be good.

This is good for the game because it adds variety, adds new RP, adds world-building, and is just overall cool. 
Who doesn't want unique cigarette vendors?
<details>

Sprites:
image
image

In Use:
image
image
image
image

</details>

# Changelog
:cl:
add: Added WY and Koorlander vendors as new vending machines
imageadd: Added Koorlander and WY vendor sprites
maptweak: Replaced ALL of the cigarette vendors on every map with either a WY or Koorlander one
/:cl:


thevandie:
I just remember HYBRISA EXISTS, so yeah I'm just waiting till both that and the sprite freeze end in like, less than a week apparently. I'll reopen when that's all in, and it'll just have American Spirits and Balajis

PR #7800, opened 2024-12-12
surv fixes and tweaks 

BOBAMAx:
# About the pull request

redo of #7704 since I deleted my repo lmao OOPS
Removes duplicate equipment from surv goons and CLF (crowbars and flashlights)
removes second pistol spawn from CLF survivor
survivor equipment no longer tries to spawn in places it will not fit (fire axe in backpack, first aid kits in backpack of non-medics)
`add_survivor_weapon_civilian()` and `add_survivor_weapon_security()` assume you have an empty belt slot, and many presets did not and thus would not receive any ammo. This had been fixed for the following presets in the following ways:

- USCM recruiter: the M4A3 holster now is in the jacket slot, so the belt slot is freed.
- Engineer Presets: the `add_survivor_weapon_civilian()` proc is now before the toolbelt, so if a weapon IS generated, the ammo belt will spawn. Otherwise, the engineer will have a tool belt as normal.
- CLF survivor: The `spawn_rebel_belt()` and `add_survivor_weapon_security()` procs will never work at the same time, so I added a new `add_survivor_weapon_rebel` proc that assumes you have a 3 slot webbing attached to your uniform, which the CLF survivor now does. The ammo will now spawn in this webbing.

Also, engineer presets have had their super special welding tool moved to a pocket slot, as engineers would frequently fail to spawn with their backpack sidearm due to a lack of room.

equipment presets should actually spawn the things they are trying to spawn. equipment presets should not spawn the same items multiple times for no reason. Survivors that spawn with a weapon should have a reason to actually use the weapon and not instantly ditch it because you didn't get any ammo for it because the ammo belt didn't spawn. Engineer presets not spawning with a tool belt 40 percent of the time is acceptable in my mind as a !!!FUN!!! random chance thing that's not that big of a deal because tool belts are easier to come by than ar10 or uzi ammo. CLF survivors having their own unique pool of weapons with ammo in a webbing adds additional flavor (CLF survivors having mp27s and stuff like the ERT version) without removing their unique belt items such as katanas.

There is the issue of trucker survivors spawning with wheels in the hand and then `add_random_survivor_equipment()` on occasion failing to spawn a fire axe or a first aid kit due to the slot already being occupied, but by this point the preset could have a weapon in one hand and wheels in the other, so there is no place to safely spawn the gear and I don't know of any good solution to this problem. Silver lining: Fire axes and first aid kits are already broken so it makes no functional difference to be broken in a different way.

<details>

surv1
surv2
surv3

</details>

# Changelog
:cl: BOBAMA
balance: Recruiter, Engineer, and CLF survivors now have ammo for their weapons.
balance: Engineer survivors sometimes do not get tool belts.
fix: Survivors should now more consistently spawn with ALL of their intended gear.
fix: CLF and Goon survivors should no longer spawn with duplicate gear.
/:cl:


kiVts (approved):

code/modules/gear_presets/survivors/survivors.dm
```diff
@@ -313,13 +313,13 @@ Standart Survivors :	/datum/equipment_preset/survivor/scientist,
...
 	new_human.equip_to_slot_or_del(new /obj/item/storage/backpack/satchel/eng(new_human), WEAR_BACK)
-	new_human.equip_to_slot_or_del(new /obj/item/tool/weldingtool/largetank(new_human), WEAR_IN_BACK)
+	new_human.equip_to_slot_or_del(new /obj/item/tool/weldingtool/largetank(new_human), WEAR_R_STORE)
 	new_human.equip_to_slot_or_del(new /obj/item/stack/sheet/plasteel/med_small_stack(new_human), WEAR_IN_BACK)
 	new_human.equip_to_slot_or_del(new /obj/item/clothing/glasses/welding(new_human), WEAR_EYES)
 	new_human.equip_to_slot_or_del(new /obj/item/clothing/gloves/yellow(new_human), WEAR_HANDS)
 	new_human.equip_to_slot_or_del(new /obj/item/clothing/shoes/jackboots(new_human), WEAR_FEET)
-	new_human.equip_to_slot_or_del(new /obj/item/storage/belt/utility/full(new_human), WEAR_WAIST)
 	add_survivor_weapon_civilian(new_human)
+	new_human.equip_to_slot_or_del(new /obj/item/storage/belt/utility/full(new_human), WEAR_WAIST)
```
kiVts, line 322: looking back at it, this will never equip because proc above always adds something to waist. Either remove it or give a possibility to not spawn in a waist from proc above.

kiVts (changes requested):

code/modules/gear_presets/survivors/survivors.dm
BOBAMAx: `add_survivor_weapon_civilian()` has a 60 percent chance to not do anything at all. If it gives you a weapon, it also gives you an appropriate ammo belt, and the utility belt will just delete itself. The reference images in the main PR body show this behavior. (second image the proc rolled no weapon so the toolbelt was equiped. third image the proc did roll a weapon so there is no toolbelt.) This behavior is not intuitive unless you are familiar with the particular proc so I will add a comment to clarify.

kiVts: Didnt notice a prob(60) :patrice:

PR #7806, opened 2024-12-12
Move revolver tricks from the S&W to all revolvers + other related things

TheKillfish:
# About the pull request

Title. Additionally adds verbs and hotkey (Toggle Revolver Unique Action, currently unbound) to switch between spinning the cylinder and performing tricks, moves gun trick buffs to its own proc (only the S&W has any buff right now, the base proc just returns and revolvers need a variable set to TRUE to be able to apply buffs), and adds a new variable to item base for preventing item storage if set to TRUE (used by flip tricks to prevent an amusing but bad bug).

Revolvers are cool, but what's cooler? Doing tricks with them. Currently only the S&W 38. revolver, an infrequent colony revolver, can to tricks. This lets people easily switch between doing tricks or spinning their revolver's cylinder.

As for the new var, it's a lazy but functional solution to a bug, and a var that could be used in a lot of other places where an item shouldn't be put in storage for one reason or another.

<details>

</details>

# Changelog

:cl: TheKillfish
code: no_storage variable added to item base, used in inventory.dm. Currently only used to prevent storing revolvers while performing trick flips with one (due to a bug related to storing a revolver while trick flipping it), and also used on quick_equip verb so you don't get told "You are unable to equip that." while trying to quickstore a revolver you are trick flipping.
code: Moves revolver trick code from just the S&W to the revolver base type. The S&W is still the only one that provides buffs for successful tricks, but the means to add trick buffs to other/future revolvers is present.
add: Verb and hotkey for switching the Unique Action of revolvers between spinning the cylinder and trying to perform slick tricks.
/:cl:


Blundir (changes requested):
As we've dicussed before, limit it's availability to a trait, also there really should be a list of guns that can do that, instead of every single one

TheKillfish:
Not a fan of limiting the roster of guns *and* having an item you need to use to even do tricks, so closing this and focusing on other things.

PR #7813, opened 2024-12-13
reduce the amount of times you have to fix chat

harryob:
this never loads, and we can detect when it hasn't loaded, so why not automatically reload it for players

:cl:
fix: fixes chat never loading correctly, by just rebooting it when it fails to load
/:cl:


cm13-github:

cm13-github:

harryob:
need to see if this persists on 516

PR #7818, opened 2024-12-14
Limits colossal sandwich to 15 layers

Blundir:
# About the pull request

Title, you can't put more than 15 items on it

Let's be real here, i know that it's fun, but it's also lrp, causes lag, entire server was lagged out because of it on multiple occasions, not to mention that it blocks vision and is generally buggy.

Discord_SyKShw3YhY

<details>

The maximum size of it

dreamseeker_5U2ZSwuGaM

</details>

# Changelog
:cl:
del: limits colossal sandwich to 15 layers
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/food/sandwich.dm
+++ b/code/game/objects/items/reagent_containers/food/sandwich.dm
@@ -31,7 +31,7 @@
 		to_chat(user, SPAN_DANGER("If you put anything else on \the [src] it's going to collapse."))
 		return
 	if(length(src.contents) >= 15)
-		to_chat(user, SPAN_DANGER("\the [src] is already massive! You can't add more without ruining it."))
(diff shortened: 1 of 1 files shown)


Git-Nivrak:
How about you optimize it instead of removing soul

Blundir:
Doesn't mitigate other things

fira:
Assuming there's no way to remove ingredients (i don't see any?) there might be an easy optimization which is just storing the sandwich top overlay and removing it instead of rebuilding the whole overlays

But i don't know if that's enough and that doesn't really solve the other issues

cuberound:
jeh, rebuilding it from bottom up every time is kinda silly, but reducing layers solves more then just performance issue

realforest2001 (approved):

PR #7827, opened 2024-12-15
Overwatch can hear speech and see emotes

ihatethisengine:
# About the pull request
Overwatch consoles let you hear speech and see emotes around the marine you are overwatching. His speech will also appear as langchat over the console. Basically close to how broadcasting works rn. Doesn't show speech and AUDIBLE emotes if there are comms issues unless on the same z-level.

Now supports ground console. Might be bug heavy, needs testing. 

It was in the movie. And we have to make observer+ role a bit more interesting.

<details>
One day I'll record something, really lazy rn though and my vm I use to code is bugged af
</details>

# Changelog
:cl: ihatethisengine
add: Overwatch can hear speech and see emotes.
/:cl:


code/modules/cm_marines/overwatch.dm
```diff
@@ -644,6 +683,11 @@
 			var/obj/item/clothing/head/helmet/marine/helm = H.head
 			return helm.camera
 
+/obj/structure/machinery/computer/overwatch/proc/get_helm_from_target(mob/living/carbon/human/H)
+	if(current_squad)
+		if(H && istype(H) && istype(H.head, /obj/item/clothing/head/helmet/marine))
+			var/obj/item/clothing/head/helmet/marine/helm = H.head
+			return helm
```
harryob, line 690:
```suggestion
	if(!current_squad)
		return

	if(istype(target) && istype(target.head, /obj/item/clothing/head/helmet/marine))
		var/obj/item/clothing/head/helmet/marine/helm = target.head
		return helm
```

code/modules/cm_marines/overwatch.dm
```diff
@@ -644,6 +683,11 @@
 			var/obj/item/clothing/head/helmet/marine/helm = H.head
 			return helm.camera
 
+/obj/structure/machinery/computer/overwatch/proc/get_helm_from_target(mob/living/carbon/human/H)
```
harryob, line 686:
```suggestion
/obj/structure/machinery/computer/overwatch/proc/get_helm_from_target(mob/living/carbon/human/target)
```

harryob (changes requested):

ihatethisengine:
It's a draft why are you suggesting changes :sob:

harryob:
> It's a draft why are you suggesting changes 😭

do what i want

harryob (approved):

code/game/machinery/computer/groundside_operations.dm
```diff
@@ -358,13 +387,52 @@
...
+
+/obj/structure/machinery/computer/groundside_operations/proc/transfer_emote(obj/item/camera, mob/living/sourcemob, emote, audible = FALSE, show_message_above_tv = FALSE)
+	SIGNAL_HANDLER
+	if(inoperable())
+		return
+	var/target_zs = SSradio.get_available_tcomm_zs(freq)
+	if(audible && !(z == sourcemob.z) && !((z in target_zs) && (sourcemob.z in target_zs)))
+		return
+	if(show_message_above_tv)
+		langchat_speech(emote, get_mobs_in_view(7, src), null, null, TRUE, LANGCHAT_FAST_POP, list("emote"))
```
harryob, line 422: one nit - please use named args instead of passing null for your langchat_speech calls

harryob (changes requested):
jk - also most of the logic in the transfer speech/transfer emote procs are identical barring the checks on prefs - no way to reduce these down?

ihatethisengine:
> also most of the logic in the transfer speech/transfer emote procs are identical barring the checks on prefs - no way to reduce these down?

Don't really see a way because emotes and speech use different sets of args. I can move Z check into a different proc but even then I would have to check if this is not an audible emote to pass it anyway.

harryob (approved):

PR #7833, opened 2024-12-17
fixes duplicate req door

Detective-Google:
req southline got bugged recently and has two stacked doors. I fixed it. it's better now.
before
image
after
image

:cl:
fix: fixed duplicate windoors on the reqline
/:cl:


tallfission:
Soul removal..

Drulikar (changes requested):
See how much work adding this lint would be (as in how much would be detected on other maps to be fixed):
multiple_doors.yml
```yml
/obj/structure/machinery/door:
  banned_neighbors:
    /obj/structure/machinery/door:
      identical: true
 ```
 Edit: (Revised path)

PR #7834, opened 2024-12-17
Adds a dedicated IO Prep Room

efzapa:
# About the pull request
IOs now have their own dedicated prep room, with all the same vendors as they had in the Intel Processing Room, their Spawn location has also been moved.

Mostly QOL for ASOs but it's a combination of not wanting to have to deal with IOs leaving heaps of gear in Intel and not cleaning up after themselves, and just that in my opinion, a prep area and work area shouldn't really be merged together as it usually leads to problems with one or the other causing problems for each other, be it one side lying gear on the floor or the other throwing it away.

<details>

image

</details>

# Changelog
:cl:
mapadd: Added an IO Prep Room
/:cl:

### DIFF
--- a/code/game/area/almayer.dm
+++ b/code/game/area/almayer.dm
@@ -628,6 +628,11 @@
 	icon_state = "livingspace"
 	fake_zlevel = 1 // upperdeck
 
+/area/almayer/living/iobunks
+	name = "\improper Intelligence Officer Bunks"
+	icon_state = "livingspace"
+	fake_zlevel = 1
+
 /area/almayer/living/commandbunks
 	name = "\improper Commanding Officer's Bunk"
 	icon_state = "livingspace"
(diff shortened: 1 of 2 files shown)


thevandie:
does this replace the armorer's workshop and the little cafe with the stuff in it?

efzapa:
> does this replace the armorer's workshop and the little cafe with the stuff in it?

yes, as far as I am concerned if Officers or Enlisted want to have a hangout RP spot for eating, they have the Officers Mess and Mess Hall respectively

armorers shed isn't really used either and is just maint fluff with gun boxes, if Nanu wants I can stuff it back in somewhere else

thevandie:
Alright thats a -1 from me

SASoperative:
Why not change the IO 'lobby' area that is only really used to store crates that have not been processed yet? 
I can't remember the last time it has been used for anything that is not just crate storage, or where trash from the crates normally go, if it even is used at all.

Even than, picking up the gear of three people and just throwing it into a rack so its out of sight is also maybe 30 seconds. IO's are not super messy in comparison to whole squads.

I won't deny, IO's having their own dedicated prep would be nice, but I don't think it should come at the cost of two parts of the ship, one of which I know is used on a frequent basis by marines (Or at least the L42s in the armory are)

efzapa:
will re-map this eventually

PR #7861, opened 2024-12-20
Fix lots of random throwing and explosion runtimes

Drulikar:
# About the pull request

This PR attempts to mitigate various oversights and runtimes that occur with explosions and throw_atom. They are really getting out of hand. This was kicked off by a round in watchtower testing that caused the master controller to crash from recursive throw code.

Should mean aspects like two crushers colliding at each other even is handled, not to mention aspects that would making throwing an atom impossible (like trying to throw an atom at an area - why is an area considered an atom? - why is an area a result in range??). There are likely performance gains from this, but then again, code not halting because of runtime could mean more is being done too...

<details>

image

</details>

# Changelog
:cl: Drathek
fix: Fixed various runtimes caused by explosions and throw_atom code
/:cl:


code/game/objects/items/shards.dm
```diff
@@ -172,7 +172,7 @@
 		playsound(M, 'sound/effects/meteorimpact.ogg', 35)
 		M.at_munition_interior_explosion_effect(cause_data = create_cause_data("Anti-Tank Rocket", U))
 		M.interior_crash_effect()
-		M.ex_act(1000, get_dir(U, T), create_cause_data("Anti-Tank Rocket", U))
+		M.ex_act(1000)
```
fira, line 175: I don't get this change, are you removing the direction to stop throwing?

code/modules/mob/living/living_defense.dm
```diff
@@ -40,8 +40,10 @@
 	var/impact_damage = (1 + O.throwforce*THROWFORCE_COEFF)*O.throwforce*THROW_SPEED_IMPACT_COEFF*O.cur_speed
 
 	var/datum/launch_metadata/LM = O.launch_metadata
+	var/launch_meta_valid = istype(LM)
```
fira, line 43: I also don't understand this bit - why use istype() for this? Can launch_metadata be another type of value? At a glance it looks to me like using ?. would be enough

code/game/objects/items/shards.dm
Drulikar: ex_act doesn't take these arguments. 
image

code/modules/mob/living/living_defense.dm
Drulikar: Replication of LM handling I saw elsewhere, so standardized across the board. If you prefer I can just `if(!LM)` everywhere? Examples on master: https://github.com/search?q=repo%3Acmss13-devs%2Fcmss13%20istype(LM)&type=code

harryob (approved):

PR #7862, opened 2024-12-20
Proc Redefinition Linting

Drulikar:
# About the pull request

This PR turns on redefinition warnings (which should fail lints since we Werror). This PR needs testing though as there are going to be actual changes from this refactor. One aspect for example I'm not sure about is whether an area power_change should just always update_icon (vendor code mistakenly was causing this). I have opted to just include that in the proc now and removed redundant calls. Another couple notable example are that `repair_robotic_organs` now has an implementation for `repeat_step_criteria` and `pkd_special` revolvers that aren't a `l_series` subtype will no longer have burst changes.

TGS DMAPI also fails this, but I am waiting advice on what to do about this, so for now it will serve just as a test to ensure lints fail. See https://github.com/tgstation/tgstation-server/issues/2056 (This is now fixed)

In almost all cases, a redefinition is not necessary. And in the few cases it is, `CAN_BE_REDEFINED(TRUE)` can be used for the proc. See code changes for examples of all the obvious mistakes allowing redefinitions has caused.

<details>

</details>

# Changelog
:cl: Drathek
balance: The PKD special burst values have changed for non-"Double" (l_series) variants
code: Code no longer uses proc redefinitions with a few exceptions for global macros
fix: Fixed various things like all backpacks (instead of just JIMA flags) processing
/:cl:


cm13-github:

cm13-github:

harryob (approved):

PR #7863, opened 2024-12-20
Fix hugger trap refilling itself

JackieEstegado:
# About the pull request

Hugger traps kept refilling themselves like this cause spaghetti code restarted the death timer on already dead huggers, which caused them to check for places they can crawl back into again. :D

https://github.com/user-attachments/assets/8ebf0095-e423-46d8-aa13-b119803c60ad

Only good bug is a dead bug.

Works.

# Changelog

:cl:
fix: Facehugger resin traps no longer refill themselves with dead huggers.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
@@ -86,7 +86,7 @@
 		return
 	addtimer(CALLBACK(src, PROC_REF(check_turf)), 0.2 SECONDS)
 
-	if(!death_timer && hivenumber != XENO_HIVE_TUTORIAL)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
Drulikar, line 1: This works, but you should also amend `/obj/item/clothing/mask/facehugger/proc/die()`
moving
```js
	if(stat == DEAD)
		return
```

To be after `death_timer = null`

So that we can ensure there isn't a jump timer either for a dead hugger (pretty sure I've seen reports of this too). Technically this would make your existing change moot since it would just del_timer the death_timer now, but its fine to keep your change.

Drulikar (changes requested):

Drulikar (approved):

PR #7867, opened 2024-12-21
jolly grinchmas (sets the festiviser to just be on christmas day)

harryob:
festivities have concluded early

im fine with a more subtle festivizer that last longer but i need to be not on my phone to do that

:cl:
code: Christmas Festivizer now only applies on Christmas Eve and Day.
/:cl:

### DIFF
--- a/code/modules/decorators/christmas.dm
+++ b/code/modules/decorators/christmas.dm
@@ -5,7 +5,7 @@
 
 // define when you wanna decorate
 /datum/decorator/christmas/is_active_decor()
-	return is_month(12) && (is_day(21) || is_day(22) || is_day(23) || is_day(24) || is_day(25) || is_day(26))
+	return is_month(12) && is_day(25)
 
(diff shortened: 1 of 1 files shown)


Ben10083:
Eh, to be honest christmas eve should also be fine

code/modules/decorators/christmas.dm
```diff
@@ -5,7 +5,7 @@
 
 // define when you wanna decorate
 /datum/decorator/christmas/is_active_decor()
-	return is_month(12) && (is_day(21) || is_day(22) || is_day(23) || is_day(24) || is_day(25) || is_day(26))
+	return is_month(12) && is_day(25)
```
realforest2001, line 8:
```suggestion
	return is_month(12) && (is_day(24) || is_day(25))
```

realforest2001:
Should be a thing anyway. No need to TM it.

PR #7882, opened 2024-12-22
Scaling basic resin construction cost

Drulikar:
# About the pull request

This PR is aimed at discouraging basic resin constructions (walls/membranes/doors) in areas that are already saturated with resin constructions. I am not certain yet if 40% saturation is the right amount to start scaling (as in I generally don't want it to really ever come in to play for most rounds), and if plasma cost is harsh enough when its active. The plasma cost will always be under plasma_max if there even is enough plasma_max needed to do the construction normally.

Current values:
image

I also made the admin niche message for enclosed dead bodies more relevant.

Should mitigate, but not make impossible, saturating areas heavily with basic resin constructions. I personally think marines have enough tools to quickly enough deal with this occurring even in caves, but I guess I'm in the minority.

<details>

</details>

# Changelog
:cl: Drathek
balance: Added scaling to the plasma cost of basic resin constructions when an area is at least 40% saturated.
admin: Niche messaging for sealing off dead bodies now only applies to revivable bodies
refactor: Refactored some resin construction code
/:cl:


HeresKozmos:
For clearing structures you already get:
- Boot knives, machetes, most melee weapons
- Grenades of various types/kinds
- Mortar shells, all but camera shells 
- CAS, several different configurations (some incredibly powerful at area clearing*)
- Actual bullets, you can shoot them if you really need to
- Breaching shotgun, annihilates most resin constructs AND normal or reinforced walls
- C4, carried by most technicians
- Breaching Hammers, carried by Synthetics

This is an absolutely pointless PR designed to make builders even more unfun to play.

Drulikar:
> For clearing structures you already get:
> 
> * Boot knives, machetes, most melee weapons
> * Grenades of various types/kinds
> * Mortar shells, all but camera shells
> * CAS, several different configurations (some incredibly powerful at area clearing*)
> * Actual bullets, you can shoot them if you really need to
> * Breaching shotgun, annihilates most resin constructs AND normal or reinforced walls
> * C4, carried by most technicians
> * Breaching Hammers, carried by Synthetics
> 
> This is an absolutely pointless PR designed to make builders even more unfun to play.

I am aware, but you aren't aware how much it takes to get to 30% saturation in an area.

TheManWithNoHands:
> I am aware, but you aren't aware how much it takes to get to 30% saturation in an area.

Bro, tell that then. That is important info to get perspective on this.

Drulikar:
> > I am aware, but you aren't aware how much it takes to get to 30% saturation in an area.
> 
> Bro, tell then how much is needed to get 30%. That is important info to get perspective on this.

In my description it says:
> (as in I generally don't want it to really ever come in to play for most rounds)

Its basic resin constructions to turfs that can be opened. The code in the Files Changed tab. Western caves for example calculates out to have 2015 openable turfs and in distress signal starts at 292. 2105 * 0.3 is 631.5. So you need multiple screen worth of walls/doors/resin to even start getting a higher cost.

It could affect small areas more than I want maybe I need to ignore areas that have say under 500 openable turfs or have more of a sliding scale depending on the cap, but the goal of this is to not really ever feel this effect unless someone is really saturating walls/membranes/doors somewhere.

Diegoflores31:
this makes lessers unable to build at some point is that intended ???

Drulikar:
> this makes lessers unable to build at some point is that intended ???

False, its capped to max_plasma

Red-byte3D:
How is this going to affect maps that are already tight enough as it is? LV 624 is a bad example because theres multiple entry points through the cave systems, esentially meaning it all crashes into one place anyway so majority of the defenses will be around the cave area you've hived in, but for example on a map like trijent dam you have to build up the west podlock, south podlock, science, and the east of the podlock area, which are all basically a total of 2-3 screens.

It's alot of places to build in a small radius, not considering the fact that none of those said areas are un-cassable so re-building after a cas strike would probably cost a lot of plasma.
Maybe the scaling should work differently if the area its in is open to cas or mortar? 

Not to mention most probably that comm holds are going to be even harder to hold due to the density you have to build them in so maybe a grace period on tiles before the cost actually even begins to increase?

All just worries i have, but balance wise i think people already dont have alot of reasons to play builders as menitoned above with how easy people can clear walls, meaning most of them dont even amount to much. I think before a pr like this is viable we need to consider either increasing the structure health, or just nerfing the efficiency & availability of the tools they have to deal with said constructions

Comxy:
I am thinking increasing the damage multiplier on melee weapons for example 25% saturation might be a better solution. It seems more player friendly for building castes and then you can also lower the initial threshold to make clearing cluttered structures less time consuming.

Drulikar:
I really can't get it into commentors heads how much it takes to even reach this threshold. So I'm sorry. I can only reiterate this is intended to never be felt in a regular round. If it can be it will be adjusted.

PhantornRU:
But logically it's the other way around... If there's a lot of resin nearby, it should be easier for you to build more. But here it's the other way around. Something "incomprehensible" based on something incomprehensible starts to interfere with building.

I'd better make another restriction if you really want it. The more plasma - the faster the building is built. The less - the slower. If you really want restrictions on construction...

Drulikar:
> But logically it's the other way around... If there's a lot of resin nearby, it should be easier for you to build more. But here it's the other way around. Something "incomprehensible" based on something incomprehensible starts to interfere with building.
> 
> I'd better make another restriction if you really want it. The more plasma - the faster the building is built. The less - the slower. If you really want restrictions on construction...

This suggestion goes completely against the point of this PR.

harryob (approved):
tm-able

code/modules/cm_aliens/XenoStructures.dm
```diff
@@ -481,14 +483,16 @@
 
 /obj/structure/mineral_door/resin/Destroy()
 	relativewall_neighbours()
-	var/turf/U = loc
+	var/area/area = get_area(src)
+	area?.current_resin_count--
 	spawn(0)
```
Segrain, line 488:
```suggestion
	area?.current_resin_count--
	var/turf/base_turf = loc
	spawn(0)
```
You cannot do `get_step(loc, cardinal)` below, because it is inside of a `spawn()`, and by the time code gets to this point, the `..()` has already deleted the source object and there is no `loc` to use. That `var/turf/U = loc` was actually meaningful, without it you can be left with a freestanding door.
image

code/modules/cm_aliens/XenoStructures.dm
```diff
@@ -481,14 +483,16 @@
...
-			if(!istype(T)) continue
-			for(var/obj/structure/mineral_door/resin/R in T)
-				R.check_resin_support()
+		var/turf/current_turf
+		for(var/cardinal in GLOB.cardinals)
+			current_turf = get_step(loc, cardinal)
+			if(!istype(current_turf))
+				continue
+			for(var/obj/structure/mineral_door/resin/door in current_turf)
+				door.check_resin_support()
```
Segrain, line 495:
```suggestion
		var/turf/adjacent_turf
		for(var/cardinal in GLOB.cardinals)
			adjacent_turf = get_step(base_turf, cardinal)
			if(!istype(adacent_turf))
				continue
			for(var/obj/structure/mineral_door/resin/door in adjacent_turf)
				door.check_resin_support()
```

Segrain (changes requested):

grandalff:
genuinely horrid idea, you are finding non issues which dont even matter much (as resin constructions, when spammed, can be meleed very swiftly and easily, plus marines have CAS, OB's) and are trying to somehow "fix" it. Noone even wants to play builders, this just makes their life even more painful for no reason.

harryob (approved):

PR #7887, opened 2024-12-22
Demo Specialist Loader Assistant Role

Geevies:
# About the pull request

Copies the implementation of the Spotter role to make one for the Demo spec as well, called the "Loader", it has no real special abilities, other than a dedicated body to carry rockets and reload them. Their pamphlet gives them a little more endurance.

Encouraging teamplay amongst the marines will always be good for the game. Plus, it's good for roleplay.

i compiled, ran the server, went pfc, got the stuff, checked that my HUD and ID updated correctly, and that i could wear the armor

# Changelog
:cl:
add: Added a Loader assistant role to the Demolition specialist
/:cl:


Red-byte3D:
I think this is a good addition but i personally cannot wait in 3 months for the "It's the loader, it should do something special" PR
(im just saying this so i can say i said so in a few months)

RainbowStalin:
I personally despise what the spotter has become, because it is now a 30/70 of either being an actual spotter, or someone who wants to PFC but with nightvision. And I can see this being another sought after thing now with an extra endurance level.

TheManWithNoHands:
Fix changelog. Your Git name needs to be on the top, on the rigth of CL.
Also you gave extra rockets and C4 to the SADAR with this, that needs to be listed as well. Thats a SADAR buff.

Diegoflores31:
i really see this as a "free gaming" role , same as current spotter . SADAR already has the extra backpack and the mechanics for the loading . there is no need to give more free gaming

private-tristan:
It really doesn't need to give extra endurance (as that would only ever be given out to robustos so that they can frag).
Instead, can't it increase the speed when reloading the SADAR directly (Demo wielding, facing away, Loader behind, loads rocket). It's a very underused feature which could use some more incentive to use it, especially with a role based around loading rockets.

harryob:
> Your Git name needs to be on the rigth of CL. 
no it does not

TheManWithNoHands:
> no it does not

*looks his prs up.* Huh, your rigth it doesnt.... i been living a lie for almost a year.

kiVts (changes requested):
the fact that you can do assisted reload with rpg is an arcane knowledge, make a guide to it similarly how surgical case has one in its description to how fix an IB.

kiVts:
Oh and also fix this 
image

Geevies:
> Oh and also fix this image

separate PR made for this one, so the fix doesn't get bogged down with this PR https://github.com/cmss13-devs/cmss13/pull/7939

chefoxara:
i like this please ignore all the "this wil be powergamed 🤓 " nerds in the comments and add this!!!!!!

harryob (approved):

PR #7891, opened 2024-12-22
Unnest fixes

JackieEstegado:
# About the pull request

1. Fixed mobs getting unnested due to the wall they are on being upgraded by checking if there is still a nestable surface there after the turf change.

2. Fixed mobs not regaining their original pixel offset after being unnested while jittering (ie. from late stage larva).
During the last change of related code the original offset was simply never stored.
Technically we aren't preserving the mob type's intended offset like this, but I am just replicating the behaviour of the parent proc (/obj/structure/bed/afterbuckle), but fixing that I consider out of scope here.

Also cleaned up a couple of redundant pieces of code.

<details>

</details>

# Changelog
:cl:
fix: Mobs no longer get unnested when the wall that they are on gets upgraded.
fix: Mobs now regain their original pixel offset upon being unnested even when jittering.

/:cl:

### DIFF
--- a/code/game/turfs/walls/walls.dm
+++ b/code/game/turfs/walls/walls.dm
@@ -103,7 +103,7 @@
 		var/list/turf/cardinal_neighbors = list(get_step(src, NORTH), get_step(src, SOUTH), get_step(src, EAST), get_step(src, WEST))
 		for(var/turf/cardinal_turf as anything in cardinal_neighbors)
 			for(var/obj/structure/bed/nest/found_nest in cardinal_turf)
-				if(found_nest.dir == get_dir(found_nest, src))
+				if(found_nest.dir == get_dir(found_nest, src) && !istype(.,/turf/closed/wall/resin/))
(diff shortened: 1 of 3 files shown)


code/game/turfs/walls/walls.dm
```diff
@@ -103,7 +103,7 @@
 		var/list/turf/cardinal_neighbors = list(get_step(src, NORTH), get_step(src, SOUTH), get_step(src, EAST), get_step(src, WEST))
 		for(var/turf/cardinal_turf as anything in cardinal_neighbors)
 			for(var/obj/structure/bed/nest/found_nest in cardinal_turf)
-				if(found_nest.dir == get_dir(found_nest, src))
+				if(found_nest.dir == get_dir(found_nest, src) && !src.density)
```
kiVts, line 106:
```suggestion
				if(found_nest.dir == get_dir(found_nest, src) && !density)
```

kiVts (changes requested):

kiVts (approved):

PR #6588, opened 2024-06-28
Squad roles positions code QoL

blackcrystall:
# About the pull request

Small change, to make it a little bit better for any further updates of roles and multi faction gameplay, plus change one letter vars or UPPERCASE names to normal, just a little bit of code formating

# Changelog

:cl: BlackCrystalic
code: squads now have lists for roles limit and taken number, not hardcoded role list in vars, plus one letter vars change
/:cl:

PR #7896, opened 2024-12-23
Put players with no squad preference in random squads again

sg2002:
# About the pull request
This fixes a problem introduced in #6588, where marines with no squad preference started to get assigned squads in a deterministic order, where if Alpha has a job slot open, you always get assigned to it, then to Bravo and so on... And since there's no cap on riflemen, all riflemen with no squad preference would always be assigned to Alpha.

Before I fully understood that such a minimalist change would work, I actually [did a bigger rewrite](https://github.com/cmss13-devs/cmss13/commit/d31d22a4fe897dbe5fd62eb0e986de64720e8f90). That version is more computationally complex, but may be simpler to understand, due to being somewhat better abstracted. Just giving that as an option for maintainers too, if they're interested.

I did a bit of testing on the local server, seems to work fine.

Do we really want Alpha to suffer me as their SL every single round?
<details>

</details>

# Changelog
:cl: sg2002
fix: players with no squad preference get correctly put into random squads
/:cl:

### DIFF
--- a/code/game/jobs/job/marine/squads.dm
+++ b/code/game/jobs/job/marine/squads.dm
@@ -1,8 +1,6 @@
 //This datum keeps track of individual squads. New squads can be added without any problem but to give them
 //access you must add them individually to access.dm with the other squads. Just look for "access_alpha" and add the new one
 
-//Note: some important procs are held by the job controller, in job_controller.dm.
-//In particular, get_lowest_squad() and randomize_squad()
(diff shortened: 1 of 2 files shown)


harryob (approved):

sg2002:
That shuffle is useful to have for lowpop purposes where if you're the only person rolling SL(engie, medic) you'd still be randomized instead of always getting Alpha.

PR #7908, opened 2024-12-25
Fix of live server silent shutdown + How to repeat this bug on LIVE server guide

blackcrystall:
# About the pull request
You think this bug non existen?
Here steps how you can repeat it on LIVE & TEST server
There exist two ways, as far as I know.
1. Using given you back in 2018 admin rights (aka you got at least for one second admin rights, so your byond client side think you can do funny), that removed next round, but you for some reason never closed your game window
 1.1 just click here, and nobody ever find out, if reboot finished without ERRORs due to sh1tcode (if it's run in to problem, your client will be logged as usr: CKEY, in runtime log), as long as nobody BAN you or force close your game window, you can repeat it (even if you got kicked, you can just click reconnect)
image

2. Using MODIFIED byond client (aka client for cheating)
 2.1 Find out by yourself what function call you need to do (I really don't remember what the hell is going with byond backend here, you can just try to copy data -but you still need to know how to burrow it- on test server that you send to it, and then resend to LIVE server)

Please don't abuse it, I just explaining what a bug I fixing here!!!
(this work probably in all byond games, find out by yourself)

I don't think I'm funny, but It's a lie if you think I'm not funny.

Also big thanks to lummox, I like byond, and more, read it here what I like also!!!
1. When do we move from 32 to 64 (probably never)?
2. maxx and maxy, I like those, they steal max cap on some objects... funny to load 3millions elements and then findig out you run out of some type and due to it get OOM/OOR error... because your map load to many tiles... where they don't need to be. what a bad idea to make it per z level... (but still have a lot of resources)

I have no idea

# Changelog

:cl: BlackCrystalic
fix: You can't do silent server shutdown from byond client side backend call to server
/:cl:


Watermelon914:
Closing in favour of #7909

PR #7909, opened 2024-12-25
Fixes maintainer access to the profiler and clears profiler access whenever admins are loaded.

Watermelon914:
# About the pull request

As explained by the title.

Prevents a vulnerability where someone who previously had access to the profiler retains access despite no longer requiring it.

No player facing change so changelog not needed

### DIFF
--- a/code/modules/admin/admin_ranks.dm
+++ b/code/modules/admin/admin_ranks.dm
@@ -64,6 +64,10 @@ GLOBAL_LIST_EMPTY(admin_ranks) //list of all ranks with associated rights
 		C.admin_holder = null
 	GLOB.admins.Cut()
 
+	//Clear profile access
+	for(var/admin in world.GetConfig("admin"))
+		world.SetConfig("APP/admin", admin, null)
+
 	load_admin_ranks()
 
 		//load text from file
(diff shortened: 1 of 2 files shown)


Drulikar (changes requested):
I do not understand why on event server this works (after a reconnect - not first connect); but not on live (even with reconnects). I can confirm for both servers the cfg/admin.txt is assigned to be `some_ckey role=admin` for only those who have connected this round, and the perms for the file are the same for both servers.

But the aspect of requiring a reconnect is wonky as is. So try setting up the admin.txt during init (e.g. after load_admin_ranks) rather than on associate?

Watermelon914:
> I do not understand why on event server this works (after a reconnect - not first connect); but not on live (even with reconnects). I can confirm for both servers the cfg/admin.txt is assigned to be `some_ckey role=admin` for only those who have connected this round, and the perms for the file are the same for both servers.
> 
> But the aspect of requiring a reconnect is wonky as is. So try setting up the admin.txt during init (e.g. after load_admin_ranks) rather than on associate?

I was doing testing with harryob, and the admin.txt was getting populated with ckeys. The problem isn't related to code, it's related to the cm-live server and it's unclear what might be causing it. In any case, this PR is still useful to merge since it clears up a potential vulnerability where people stay in the admin.txt despite probably not being an admin anymore.

PR #7933, opened 2024-12-28
Change the color of the Squad Leader pointing arrow to the color of the squad

casperr04:
# About the pull request

This will change it so the big green arrow SL's get will now change color to the respective color of the squad. This also works for or any other squads that have their own equipment color, like UPP squads.

Makes it easier to spot from which squads SL the arrow is coming from. Also looks cool.

<details>

https://github.com/user-attachments/assets/3251caf6-2cb4-471f-a319-df2f842f389f

</details>

# Changelog
:cl:
add: Squad Leader pointer arrows are now colored based on the color of the squad.
/:cl:


Git-Nivrak:
Have you tested this on xenos? It looks like you will get a null runtime if a xeno tries to point as they do not have a assigned_squad variable

casperr04:
> Have you tested this on xenos? It looks like you will get a null runtime if a xeno tries to point as they do not have a assigned_squad variable

Oops.

harryob (approved):

PR #7942, opened 2024-12-29
No more Forklift man

DoomLaser9:
# About the pull request

Changes all forklift ID names to generic names.

Joke names are not allowed. If using these items is a bannable offense, then they need to be either removed or changed.

N/A

# Changelog
:cl: Laser9
maptweak: Replaced all forklift ID names with generic names.
/:cl:


Detective-Google:
Last names often come from professions, for example your PR aims to change some of these IDs to say "John Smith", the name Smith comes from blacksmiths. By this logic, in the future of 2182 a similar thing likely has happened with forklift operators.

Nanu308 (approved):
Mapping fine, this is a @harryob decision tho as i presume its staff related

irakligabunia:
> Last names often come from professions, for example your PR aims to change some of these IDs to say "John Smith", the name Smith comes from blacksmiths. By this logic, in the future of 2182 a similar thing likely has happened with forklift operators.

John Forklift is a name that will get you noted, getting people the idea that it's okay only to bait them into a ticket is a bad thing

PR #7943, opened 2024-12-29
Show all channels a headset has on examining it.

Awantje:
# About the pull request

Tl:DR: If you examine your headset the same ui will pop up that pops up when you normally click it.

Shift click a headset to examine it. And it will have a pop up listing it's frequencies open up.
The pop up shows you all frequencies it has and is always updated based upon the keys inserted.

The first line I override the examine proc.
The second line we ensure you cannot remotely interact with it. Or get the menu if you normally would not get the menu.
So a shift click cannot be used to bypass checks on the pop up. Such as remote interaction from far away. Yautja and synths still/also get the pop up.
Next we ensure we stop updating changes if the user closes the window. 
Then we open the window.
And last we make sure to always call the parent proc so that we also display the description.
It also fixes "Fixes issue #7918 "

This is good because currently the channels are listed in the description when clicking the headset.
Now, there is no need to manually add it to each headset it's description. And it updates when keys are inserted or removed.
Currently, some descriptions are missing some channels. And some headsets do not have any channels in their description.
This helps also helps newer players, because it's more obvious, always correct/self updating.
Because it's a pop up it won't get lost in the chat log and is also easier to see/write down.
 
Unless I also managed to tape my inputs you would not know what buttons I was pressing.
But, I spawned in as Human and shift clicked a headset in my hand, one I was wearing, one in a backpack and one on the floor twice. Once next to me and once remotely. I got the expected result each time. I also did the same as a synth, a yautja and a monkey. The results were as expected. As a xeno I could not pick it up only slash it.

# Changelog

:cl:Awan
qol: If you examine a radio that is on your person it's menu listing all frequencies will now pop up.
/:cl:


Awantje:
Now, for a follow up question. Isn't having the channels listed in the description kinda redundant now?
I did it this way because it is future proof. Instead of manually having to edit/update descriptions.
But as stated there is still headsets that do not (correctly) list their channels.
The only reason I did not remove it is because currently you can examine from a distance and see the description.
If I remove the strings in the desc variables you can no longer from a distance see what channels a headset has.

harryob:
the title of your pr needs to say what you're trying to fix, or accomplish

Awantje:
> the title of your pr needs to say what you're trying to fix, or accomplish

Sorry, my bad. Fixed it.

harryob:
> > the title of your pr needs to say what you're trying to fix, or accomplish
> 
> Sorry, my bad. Fixed it.

you’re all good, the title is just the name of the squash commit when we merge the pr - best it’s helpful!

code/game/objects/items/devices/radio/headset.dm
```diff
@@ -137,6 +137,12 @@
 	user.set_interaction(src)
 	tgui_interact(user)
 
+/obj/item/device/radio/headset/examine(mob/user as mob)
+	if(ishuman(user) && loc == user)
+		user.set_interaction(src)
```
harryob, line 142: i wouldn't really worry about this, interactee stuff is mostly dead code

```suggestion
```

harryob (changes requested):

code/game/objects/items/devices/radio/headset.dm
Awantje: done

harryob (approved):

PR #7945, opened 2024-12-29
roofs using queue instead of recursion

cuberound:
# About the pull request
dreammaker had recursion issues with big roofs, this might solve it using using queue

allows for big roofs

<details>

</details>

# Changelog
:cl:
code: improves initialization of masternode for structure/roof
/:cl:


code/game/objects/structures/roof.dm
```diff
@@ -108,13 +115,13 @@
...
-		for(var/obj/structure/roof/roof in connected_roof)
-			qdel(roof)
-	return ..()
+    if(connected_nodes)
+        for(var/obj/effect/roof_node/roof_node in connected_nodes)
+            qdel(roof_node)
+    if(connected_roof)
+        for(var/obj/structure/roof/roof in connected_roof)
+            qdel(roof)
+    return ..()
```
Drulikar, line 124: You've replaced tab indentation with space indentation here.

code/game/objects/structures/roof.dm
```diff
@@ -154,10 +161,27 @@
 	remove_under_roof(living)
 
 /datum/roof_master_node/proc/connect(location)
+	var/list/unprocesed_nodes = list()
 	for(var/obj/effect/roof_node/node in location)
-		node.link_master(src)
+		unprocesed_nodes += node
```
Drulikar, line 166: Would be better to just figure out all the work that needs to be performed, and have that in a list. Likely better to add using |= so you don't add duplicates; but you may want to test what performs better w/ large use cases: as in is byond faster at checking for duplicates than just enqueueing them anyways and finding out we don't need to work on them when you come across a node already set up.

code/game/objects/structures/roof.dm
```diff
@@ -154,10 +161,27 @@
...
+
+
+	while(unprocesed_nodes.len)
```
Drulikar, line 169: use length() over .len where possible

code/game/objects/structures/roof.dm
```diff
@@ -154,10 +161,27 @@
...
+		if(new_nodes)
+			unprocesed_nodes += new_nodes
+
+	var/list/unprocesed_roofs = list()
+
 	for(var/obj/structure/roof/roof in location)
-		roof.link_master(src)
+		unprocesed_roofs += roof
+
+	while(unprocesed_roofs.len)
```
Drulikar, line 180: use length() over .len where possible

code/game/objects/structures/roof.dm
```diff
@@ -154,10 +161,27 @@
...
+		var/obj/effect/roof_node/node = popleft(unprocesed_nodes)
+		var/list/new_nodes = node.link_master(src)
+		if(new_nodes)
+			unprocesed_nodes += new_nodes
+
+	var/list/unprocesed_roofs = list()
+
 	for(var/obj/structure/roof/roof in location)
-		roof.link_master(src)
+		unprocesed_roofs += roof
```
Drulikar, line 178: Would be better to just figure out all the work that needs to be performed, and have that in a list. Likely better to add using |= so you don't add duplicates; but you may want to test what performs better w/ large use cases: as in is byond faster at checking for duplicates than just enqueueing them anyways and finding out we don't need to work on them when you come across a node already set up.

code/game/objects/structures/roof.dm
```diff
@@ -96,9 +100,12 @@
 		return
 	master.connected_nodes += src
 	linked_master = master
+	var/list/nerby_nodes = list()
 	for(var/direction in CARDINAL_ALL_DIRS)
 		for(var/obj/effect/roof_node/node in get_step(src,direction))
-			node.link_master(master)
+			if(!node.linked_master)
+				nerby_nodes += node
+	return nerby_nodes
```
Drulikar, line 108: This proc is likely moot to exist with suggested changes to connect.

code/game/objects/structures/roof.dm
```diff
@@ -49,12 +49,13 @@
...
-	for(var/icon in GLOB.player_list)
-		var/mob/mob = icon
-		mob.client.images -= normal_image
-	return ..()
+    if(linked_master)
+        linked_master.remove_roof(src)
+    linked_master = null;
+    for(var/icon in GLOB.player_list)
+        var/mob/mob = icon
+        mob.client.images -= normal_image
```
Drulikar, line 57:
```suggestion
    for(var/mob/mob as anything in GLOB.player_list)
        mob.client?.images -= normal_image
```

code/game/objects/structures/roof.dm
Drulikar, line 1: Please have node setup delayed until after round start by:
- In Initialize, call parent `..()` and then return `INITIALIZE_HINT_ROUNDSTART`
- Move set up of nodes from Initialize to LateInitialize

The goal of this is so that nightmare inserts can change roofs around and *then* roofs get set up.

Drulikar (changes requested):

code/game/objects/structures/roof.dm
cuberound: okey I have replaced lateinit hint with roundstart one, I knew it needs to happen after it all spawns but did not realize that nightmare inserts exist

code/game/objects/structures/roof.dm
```diff
@@ -34,7 +34,7 @@
 			neighbor = locate() in adjacent_loc
 			if(!neighbor)
 				neighbor = new(adjacent_loc)
-	return INITIALIZE_HINT_LATELOAD
+	return INITIALIZE_HINT_ROUNDSTART
```
Drulikar, line 37: I don't imagine lazy_nodes will work correctly if a nightmare insert modifies a roof. You really ought to figure out how to get them to load after roundstart.

If the issue is that master lateloads before the nodes lateload, have the nodes when they're done loading call a proc to the master node that increments a ready count in the master node.

If the master gets signaled a node is ready, and the count of ready nodes (calls to this proc) now matches the count of linked nodes, then master initializes. You just need to ensure that the count of linked nodes will never be incorrect.

Drulikar (changes requested):

cm13-github:

cm13-github:

Drulikar:
I just don't think I'm happy with the implementation still. But also for example 
meh
Two masters created, but one master took all the nodes. In this situation I think we probably could:
1. Don't create nodes until INITIALIZE_HINT_ROUNDSTART occurs for lateload
2. Once roofs late load, they find all adjacent roofs, and that group makes a master (maybe keep the !lazy_nodes separate?)
3. Perform INVOKE_NEXT_TICK on that master (so all the roofs can get linked) to:
4. For master's roofs that are lazy_nodes: for every open turf adjacent that doesn't have a roof make a node linked to that master

(I don't recall the usecase for setting !lazy_nodes though)

(PS: Yes this would intentionally allow nodes to overlap - if that would need to be avoided then you need the capability for a node to have multiple masters but those being all potentially lists just seems like a waste of memory)

I may fiddle with this more in the future, but I don't know if I have the energy at the moment to tackle this.

hry-gh:
closed as stale, please recreate this pr if you want to resume working on it

PR #7948, opened 2024-12-30
Reduces re-evolve cooldown to 4 minutes 

MeH0y:
# About the pull request

As title, the cooldown to re-evolve into the same caste is now four minutes down from five.

The addition of this mechanic has had consequences outside of preventing the niche abuse it is targeting. 
The burrower tunnel cooldown is 4 minutes, if you want to de-evolve and build as a drone you are now adding a minute to the cooldown for trying to be more useful, I don't believe this should be the case. 
Prevents a niche scenario that has occurred when willing drone participants need to re-evolve into drone in order to fill Queen slot but are unable to because of the five minute timer.

I tested it I promise

# Changelog

balance: Xenomorph re-evolve cooldown reduced to four minutes from five.
/:cl:

### DIFF
--- a/code/datums/components/deevolve_cooldown.dm
+++ b/code/datums/components/deevolve_cooldown.dm
@@ -1,4 +1,4 @@
-#define XENO_DEEVOLVE_COOLDOWN 5 MINUTES
+#define XENO_DEEVOLVE_COOLDOWN 4 MINUTES
 
 /**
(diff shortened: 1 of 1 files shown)


Drulikar:
This goes against the point of the cooldown. It's to discourage doing stuff like you're trying to do.

PR #7967, opened 2024-12-31
New xenos special structutre "xenomorpher"

Lengord:
# About the pull request

Add one more special structure at the hive where xenos can sacrifice bodies of bursted marines (and alive xenos for fun or roleplay, or if it is bad can remove it), and increase charges of structure, then at the wish, player can use structure to buy upgrades from the list for his character. 

Upgrades done: increase health, armor, plasma, damage, standing regeneration, resting regeneration, plasma regeneration, pheromone power, ability to heall off weed 

More possible upgrades to add  (if i understand how to them): buy larva, buy evo points, acid claws, speed (may be too op to add), flame immunity and ignite immunity (one denies the other to avoid immortal mob)

hugger, lesser, larva and queen can't use it
huggers, lessers can't be used to get charges
upgardes are capped to avoid ovrpowered xenos

P.S. I know code sucks, but it's the best i can do for now, belive me it took me lot of time to make even this. I want to see if this idea will be good at all and if it worth to improve.
Adds more flexibility to gameplay as a Xeno. It is intended to work like Xeno Req and shouldn't be OP because it is tied to the number of burst bodies, which are limited at the start and by the number of captured marines. In the end, the more marines that are lost as captures, the more Xenos.

<details>

</details>

# Changelog
:cl:
add: Added new xenomorph special structure that allows xeno players to upgrade stats
code: Added code for new structure and change connected code.  

/:cl:

### DIFF
--- a/code/__DEFINES/xeno.dm
+++ b/code/__DEFINES/xeno.dm
@@ -639,6 +639,7 @@
 #define XENO_STRUCTURE_EGGMORPH  "egg morpher"
 #define XENO_STRUCTURE_RECOVERY  "recovery node"
 #define XENO_STRUCTURE_NEST  "thick resin nest"
+#define XENO_STRUCTURE_XENOMORPHER "xenomorpher"
 
 #define RESIN_TRAP_EMPTY 0
 #define RESIN_TRAP_HUGGER 1
--- a/code/datums/construction/xenomorph/construction_template_xenomorph.dm
+++ b/code/datums/construction/xenomorph/construction_template_xenomorph.dm
@@ -59,6 +59,12 @@
 	build_type = /obj/effect/alien/resin/special/eggmorph
 	build_icon_state = "eggmorph_preview"
 
+/datum/construction_template/xenomorph/xenomorpher
+	name = XENO_STRUCTURE_XENOMORPHER
+	description = "A xenomorph structure that allows them to change their nature in exchange for biomass."
+	build_type = /obj/effect/alien/resin/special/xenomorpher
+	build_icon_state = "pool_preview"
+
 /datum/construction_template/xenomorph/recovery
 	name = XENO_STRUCTURE_RECOVERY
 	description = "Hastily recovers the strength of sisters resting around it."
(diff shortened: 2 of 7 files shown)


Drulikar:
I don't know what from this is really salvageable. Maybe explore better end game hive buffs instead. Yes bursts have no use, but maybe keep it that way. Shoving living xenos in the pool to damage them is silly, but needs to not be a thing too. Fire immunity and ignite immunity absolutely need to not be a thing for an individual indefinitely; *maybe* as a hive buff but only as a temporary thing like the end game buffs are.

If we really wanted to go w/ something like this, it would need to be in a fashion of risk/reward. As in xenos that kill or assist a kill get a small accumulating health bonus to a point. But I'm not sure we need something like that.

Red-byte3D:
This got closed while i was writing this but i wanted to post this anyway incase this ever gets re-opened in the future here are some of my opinions on it.

Some maps start out with /significantly/ more burst bodies then others, that will make xenomorphs way stronger on the first drop already, we're trying to lower how fast xenomorphs stomp after the first drop anyway, so i dont think this is a good idea unless theres a time limit, or the structure /only/ accepts marine burst bodies, or survivor ones.

As much as i dislike it, we're against personal buffs as people will only give it to "good players" Yes, i think you cant really avoid this but this is a really big issue that comes with this.

The name doesnt work

Text needs to be "We" instead of "You"

The current queen "give evolution (sacrificing a larva for 250 evo)" is kind of useless, i could see actual bodies being used to store a currency that the queen can hand out evo with on a big cooldown instead of using larva, as in most cases iit never gets used because usually a larva is more important then evo and or theres just literally never burrowed larva. This could be useful in that front but thats about it.

RainbowStalin:
I could see induvidual buffs if you could not stack them. Makes it so you want to spread them out and find the best buff for each xeno, giving a managerial field for the queen while ovi. What I never see happening is that it takes any bursted body. Because either we would need to remove most bursted bodies from maps (no), make map bursts give less points to an upgrade (still removes why they are there tbh), or make it marine bursts only.

And then comes the final solution. Make it royal jelly based and it effectivly becomes a xeno lategame that might be viable (in my opinion). It can work, but would need a solid resource to go off of.

PR #7975, opened 2025-01-02
Predators can no longer set off OT claymores 

Joelampost:
# About the pull request
Predators can no longer trigger OT claymores or prox sensors.

Removing OT claymores or just simply removing OT from the game entirely would also work but that is not up to me.

Dying instantly as a one life role is not fun. 

Zero counter play is not fun.
<details>
I tested both for xenos and any other mob without the marine IFF tag and it set off so it should work

</details>

# Changelog
:cl: Joe Lampost
balance: Predators can no longer trigger OT claymores or prox sensors 
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/mine.dm
+++ b/code/game/objects/items/explosives/mine.dm
@@ -326,6 +326,11 @@
 	matter = list("metal" = 3750)
 	has_blast_wave_dampener = TRUE
 
+/obj/item/explosive/mine/custom/try_to_prime(mob/living/carbon/hunter)
+	if(isyautja(hunter))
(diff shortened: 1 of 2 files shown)


thevandie:
This is a cope pr

Joelampost:
> This is a cope pr

yes and

Joelampost:
> This is a cope pr
> 
> The counterplay is _look at the ground you're sprinting over_ Survivors deal with this all the time, just get

Drulikar:
As a predator you have the most opportunity to observe for traps. Making them immune is not the answer.

PR #7996, opened 2025-01-03
adds a search to the hair/facial hair pickers, colorizes on 516

harryob:
:cl:
ui: adds a search bar for hair/facial hair picking, also sorts them alphabetically
ui: if you're on 516, the hair/facial hair picker now previews hair color too
/:cl:

https://github.com/user-attachments/assets/a05b8abb-f3c7-49ad-8cdb-a9165c33bc49

hair color preview
EPbY73zJHsSK4Zr6@2x


DexterDude:
thanks pookie

Drulikar (changes requested):
Do you know a way to mitigate the selector somereason growing when the elements are limited?
image

tgui/packages/tgui/interfaces/HairPicker.tsx
```diff
@@ -208,26 +215,33 @@ export const HairPickerElement = (props: {
...
+          .filter(
+            (val) =>
+              search.length === 0 || val.name.toLowerCase().includes(search),
+          )
+          .sort((a, b) => (a.name > b.name ? 1 : -1))
+          .map((hair) => (
+            <Stack.Item
+              key={hair.name}
+              className={`Picker${active === hair.icon ? ' Active' : ''}`}
+              height="fit-content"
```
Drulikar, line 228:
```suggestion
              maxHeight="66px"
```
This doesn't seem to make any effect. However you could do instead above. But its not exactly perfect: 
image
image
Eliminating the black space just makes it so there's no separate in the normal view e.g. 64px:
image
image

Drulikar (changes requested):

Drulikar (approved):

PR #8003, opened 2025-01-04
allows predators to make plasma breach charges + makes them actually stun and makes them better

Lagomorphica:
# About the pull request

allows predators to make plasma breaching charges from their bracer on a 4 minute cooldown with an energy cost of 1000, and buffs the charges to actually stun as is intended + gives them more shrapnel (20 instead of 10) and uses a different icon to communicate that it is actually a plasma wave

the icon for the plasma breach charge is the old claim equipment icon from beaglegaming1s pr

better options for removing environmental obstacles + better shipside gear recovery, will need to be added to hc if it is supposed to be dishonourable etc etc

image
tested on local and everything seemed to work fine, unless i'm an idiot

<details>

n/a

</details>

# Changelog

:cl:
balance: predators can produce plasma breaching charges from their bracer
balance: plasma breaching charges now actually stun when the shrapnel hits and they detonate in 3 seconds rather than 5, alongside having additional shrapnel - this does not impact xenos
/:cl:

### DIFF
--- a/code/__DEFINES/keybinding.dm
+++ b/code/__DEFINES/keybinding.dm
@@ -186,6 +186,7 @@
 #define COMSIG_KB_YAUTJA_ACTIVATE_SUICIDE "keybinding_yautja_activate_suicide"
 #define COMSIG_KB_YAUTJA_INJECTORS "keybinding_yautja_injectors"
 #define COMSIG_KB_YAUTJA_CAPSULE "keybinding_yautja_capsule"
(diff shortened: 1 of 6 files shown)


Zonespace27:
Thank you for the PR, but the maintainer team agrees that this is not necessary at the moment. Predators have quite a few tools for breaching as it stands, and we don't see the need for another one that can also be pulled out of thin air.

PR #8029, opened 2025-01-07
Bug where we forget to unregister signal

blackcrystall:
# About the pull request
Bug, we don't remove signal in right moment, so just changed 4 lines of code for both CO guns

# Changelog
no playeside facing changes


kiVts:
average player has no idea what you're trying to say in the changelog both due to poor English skills and too much depth on technical part, rewrite it.

harryob (approved):

PR #8058, opened 2025-01-09
nerfs combi stick block chance from 66% to 33%

irakligabunia:
# About the pull request

its a nerf to the griefer wl that exists to hunt pvts and young xenos because they're on a powertrip
In all seriousness, it nerfs the combi-stick weapon that Yautja use. Specifically it nerfs the block chance of it from 66% to 33%

I already buffed the durability of predators, I don't think they need this anymore. Right now it's just rolling a dice everytime you play against a pred, especially if you're a xenomorph. I'd prefer if this was testmerged first though.

god told me this would work

# Changelog

:cl: Pewdiepie
balance: nerfs combi stick block chance from 66% to 33%
/:cl:

### DIFF
--- a/.github/workflows/ci_suite.yml
+++ b/.github/workflows/ci_suite.yml
@@ -67,6 +67,9 @@ jobs:
     name: Lint with OpenDream
     runs-on: ubuntu-22.04
     steps:
+      - uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: 9
       - uses: actions/checkout@v4
       - uses: robinraju/release-downloader@v1.9
         with:
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -97,7 +97,7 @@ Contains most of the procs that are called when a mob is attacked by something
 	var/block_effect = /obj/effect/block
 	var/owner_turf = get_turf(src)
 	if(l_hand && istype(l_hand, /obj/item/weapon))//Current base is the prob(50-d/3)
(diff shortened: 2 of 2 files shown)


Strobia:
togif

RenaRenaRe:
So I've mentioned this before and everyone ignored it so maybe I'm crazy and missing something but I'm pretty sure this only affects the block chance for pounces. These numbers are only used when "combistick" is true. You can search to find all instances where check_shields() is called and the only time where the combistick parameter is set is for pounces. Otherwise "combistick" is false so it does not use these numbers. The numbers that are actually used is the 50 - damage/3 except the "damage" parameter is sent as 0 for xeno slashes regardless of the actual damage so for xeno slashes the block chance is just 50%.

You can test this by setting the values you changed to 0 and observing that combistick still blocks. Then change the 50-damage/3 to 100 and observe that combistick always blocks.

harryob:
> So I've mentioned this before and everyone ignored it so maybe I'm crazy and missing something but I'm pretty sure this only affects the block chance for pounces. These numbers are only used when "combistick" is true. You can search to find all instances where check_shields() is called and the only time where the combistick parameter is set is for pounces. Otherwise "combistick" is false so it does not use these numbers. The numbers that are actually used is the 50 - damage/3 except the "damage" parameter is sent as 0 for xeno slashes regardless of the actual damage so for xeno slashes the block chance is just 50%.
> 
> You can test this by setting the values you changed to 0 and observing that combistick still blocks. Then change the 50-damage/3 to 100 and observe that combistick always blocks.

you're correct, this just nerfs it blocking pounces. fix it

PR #8080, opened 2025-01-11
Tackle Rework

Git-Nivrak:
# About the pull request
Reworks tackle to be more teamplay based than solo
Tackles from multiple xenos on the same person will now stack rather than being separate counters, Xenomorph tackle min and max increased across the board as well as lowering tackle strengths (duration of the tackle stun), Some problematic xenos got harsher treatment (Runner, Drone, Carrier, Hivelord)

I think everyone can agree the current tackle system is bad and cheesy, You can call it salt but no one enjoys being dragged forever by a drone or a runner because you have been clicked X amount of times
In an attempt to change the focus of tackle from solo to team based I think the system would be much more rewarding when xenos work together and much less frustrating to play against overall.
Values can be adjusted accordingly if xeno capping increases or decreases significantly as that is a non-goal of this pr (whatever was previously being capped solo should turn into team caps)
<details>

</details>

# Changelog
:cl:
balance: Tackles from multiple xenos now stack rather than having their own separate counters
balance: Increased min and max amount of attempts xenos need to tackle across the board
balance: Decreased duration of tackle stun across the board
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/Runner.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Runner.dm
@@ -19,11 +19,11 @@
 	evolves_to = list(XENO_CASTE_LURKER)
 	deevolves_to = list(XENO_CASTE_LARVA)
 
-	tackle_min = 4
-	tackle_max = 5
+	tackle_min = 8
+	tackle_max = 10
 	tackle_chance = 40
-	tacklestrength_min = 4
-	tacklestrength_max = 4
+	tacklestrength_min = 2
+	tacklestrength_max = 2
 
 	heal_resting = 1.75
 
--- a/code/datums/components/tackle_counter.dm
+++ b/code/datums/components/tackle_counter.dm
@@ -1,0 +1,31 @@
+/datum/component/tackle_counter
+	var/tackles = 0
+	var/reset_timer
+
+/datum/component/tackle_counter/Initialize()
+	if(!isliving(parent))
+		return COMPONENT_INCOMPATIBLE
+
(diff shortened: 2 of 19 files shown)


Red-byte3D:
The entire point of runner is its a guaranteed 4 tackle capture if you dont respect it, same as drone, thisi s literally killing both runner as a caste and any change of self defense as a drone

code/modules/mob/living/carbon/xenomorph/castes/Runner.dm
```diff
@@ -19,11 +19,11 @@
 	evolves_to = list(XENO_CASTE_LURKER)
 	deevolves_to = list(XENO_CASTE_LARVA)
 
-	tackle_min = 4
-	tackle_max = 5
+	tackle_min = 8
+	tackle_max = 10
```
ihatethisengine, line 23: Might as well just delete the caste lol. Cannot cap nor even transport caps.

If you want to make this work you need to do a lot more than just nerfing solotackle. You need proper telegraphing, you need to fix leaders accidentally tackling other xenos, etc etc. And for what really? Getting solocapped is a skill issue most of the times.

private-tristan:
Sentinel's entire gimmick around paraslash ability is to capture the marine once they're down in a 1v1.

Red-byte3D:
A more detailed comemnt.
As we do need to mention alot of things xenos are strugglingt wtih
1.) Capping is barely worth it already, most people just get shoved by a leader who has 0.5 hours on the game because he got leader due to being tier 3 or dir click etc, this is going to basically make it impossible to capture marines
2.) The only things that really need to change from the xeno side is one, warrior lunge being able to shove people. making it so they can only shove while having something lunged is a way better change ot counteract that
3.) sentinel paralyzing slash needs a looking at

Git-Nivrak:
After some consideration of the current state of the game I will closed this pr for now

Will reopen later when the game is in a better spot for it

PR #8082, opened 2025-01-11
Breaching nerfs

Git-Nivrak:
# About the pull request
Nerfs XM51 breaching power

The map have become too destructible, get two or three guys with XM51 and the entire frontline becomes an open field while previously you had to use precious C4, Breaching charges or even deconstruct which takes a *lot* of time, to this end the damage to rwalls for XM51 has been reduced significantly, increasing the amounts of shots you need from 4 to 7 and the amount of breacing shells you can carry has been reduced (by nerfing the ammo box max capacity)
<details>

</details>

# Changelog
:cl:
balance: Nerfed XM51 bonus damage against reinforced walls.
balance: Reduced amount of breaching shells its ammo box can carry from 120 to 60.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -2060,9 +2060,11 @@
 
 /obj/item/weapon/gun/rifle/xm51/set_bullet_traits()
 	LAZYADD(traits_to_give, list(
-		BULLET_TRAIT_ENTRY_ID("turfs", /datum/element/bullet_trait_damage_boost, 30, GLOB.damage_boost_turfs), //2550, 2 taps colony walls, 4 taps reinforced walls
+		BULLET_TRAIT_ENTRY_ID("turfs", /datum/element/bullet_trait_damage_boost, 30, GLOB.damage_boost_turfs), //2550, 2 taps colony walls
+		BULLET_TRAIT_ENTRY_ID("rwalls", /datum/element/bullet_trait_damage_boost, 0.4, GLOB.daamge_boost_rwall), // 1020, 9 taps for rwall
 		BULLET_TRAIT_ENTRY_ID("xeno turfs", /datum/element/bullet_trait_damage_boost, 0.23, GLOB.damage_boost_turfs_xeno), //2550*0.23 = 586, 2 taps resin walls, 3 taps thick resin
 		BULLET_TRAIT_ENTRY_ID("breaching", /datum/element/bullet_trait_damage_boost, 15, GLOB.damage_boost_breaching), //1275, enough to 1 tap airlocks
+		BULLET_TRAIT_ENTRY_ID("xeno_door", /datum/element/bullet_trait_damage_boost, 0.5, GLOB.damage_boost_doors_xeno), // 637.5 one tap normal resin door 2 tap thick resin door
 		BULLET_TRAIT_ENTRY_ID("pylons", /datum/element/bullet_trait_damage_boost, 6, GLOB.damage_boost_pylons) //510, 4 shots to take out a pylon
 	))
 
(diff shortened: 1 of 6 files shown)


Staykeu:
This is stupid. You're on crack.

squidqueen:
I feel a better solution is making the XM51 more scarce. As it is, the gun runs out of ammo extremely fast into a round and then getting a resupply can be nearly impossible. Reduce overall saturation of xm51 and it should handle the breaching issue overall. 

Being able to remove resin doors with UBS is less of an issue than marines being able to flame under resin doors, which invalidates them as a defense because the fire immediately breaks the door

DrAmazing343:
Letting USB break doors is great for the breacher style and a nerf to it basically invalidates that since the more time spent facing a door the more time xenos have to stack up and turbo juggle you to hell

Diegoflores31:
> I feel a better solution is making the XM51 more scarce. As it is, the gun runs out of ammo extremely fast into a round and then getting a resupply can be nearly impossible. Reduce overall saturation of xm51 and it should handle the breaching issue overall.

i think there is plenty of ammo available on normal play. there is an excess of XM51 in req and you also get its ammo literally for free on ASRS drops so . idk

simonset55:
Good change, xenos need their walls and doors to fight since being in the open is the marine's playground.

Breaching stuff that was added quite a while ago just invalidates that and is the reason for xeno wipes most of the time, especially on lowpop because they never get time to rest behind defenses. Inevitably they're pushed back to a wall at the hive and then they're all dead.

Could also just give xenos an ability that instantly destroys cades to balance it out, that'd be sooo funny.

ihatethisengine:
People build their loadouts around carrying an XM51, it's not like you can fit it in a pocket. Kind of split on this one. I see where it's coming from, but I feel like xenos already have enough invincible walls already (caves) and also nothing stops you from building new walls as a xeno. Plus for example on Solaris LZ1 is made of Rwalls, and you need to deconstruct them every round to make FOB work, which would be a lot more tedious if this PR gets merged.

As for the masterkey nerf, I think this should be in a separate PR and I don't think the nerf is necessary. You either need to nerf other options that can quickly deal with the doors as well (m39 with stock and bayo, machete, scout gun, axe), or don't touch the attachment one of the main selling points of which is oneshotting doors. There are downsides in taking UBS over other attachments such as agrip, so I think the trade-off of getting a situational advantage is fair enough. Also we don't have any telegraphs for damaged doors as far as I am aware, so this will make things a lot more confusing for everyone involved.

thwompa:
Splice in the changes to explosive availability and this will go a lot better. We can [make a new explosive](https://en.wikipedia.org/wiki/Detonating_cord) if you want to go further here. I don't think flat nerfs will fly on their own.

Red-byte3D:
good change, nobody builds or plays builder becasue either building sucks or peole just instantly destroy it from 10 years away with one shot, good nerf dont listen to literally anything above

code/modules/projectiles/guns/rifles.dm
```diff
@@ -2060,7 +2060,8 @@
 
 /obj/item/weapon/gun/rifle/xm51/set_bullet_traits()
 	LAZYADD(traits_to_give, list(
-		BULLET_TRAIT_ENTRY_ID("turfs", /datum/element/bullet_trait_damage_boost, 30, GLOB.damage_boost_turfs), //2550, 2 taps colony walls, 4 taps reinforced walls
+		BULLET_TRAIT_ENTRY_ID("turfs", /datum/element/bullet_trait_damage_boost, 30, GLOB.damage_boost_turfs), //2550, 2 taps colony walls
+		BULLET_TRAIT_ENTRY_ID("rwalls", /datum/element/bullet_trait_damage_boost, 0.51, GLOB.daamge_boost_rwall), // 1147, 7 taps for rwall
```
Red-byte3D, line 2064: 7 taps for a rwall might be too much but otherwise looks gtm

PR #8095, opened 2025-01-13
516 byondstorage can fail to appear for a bit for some reason

harryob:
so we just wait for it to appear instead

:cl:
fix: on 516, sometimes the storage backend would not find your preferences correctly. this is fixed, but you still need to be on version 516.1650+
/:cl:


Drulikar (approved):
As far as I can tell this works; but 516 is unplayable.

PR #8109, opened 2025-01-14
M46C NERF (makes it mk1 equal)

Red-byte3D:
# About the pull request

Makes the CO rifle an equivelant of the MK1, removes its ability to remove the ID lock and gives it full auto while the iff is on.

Requested by the co senator

<details>

</details>

# Changelog
:cl:
balance: Statistically the m46c is now equal to the mk1 and should be kept this way.
balance: The co can no longer remove the ID lock from his rifle.
balance: The m46c can now use full auto while IFF is enabled
/:cl:


So-mii:
I asked for him to do this :D

thwompa:
I was getting ready to ask Drathek to do it without a PR and see if anybody notices LOL good change

code/modules/projectiles/guns/rifles.dm
```diff
@@ -648,8 +609,7 @@
 	. = ..()
 	if(iff_enabled)
 		modify_fire_delay(FIRE_DELAY_TIER_12)
```
Drulikar, line 611: If the firemodes now don't change when toggling `iff_enabled` don't bother setting them here. Just make the gun defined with `start_automatic`

In other words the only thing you want here is `modify_fire_delay(FIRE_DELAY_TIER_12)`

Drulikar (changes requested):
Otherwise looks fine.

TheGamerdk:
Probably want to remove the gun_firemode = GUN_FIREMODE_SEMIAUTO on line 636

Also probably want to remove the m46c/stripped

Also could possibly make it a subtype of the MK1 if you want to enforce the
balance: Statistically the m46c is now equal to the mk1 and should be kept this way.

Red-byte3D:
subtypes would be too much of work if people wanted to revert back, this is just easier to uphold and the comment was a statement for any person re-adjusting the mk1

TheGamerdk:
> subtypes would be too much of work if people wanted to revert back, this is just easier to uphold and the comment was a statement for any person re-adjusting the mk1

it's actually only a ctrl-f replace /obj/item/weapon/gun/rifle/m46c to /obj/item/weapon/gun/rifle/m41aMK1/m46c
and then starting_attachment_types = list() and remove set_gun_config_values()

but fair enough, makes sense either way.

Drulikar (approved):

ihatethisengine:
Maybe the gun should use alt IFF instead of full IFF? The gun might still be a bit problematic as a basically better SG.

code/modules/projectiles/guns/rifles.dm
```diff
@@ -633,7 +595,6 @@
 		to_chat(usr, SPAN_WARNING("[icon2html(src, usr)] Action denied by [src]. Unauthorized user."))
 		return
 
-	gun_firemode = GUN_FIREMODE_SEMIAUTO
 	iff_enabled = !iff_enabled
```
ihatethisengine, line 598: in case you are interested, it's a really easy change
```suggestion
	iff_enabled = !iff_enabled
	SEND_SIGNAL(src, COMSIG_GUN_ALT_IFF_TOGGLED, iff_enabled)
```
and you'll need to put this in Initialize() of the gun
`AddComponent(/datum/component/iff_fire_prevention, FIRE_DELAY_TIER_5)`

(FIRE_DELAY_TIER_5 is the additional delay you get when trying to shoot over a friendly, I am not actually sure if it's working or needed, you can just remove the arg)

Red-byte3D:
senator was against that, if its not fine it iwll be revisited

PR #8131, opened 2025-01-17
Prohibits var plane from being edited at the mapping

blackcrystall:
# About the pull request
Simply as...
Added it as banned variables for all assets on maps

Why do we need this? Because it's very hard to track and keep up to date with defines in game for planes.

# Changelog
No player facing changes


Nanu308 (approved):
mapping seems fine

harryob (approved):

PR #8154, opened 2025-01-19
Research: Drawbacks/Nerf to MST and NST

private-tristan:
# About the pull request
Muscle Stimulating (MST) Now reduces your nutrition its' potency times hunger factor (which stacks ontop of hunger factor). meaning that you're getting an extra 0.05 nutrition loss per tick for each level of MST.

Nerve-Stimulating (NST) now reduces your **SPEED** by .1 for every level it has. For reference, a splinted limb gives you .5 slowdown.

NST now only reduces STUN by .5 per level, instead of .5 per level AND. 25 per level, meaning that it now only provides .5 seconds of stun resistance per level, instead of .75
**Note that STUN is an immobilize** that locks you in place (such as being between 5-7 tiles away when a queen screeches), **not to be confused with WEAKEN** which includes queen screech (within 4 tiles) and being tackled.
The intent is not to make MST/NST too weak, but rather to give them at least a minor drawback for marines using them, so that godstims are marginally less of a stomp (And also nerf NST like wow that thing is OP).

Creates a variant of NST called antistun, which is a clone of NST except it doesn't reduce your movespeed. This property is **unobtainable** and is reserved exclusively for neuraline (CL injector) and MARSOC (Deathsquad Stims)
Also gives speed stim (Marsoc stim injector) a single level of nutritious, to prevent marsoc from immediately starving upon using their speed stim.

MST and NST are extremely powerful in their current state, allowing marines to rush down xenos extremely easily while being extremely hard to kill. I'd like them to at least have a minor downside for marines, MST's nutrient drain can be worked around but forces researchers to either spend points on the nutritious property or for marines to keep track of their hunger more often.
<details>

Here's a quick video showing queen screech (7 stun/weaken) against no NST/5NST/10NST. With level 5 NST, it takes 2 ticks for the stun to wear off, while with level 10 (or rather, level 6/7 NST) it takes only 1 tick.

https://github.com/user-attachments/assets/92c3011c-8299-490f-8960-784b78819c9c

</details>

# Changelog
:cl:
balance: Nerve Stimulating (NST) property now makes you slower by .05 per level (for reference, a splinted limb is .5 slowdown). This does not affect the CL's Neuraline or MARSOC stims.
balance: NST now reduces immobilizes (not roots) by .5 per level, instead of .75
balance: Muscle Stimulating (MST) property now makes you get hungry faster, depending on potency. 
balance: MARSOC speed stims and ultrazine now have nutritious to counteract the hunger loss 
/:cl:


cm13-github:

cm13-github:

PR #8168, opened 2025-01-19
Fix of auto shove for dancer/lying&dead zenos

### LINKED ISSUE #7362: Queen antibodyblock autotackle does not take dancer dodge into account
### Testmerges

#6808, #7242, #6334, #7330, #6248, #7217

### Round ID

24470

### Description of the bug

Dancer can pass through other xenos using his dodge ability. But if it tries dodging through the queen he keeps getting autotackled.

Dancer on dodge shouldn't be getting autotackled because it's not bodyblocking the queen. And it being able to run through her is actually a key feature of dancer. This bug breaks it.

This is possibly unfixable because it may require queen autotackle logic to know about the dancer dodge, but still needs to be looked into.

1. Get a dancer
2. Press dodge
3. Run into a queen
...

### Issue Bingo

blackcrystall:
# About the pull request
We simply solve issue, nothing more... we don't shove dead zenomorphs and lying down zenomorphs, and dancer in his "DODGE" effect

Less bugs and "distracting" things.

# Changelog

:cl: BlackCrystalic
fix: Auto shove ignore dead/lying xenomorphs
fix: Dancer can now safely play around queen/king with auto shove enabled
/:cl:

Closes #7362

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/XenoProcs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/XenoProcs.dm
@@ -764,3 +764,9 @@
 	SetRoot(0)
 	SetEyeBlur(0)
 	updatehealth()
+
+/mob/living/carbon/xenomorph/BlockedPassDirs(atom/movable/mover, target_dir)
+	var/mob/living/carbon/xenomorph/xeno = mover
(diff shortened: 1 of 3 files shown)


cm13-github:

PR #8169, opened 2025-01-19
Boiler neuro purges stim, buffs neuro tailstab duration, reduce dizziness 

private-tristan:
# About the pull request
Boiler neurotoxin now purges 2u of stim per tick of the effect
Boiler Tailstab now adds 20 duration*** of neuro (from 16).

Decreases the amount of dizziness inflicted per tick by neuro from 12 to 8

***Note that neurotailstab won't inflict blindness despite adding 20 duration, as it ticks down once(?) before taking effect (Plus, an instant 5 second blind on toxic tailstab would be crazy).

Xenos should have more way to purge stims (especially godstims) that don't include being the weakest T1 in the game.

As for the neuro tailstab buff, I think that it's very underused and underpowered compared to the 90(!) DOT from acidic tailstab.

The dizziness nerf is because it's such an annoying  effect, it lingers for a LONG time after the neuro ends, I'd rather it ended a bit quicker.

made a custom stim with no effects, injected 100u, boiler tailstab, watch stims get drained.
works.

# Changelog
:cl:
add: Boiler neurotoxin now quickly purges research stimulants.
balance: Boiler's neurotoxin tailstab now inflicts 4 more neurotoxin duration (16 to 20). Acid Tailstab is unaffected.
qol: Duration of dizziness caused by neurotoxin has been reduced.
/:cl:

### DIFF
--- a/code/datums/effects/neurotoxin.dm
+++ b/code/datums/effects/neurotoxin.dm
@@ -20,6 +20,8 @@
 	var/chat_cd = 0
 	/// Stamina damage per tick. Major balance number.
 	var/stam_dam = 7
+	/// Stimulant drain per tick.
+	var/stim_drain = 2
 
 /datum/effects/neurotoxin/New(atom/thing, mob/from = null)
 	..(thing, from, effect_name)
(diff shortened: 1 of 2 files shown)


Red-byte3D:
i was going to do it originally but you should make the queen NEURO SPIT remove stims too

harryob (approved):

PR #8174, opened 2025-01-20
forensic spray can no longer generate randomly

iloveloopers:
# About the pull request
Forensic spray can no longer generate in recipes, mutated plants and colony vials.

Doesn't make sense for this to show up in plants and just makes recipes needlessly harder.

Probably works

# Changelog

:cl:
qol: Forensic spray can no longer generate in recipes, mutated plants and colony vials.
/:cl:


harryob (approved):
why is chemcode so shit why do you need 200 bitflags to do one thing

PR #8175, opened 2025-01-20
Research chems now spawn with higher ODs

private-tristan:
# About the pull request
Taken directly from #7023 

Research chems (those found in vials or in botany) now generate with significantly higher ODs (Minimum 30u), to encourage amping/relating them to make them useful instead of just ignoring them. 

Currently chem papers are extremely hit and miss, while you may get papers with good properties, and chems that would "fit in" with the paper, the combo of them would result in an OD so low that it would be effectively worthless. With a higher OD, researchers can play around with chems more, relating/amping without making it completely useless.
spawned in as researcher
spammed Spawn "fancy/vials/random
scanned vials
ODs minimum 30u, typically 40-45u, with 1 50u.

# Changelog
:cl: KiVts, Private-Tristan
balance: Research chems now spawn with a significantly higher overdose threshold
/:cl:

### DIFF
--- a/code/modules/reagents/Chemistry-Generator.dm
+++ b/code/modules/reagents/Chemistry-Generator.dm
@@ -200,9 +200,8 @@
 
 	//OD ratios
 	overdose = 5
-	for(var/i=1;i<=rand(1,11);i++) //We add 5 units to the overdose per cycle, min 5u, max 60u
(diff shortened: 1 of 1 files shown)


private-tristan:
seems like the #7023 is on TM for now, closing.

PR #8248, opened 2025-01-26
Fix predator limit

Drulikar:
# About the pull request

This PR fixes an oversight where predator limit was not working because for whatever reason predators are the only whitelist where `get_whitelist_status` doesn't return the defines in mode.dm and instead it uses the ranks from clans.dm. It also was only skipping the slot tally for ancients. Now predators will properly be limited in slots at 4 + 1 per 80 clients + 0 (admin adjustment). Council do not check for a slot, and both ancient & council do not consume a slot.

Always intended for there to be a limit, thus the issue where aboms get swamped with late joining predators.

<details>

Tests were conducted where starting count was 0.

testing

As council/ancient:
image

Non ancient: 
image

</details>

# Changelog

:cl: Drathek
fix: Fixed predators having unlimited slots
/:cl:

### DIFF
--- a/code/game/gamemodes/cm_initialize.dm
+++ b/code/game/gamemodes/cm_initialize.dm
@@ -58,7 +58,7 @@ Additional game mode variables.
 	var/marine_starting_num = 0 //number of players not in something special
 	var/pred_current_num = 0 //How many are there now?
 	var/pred_per_players = 80 //Preds per player
-	var/pred_start_count = 4 //The initial count of predators
+	var/pred_start_count = 0 //The initial count of predators
 
 	var/pred_additional_max = 0
 	var/pred_leader_count = 0 //How many Leader preds are active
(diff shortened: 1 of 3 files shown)


Drulikar:
Starting count increased from 4 to 6. Leaders now also do not count for slots.

code/game/gamemodes/cm_initialize.dm
```diff
@@ -56,13 +56,14 @@ Additional game mode variables.
...
 	var/merc_starting_num = 0 //PMC clamp.
 	var/marine_starting_num = 0 //number of players not in something special
-	var/pred_current_num = 0 //How many are there now?
-	var/pred_per_players = 80 //Preds per player
-	var/pred_start_count = 4 //The initial count of predators
-
-	var/pred_additional_max = 0
-	var/pred_leader_count = 0 //How many Leader preds are active
-	var/pred_leader_max = 1 //How many Leader preds are permitted. Currently fixed to 1. May add admin verb to adjust this later.
+	/// How many predators utuilize slots currently
```
harryob, line 59:
```suggestion
	/// How many predators utilize slots currently
```

harryob (changes requested):
minor spelling mistake

harryob (approved):

PR #7802, opened 2024-12-12
Removes non combat click delay

Git-Nivrak:
# About the pull request
Basically #2130 but coded differently

# Explain why it's good for the game
#2130
# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
del: Removed non-combat click delay, interacting with items not on your person should be significantly smoother
balance: Clicking on the floor (open turfs) with an item will no longer cause a click delay
/:cl:

PR #8260, opened 2025-01-27
Revert "Removes non combat click delay"

Strobia:
Reverts cmss13-devs/cmss13#7802
This PR clearly states removing NON COMBAT click delay and that is misleading due to the fact that it HEAVILY affects combat and the current balance of the Yautja whitelist/all melee weapons + point blanks as marine.

🆑  LOLVAX1
Balance: Reverts the ''non combat'' click delay change due to it affecting combat heavily.

🆑

### DIFF
--- a/code/__DEFINES/_click.dm
+++ b/code/__DEFINES/_click.dm
@@ -1,4 +1,4 @@
-//Defines file for click related parameters
+//Defines file for byond click related parameters
 //this is mostly for ease of use and for finding all the things that use say RIGHT_CLICK rather then just searching "right"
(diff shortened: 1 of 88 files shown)


Zonespace27 (changes requested):
Please use the PR template.

Strobia:
CAN DO REVERT FEATURE SUCKS 1 SECOND!!!!!!!!!!!!

Drulikar:
We're well too late to be reverting this. Simply add delays to the items that are problematic.

PR #8263, opened 2025-01-27
Increases nuke timer to 3 minutes and makes it uncancelable, nuke autotriggers after decryption is finished.

ihatethisengine:
# About the pull request
Raises the timer of **techwebs** nuke to 3 minutes and makes it uncancelable.

Nuke autotriggers after decryption is finished.

Every time decryption is halted you lose 2 minutes of progress (to discourage people from holding "almost decrypted" nuke).

During decryption phase Queen can reset decryption by interacting with the nuke (just more consistent, it's weird that you would have to wait until it's decrypted otherwise).

Nuke is meant to end the round. But command often keeps it for later instead, still trying to win the game with regular methods while having a reliable backup plan (the nuke). This PR aims to discourage holding nuke for later, and to encourage using it right away, to end the round. This would also give xenos a better shot to contest the nuke itself. 

Again, nuke is meant to be a round-ender, not some soft of a backup plan you can save for later.

<details>

</details>

# Changelog
:cl: ihatethisengine
balance: raised the timer of techwebs nuke to 3 minutes and made it uncancelable, nuke autotriggers after decryption is finished.
/:cl:


Drulikar:
> Honestly I would even go as far as starting the nuke as soon as it is decrypted, but Drathek said no.
This is not what I said

tgui/packages/tgui/interfaces/NuclearBomb.jsx
```diff
@@ -7,6 +7,7 @@ export const NuclearBomb = () => {
 
   const cantNuke = (!data.anchor, !!data.safety, !data.decryption_complete);
   const cantDecrypt = (!data.anchor, data.decryption_complete);
+  const cantDisengage = (!data.anchor, !data.cantDisengage);
```
ihatethisengine, line 10:
```suggestion
  const cantDisengage = (!data.anchor, !data.can_disengage);
```

efzapa:
it should be clarified preferably in the last decryption stages that the nuke activation will start on decryption they *will not* be able to deactivate it.

because alooot of people won't be used to this auto trigger.

ihatethisengine:
> because alooot of people won't be used to this auto trigger.

Eventually, they will.

efzapa:
> > because alooot of people won't be used to this auto trigger.
> 
> Eventually, they will.

the notice would still be useful for clarity sake

code/game/machinery/nuclearbomb.dm
```diff
@@ -466,8 +467,24 @@ GLOBAL_VAR_INIT(bomb_set, FALSE)
...
+				to_chat(ui.user, SPAN_INFO("[src] is impossible to disengage now!"))
+				return
+		if("toggleSafety")
+			if(decrypting)
+				to_chat(ui.user, SPAN_INFO("Stop decryption first!"))
+				return
+		if("toggleAnchor")
+			if(decrypting)
+				to_chat(ui.user, SPAN_INFO("Stop decryption first!"))
+				return
```
kiVts, line 487: why add a second switch?

code/game/machinery/nuclearbomb.dm
```diff
@@ -621,11 +646,23 @@ GLOBAL_VAR_INIT(bomb_set, FALSE)
...
 	SIGNAL_HANDLER
 
 	if(!decrypting)
 		return
 
 	decrypting = FALSE
+	decryption_time = min(initial(decryption_time), decryption_time + 2 MINUTES)
 	announce_to_players()
+
+/obj/structure/machinery/nuclearbomb/tech/attack_hand(mob/user as mob)
```
kiVts, line 660: dont need `as mob`, ever.

kiVts (changes requested):

code/game/machinery/nuclearbomb.dm
ihatethisengine: To override parent's?

code/game/machinery/nuclearbomb.dm
ihatethisengine, line 660:
```suggestion
/obj/structure/machinery/nuclearbomb/tech/attack_hand(mob/user)
```

code/game/machinery/nuclearbomb.dm
kiVts: the you can move the parent call to the only action below

kiVts: the issue is when someone inevitably comes around to change things, they will see two switches and think of where to put their code. its not too big of a hussle to conjoin them either.

kiVts (approved):

code/game/machinery/nuclearbomb.dm
```diff
@@ -579,7 +579,7 @@ GLOBAL_VAR_INIT(bomb_set, FALSE)
 	var/list/humans_other = GLOB.human_mob_list + GLOB.dead_mob_list
```
Drulikar, line 590: Any reason to include dead mob list then if you are removing every mob that isn't conscious?

Drulikar (changes requested):

zzzmike:
In round 26819, I don't know if this was TM'd at the time but decryption was able to be 'started' after decryption had already completed - it immediately completed but either way this shouldn't happen. So, you may want to add something like a decrypted var and add a return if it's true in the if("toggleEncryption") flow.

cm13-github:

code/game/machinery/nuclearbomb.dm
Drulikar: Yes thus what I'm saying:
```js
	if(current_mob.stat != CONSCIOUS || isyautja(current_mob))
			humans_other -= current_mob
```

If your stat is `DEAD` it does not equal `CONSCIOUS` so you remove them no?

ihatethisengine: oh right

cm13-github:

ihatethisengine:
remove sprite and mapping tags please

PR #8273, opened 2025-01-29
Smartscope now uses alt IFF.

Detective-Google:
# About the pull request

This PR adds a new smartscope that replaces all current instances of the smartscope and uses the Alt IFF system(E.g, it prevents you from shooting at friendlies rather than firing through them.) 

the old smartscope has been removed entirely

The B8(new smartscope) now applies a 10% damage reduction modifier, and a .2 falloff modifier, but otherwise does not alter the gun's stats. 
The B8(new smartscope) Does not have the zoom feature.

The M41A MK2 can now equip the smartscope.

The smartscope largely exists as "training wheels" for newer/more casual players to play, however as it currently exists, completely destroys a weapon's stats-- making a player individually near useless, while also allowing for IFF stacking with multiple rifles, in which case it is very very strong. It is also restricted to some of the least used guns in the game, meaning the pulse rifle a newer player is often handed can't even mount it.

To remedy this, the smartscope now uses the alt IFF system, instead of being a DPS stacking item you can add to your frontline riflemen it instead acts as an assist item, similar to the recoil compensator, that allows you to trade some damage for the assist. 

<details>

https://github.com/user-attachments/assets/ceb47ac1-b0ba-4b57-b992-52c961248159

</details>

# Changelog
:cl: detectivegoogle, ihatethisengine
balance: The smartscope now uses alt IFF and cannot fire through friendlies.
balance: The smartscope now only applies a small damage and falloff malus to the mounted weapon.
balance: The smartscope can be mounted to the m41a MK2
balance: The smartscope has lost it's zoom functionality.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -926,6 +926,36 @@ Defined in conflicts.dm of the #defines folder.
 			. = TRUE
 	return .
 
+/obj/item/attachable/alt_iff_scope
+	name = "B8 Smart-Scope"
+	icon = 'icons/obj/items/weapons/guns/attachments/rail.dmi'
+	icon_state = "iffbarrel"
+	attach_icon = "iffbarrel_a"
+	desc = "An experimental B8 Smart-Scope. Based on the technologies used in the Smart Gun by ARMAT, this sight has integrated IFF systems. It can only attach to the M4RA Battle Rifle and M44 Combat Revolver."
+	desc_lore = "An experimental fire-control optic capable of linking into compatible IFF systems on certain weapons, designated the XAN/PVG-110 Smart Scope. Currently programmed for usage with the M4RA battle rifle and M44 Combat Revolver, due to their relatively lower rates of fire. Experimental technology developed by Armat, who have assured that all previously reported issues with false-negative IFF recognitions have been solved. Make sure to check the sight after every op, just in case."
+	slot = "rail"
+	pixel_shift_y = 15
+
+/obj/item/attachable/alt_iff_scope/New()
+	..()
+	damage_mod = -BULLET_DAMAGE_MULT_TIER_2
+	damage_falloff_mod = 0.2
+
+/obj/item/attachable/alt_iff_scope/set_bullet_traits()
+	LAZYADD(traits_to_give, list(
+		BULLET_TRAIT_ENTRY(/datum/element/bullet_trait_iff)
+	))
+
+/obj/item/attachable/alt_iff_scope/Attach(obj/item/weapon/gun/attaching_gun)
+	. = ..()
+	if(!GetComponent(attaching_gun, /datum/component/iff_fire_prevention))
+		attaching_gun.AddComponent(/datum/component/iff_fire_prevention)
+	SEND_SIGNAL(attaching_gun, COMSIG_GUN_ALT_IFF_TOGGLED, TRUE)
+
+/obj/item/attachable/alt_iff_scope/Detach(mob/user, obj/item/weapon/gun/detaching_gun)
+	. = ..()
+	SEND_SIGNAL(detaching_gun, COMSIG_GUN_ALT_IFF_TOGGLED, FALSE)
+
 /obj/item/attachable/scope
 	name = "S8 4x telescopic scope"
 	icon = 'icons/obj/items/weapons/guns/attachments/rail.dmi'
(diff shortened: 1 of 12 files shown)


private-tristan:
should either remove the old smartscope or make sure that it's still able to be applied to rifles (as right now it looks like, even if a mini iff scope spawns in, it'll be useless).

Detective-Google:
> should either remove the old smartscope or make sure that it's still able to be applied to rifles (as right now it looks like, even if a mini iff scope spawns in, it'll be useless).

the pre-existing IFF scope has not been touched at all except changing it's name and description to iterate it from B8 -> B9 so that it is identifiable from the new one. It still maintains all its functionality but is unobtainable as every source of mini_iff has been find&replaced with alt_iff

thwompa:
This is supported by me.

Syndro101:
Claiming this is "training wheels" for newer players who 99% of the time wont mount it seems pointless to me, not to mention it removing full-auto makes mounting it on the mk2 completely impractical. More experienced players will see the rancid downsides the attachment contributes and completely avoid it.

We already went through alt-IFF once with SG and the majority of the community absolutely despised it, doing the same to this scope that's already seen nerfs prior will just make it even less used.

I'd prefer if the smart scope was changed to be less available to prevent the "IFF-stacking" you've mentioned instead of turning the attachment into a gimmick that won't ever see use again.

I'd like to state I have no objection to balancing IFF, I just feel this isn't the direction it should take.

DrAmazing343:
The only way this is good training wheels for new players is if it still allows full auto because new players won’t understand semi auto carpal tunnel that makes the mk2 a fragger gun

BasilHerb:
Sounds like a well deserved change. IFF has been trending this way for a while (as it should), since IFF forces individual guns into being dogshit while simultaneously making them the most OP thing imaginable due to stacking capability. 

I am very interested to see how it will fare on the MK2 rifle- Speaking frankly, in 60% of cases marines that could fire, won’t due to fear of FF. The overall damage nerf doesn’t seem outrageous, so this would make the B8 unironically worth taking. People saying this makes the attachment ‘useless’ are massively underestimating how much fearRP impacts marines- Currently, if you want to unga without fear of FF, you basically need to frontline. Most marines behind you will never contribute a single shot to the battle, making them dead weight.

I have to hard agree with the others in letting MK2 keep full auto with the scope. You can already reach full auto fire rate by clicking on semi, so this change effectively accomplishes nothing and does not nerf the weapon’s fire rate. All it’s nerfing is the player’s chances of avoiding arthritis by 30. 

Marine buffs disguised as nerfs stay winning 🔥🔥🔥

code/modules/projectiles/gun_attachables.dm
```diff
@@ -1171,46 +1202,6 @@ Defined in conflicts.dm of the #defines folder.
 	select_gamemode_skin(type)
 	attach_icon = icon_state
 
-/obj/item/attachable/scope/mini_iff
```
realforest2001, line 1174: I would personally leave this in for potential event use, but that's up to you.

realforest2001 (approved):

code/modules/projectiles/gun_attachables.dm
Detective-Google: I had originally intended to do so, but I found that in practice a similar amount of varediting is needed to just make a gun IFF as to spawn in an un-obtainable item and varedit the desired guns to accept it, I decided it was not worthwhile to maintain the code for this niche use case as a result.

cm13-github:
#### Test Merge Updated

<details><summary>Details</summary>

##### Server Instance
CM13 Live

##### Revision
Origin: 8fde04edfa037e0e0ad1a4ee6b2c5cb661e30130
Pull Request: 61e30f00b73e3afca460bbd1ba3cad9dcbbb259f
Server: 46172eeae6604a1c4fa7d0c121070780ea0cca41

##### Comment

</details>

Neskah:
I like the idea, but this seems kind of random. The B8 was super niche before, but it was by no means a form of 'training wheels'. It allowed the weapons it was mounted on to perform an otherwise impossible action, thereby adding a new playstyle. The only thing it might 'train' you to do is friendly fire your allies.

 The new scope is technically a straight downgrade on a weapon, but it _can_ help suppression-heavy marines avoid friendly firing as much, and does act as a nice form of training wheels for new marines. If anything, I really think this change should be reverted, but the new B8 functionality should be added as some sort of smart holo or reflex sight, mountable on M41As and potentially other weapons.

And if people still think it's useless and bad, and you're trying to buff the B8, maybe add the new functionality along with a limited/weaker motion detector as a smart holo-sight.

code/modules/projectiles/gun_attachables.dm
```diff
@@ -926,6 +926,37 @@ Defined in conflicts.dm of the #defines folder.
 			. = TRUE
 	return .
 
+/obj/item/attachable/alt_iff_scope
+	name = "B8 Smart-Scope"
+	icon = 'icons/obj/items/weapons/guns/attachments/rail.dmi'
+	icon_state = "iffbarrel"
+	attach_icon = "iffbarrel_a"
+	desc = "An experimental B8 Smart-Scope. Based on the technologies used in the Smart Gun by ARMAT, this sight has integrated IFF systems. It can only attach to the M4RA Battle Rifle, the M44 Combat Revolver, and the M41A MK2 Pulse Rifle."
+	desc_lore = "An experimental fire-control optic capable of linking into compatible IFF systems on certain weapons, designated the XAN/PVG-110 Smart Scope. Currently programmed for usage with the M4RA battle rifle the M44 combat revolver, and the M41A MK2 pulse rifle. Experimental technology developed by Armat, who have assured that all previously reported issues with false-negative IFF recognitions have been solved. Make sure to check the sight after every op, just in case."
```
kiVts, line 935:
```suggestion
	desc = "An experimental B8 Smart-Scope. Based on the technologies used in the Smart Gun by ARMAT, this sight has integrated IFF systems. It can only attach to the M4RA Battle Rifle, the M44 Combat Revolver, and the M41A MK2 Pulse Rifle."
	desc_lore = "An experimental fire-control optic capable of linking into compatible IFF systems on certain weapons, designated the XAN/PVG-110 Smart Scope. Experimental technology developed by Armat, who have assured that all previously reported issues with false-negative IFF recognitions have been solved. Make sure to check the sight after every deployment, just in case."
```


Already said that in the description.

kiVts (approved):

PR #8278, opened 2025-01-29
Makes egg-sac eggs not support playable huggers 

Red-byte3D:
# About the pull request

If there is eggs BEFORE you put a cluster down as a carrier, the eggs will not support a playable hugger.  However any egg put after a cluster was placed will still support huggers(eggsac only)

Alot of people were complaining about how an oppressor + eggsac combo, or eggs designated to trap people by carrier players were being used up by playable huggers, this should remedy that while still allowing some huggers to spawn out of correct eggs.

<details>

</details>

# Changelog
:cl:
add: Eggsac carrier eggs no longer support playable huggers IF there is no cluster BEFORE they are placed, eggs put on clustered/hive weeds still allow huggers
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/egg.dm
+++ b/code/modules/cm_aliens/structures/egg.dm
@@ -457,6 +457,11 @@ SPECIAL EGG USED WHEN WEEDS LOST
 
 #define ORPHAN_EGG_MAXIMUM_LIFE 6 MINUTES // Should be longer than HIVECORE_COOLDOWN
 
+/obj/effect/alien/egg/carrier_egg/orphan
+	huggers_can_spawn = TRUE
+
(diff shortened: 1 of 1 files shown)


ihatethisengine:
Can't you make this a setting? Normal carrier can prevent hugs from spawning on him, why not do the same for eggsac?

Red-byte3D:
its how it was requested

Drulikar (dismissed):
`/obj/effect/alien/egg/carrier_egg/orphan` type should have huggers allowed again. These are the normal eggs that convert when they lose weeds, but restore to normal eggs if weeds are restored.

harryob (approved):

PR #8327, opened 2025-02-02
Painkilling/Electrogenetic Value and EGN triggers on defibrillating

private-tristan:
# About the pull request
Painkilling and Electrogenetic (Epi) are now twice as effective, halved their property level in chems (resulting in **no difference** in base painkillers)
Cardiostabilziing isn't changed by this to prevent inap from being better from tram
Electrogenetic now works when combined with the defibrillating property, draining and triggering a heal.

Each level of painkilling now gives you 20 pain reduction, with level 10 PNK letting you run around with no painslow at -149 health.

Each level of electrogenetic now heals 10 brute/burn/tox per trigger. See clip below for egn10 testing

Both EGN and PNK are, despite buffs, still really unused, especially painkilling, which was completely useless as oxycodone existed and gave you level 8 painkilling roundstart. By halving the levels and doubling the power it means that researchers making have much impact per level if they want to use this over NST/MST godstim slop.
EGN10 testing on an maxburn corpse that's still burning

https://github.com/user-attachments/assets/3a2f3be8-621a-4fb0-89d2-3306672e4567

# Changelog
:cl:
add: Electrogenetic property now triggers with defibrillating.
balance: Painkilling and Electrogenetic are twice as powerful per level, halved the property levels of Painkilling and Electrogenetic in chemicals, meaning that they act the same.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry_properties/prop_positive.dm
+++ b/code/modules/reagents/chemistry_properties/prop_positive.dm
@@ -580,11 +580,11 @@
 		return
 	var/mob/living/carbon/human/dead = M
 	var/revivable = dead.check_tod() && dead.is_revivable()
-	for(var/datum/reagent/R in M.reagents.reagent_list)
-		var/datum/chem_property/property = R.get_property(PROPERTY_ELECTROGENETIC) //Adrenaline helps greatly at restarting the heart
+	for(var/datum/reagent/electrogenetic_reagent in M.reagents.reagent_list)
+		var/datum/chem_property/property = electrogenetic_reagent.get_property(PROPERTY_ELECTROGENETIC) //Adrenaline helps greatly at restarting the heart
 		if(property)
 			property.trigger(M)
-			M.reagents.remove_reagent(R.id, 1)
+			M.reagents.remove_reagent(electrogenetic_reagent.id, 1)
 			break
 	if(revivable && (dead.health > HEALTH_THRESHOLD_DEAD))
 		addtimer(CALLBACK(dead, TYPE_PROC_REF(/mob/living/carbon/human, handle_revive)), 5 SECONDS)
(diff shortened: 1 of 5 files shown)


Git-Nivrak:
PNK8 (current) already keeps you alive almost until you die, doubling it is probably overkill. Do some testing on local maybe

private-tristan:
> PNK8 (current) already keeps you alive almost until you die, doubling it is probably overkill. Do some testing on local maybe

while you'll still be standing and able to fight, you'll still be severely painslowed.


Testing with PNK4: (Oxy)
4.5 painslow at 146 burn
3.5 painslow at 145 brute + unsplinted frac

Testing with PNK6: 
2.75 painslow at 146 burn
2.75 painslow at 145 brute + unsplinted frac

Testing with PNK8:
1.5 painslow at 146 burn
1 painslow at 145 brute + unsplinted frac

Testing with PNK10:
0 painslow at 146 burn
0 painslow at 145 brute + unsplinted frac.
1.5 painslow at 124 brute + 3 unsplinted fracs

Diegoflores31:
> PNK8 (current) already keeps you alive almost until you die, doubling it is probably overkill. Do some testing on local maybe

yes but all chems are going to be halved . i guess the objective is research being able to make something that is stronger than what you can get roundstart (oxycodone)

code/modules/reagents/chemistry_properties/prop_positive.dm
```diff
@@ -580,6 +580,12 @@
 		return
 	var/mob/living/carbon/human/dead = M
 	var/revivable = dead.check_tod() && dead.is_revivable()
+	for(var/datum/reagent/R in M.reagents.reagent_list)
+		var/datum/chem_property/property = R.get_property(PROPERTY_ELECTROGENETIC) //Adrenaline helps greatly at restarting the heart
+		if(property)
+			property.trigger(M)
+			M.reagents.remove_reagent(R.id, 1)
+			break
```
kiVts, line 588: single letter var >:(

kiVts (changes requested):
100+ damage on EGN 10 heal is bit cursed, but alright.

kiVts (approved):

PR #8349, opened 2025-02-04
Makes the vendor inflation announcements silent

zzzmike:
# About the pull request

Makes the frequent announcements about souto being 5% more expensive or whatever not play sound for anyone. (Also tried to improve the one letter var situation)

I feel like playing interference.ogg for such a frequent fluff-RP "event" a little annoying and pointless. (If the consensus is that I'm being a grinch then you can reject this PR, no problem.)

I did test it on local and it worked.

# Changelog
:cl:
sounddel: the vendor inflation event no longer plays sound
/:cl:

### DIFF
--- a/code/defines/procs/announcement.dm
+++ b/code/defines/procs/announcement.dm
@@ -155,7 +155,7 @@
 		if(istype(target, /mob/new_player))
 			continue
 
-	to_chat_spaced(target, html = "[SPAN_ANNOUNCEMENT_HEADER(title)]<br><br>[SPAN_ANNOUNCEMENT_BODY(message)]", type = MESSAGE_TYPE_RADIO)
+		to_chat_spaced(target, html = "[SPAN_ANNOUNCEMENT_HEADER(title)]<br><br>[SPAN_ANNOUNCEMENT_BODY(message)]", type = MESSAGE_TYPE_RADIO)
 		if(!quiet)
 			if(isobserver(target) && !(target.client?.prefs?.toggles_sound & SOUND_OBSERVER_ANNOUNCEMENTS))
 				continue
--- a/code/defines/procs/announcement.dm
+++ b/code/defines/procs/announcement.dm
@@ -127,36 +127,36 @@
 
 //AI shipside announcement, that uses announcement mechanic instead of talking into comms
 //to ensure that all humans on ship hear it regardless of comms and power
-/proc/shipwide_ai_announcement(message, title = MAIN_AI_SYSTEM, signature, ares_logging = ARES_LOG_MAIN, sound_to_play = sound('sound/misc/interference.ogg'), quiet = FALSE)
-    var/list/targets = GLOB.human_mob_list + GLOB.dead_mob_list
-    for(var/mob/target in targets)
-        if(isobserver(target))
-            continue
-        if(!ishuman(target) || isyautja(target) || !is_mainship_level(target.z))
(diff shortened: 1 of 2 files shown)


ihatethisengine:
can we make this "event" proc a bit less often?

hry-gh (changes requested):
snakecase (so target not Target)

code/defines/procs/announcement.dm
```diff
@@ -162,11 +180,19 @@
...
 			continue
 
-		to_chat_spaced(T, html = "[SPAN_ANNOUNCEMENT_HEADER(title)]<br><br>[SPAN_ANNOUNCEMENT_BODY(message)]", type = MESSAGE_TYPE_RADIO)
-		if(isobserver(T) && !(T.client?.prefs?.toggles_sound & SOUND_OBSERVER_ANNOUNCEMENTS))
+		to_chat_spaced(target, html = "[SPAN_ANNOUNCEMENT_HEADER(title)]<br><br>[SPAN_ANNOUNCEMENT_BODY(message)]", type = MESSAGE_TYPE_RADIO)
+		if(isobserver(target) && !(target.client?.prefs?.toggles_sound & SOUND_OBSERVER_ANNOUNCEMENTS))
 			continue
-		playsound_client(T.client, sound_to_play, T, vol = 45)
+		playsound_client(target.client, sound_to_play, target, vol = 45)
+/proc/announcement_helper_quiet(message, title, list/targets)
```
hry-gh, line 191:
```suggestion

/proc/announcement_helper_quiet(message, title, list/targets)
```

code/defines/procs/announcement.dm
```diff
@@ -145,14 +145,32 @@
 
 	announcement_helper(message, title, targets, sound_to_play)
 
+/proc/shipwide_quiet_ai_announcement(message, title = MAIN_AI_SYSTEM, signature, ares_logging = ARES_LOG_MAIN)
```
hry-gh, line 148: why not make this an additional argument?

code/defines/procs/announcement.dm
hry-gh: also ,why not make this an additional argument?

hry-gh (changes requested):

code/defines/procs/announcement.dm
```diff
@@ -127,13 +127,13 @@
 
 //AI shipside announcement, that uses announcement mechanic instead of talking into comms
 //to ensure that all humans on ship hear it regardless of comms and power
-/proc/shipwide_ai_announcement(message, title = MAIN_AI_SYSTEM, sound_to_play = sound('sound/misc/interference.ogg'), signature, ares_logging = ARES_LOG_MAIN)
```
hry-gh, line 130: there are a lot of calls of shipwide_ai_announcement, most of them use positional arguments for sound_to_play - why not just leave the order of arguments how they were and put quiet at the end?

hry-gh (changes requested):

code/defines/procs/announcement.dm
zzzmike: > there are a lot of calls of shipwide_ai_announcement, most of them use positional arguments for sound_to_play - why not just leave the order of arguments how they were and put quiet at the end?

agreed, my bad!

hry-gh (approved):

PR #8398, opened 2025-02-07
removes all inline control statements

hry-gh:
i've been telling people not do this for ages but i suppose copypasting is eternal


iloveloopers:
This PR removes CM history and soul

PR #8406, opened 2025-02-08
Fix surv tracking

Zonespace27:
# About the pull request
I didn't account for the fact that survs don't always have the survivor job, so stats were showing up skewed


code/datums/statistics/entities/survivor_survival.dm
```diff
@@ -15,12 +15,13 @@
...
-	for(var/mob/living/carbon/human/human as anything in GLOB.alive_human_list)
-		if(QDELETED(human))
+	for(var/datum/weakref/ref as anything in GLOB.spawned_survivors)
+		var/mob/living/carbon/human/human = ref.resolve()
+		if(human?.stat == DEAD)
 			continue
 
-		if(issurvivorjob(human.job))
-			remaining_survivors++
+		remaining_survivors++
```
hry-gh, line 23: on the offchance that a surv is gibbed, wouldn't this still increment the remaining_survivors, as human?.stat == DEAD will just resolve to false with the conditional access

Zonespace27: True enough. Fixing.

hry-gh (approved):

PR #8424, opened 2025-02-09
Lifeboat launch now checks for held ID

zzzmike:
# About the pull request

Previously you had to be wearing the ID properly, now you can also hold the ID

edge cases accounted for

tested it on local, worked

<details>

</details>

# Changelog
:cl:
qol: Previously you had to be wearing a valid ID to launch a lifeboat, now you can also hold it in hand
/:cl:

### DIFF
--- a/code/modules/shuttle/computer.dm
+++ b/code/modules/shuttle/computer.dm
@@ -416,9 +416,10 @@
 
 				var/mob/living/carbon/human/human_user = user
 				var/obj/item/card/id/card1 = human_user.get_idcard()
+				var/obj/item/card/id/card2 = human_user.get_idcard()
 
 				if(!card1 || (!(ACCESS_MARINE_SENIOR in card1.access) && !(ACCESS_MARINE_DROPSHIP in card1.access))) // if no card or not enough access, check for held id
-					var/obj/item/card/id/card2 = locate(/obj/item/card/id) in human_user
+					card2 = locate(/obj/item/card/id) in human_user
 
 				if(!card2 || (!(ACCESS_MARINE_SENIOR in card2.access) && !(ACCESS_MARINE_DROPSHIP in card2.access))) // still no valid card found?
 					to_chat(user, SPAN_NOTICE("[src]'s screen says \"Unauthorized access. Please inform your supervisor\"."))
--- a/code/modules/shuttle/computer.dm
+++ b/code/modules/shuttle/computer.dm
@@ -415,13 +415,12 @@
 					return
 
 				var/mob/living/carbon/human/human_user = user
-				var/obj/item/card/id/card1 = human_user.get_idcard()
-				var/obj/item/card/id/card2 = human_user.get_idcard()
+				var/obj/item/card/id/card = human_user.get_idcard()
 
-				if(!card1 || (!(ACCESS_MARINE_SENIOR in card1.access) && !(ACCESS_MARINE_DROPSHIP in card1.access))) // if no card or not enough access, check for held id
-					card2 = locate(/obj/item/card/id) in human_user
+				if(!card || (!(ACCESS_MARINE_SENIOR in card.access) && !(ACCESS_MARINE_DROPSHIP in card.access))) // if no card or not enough access, check for held id
+					card = locate(/obj/item/card/id) in human_user
 
-				if(!card2 || (!(ACCESS_MARINE_SENIOR in card2.access) && !(ACCESS_MARINE_DROPSHIP in card2.access))) // still no valid card found?
+				if(!card || (!(ACCESS_MARINE_SENIOR in card.access) && !(ACCESS_MARINE_DROPSHIP in card.access))) // still no valid card found?
 					to_chat(user, SPAN_NOTICE("[src]'s screen says \"Unauthorized access. Please inform your supervisor\"."))
 					return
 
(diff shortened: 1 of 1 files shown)


ihatethisengine:
not quite following, why? Why would you interact with a console while holding your ID in your hand? I can understand holding an ID of another person to launch

zzzmike:
> not quite following, why? Why would you interact with a console while holding your ID in your hand? I can understand holding an ID of another person to launch

the person that made me discover this as an issue said that they were used to having id in hand to interact with the keycard authentication console to set red alert, so they did the same here

ihatethisengine:
> the person that made me discover this as an issue said that they were used to having id in hand to interact with the keycard authentication console to set red alert, so they did the same here

I mean keycard authentication is a thing you are supposed to swipe, why would you use it on a computer I still do not really understand...

zzzmike:
Well, in any case, they tried. And this wouldn't be the only thing that can function via held id or worn id - airlocks for example. You can imagine the airlock detects any nearby ID or whatever, so same for the console. And in my real life job I did have to stick a chipped card into a card reader to log into a work computer, so there's precedent in that sense.

code/modules/shuttle/computer.dm
```diff
@@ -417,7 +417,13 @@
 				var/mob/living/carbon/human/human_user = user
 				var/obj/item/card/id/card = human_user.get_idcard()
 
-				if(!card)
+				if(!card) // if no card is worn, check for held id
+					for(var/obj/item/held_item in human_user.contents)
+						if(istype(held_item, /obj/item/card/id))
+							card = held_item // held id is set as the card
+							break
```
hry-gh, line 424: this is all equivalent to

```suggestion
					card = locate(/obj/item/card/id) in human_user
```

hry-gh (changes requested):

ihatethisengine:
> Well, in any case, they tried. And this wouldn't be the only thing that can function via held id or worn id - airlocks for example. You can imagine the airlock detects any nearby ID or whatever, so same for the console. And in my real life job I did have to stick a chipped card into a card reader to log into a work computer, so there's precedent in that sense.

Why do you compare the lifeboat's console with doors? Are there any other consoles that can work with ID being held, not worn? I am not aware of any. So this change would make it inconsistent if anything.

zzzmike:
> Why do you compare the lifeboat's console with doors?
Because you're using your ID on both to get access.

ihatethisengine:
> Because you're using your ID on both to get access.

But it is different though? You usually open doors by bumping into them, but there is another mechanic that allows you to open them by clicking with held ID. You still cannot bump-open doors by holding an ID in your hand. Yet again, there is no console that works like that. So, this is just inconsistent. 

More than that, your implementation is just confusing:

For example, let's use doors. You are a private, you found a SO corpse and looted their ID. Then you clicked a CiC door with the SO's ID and entered it. Good job!
Now to your current implementation. You are a private, you found a SO corpse and looted their ID. Then you clicked on a lifeboats launch console with the SO ID and... Access denied. Because the console checks for your worn ID first anyway. 

To make this work, you should either loop through all ID in player's content for sufficient access, or use attackby proc. 

But I am telling you again, this will be inconsistent. Currently if player tries to access a console by clicking it with ID, he fails. So he never does this again. If this gets merged, player will learn that he's actually able to access consoles by clicking them with ID, so he'll try the same on other consoles and fail. And there will be no way for player to tell if a console is accessible with held ID or not. Consistency matters, because player should be able to assume what he can and cannot do.

zzzmike:
> If this gets merged, player will learn that he's actually able to access consoles by clicking them with ID, so he'll try the same on other consoles and fail.

No. The player that tried this is a whitelisted synth. If your whole thing about "everyone uses every different consoles the same way after learning about one console (but airlocks and the authentication devices and whatever other devices aren't included because they're different somehow)" is true, this must be the first time a _whitelisted synth player_ used **any** console, right? Or they would've learned to not try this?

I'm not trying to make the entire game into some kind of perfect consistent state. I'm just reacting to an experienced player who decided to put their ID into their hand before interacting with the lifeboat and got 'access denied' and so figured that their class didn't have access. This PR is not perfect. Do I think it's still an improvement? Yup. That's my bar.

zzzmike:
In any case, it should at least grant access if either held or worn id has the right access, so changed it for that.

ihatethisengine:
https://github.com/cmss13-devs/cmss13/pull/8424/files#diff-7e56ae828b611a418f708858a7048ba827119dd767af422080fcf90e28b0b85dR365

Put required access there.

Replace your code with
`if(!allowed(human_user))`

ihatethisengine:
Wait, it all is still in attack_hand proc. Which means clicking your object (console) with held ID simply won't work. Same goes for the most of the machinery, they only react to attack_hand, while clicking something with an ID is attackby. So there is actually no sane and consistent way to do what you are trying to do.

Flight console, remote dropship control, ERT console all work with attack_hand, thus you cannot use them by clicking them with ID.

hry-gh (approved):
ihatethis is right though we should make all the id stuff more consistent

zzzmike:
> Wait, it all is still in attack_hand proc. Which means clicking your object (console) with held ID simply won't work.

I get what you're saying but I did test using the console with held ID (and active hand being the held ID) - it did work.

PR #8425, opened 2025-02-09
Removes pixel shifting

Git-Nivrak:
# About the pull request
Removes pixel shifting items

# Explain why it's good for the game
Cool idea in theory but in reality it is mostly being used for stuff like hiding claymores and in general the implementation is pretty buggy
# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
del: Removed pixel shifting
/:cl:

PR #8443, opened 2025-02-10
Reverts #8425, makes active claymores layer above (almost) all items so they cannot be hidden.

efzapa:
# About the pull request

Re-adds pixel shifting and makes it so claymores only appear under mobs (and whatever else is layered above mobs)

Nuking an entire feature because you died and not even looking to see how you can fix the root problem is bad, which is basically what  #8425 did.
While I am not disparaging the idea behind the PR, it *was* a good idea, you shouldn't be able to hide claymores, it's scummy.
The issue I have with the PR is that it took the nuclear route and the execution was near-abysmal.

On top of this, there is a PR up right now reworking Yautja Ship that utilizes, dare i say requires pixel shifting as a mechanic for the trophy racks.

There is no reason this should have been nuked when we apparently ALREADY had commented out code that would solve this instantly.

<details>

image

</details>

# Changelog
:cl:
add: Re-Added Pixel Shifting.
balance: Claymores will now appear above all items and cannot be hidden by them, you can however still trap bodies and tall grass.
/:cl:

### DIFF
--- a/code/__DEFINES/keybinding.dm
+++ b/code/__DEFINES/keybinding.dm
@@ -54,6 +54,10 @@
 #define COMSIG_KB_HUMAN_ROTATE_CHAIR "keybinding_human_rotate_chair"
 #define COMSIG_KB_HUMAN_SHOW_HELD_ITEM "keybinding_human_show_held_item"
 #define COMSIG_KB_HUMAN_CYCLE_HELMET_HUD "keybinding_human_cycle_helmet_hud"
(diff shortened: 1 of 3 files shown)


Drulikar:
Not interested thank you.

efzapa:
> Not interested thank you.

not interested in an objectively good fix/addition to the codebase?
not interested in pixel shifting?

I did what the guy was ***trying*** to do from the start, literally better

Drulikar:
Elaborated on discord, but for visibility: Pixel shifting was cited as likely problematic when it was originally proposed. Its already gone through several exploit passes. It is no longer deemed something worth trying to maintain.

PR #7483, opened 2024-11-03
Makes Sentinel's Scatter Spit a bit more consistent

TheKillfish:
# About the pull request

Moves scatter value from Scatter Spits main projectile to its bonus projectiles.

# Explain why it's good for the game

Since Scatter Spits scatter normally applies its scatter to every projectile including the main one, it means you can completely miss whoever it is you're actaully targeting, and in practice this feels like it happens more often than not. This gives consistency by ensuring at least 1 of your total 5 projectiles will go in the direction you want it to, as opposed to currently where it's damn a near coin toss if they hit who you're targeting or veer of around them instead.

Balance tag may be warrented at maintainer discretion.

# Testing Photographs and Procedure

Found Scatter Spit projectile code, moved scatter value from main projectile to additional projectile. Debated adding comparatively minor scatter (15) to main projectile too, but found it was not meaningfully inaccurate enough to justify having that over no scatter at all.

# Changelog

:cl: Killfish
balance: Improved consistency of Scatter Spits accuracy. Now one of your projectiles is guaranteed to to go in the direction you're aiming at, instead of potentially going everywhere except where you're aiming and missing everything.
/:cl:

PR #8444, opened 2025-02-11
Nerfs sentinel

Red-byte3D:
# About the pull request

Nerfs sentinels knockdown back to 2 instead of 2.5
Nerfs sentinels scatterspit cd to 7, not 8 like the old one but 7 where its a middle ground.
Nerfs sentinels para slash drop back to 4 seconds.

#7483
#7197
These two prs made sentinel very overpowered.
I believe that a 2.5 stun (as long as a lurker, off by .5) is NOT good for a tier one, there is quite literally no other caste that is as safe and as strong as sentinel right now, sentinel could reliably capture without these buffs and they most certainly still will be able to. With the difference now that you wont instnatly go down and forever, meaning the sentinel has to actually have a funcitoning strategy instead of just run at you.

The scatter spit is still VERY accurate, it will still always hit unless youre spitting at edge of the screen, meaning you can still follow up with a slash.

Also the goldendarkness55 pr was not closed even though he was CBANNED

<details>

</details>

# Changelog
:cl:
balance : Nerfs sentinels knockdown back to 2 instead of 2.5
balance : Nerfs sentinels scatterspit cd to 7, not 8 like the old one but 7 where its a middle ground.
balance : Nerfs sentinels para slash drop back to 4 seconds.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/sentinel/sentinel_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/sentinel/sentinel_abilities.dm
@@ -15,7 +15,7 @@
 	macro_path = /datum/action/xeno_action/verb/verb_scattered_spit
 	action_type = XENO_ACTION_CLICK
 	ability_primacy = XENO_PRIMARY_ACTION_2
-	xeno_cooldown = 6 SECONDS
+	xeno_cooldown = 7 SECONDS
 	plasma_cost = 30
 
 // Paralyzing slash
--- a/code/modules/mob/living/carbon/xenomorph/castes/Sentinel.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Sentinel.dm
@@ -68,7 +68,7 @@
 	// State
 	var/next_slash_buffed = FALSE
 
-#define NEURO_TOUCH_DELAY 3 SECONDS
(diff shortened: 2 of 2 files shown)


NexusStorm1:
I think the other changes are good, but you really should keep the drop at 3 seconds. Back when it was 4s it was way too easy for marines to escape into safety and it was difficult to capitalize on the stun, especially with the slash knockdown buff being reverted. 

Also, the changelog isn't accurate, as it seems like you're changing the scatterspit CD back to 7s, not the para slash CD?

Red-byte3D:
> I think the other changes are good, but you really should keep the drop at 3 seconds. Back when it was 4s it was way too easy for marines to escape into safety and it was difficult to capitalize on the stun. Otherwise, this PR is basically nerfing sentinel back to it's original state.
> 
> Also, the changelog isn't accurate, as it seems like you're changing the scatterspit CD back to 7s, not the para slash CD?

fixed

chefoxara:
spitter is way weaker than sent rn and sent got unneccessarily buffed, i approve of this PR
give spitters some strains or smth pls too
would be nice

PhantornRU:
Spitter and Sentinel - still most weaknes castes. Do we srsly need nerf sentinel, who died like 2 sec when marine shoot?

hry-gh (approved):
love twiddlign numbers

PR #8470, opened 2025-02-13
faster cellauto explosions

### LINKED ISSUE #8458: Walking through explosion shockwaves (blast walking)
### Testmerges

N/A

### Round ID

_No response_

### Description of the bug

Sometimes, you can walk into a shockwave tile the same time it travels to your tile, thus bypassing and "dodging" the explosion. This lets you avoid stuns.

https://github.com/user-attachments/assets/56685dad-59b1-4b19-acba-910f612bd8b5

The shockwave should apply its effects (stun, damage) to you every single time, as opposed to seemingly being unphased by it.

1. Throw grenade
2. Get lucky while walking towards it
3. Blast walk through the explosion, thus avoiding injury
...

### Issue Bingo

jupyterkat:
# About the pull request

makes explosions near instant
please tm this, there might be unforeseen consequences
Fixes #8458

no more dodging explosions by walking between the blastwaves

<details>

[Video.webm](https://github.com/user-attachments/assets/b9131acb-f881-4b3b-9fc3-169d51f82531)

</details>

# Changelog
:cl:
refactor: explosions are now faster
/:cl:


Willzadl:
I don't know whether this is verifiable, but in game it seems to have made grenades unpredictably laggy, sometimes taking a longer time to detonate

private-tristan:
This causes claymores to deal significantly more damage than previously.

code/modules/shuttles/marine_ferry.dm
```diff
@@ -467,14 +467,8 @@
...
-	var/explosion_alive = TRUE
-	while(explosion_alive)
-		explosion_alive = FALSE
-		for(var/datum/automata_cell/explosion/E in GLOB.cellauto_cells)
-			if(E.explosion_cause_data && E.explosion_cause_data.cause_name == "dropship crash")
-				explosion_alive = TRUE
-				break
-		sleep(1)
+	while(SSexplosions.is_exploding())
+		sleep(10)
```
fira, line 471: sleep(world.tick_lag)
or UNTIL() if you're fine with stoplag()

code/datums/autocells/explosion.dm
```diff
@@ -57,15 +54,12 @@
...
-	// Workaround to account for the fact that this is subsystemized
-	// See on_turf_entered
-	var/list/atom/exploded_atoms = list()
-
 	var/obj/effect/particle_effect/shockwave/shockwave = null
 
 // If we're on a fake z teleport, teleport over
 /datum/automata_cell/explosion/birth()
 	shockwave = new(in_turf)
+	QDEL_IN(shockwave, 0.2 SECONDS)
```
fira, line 62: I haven't looked in great detail but I don't think you want to tie in explosion lifecycle to time, because the whole contentious point with them and the reason this PR exists to begin with is that there is a noticeable drift between subsytem ticks and world time

code/controllers/subsystem/explosions.dm
```diff
@@ -0,0 +1,49 @@
+
+SUBSYSTEM_DEF(explosions)
+	name  = "Explosions"
+	wait  = 0.05 SECONDS
+	priority = SS_PRIORITY_EXPLOSIONS
+	flags = SS_NO_INIT
+
+	var/list/exploded_turfs = list()
+	var/list/currentrun = list()
```
fira, line 9:
```suggestion
	var/list/list/currentrun = list() //! List of currently processed explosion information
```

code/controllers/subsystem/explosions.dm
```diff
@@ -0,0 +1,49 @@
...
+		currentrun.len--
+
+		var/turf/exploded_turf = ex_args["exploded_turf"]
+
+		if (!exploded_turf || QDELETED(exploded_turf))
+			continue
+
+		exploded_turf.ex_act(ex_args["power"], ex_args["direction"], ex_args["explosion_cause_data"])
+		for(var/atom/exploded_article in exploded_turf)
+			if(!exploded_article || QDELETED(exploded_article))
```
fira, line 44: if it's in the turf, it's not deleted, considering this is ran on MC scheduling, not as a side effect

code/controllers/subsystem/explosions.dm
```diff
@@ -0,0 +1,49 @@
...
+		currentrun = exploded_turfs.Copy()
+		exploded_turfs.Cut()
+
+	while(length(currentrun))
+		var/list/ex_args = currentrun[length(currentrun)]
+		currentrun.len--
+
+		var/turf/exploded_turf = ex_args["exploded_turf"]
+
+		if (!exploded_turf || QDELETED(exploded_turf))
```
fira, line 39: turfs never delete 'proper', you don't need to check

code/controllers/subsystem/explosions.dm
```diff
@@ -0,0 +1,49 @@
...
+			continue
+
+		exploded_turf.ex_act(ex_args["power"], ex_args["direction"], ex_args["explosion_cause_data"])
+		for(var/atom/exploded_article in exploded_turf)
```
fira, line 43: as anything

code/datums/autocells/explosion.dm
```diff
@@ -209,40 +189,13 @@
...
-
-	exploded_atoms += A
-
-	// Note that we don't want to make it a directed ex_act because
-	// it could toss them back and make them get hit by the explosion again
-	if(A.gc_destroyed)
-		return
-
-	INVOKE_ASYNC(A, TYPE_PROC_REF(/atom, ex_act), power, null, explosion_cause_data)
-	log_explosion(A, src)
```
fira, line 245: Where did this go? It's dumb but it's important, because that's how game ensures you can't just walk through an explosion, and that chained explosions work properly

fira (commented):
Not a fan honestly

code/datums/autocells/explosion.dm
jupyterkat: this is the shockwave effect though, not the explosion itself

code/datums/autocells/explosion.dm
jupyterkat: not needed anymore, since explosions are sped up so much, this actually makes the performance worse.

code/modules/shuttles/marine_ferry.dm
jupyterkat: huh

code/datums/autocells/explosion.dm
fira: It's here for a reason. It's not just about things moving normally into each other at walkspeed, it also is relevant for things like recursive throws into explosions. Also, even if explosions happened in perfect sync with the rest, you'd still need to scan turfs twice per tick to ensure you sample player positions correctly to pick them up during subsytem fire. Otherwise someone can walk through an explosion before you scan for them if they moved in another way during previous tick. Think of it like Shannon-Nyquist theorem

jupyterkat: well then it clearly doesn't work anyways considering that blast walking is still a thing

jupyterkat: Its also not relevant for throws

PR #8491, opened 2025-02-15
Valkyrie Damage Mitigation Statistic

Zonespace27:
# About the pull request
Tracks the damage that valk mitigates for itself and others through use of its tantrum ability.

It's cool and it's in line with the already-existing healing tracking statistic for valk

image

(I have since added a floor() to this)
# Changelog

:cl:
add: Valkyries can now see how much damage they have mitigated through use of their Tantrum ability.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
@@ -121,7 +121,8 @@
 		"penetration" = penetration,
 		"armour_break_pr_pen" = armour_break_pr_pen,
 		"armour_break_flat" = armour_break_flat,
-		"armor_integrity" = armor_integrity
+		"armor_integrity" = armor_integrity,
+		"armour_type" = armour_type,
 	)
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -121,7 +121,8 @@
 		"penetration" = penetration,
 		"armour_break_pr_pen" = armour_break_pr_pen,
 		"armour_break_flat" = armour_break_flat,
-		"armor_integrity" = armor_integrity
+		"armor_integrity" = armor_integrity,
+		"armour_type" = armour_type,
```
Red-byte3D, line 125: should be armor not armour

Zonespace27: I'd normally agree but that'd be inconsistent with the other uses of the word "armour" in the damagedata lists throughout the code, and it's out of scope to fix that here

kiVts (approved):

PR #8492, opened 2025-02-15
chunk tgui topic payloads on 516

hry-gh:
can't send ultra big topics on 516 anymore, so we chunk it up instead


code/modules/tgui/tgui.dm
```diff
@@ -341,6 +344,30 @@
...
+			id = text2num(id)
+
+			var/total = text2num(href_list["totalPackets"])
+			if(id == 1)
+				if(total > MAX_MESSAGE_CHUNKS)
+					return
+
+				partial_packets = new /list(total)
+
+			partial_packets[id] = href_list["packet"]
```
Drulikar, line 359: When submitting a complex tacmap drawing:
image

Drulikar (changes requested):

Drulikar (approved):

PR #8501, opened 2025-02-16
Adds MC_TICK_CHECK to SSInactivity

Zonespace27:
# About the pull request

Adds MC_TICK_CHECK to SSInactivity, resuming the SS the next tick if it hasn't finished its job yet.

SSInactivity uses a lot of tick and overtimes more than you'd expect, so hopefully this should help. Would recommend TMing first.


Zonespace27:
image

(Yellow is the part we care about)
Works!

Git-Nivrak (approved):

PR #8534, opened 2025-02-19
Emergency lifeboat launch abort

zzzmike:
# About the pull request

Humans (with launch access) now can do an emergency abort of the lifeboat launch process once per game (potentially twice if queen hacks dropship after an abort). Takes 15 seconds to reset the console after an aborted launch before anyone can launch again.

Queens can already stop lifeboats from launching during the launch process. Why should xenos be able to stop the launch but a human with proper launch access cannot?

There are a few reasons why a human might want to, such as: someone unauthorized launching (with a stolen ID), the CO getting stuck in between the two lifeboat doors (aka the permadeath zone), someone obviously grief-launched, etc.

Have tested on local and made adjustments based on tests.

</details>

# Changelog
:cl:
add: Humans (with launch access) can now do an emergency abort of a lifeboat launch, once per game (potentially twice if queen hacks a dropship after an abort).
/:cl:


ihatethisengine:
Bloat. You should really keep it simple instead. IMO even emergency launch should be removed as just regular launch with 10 seconds warning is sufficient enough. There is no good reason to add more unnecessary mechanics.

PR #8606, opened 2025-02-27
refactors audio player (by adding a 516 version with nicer code)

hry-gh:
~~and when i no longer need to support 515 i can simply delete the old one~~ the old one was targeting ie8 compatibility and the new one works on ie11 so

:cl:
admin: the audio player works on 516 again - go crazy
/:cl:

### DIFF
--- a/tgui/packages/tgui-panel/audio/middleware.js
+++ b/tgui/packages/tgui-panel/audio/middleware.js
@@ -4,10 +4,12 @@
  * @license MIT
  */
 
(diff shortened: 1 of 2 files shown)


Bm0n:
From my testing this doesn't stop music if a song is already playing anymore meaning multiple songs can play at once

This can easily be fixed by calling `stop_music` right before `play_music` on the DM side of things- not sure how to fix it in typescript though!

Drulikar (approved):

PR #8607, opened 2025-02-27
optimization to chat saving via indexeddb improvements

hry-gh:
# About the pull request

we batch up chat messages and send them over in groups to indexeddb across the iframe, and write them in a ring which loops round at MAX_MESSAGES. when being read, we read from the start of the loop and reassemble the last MAX_MESSAGES number of messages

title if that makes sense to like anyone

# Changelog

:cl:
code: Chat saving to indexdb is reworked
/:cl:

### DIFF
--- a/tgui/packages/common/storage.js
+++ b/tgui/packages/common/storage.js
@@ -107,7 +107,7 @@ class IFrameIndexedDbBackend {
   async get(key) {
     const promise = new Promise((resolve) => {
       window.addEventListener('message', (message) => {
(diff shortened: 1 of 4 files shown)


Drulikar (approved):
Seems to work locally but will TM it for a bit.

Drulikar:
Seems to not gracefully handle old data for most people some reason, but otherwise fine after a full webview reboot and cache clear?

Drulikar (approved):

PR #8626, opened 2025-03-01
Halves the untrained pointing cooldown

Willzadl:
# About the pull request

pointing cooldown now = 25 vs 50 ticks prior, leadership cooldown remains at 10 and with big arrows

as it stands, the pointing cooldown of 50 is awkwardly long, and while it prevents spamming it also negatively impacts tactical awareness and easy non-verbal communication

halved cooldown should still prevent spamming, shouldn't outclass the pointing abilities of leaders, but allow for more efficient communication (especially for indicating cross-referencing between 2 pointed-at objects)

its just a constant edit

</details>

# Changelog
:cl:
qol: halved the untrained pointing cooldown
/:cl:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -385,7 +385,7 @@
 	if(ishuman(mob))
 		squad = mob.assigned_squad
 	if(!check_improved_pointing()) //Squad Leaders and above have reduced cooldown and get a bigger arrow
-		recently_pointed_to = world.time + 50
+		recently_pointed_to = world.time + 25
 		new /obj/effect/overlay/temp/point(T, src, A)
 	else
 		recently_pointed_to = world.time + 10
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -385,7 +385,7 @@
 	if(ishuman(mob))
 		squad = mob.assigned_squad
 	if(!check_improved_pointing()) //Squad Leaders and above have reduced cooldown and get a bigger arrow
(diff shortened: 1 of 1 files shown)


code/modules/mob/mob.dm
```diff
@@ -385,7 +385,7 @@
 	if(ishuman(mob))
 		squad = mob.assigned_squad
 	if(!check_improved_pointing()) //Squad Leaders and above have reduced cooldown and get a bigger arrow
-		recently_pointed_to = world.time + 50
+		recently_pointed_to = world.time + 25
```
Red-byte3D, line 388: Use SECONDS

Willzadl: what does that mean

Detective-Google: word.time + 3 SECONDS (with 3 being however long the cooldown should be)

Willzadl: i dont understand, it's not like i can just write '3 seconds' can i ??

Willzadl: oh wait it's built in as a module, that's great man i will change that

Willzadl: unfortunately i have no idea how the original code translates from '50' into seconds, if anybody else wants to approach this as a refactor that would be great but i don't

Red-byte3D: 50 is just byond ticks which is 5 seconds, so you would write 2.5 SECONDS

kiVts (approved):

PR #8628, opened 2025-03-01
Adds Fan The Hammer to the M44 revolver

Red-byte3D:
# About the pull request

Adds fan the revolver, a new weapon action that lets you fire all 7 shots of the revolver in succession (think burst fire)

 i think this will add a fun way to play the weapon, make it  high risk high reward (it does alot of damage if youre next to the xeno) and i think it will make the weapon a bit more relevant

I'd like a tm for a few thousand rounds to see what happens or how it plays

https://github.com/user-attachments/assets/553bbf34-98ed-4806-bfc1-91e19e6d62ae

# Changelog
:cl:
add: M44 Can now "Fan The Hammer" to fire all 7 shots in succession.
/:cl:


private-tristan:
if you enable fan the hammer you can toggle firemode and return to single fire (but able to swap back to burst).

Due to how burst PBs work, you fire all 7 shots **instantaneously**, this means dealing 504 dmg in a single tick if you're PBing a large mob, even 1handed. If you manage to land 2 PBs you crit anything short of a crusher or hedgehog ravager when using base ammo.

This means that, using base ammo, you're able to crit a praetorian using 2 bursts in a bit under 2 seconds.

https://github.com/user-attachments/assets/a28db6ef-c9b2-471f-9996-9afbe70b98c0

easiest way to stop this would probably be making it unable to PB while fan firing?

SomeoneYouProbablyKnow:
The shotgunification of all weapons

Mrrpip:
Figuring I nay aswell ask for my curiousity's sake.

But how does this interact with... Well... Already holding an item in your offhand? I may be wrong here, but the code doesn't mention any restrictions when the firemode is on.

I know you could probably just handwave it with CMB styled burst fire, but that sounds less cool than yeehaw pow pow pow and all that.

SomeoneYouProbablyKnow:
Should make you drop offhand item. that's make sense and be cool

SomeoneYouProbablyKnow:
Told you its the shotgunification of all weapons

code/modules/projectiles/guns/revolvers.dm
```diff
@@ -351,11 +386,21 @@
...
 /obj/item/weapon/gun/revolver/m44/able_to_fire(mob/user)
-	if (folded)
+	if(folded)
 		to_chat(user, SPAN_NOTICE("You need to unfold the stock to fire!"))//this is stupid
-		return 0
-	else
+		return FALSE
+	if(!currently_fanning)
 		return ..()
+	if(currently_fanning && user.get_inactive_hand() && !istype(user.get_inactive_hand(), /obj/item/weapon/twohanded/offhand))
```
ihatethisengine, line 394:
```suggestion
	if(user.get_inactive_hand() && !istype(user.get_inactive_hand(), /obj/item/weapon/twohanded/offhand))
```

ihatethisengine: already checked

code/modules/projectiles/guns/revolvers.dm
```diff
@@ -351,11 +386,21 @@
...
+		to_chat(user, SPAN_NOTICE("You can't fan the hammer when something else is in your hand"))
+		return FALSE
+	var/hand_to_check = user.r_hand == src ? "l_hand" : "r_hand"
+	var/mob/living/carbon/human/revolver_wielder = user
```
ihatethisengine, line 398:
```suggestion
	var/mob/living/carbon/human/revolver_wielder = user
	if(!istype(revolver_wielder))
		return ..()
```

ihatethisengine: to make sure it won't run time is the wielder is event xeno

Generallemon07:
One small issue. You cannot fan a double action revolver, like physically.

PR #8633, opened 2025-03-02
Stricter AFK grace timer

Drulikar:
# About the pull request

This PR simply changes an old 20 min grace period to 3 mins after round start.

This is likely the reason why early game larva are afk and also not getting taken over by ghosts.

# Changelog

:cl: Drathek
balance: Grace timer for being AFK is reduced from 20 mins down to 3 mins after round start
/:cl:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -240,7 +240,8 @@
 	SHOULD_NOT_SLEEP(TRUE)
 	if(client == null)
 		away_timer++
-	if(client == null || (client.inactivity > 1 && world.time > 20 MINUTES)) //Do not start away_timer on connected clients until 20 minutes has passed.
+	var/game_started = SSticker.current_state == GAME_STATE_PLAYING && ROUND_TIME > 3 MINUTES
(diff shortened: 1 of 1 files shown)


ihatethisengine:
Who cares about early game larva being afk? It's not like there is much to do roundstart but wait for evo. A lot of players use this time to do stuff IRL without having to worry about latejoining via queue that can take forever.

Red-byte3D:
To add on ii dont see why we cant just ghost larva mobs instead of everyone pre 20 minutes, that way the evo isnt lost either

Drulikar:
Not budging on something that does get ahelped decently frequently and I was asked about specifically. If you're AFK you're AFK.

ihatethisengine:
The fact something getting ahelped doesn't make it problematic  :man_facepalming: 

Rounds already take hours, how do you expect people to not afk at all. Especially when it doesn't matter (roundstart).

hry-gh:
> It's not like there is much to do roundstart but wait for evo.

add better game content then

PR #8652, opened 2025-03-04
some Tgui fixes

Kashargul:
# About the pull request

There's a rare case where the chat might not init, this also adds a second check if telemetry was received and if not, re-requests it once. 

Fixing some tgui issues.

recommending to TM.

Brings the chat closer in sync with https://github.com/tgstation/tgstation again, some of the other changes that exist on TG are likely better for you to check on your own, especially the option to link / unlink font sizes, etc.

<details>

</details>

# Changelog
:cl:
fix: numberInput works properly again with mouse movement
fix: the chat will no longer init before settings are loaded
qol: chat settings can be in / exported
qol: chat tabs can be reordered
/:cl:


cm13-github:

cm13-github:

Willzadl:
you broke chat highlights!

Kashargul:
> you broke chat highlights!

Checking later

tgui/packages/tgui-panel/settings/middleware.js
```diff
@@ -1,73 +0,0 @@
...
-  removeHighlightSetting,
-  updateHighlightSetting,
-  updateSettings,
-} from './actions';
-import { FONTS_DISABLED } from './constants';
-import { selectSettings } from './selectors';
-
-const setGlobalFontSize = (fontSize) => {
-  document.documentElement.style.setProperty('font-size', fontSize + 'px');
-  document.body.style.setProperty('font-size', fontSize + 'px');
```
Drulikar, line 22: This appears to be the cause for the ping layout change. As far as why: that top panel I think is now handling font size differently than it used to (e.g. 12-14pt looks fine but not higher or lower). If I revert changes to middleware.ts/.js it then it resizes correctly.

image

tgui/packages/tgui-panel/chat/reducer.ts
```diff
@@ -188,5 +210,52 @@ export const chatReducer = (state = initialState, action) => {
     }
     return nextState;
   }
+  if (type === moveChatPageLeft.type) {
```
Drulikar, line 213: The first tab cannot be moved right, and the second tab cannot be moved left.

image

Drulikar (changes requested):
- Import/Export appears to be working great.
- I haven't been able to find any issues with the rework to NumberInput.
- As far as I can tell you ported linked statpanel font size, but I don't see a means to do this by a user. If you wanted to just utilize our existing stat font size, you likely only need to send a message to the statpanel. E.g. in dm code `client.stat_panel.send_message("change_fontsize", new_fontsize)` or a tgui equivalent to trigger `Byond.subscribeTo("change_fontsize", function (new_fontsize)`.

tgui/packages/tgui-panel/chat/reducer.ts
Kashargul: Yes, you can never swap the main tab, it always stays first.

tgui/packages/tgui-panel/settings/middleware.js
Kashargul: Ye the element differs from TG, needs a separate exclusion. Will check later.

Kashargul: WE just make a proper custom button and don't import and override an element

Drulikar (approved):

PR #8671, opened 2025-03-05
Fixes lizard recursive check and infinite knockout

VileBeggar:
# About the pull request

fixes recursive loop with giant lizards that happened due to receiving damage while out of range of any enemies
fixes lizards being knocked out for all eternity due to explosives

bug fix very good.

<details>

https://github.com/user-attachments/assets/2216104c-d643-4ff5-ab70-0a8b5d16c442

</details>

# Changelog
:cl:
fix: Giant Lizards will no longer cause recursive loops (the server won't randomly decide to die anymore). They will now no longer be permanently stunned by explosions.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/retaliate/giant_lizard.dm
+++ b/code/modules/mob/living/simple_animal/hostile/retaliate/giant_lizard.dm
@@ -778,7 +778,7 @@
 	alert_others()
 
 /mob/living/simple_animal/hostile/retaliate/giant_lizard/proc/alert_others()
-	for(var/mob/living/simple_animal/hostile/retaliate/giant_lizard/pack_member in GLOB.giant_lizards_alive)
+	for(var/mob/living/simple_animal/hostile/retaliate/giant_lizard/pack_member as anything in GLOB.giant_lizards_alive)
 		if(pack_member == src || pack_member.target_mob_ref?.resolve() || get_dist(src, pack_member) > 7)
 			continue
 		pack_member.Retaliate(pack_attack = TRUE)
--- a/code/modules/mob/living/simple_animal/hostile/retaliate/giant_lizard.dm
+++ b/code/modules/mob/living/simple_animal/hostile/retaliate/giant_lizard.dm
@@ -762,9 +762,11 @@
 
 	target_mob = FindTarget()
 	target_mob_ref = WEAKREF(target_mob)
-	if(target_mob_ref) // qdeleted check
-		growl(target_mob)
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/simple_animal/hostile/retaliate/giant_lizard.dm
```diff
@@ -762,9 +762,11 @@
...
-		growl(target_mob)
-		MoveToTarget()
+	if(!target_mob_ref)
+		return
+
+	growl(target_mob)
+	MoveToTarget()
 
 	//basic pack behaviour
 	for(var/mob/living/simple_animal/hostile/retaliate/giant_lizard/pack_member in view(12, src))
```
Watermelon914, line 772: Isolate the recursive nature of this proc into its own proc that you then call.

Assign a depth value that decreases with each call so that this doesn't recurse infinitely if there are, say, a bunch of lizards scattered over the map.

Watermelon914: I think you should make this only go 1 layer deep btw, it shouldn't be recursing more than once because it doesn't really need to and it leads to more problems that way.

Drulikar: Alternatively make a GLOB list of the lizards, and just iterate that list and distance check them.

code/modules/mob/living/simple_animal/hostile/retaliate/giant_lizard.dm
```diff
@@ -768,11 +772,16 @@
...
-	//basic pack behaviour
-	for(var/mob/living/simple_animal/hostile/retaliate/giant_lizard/pack_member in view(12, src))
-		if(pack_member == src || pack_member.target_mob_ref?.resolve())
+	if(pack_attack)
+		return
+
+	alert_others()
+
+/mob/living/simple_animal/hostile/retaliate/giant_lizard/proc/alert_others()
+	for(var/mob/living/simple_animal/hostile/retaliate/giant_lizard/pack_member in GLOB.giant_lizards_alive)
```
Drulikar, line 781:
```suggestion
	for(var/mob/living/simple_animal/hostile/retaliate/giant_lizard/pack_member as anything in GLOB.giant_lizards_alive)
```

Drulikar (changes requested):
Looks good otherwise. Be sure to undraft when you think its ready.

Drulikar (approved):

PR #8674, opened 2025-03-05
fixes the sticky keys

hry-gh:
the window was still keeping focus because it reset its geometries after being closed. stupid

:cl:
fix: tgui windows closing should no longer cause sticky keys
/:cl:


TheGamerdk:
ehm, harry...
image

Drulikar (approved):

PR #8686, opened 2025-03-07
Increases Mortar Belt and Mortar Backpack Storage

MeH0y:
# About the pull request

Increases the number of shells able to be carried in both the mortar belt and backpack.
Mortar belt storage: 4 > 8
Mortar backpack storage: 8 >12
While my initial focus was on the belt, it didn't make sense for the belt and backpack to be on the same level.
These storages were not altered alongside the mortar rework and were already useless beforehand. Since the mortar is now mobile, these storage items should be viable.

I fit more shells in them
# Changelog
:cl:
balance: Increased shell storage capacity of the mortar belt and mortar shell backpack
/:cl:


DoguhanOzgurAkca:
Mortar really needs to become way more wieldy if the semi-frontline mortar is to be a thing honestly. Most of the time, its not used from my experience

Git-Nivrak (approved):

PR #8707, opened 2025-03-09
No more T0 triggering claymore

seanlol092:
# About the pull request

Title

This prevents hugger and lesser drones from triggering mines. A hugger or lesser drone triggering a minefield by thier own (one person clearing the minefield) shouldnt be possible. They could still trigger proxy claymores

<details>

</details>

# Changelog
:cl:
balance: Claymore cant be trigger by huggers and Lesser drones
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/mine.dm
+++ b/code/game/objects/items/explosives/mine.dm
@@ -202,6 +202,10 @@
 /obj/item/explosive/mine/proc/try_to_prime(mob/living/L)
 	if(!active || triggered || (customizable && !detonator))
 		return
+	if(isxeno(L))
+		var/mob/living/carbon/xenomorph/xeno = L
+		if(xeno.mob_size <= MOB_SIZE_XENO_VERY_SMALL)
+			return
 	if(!istype(L))
 		return
 	if(L.stat == DEAD)
@@ -241,6 +245,11 @@
 	if(M.a_intent == INTENT_HELP)
 		to_chat(M, SPAN_XENONOTICE("If you hit this hard enough, it would probably explode."))
 		return XENO_NO_DELAY_ACTION
+
+	if(tripwire)
+		if(M.mob_size <= MOB_SIZE_XENO_VERY_SMALL)
+			to_chat(M, SPAN_XENONOTICE("You are too weak to slash this claymore."))
+			return XENO_NO_DELAY_ACTION
 
 	M.animation_attack_on(src)
 	M.visible_message(SPAN_DANGER("[M] has slashed [src]!"),


ihatethisengine:
In fact this is the only way to "disarm" OT without losing actually valuable xenos.

efzapa:
i feel like making them explode by spits was probably the best solution for all of this

seanlol092:
> i feel like making them explode by spits was probably the best solution for all of this

Will try to make it but it will take time since i have limited knowledge about it

seanlol092:
ngl should have made seperate PR for it

DrAmazing343:
EOD Spitter Jellies incoming

seanlol092:
Also this is not just OT claymore its all of claymores

seanlol092:
> EOD Spitter Jellies incoming

EOD Spitter have risk of accidentally stepping on hidden claymore rather than hugger / lesser drone disarming them without any risk for the xenos

seanlol092:
I would love to see spitter spitting at claymore to kill marines

Diegoflores31:
i dont know if  its a good idea that spits detonate the mines , sentinels can just blow them up from 7 tiles away and very quickly .
maybe it should be with acid spray that has more cooldown and shorter range

Diegoflores31:
> i dont know if its a good idea that spits detonate the mines , sentinels can just blow them up from 7 tiles away and very quickly . maybe it should be with acid spray that has more cooldown and shorter range

alternatively the explosion when disarmed should be hardcoded rather than prime()

efzapa:
> i dont know if  its a good idea that spits detonate the mines , sentinels can just blow them up from 7 tiles away and very quickly .
> maybe it should be with acid spray that has more cooldown and shorter range

if we limit it to castes with only acid spit then it's only like 3 castes in the whole game that can do it, PLUS you have to actually notice it to spit it to begin with

I dont think this will be that huge of a change

seanlol092:
> i dont know if  its a good idea that spits detonate the mines , sentinels can just blow them up from 7 tiles away and very quickly .
> maybe it should be with acid spray that has more cooldown and shorter range


Acid spray doesnt trigger them since for some reason they aren't counted as getting hit by claymore. The claymore needs to get hit **twice** by a projectile spit. There's a spit test of all xeno projectile I've tested ( / = detonates, x = doesn't detonates).

I guess making it detonates at three hits would be fine or four if it's neuro spit

seanlol092:
> if we limit it to castes with only acid spit then it's only like 3 castes in the whole game that can do it, PLUS you have to actually notice it to spit it to begin with
> 
> I dont think this will be that huge of a change 

It isn't just limit to acid spit. It's all xeno spits like neuro, scatter spit, etc.

About the layer of the claymore I think it should be always on top

kiVts (changes requested):

code/game/objects/items/explosives/mine.dm
```diff
@@ -264,6 +274,15 @@
 	if(!QDELETED(src))
 		disarm()
 
+/obj/item/explosive/mine/bullet_act(obj/projectile/xenoprojectile)
```
Detective-Google, line 277: xeno_projectile

we use a formatting called snake case where words are all lowercase separated by underscores (can see it there with bullet_act)

code/game/objects/items/explosives/mine.dm
```diff
@@ -202,6 +203,10 @@
 /obj/item/explosive/mine/proc/try_to_prime(mob/living/L)
 	if(!active || triggered || (customizable && !detonator))
 		return
+	if(isxeno(L))
+		var/mob/living/carbon/xenomorph/xeno = L
+		if(xeno.mob_size <= MOB_SIZE_XENO_VERY_SMALL)
```
Drulikar, line 208: This should only be prevented if this mine is configured to only trigger for larger targets.

code/game/objects/items/explosives/mine.dm
```diff
@@ -264,6 +273,15 @@
 	if(!QDELETED(src))
 		disarm()
 
+/obj/item/explosive/mine/bullet_act(obj/projectile/xeno_projectile)
+	if(!triggered && istype(xeno_projectile.ammo, /datum/ammo/xeno)) //xeno projectile
+		hit_count++
+		if(hit_count >= 2) // Check if hit two times
+			visible_message(SPAN_DANGER("\The [src] is hit by [xeno_projectile] and violently detonates!")) // Acid is hot for claymore
```
Drulikar, line 280:
```suggestion
			visible_message(SPAN_DANGER("[src] is hit by [xeno_projectile] and violently detonates!")) // Acid is hot for claymore
```

code/game/objects/items/explosives/mine.dm
```diff
@@ -242,6 +246,11 @@
 		to_chat(M, SPAN_XENONOTICE("If you hit this hard enough, it would probably explode."))
 		return XENO_NO_DELAY_ACTION
 
+	if(tripwire)
+		if(M.mob_size <= MOB_SIZE_XENO_VERY_SMALL)
+			to_chat(M, SPAN_XENONOTICE("You are too weak to slash this claymore."))
```
Drulikar, line 251:
```suggestion
			to_chat(M, SPAN_XENOWARNING("Slashing that would be suicide!"))
```

Drulikar (changes requested):
Rather than mines always ignoring small xenos, make it so mines can be configured to ignore small xenos.

code/game/objects/items/explosives/mine.dm
seanlol092: I have now made it so it can be toggle by using a wirecutter, i was hoping to use cable coil but i dont see any trait_tool for it

code/game/objects/items/explosives/mine.dm
```diff
@@ -186,32 +197,39 @@
...
-	if(!istype(L))
+
+	if(sensor_for_small == FALSE)
+		if(isxeno(enemy))
+			var/mob/living/carbon/xenomorph/xeno = enemy
+			if(xeno.mob_size <= MOB_SIZE_XENO_VERY_SMALL)
+				return
+		return
+	if(!istype(enemy))
 		return
```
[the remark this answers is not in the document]
seanlol092: Should be working properly now. Right side is modified

https://github.com/user-attachments/assets/6e9475a7-5690-4bde-8bba-721bfd1782b4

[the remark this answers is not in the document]
seanlol092: nvm. placing it as UPP or other IFF sets the claymore's IFF to theirs

Drulikar (approved):
Let me know if you're okay with the refactoring and changes I did. Github suggestions only allow suggestions within a limited range of context so just committed suggestions instead and easier for me to validate I don't suggest bugs. Will just TM it in the meantime.

Drulikar (changes requested):
Actually nevermind, pushing using the github extension does it differently than I expected. See https://github.com/seanlol092/cmss13/pull/1 instead.

code/game/objects/items/explosives/mine.dm
```diff
@@ -8,7 +8,7 @@
 	icon_state = "m20"
 	force = 5
 	w_class = SIZE_SMALL
-	//layer = MOB_LAYER - 0.1 //You can't just randomly hide claymores under boxes. Booby-trapping bodies is fine though
+	layer = MOB_LAYER - 0.1 //You can't just randomly hide claymores under boxes. Booby-trapping bodies is fine though
```
Drulikar, line 11:
```suggestion
	layer = ABOVE_LYING_MOB_LAYER //You can't just randomly hide claymores under boxes. Booby-trapping bodies is fine though
```
Same number

Drulikar (approved):

PR #8716, opened 2025-03-10
Tactical Prybars will no longer attack cades while linking

Willzadl:
# About the pull request

fixes a strange code interaction at #8693 
we dont want to attack our cades

tested and it works sorry i didnt want to make a video

# Changelog

:cl:
fix: tactical prybars will no longer attack cades when trying to link them together
/:cl:


kiVts (changes requested):
Thats not the way to do it, instead when you link plasteel cades, return.

Willzadl:
> Thats not the way to do it, instead when you link plasteel cades, return.

my bad, i was just comparing the differences between that one and the other crowbar variants. i dont understand what the problem is otherwise, so i'll close this

PR #8723, opened 2025-03-10
dpi touchups

hry-gh:
:cl:
add: there's now a new tgui scale option in settings for people with high dpi monitors using windows scaling that allows you to toggle between having windows be larger but their contents the correct size, or having windows be the normal size but their contents shrunk down
fix: people with a different windows scaling option than 100% on 516 can drag windows normally again
/:cl:


Drulikar (changes requested):
Behavior is a little weird when swapping back (doesn't seem to properly return the setting back):
First launch (at 150%) - this is ok:
image
Change to smaller scale setting, save, and reopen windows - this is ok:
image
Restore setting back to default, save, reload windows - not ok?:
image

hry-gh:
> Behavior is a little weird when swapping back (doesn't seem to properly return the setting back):

resolved

Drulikar (approved):

PR #8734, opened 2025-03-11
yet more dpi shenanigans

hry-gh:
# About the pull request

follow up to #8723, just respects the tgui scaling (now window scaling) pref globally, with windows either becoming bigger or zooming out

this now applies to most simple html browsers, tgui say and some other miscellany 

looks less silly :)

4k monitor with 200% scaling:

on "smaller zoom" mode:

rPrHMkkBa310nEas@2x

on "larger windows" mode:
cfhepGVohf94UYRQ@2x

# Changelog

:cl:
qol: people with window scaling that is greater or less than 100% will now have windows scale, appropriate to the "window scaling" option in settings
/:cl:

### DIFF
--- a/code/datums/browser.dm
+++ b/code/datums/browser.dm
@@ -70,7 +70,7 @@
 	head_content += "<link rel='stylesheet' type='text/css' href='[common_asset.get_url_mappings()[stylesheet]]'>"
 	head_content += "<link rel='stylesheet' type='text/css' href='[other_asset.get_url_mappings()["loading.gif"]]'>"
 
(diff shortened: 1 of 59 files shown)


Drulikar (approved):
Appears to work fine. Though of note character preferences do not use the scaling, but may be easier to finish the tguification of preferences rather than wrestling the byondui.

Drulikar:
For fireteam scaling issue: 
[patch.diff.txt](https://github.com/user-attachments/files/19221282/patch.diff.txt)

DarkLordCabbage:
It's fixed!!! Thank you!!!!!!!!!

PR #8768, opened 2025-03-15
mar-50 lmgification

BOBAMAx:
# About the pull request

Reduces the mar-50's ROF and increases scatter and makes the bipod a starting attachment

Makes the Mar-50 LMG still effective while used defensively with the bipod while making it less effective on the move with the bipod attached. You can still remove the bipod to make it more effective offensively. higher damage per shot than the HPR with a lower ROF. Primarily designed around CLF inserts and the odd ERT that gets one.

<details>

</details>

# Changelog
:cl:
balance: Mar-50 is now more like the HPR
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -868,7 +868,7 @@
 	reload_sound = 'sound/weapons/handling/gun_mar40_reload.ogg'
 	unload_sound = 'sound/weapons/handling/gun_mar40_unload.ogg'
 
-	starting_attachment_types = list(/obj/item/attachable/mar50barrel)
+	starting_attachment_types = list(/obj/item/attachable/mar50barrel, /obj/item/attachable/bipod)
 
 	current_mag = /obj/item/ammo_magazine/rifle/mar40/lmg
 	attachable_allowed = list(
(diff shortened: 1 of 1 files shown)


Git-Nivrak:
This will have the same fire-rate as HPR since the bipod reduces the fire rate, combine that with its higher damage and already decent accuracy, I don't think this should be available by default

BOBAMAx:
> This will have the same fire-rate as HPR since the bipod reduces the fire rate, combine that with its higher damage and already decent accuracy, I don't think this should be available by default

https://github.com/user-attachments/assets/29f3c032-00aa-4bb7-9852-288198071817

this is how the gun works on master right now.

https://github.com/user-attachments/assets/e00ff8d7-69bf-4981-8459-0d90a64a8c04

this is how it works in my pr. I don't understand what you mean by "this will have the same fire-rate as the HPR" because my PR makes the mar-50 have HALF the ROF of the HPR when using a bipod specifically because it does more damage. I'd appreciate it if you could re-open my PR so that survivors don't have access to deathsquad tier weapons anymore thanks.

Git-Nivrak:
If that's your concern you can do this with only the firerate change, leave the scatter and bipod out

PR #8774, opened 2025-03-16
weather cleans up floor cleanables

hry-gh:
like blood and guts and whatnot

:cl:
add: weather now cleans up blood and other cleanables from the floor
/:cl:

video:
https://files.catbox.moe/pklbmt.mp4

### DIFF
--- a/code/controllers/subsystem/weather.dm
+++ b/code/controllers/subsystem/weather.dm
@@ -1,9 +1,9 @@
 GLOBAL_LIST_EMPTY(weather_notify_objects)
 
 SUBSYSTEM_DEF(weather)
-	name   = "Weather"
-	wait   = 5 SECONDS
-	priority   = SS_PRIORITY_LIGHTING
+	name = "Weather"
+	wait = 5 SECONDS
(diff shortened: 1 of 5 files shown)


chefoxara:
less lag very good i approv.

Drulikar:
I assume you intended to do something about existing map decals per https://discord.com/channels/150315577943130112/745447048261795890/1359157867763990680? I'm not sure a good approach for this other than I dunno certain types or all when maploaded unregister themselves from weather list?

hry-gh:
> I assume you intended to do something about existing map decals per [discord.com/channels/150315577943130112/745447048261795890/1359157867763990680](https://discord.com/channels/150315577943130112/745447048261795890/1359157867763990680)? I'm not sure a good approach for this other than I dunno certain types or all when maploaded unregister themselves from weather list?

im fine with decals that mappers have placed that are cleanables outside being cleaned up. if mappers don't want them to be cleaned up, they should not be a cleanable. however, pre-first drop, they won't be cleaned up by weather effects

Drulikar (approved):

PR #8778, opened 2025-03-16
verbs were being used like proc in xenomorph init

4soup:
# About the pull request
turns **verbs into procs** inside of xenomorph initialization.
verifies **user** instead of using **usr**.
fixes hud rendering error
verbs were being used like procs in xenomorph init.

# Changelog
:cl:
fix: fixed a few things
/:cl:


Strobia:
🐷🐖🐽🐗

lanse12:
lgtm thanks for the picture

Drulikar (changes requested):
This PR is dead in the water until the description is amended. Not going to squash that into the repo. Do it again and you'll just land a gitban.

As far as calling verbs as procs - don't really care. They are procs with just additional features. So please elaborate what even is being fixed.

4soup:
**"fixes hud rendering error."**
Verbs are assumed to have valid `usr` and so calling them in places where they might not have one causes issues. 
In the xenomorph initialization procedure, a verb (`toggle_xeno_hostilehud)` can be called without valid usr (`usr = null`). 
If you spawned a groid and there was no player attached, toggle_xeno_hostilehud would crash with **`cannot read null.client`** because **`usr = null.`** 
**This fixes a crash inside xenomorph hud rendering without a player**

Drulikar:
PR description was not amended to remove weird image. They can reopen if they want after gitban ends.

PR #8827, opened 2025-03-22
Antiparasitic buff

JackieEstegado:
# About the pull request

We lower the burn damage antiparasitic does per tick by 25%.

Since the APS fix 3 weeks ago I played dozens of hours of researcher. I never once saw another researcher that knew about the burn damage even attempt to make an APS chem for marines and even I myself rarely felt it was worthwhile the investment of credits.
The reason for this is that multiple problems stack together:
1. The amount burn of damage it creates is very hard to counter
2. To cure a latestage patient it requires at minimum APS4 with 30u of the chem (Problem with this is that most medics wont have the requisite knowledge to know when a patient is cured + a medic cant be expected to babysit a marine for several minutes to keep injecting more of the chemical and treat burns. Additionally it decreases the viability of it being creatable solely via relating, since 30u+ OD chems are rare).
3. Due to the above two problems the researcher has to figure out how to deal with the burn damage. (Math for this will be below) There are two ways to handle it, one is to include traditional healing chemicals in the pill/autoinjector that the APS will be in, however the problem is that at level 4 even a mix of kelotane, tricord and dermaline (which by the way restricts the APS in one pill/injector be maximum 15u due to it's OD of 15) wont completetly negate the burn damage.
Or the researcher has to include some levels of ACR in the chem itself, which increases credit costs, and further decreases the chances of being able to make such a chem to near zero without the use of CREATE which will be even more expensive.
4. Including multiple chems to counter the burn makes mixing and distributing the mix very slow and error prone, prettymuch requiring a reseacher fully dedicating his time to produce it.

<details>
<summary>Math</summary>

All numbers included are of 1 in-game tick.

Rate of metabolism of Kelotane/Dermaline/Tric/Research chems: 0.2

Burn damage per levels of APS: 1
Burn heal per levels of ACR: 0.5 per level + (if above ACR4) additional 0.25 per level
Heals: Kelotane = 1 Dermaline = 1.5 Tricordrazine = 0.5

Rate of progression of embryo (When not in stasis/nest/etc): 2
Rate of regression by APS per level: 1 (This means that to pause the progression of embryo we need APS2)
There are (effectively) 5 different stages (0-4) that the embryo goes through each needing 90 progress to advance to the next.
When the embryo reaches stage 4's end and advances it will become a larva and be ready to burst (total of 450 progress needed).
The APS regresses it to the previous stage when it reaches 0 progress in that stage.
When progress 0 is reached at stage 0 OR 1 (this effectively means we need to cure 90 progress less) the embryo will be aborted and the patient cured.

Now lets take an example case of a patient that just entered stage 4, to cure them from this point we need to regress the embryo by 270 progress. With an APS level of 4 we will be regressing by 2 each tick, meaning we will need a total of (Amount of ProgressToRegress\/RegressRate\*MetabolismRate) 270\/2\*0.2 = 27u of the APS chem to cure the patient.
However while that curing happens it will be dealing burn damage, precisely 4 per tick, we can mostly offset it by administering kelotane, dermaline and tricord, which will come out to 3.0 heal per tick, meaning a net 1.0 of damage will be dealt per tick untill the patient is cured. However using dermaline has a downside, it has an OD limit of 15u. Meaning if we want to make a pill/injector mix that offsets our damage we will have to restrict our dose of all the chems to 15u to prevent a dermaline OD (thus increasing the amount of babysitting work a medic has to do).

Another way to deal with the burn damage is by including ACR levels in our chem, just 2 levels will allow us to make a mix that completetly offsets the burn damage, taking work off the medic, however they would still have to apply two doses to fully cure.
To fix that we will need to remove the need for dermaline out of the mix, which will require having ACR5, thus we now would only need a mix of the APR-ACR chem and kelotane. This will finally allow the medic to not have to babysit the marine, HOWEVER this mix would still have the very large problem that if the patient was to have kelotane in them before administering that it will OD them, you would think that this means that the medic just has to wait for it to metabolize first, but from my experience most medics will be oblivious to the danger and just proceed to OD the patient, often fatally.

To remove the risk of these sorts of ODs we need to make the chem have enough ACR to fully offset the burn by itself, which would be an ACR of 6.

Credit cost of CREATEing mentioned APR chems by traditional method of setting complexity to all rare, and using relates to make the recipe viable/easy (thus also requiring OD level to be increased by 15):
(The "+" sign means atleast 3 relates, meaning atleast 3 more credits, but depending on luck it can be WAY more)
APS4 (with final OD of 15): 10+ credits
APS4+ACR2: 20+ credits
(From here on consider that we also require clearance level 3 to be able to create properties above lvl 4 which will cost us 9 credits if we dont have it already)
APS4+ACR5: 23+ credits
APS4+ACR6: 25+ credits

(For reference a good bonemend costs about 3-6 credits to make via relate, 5+ via use of CREATE, and will be of much more use than Antipara)

So if we want to make an Antipara chem that will not be a huge amount of work both for the medics and the researchers (mixing proper ratios of multiple chemicals is prettymuch a fulltime job) _**we will have to spend about half of what the usual round winning stims costs.**_

With the change made in this PR the burn damage will be lowered by 25%, thus allowing the researchers the two options of either making a mostly viable APS for cheap, or spending some more credits (but much less than what would be needed right now) for a great one.

</details>

<details>

</details>

# Changelog
:cl:
balance: Lowered the burn damage of the Antiparasitic chemical property by 25%.
/:cl:


kiVts (changes requested):
APS 4 will be outhealed with KD mix by 0.5, level 5 will be matched with KD mix, level 6 will deal 0 burn if you have KD+tric

half is too much, make it a 50% increase.

code/modules/reagents/chemistry_properties/prop_positive.dm
```diff
@@ -492,7 +492,7 @@
 		if(embryo && istype(embryo))
 			if(embryo.counter > 0)
 				embryo.counter = embryo.counter - (potency * delta_time)
-				current_human.take_limb_damage(0, potency)
+				current_human.take_limb_damage(0, POTENCY_MULTIPLIER_MEDIUMLOW*potency)
```
kiVts, line 495: 1.5 not 0.75

kiVts (changes requested):

code/modules/reagents/chemistry_properties/prop_positive.dm
JackieEstegado: I am dum dum :|

kiVts (approved):

PR #8840, opened 2025-03-23
Allows Drones to evolve into any T2 before first drop

silverplatedelta:
# About the pull request

This PR allows drones to evolve into any T2 caste (now allowing them to evolve into spitter, warrior, lurker) before the marines first drop.

As discussed earlier on the forums, as well as in coding chat I feel that this could eliminate risk of missing t2/t3 evo as a result of paying drone tax. Players wanting to evo into a combat caste can provide building/support to the hive for a little while they wait, as opposed to just sitting in the hive or throwing themselves at survs

While I don't think this will solve the issue of xenos chronically needing builders in the early game, I do feel its a step in the right direction, and wont make early game building be as punishing to players that also wish to partake in combat later in the round.

Code complies,

Became drone, evolved into all possible castes. Became sent, evolved into spitter, devo'd back into sent. Became drone, set first drop variable to true. Unable to evo into spitter/lurker/warrior.

# Changelog
:cl:
balance: Drones can now evolve into any T2 caste before first drop
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/caste_datum.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/caste_datum.dm
@@ -27,6 +27,8 @@
 	var/list/evolves_to = list()
 	/// what caste or castes to de-evolve to.
 	var/list/deevolves_to = list()
+	///This is where you add castes drones can evolve too before first drop.
+	var/list/early_evolves_to
 	///If they can use consoles, etc. Set on Queen
 	var/is_intelligent = 0
 	var/caste_desc = null
--- a/code/modules/mob/living/carbon/xenomorph/castes/caste_datum.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/caste_datum.dm
@@ -27,7 +27,7 @@
 	var/list/evolves_to = list()
 	/// what caste or castes to de-evolve to.
 	var/list/deevolves_to = list()
-	///This is where you add castes drones can evolve too before first drop.
+	///This is where you add castes drones can evolve to before first drop.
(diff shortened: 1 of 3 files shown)


sargashC:
I don't know how I feel about them having the ability to always evolve into every caste, but another discussed was allow it up until the first drop, afterwards they still need to devolve to larva. But before then it seems sort of a no brainer.

Willzadl:
I don't like the notion of backliner drones building their own nest, then evolving straight to lurker. I think the original idea to have it only be pre-15 was better, since this is an actual buff instead of the solution to drone tax it was meant to be.

thymol-blue:
I think prior first drop is perfect like others are saying is perfect, what stops me from paying drone tax is literally the fear of losing out on the T2 slots because I didn't devolve in time.

This is a really great change!

silverplatedelta:
Thanks for the feedback guys! Im going to change it here shortly to only occur before the first drop.

code/modules/mob/living/carbon/xenomorph/castes/caste_datum.dm
```diff
@@ -27,6 +27,8 @@
 	var/list/evolves_to = list()
 	/// what caste or castes to de-evolve to.
 	var/list/deevolves_to = list()
+	///This is where you add castes drones can evolve too before first drop.
```
private-tristan, line 30: to, not too.

silverplatedelta: guh. thanks!

Detective-Google (approved):
code, basic testing

PR #8841, opened 2025-03-23
Chem property fixes

JackieEstegado:
# About the pull request

Defibrillation property:
-Notification of the person being revived was in a branch, causing it to often never get reached. This caused marines to get revived without them ever realizing it. Solution: Reviving and the notification are in one branch now
-Kept triggering healing and sound notification even after the mob was above the dead treshold, causing them to often get healed to full health, which is not intended. Solution: Once the revive timer is underway it will not attempt to do anything anymore
-If the mob got damaged to below the dead threshold while the revive timer was already underway it would still revive them, causing them to inmidietly die again. Solution: If the mob gets damaged to below threshold again we just cancel the revive, and start the whole process over.

Paining would never apply pain because it was expecting a TRUE return from a proc that didnt return anything ಠ_ಠ

Photosensitive's process proc was typoed, so it never got triggered.

Xeno benefiting bugs bad

<details>

</details>

# Changelog
:cl:
fix: Defibrillation chems now always notify the player that they are being revived
fix: Defibrillation property no longer keeps healing the patient once its above -100 health
fix: Fixed a case where defibrillation property would prematurely revive a patient despite them being below -100 health
fix: Paining chem property now causes pain
fix: Photosensitive chem property is no longer harmless
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry_properties/prop_negative.dm
+++ b/code/modules/reagents/chemistry_properties/prop_negative.dm
@@ -6,7 +6,7 @@
 /datum/chem_property/negative/process(mob/living/M, potency = 1, delta_time)
 	M.last_damage_data = create_cause_data("Harmful substance", holder.last_source_mob?.resolve())
 
-	return TRUE
+	return ..()
 
 /datum/chem_property/negative/can_cause_harm()
 	return TRUE
--- a/code/modules/reagents/chemistry_properties/prop_negative.dm
+++ b/code/modules/reagents/chemistry_properties/prop_negative.dm
@@ -6,6 +6,8 @@
 /datum/chem_property/negative/process(mob/living/M, potency = 1, delta_time)
 	M.last_damage_data = create_cause_data("Harmful substance", holder.last_source_mob?.resolve())
 
+	return TRUE
+
 /datum/chem_property/negative/can_cause_harm()
 	return TRUE
 
--- a/code/modules/reagents/chemistry_properties/prop_positive.dm
+++ b/code/modules/reagents/chemistry_properties/prop_positive.dm
@@ -552,7 +552,7 @@
 	category = PROPERTY_TYPE_REACTANT
 	value = 3
 	cost_penalty = FALSE
-	COOLDOWN_DECLARE(ghost_notif)
+	var/revivetimerid
 
 /datum/chem_property/positive/defibrillating/on_delete(mob/living/M)
 	..()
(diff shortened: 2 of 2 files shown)


code/modules/reagents/chemistry_properties/prop_negative.dm
```diff
@@ -6,6 +6,8 @@
 /datum/chem_property/negative/process(mob/living/M, potency = 1, delta_time)
 	M.last_damage_data = create_cause_data("Harmful substance", holder.last_source_mob?.resolve())
 
+	return TRUE
```
kiVts, line 9: call parent in this proc, dont return true,

code/modules/reagents/chemistry_properties/prop_positive.dm
```diff
@@ -602,15 +602,13 @@
...
-			COOLDOWN_START(src, ghost_notif, 30 SECONDS)
-			playsound_client(ghost.client, 'sound/effects/adminhelp_new.ogg')
-			to_chat(ghost, SPAN_BOLDNOTICE("Your heart is struggling to pump! There is a chance you might get up!(Verbs -> Ghost -> Re-enter corpse, or <a href='byond://?src=\ref[ghost];reentercorpse=1'>click here!</a>)"))
+	if(!COOLDOWN_FINISHED(src, ghost_notif))
+		return
+	var/mob/dead/observer/ghost = dead.get_ghost()
+	if(ghost?.client)
+		COOLDOWN_START(src, ghost_notif, 30 SECONDS)
+		playsound_client(ghost.client, 'sound/effects/adminhelp_new.ogg')
+		to_chat(ghost, SPAN_BOLDNOTICE("Your heart is struggling to pump! There is a chance you might get up!(Verbs -> Ghost -> Re-enter corpse, or <a href='byond://?src=\ref[ghost];reentercorpse=1'>click here!</a>)"))
```
kiVts, line 611: this will send a ghost notification everytime DFB heals on failed defib. Move this to the line where its actually reviving them.

kiVts (changes requested):

code/modules/reagents/chemistry_properties/prop_positive.dm
JackieEstegado: Overhauled the whole process. There is now only one notification per actual revive attempt. Kept notification cooldown in due to several rapid revive timer resets being a possibility if the person keeps getting damaged.

code/modules/reagents/chemistry_properties/prop_positive.dm
```diff
@@ -580,17 +581,33 @@
...
 	var/mob/living/carbon/human/dead = M
 	var/revivable = dead.check_tod() && dead.is_revivable()
+	if(!revivable)
+		return
+
+	if(revivetimerid)
+		if(dead.health <= HEALTH_THRESHOLD_DEAD) //If the mob got damaged to below the threshold while the timer was ticking then we reset
+			deltimer(revivetimerid)
+			revivetimerid = null
+		return;
```
kiVts, line 591:
```suggestion
		return
```

code/modules/reagents/chemistry_properties/prop_positive.dm
```diff
@@ -580,17 +581,33 @@
 		return
 	var/mob/living/carbon/human/dead = M
```
kiVts, line 582: because you refactored this, you're will have to fix the single letter var

kiVts (dismissed):

Detective-Google (approved):

PR #1626, opened 2022-11-20
Leadership orders shout phrases again

kugamo:
## About The Pull Request

https://gitlab.com/cmdevs/colonial-warfare/-/merge_requests/2806 removed the phrases when someone gave orders, this PR readds them back into the game but without having the orders said over the radio.

https://user-images.githubusercontent.com/16618648/202879931-fc39a020-2f74-42a8-a4ea-a01de1abf068.mp4

## Why It's Good For The Game

These were awesome, its strange behavior to suddenly get a buff without any apparent cause. At least these phrases had a continuance of cause and effect with them, even if their exact meaning isnt up to the player. Really should have been refactored so a phrase could be input instead of just a complete removal if devs wanted to give leaders more autonomy imho. 

For me these phrases grant leaderships roles accessibility to acting the role without any hurdles, and they really add to the battlefield/command atmosphere in a cool and impactful way thats sorely missed by alot of players.

Gladly will give my hard earned GBP for this.

## Changelog

:cl:
add: Readded back shouting phrases when issuing leadership orders but without radio broadcast of them.
/:cl:

PR #8859, opened 2025-03-25
Revives Leadership order phrases, 2025 edition

Nihisohel:
# About the pull request

For those who want to larp like a military officer in the battlefield.
Essentially a revival of https://github.com/cmss13-devs/cmss13/pull/1626 with added lines and as a preference.

Honestly, these were awesome. Anything I would type here, Kugamo typed it first in his PR.
Adds it as a preference for those that don't wanna be loud.
Total of 20 something messages per order was added as well compared to the old 5 per.

Also, thank god I don't need to spend GBP for this unlike Kugamo.

It works. I do hope pick doesn't lag out the server though.

<details>

https://cdn.discordapp.com/attachments/1313663544096587847/1353930842917048480/Desktop_2025.03.25_-_11.11.28.12.mp4?ex=67e371db&is=67e2205b&hm=d0f7b6c97ce4859f6bc4c05f86a9b0c356dab2a851f54410b4c4ac84f6c55112&

</details>

# Changelog
:cl:
add: A preference for spoken leadership orders have been added. Fulfill your milsim larp fantasies with this. Set to on as default.
/:cl:


ihatethisengine:
The problem with the old orders was that they were sent over the comms and therefore did not show up in local/abovehead chat. So yeah this should be good

Mister-moon1:
this is awesome

dawsonkeyes:
if there’s complaints about the relevancy of an order’s speech (like a hold order being used while retreating to better soak up damage) you could instead make it a louder warcry emote? Not that I think it’ll be a problem in the first place; just if someone thinks it might be

Nihisohel:
> if there’s complaints about the relevancy of an order’s speech (like a hold order being used while retreating to better soak up damage) you could instead make it a louder warcry emote? Not that I think it’ll be a problem in the first place; just if someone thinks it might be

I think calling out "HOLD THE LINE, NOT ONE STEP BACK" and then retreating is funny even if unintended

code/modules/client/preferences_toggles.dm
```diff
@@ -381,6 +382,15 @@
 		to_chat(src,SPAN_BOLDNOTICE( "Click-dragging now blocks clicks from going through."))
 	prefs.save_preferences()
 
+/// Toggles whether activating marine leader orders will be spoken or not, on by default
+/client/proc/toggle_leadership_spoken_orders()
+	prefs.toggle_prefs |= TOGGLE_LEADERSHIP_SPOKEN_ORDERS
```
fira, line 387: This only enables, doesn't toggle, you want `^=`

code/modules/mob/living/carbon/human/powers/issue_order.dm
```diff
@@ -50,7 +50,18 @@
 	// 1min cooldown on orders
 	addtimer(CALLBACK(src, PROC_REF(make_aura_available)), COMMAND_ORDER_COOLDOWN)
 
-	visible_message(SPAN_BOLDNOTICE("[src] gives an order to [order]!"), SPAN_BOLDNOTICE("You give an order to [order]!"))
+	if(src.client?.prefs && (src.client?.prefs?.toggle_prefs & TOGGLE_LEADERSHIP_SPOKEN_ORDERS))
```
fira, line 53: don't need to double it up, the second check already does the first

fira (changes requested):

Detective-Google (approved):

fira (approved):
Code LGTM but I'd really like someone else to doublecheck the savfile migration bit

Detective-Google (approved):
save file looks (hopefully) correct.

PR #8874, opened 2025-03-26
ignore mouse 4/5

### LINKED ISSUE #8873: Mouse 4 and 5 inputs effectively count as a mouse 1 input
### Testmerges

N/A

### Round ID

N/A

### Description of the bug

When using a mouse that has more than three inputs, the additional buttons all act as mouse 1. This can be an issue if you have other things bound to said buttons resulting in friendly fire and accidental selection of items on menus.

It should not, quite simply it should not do the same thing as mouse 1, or at least be rebind-able to have no input on that button

1. have a mouse with more than 3 buttons
2. press said buttons
3. friendly fire/ buy things you didn't want to in venders.

### Issue Bingo

hry-gh:
516 finally implemented mouse 4/5 correctly, but we assume that every click is always this

closes #8873

eventually we should make this rebindable to stuff but uhh not today man

:cl:
fix: mouse 4/5 presses no longer count as mouse 1 presses
/:cl:

### DIFF
--- a/code/__DEFINES/_click.dm
+++ b/code/__DEFINES/_click.dm
@@ -10,6 +10,9 @@
 ///if(modifiers[BUTTON] == LEFT_CLICK)
 #define BUTTON "button"
 
+#define BUTTON4 "xbutton1"
+#define BUTTON5 "xbutton2"
+
 //Keys held down during the mouse action
 #define CTRL_CLICK "ctrl"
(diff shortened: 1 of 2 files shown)


Drulikar (approved):

XxSWAG-MASTER-OVERLORDxX:
This would have taken effect by now, correct? if so, this has not worked, and the issue persists

PR #8875, opened 2025-03-26
stops any weird audio player crashes

hry-gh:
:cl:
fix: the audio player shouldn't blow up in weird circumstances
/:cl:


Drulikar (approved):
Dunno if this would fix their issue but doesn't hurt.

hry-gh:
> Dunno if this would fix their issue but doesn't hurt.

not the underlying cause, but the bluescreen is caused by an uncaught error within this promise, which we're catching now

PR #8879, opened 2025-03-26
Small non-underground area in the LV lake house

DoguhanOzgurAkca:
# About the pull request

This change adds a small 5x5 area in the LV north central caves. It has no roof, and in the middle of it is a broken comms tower. 

This is my first proper PR for byond, SS13, and of course, first time doing anything map-related. It seemed to work well when I was testing it, though. I would be open to ideas about what else to place in this little area, as I was unsure. It is meant to be a broken comms relay of sorts. 

The reason why I am making this change is because I have noticed that the LV is the only map in the game where the sensor tower is neither in an open area nor close to one. It's pretty deep. I have also enjoyed non-underground areas very deep in the map quite a bit in other maps such as Hybrisa. It shakes things up a bit honestly. 

<details>

Screenshot 2025-03-26 at 21 45 42

</details>

# Changelog
:cl:
add: Addeded a new area to LV lake house area that isn't underground.
/:cl:


DoguhanOzgurAkca:
Screenshot 2025-03-27 at 01 20 45

Updated how it looks just a bit. If anyone has ideas, especially in terms of adding a very small light to it, I am still very much open to it.

dawsonkeyes:
Is there any better of a way to indicate there's no cave above the area? I feel like only people that are in-the-know would be able to understand that it's an open zone

Steelpoint:
There is currently no real way to discern an open area from a non-open area side from directly examining the tile, so this being a 5x5 open area is not going to be noticeable to anyone on first glance.

DoguhanOzgurAkca:
> Is there any better of a way to indicate there's no cave above the area? I feel like only people that are in-the-know would be able to understand that it's an open zone

Not really, sadly. I would love to have some kind of moonlight shining through the hole in the cave but well, that does not exist, for now at least.

Steelpoint:
> > Is there any better of a way to indicate there's no cave above the area? I feel like only people that are in-the-know would be able to understand that it's an open zone
> 
> Not really, sadly. I would love to have some kind of moonlight shining through the hole in the cave but well, that does not exist, for now at least.

Theoretically you can set the area to be somewhat illuminated via using 'base_lighting_alpha = X' . This does present a minor balance consideration since this will light the area up permanently, furthermore it likely won't look very aesthetically pleasing since it illuminates the tiles with no soft edges, so it will be a hard box edge that stands out from the other tiles.

DoguhanOzgurAkca:
So I was testing the tower, and I noticed that it completely blocks movement through its 1x3 area. 
I did it so that only the base actually blocks you (not pushed for the PR), but now the problem is that it can block sprites quite well, and Xenos cannot see through it. But it can be destroyed by C4. It is rather awkward for it to completely block the movement, but being able to hide behind it also sucks.

image


https://github.com/user-attachments/assets/e4fb0106-e7a5-4424-9bbb-bff971fe2da1

ihatethisengine:
Carp lake on sorokyne is also in the open and very few people know about that...

Can you apply weather effect to an area? If it always was raining there it would make it more obvious.

Detective-Google (approved):
code, needs mapping approval

DoguhanOzgurAkca:
> code, needs mapping approval

Ah, I forgot to push the new code. I changed it a bit.

DoguhanOzgurAkca:
Latest version has 3 changes total.
First of all, the non underground area was moved a bit north to be against the cave wall. I removed like 3 tiles of rock to make a bit more way. The corpse is an engineering corpse. I asked a few survivor players I know and they said it wouldn't cause big problems or anything.
Screenshot 2025-04-01 at 23 59 12

Then two more things. One, a little fishing area. I am trying to sprite plastic chairs into the game; when that is done, I plan on replacing the wooden chairs with them. For now, think of them as being borrowed from the underground lakehouse.

Screenshot 2025-04-01 at 23 59 21


Last thing is that sensor tower has some dirt on its tiles now. 

Screenshot 2025-04-01 at 23 59 29

Detective-Google:
image

you have step_x/step_y pixel shifts in your PR, we use pixel_x/pixel_y. make sure to fix that, IIRC there's a strongdmm setting that changes which one it does when you pixel shift things.

Drulikar:
Stale exempt until Nanu has a moment to look.

Nanu308 (approved):
I am fine with this change per now, only thing might be marines trying a paradrop meme at some point so i'll have to keep an 👁️ open

DoguhanOzgurAkca:
> I am fine with this change per now, only thing might be marines trying a paradrop meme at some point so i'll have to keep an 👁️ open

I do think paradrops in very tight non underground parts being not allowed or being risky could be a good change. Either a new roof type or a change to the paradrop system where its like "if the area that is not underground isn't this big, don't allow it"

PR #8890, opened 2025-03-27
Impact ammo changes for scout

Mister-moon1:
# About the pull request
Stun time reduced over distance, slowdown reduced for humans, slowdown removed for xenos, AP values reduced to incendiary levels

Scout is good, very good, if you aren't running up to doors without a clue whats behind it then realistically you can avoid 9/10 fights you don't want to take, add onto this a very powerful gun then it becomes insanely good.

As it stands the solution people give for scouts is have someone near you, only that its a bad faith answer considering the fact scouts can 1v2 depending on castes, this is due to the fact that stun time barely changes for on screen mobs and superslow is applied to those hit, leaving them borderline immobile allowing the scout to just switch between targets and kill both. 

Yes scout should be strong due to the risks they can choose to take, however I feel like they should have to commit more to a fight and be less effective shooting at range to allow those stunlocked to escape should other xenos push scout away.

On paper scout should still be able to 1v1 xenos and have the edge, just that 1 landed hit doesn't ensure victory and they will have to be within closer proximity to the xeno to avoid them escaping.

AP being brought down because this is not a sniper caliber, it does not need 50 AP.

for anyone saying cope pr this is from experience as scout, its stupid as is.
https://imgur.com/Mp6GqMT

<details>

</details>

# Changelog

:cl:
balance: M4RA custom impact ammo no longer slows xenos, stuns reduced over range, AP brought down to that of incendiary ammo
/:cl:

### DIFF
--- a/code/datums/ammo/bullet/rifle.dm
+++ b/code/datums/ammo/bullet/rifle.dm
@@ -179,8 +179,8 @@
 	if(iscarbonsizexeno(living_mob))
 		var/mob/living/carbon/xenomorph/target = living_mob
 		to_chat(target, SPAN_XENODANGER("You are shaken by the sudden impact!"))
-		target.KnockDown(0.5-fired_projectile.distance_travelled/200) // purely for visual effect, noone actually cares
-		target.Stun(0.5-fired_projectile.distance_travelled/200)
+		target.KnockDown(0.5-fired_projectile.distance_travelled/100) // purely for visual effect, noone actually cares
+		target.Stun(0.5-fired_projectile.distance_travelled/100)
 	else
 		if(!isyautja(living_mob)) //Not predators.
 			living_mob.apply_effect(0.5, SUPERSLOW)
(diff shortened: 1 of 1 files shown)


Mister-moon1:
will specify that if you see a lone xeno you can literally click on them and walk them down they still cant do anything about that, this just covers the scenarios where scouts play safe and get rewarded for it

neeshacark:
May or may not be a little overkill to also nuke the AP, I still think we should just remove impact burst fire but I don't think we have existing code to allow some ammo types to burst and not others. Nuking just the AP may also work. I honestly feel it's the combination of fuck-off dps and the stun combined that makes it insane.

Full disclosure 
image

Mister-moon1:
Removing burst fire wont address much of anything other than frontlining capability, it just means that the time it takes to kill your target(s) is longer without much a change at all.

Also realistically you wont be finding anything outside of fortified defenders, buffed valks and the king with enough armour to even see a use for 50 AP

PR #8894, opened 2025-03-28
Accessory fixes & Tactical masks can be worn as accessories

### LINKED ISSUE #9035: Uniforms with webbings can be put into a backpack for extra storage
### Testmerges

N/A

### Round ID

N/A

### Description of the bug

Per the title. If you put a uniform with a webbing into your backpack you can have tons of extra storage.

probably not this much extra storage maybe

1. put a webbing on a uniform
2. put the uniform(s) into your backpack
3. extra storage
...

### Issue Bingo

Nihisohel:
# About the pull request

Title. Allows the wearing of scarves and other masks on your armor or uniform through use item interaction.
Thank you CM Dev.

Adds the functionality to make masks and other clothing items such as hats to be treated as accessories without refactoring the entirety of accessory code.

Currently, you need to use it while in your hand to change it from and to an accessory.

By god, I am not refactoring accessory code.

This means you can smoke something while looking cool or wear another mask under your mask, I guess.

Prevents accessories from attaching to uniforms while its in a storage item as an alternative of #9110
closes #9035

image

# Changelog
:cl:
add: Scarves and some masks can be used as accessory items! Use them inhand to change them back and to an accessory.
fix: Armbands no longer hide when sleeves are cut or jacket is removed
fix: Accessories may no longer be attached while it is in a storage item.
/:cl:


Strobia:
FIRE🔥MERGE THIS NOW

code/modules/clothing/clothing.dm
```diff
@@ -12,14 +12,74 @@
...
 	var/clothing_traits_active = TRUE //are the clothing traits that are applied to the item active (acting on the mob) or not?
 
+	// accessory stuff
+	var/list/accessories
+	var/list/valid_accessory_slots = list()
+	var/list/restricted_accessory_slots = list()
+	// vars below related to converting clothing to accessories
+	var/can_become_accessory = FALSE //can this clothing item be turned into an accessory?
+	var/slot = ACCESSORY_SLOT_DECOR //default slot for accessories, pathed here for use for non-accessories
+	var/accessory_path = "/obj/item/clothing/accessory"
```
fira, line 27: string paths are a no-no. what's this for anyway? i don't see it used anywhere other than directly below

fira (dismissed):

code/modules/clothing/clothing.dm
Nihisohel: in my mind, i figured it would be good for future implementation for clothing turned into accessories that have specific mechanics to it, rather than just pathed generically to /obj/item/clothing/accessory so you could theoretically have a converted accessory inherit that string-path specific mechanics or w/e

code/modules/clothing/clothing.dm
```diff
@@ -12,14 +12,74 @@
...
 	var/list/clothing_traits // Trait modification, lazylist of traits to add/take away, on equipment/drop in the correct slot
 	var/clothing_traits_active = TRUE //are the clothing traits that are applied to the item active (acting on the mob) or not?
 
+	// accessory stuff
+	var/list/accessories
+	var/list/valid_accessory_slots = list()
+	var/list/restricted_accessory_slots = list()
+	// vars below related to converting clothing to accessories
+	var/can_become_accessory = FALSE //can this clothing item be turned into an accessory?
+	var/slot = ACCESSORY_SLOT_DECOR //default slot for accessories, pathed here for use for non-accessories
```
Drulikar, line 26: Refactor this to `accessory_slot` or something. Make the comment on the right of it instead an [autodoc](https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/AUTODOC.md) comment.

code/modules/clothing/clothing.dm
```diff
@@ -12,14 +12,74 @@
...
+	var/accessory_path = /obj/item/clothing/accessory // for pathing to different accessory subtypes with unique mechanics
```
Drulikar, line 27: Make the comment on the right of it instead an [autodoc](https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/AUTODOC.md) comment.

code/modules/clothing/clothing.dm
```diff
@@ -12,14 +12,74 @@
...
+		to_chat(user, SPAN_NOTICE("[src] cannot be reverted because the original item path is missing."))
+		return
+
+	var/obj/item/clothing/original_item = new access.original_item_path(loc)
+	if(!original_item)
+		to_chat(user, SPAN_NOTICE("Failed to revert [src] to its original item."))
+		return
+
+	if(ismob(loc) && loc == user)
+		user.put_in_hands(original_item)
```
Drulikar, line 73: This never occurs because `user` is null because `revert_from_accessory` never passes the user as an argument.

code/modules/clothing/under/ties.dm
```diff
@@ -20,6 +19,11 @@
 	var/removable = TRUE
 	flags_equip_slot = SLOT_ACCESSORY
 	sprite_sheets = list(SPECIES_MONKEY = 'icons/mob/humans/species/monkeys/onmob/ties_monkey.dmi')
+	var/original_item_path = /obj/item/clothing/accessory
+
+/obj/item/clothing/accessory/attack_self(mob/user)
+	if(can_become_accessory)
+		revert_from_accessory()
```
Drulikar, line 26: `user` is not passed in `revert_from_accessory`

code/modules/clothing/masks/breath.dm
```diff
@@ -322,6 +326,8 @@
 	flags_inventory = ALLOWCPR
 	flags_cold_protection = BODY_FLAG_HEAD
 	min_cold_protection_temperature = ICE_PLANET_MIN_COLD_PROT
+	can_become_accessory = TRUE
+	slot = ACCESSORY_SLOT_MASK
```
Drulikar, line 330: This is already inherited from `/obj/item/clothing/mask/tornscarf` so should be removed since this doesn't do anything nor does every kind specify it.

Drulikar (changes requested):

code/modules/clothing/clothing.dm
```diff
@@ -12,14 +12,74 @@
...
+	// we dont want duplicates man
+	qdel(src)
+
+/obj/item/clothing/proc/revert_from_accessory(mob/user)
+	var/obj/item/clothing/accessory/access = src
+	if(!access.original_item_path)
+		to_chat(user, SPAN_NOTICE("[src] cannot be reverted because the original item path is missing."))
+		return
+
+	var/obj/item/clothing/original_item = new access.original_item_path(loc)
```
Drulikar, line 67: Rather than loc, you should use get_turf(user) or get_turf(loc) so it can be on the floor if their hands or full rather than inside the mob's contents inaccessible.

Nihisohel: passing user for put_in_hands already defaults the item to the floor if both hands are full

Drulikar (changes requested):
Two remaining issues (one unresolved from previous review). Otherwise ready.

code/modules/clothing/clothing_accessories.dm
```diff
@@ -54,6 +54,10 @@
 			to_chat(usr, SPAN_WARNING("You cannot attach accessories of any kind to \the [src]."))
 			return
 
+		if(isstorage(src.loc))
+			to_chat(usr, SPAN_WARNING("You cannot attach accessories to [src] while it is in a storage item."))
```
Drulikar, line 58:
```suggestion
			to_chat(user, SPAN_WARNING("You cannot attach accessories of any kind to [src]."))
			return

		if(isstorage(loc))
			to_chat(user, SPAN_WARNING("You cannot attach accessories to [src] while it is in a storage item."))
```

Drulikar (changes requested):

PR #8898, opened 2025-03-28
Latejoins now keep the same ratio of marines to xenomorphs

Git-Nivrak:
How latejoining currently works is that every "combat" role is worth 1 latejoin and every "shipside" role is worth 0.25 latejoins,
Before 20 minutes every 4 latejoins give 1 larva and after that it is 2.5 latejoins for 1 larva.

The problem: Over the course of long rounds latejoins can tip the scales numbers wise significantly and turn either side from losing to winning

The solution: Making latejoins keep the same ratio of marines to xenos so if there are more xenos compared to marines they will get more larvae and vice versa (This keeps the static 1:4 ratio before 20 mins and 0.25 value of shipside roles)
<details>

</details>

# Changelog
:cl:
balance: Latejoins now keep the same ratio of marines to xenomorphs instead of being a static 2.5
/:cl:

### DIFF
--- a/code/modules/mob/new_player/new_player.dm
+++ b/code/modules/mob/new_player/new_player.dm
@@ -179,7 +179,6 @@
 
 	if(latejoin_larva_drop && SSticker.mode.latejoin_tally - SSticker.mode.latejoin_larva_used >= latejoin_larva_drop)
 		SSticker.mode.latejoin_larva_used += latejoin_larva_drop
-		var/datum/hive_status/hive
 		for(var/hivenumber in GLOB.hive_datum)
 			hive = GLOB.hive_datum[hivenumber]
 			if(hive.latejoin_burrowed == TRUE)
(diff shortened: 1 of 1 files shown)


Drulikar:
I'm a bit skeptical if this will be desirable. It looks like it will just lock in whoever is winning at the 20 minute mark. Ensuring the game snowballs is fine eventually, but doing so at the 20 minute mark is likely too early.

PhantornRU:
> I'm a bit skeptical if this will be desirable. It looks like it will just lock in whoever is winning at the 20 minute mark. Ensuring the game snowballs is fine eventually, but doing so at the 20 minute mark is likely too early.

I've often seen situations where a lot of latejoined Marines simply swarm the Xenomorphs, even though the Xenomorphs had all the advantages before.

SomeoneYouProbablyKnow:
I gotta be honest my concern isnt with what faction wins, i don't care, im here for fun stuff. My worry is this system will make actual rounds less interesting to be in. If you join a faction now it's even more likely they've already won and you won't get to do jack.
I'd gladly sacrifice global winrate if it means more interesting rounds for me

cm13-github:

<details><summary>Test Merge Deployed @ 04/02/2025 15:00:14 +00:00:</summary>

##### Server Instance

##### Revision
Origin: 76bbebc9083bcc8eed201c10fcae01443fa3c731
Pull Request: e69536beec808e2d5a7647541e06b23a42b6247f
Server: 42bb0f26c1efa7473a3651b897d7f1d75e996a94
</details>

Git-Nivrak:
I don't think this is working, going to try a different approach

PR #8905, opened 2025-03-29
Attempts to fix permanent zoom, 2025 edition

Nihisohel:
# About the pull request

Title. 
Honest to god I really shouldn't have wasted my time on this niche ass bug.

Rather than using signals like what morrow attempted, checks instead for zoom in the destroy proc and attempts to unzoom the user before qdeleting.

user.interactee is still gonna return null but I think thats gonna be a permanent problem.

Also fixes binocular throw range, while funny, is still a bug.
Would not want to throw my binoculars to a group of bugs 1 screen away.

By God I hope it works.

I've looped around 200 HE OBs trying to replicate this bug.

</details>

# Changelog

:cl:
fix: Attempts to fix permanent zoom bugs once more.
fix: Fixes binocular throw range.
/:cl:

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -201,6 +201,11 @@
 	if(ismob(loc))
 		dropped(loc)
 
+	if(zoom)
+		for(var/mob/living/user in viewers(src, GLOB.world_view_size))
+			if(user.client && user.interactee == src) // user.interactee is necessary here to prevent everyone else zooming in from being unzoomed
+				unzoom(user)
+
 	return ..()
 
 /obj/item/ex_act(severity, explosion_direction)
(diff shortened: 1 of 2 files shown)


code/game/objects/items.dm
```diff
@@ -201,6 +201,11 @@
 	if(ismob(loc))
 		dropped(loc)
 
+	if(zoom)
+		for(var/mob/living/user in viewers(src, GLOB.world_view_size))
+			if(user.client && user.interactee == src) // user.interactee is necessary here to prevent everyone else zooming in from being unzoomed
+				unzoom(user)
+
```
Detective-Google, line 208: like harry said in discord, a for in viewers loop in Destroy(already super expensive) is a no go

fira: it's also a super backwards hack that just shows interactee relationship is not correct 😅

Nihisohel: yeah i have another fix in mind that ill commit when i get to coding it

PR #8906, opened 2025-03-29
[CONFIG] Block nuclear device invocation until marines still alive.

PhantornRU:
# About the pull request
Sometimes there are situations where there are enough combat marines on the ground and they decide to call in a nuke too early instead of finishing off the xenomorphs. We now stop the command in such cases so that they lead the marines instead of wanting to abruptly end the game.

Through the config you can control what percentage of marines you want in the game.

# Test
Was tested on BandaMarines (https://github.com/ss220club/BandaMarines/pull/242).
Works fine. 

# Changelog

:cl:
config: Add to Config - Block nuclear device invocation until more marines than this percentage are alive. (0-100)
/:cl:

### DIFF
--- a/code/modules/cm_tech/techs/marine/tier4/nuke.dm
+++ b/code/modules/cm_tech/techs/marine/tier4/nuke.dm
@@ -44,7 +44,7 @@
 		var/marines_count = SSticker.mode.count_marines() // Counting marines on land and on the ship
 		var/marines_peak = GLOB.peak_humans * nuclear_lock / 100
 		if(marines_count >= marines_peak)
-			to_chat(user, SPAN_WARNING("You cannot purchase this while there are more than [marines_peak] Marines and USCM crew alive on this operation."))
+			to_chat(unlocking_mob, SPAN_WARNING("You cannot purchase this while there are more than [marines_peak] Marines and USCM crew alive on this operation."))
 			return FALSE
 
 	return TRUE
--- a/code/modules/cm_tech/techs/marine/tier4/nuke.dm
+++ b/code/modules/cm_tech/techs/marine/tier4/nuke.dm
@@ -44,7 +44,7 @@
 		var/marines_count = SSticker.mode.count_marines() // Counting marines on land and on the ship
 		var/marines_peak = GLOB.peak_humans * nuclear_lock / 100
 		if(marines_count >= marines_peak)
-			to_chat(unlocking_mob, SPAN_WARNING("You cannot purchase this while there are more than [marines_peak] Marines and USCM crew alive on this operation."))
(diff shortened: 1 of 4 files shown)


Aurallianz:
I don't know if this is a good idea.
This situation likely leads to a more prolonged stalemate, especially since games that involve a nuclear option already take around two hours. 
The only potential change regarding the nuclear option is restricting the Marines from using the nuke if there are fewer than 2-3 Xenos on the map. 
However, this change WILL slightly undermine the player's agency and somewhat diminish the efforts of the Intelligence Officers.

If xenos allow the marines to hold BOTH comms towers for the required duration, xenos DESERVE to lose.

config/example/game_options.txt
```diff
@@ -58,3 +58,6 @@ ANIMAL_DELAY 0
 
 ## Remove the # in front of this config option to have loyalty implants spawn by default on your server.
 #USE_LOYALTY_IMPLANTS
+
+## Block nuclear device invocation until more marines than this percentage are alive. (0-100)
+nuclear_lock_marines_percentage
```
fira, line 63: need a value

code/modules/cm_tech/techs/marine/tier4/nuke.dm
```diff
@@ -39,6 +39,14 @@
 		to_chat(unlocking_mob, SPAN_WARNING("You cannot purchase this node before [ceil((NUKE_UNLOCK_TIME + SSticker.round_start_time) / (1 MINUTES))] minutes into the operation."))
 		return FALSE
 
+	var/nuclear_lock = CONFIG_GET(number/nuclear_lock_marines_percentage)
+	if(nuclear_lock > 0 && nuclear_lock != 100)
+		var/marines_count = SSticker.mode.count_marines() // Counting marines on land and on the ship
+		var/marines_peak = GLOB.peak_humans * nuclear_lock / 100
+		if(marines_count >= marines_peak)
+			to_chat(unlocking_mob, SPAN_WARNING("You cannot purchase this while there are more than [marines_peak] Marines and USCM crew alive on this operation."))
```
fira, line 47: i think there's no point to expose the real internal value to players. the config % (or nothing) would be enouhg

config/example/game_options.txt
PhantornRU:
```suggestion
nuclear_lock_marines_percentage 50
```

code/modules/cm_tech/techs/marine/tier4/nuke.dm
PhantornRU, line 47:
```suggestion
			to_chat(unlocking_mob, SPAN_WARNING("You cannot purchase this while there are more than [nuclear_lock]% Marines and USCM crew alive on this operation."))
```

Git-Nivrak (approved):

PR #8373, opened 2025-02-06
Alternative Surv Nerf.

Detective-Google:
# About the pull request

This PR is an alternative to #8303, and uses some code from that PR.

The main difference between this PR and that one is:

Wherever a weapon was removed, it was replaced with a new, less strong weapon.
Wherever a weapon was removed, so was it's ammo.
The flamethrowers on all maps were not removed
The IMS' on all maps were not removed
Booze vendors now spawn with significantly less bottles in them(uses the same system as weymeds.)
Removed some OP gear that nivrak missed.
Didn't touch the out of rotation maps.

I spent a pretty long time testing various weapons that are available to survs, and went through each (in rotation) map and replaced OP/broken weapons with weapons that are good, but not better than their marine equivalent.

also it's always bothered me that solaris has no dead WY management since they are a valuable RP prop, so now one spawns in admin.

# Explain why it's good for the game

Survs have a lot of dumb OP stuff that encourages people to roll it to snag the loot and be LRP pfc+s once the round begins proper, this is not desirable, however I believe #8303 made some changes that were overall detrimental to the experience and removed things that did not warrant removal, such as flamethrowers, which provide a valuable niche to deterring certain types of xenomorphs, and come with limits that in my opinion make it balanced, as well as removing items a-la the IMS, which are generally harmless in terms of direct gameplay impact, are available to marines via research, and in my opinion offer a "cool" factor to exploring the maps. I told the author as much but I suspect they were so inundated with people pinging them with cope they simply didn't parse it, which is understandable, so I'm making an alternate PR.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl: nivrak, detectivegoogle
del: removed thermals and other OP items from all maps.
balance: booze vendors spawn less bottles, molotov fire is weaker
maptweak: certain OP weapons on various maps were replaced with more modest counterparts.
maptweak: solaris now has a dead liaison spawner on it.
/:cl:

PR #8910, opened 2025-03-29
colony boozeomats no longer spawn full

Detective-Google:
# About the pull request

colony side boozeomats now use the same system as weymeds to randomly unstock some of their contents

part of the atomization of #8373 

part of the reason molotov spam is so bad right now is that for flavor reasons we have like 90000 types of booze in the vendor each with a few bottles, reducing the count of bottles in the vendor reduces the amount of molotovs that can be made and nudges survivors towards thinking about when and where to use them.

<details>

image

</details>

# Changelog
:cl:
balance: colony boozeomats are no longer always full
/:cl:


fira (approved):

Dutch-VanDerLinde:
@Detective-Google It says in the note that the amount selected is between 1 and the original amount, yet in the actual code it says 1 and 3? Is that intended? Is the note just wrong?

PR #8940, opened 2025-03-31
 fix: fix concurrent map reading of same map files

fira:
# About the pull request

Mirror PR from PVE : https://github.com/cmss13-devs/cmss13-pve/pull/867

Basically loading the same template several times in parallel causes BYOND to just carry over Regex cursor between readings, messing up the file reading. This seems to fix it.

Less chance of future issues

No runtimes when loading Otogi on PVE

# Changelog
:cl:
fix: Fix an issue when concurrently loading identical maps.
/:cl:

### DIFF
--- a/code/modules/mapping/reader.dm
+++ b/code/modules/mapping/reader.dm
@@ -207,6 +207,7 @@
 	var/stored_index = 1
 	var/list/regexOutput
 	//multiz lool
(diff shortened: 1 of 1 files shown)


Aurallianz:
Yo solid!! I had this happen once when messing around with stuff thanks fira, this appears to work ok!

Drulikar (approved):
As far as I can tell there have been no issues from this change.

PR #8944, opened 2025-04-01
[FIX] Buckled mob no longer gets bugged when walking on a trap

PhantornRU:
# About the pull request

There was a situation where if a MOB was riding a wheelchair, a Souto-mobile or any other "bed" based ride (like a wheeled chair or a medical wheelchair), then riding it would trigger a bug.

The bug is that you continue to control the vehicle (not the multiz vehicle), even if the trap has hooked you.
Now getting into a "hole" throws you off and punishes you.

Fixes are good.
No more abuse on strollers and chairs.

<details>
Fixed:

image

image

</details>

# Changelog

:cl:
fix: Traps now activate correctly on buckled_mob
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/trap.dm
+++ b/code/modules/cm_aliens/structures/trap.dm
@@ -344,6 +344,16 @@
 	..()
 
 /obj/effect/alien/resin/trap/Crossed(atom/A)
+	if((isStructure(A) && istype(A, /obj/structure/bed)) || (isVehicle(A) && !isVehicleMultitile(A)))
+		if(O.buckled_mob)
+			var/mob/living/M = O.buckled_mob
+			O.unbuckle()
+			M.forceMove(get_turf(O))
+			HasProximity(M)
+			to_chat(M, SPAN_XENOHIGHDANGER("You've fallen into a pit full of resin!"))
+	if(trap_type == RESIN_TRAP_EMPTY)
+		return
+
 	if(ismob(A) || isVehicleMultitile(A))
 		HasProximity(A)
 


Git-Nivrak (changes requested):
Fix linters

code/modules/cm_aliens/structures/trap.dm
```diff
@@ -344,6 +344,17 @@
 	..()
 
 /obj/effect/alien/resin/trap/Crossed(atom/A)
+	if((isStructure(A) && istype(A, /obj/structure/bed)) || (isVehicle(A) && !isVehicleMultitile(A)))
+		var/obj/O = A
```
Detective-Google, line 348: https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#dont-use-abbreviations
one letter var needs to be changed

Detective-Google (changes requested):

PhantornRU:
nooo it's not stale =(

Detective-Google (changes requested):
in testing, this didnt' actually fix the fact that you can wheel around through acid traps, only throwing you off if you hit the very center tile. Overall behaved in a janky unintuitive way. Your code should probably also apply to the invisible tripwires placed by those traps. 

It should only unbuckle the user if they actually set the trap off;
You should remove the forcemove, this looks weird in practice during testing as opposed to just buckling.

code/modules/cm_aliens/structures/trap.dm
```diff
@@ -344,6 +344,17 @@
 	..()
 
 /obj/effect/alien/resin/trap/Crossed(atom/A)
+	if((isStructure(A) && istype(A, /obj/structure/bed)) || (isVehicle(A) && !isVehicleMultitile(A)))
+		var/obj/target = A
+		if(target.buckled_mob)
+			var/mob/living/user = target.buckled_mob
+			target.unbuckle()
+			user.forceMove(get_turf(target))
```
PhantornRU, line 352:
```suggestion
```

Detective-Google:
https://github.com/user-attachments/assets/a61fee7f-1cc2-474b-9cf1-84fdc124c29d

still hasnt addressed the interaction with tripwires on multi tile traps

PR #8953, opened 2025-04-03
XOs and COs now have basic piloting skill

thatrandomguy12131:
# About the pull request

Simple change to give executive and commanding officers basic piloting.

This is a very simple change, it will likely be used more rarely however I feel it could show to be important. In a lore sense you could argue the CO and XO wouldn't know how to fly, but that same argument would throw the ASO and SEA out the window as well so I feel this is fine. I saw some folks on the discord that said it was annoying COs didn't have pilot as well so I just went ahead and made this.

# Changelog
:cl: Hahafunnyname
add: COs and XOs now have basic piloting skill
/:cl:


Git-Nivrak:
I don't think this is needed, your "why is it good for the game" isn't convincing either

PR #9023, opened 2025-04-10
Reduces stun from unburrowing to 1 second

TheKillfish:
# About the pull request

Lowered stun from unburrow from 2 seconds to 1 second.

Oh, and redefined some vars used for timing to SECONDS. Is better code.

# Why this is good for the game

Fixes: #7076, probably.

Stunlocking bad, especially one as easy as this one was.

# Changelog
:cl: Killfish
balance: Reduced stun from unburrowing from 2 seconds to 1 second. Stunlock begone, maybe.
/:cl:


Drulikar (changes requested):
```js
if(!action_cooldown_check())
	return
```
Is required to do anything more than visually change the button. Please test and tinker with it. You likely only want cooldown applied for a full sequence of burrowing, but try it out.

TheKillfish:
Ah, piss. Knew I probably forgot something

code/modules/mob/living/carbon/xenomorph/abilities/burrower/burrower_abilities.dm
```diff
@@ -7,16 +7,21 @@
...
+	if(!action_cooldown_check())
+		return
+
 	if(SSticker?.mode?.hardcore)
 		to_chat(xenomorph, SPAN_XENOWARNING("A certain presence is preventing us from burrowing here."))
 		return
 
 	if(HAS_TRAIT(xenomorph, TRAIT_ABILITY_BURROWED))
 		xenomorph.tunnel(get_turf(A))
+		apply_cooldown()
```
Red-byte3D, line 24: are you sure this actually does anything? this looks like it just applies the cooldown when you initially use the ability, which would be made moot by the fact theres a tunnel time etc?

TheKillfish: You are correct. Dangit. I wanted this to be an easy PR, _why must you be such, Burrower code!?_

Red-byte3D: The easy way to fix this sto apply_cooldown() after the final proc goes through, just follow what xenomorph.tunnel goes to

Drulikar (changes requested):
> Fixes: https://github.com/cmss13-devs/cmss13/issues/7076, probably.

It doesn't now. I implore you to actually test and tinker with the changes to achieve what you're trying to do.

- Take in to consideration some marines will get up faster than others - so you should test against equipped riflemen.
- You cannot spam the ability else you will get set on a cooldown - but timed correctly it can be very quick for 0 tile burrowing/unburrowing.

Things to tinker with:
- Minimum duration when burrowed to unburrow (even if distance is 0)
- Stun duration when unburrowing under a marine.

PR #9030, opened 2025-04-11
Gun Jamming additions

Nihisohel:
# About the pull request

Adds a few more tedium and tweaks to gun jamming.

Clarifies durability chances, it now goes from
Perfect at 100
Fine at 99-81
Normal at 80-51
Poor at 50-21
Bad at 20-1
Broken at 0

Refactors the way jamming is calculated, it is now calculated exponentially rather than linearly (Or at least I think my calculation did it properly) as a result, jamming chances are extremely lower than it should be but having it happen a bit more rarely is better than happening frequently anyhow. 
Most jamming chances still occur at a threshold of 80, shotgun jamming threshold is now lower at 50 durability.

Fixes bullet cocking, it won't increase infinitesimally each time you attempt an unjam.

ES4 Wey-Yu pistol no longer guarantees durability loss each time it fires, instead becomes a coinflip probability at 50

Fixes grenade launchers and flare guns from being unable to be repaired
closes: #9048
closes: #9096 

Your gun now gets damaged from more sources: 
Xenos can also melee guns to damage them, but not destroy them, this should make gun deletion a bit fairer and less tedious.
Your gun can also be damaged by flames, try not to cook it in an OB.

It works.
image

# Changelog
:cl:
balance: Xenos can now damage guns on the ground
balance: Jamming is now calculated exponentially rather than linearly
qol: Clarifies durability examine wording, gives it some color too
fix: Fixes infinite gun cocking delay via unjamming
fix: Fixes grenade launchers and flare guns from being unable to be repaired
fix: Fixes bullets loaded in guns without a traditional chamber defaulting its values rather than incorporating it
balance: ES4 pistol no longer guarantees durability loss per fire, instead coinflips the chances
/:cl:


code/modules/projectiles/magazines/rifles.dm
```diff
@@ -92,7 +92,7 @@
 	name = "\improper M41A MK1 magazine (10x24mm)"
 	desc = "A long rectangular box of rounds that is only compatible with the older M41A MK1. Holds up to 95 rounds."
```
private-tristan, line 93:
```suggestion
	desc = "A long rectangular box of rounds that is only compatible with the older M41A MK1. Holds up to 99 rounds."
```

Steelpoint:
If you want to make this extremely lore accurate, firing a Mk1 with a 99rnd magazine should have a significantly higher chance to trigger a jam/misfire. Of course this raises a "noob trap" issue unless its made extremely clear why this is happening, and a easy way to remove the 4 extra bullets is added. (Or make it you must manually add 4 extra bullets when the magazine is already up to 95rnd  with the chat noting this increases the risk of a jam)

DrAmazing343:
Certainly make them come pre-loaded with 95 and allow bullet insertion up to 99. 

According to Lee Brimmicombe-Wood’s Aliens: Colonial Marines Technical Manual, it is actually standard practice to only load up to 95 to prevent jamming. (Thanks, Alien wiki!)

Given that it’s standard practice, I’d expect the pre-loaded kits we’re given to not be loaded beyond jamming point.

Detective-Google (changes requested):
separate misfiring into it's own feature PR without tweaks/fixes bundled to be evaluated for balance, ditto for Mk1 mag size change.

code/modules/projectiles/guns/specialist/launcher/grenade_launcher.dm
```diff
@@ -109,6 +109,10 @@
 
 
 /obj/item/weapon/gun/launcher/grenade/attackby(obj/item/I, mob/user)
```
Detective-Google, line 111: fix the single letter var in this proc while you're here

Nihisohel: death to all cm dev changes

Detective-Google (approved):

PR #9034, opened 2025-04-11
lurker rework, Pounce changes, invis decreased on pounce/damage

private-tristan:
# About the pull request

please TM before merging, I want to see how it actually plays out ingame.

Buff - Lurker pounce now acts like runner pounce when not cloaked, stunning for 1s.
* This pounce will be identical to runner pounce, giving the lurker enough time for 1 slash before leaving (due to lurker's slower attack speed), and a 2nd slash right as the marine gets up (assuming endurance 1). Lurker pounce stun/freeze **unchanged if you're cloaked**

Buff - Lurkers heal a bit faster when resting, from 1.5 to 2
* Previously they healed (relatively) slower when resting compared to runner's 1.75

Nerf - Lurker pounce CD increased to 7s (from 6)
* gives the lurker a little delay so that they can't immediately pounce the marine after they get back up, assuming the marine has endurance 1, there will be a roughly 1.5s~ delay before the lurker can pounce again. 

Nerf - Lurker pounce now **decreases invis**. 
* punishes lurkers for missing their pounce more, especially if the marine they attempted to pounce didn't know a lurker was trying to pounce them.  Lurkers can still run away, but they'll get, yaknow, shot at as they do, instead of invisibly scampering off into the darkness.
* Regular cloaked pounces still fully uncloak you.

Nerf - Lurkers taking damage while cloaked will cause their visibility to increase, if visibility increases too much, your invisibility is cancelled with a 50% cloak refund. **the speed boost from being cloaked will linger for 4 seconds**

Nerf - Decreases the slow-per-slash on floored targets from 3.5s to 2.5s per slash.
* with a cloak pounce, the marine will have roughly 8s of slowdown when they get up (from about 13s of slowdown)

QoL - Bumping no longer causes the generic "invis lost" message alongside the "you bumped and lost invis" message. 

**Reverted due to causing issues with GL prot** ~~Misc. Change - Runners/Lurkers/Hellhounds can no longer move during pounce (canceling it)~~
* ~~It's a bit janky how moving canceled your pounce, especially now that lurkers will lose their invis if they move mid-pounce.~~
*  ~~If someone feels strongly about it, I'm fine making it specifically only for cloaked lurkers or finding some way to have it as a toggle.~~

oh, and while I was looking through general-powers I decided to fix up some single letter vars.

It came up in CM discussion and I felt like throwing this PR together, I've generally felt that lurker hasn't been the most engaging to play, as you're forced to play around the 20s cooldown for invis so that you can access your stunning pounce.

With lurkers always uncloaking when pouncing, missing a pounce means that the lurker loses access to their escape tools while also probably being directly adjacent to the marine that they tried to pounce. Meaning that the marine is immediately made aware of the lurker (if they didn't notice/have a MD) and can fight back during the 8/10 seconds that the lurker has no escape tools outside of "crippling slash and run away"

Strob put it as:
>  adds skill expression to lurker / opens more opportunities to punish them for missing pounce ( while also allowing them more room to outplay without cloak )

As for the slowdown-per-slash change, it's a bit absurd that getting lurker pounce gave you 13 seconds of slowdown, so I toned it down a bit.
<details>
<summary>Demo Videos [UPDATED!]</summary>

Damaging a lurker will disrupt their cloak

https://github.com/user-attachments/assets/444f83b8-3619-4494-8d0f-eb909c00c76e

Uncloaked pounce (ignore the "We use our invisibility to empower our pounce!" msg, fixed).

https://github.com/user-attachments/assets/9805b259-82f1-4d7a-88cd-337ea3bbdfbb

Full Combo Demo

https://github.com/user-attachments/assets/b71c3e0c-36c6-4607-ba84-02e90695b2aa

Pouncing while your cloak is too damaged will cause it to be ended.

https://github.com/user-attachments/assets/34bc3794-9cbf-4b6a-ae99-68b4d2ac5096

Bumping a marine while invis doesn't print 2 messages to your chatbox

https://github.com/user-attachments/assets/b6a2f52a-e6d6-4b2c-a293-414545800bee

</details>

# Changelog
:cl:
balance: Lurker's Pounce now acts like a runner pounce if the lurker isn't cloaked, stunning for 1s. Cloak pounce stays at 2.5s
balance: Lurker's Pounce now makes you more visible 
balance: lurkers taking damage while cloaked will have their visibility increased, becoming too visible will cause cloak to be cancelled. The speed boost linger for 4 seconds after the cloak is disrupted.
balance: Lurker's Pounce CD increased to 7s (from 6s)
balance: Decreased Lurker's slow-per-slash on floored targets from 3.5s per slash to 2.5s per slash. A full cloak combo will generally leave the marine with 8s of slowdown
balance: Increased lurker's health regen mult while resting from 1.5 to 2(for reference, runner has 1.75). This translates to about 13 hp per tick while resting.
qol: Lurker's no longer get 2 different "invis loss" texts when bumping into a visible marine.
code: Replaced single letter vars
/:cl:


Mister-moon1:
You should probably fix the part where pounces and lunges end early if you are still moving or use the ability multiple times etc before actively punishing them for janky mechanics, imo though why play lurker after this just play runner

private-tristan:
> You should probably fix the part where pounces and lunges end early if you are still moving or use the ability multiple times etc before actively punishing them for janky mechanics

Runners/Lurkers/Hellhounds can no longer move mid-pounce.

code/modules/mob/living/carbon/xenomorph/abilities/general_abilities.dm
```diff
@@ -211,6 +211,8 @@
 	var/freeze_timer_id = TIMER_ID_NULL // Timer to cancel the end freeze if it can be cancelled earlier
 	var/freeze_play_sound = TRUE
 
+	var/move_during_pounce = TRUE  // Can you move while pouncing to cancel it?
```
Drulikar, line 214: All the comments here really ought to be [autdoc'd](https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/AUTODOC.md#documenting-a-variabledefine) comments instead.

Drulikar (changes requested):

cm13-github:

<details><summary>Test Merge Deployed @ 04/18/2025 12:09:54 +00:00:</summary>

##### Server Instance

##### Revision
Origin: be723fd4d5eb7a20d59ab9fe36c6a1383e6814ce
Pull Request: 215806ee6a4269c1997acd7fe03b855eb11631cc
Server: 4eafccc750488f7fc4919dc3b8ed8e7b712568bf
</details>

Kyogon:
Based off the description this feels like lurker is being gutted and turned into a runner side-grade instead of a t2 compared to a t1. The pounce nerf was already enough to make them much weaker, but making their invisibility cancel on using it alone just feels weird. Was the goal to make lurker into a runner who sometimes goes invisible? As it stands this is most likely going to make runner the least-played caste. The least they could get is some armor considering a PB will still put a lurker into near-crit, and the stun is now non-existent.

code/modules/mob/living/carbon/xenomorph/abilities/general_abilities.dm
Drulikar: `///` is required to autodoc

private-tristan: Oops, fixed.

RainbowStalin:
When you normal pounce (atleast testing with a nested human), you don't face towards the pounce. So you get sidewayds/backwards pounces.

private-tristan:
> sidewayds/backwards pounces.
this is due to Dirlocking, which isn't interrupted by pouncing.

Red-byte3D:
This gives lurker way too much presence, you should not be able to stun people while uncloaked, runners can do this because the damage they deal is low, they have lower health and have to pick their fights and have to display a relatively higher skill level to not instantly die

This makes lurkers just do what runners can but at a higher level with more damage  + the animation attack cancel on crippling slash

cloaking system is just a straight nerf as you'll basically be useless the moment theres an sg on your screen

private-tristan:
reverted my pounce tracking buff, **spriteclicked pounces no longer track the marine** (unless you're already adjacent, which is how they used to work).

image

Detective-Google (changes requested):
the changes to lurker pounces make pounce protection (e.g. shields, GL armor) not work properly

PR #9036, opened 2025-04-11
Small-ish Burrower Rebalance

TheKillfish:
# About the pull request

Basically goes about doing what my previous PR (#9023) aimed to do but better (AKA no cooldowns to make things harder to use).

Reduces unburrow stun from 2 seconds to 1.5 seconds and ensures the travel time for burrowing has a minumum of 1 second. This sorts out #7076, also has the side-effect of making it harder to solocap by unburrowing on top of someone by taking away 1 guaranteed tackle (from 4 guaranteed tackles to 3 guaranteed tackles).

Tremor has had its cooldown reduced from 45 seconds to 30 seconds to compensate and make it a bit less miserable.

Additionally kills two vars not used anywhere.

Burrow stunlocking sucks, so it's dead now.

Solocapping being harder is an unintended effect of killing burrow stunlocking, however speaking from personal experience as things are now, it is not very engaging being solocapped in FOB by an enemy you didn't even know was nearby as you have barely any chance to resist or attempt an escape; your fate is in the hands of tackle RNG, how fast you can type a cry for help over comms (assuming comms are even up), and the reaction time/gear of your fellow marines. *I am asking feedback from those more experienced as to how bad this may or may not be.*

Tremor cooldown reduction is to both compensate for the above, as it is the only other get-away tool Burrower has if caught in a bad position, and since 45 seconds for a 1 second stun (often shorter due to Endurace skill) feels a bit excessively long. For a T2, 30 seconds for an AOE with 1 second stun feels as if it should be fair, however this can be reverted or tweaked further if need be.

# Changelog

:cl: Killfish
del: Removed two entirely unused variables
balance: Burrower now has a minimum travel time of 1 second, and unburrow stun has been reduced from 2 to 1.5. No more stunlocking, and surprise solocapping in FOB is now a bit harder.
balance: Tremor has had its cooldown reduced from 45 seconds to 30 seconds.
/:cl:


code/modules/mob/living/carbon/xenomorph/castes/Burrower.dm
```diff
@@ -265,7 +263,10 @@
 		to_chat(src, SPAN_NOTICE("We cannot tunnel to there!"))
 	tunnel = TRUE
 	to_chat(src, SPAN_NOTICE("We start tunneling!"))
-	tunnel_timer = (get_dist(src, target)*10) + world.time
+	var/target_distance = get_dist(src, target)
+	if(target_distance == 0)
+		target_distance = 1 // This is to prevent stunlocking
```
Drulikar, line 268:
```suggestion
	var/target_distance = max(get_dist(src, target), 1) // Min distance of 1 is to prevent stunlocking
```

Drulikar (approved):

PR #9067, opened 2025-04-14
UPP HvH fixes, removes shipside announcements and echo-squad from UPP.

Aurallianz:
Removes shipside announcements for HvH, additionally removes echo squad from UPP Computer.

# About the pull request

bug fixes

bug fixes

<details>

</details>

# Changelog

:cl:
fix: Removed shipwide announcements as both ships share ship announcements.
fix: Removed echo-squad from the UPP computer.
/:cl:

### DIFF
--- a/code/game/machinery/computer/almayer_control.dm
+++ b/code/game/machinery/computer/almayer_control.dm
@@ -207,6 +207,10 @@
 				to_chat(human_user, SPAN_WARNING("Biometrics failure! You require an authenticated ID card to perform this action!"))
 				return FALSE
 
+			if(GLOB.master_mode == GAMEMODE_FACTION_CLASH_UPP_CM)
+				to_chat(user, SPAN_WARNING("You cannot make a shipwide announcement due to Operational Security."))
+				return FALSE
+
 			if(!COOLDOWN_FINISHED(src, cooldown_message))
 				to_chat(user, SPAN_WARNING("Please allow at least [COOLDOWN_TIMELEFT(src, cooldown_message)/10] second\s to pass between announcements."))
 				return FALSE
--- a/code/game/machinery/computer/groundside_operations.dm
+++ b/code/game/machinery/computer/groundside_operations.dm
@@ -21,6 +21,7 @@
 	var/add_pmcs = FALSE
 	var/lz_selection = TRUE
 	var/has_squad_overwatch = TRUE
+	var/echo_enabled = TRUE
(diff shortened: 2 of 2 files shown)


code/game/machinery/computer/almayer_control.dm
```diff
@@ -207,6 +207,10 @@
 				to_chat(human_user, SPAN_WARNING("Biometrics failure! You require an authenticated ID card to perform this action!"))
 				return FALSE
 
+			if(GLOB.master_mode == GAMEMODE_FACTION_CLASH_UPP_CM)
+				to_chat(user, SPAN_WARNING("You cannot make a shipwide announcement due to Operational Security."))
+				return FALSE
+
```
Detective-Google, line 213: Don't really love this solution to this problem; + people reported that non-ship announcements were also being heard.

code/game/machinery/computer/groundside_operations.dm
```diff
@@ -83,7 +84,7 @@
 	dat += "<BR><A href='byond://?src=\ref[src];operation=mapview'>Tactical Map</A>"
 	dat += "<BR><hr>"
 	var/datum/squad/marine/echo/echo_squad = locate() in GLOB.RoleAuthority.squads
-	if(!echo_squad.active && faction == FACTION_MARINE)
+	if(echo_enabled && !echo_squad.active && faction == FACTION_MARINE)
```
Detective-Google, line 87: this line already checks for the `faction` of the console, why not set the `faction` of the UPP console to UPP instead?

Detective-Google (changes requested):

code/game/machinery/computer/almayer_control.dm
Aurallianz: Based on my testing, I haven't encountered this issue with command announcements and I haven't seen any evidence that it occurs for either side. Both maps are configured to the master shipmap, which is how the announcement is transmitted. The challenge is that I would need to implement a method to check if the master map is the Almayer, but I prefer not to do that since it would impact the master gamemode. If you can think of an alternative workaround that doesn't affect HvX, please share your ideas.

code/game/machinery/computer/groundside_operations.dm
Aurallianz: That for some reason doesn't resolve the issue; it seems ineffective, I don't know where it pulls from. My implementation simply removes it entirely.

PR #9078, opened 2025-04-16
Autowiki Tweaks. Skips/Armour damage table 0s/Xeno strain comparison

MistChristmas:
# About the pull request

Skips some things like bursting pipes which are a... subtype of grenade.

SG barrels? These aren't removable it's not really an attachment ugh.
88 Burst Stock ??

Other barrels and stocks that are basically just the weapon because they're not removable.

14 different kinds of bayonet which are basically the same. The urge to add some sort of generic bayonet for just autowiki is high but my effort is low. Cause I'd need to add it to every gun that has bayonets.

Armour damage table thingie uses isnull to allow 0 values in the table.

Implement Drathek's thingie for looking at the differences for strains and base. Though just using a loop and adding colour.

Uhhhh wiki shouldn't be flooded with bayonets, and random attachies. And grenades that pretend they're bursting pipes. Armour damage tables shouldn't say armour-40 when it should be 0.
<details>

</details>

# Changelog
:cl: MistChristmas, Drathek
qol: Autowiki: Ignores most bayonets, and other misc. attachments. Compares difference in base vs strain stats. Switches to using isnull so armour damage table populates 0s.
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/grenades/marines.dm
+++ b/code/game/objects/items/explosives/grenades/marines.dm
@@ -908,11 +908,15 @@
 // abstract grenades used for hijack explosions
 
 /obj/item/explosive/grenade/high_explosive/bursting_pipe
+	AUTOWIKI_SKIP(TRUE)
+
 	name = "bursting pipe"
 	alpha = 0
 	mouse_opacity = MOUSE_OPACITY_TRANSPARENT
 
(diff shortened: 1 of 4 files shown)


Drulikar (approved):

Drulikar:
For future:
- mags may want to be lined up with damage tables or separated.

PR #9062, opened 2025-04-14
IASF insert & Hybrisa changes & Fixes.

Zenith00000:
# About the pull request

# Explain why it's good for the game

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl: Zenith, BOBAMA, Drathek, vandujr
add: Added the L54 service pistol & two (custom) variants
fix: fixed RMC boots pathing
fix: fixed Scout cloak on urban maps
fix: fixed RMC boots pathing
code: changed some code
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
imageadd: added some icons and images
imagedel: deleted some icons and images
maptweak: Many fixes and small changes across Hybrisa.
maptweak: revamped/fixed up Steelpoints IASF insert.
maptweak: revamped/fixed up the CLF insert, now it spawns at the space-port north-west.
/:cl:

PR #9098, opened 2025-04-18
CLF Coordinator CO Role

Comxy:
# About the pull request
After seeing https://github.com/cmss13-devs/cmss13/pull/9056 I thought it would be cool to add a CO specific role for CLF. I came up with the CLF coordinator role which is important in universe to coordinate and help with CLF recruiters and CLF cells to coordinate attacks against UA. I figured CLF organizer would next up in this side hierarchy of the CLF. They wouldn't be commanding over the cell since we have the leader for that already but would be working together with the cell lore wise. Also adds a skin for colony knives instead of it being a marine knife.

I thought it would be cool to add a new CLF coordinater role which can only appear on NV for now. It will add more world building to the CLF. Also I changed the knife sprite survival pouches start with because I don't like them having a marine issued knife instead of a survival knife.

<details>

promoclf

</details>

# Changelog
:cl:
add: CLF Coordinator role
imageadd: CLF Coordinator armor, helmet, mask and colony knife
/:cl:


efzapa:
unless I am blind, I cannot find any supervisor vars in the job, I assume the Coordinator is still below and answers to the Commander?

Zenith00000:
Cool cool. Can I make some suggestions though? The helmet should probably have the red eyes be a separate thing, make them a mask or goggles rather.

Also the whole hood thing is cool, but I feel like a CO role should have something a bit different, although for the CLF it's a bit difficult to say, it's not bad though.

I think the HUD icon should be a star, as the CLF flag is the three stars, would be appropriate, and the one you're using is the same as the sniper spec, or looks very similar. CO roles have always used a star, and keeping to that would help make it less confusing.

I think also yes the name should just be Commander, or something similar to keep the role less confusing, as I'm pretty sure the CLF also have Commanders or even more appropriate would probably be "Cell Leader", but perhaps you should consult the lore guys, they'll know better.

psyendrocronologicalwarfare:
>Also the whole hood thing is cool, but I feel like a CO role should have something a bit different, although for the CLF it's a bit difficult to say, it's not bad though.

CLF are trained and supplied by the UPP at times so that could be a thing
like a UPP advisor or something /shrug

BOBAMAx:
This implementation would only work on NV because CO survivor slots don't open unless the map has a path set for them in the JSON. See here: https://github.com/cmss13-devs/cmss13/blob/9fa97f81bf7ba6b8db84c0e934b24e248f1bffe6/code/game/jobs/job/civilians/other/survivors.dm#L213

code/game/objects/effects/landmarks/survivor_spawner.dm
```diff
@@ -42,7 +42,7 @@
 	hostile = TRUE
 	equipment = /datum/equipment_preset/survivor/clf
 	synth_equipment = /datum/equipment_preset/synth/survivor/clf
-	CO_equipment = /datum/equipment_preset/survivor/clf //to prevent NV CO from spawning as a CO
+	CO_equipment = /datum/equipment_preset/survivor/clf/coordinator
```
efzapa, line 45: I am not 100% sure how surv code works, but would it be possible for multiple CLF COs to spawn, given they ALL have the CO gear flag now, I would just remove the vars for non lead roles

Zenith00000: Nah only one Synth and one CO Survivor role can spawn.

You will however need some extra code to get the CO role to not just spawn on normal rounds, as all the CLF spawns are part of an insert.

BOBAMA did it for my PR here, he was very helpful, since I basically did the same thing with having a CO role only spawn on an insert.

https://github.com/cmss13-devs/cmss13/pull/9062/commits/9ef88fa63d2a7fbae9fe69b858dfcca3ad66e17b

Zenith00000: You might still need the code for maps without a CO survivor, and still have to add the code BOBAMA did in any case, you'll have to test it on different maps.

Let's say you have CO survivor on high on LV, you'll spawn in as a CLF CO survivor I'm pretty sure, even on a NON CLF round. I was having the same sort of issues with Hybrisa on my PR. Please test it some more.

efzapa:
> Basically I had something like this structure in mind. Since CLF works in independent cells, things like coordinators, organizers and diplomats towards other factions like UPP are also necessary for the CLF to function. That is why the coordinator wouldn't be in charge of the cell but works independently together with them. I thought this would be a cooler RP role as CO and wouldn't follow the, what I find to be boring, classic hierarchy. 

imho I would place them under the Cell Commander unless you feel like having to make another Admin Role or a role higher both answer to like a Rebel Overseer or smth

there should *always* be some sort of higher role if there is a faction that players can regularly play, so that people instantly know "oh shit this is an admin" IC and they can issue orders ~~and BEs~~ if needbe

code/game/objects/effects/landmarks/survivor_spawner.dm
Zenith00000: > I am not sure what you mean. From what I understand CO surv cannot be rolled on LV since there is no CO survivor type attached to that map. Only LV522 and NV have them currently, so this would only spawn on NV since there is no CLF insert on LV522.

Well if you're just doing it for NV only then perhaps it won't, but I'd test to see if you can spawn in on other maps just incase.

cm13-github:

cm13-github:

cm13-github:

cm13-github:

PR #9103, opened 2025-04-19
Adds USCM Black Insulated Gloves to vendors

psyendrocronologicalwarfare:
# About the pull request

It adds black insulated gloves to engis, IOs, FTLs, 
SLs get it for 3 points, as their insuls arent free (same price as tan insuls)

It looks nice, and I don't think it would have any gameplay difference. 

# Changelog

:cl: stalkerino
add: adds black insul gloves to CT, FTL, SL and IO, gave it a new icon when its on the floor too
fix: fixed black gloves not being black
/:cl:

### DIFF
--- a/code/game/objects/items/storage/backpack.dm
+++ b/code/game/objects/items/storage/backpack.dm
@@ -541,7 +541,7 @@
 	icon_state = "marinesatch"
 	max_storage_space = 20
 
-/obj/item/storage/backpack/satchel/intelbrown
+/obj/item/storage/backpack/marine/satchel/intel/satchel/brown
 	name = "\improper USCM expedition leather satchel"
 	desc = "A kevlar reinforced IMP based satchel, can quickly be accessed with only one hand. It's made out of a kevlar-leather mix. Usually issued to USCM intelligence officers."
 	icon = 'icons/obj/items/clothing/backpack/backpacks.dmi'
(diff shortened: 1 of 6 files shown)


code/game/objects/items/storage/backpack.dm
```diff
@@ -535,6 +535,36 @@
 	icon_state = "intel_chestrig"
 	max_storage_space = 20
 
+/obj/item/storage/backpack/marine/satchel/intel/satchel
+	name = "\improper USCM expedition satchel"
+	desc = "A heavy-duty IMP based satchel, can quickly be accessed with only one hand. Reinforced with Kevlar so it doesn't rip. Usually issued to USCM intelligence officers."
+	icon_state = "marinesatch"
+	max_storage_space = 20
+
+/obj/item/storage/backpack/satchel/intelbrown
```
Red-byte3D, line 544: you should probably make this obj/item/storage/backpack/marine/satchel/intel/satchel instead of the wrong path
also use snake case intel_brown

Red-byte3D: although you dont really need to make intel/satchel you can just probably use satchel/intel_brown or something like that, but do fix it so the paths are the same

psyendrocronologicalwarfare:
I don't agree with the opinions you've stated except the one about the leather satchels, those can go.
This PR can be closed.

I don't see a need nor do I think it would look good for IOs to be always different
Black gloves are already something you can get, and I don't see any need for them to be yellow. It also doesn't affect me because I swap my insuls out for black insuls every map I deploy on.
They're marines, not a separate entity that HAS to have everything different designed for them.

PR #9114, opened 2025-04-19
Portable Cades Can Be Melted

DarkLordCabbage:
# About the pull request

Enables xenos to melt m-6 portable composite barricades.  Did some quick tests to ensure it worked and didn't insta-die under acid (It doesn't- a base barricade and portable cade take about the same amount of acid to destroy). 

When portable cades were added 6 years ago, they were much stronger, but also a more limited resource. Now, portable cades are worse than normal cades, albeit usable by everyone and in greater supply than they were. It doesn't make a lot of sense that you can melt barricades but not a subpar portable version, so this brings it in line with that.

<details>

image

image

</details>

# Changelog

:cl:
balance: portable composite barricades are now meltable
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade/deployable.dm
+++ b/code/game/objects/structures/barricade/deployable.dm
@@ -1,6 +1,6 @@
 /obj/structure/barricade/deployable
 	name = "portable composite barricade"
-	desc = "A plasteel-carbon composite barricade. Resistant to most acids while being simple to repair. There are two pushplates that allow this barricade to fold into a neat package. Use a blowtorch to repair."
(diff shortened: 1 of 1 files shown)


Butlerblock:
Could you change it so they can be manually climbed over as well?

DarkLordCabbage:
I thought about it Butler, but you can't barb portable cades, so it'd be unfair to marines to give them cades that xenos can climb over and they can't do much about. At the same time, making only marines able to climb over cades and not xenos would be too one-sided. If I were to add the ability to barb the cades, they'd lose their niche as a protective measure for everyone- since the hapless doctors trying to use them wouldn't be able to barb them. So as unfortunate as it is to run to the hold and die looking over portacades you can't climb, I can't think of a way to circumvent this.

TL;DR: It'd be nice, but I can't

Detective-Google (approved):
code, needs balance approval

PR #9120, opened 2025-04-20
multitile vehicle hvh stuff

BOBAMAx:
# About the pull request

Adds the ability to throw grenades through vehicle doors, both ways, respecting the door lock if you are throwing into it.
Getting hit by the van no longer is a knockdown and shoves you around like it does to xenos.
Dead people no longer take up passenger slots if they die inside vehicles as a failsafe to prevent the dead from clogging up the vehicle and making it impossible to enter it as a human without admin intervention.

CLF survivors have taken to using the van as a stronghold against marines as the marines can usually only send 2-3 people in due to the passenger limit. This means that when those marines die, the van completely fills up and the CLF inside basically become untouchable, which is frankly very silly. This makes it so that door-camping vehicle exits is a dangerous prospect, on both sides, as you could easily eat a grenade. The dead no longer taking up passenger slots if they die inside of the vehicle also prevents situations where everyone in the vehicle dies and nobody else can enter it due to the passenger limit, which is much more likely to become a thing in a world where you can throw nades through vehicle doors. Running people over with the van no longer stuns to discourage using the van in combat, and to either drive away or dismount and shoot.

My goal is for van combat to play out like this:
Marines that can see the van, shoot it until it stops moving, which is now much easier now that you can't get permastunned by it. They then, if the CLF inside choose to not dismount, simply blow up the occupants with grenades instead of sending in one guy at a time on a suicide mission. The marines can then reliably enter the van and remove the dead occupants because the slots are not all filled by dead people.

<details>
https://streamable.com/bfudas

</details>

# Changelog
:cl:
qol: you can now throw grenades through vehicle doors both ways
balance: vans no longer stun humans. dead people no longer take up space in vehicles if they die inside of it.
/:cl:

### DIFF
--- a/code/modules/vehicles/interior/interactable/doors.dm
+++ b/code/modules/vehicles/interior/interactable/doors.dm
@@ -58,8 +58,32 @@
 		interior.exit(M)
 	return XENO_NO_DELAY_ACTION
 
-/obj/structure/interior_exit/vehicle/attackby(obj/item/O, mob/M)
-	attack_hand(M)
+/obj/structure/interior_exit/vehicle/attackby(obj/item/O, mob/user)
+	if(istype(O, /obj/item/explosive/grenade))
+		var/obj/item/explosive/grenade/nade = O
+		if(nade.antigrief_protection && user.faction == FACTION_MARINE && explosive_antigrief_check(nade, user))
+			to_chat(user, SPAN_WARNING("\The [nade.name]'s safe-area accident inhibitor prevents you from priming the grenade!"))
+			// Let staff know, in case someone's actually about to try to grief
+			msg_admin_niche("[key_name(user)] attempted to prime \a [nade.name] in [get_area(src)] [ADMIN_JMP(src.loc)]")
+			return
+
+		user.visible_message(SPAN_WARNING("[user] takes position to throw [nade] through the door of the [src]."),
+		SPAN_WARNING("You take position to throw [nade] through the door of the [src]."))
+		if(!do_after(user, 10, INTERRUPT_ALL, BUSY_ICON_HOSTILE))
+			return
+
+		user.visible_message(SPAN_WARNING("[user] throws [nade] through the door of the [src]!"),
+		SPAN_WARNING("You throw [nade] through the door of the [src]"))
+
+		var/turf/exit_turf = get_exit_turf()
+		user.drop_held_item()
+		interior.exit(nade, exit_turf)
+		if(!nade.active)
+			nade.activate(user)
+		return
+
+	else
+		attack_hand(user)
 
 /obj/structure/interior_exit/attack_ghost(mob/dead/observer/user)
 	if(!interior)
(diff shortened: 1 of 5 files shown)


code/modules/vehicles/interior/interactable/doors.dm
```diff
@@ -58,8 +58,32 @@
 		interior.exit(M)
 	return XENO_NO_DELAY_ACTION
 
-/obj/structure/interior_exit/vehicle/attackby(obj/item/O, mob/M)
-	attack_hand(M)
+/obj/structure/interior_exit/vehicle/attackby(obj/item/O, mob/user)
```
Detective-Google: fix one letter vars while editing

code/modules/vehicles/interior/interactable/doors.dm
```diff
@@ -58,8 +58,32 @@
...
+		var/obj/item/explosive/grenade/nade = O
+		if(nade.antigrief_protection && user.faction == FACTION_MARINE && explosive_antigrief_check(nade, user))
+			to_chat(user, SPAN_WARNING("\The [nade.name]'s safe-area accident inhibitor prevents you from priming the grenade!"))
+			// Let staff know, in case someone's actually about to try to grief
+			msg_admin_niche("[key_name(user)] attempted to prime \a [nade.name] in [get_area(src)] [ADMIN_JMP(src.loc)]")
+			return
+
+		user.visible_message(SPAN_WARNING("[user] takes position to throw [nade] through the door."),
+		SPAN_WARNING("You take position to throw [nade] through the door."))
+		if(!do_after(user, 10, INTERRUPT_ALL, BUSY_ICON_HOSTILE))
```
Detective-Google: 1 SECONDS

Drulikar:
Stale exempt until reviewed and tested.

Detective-Google (changes requested):
some little things for now.

 Concerned about the possibility of people driving the van around throwing molotovs out the doors (the van can eat a LOT of bullets and this is already a problem with people driving over stuff *without* being able to attack from within.) Equally concerned about survs being able to effectively surround the van in a sea of molotov fire using this.

code/modules/vehicles/interior/interactable/doors.dm
BOBAMAx: done

code/modules/vehicles/interior/interactable/doors.dm
BOBAMAx: done

BOBAMAx:
> Concerned about the possibility of people driving the van around throwing molotovs out the doors (the van can eat a LOT of bullets and this is already a problem with people driving over stuff _without_ being able to attack from within.) Equally concerned about survs being able to effectively surround the van in a sea of molotov fire using this.

The van no longer stuns when driving into people so marines can kill it pretty quick because they dont have to stand back up, pick up their gun, wait for the wield delay etc. You can just keep shooting it. It dies in a few mk2 mags which go by pretty quick with a bunch of marines shooting at you at once. surrounding the van in fire is just a worse version of throwing the fire in the van which you can already do, but nobody does because it's a losing strategy. People resort to van holds as a last ditch desperation move not because it's actually good.

 My only concern is FOB rework getting merged and hostile groundside humans being able to walk into the fob, with or without a van, because it's just straight up undefended. Might be worth making multitile vics targeted by sentry guns based on last known driver IFF in the future and somehow making it so that humans also get deterred by FOB defenses without making it so that taking human prisoners results in rockets blowing up the FOB.

PR #9171, opened 2025-04-25
fixes a say_understand runtime

kiVts:
# About the pull request

TV calls say_understands with first argument as itself. We always expect Mob there, so theres  a typecheck now.
1k runtimes every round bad

<details>

</details>

# Changelog
:cl:
fix: Fix spammed runtimes related to tvs relaying messages
/:cl:

### DIFF
--- a/code/modules/mob/say.dm
+++ b/code/modules/mob/say.dm
@@ -110,12 +110,13 @@
 	if(universal_understand)
 		return TRUE
 
-	if(istype(other) && other?.universal_speak)
+	if(istype(other) && other.universal_speak)
 		return TRUE
 	//Language check.
-	for(var/datum/language/L in languages)
-		if(speaking?.name == L.name)
-			return TRUE
+	if(speaking)
+		for(var/datum/language/known_languages in languages)
+			if(speaking.name == known_languages.name)
+				return TRUE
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/say.dm
```diff
@@ -109,7 +109,8 @@
 
 	if(universal_understand)
 		return TRUE
-	if(other?.universal_speak)
+
+	if(istype(other) && other?.universal_speak)
```
Drulikar, line 113:
```suggestion
	if(istype(other) && other.universal_speak)
```

code/modules/mob/say.dm
```diff
@@ -109,7 +109,8 @@
...
 		return TRUE
 	//Language check.
 	for(var/datum/language/L in languages)
```
Drulikar, line 116:
```suggestion
	for(var/datum/language/known_lang as anything in languages)
```

- update `L` to `known_lang`
- Also wrap this loop in a `if(speaking)` rather than have the null check via `?` on speaking each loop iteration.

Drulikar (changes requested):

Detective-Google (approved):
lgtm, want another pair of eyes on it since evil saycode.

PR #9181, opened 2025-04-26
lungemines now gib the user

Lagomorphica:
# About the pull request

my main goal is to blow up
lunge mines will now gib the user on hit, this is to make it more in line with the fact that for most marine roles it will gib. i considered making it os that the user blows up with the same severity as the hit so if they have extremely heavy armour (B18 or similar) they can survive but decided to simply call gib although i can easily change it

damaged lunge mines do not gib the user, they deal relatively low damage in comparison

i don't find lunge mines to be bad for the game in their current state, BUT i think it's bad that you can use it and sometimes entirely avoid damage or die and be revived anyway, this item is niche as well. 

for further context, i think that it's fine for marines to charge into suicidal situations or OD/PB themselves to prevent being captured/dive into risky situations to secure a kill, but i DON'T think it's fine to basically guarantee a kill against a xeno/pred/whatever with zero consequences to yourself besides waiting for recovery - nothing else in the game incentivizes suiciding besides this item and everything else comes with a greater degree of risk than this item because you have to actually win the fight first for the kill and you can still be captured/dragged/flinged away.

n/a, i did test on local to make sure it wouldn't gib the wrong person, everything else works normally

# Changelog

:cl:
balance: undamaged lungemines now gib the user on hit, ex_act was removed due to being bugged (caused runtimes, did not do anything to xenomorphs due to checking the user for the explosive pen so ultimately won't change anything about it)
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -332,14 +332,13 @@
 	detonating = TRUE
 	playsound(user, 'sound/items/Wirecutter.ogg', 50, 1)
 
-	sleep(3)
+	addtimer(CALLBACK(src, PROC_REF(lungemine_detonate), target, user), 0.3 SECONDS)
 
+/obj/item/weapon/twohanded/lungemine/proc/lungemine_detonate(atom/target, mob/user)
+	var/turf/epicenter = get_turf(target)
+	cell_explosion(epicenter, detonation_force, 50, EXPLOSION_FALLOFF_SHAPE_LINEAR, null, create_cause_data(initial(name), user))
 	if(gib_user)
 		user.gib()
-
-	var/turf/epicenter = get_turf(target)
-	target.ex_act(400, null, src, user, 100)
-	cell_explosion(epicenter, detonation_force, 50, EXPLOSION_FALLOFF_SHAPE_LINEAR, null, create_cause_data(initial(name), user))
 	qdel(src)
 
 /obj/item/weapon/twohanded/lungemine/damaged
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -334,7 +334,7 @@
 
 	sleep(3)
 
-	if(gib_user == TRUE)
+	if(gib_user)
 		user.gib()
 
 	var/turf/epicenter = get_turf(target)
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -281,6 +281,7 @@
 	hitsound = "swing_hit"
 
 	var/detonating = FALSE
+	var/gib_user = TRUE
 	var/wielded_attack_verb = list("charged")
 	var/wielded_hitsound = null
 	var/unwielded_attack_verb = list("whacked")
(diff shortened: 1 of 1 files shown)


AwkwardStereo:
this change is based and i look forward to exploding instantly

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -333,6 +334,9 @@
 
 	sleep(3)
 
+	if(gib_user == TRUE)
```
Git-Nivrak, line 337:
```suggestion
	if(gib_user)
```

Git-Nivrak (changes requested):

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -333,6 +334,9 @@
 
 	sleep(3)
 
+	if(gib_user)
+		user.gib()
+
 	var/turf/epicenter = get_turf(target)
 	target.ex_act(400, null, src, user, 100)
 	cell_explosion(epicenter, detonation_force, 50, EXPLOSION_FALLOFF_SHAPE_LINEAR, null, create_cause_data(initial(name), user))
```
Detective-Google, line 342:
```suggestion

	addtimer(CALLBACK(src, PROC_REF(lungemine_detonate), target, user), 0.3 SECONDS)

/obj/item/weapon/twohanded/lungemine/proc/lungemine_detonate(atom/target, mob/user)
	var/turf/epicenter = get_turf(target)
	cell_explosion(epicenter, detonation_force, 50, EXPLOSION_FALLOFF_SHAPE_LINEAR, null, create_cause_data(initial(name), user))
	if(gib_user)
		user.gib()
```

Detective-Google: - use addtimer rather than sleep
- the ex_act proc was formatted incorrectly and didn't actually apply to xenos at all, it could be fixed by removing `user` from the args, however this makes the lungemine gib almost anything you hit with it (even crushers) which would be a large departure from the current behavior (e.g., people are used to it being bugged to ignore the ex_act anyway.) Removing it has the side effect of making it less effective against preds.
either remove the `user` arg from the ex_act or remove the line entirely as in this suggestion.
Also moved the gib to the end, this is because `cell_explosion` references the `user`, therefore we should wait to destroy the mob that is `user` until after that code has run.

Detective-Google (changes requested):

Detective-Google (approved):
code, testing, needs balance approval

lanse12:
fyi lungemines are basically the same as PB'ing a xeno, except you get stunned and damaged by the explosion
these only 1 shot runners.

RainbowStalin:
> fyi lungemines are basically the same as PB'ing a xeno, except you get stunned and damaged by the explosion these only 1 shot runners.

>damaged lunge mines do not gib the user, they deal relatively low damage in comparison

I think its these that you are thinking of. Normal lunge mines gib T3s.

lanse12:
> > fyi lungemines are basically the same as PB'ing a xeno, except you get stunned and damaged by the explosion these only 1 shot runners.
> 
> > damaged lunge mines do not gib the user, they deal relatively low damage in comparison
> 
> I think its these that you are thinking of. Normal lunge mines gib T3s.

normal lunge mines do not gib anything above runners, tested it on local

Drulikar (approved):

PR #9197, opened 2025-04-28
Fixes Oppressor Praetorian being able to pull from outside of view range.

Lagomorphica:
# About the pull request

oppressor praetorian can no longer pull outside of view range

due to how oppressor praetorian range is counted, it does not include your starting tile - this PR does not nerf the ability to pull on-screen (you can still reach the edge of the screen diagonally or in any other combination), but it does prevent you from being able to pull marines OFF-SCREEN by clicking on parts of their sprite that are sticking onto the next tile.

i tested this and tested the bug, 6 tile range still allows for grabbing anything within view but does not let you grab outside of your screen

# Changelog
:cl:
fix: oppressor praetorian can no longer click on marine muzzle flashes to grab you from off-screen
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_abilities.dm
@@ -124,7 +124,7 @@
 	plasma_cost = 180
 
 	// Config
-	var/max_distance = 7
+	var/max_distance = 6
 	var/windup = 8 DECISECONDS
 
 /datum/action/xeno_action/activable/oppressor_punch


Red-byte3D:
this is less of a oppressor problem, more of a sprite muzzle countinga s you problem, bandaid fix that will just cause more issues with other stuff, probably best to look at fixing the muzzle instead of this

Red-byte3D:
you can fix this by returning if the given distance to the hook is above 7

BathSaltsAddict:
> this is less of a oppressor problem, more of a sprite muzzle countinga s you problem, bandaid fix that will just cause more issues with other stuff, probably best to look at fixing the muzzle instead of this

The trouble is, Oppressors can still use the bug even if they can't click your sprite by moving back one tile and using Abduct in the transitory state between tiles, allowing them to hook from off-screen same as before, albeit having to, for a split second, briefly step into viewing range before moving back - rather than clicking the sprite.

Git-Nivrak (approved):

PR #9225, opened 2025-05-02
prevents fluffing growing unremovable facial hair on female characters

Lagomorphica:
# About the pull request
fluffing no longer gives you a beard as a female

currently you can't get rid of beards as a female character due to an arbitrary lock on picking one (you can circumvent this by simply changing your gender to male, picking one, then swapping to female regardless)

fluffing will give you one and it is unremovable aside from admin powers due to the fact that mirrors, razors, etc all check your gender and prevent it

i personally would have no issue removing this restriction and leaving it so you can grow a beard and remove it via mirror, but i think permanent cosmetic changes like this are bad especially when they look horrible

i did test on local but i didn't take photos

# Changelog

:cl:
fix: fluffing no longer gives you an unremovable beard as a female character
/:cl:


TheGamerdk:
Just add an exception to mirrors/razors to allow you to go to 'No beard' regardless of gender?

sippykot:
just remove gender restrictions from facial hair tbh

DarkLordCabbage:
I would agree on tweaking this so that instead, gender no longer affects facial hair.

Detective-Google (approved):
would appreciate a followup to remove the gender restriction all together, but this is fine for now.

PR #9066, opened 2025-04-14
Obliterates thermal glasses from in-rotation maps and goggle spawners.

Lagomorphica:
# About the pull request

Removes thermal goggles from random chance spawners and preset spawns on every map that I have found to have them.

# Explain why it's good for the game
Survivor, over the years, has progressively become stronger and stronger and stronger. This is an extremely small step, although it is one of the most egregious examples of overpowered map loot available to both survivors and marines that does not fit the current version of CM. 

Having permanent, full awareness of every single xenomorphs position using binoculars and being able to relay it is extremely unhealthy for the game, and is bar nothing one of the single most powerful things in the game - and, frankly, it shouldn't be available at all to a role that is designed to survive rather than to be a doom frag 9000 role, and much less should it fall into marine hands such as being given to the scout specialist or demolitionist.

# Testing Photographs and Procedure
N/A.

# Changelog

:cl:
balance: Removes thermals as a possible selection from random goggle spawners such as on Fiorina and from pre-selected map spawns on Sorokyne Strata and Kutjevo. 
/:cl:

PR #9228, opened 2025-05-03
marsochelmet removal shiva

psyendrocronologicalwarfare:
# About the pull request

removes marsoc helmet

#9066 

Survivors shouldn't be running around with light armor, marsoc helmet with unlimited night vision, a MK1, a subdermal armor implant

The nvgs are so easy to get that there's no point contesting it, because by the time you get there they already have that and a flamer so you die

survivors shouldn't have the same power as specialists, or marines

# Changelog
:cl: stalkerino
balance: removed marsoc helmet from shiva
/:cl:


Aurallianz:
Heyo!

You'll need to run /tools/mapmerge2/I Forgot To Map Merge.bat, which should be from the root file directory.

realforest2001 (approved):

PR #9231, opened 2025-05-04
Adds reset keys to survivor synthetic loadouts, ever-so-slightly tweaks UPP & CLF Synth Survivor loadouts 

tallfission:
# About the pull request

Survivor synthetics will now spawn with reset keys in their survival pouches by default, save for CLF survivors and UPP survivors, who will get reset keys in their backpacks as they always have. The addition of the key to the pouch /does/ add an additional slot. 

CLF synth survivors will now spawn with a regular tool pouch. 

CLF and UPP synth survivors will spawn with a Type 80 bayonet in their survival pouch instead of an M5. 

Almost all synthetic survivors currently do not spawn with any kind of reset key, which leaves both their surviving buddies and the marines very little in the way of reviving them if they are recovered while dead. Currently, the only way for marines to get reset keys is: 
1. The CL uses their briefcase to buy one
2. A combat technician, MT, or the CE using their limited points pool to buy one
3. Another synthetic is playing in the round 

None spawn on the Almayer naturally and this can be especially frustrating on lowpop, where most of these roles are not filled, and combat technicians don't often consider taking one, which can lead to the unfun circumstance of sitting dead in medical for 30 minutes while a doctor tries to find someone to buy a key or wait for the Liasion to wake up. Adding one to the synthetic, to be used when they go offline, should fix this. 

It also just feels kind of silly to me that a synthetic wouldn't carry one with them in the event of an emergency. 

I have tweaked the CLF loadout to carry a regular tools pouch as more of a QOL thing, they no longer carry duplicates of some tools and lack others. 

Adding type 80 bayonets to their pouches is also more of a fluff thing than anything, I'm marking it as balance though because they do a lot more damage on throwing. 

compiles!! does the things!! 

# Changelog

:cl:
balance: All survivor synthetic types now start with reset keys in their survival pouch 
balance: Synthetic survival pouch has increased storage capacity by 1 
balance: CLF and UPP synthetic survivors get Type 80 bayonets in their survival pouches
balance: CLF synthetic survivors get a regular tools pouch instead of a synthetic tools pouch 
/:cl:


tallfission:
redoing this better

PR #9243, opened 2025-05-08
Allows Yautja to Flay Prey with the Ceremonial Dagger Instead of Butcher Verb

matttheficus:
# About the pull request

Allows Yautja to flay a target with the Ceremonial Dagger and have it unset the current target (like the `butcher` proc does).

Tested it and ensured that `butcher` still functions correctly and is unchanged. Time-to-flay is approximately the same as the `butcher` verb so unless I need to change it to be exactly equal I'm leaving it as-is.

The proc only tracks the target trophy being taken when the scalp is removed.

During my first Youngblood round, I was unaware there even **was** a `butcher` verb. So, logic dictated that you use your bootknife (the Ceremonial Dagger) to skin downed Prey.

This, apparently, doesnt do anything. Well now it does.

<details>

</details>

# Changelog
:cl:
qol: flaying a target with the Ceremonial Dagger now announces and unsets Prey like the Butcher Verb.
/:cl:


code/modules/cm_preds/yaut_weapons.dm
```diff
@@ -656,6 +656,11 @@
 					user.put_in_inactive_hand(cut_scalp) //Put it in the user's offhand if possible.
 					victim.h_style = "Bald"
 					victim.update_hair() //tear the hair off with the scalp
+					if(user.hunter_data.prey == target)
+						to_chat(src, SPAN_YAUTJABOLD("You have claimed the scalp of [target] as your trophy."))
+						user.emote("roar2")
+						message_all_yautja("[user.real_name] has claimed the scalp of [target] as their trophy.")
+						user.hunter_data.prey = null
```
Detective-Google, line 663: this should be where they're made permanently dead in line with the butcher verb, no?

Detective-Google (changes requested):

code/modules/cm_preds/yaut_weapons.dm
matttheficus: The only additional "mechanic" is the fact that the scalp now counts as a trophy and completes your hunt. I didnt add any permakill that wasnt there due to it not being the main intent.

If there **should** be a permakill in this proc, I can always add it if desired.

Detective-Google (approved):

PR #9253, opened 2025-05-09
All pens are now clicky!

DarkLordCabbage:
# About the pull request

Now all pens make clicking sounds when you click them on/off. Changed var name and some descs to reflect this.

Has been tested, pens can still write fine, everything works on the surface!

Immersion! It honestly made so little sense to me that pens didn't click when I used them that I found a free pen clicking sound, downloaded audacity, and then discovered we already had pen clicking sounds. They were just only in effect for some pens???

# Changelog

:cl:
add: All pens now make clicking sounds when you turn them off/on
/:cl:


efzapa:
the reason *some* pens are clicky is because they are WY pens, and hence more fancy

Bert-Stein:
> hence more fancy
Solution: add less fancy clicky sound.

DarkLordCabbage:
> the reason *some* pens are clicky is because they are WY pens, and hence more fancy

You can buy clicky pens at Safeway. The pens must click.

JackieEstegado:
No.

Nanu308:
Mapping fine, but I agree, clicking noise annoyance I'd rather deal without.

PR #9257, opened 2025-05-10
Shotguns can no longer mount the underbarrel flamer.

Detective-Google:
# About the pull request

This PR removes the underbarrel flamer and it's advanced subtype from the allowed attachments list on shotguns.

- Shotguns and flamethrowers are both short range, high burst damage weapons largely balanced by the infrequency with which you can fire them. 

- Most skirmisher/backline castes can just about survive one hit from either, and have a chance to escape/fight back/etc.

- Combining the two burst damage weapons lead to effective instant kills in 1v1s where a caste like a warrior or lurker is meant to have heavy advantage, with no extra risk or requirement of much skill from the player using it.

- This undermines the purpose of those castes, and is un-fun for the players playing them.

# Changelog
:cl:
balance: underbarrel flamethrowers no longer fit onto shotguns.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/shotguns.dm
+++ b/code/modules/projectiles/guns/shotguns.dm
@@ -473,8 +473,6 @@ can cause issues with ammo types getting mixed up during the burst.
 		/obj/item/attachable/bayonet/upp,
 		/obj/item/attachable/verticalgrip, // Underbarrel
 		/obj/item/attachable/flashlight/grip,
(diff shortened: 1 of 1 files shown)


deathrobotpunch:
there is no reason in this PR that explains why you are removing the base underbarrel flamer?

realforest2001 (approved):

PR #9259, opened 2025-05-10
trijent LZs now powerable

Detective-Google:
# About the pull request

the LZs on trijent had UNPOWERABLE set to 1 which caused confusion as this was a departure from other maps where the expectation is that the marine FOB can be powered.

the marine FOB being able to be powered is standard, and is part of the incentive for controlling power generation as the marine faction, arbitrarily disallowing power from applying to this area is confusing.

# Changelog

:cl:
maptweak: trijent LZs can now be powered
/:cl:


Nanu308 (approved):
This is fine, although in the future we'll need to double check how we want outdoor areas handled further.

Git-Nivrak (approved):

PR #7714, opened 2024-12-01
Praetorian Dancer Dodge now creates Afterimage(s).

Venuska1117:
# About the pull request

Add "afterimage" effect to Dancer "Dodge" skill

**Make sure to TMs it first, i was testing it alone, and i have suspicion that it will break something**

# Explain why it's good for the game

Add Afterimage effect to Dancer prae Dodge, it lack any indication that it uses "Dodge", this PR is to give prae dancer visual indication of dodge ability, and make them look like they are "Hard to Shoot".

(this is intended to be only visuals)

# Important Info for adminbus:

**Changing (adding) color in VV apply on afterimages.**

**Changing .dmi apply changes to afterimages.**

_(see vids below to know what i am talking about here)_

# Demonstration Video:

<details>
<summary>Click HERE to see Video</summary>

Final Version: 

https://github.com/user-attachments/assets/9fa1c2bb-d582-49d0-b4af-c1839900d62c

|
**BELOW FOR ADMINS:**
|

VV "Color" change support:

https://github.com/user-attachments/assets/5d05acb5-a9f1-4a75-8023-1c419cf828df

Custom .dmi Sprite support:

https://github.com/user-attachments/assets/23ac862b-66ea-4e8c-bb4f-42a8d519e8a2

</details>

# Changelog
:cl: Venuska1117
add: Add "Afterimage" effect to Dancer Praetorian "Dodge" ability.
/:cl:

PR #9273, opened 2025-05-13
Praetorian Dancer Buff.

Venuska1117:
Requesting Testmerge for that dodge stat.

# About the pull request

This PR buff dancer strain for praetorian, but it still require high skills to execute correctly.

This is very old strain, it already exist for 6 years, and during this time its got its own changes made for it, but during past 4 years there was no significal changes to this strain (other than acid removal), and considering current times changes, this strain deserves its own buffs to keep up with high demand of marines getting stronger.

This change focuses on keeping dancer "glass cannon" apperance in exchange it gains:
- passive dodge of 18% 
- more health regeneration on weeds (1.35 multiplier)
- increased range of impale by 1 tile
-  dodge ability increases evasion when activated up to 38%.

Disclaimer: dodge allows you to dodge any projectile with set %, dodge only works when standing, not when lying, and it doesn't make you dodge explosions/fire.

Dancer keeps its "glass cannon" image but in exchange it gain evasion and regeneration, if they are fragile it should be compensated with other benefits.

Also added "Afterimage" visual effect to dancer "dodge" ability, its main focus was to indicate to user and surrounding players when dodge is still on effect. (addition from #7714)

For interested, now regeneration on weeds without phermones take 100 seconds (1minute 40 seconds) instead of 162,5 seconds (2 minutes 43 seconds), default value for regenerating when resting  as praetorian on weeds is 8, with buff its 13 now (+5)

# Videos of afterimages you can find here: #7714

# Changelog

:cl: Venuska1117
balance: Dancer gains passive dodge of 18% and regeneration Tier 7 (1.35 multiplier).
balance: Dancer Impale ability can now be casted from 2 tiles away, instead of 1 tile.
balance: Dancer Dodge ability now increase dodge value by +20% to total of 38% dodge when active.
add: Dancer Dodge ability now creates afterimages as visual effect and indicatior of your ability activation.
add: New mechanic for xenos called "dodge_chance" this mechanic allow to dodge any projectile with % chance.
/:cl:


Venuska1117:
> I am just gona warn you that accuracy is not exactly a balanced value per say on CM, rifle with basic attachements will hit all of its shots as accuracy is powercreeped to hell so it has more then 160% hit chance, but some weapons have accuracy of "only" 100% or 90% and then you more then halve the DPS, this might end up being frustrating both for the dancer and marine players as those who know will grab correct attachments and delete you and the ones who do not will not be able to cause any harm. And with evasion on most players will not be able to harm you at all, but few powergamers who get ready for it just still land all the shots, I think that V grip E barrel Reflex and stock get you at more the n185% hitchance, meanwhile most weapons just land none of the shots with no way even for the dancer to tell who can counter it as both just look like m41a

if someone wonders, dodge_chance is that mechanic, its already added.

PR #9281, opened 2025-05-15
Fix github actions with versioning, download attemts and windows strict

### LINKED ISSUE #9262: Fix linter byond.zip
### Testmerges

-

### Round ID

_No response_

### Description of the bug

Linter gives errors .zip Byond
Image

Text:
```
  End-of-central-directory signature not found.  Either this file is not
Archive:  byond.zip
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of byond.zip or
        byond.zip.zip, and cannot find byond.zip.ZIP, period.
Error: Process completed with exit code 9.
```

# Fix on TG:

Create PR

### Issue Bingo

biologed:
# About the pull request

Close #9262

Its good for devs

<details>

-

</details>

# Changelog
:cl:
fix: Fix github actions with versioning, download attemts and windows strict
/:cl:


Drulikar:
Do not use webarchive as a mirror. We are currently waiting on https://github.com/cmss13-devs/byond-build-mirror/actions Do not perform multiple attempts to download. If it fails let it fail. There is already a flaky workflow to rerun some actions as needed.

biologed:
> If it fails let it fail

You might as well not use CI/CD at all

PR #9304, opened 2025-05-19
Guarantees a fog gap on LV-624

Willzadl:
# About the pull request

Ensures that one fog gap will always be generated on LV.

As it stands, many people think that fog on LV is a relic and should be removed entirely. Furthermore, fog gaps (usually) lead to somewhat interesting close-quarters combat more often than not. This shouldn't lead to the same issues encountered in #4997 as it is still much more difficult to push through this gap than to just let the fog go down, as observed in rounds currently.

Additionally, removing the fog gap will give survivors access to the rest of the map, allowing them to get to the armory and crashed ship which otherwise seem quite pointless on an already weapon-scarce and extremely small map. Hopefully this will discourage and make drone rushing somewhat more difficult and allowing access to some (risky) loot with similar parity to other maps.

This may lead to unintended consequences of a new meta for the CLF crashed ship insert despite the benefits given to their hold anyway, but can be observed over a long time period and potentially always have no gap instead (meta-able? problem for later). The RNG isn't fun as it stands now anyway.

I tested it and I think it works, code should be pretty obvious. I don't think that recording a single or even a few rounds would prove that it does anyway.

<details>
nil

</details>

# Changelog
:cl:
maptweak: LV-624 will now always generate a fog gap.
/:cl:

### DIFF
--- a/code/game/objects/items/ore.dm
+++ b/code/game/objects/items/ore.dm
@@ -7,6 +7,8 @@
 
 /obj/item/ore/uranium
 	name = "pitchblende"
+	desc = "An ore containing Uranium. Just looking at it makes your head feel fuzzy... it's slightly luminescent."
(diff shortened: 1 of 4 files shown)


cm13-github:

Willzadl:
stupid github i hate this

PR #9311, opened 2025-05-20
CLF survivor changes/nerfs

Lagomorphica:
# About the pull request
Adjusts the sentries of CLF survivors to spawn with a random amount of ammo between 100 - 500 instead of having 99,999 rounds, makes it so only the SL can drive, bumps up the Engineers skill level and gives the SL the ability to treat internal bleeding and do basic medical tasks. This also removes the SL having Endurance 5 and SO equivalent orders, giving all of them Endurance 1 instead.

CLF survivors are incredibly secure especially on lowpop when they have infinite ammo turrets surrounding them, enabling them to simply wall off or fortify the shit out of one specific area.

Endurance 5 also should not be available in regular gameplay to survivors from what I understand.

Having less people able to drive the van means that it is a risk vs reward of who you send out rather than having everyone capable of it, so now you need to keep your synthetic/CLF leader alive.

Overall, brings them more in line with other survivors and stops them from being as powerful as the actual CLF ERT due to them having an impact on the gameplay loop.

# Changelog
:cl:
balance: CLF squad leader no longer has Endurance 5 or full construction/engineering skill, now having basic medical and surgical skills to treat IB. Most CLF roles aside from the squad leader and synthetic have lost the ability to drive the van. CLF engineer now has proper engineering training, and all CLF survivors have Endurance 1.
balance: CLF insert sentries no longer have unlimited ammo, instead spawning with a random amount between 100 - 500 to simulate them having been crashed for a while.
/:cl:


BOBAMAx:
Because the TL medic and engi all have the same weight to roll, and the minimum survivors spawned is 2, you could easily end up in a situation where nobody can actually drive the van. I'd recommend re-adding vehicle skills to the engineer so that there will always be at least 1 person who can drive the van round start.

Lagomorphica:
> Because the TL medic and engi all have the same weight to roll, and the minimum survivors spawned is 2, you could easily end up in a situation where nobody can actually drive the van. I'd recommend re-adding vehicle skills to the engineer so that there will always be at least 1 person who can drive the van round start.

i agree, done - i had assumed the SL would always spawn

Nanu308 (approved):
This is fine

Git-Nivrak (approved):

PR #9328, opened 2025-05-21
First-drop Sentry lose power if Marines evac while they're still active

TheKillfish:
# About the pull request

If there are fewer than 12 humans groundside while First Drop Turrets are active, their batteries will die.

If Marines evac before the First Drop Sentries lose power, they've forfeited groundside. Xenos do not need to be kept out of FOB if Marines forfeit.

<details>

</details>

# Changelog
:cl: Killfish
balance: First Drop Sentries immediately run out of battery if Marines evac before they run dry normally.
/:cl:


Drulikar (changes requested):
I do not want the sentries auto undeploying since they are a defense that the marines cannot move nor had say in where they are placed. So if they don't undeploy, you can know it'll atleast still have to be destroyed or undeployed manually to get past. Xenos can already examine them to know they are down, uses the same powered down icon state as other turrets so can verify visually, and they make noise & visible message when they power down.

code/modules/defenses/sentry.dm
```diff
@@ -602,18 +602,54 @@
...
+			set_battery_state(TURRET_BATTERY_STATE_DEAD)
+			forfeit_shutoff = TRUE
+			sleep(45)
+			playsound(loc, 'sound/weapons/smg_empty_alarm.ogg', 15, 1)
+			deployment_system.visible_message(SPAN_WARNING("[name] beeps steadily as it automatically depowers itself."))
+
 /obj/structure/machinery/defenses/sentry/premade/deployable/colony/landing_zone/get_examine_text(mob/user)
 	. = ..()
+	if(forfeit_shutoff)
+		return
```
Drulikar, line 651: No point this being here, but it would be useful in `set_battery_state`

code/modules/defenses/sentry.dm
```diff
@@ -602,18 +602,54 @@
...
+				if(groundside_humans > 12)
+					break
+
+		// Number of humans needed is same value used in the _check_danger() proc in hivebuff.dm and king_cocoon/process()
+		if(groundside_humans < 12)
+			set_battery_state(TURRET_BATTERY_STATE_DEAD)
+			forfeit_shutoff = TRUE
+			sleep(45)
+			playsound(loc, 'sound/weapons/smg_empty_alarm.ogg', 15, 1)
+			deployment_system.visible_message(SPAN_WARNING("[name] beeps steadily as it automatically depowers itself."))
```
Drulikar, line 646:
```suggestion
```
Don't sleep - use timers. But pointless to power it down fully, and then later make it beep again.

If you really wanted it to go critical immediately then go dead in 4.5 seconds, use a timer, and either set a `force` argument to `set_battery_state` to ignore `forfeit_shutoff` or convey that information in `forfeit_shutoff` as a different value.

Ultimately all `forfeit_shutoff` should do is just make the prepared timers do nothing - but you could also just alternatively keep a reference to those timers and kill stop instead.

PR #9334, opened 2025-05-22
Fix eraly latejoin larba 

kalazus:
# About the pull request

Reduces early latejoin marines per larva to 3. 
https://github.com/cmss13-devs/cmss13/blob/aa7ea27f611f94198c285d67cc07d84e816242d6/code/game/gamemodes/cm_initialize.dm#L123

https://github.com/cmss13-devs/cmss13/blob/aa7ea27f611f94198c285d67cc07d84e816242d6/code/controllers/configuration/entries/combat_config.dm#L28-L29

Xeno amount = ready amount (including xenos) / 4
Early latejoin (first 15 minutes), adds larva for every 4 players, but you cant latejoin as xeno.
First one counts xeno players, while second one doesnt.
P.s. maybe these values should be balanced, but second one should always be = first - 1 

Latejoining 1 second after round start gives the same amount of xenos as pressing "ready".

# Changelog

:cl:
balance: early latejoin marines counts toward larvas the same way as if they joined on round start.
/:cl:


Detective-Google:
For reference, this is an intentional discrepancy from https://github.com/cmss13-devs/cmss13/pull/6291

PR #9344, opened 2025-05-24
CMO given upgraded fireman carry as well as firearms one, researcher …

thatrandomguy12131:
…given basic fireman carry.

CMO now has fireman carry two and researcher fireman carry one.

CMO is also now given firearms one (if I changed the code correctly) and this is on account of them literally having a gun in their vendor as well as them being a command role.
# About the pull request

This is meant to make both the researcher Skillset a tad bit closer to that of a doctor as well as make the CMOs skillset more attuned to that of both a senior medical officer and a command role.

I have been racking up researcher hours and recently noticed how moving the injured is very difficult without fireman carry, as well as this during hijack I try and play a leadership like role as CMO and without any firearm skill my accuracy ends up hitting more of my fellow marines than it does any xeno in front of me. The Fireman carry two for CMO is just to make them a better doctor.

# Changelog
:cl: Hahafunnyname
add: CMO fireman carry/firearm skill buffed by one each, researcher given trained fireman carry.
/:cl:


Detective-Google:
We're not interested in this change right now, researchers are really only doctors secondarily and hauling around wounded isn't something they need to be able to do without a roller bed, really.

The CMO intentionally is *not* just doctor+ anymore either, we're gradually moving away from that in it's design philosophy, the CMO should be empowered to delegate their staff rather than incentivized to do everything themselves due to being faster at it.

CMO doesn't get firearms skill because they're not supposed to be fragging out they're meant to be supporting other marines.

PR #9366, opened 2025-05-27
Flameshape refactor

kalazus:
# About the pull request

Refactor of flameshape.dm to make it readable, removed recursion from FLAMESHAPE_DEFAULT and a bit easier to understand OB recursion. Also removed strange usage of OB recursion in /obj/structure/ship_ammo/laser_battery to place  1 tile of fire...
The only thing changed is that line flame will check for new obstacles when spreading, and green fire is able to replace itself.

Cleaner and faster code.

Checked OB, pyro flamer and datum proc called laser_battery and mortar_shell/incendiary.

# Changelog

:cl:
refactor: refactored some related to fire spreading code
/:cl:

### DIFF
--- a/code/modules/cm_marines/dropship_ammo.dm
+++ b/code/modules/cm_marines/dropship_ammo.dm
@@ -254,10 +254,18 @@
 	set waitfor = 0
 	var/list/turf_list = RANGE_TURFS(3, impact) //This is its area of effect
 	playsound(impact, 'sound/effects/pred_vision.ogg', 20, 1)
+
+	var/datum/reagent/fire_reag = new()
+	fire_reag.intensityfire = 75
+	fire_reag.durationfire = 5
+	fire_reag.burn_sprite = "dynamic"
+	fire_reag.burncolor = "#EE6515"
+
 	for(var/i=1 to 16) //This is how many tiles within that area of effect will be randomly ignited
-		var/turf/U = pick(turf_list)
-		turf_list -= U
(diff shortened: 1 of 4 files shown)


fira:
i beg of you please do something about these zero timers while you're here 😭 
Also the whole thing shouldn't be waitfor = 0, just use INVOKE_ASYNC on gameplay sideeffect calls
Gonna sound dumb too but stepping all 3 at same time sounds like it's gonna make it look clunky in comparison (probably doesn't matter?)

kalazus:
0 timers are good to prevent procs overtime (real lags). While they can be redundant in something like stars, line, and triangle (will remove it at least in line and triangle because it is more "player experience sensitive"), they arent in "default" and "irrefular".
I added only 1 waitfor (forgot to mention) to prevent line fire from penetrating walls if they were placed after initial click. So in line and triangle they have a reason to be here.
I dont undestand, why use INVOKE_ASYNC anywhere here? it is just waitfor wrapper.
And yeah, considering that i will remove timers from triangle, no reason to change it, reverting to previous behaviour.

fira:
INVOKE_ASYNC just means you know & tell future coders where you *actually* need to decouple your synchronous call, instead of slapping waitfor=0 and nobody being able to tell wtf it's for next year

Zero/One-timers are a crutch 99% of time, and most of the time don't actually help overtime. In this case rescheduling works still for things like the handheld weapon, or the OB, so yeah, sure, w/e

kalazus:
Can add comments to prevent "nobody being able to tell wtf it's for next year", but in line and triangle INVOKE_ASYNC wouldn't work (like, there are at most 3 lines that can be separated from "fire walking" proc).

PR #9388, opened 2025-05-29
Jungle Vines are now untrappable

Willzadl:
# About the pull request

In current LV, all vines (i believe) are placed on unweedable tiles. Hence, this PR should change nothing in current LV and will only affect the LV rework. The upcoming LV rework intends to make these tiles unweedable, but keep them untrappable as that is one of the original major balance intentions of the unweedable jungle in current LV. Requested by Kyo, this PR can be kept open until that PR is ready.

It's hard/impossible to see traps under vines, and there is a LOT of them in the jungle in very tight corridors, hence allowing them to be trappable would be a very significant, and undesired outcome (Kyo please confirm as lead mapper...)

# Changelog
:cl:
maptweak: Jungle Vines on LV are now untrappable.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
@@ -595,6 +595,10 @@
 			to_chat(xeno, SPAN_XENOWARNING("We cannot make a hole on a light!"))
 			return FALSE
 
+		if(locate(/obj/structure/flora/jungle/vines) in src)
+			to_chat(xeno, SPAN_XENOWARNING("We cannot make a hole under the vines!"))
+			return FALSE
+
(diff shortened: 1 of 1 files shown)


Kyogon:
Can confirm, this was requested by me for the incoming LV-624 rework. It is necessary and will not affect current balance in any large way due to most vines being in an area that cannot be trapped regardless. Only exceptions are a few spots in sand temple.

cuberound:
possibly prevent egg morpher and egg planting too. as under thickest wines (that behave like curatins) you can walk to adjacent tile without having chance to see them

Red-byte3D:
this also needs a balance tag

Willzadl:
> this also needs a balance tag

Like I said, this doesn't change balance now and won't change the balance of jungle when the LV rework is merged, as intended.

PR #9229, opened 2025-05-03
Minor Base Prae Buffs

DarkLordCabbage:
# About the pull request

This pull request hopefully makes the base strain of praetorian more viable without unbalancing it.

1. Removes the wind up from acid spray, it's now instant like spitter's or boiler's
2. Removes the wind up from acid ball, while firing it is now instant, the throw speed and detonation speed are the same
3. The dash can now pass over mobs instead of getting stuck if there's one in the way

Everything's tested and should be good to go.

# Explain why it's good for the game

Base prae needs a little bit of love. It and dancer are consistently overshadowed and outperformed by vanguard, oppressor, and valkyrie. These changes should make it a little less clunky to play and  feel a little smoother. The wind up removal owes itself to the fact that prae's acid spray doesn't stun or do insane amounts of damage, and its acid ball still takes time to land and detonate- meaning it can still be dodged or ducked. Base prae is supposed to be better than the other strains for survivability, so its dash ability passing over mobs brings it in line with that idea. No more will your retreat path be blocked by a boiler; bodyblock begone!

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

https://github.com/user-attachments/assets/53a3660d-423f-4028-b5bc-f76becbf1019

</details>

# Changelog

:cl:
balance: Base prae's acid spray/ball no longer have a wind up, its dash can now pass over mobs
/:cl:

PR #9392, opened 2025-05-30
balance: Minor Base Prae Buffs

LiquidPotroh:
# About the pull request
Just copy of [this PR](https://github.com/cmss13-devs/cmss13/pull/9229), but  removed the acid ball buff. As suggested in the comments. Author of that request wanted to do it himself, but apparently died.

1. Removes the wind up from acid spray, it's now instant like spitter's or boiler's
2. The dash can now pass over mobs instead of getting stuck if there's one in the way
Everything's tested and should be good to go.

Base prae needs a little bit of love. It and dancer are consistently overshadowed and outperformed by vanguard, oppressor, and valkyrie. These changes should make it a little less clunky to play and feel a little smoother. The wind up removal owes itself to the fact that prae's acid spray doesn't stun or do insane amounts of damage. Base prae is supposed to be better than the other strains for survivability, so its dash ability passing over mobs brings it in line with that idea. No more will your retreat path be blocked by a boiler; bodyblock begone!

<details>

https://github.com/user-attachments/assets/8134d548-7424-45e8-a874-bba39a27cefa

</details>

# Changelog
:cl:
balance: Base prae's acid spray, its dash can now pass over mobs
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_abilities.dm
@@ -228,6 +228,9 @@
 	slash = FALSE
 	freeze_self = FALSE
 
+/datum/action/xeno_action/activable/pounce/base_prae_dash/initialize_pounce_pass_flags()
+	pounce_pass_flags = PASS_MOB_THRU|PASS_OVER_THROW_MOB
+
 /datum/action/xeno_action/activable/prae_acid_ball
 	name = "Acid Ball"
 	action_icon_state = "prae_acid_ball"
@@ -289,8 +292,6 @@
 	spray_type = ACID_SPRAY_LINE
 	spray_distance = 7
 	spray_effect_type = /obj/effect/xenomorph/spray/praetorian
-	activation_delay = TRUE
-	activation_delay_length = 5
 
 ///////////////////////// VALKYRIE PRAE
 


LiquidPotroh:
still ready for the review

code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_abilities.dm
```diff
@@ -228,6 +228,9 @@
 	slash = FALSE
 	freeze_self = FALSE
 
+/datum/action/xeno_action/activable/pounce/base_prae_dash/initialize_pounce_pass_flags()
+	pounce_pass_flags = PASS_MOB_THRU|PASS_OVER_THROW_MOB
```
Drulikar, line 232: From the other pr:
> num 3 [dash changes] could be fine, but I would rather that to then be the behavior for ground targeted dashes generally (and only passing friendlies) since base prae isn't something special like dancer is where its gimmick is moving around mobs freely. So if thats something you really wanted, do that in its own PR because it'll be a decent bit complicated to handle the pass flags only for friendlies, and then us getting the desired behavior across the board.

So unless you're wanting to tackle making this the behavior generally for xenos, and only friendly xenos, I'd just stick to removing the windup for the acid spray.

Drulikar:
Just going to do the change myself https://github.com/cmss13-devs/cmss13/pull/9843

PR #9416, opened 2025-06-01
Adds holocards to stasis bags

SpypigDev:
# About the pull request

Makes holocards visible on stasis bags using a small overlay sprite

Shamelessly stole this idea from Jack Morrison on the forums

See original ideasguys thread here: https://forum.cm-ss13.com/t/holocard-on-stasis-bag/9855

Screenshots & Videos

https://github.com/user-attachments/assets/283f9308-86ea-4284-954d-acd741783f5e

# Changelog
:cl:
add: Holocards now visible on stasis bags
/:cl:


Aurallianz:
Excellent PR; I feel like it should be slightly more visible than it is. It's a little hard to notice, no?

SpypigDev:
> Excellent PR; I feel like it should be slightly more visible than it is. It's a little hard to notice, no?

My grainy video probably doesnt help, but the actual sprite looks like this, and the card flashes back and fourth, making it just noticeable enough that a fast moving runner would miss it, but a sharp eyed doctor would spot it

image

fira:
you could use a signal handler for that rather than strapping istype in parent code

TheManWithNoHands:
> My grainy video probably doesnt help, but the actual sprite looks like this, and the card flashes back and fourth, making it just noticeable enough that a fast moving runner would miss it, but a sharp eyed doctor would spot it

The purpel for the infected tag is too desaturated. Keep it the same shade as the normal holo tag, that one is brigth and saturated well already. What aint broke, dont need fixing and all that.

SpypigDev:
will be away for a few days. code will be updated shortly

SpypigDev:
> you could use a signal handler for that rather than strapping istype in parent code

Untitled

code/__DEFINES/dcs/signals/atom/mob/living/signals_human.dm
```diff
@@ -86,3 +86,6 @@
 
 /// From /mob/living/carbon/human/UnarmedAttack()
 #define COMSIG_HUMAN_UNARMED_ATTACK "human_unarmed_attack"
+
+/// From /obj/structure/closet/bodybag/cryobag/close()
```
Detective-Google: Wouldn't this be sent by `hud_set_holocard`, and not `cryobag/close`?

SpypigDev: yeah my B

PR #9425, opened 2025-06-02
Enables the storage of flags and tesla into the construction belt.

MemesLauncher:
# About the pull request

Enables the storage of flags and teslas into the construction rig belt.

I do believe it might be an oversight, considering all sentries, flame sentries and their "upgraded" counterparts are able to be put inside them.

Puts the tesla and flag in line with the other two sentries when it comes to storage and prevents attempts at putting the flags/teslas into the construction rig thinking it could be done the same as the flamer/normal sentries, possibly wasting their belt choice.

<details>

Capture d'écran 2025-06-02 140712

</details>

# Changelog
:cl:
balance: Enabled the storage of flag/teslas into the construction rig.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -155,7 +155,7 @@
 		/obj/item/stack/sandbags_empty,
 		/obj/item/stack/sandbags,
 		/obj/item/stack/barbed_wire,
-		/obj/item/defenses/handheld/sentry,
+		/obj/item/defenses/handheld,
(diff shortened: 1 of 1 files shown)


DrAmazing343:
THANK FUCKING GOD I KNEEL

Detective-Google (approved):
I think the construction belt could use a re-evaluation in general because it's *really* good, imo it's a bit of a storage creep for engineers.

that being said, I agree that if it should hold turrets at all, it should also hold the support ones.

PR #9410, opened 2025-06-01
Hand labelers can be used on crates

DarkLordCabbage:
# About the pull request

You can now use the hand labeler to label crates

# Testing Photographs and Procedure

# Changelog

:cl:
add: You can now use the hand labeler to label crates
/:cl:

PR #9428, opened 2025-06-03
Fix inconsistent labeler removal messaging

Drulikar:
# About the pull request

This PR is a follow up to #9410 where it was discovered that you can endlessly attempt to removal labels from things. Now things that would give feedback for label removal checks if its actually labeled (and its not simply that the component was left around). We could just always remove the component, but it looks like some things assume it exists and I dunno could be useful to see the last label I guess in VV?

More correct feedback for label removal.

<details>

image

</details>

# Changelog

:cl: Drathek
fix: Removing labels now has correct messaging if you attempt to remove an already removed label
/:cl:

### DIFF
--- a/code/game/objects/items/tools/misc_tools.dm
+++ b/code/game/objects/items/tools/misc_tools.dm
@@ -110,14 +110,14 @@
 
 */
 
-/obj/item/tool/hand_labeler/proc/remove_label(atom/A, mob/user)
-	var/datum/component/label/label = A.GetComponent(/datum/component/label)
+/obj/item/tool/hand_labeler/proc/remove_label(atom/target, mob/user)
+	var/datum/component/label/label = target.GetComponent(/datum/component/label)
 	if(label && label.has_label())
-		user.visible_message(SPAN_NOTICE("[user] removes label from [A]."),
-						SPAN_NOTICE("You remove the label from [A]."))
+		user.visible_message(SPAN_NOTICE("[user] removes label from [target]."),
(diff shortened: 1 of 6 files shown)


code/game/objects/items/tools/misc_tools.dm
```diff
@@ -110,14 +110,14 @@
...
-		label.remove_label()
-		log_admin("[user] has removed label from [A.name]. (CKEY: ([user.ckey]))")
-		playsound(A, remove_label_sound, 20, TRUE)
+/obj/item/tool/hand_labeler/proc/remove_label(atom/target, mob/user)
+	var/datum/component/label/label = target.GetComponent(/datum/component/label)
+	if(label && label.has_label())
+		user.visible_message(SPAN_NOTICE("[user] removes label from [target]."),
+						SPAN_NOTICE("You remove the label from [target]."))
+		label.clear_label()
+		log_admin("[user] has removed label from [target.name]. (CKEY: ([user.ckey]))")
```
Detective-Google:
```suggestion
		log_admin("[key_name(user)] has removed label from [target.name].")
```

Detective-Google: may potentially want to include what the label was in the log as well e.g. move it up a line and add [label.label_name]?

Drulikar: image

Detective-Google (approved):

PR #9429, opened 2025-06-03
Update byond mirror path for CI

Drulikar:
# About the pull request

This PR is a follow up to #9343 where the automatic workflow finally succeeded but I guessed wrong that the domain would be included in the url path.

See checks.

# Changelog
No player facing changes.


Detective-Google (approved):
looks fine, tested URL.

PR #9430, opened 2025-06-03
fixes micro tesla fire delay

cuberound:
# About the pull request

micro tesla has 1 second fire delay variable defined, but not applied, obvious oversight as with other teslas the variables are used

Makes micro tesla work as inteanded, without it it was just worse normal tesla (that did not block movement)

<details>

</details>

# Changelog
:cl:
balance: applies 1 second fire delay instead of 2 seconds (same as normal tesla) on micro tesla as intended
/:cl:


fira:
Can you clarify the balance effect in changelog (2 -> 1 seconds) ?
Also I'm of the mind this is still a Balance PR. If it's been like this for 4 years, there's no reason to "fix" it "just because" now. It doesn't really have much grounding in reality anymore.

PR #9447, opened 2025-06-05
Buffs Carriers max plasma to 800, keeps Eggsac at 600

TheKillfish:
# About the pull request

Title.

600 plasma makes it very tedious to build special structures as Carrier on account of your slow plasma regen and the high plasma cost of building those special structures (400 to make the template + 450 to 500 for common structures you'll be building, or 1000 if you're rebuilding the Hive Core). With 800, there'll still be waiting, but it'll be less than before.

My rationale for keeping Eggsac at 600 is with egg generation, you'll need to manage your plasma better.

# Changelog
:cl: Killfish
balance: Carrier gets a max of 800 plasma, while Eggsac remains the same.
/:cl:


squidqueen:
both should be buffed, eggsac is miserable to play because they are far more reliant on special structures than base carrier in order to have eggs that last longer than 5 seconds

Drulikar:
I would rather other things explored for carrier. This doesn't really do anything but make some initial set up faster and will just lead to other plasma changes that we don't need right now.

PR #9494, opened 2025-06-07
Makes Xeno Spit abilities actually use defined Cooldown vars

TheKillfish:
# About the pull request

Nukes spit delay vars on Xenos and the cooldown modifying part of Xeno Spit, and adding Xeno-specific subtypes where applicable with their defined cooldowns (and an appropriate name, like Praetorian's instance of Xeno Spit, which has also been stripped down to just the vars different to original type). This also does away with spit cooldown messages, since the only Xeno with slow enough spit to actually justify it is Boiler, which had its own specified message anyway.

Less shitcode, more in-line with other xeno abilities. This may add variants that only have two lines different to the original, but that's better than 6 year old shitcode being left to rot and be ugly/confusing.

# Changelog
:cl:
code: Xeno Spit now uses xeno_cooldown var, like a normal ability.
del: Deletes now-useless vars related to Xeno Spit cooldown
/:cl:


code/modules/mob/living/carbon/xenomorph/abilities/boiler/boiler_abilities.dm
```diff
@@ -53,6 +53,8 @@
 	cooldown_message = "Our stomach fills with another gas glob. We are ready to bombard again."
 	sound_to_play = 'sound/effects/blobattack.ogg'
 	aim_turf = TRUE
+	xeno_cooldown = 30 SECONDS
+	no_cooldown_msg = TRUE // Boiler spits slow enough that having a cooldown message is justifiable
```
fira: comment is reversed vs value, no?

TheKillfish: You are correct, thanks for catching

code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_abilities.dm
```diff
@@ -1,5 +1,6 @@
-//// Abilities PRIMARILY used by Praetorian.
-
+/datum/action/xeno_action/activable/xeno_spit/praetorian
```
Detective-Google: should go with the other base prae abilities, no?

PR #9542, opened 2025-06-08
Fix Autodoc bug caused by patient expiration

FebrezeNinja:
# About the pull request

Observed during a round, if patient expires during surgery then the external surgeries (brute/burn/tox/blood) don't reset properly when the patient gets ejected, meaning they can't be selected again until the reset is reached through another means.

This PR moves the reset to go_out which is called by every exit method.

Minor issue with a workaround (complete any different surgery) but still a bugfix.

<details>

</details>

# Changelog

:cl:
fix: Autodoc no longer locks procedures because of dead patient
/:cl:

### DIFF
--- a/code/game/machinery/medical_pod/autodoc.dm
+++ b/code/game/machinery/medical_pod/autodoc.dm
@@ -51,6 +51,12 @@
 
 /obj/structure/machinery/medical_pod/autodoc/go_out()
 	. = ..()
+	surgery = 0
+	heal_brute = 0
+	heal_burn = 0
+	heal_toxin = 0
+	filtering = 0
+	blood_transfer = 0
 	surgery_todo_list = list()
 	stop_processing()
 	if(connected)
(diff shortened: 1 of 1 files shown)


code/game/machinery/medical_pod/autodoc.dm
```diff
@@ -51,6 +51,12 @@
 
 /obj/structure/machinery/medical_pod/autodoc/go_out()
 	. = ..()
+	surgery = 0
+	heal_brute = 0
+	heal_burn = 0
+	heal_toxin = 0
+	filtering = 0
+	blood_transfer = 0
```
fira:
```suggestion
	surgery = FALSE
	heal_brute = FALSE
	heal_burn = FALSE
	heal_toxin = FALSE
	filtering = FALSE
	blood_transfer = FALSE
```

fira (approved):

PR #9554, opened 2025-06-10
Xenos with ranged building can deconstruct resin walls and doors

TheKillfish:
# About the pull request

Part of Queen Rework.

Xenos with ranged building can also use Secrete Resin to deconstruct Resin Walls that they are not adjacent to, deconstruct Resin Doors they are not adjacent to and if on Harm intent, and can fill Special Structure templates if not on Harm intent or remove them if on Harm intent. This method of deconstruction cannot be used against buildings not belonging to your hive.

Dead non-mob Facehuggers no longer block building.

Being able to deconstruct at range also helps Queens on Ovi construct more effectively and correct mistakes without needing to herd a Xeno around to break Walls/Doors.

Building being blocked because of a dead non-mob Facehugger is a bit silly.

# Changelog
:cl:
qol: Xenos with ranged building can use Secrete Resin to deconstruct Resin Walls and Membranes belonging to their hive at range.
qol: Xenos can now build over dead non-mob Facehuggers.
/:cl:


Drulikar (changes requested):
I am not keen on this refunding plasma. But it must not be allowed if the wall is already damaged. If it doesn't refund plasma then the damage aspect isn't as important (other than you're offering a way to grief hives).

code/game/turfs/walls/wall_types.dm
```diff
@@ -1284,23 +1284,42 @@
...
+		playsound(src, "alien_resin_break", 25)
+		return XENO_ATTACK_ACTION
+
+	if(xeno.a_intent == INTENT_HELP)
 		return XENO_NO_DELAY_ACTION
 
-	M.animation_attack_on(src)
-	M.visible_message(SPAN_XENONOTICE("\The [M] claws \the [src]!"),
+	xeno.animation_attack_on(src)
+	xeno.visible_message(SPAN_XENONOTICE("\The [xeno] claws \the [src]!"),
```
Drulikar:
```suggestion
	xeno.visible_message(SPAN_XENONOTICE("[xeno] claws [src]!"),
```
Generally we want to remove `\the` usage.

code/modules/mob/living/carbon/xenomorph/Powers.dm
```diff
@@ -32,67 +32,144 @@
...
+				if(message)
+					visible_message(SPAN_XENONOTICE("[src] regurgitates a thick substance and thickens [target]."),
+						SPAN_XENONOTICE("We regurgitate some resin and thicken [target], using [total_resin_cost] plasma."), null, 5)
+					if(use_plasma)
+						use_plasma(total_resin_cost)
+					playsound(loc, "alien_resin_build", 25)
+				target.add_hiddenprint(src) //so admins know who thickened the walls
+				return TRUE
+
+		if(can_deconstruct && !src.Adjacent(target))
```
Drulikar:
```suggestion
		if(can_deconstruct && !Adjacent(target))
```

code/modules/mob/living/carbon/xenomorph/Powers.dm
```diff
@@ -32,67 +32,144 @@
 			return SECRETE_RESIN_FAIL
 
 	var/turf/current_turf = get_turf(target)
+	var/can_deconstruct = TRUE
 
 	if(extra_build_dist != IGNORE_BUILD_DISTANCE && get_dist(src, target) > src.caste.max_build_dist + extra_build_dist) // Hivelords and eggsac carriers have max_build_dist of 1, drones and queens 0
```
Drulikar:
```suggestion
	if(extra_build_dist != IGNORE_BUILD_DISTANCE && get_dist(src, target) > caste.max_build_dist + extra_build_dist) // Hivelords and eggsac carriers have max_build_dist of 1, drones and queens 0
```

Drulikar (changes requested):

code/modules/mob/living/carbon/xenomorph/Powers.dm
```diff
@@ -32,67 +32,144 @@
...
-			if(message)
-				visible_message(SPAN_XENONOTICE("[src] regurgitates a thick substance and thickens [target]."),
-					SPAN_XENONOTICE("We regurgitate some resin and thicken [target], using [total_resin_cost] plasma."), null, 5)
-				if(use_plasma)
-					use_plasma(total_resin_cost)
-				playsound(loc, "alien_resin_build", 25)
-			target.add_hiddenprint(src) //so admins know who thickened the walls
-			return TRUE
+
+	if(istype(target, /obj/structure/mineral_door/resin) || istype(target, /turf/closed/wall/resin) || istype(target, /turf/closed/wall/resin/membrane))
```
Drulikar: There is a lot of unnecessary redundant typechecking being done in this proc. At a minimum store the resulting type into a var rather than typechecking 3+ times. Also the `istype(target, /turf/closed/wall/resin/membrane)` here will *never* be checked because a `/turf/closed/wall/resin/membrane` is a `/turf/closed/wall/resin`. `istype` includes inheritance. If you look at the existing logic here you'll see it is using `type` directly to perform strict typechecking. But it really shouldn't be doing this either because it doesn't future proof well: subtypes of these types should be treated the same and have the same behavior unless overridden.

Note: A cleaner approach would be to call a proc that performs the relevant logic (meaning proc implementation inheritently performs the `istype` checking), but between these three the common parent type is /atom and its pretty snowflaky for this xeno construction/deconstruction logic to be an atom proc that only has 3 implementations. If the common type between them was something like `/turf/closed/resin` then I would instead push for you to do it that way.

To explain what I mean by put the type checking into a var and to fix the bad inheritance logic issues with the existing code:
```js
#define RESIN_DOOR 1
#define RESIN_WALL 2
#define RESIN_WALL_MEMBRANE 3
/mob/living/carbon/xenomorph/proc/build_resin(atom/target, thick = FALSE, message = TRUE, use_plasma = TRUE, add_build_mod = 1)
//...
var/structure_kind = NONE
if(istype(target, /turf/closed/wall/resin/membrane)) // We must check this one before /turf/closed/wall/resin
	structure_kind = RESIN_WALL_MEMBRANE
else if(istype(target, /turf/closed/wall/resin))
	structure_kind = RESIN_WALL
else if(istype(target, /obj/structure/mineral_door/resin))
	structure_kind = RESIN_DOOR
//... (after the end of the proc)
#undef RESIN_DOOR
#undef RESIN_WALL
#undef RESIN_WALL_MEMBRANE
```

Then you can switch/if using `structure_kind`

PR #9412, opened 2025-06-01
Capturing Rebalance

Drulikar:
# About the pull request

This PR re-balances several aspects about capturing:
- Reverse bursting now drops the hauled mob (no stuns for either party) rather than gibbing the xeno
- Xenos can no longer tunnel while hauling someone
- An additional larva is granted if a nested human bursts (utilizing a new config `extra_larva_per_nested_burst` value)
- Struggling damage inflicted is capped to 35
- While below larva stage 4 and nested, bleeding will be cured, oxygen restored over 20, and brute/burn damage healed gradually if either are above max health in damage

Regarding force changes:
- Force of the item you struggle with is now limited to 35 but the damage is still random between its force/4 and the force.
- /obj/item/attachable/bayonet has a force of 25
- /obj/item/weapon/sword/machete has a force of 35
- /obj/item/weapon/gun/rifle/m41a has a force of 10
- /obj/item/weapon/gun/rifle/m41a with a bayonet has a force of 30
- /obj/item/weapon/gun/rifle/m4ra_custom has a force of 26
- /obj/item/weapon/gun/rifle/m4ra_custom with a bayonet has a force of 46

Things also under consideration:
- Gradual knockdown resistance to discourage tackle dragging (at the moment I think only runner can effectively do this, and it is tedious - but if players think this is the only way to transport safely now then I will likely implement it)
- Adjustments to burst timers as well as potentially different speeds depending on situation such as proposed in https://github.com/cmss13-devs/cmss13/pull/6682

Other notes:
-  Struggling w/ a weapon while hauled is now attack logged.
- The config value change: `extra_larva_per_burst` default changing to 0 is already what live has it configured to.

# Explain why it's good for the game

- Reverse bursting is a niche mechanic that doesn't occur very often so rarely is considered when hauling someone that is resisting, so now it doesn't cause additional sudden damage when the player is already going to be monitoring their health. It sort of made sense for a devoured marine, but not really any more for hauling.
- Not allowing tunneling when hauling is to make capturing more engaging for both parties: For the marine, they will either likely have a chance to start resisting the haul before hugger stun wears off before the xeno reaches their destination or they will be nested closer to the front allowing marines to more likely save fellow captured marines. For the xeno, they must strategize more how to effectively transport a capture (e.g. do I nest near the front, do I think I can route fast enough to the hive and know the way, can I handle the marine if they start resisting, etc).
[...]

PR #9570, opened 2025-06-11
Removes stun from a Marine when a Xeno chooses to un-haul them.

HumiliatedGoblin:
# About the pull request
Removes the two second stun given to a marine when they are dropped from haul by the player.

Bringing attention to an oversight to https://github.com/cmss13-devs/cmss13/pull/9412, where you can bypass the no tunnel balance by being able to infinitely stun someone by dropping and tackling spam due to the two second stun given.

Seems silly that when a marine breaks out, or if the timer runs out naturally, the marine gets no stun, but if the player chooses to drop the marine they get a free two seconds to get at least 3 tackles in (4 if they're a runner). 

I also believe letting a hugged marine instantly able to fight back once dropped in hive is in line with Drathek's original intention with the PR, with having to take into account if bringing a marine who's hugger stun has worn off to the hive can lead to abandonment or needing teamwork for potentially dangerous hosts. Also, queen screech will nullify any resistance, and gives a heightened importance of queen being inside the hive to deal with hosts.

I can add back the stun to have 0.2 seconds or a small number under 0.5 seconds to allow huggers to have a chance to attach once dropped, but seems unnecessary due to huggers being instant. (not taking player huggers into account, as they should be attaching to capable hosts that aren't already in possession of a hauled xeno, at least imo)

<details>

Idk i think it works, i cant test it alone

</details>

# Changelog
:cl: bostonthebear
balance: removes stun from choosing to un-haul
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/XenoProcs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/XenoProcs.dm
@@ -454,8 +454,6 @@
 	SPAN_XENOWARNING("We release [user] from our grip!"), null, 5)
 	playsound(src, 'sound/voice/alien_growl1.ogg', 15)
 	log_interact(src, user, "[key_name(src)] released [key_name(user)] at [get_area_name(loc)]")
-	if(stuns)
-		user.adjust_effect(2, STUN)
 	UnregisterSignal(user, COMSIG_MOB_DEATH)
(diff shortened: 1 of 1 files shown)


Drulikar:
There is no oversight that dropping a capture still stuns per usual. Tackle dragging however I specifically mentioned in said pr that may be changed in the future.

HumiliatedGoblin:
image_2025-06-11_185052568

PR #9578, opened 2025-06-12
Dropship Lock now takes 5 seconds

realforest2001:
# About the pull request

As title

I dislike seeing queens running in on a screech and one-click locking the dropship down then running off again. It is very frustrating to play against and to observe, and doesn't make much sense to me, especially with how long it takes to override it as a pilot.

<details>

</details>

# Changelog
:cl:
balance: Locking the dropship as Queen now takes 5 seconds.
/:cl:

### DIFF
--- a/code/modules/shuttle/computers/dropship_computer.dm
+++ b/code/modules/shuttle/computers/dropship_computer.dm
@@ -300,7 +300,7 @@
 		return
 
 	// door controls being overridden
-	if(!dropship_control_lost)
+	if(!dropship_control_lost && do_after(xeno, 5 SECONDS, INTERRUPT_ALL, BUSY_ICON_HOSTILE))
 		dropship.control_doors("unlock", "all", TRUE)
 		dropship_control_lost = TRUE
 		update_icon()
--- a/code/modules/shuttle/computers/dropship_computer.dm
+++ b/code/modules/shuttle/computers/dropship_computer.dm
@@ -300,7 +300,7 @@
 		return
 
 	// door controls being overridden
-	if(!dropship_control_lost && do_after(xeno, 5 SECONDS, INTERRUPT_ALL, BUSY_ICON_HOSTILE))
+	if(!dropship_control_lost && do_after(xeno, 3 SECONDS, INTERRUPT_ALL, BUSY_ICON_HOSTILE))
 		dropship.control_doors("unlock", "all", TRUE)
 		dropship_control_lost = TRUE
 		update_icon()


Diegoflores31:
wont this allow pilots to just launch while the queen is locking ?

specially synthetics could just crawl under and press launch in those 5 seconds

plus opening the cockpit already takes like 3 seconds

Ben10083:
> wont this allow pilots to just launch while the queen is locking ?
> 
> specially synthetics could just crawl under and press launch in those 5 seconds
> 
> plus opening the cockpit already takes like 3 seconds

This could be fixed with a variable that is set if a hijack attempt is in progress, preventing remote launch

ihatethisengine:
Hijacking Alamo by rushing it is a very risky but rewarding move. It's a very (probably the most) satisfying thing to do as the Queen. You're betting everything you have on such an attempt, trusting that the marines, and the DP in particular, will make grave mistake by allowing you to pull this off, and that your fellow xenos will follow you into a basically death trap. I played this situation many times on both sides, both as the queen and as the DP, both as a regular marine and a regular xeno. 

A good DP will never allow the Queen to lock the ship, unless the FOB has fallen mid-transit and Queen is camping the landing pad. It's not frustrating to play against. It's only frustrating when you have an inexperienced DP, but handholding won't improve this situation. The rule of thumb for DP is: keep the cockpit door locked and launch as soon as Queen attempts to enter the DS. Screech stun isn't long enough to pry open the cockpit door. So by following this rule you will get away EVERY time. Unless you are on cooldown. But that's another story. This is why you never cycle when FOB is under siege.

Making such a change will do nothing but make Queen experience worse. If Queen is able to lock the DS, the game is basically already lost for marines. Letting the Queen to wipe the DS is at least spectacular. If losing is fun, then losing the entire DS must be very fun as well. More than that, this is an interesting scenario for marines too. Retaking the DS feels great. Desperately standing your ground while the DP is trying to override the lockout is **cool**. Even if you fail and lose. This desperation is very special, and probably the most cinema-like scenario we have in the game. Why would we want to remove (or in this case, make **a lot** less likely) something cool? Just because someone got frustrated over a lost game? I don't think this is a good approach.

I think a better approach to this would be to make overriding the lockout faster. Then, if the Queen cannot sustain control over DS, marines would have a better chance to remove lockout and get away. I feel this solution would be more engaging for both sides.

DarkLordCabbage:
I think that xenos making such a bold move should be rewarded, however I agree that the time needed to fix the dropship is a bit long. Having marines block the way to the cockpit with their bodies while the PO desperately fixes the dropship which takes off just in time- very fun. Having everyone die because it takes an absurd amount of time to fix feels like marines never really had a chance in the first place.

Wintermote:
There's no risk/reward. It's all reward. Queen yeets into the dropship and touches it, gg every marine just died on the planet because it takes too long to unlock.

Queen takes 5 seconds to do it - still doesn't change the fact there's no reason not to yeet into the cockpit, as queen death as a loss pales in comparison to cementing the deaths of every marine still on the planet. It's actually one of the most riskless moves you can make as a queen player. Even dying in the process practically guarantees overall success.

Diegoflores31:
> There's no risk/reward. It's all reward. Queen yeets into the dropship and touches it, gg every marine just died on the planet because it takes too long to unlock.
> 
> Queen takes 5 seconds to do it - still doesn't change the fact there's no reason not to yeet into the cockpit, as queen death as a loss pales in comparison to cementing the deaths of every marine still on the planet. It's actually one of the most riskless moves you can make as a queen player. Even dying in the process practically guarantees overall success.

yeah except for the fact that you can be locked into the dropship and killed , DP can literally just launch the ship before you even get inside of it.
Queen diying is literally something that can entirely flip the round idk what you mean that the queen diying is "riskless"

Red-byte3D:
Why not just make the dropship computer start unlocking itself  post click instead of having a dp who has to stay there solves all issues

ihatethisengine:
> Why not just make the dropship computer start unlocking itself post click instead of having a dp who has to stay there solves all issues

This is how it works already, DP just makes it faster. It takes about 10 minutes for DS to unlock itself, DP can speed it up to 5 minutes.

kiVts (changes requested):
as discussed, 3 seconds would be more fair.

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -300,7 +300,7 @@
 		return
 
 	// door controls being overridden
-	if(!dropship_control_lost)
+	if(!dropship_control_lost && do_after(xeno, 5 SECONDS, INTERRUPT_ALL, BUSY_ICON_HOSTILE))
```
realforest2001:
```suggestion
	if(!dropship_control_lost && do_after(xeno, 3 SECONDS, INTERRUPT_ALL, BUSY_ICON_HOSTILE))
```

kiVts (approved):

Drulikar:
I don't see what this solves and just causes complications. Increase door timer a few seconds if you care instead.

PR #9755, opened 2025-06-18
Hijack Transit Changes

MeH0y:
# About the pull request

Reduces the flight time of xenomorphs hijacking from 3 minutes to 45 seconds and increases the bonus evolution during flight from 10 to 50 to allow the hive to still evolve in time. Also removes the marine announcement when the dropship departs from the ground, leaving only the imminent collision message

Little less of a slog for xenos, hijack being unexpected hopefully makes it more fun and the panic a little more real for the marines.

4 line change

# Changelog

:cl:
balance: reduced hijack crash time, removes marine announcement about incoming hijack
/:cl:


deathrobotpunch:
Maybe make the flight time random between 45 seconds and 3 minutes because why not

SomeoneYouProbablyKnow:
This does cut down on yhe time marines get to heal up arm up chill a bit and organize a defence. Maybe as a counterweight let them know where the ship hits. Because with elss time to actually get back on my feet id love to at least know if my prep room is overrun by xenos when i got to get mags

deathrobotpunch:
> This does cut down on yhe time marines get to heal up arm up chill a bit and organize a defence. Maybe as a counterweight let them know where the ship hits. Because with elss time to actually get back on my feet id love to at least know if my prep room is overrun by xenos when i got to get mags

You would have to remove the AA mechanics though, but I agree it does cut down on RP

SomeoneYouProbablyKnow:
Doesnt have to remove AA (even though aa is silly imo) but rather just tell em after the fact. The ship has collided at X

MeH0y:
im not too worried about marines having time to completely heal up and re-arm before hijack begins
while I also enjoy the brief moment of calm before the storm as a marine player to talk with my comrades, there will still be at least a two-minute window before the hijack actually begins and this rp is hopefully replaced with some real panic once the xenos hijack

IroquoisLou:
> Also removes the marine announcement when the dropship departs from the ground, leaving only the imminent collision message

If the xenos cut all the cameras after calling it down how are marines supposed to know if they're meta preparing or not?

realforest2001:
DNM for now. I fail to see a reason to just give the xenos a 'lets win faster' ticket. They already win by default of launching the dropship, removing the chance for any defence or coordination does feel thematic, but it's also not realistic as they would absolutely know their own dropship is approaching.

Frfor17:
You're a xenomainer who's just trying to make the game better for the xenos. Throw your playtime here to prove that this is not the case, because otherwise you are a xenomainer who improves the game just for yourself and your faction.

PR #9762, opened 2025-06-20
Minor Fix: No longer Click drag items whilst Hauled, Linter fix for SHOULD_CALL_PARENT in attack_self

MistChristmas:
# About the pull request
Adds a check for hauling when you click drag an item to hand; ~~Currently just returns doesn't check type.~~ Edit: Does check for gun or weapon. Doesn't work for click dragging weapon/gun to hand from non storage, I.E. M41 in back slot. Will work for pistols in a holster or other storage though.
Sprinkles more checks around for Alt+Clicking storage, internal storages, and defaults to opening the storage if they are hauled but don't have the allow inv haul flag. Instead of doing nothing.
Moves a check below a Linter needed thing so it doesn't break the Linter

We need Linters for linting
Probably shouldn't let items be grabbed that you can't get by clicking on them.
<details>

</details>

# Changelog
:cl: MistChristmas, Ediblebomb, Drathek
fix: Adds a haul check for click dragging, moves a Lint Proc to the top for Linting; Fixes the things that would get caught in the fixed lint. Left clicking storages whilst hauled will open them instead of doing nothing.
/:cl:


code/modules/clothing/clothing.dm
```diff
@@ -83,6 +83,8 @@
 /obj/item/clothing/attack_self(mob/user)
 	if(can_become_accessory)
 		convert_to_accessory(user)
+		return
+	return ..()
```
Drulikar:
```suggestion
	. = ..()
	if(can_become_accessory)
		convert_to_accessory(user)
```

code/modules/clothing/under/ties.dm
```diff
@@ -22,6 +22,7 @@
 	worn_accessory_slot = 1
 
 /obj/item/clothing/accessory/attack_self(mob/user)
+	. = ..()
 	if(can_become_accessory)
 		revert_from_accessory(user)
```
Drulikar:
```suggestion
	if(can_become_accessory)
		revert_from_accessory(user)
		return
	return ..()
```

Drulikar (changes requested):
attack_self implementations are reversed from how they need to be.

>When its a `/obj/item/clothing` (not any type of `/obj/item/clothing/accessory`) it should only call the accessory conversion code to become an accessory if it can.
>
>When its a `/obj/item/clothing/accessory` it should only call the accessory conversion code to become a `/obj/item/clothing` if it can.
>
>You achieve this by doing the code I shared originally. So a `/obj/item/clothing` (not a subtype of `/obj/item/clothing/accessory`) can only ever call `convert_to_accessory` and a `/obj/item/clothing/accessory` can only ever call `revert_from_accessory`

code/modules/clothing/under/ties.dm
MistChristmas: Whoopsie. Fixed.

PR #9302, opened 2025-05-18
Removes burst from the custom m4ra

Red-byte3D:
# About the pull request
Title

# Explain why it's good for the game

Scout is too strong, i dont think theres any other way to nerf scout in a way that doesnt completetly gut its usefulness, i believe that scould should win in one v ones yes, but with burst i believe its made too easy and basically impossible to counter unless the scout sucks.

I believe this change will not affect lower skill players, i think they are still going to just hive dive and get captured.

For the higher to mid level skill players (which, lets be honest, you dont need to be good as a scout player to get kills anyway) this just means they'll have to pick battles more carefully.

Now, before any obviously marine one sider player with below 75 hours comments on the pr

1.) Scouts strongest power is the ability to pick its fights, you have nvgs, binoculars and a cloak. This means you can engage and disengage in most circumstances, this just means you're going to have to pick your fights better instead of just decloaking offs screen and then immidietly nuke any tier 2 / 1 thats resting.

2.) This affects more then just the incin / high impact combo, the HV rounds for scout are pretty strong in themselves, having all the ap and damage that they have, albeit less then high impact they are still pretty good at nuking any tier 3 or even the Queen,  and yes just because only a few people know about this, does make it nerf worthy, something being rare does not justify it being overpowered.

3.) "But scout is a specialist, shouldnt he have an easy time killing tier 2's and ones." Yes, he should however again, burst fire makes this that basically literally the moment you catch something overextending, it means it dies. This pr will make it so it takes a bit longer to kill things, but if youre competent (able to fire at an enemy standing infront of you) nothing will really change, other then at most giving the lurker ONE chance to pounce away, if they can even react, also no other specialist can as easily and as reliably one v one or even one v three xenos as scout.

4.) "But you already nerfed ammo, hecking xenomorph dev why do you keep nerfing my fricking underdog marine faction" I think if this change is properly test merged, ammo per req buy and spawn can be increased again to compensate for the fact that you'll be spending more ammo per each encounter.

5.) "So if youre just going to give it more ammo what is the point in making it fire less" Because xenomorphs believe it or not are supposed to have hand holding, killing one tier two might not have an immidediate effect, but when people who are DECENT can perform very well and get up to 10-15 kills per each time they pick it, its going to have a big effect on the game in longer terms. This just basically means they have a fraction of a second to fight you off or atleast, try to get away. Yes its a team game, but remember its one xenomorph to 3 marines, not one spec vs 3 xenomorphs, refer to my previous point about scouts power.
[...]

PR #9766, opened 2025-06-20
Reduces A19 high velocity impact bullet damage from 40 to 30.

AnCopper:
# About the pull request

Reduces A19 high velocity impact bullet damage from 40 to 30. This addresses the current stunlock meta where these bullets deal excessive damage while maintaining high armor penetration, creating unfun instant-death scenarios.

The A19 HV bullet is currently overtuned, it combines devastating damage (40) with tier 10 armor penetration, essentially creating a "delete button" weapon that removes tactical gameplay. Players getting stunlocked and melted in seconds isn't engaging combat.
This nerf maintains the bullet's role as an armor-piercing round while preventing the bullshit instant-kill scenarios that make firefights feel like coin flips rather than skill-based encounters. 30 damage is still lethal but gives players actual time to react, take cover, and engage in meaningful tactical play instead of just getting deleted from full health.
The high penetration stays intact so it's still effective against armored targets, we're just removing the oppressive damage that makes these rounds feel broken.

Tested damage values in firing range against various armor types. Bullet still effectively penetrates heavy armor but no longer creates instant-death scenarios against unarmored targets.

# Changelog
:cl:
balance: reduced A19 high velocity impact bullet damage from 40 to 30 to address stunlock gameplay and promote tactical combat
/:cl:

### DIFF
--- a/code/datums/ammo/bullet/rifle.dm
+++ b/code/datums/ammo/bullet/rifle.dm
@@ -167,7 +167,7 @@
 	name = "A19 high velocity impact bullet"
 	flags_ammo_behavior = AMMO_BALLISTIC
 
-	damage = 40
+	damage = 30
 	accuracy = -HIT_ACCURACY_TIER_2
 	scatter = -SCATTER_AMOUNT_TIER_8
 	penetration = ARMOR_PENETRATION_TIER_10


Somniworld:
Calling 40 damage "devastating" is being rather dramatic : it seems to be the exact same damage done by M41As : 
image

The stun element is strong, but remember that the "super imba cloaking specialist that kills you super hard with his crazy rifle" can be countered...  by not being alone, just like you would with lurkers. It's also to be noted that the gun has only 12+1 bullets, this means that while you can absolutely body a xeno with it (if they've overextended, and aren't alone, and aren't on hive weeds, and aren't being healed by the queen), you will eventually have to swap to another ammo type, or refill through requisitions.  

From personal experience, using Impact at T3s is really silly : you're shooting the 800-something HP murdermachine with a super-expensive magazine that will at most down it's HP by half, assuming 100% of your shots hit. Remember that you lose the stun component against T3s. Lowering the damage further makes them thus completely irrelevant, since you might as well just use a regular M4RA AP mag (same damage, same AP value, _cheaper_).

My point is that what really makes these rounds feel bad is that they stun, but, like, so do Lurkers and runners and warriors?  Furthermore, Burst fire on the rifle was already removed, heavily gimping the rifle. At this point, this feels a lot like the dead horse is being beaten with a stick.

Red-byte3D:
This is no longer needed as per #9302

Detective-Google:
With the recent changes to the scout's primary, I'm going to shelf this for now, give it some time to see how the recent change works, and if you still feel the weapon is over-performing we can look into it again.

PR #9798, opened 2025-06-23
CBRN ERT changes

Lagomorphica:
# About the pull request
completely changes the loadout of all CBRN ERT members

the CBRN ERT is incredibly cool and very rare, but also unlike most 'rare' things is very weak - you spawn with either only a flamer or only a mk2, and the medic spawns with only a pistol. this attempts to reflavor them into all having flamethrowers while not making them overpowered, and gives them their own skills instead of using PMC skills (also nerfing their skills in the process)

they have no attachments, no AP, etc when they arrive - they will have to make use of their ability to firewalk if they want to truly be effective. this also removes some other things from them such as master endurance (end 3), master engineering, et cetera. this also fixes the standard CBRN rifleman not even being able to use most of his equipment due to spawning with a medkit pouch and no medical skill by swapping it out

# Changelog

:cl:
balance: CBRN ERT has had a passover and lost some of its more questionable skills in exchange for better armaments
/:cl:

### DIFF
--- a/code/datums/skills/uscm.dm
+++ b/code/datums/skills/uscm.dm
@@ -539,7 +539,7 @@ CIA
 CBRN
 ---------------------
 */
-/datum/skills/cbrnrifleman
+/datum/skills/cbrn_rifleman
 	name = "CBRN Rifleman"
 	skills = list(
 		SKILL_ENGINEER = SKILL_ENGINEER_TRAINED,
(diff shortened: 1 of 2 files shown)


code/datums/skills/uscm.dm
```diff
@@ -533,3 +533,63 @@ CIA
 		SKILL_FIREARMS = SKILL_FIREARMS_EXPERT,
 		SKILL_ANTAG = SKILL_ANTAG_AGENT,
 	)
+
+/*
+---------------------
+CBRN
+---------------------
+*/
+/datum/skills/cbrnrifleman
```
Detective-Google: snake_case your paths/vars

Detective-Google (changes requested):
this makes these guys pretty nutty fraggers as well, which I'm not telling you to change now but will need to be considered. testing works

Lagomorphica:
don't have any good ideas to balance this atm, will revisit later

PR #9804, opened 2025-06-24
Language prefix to '!', Minor ReOrg of related lists, Provost to '#l', pass prefixes as lowercase, Remove uppercase instances. + Misc.

MistChristmas:
# About the pull request
~~Removes some Radio say prefixes from hyperdyne and CIA, add comments reserved chars for language use for future stuffs.~~
EDIT: The _remove that grid reference_ option:

- Changes langauge prefix char to one not used by radio channels. In this case: `!` ~~$~~ Smart guy forgot about "$20"

Resolves: #9803

EDIT 2: I love scope creep:

- Reorganised Languages and Radio Channels to be mostly alphabetical apart from Special interactions such as intercom or Left hand. Or appolo etc etc.
- Changed provosts channel from .r and :r to #l as .r and :r are used for Right Hand.
- Removed all uppercase instances as we already half forced the fucking things to lowercase??? Ok.
- Added a new constant for LANGUAGE_PREFIXES, so we can display them on TGUI Say.
- Added some missing channels to RADIO_PREFIXES - I think this was just Yautja, and CIA

EDIT 3: Please end me:

- Added missing channel colours for tgui say. (CIA, Hyperdyne, Yautja for some reason..)
- Break then unbreak Multibroadcasting. Cause I forgot to lowercase less and then more things

Fixes an issue for Synths, ~~creates one for Hyperdyne and CIA? Eeeeeeh I mean they can access they're radio channels.
Someone should really reserve a prefix char specifically just for languages rather than having it be the same as the radio ones.~~
Moving Languages to a unique prefix will stop any overlap and also free up 0-9 for Radio Channels.
ReOrg so my eyes don't bleed looking for channel assigments.
Only lowercase because we want consistency between upper and lowercase at the very least. - It's a start.
.l and :l Get used for hands stuff which no one uses but we preserve anyway. #l is free so we steal it.
<details>
<summary>Here have an outdated testing Video</summary>

https://github.com/user-attachments/assets/798a2f79-871c-492f-bb95-700c3ae84a8f

</details>

# Changelog
:cl: MistChristmas, Drathek
add: Languages use '!' prefix instead of sharing prefixes with Radio Channels. Provost channel changed to '#l'.
code: Re-org RADIO_PREFIXES, langauges, and department radio keys into mostly alphabetical order. Remove Uppercase instances. Add missing colours for tgui say.
qol: Languages pop up in TGUI-say if you switch to them whilst not being on a radio channel
/:cl:


MistChristmas:
FWIW I could change this to just change the language prefix to use one(s) not used by Radio, but that would be a noticeable change for many people.

tgui/packages/tgui-say/constants.ts
```diff
@@ -20,580 +20,383 @@ export enum LineLength {
...
+    id: 'hellhound',
+    label: 'Hellhound',
+  },
+  '$q ': {
+    id: 'hivemind',
+    label: 'Hivemind',
+  },
+  '$_ ': {
+    id: 'primative',
+    label: 'Primative',
```
Drulikar: Misspelled

Drulikar (changes requested):
Additionally as mentioned on discord:
- [x] human, silicon, and whisper don't strip language if the speaker can't speak an attempted language they don't have (patch provided)
- [ ] xenos still parse languages via . and : rather than the new $
- [ ] xeno say proc likely should mimic how human and silicon use `parse_language` rather than it having a hardcoded implementation.

code/modules/mob/living/carbon/xenomorph/say.dm
```diff
@@ -20,71 +18,58 @@
 		if(!findtext(message, "*", 2)) //Second asterisk means it is markup for *bold*, not an *emote.
```
Drulikar, line 19: Please move `var/prefix = copytext(message, 1, 2)` up above this line, and have this line also use it

tgui/packages/tgui-say/helpers.ts
```diff
@@ -49,22 +49,26 @@ function setWindowVisibility(visible: boolean, scale: boolean): void {
   });
 }
 
-const CHANNEL_REGEX = /^[:.#]\w\s/;
+const CHANNEL_REGEX = /^[$:.#]\w\s/;
```
Drulikar:
```suggestion
const CHANNEL_REGEX = /^[!:.#]\w\s/;
```

code/modules/mob/living/carbon/xenomorph/say.dm
```diff
@@ -20,71 +18,58 @@
...
+		hivemind_speak = FALSE
+	if(hivemind_speak && can_hivemind_speak)
+		if (findtext_char(message, " ") < 4)
+			message = capitalize(trim(copytext(message, 3)))
+		else
+			message = capitalize(trim(copytext(message, 2)))
+		if(!message)
+			return
+
+		// Automatic punctuation
```
Drulikar:
```suggestion
	if(prefix == ";")
		message = capitalize(trim_left(copytext(message, 2)))
		hivemind_speak = TRUE
	else if(prefix == "." || prefix == "#" || prefix == ":" || prefix == ",")
		message = capitalize(trim_left(copytext(message, 3)))
		hivemind_speak = TRUE

	if(!message)
		return

	if(hivemind_speak && can_hivemind_speak)
		// Automatic punctuation
```

Whitespace isn't guaranteed to to be after the prefix so we need to do it this way.

code/modules/mob/living/carbon/xenomorph/say.dm
```diff
@@ -20,71 +18,58 @@
...
+			if(!(copytext(message, -1) in ENDING_PUNCT))
+				message += "."
+
+		hivemind_talk(message)
+		return
+
+	var/datum/language/speaking = parse_language(message)
+	if(speaking)
+		verb = speaking.speech_verb
+		message = capitalize(trim(copytext(message,3)))
```
Drulikar:
```suggestion
		message = capitalize(trim_left(copytext(message, 3)))
```

code/modules/mob/living/carbon/xenomorph/say.dm
```diff
@@ -20,71 +18,58 @@
...
 	else
-		if(!speaking || isnull(speaking))
-			for(var/datum/language/L in languages)
-				if(L.key == "0")
-					verb = L.speech_verb
-					speaking = L
-					forced = 1
-					break
+		speaking = GLOB.all_languages[LANGUAGE_XENOMORPH]
```
Drulikar:
```suggestion
		speaking = get_default_language()
```
This allows you to set hivemind as default

code/modules/mob/living/carbon/xenomorph/say.dm
```diff
@@ -20,71 +18,58 @@
...
+		verb = speaking.speech_verb
+		message = capitalize(trim(strip_language(message)))
 
 	if(!(speaking.flags & HIVEMIND) && HAS_TRAIT(src, TRAIT_LISPING)) // Xenomorphs can lisp too. :) Only if they're not speaking in hivemind.
```
Drulikar:
```suggestion
	// Xenomorphs can lisp too. :) Only if they're not speaking in hivemind.
	if((!(speaking.flags & HIVEMIND) || !can_hivemind_speak) && HAS_TRAIT(src, TRAIT_LISPING))
```

code/modules/mob/living/carbon/xenomorph/say.dm
```diff
@@ -20,71 +18,58 @@
...
-		if(!speaking || isnull(speaking))
-			for(var/datum/language/L in languages)
-				if(L.key == "0")
-					verb = L.speech_verb
-					speaking = L
-					forced = 1
-					break
+		speaking = GLOB.all_languages[LANGUAGE_XENOMORPH]
+		verb = speaking.speech_verb
+		message = capitalize(trim(strip_language(message)))
```
Drulikar:
```suggestion
		message = capitalize(trim_left(strip_language(message)))
```

code/modules/mob/living/carbon/xenomorph/say.dm
```diff
@@ -20,71 +18,58 @@
...
-
-	if(!message || stat)
+	if(!message)
 		return
 
 	// Automatic punctuation
-	if(client && client.prefs && client.prefs.toggle_prefs & TOGGLE_AUTOMATIC_PUNCTUATION)
+	if(client?.prefs?.toggle_prefs & TOGGLE_AUTOMATIC_PUNCTUATION)
 		if(!(copytext(message, -1) in ENDING_PUNCT))
 			message += "."
```
Drulikar, line 69:
```suggestion
	if((speaking.flags & HIVEMIND) && can_hivemind_speak)
		hivemind_talk(message)
		return

```

Drulikar (approved):

PR #9813, opened 2025-06-25
Increases BE log visibility and clarifies it slightly.

Detective-Google:
# About the pull request

Generally speaking battlefield executions are something staff are always going to want to be somewhat aware of due to the relatively high chance they become tickets/reports later.

this PR moves it to the FF log and reformats it slightly. 

:cl:
admin: Logging for BEs is updated
/:cl:

### DIFF
--- a/code/datums/ammo/bullet/bullet.dm
+++ b/code/datums/ammo/bullet/bullet.dm
@@ -72,8 +72,7 @@
 	user.count_niche_stat(STATISTICS_NICHE_EXECUTION, 1, firing_projectile.weapon_cause_data?.cause_name)
 
 	var/area/execution_area = get_area(execution_target)
-
-	msg_admin_attack(FONT_SIZE_HUGE("[key_name(usr)] has battlefield executed [key_name(execution_target)] in [get_area(usr)] ([usr.loc.x],[usr.loc.y],[usr.loc.z])."), usr.loc.x, usr.loc.y, usr.loc.z)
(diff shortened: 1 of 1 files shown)


code/datums/ammo/bullet/bullet.dm
```diff
@@ -72,8 +72,7 @@
 	user.count_niche_stat(STATISTICS_NICHE_EXECUTION, 1, firing_projectile.weapon_cause_data?.cause_name)
 
 	var/area/execution_area = get_area(execution_target)
-
-	msg_admin_attack(FONT_SIZE_HUGE("[key_name(usr)] has battlefield executed [key_name(execution_target)] in [get_area(usr)] ([usr.loc.x],[usr.loc.y],[usr.loc.z])."), usr.loc.x, usr.loc.y, usr.loc.z)
+	msg_admin_ff("[key_name(user)] [ADMIN_JMP(user)] [ADMIN_PM(user)] has <b>battlefield executed</b> [key_name(execution_target)] [ADMIN_JMP(execution_target)] [ADMIN_PM(execution_target)] at [get_area(user)] ([user.loc.x],[user.loc.y],[user.loc.z]) using [fired_from].")
```
Drulikar:
```suggestion
	msg_admin_ff("[key_name(user)] [ADMIN_JMP_USER(user)] [ADMIN_PM(user)] has <b>battlefield executed</b> [key_name(execution_target)] [ADMIN_JMP(execution_target)] [ADMIN_PM(execution_target)] at [get_area(user)] ([user.loc.x],[user.loc.y],[user.loc.z]) using [fired_from].")
```

More useful this way because it'll keep the killer (the CO) position always relevant and the executed is then where it happened.. Previously it would just be two static locations next to eachother.

Drulikar (approved):
With changes:
image

PR #9834, opened 2025-06-27
Added terrain-specific versions of various loadout items

Generalcamo:
# About the pull request

This adds terrain-specific variations of the helmet netting, rain cover, and marine boots to the loadout.

This removes the need to constantly change your loadout based on the set map, if you want to blend in with the surroundings. Now the game will handle this for you.

<details>

image

</details>

# Changelog

:cl:
add: Added terrain-specific versions of rain covers, helmet nettings, and marine combat boots to the loadout.
/:cl:


Zenith00000:
Let me make some snow head covers for you.

Also not sure I agree with the boots being the way they are, the grey ones are supposed to be for urban, but I guess they could work for snow as well, but black probably works better for snow.

Staykeu:
Would it not be better to have camo conforming items be the default that every marine can vend from their machines? I feel like marines having items that don't match up with the current map's terrain is more of a personalization thing (thus leaning more towards a character's spawn loadout) than having items that do match with the current map's terrain.

PR #9847, opened 2025-06-28
fixes brute rockets runtiming and deleting

Detective-Google:
# About the pull request

BRUTE rockets detonate at max range, passing `null` to prime()'s `atom` arg, which causes runtimes due to `atom.var` existing in the admin logs, meaning that fired at max range the rocket would just delete itself without exploding since `prime()` early returned. now if `atom` is null we use an alternate log that doesn't require it.

I tested this it works
image
offending runtime

# Changelog
:cl:
fix: BRUTE rockets no longer malfunction at max range.
/:cl:


cuberound:
thanks, did not expect that

cm13-github:

cm13-github:

Drulikar (approved):

PR #9876, opened 2025-07-01
Pounces now need a windup to go between z levels.

Red-byte3D:
# About the pull request

Pounces now need about 2 seconds to go up or down a z level

I think just being unable to chase anything around because they can just click the parkour ability is going to be very frustrating for marines, and really strong for xenos.

This should make it so you can take a risk for safety.
im willing to budge on maybe runners being able to do this without a windup, will need to talk

<details>

</details>

# Changelog
:cl:
balance: Pounces now take two seconds if you're aiming for another z level.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
@@ -388,7 +388,7 @@
 
 	if(A.z != X.z)
 
-		if (!do_after(X, 3 SECONDS, INTERRUPT_NO_NEEDHAND, BUSY_ICON_HOSTILE))
+		if (!do_after(X, 2 SECONDS, INTERRUPT_NO_NEEDHAND, BUSY_ICON_HOSTILE))
 			return
 
 		X.throw_atom(A, distance, throw_speed, X, launch_type = LOW_LAUNCH, pass_flags = pounce_pass_flags, collision_callbacks = pounce_callbacks, tracking=TRUE)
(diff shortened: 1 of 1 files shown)


RainbowStalin:
Does this affect both up and down? Because I can maybe get pouncing up being somewhat high effort, but down as well? Won't this just kill off multiz pounce ambushes?

DarkLordCabbage:
I have to say I disagree with this, I think the trade off that comes with marines being able to shoot down a z-level is some xenos being able to get up and retaliate quick fast and in a hurry. Z-levels can be lethal for xenos due to the amount of gunfire that marine can rain down. In turn, xenos offensive capabilities on higher Zs are pretty much spit/pounce. I think adding a cooldown would make pouncing pretty much impossible. In my mind, this change would affect the balance of the game too much, more than it's intending too. 

Also, ambusher xenos swarming over and under catwalks is 🔥

cuberound:
codewise  this is called too early, before checking plasma status and such, also this looks like if you pounce twice with the extra throw atom

the general line that I was given to go along when balancing xenos on multiz is that small xenos are supposed to be real agile but not big xenos, this is needed for any t3 ability that can travarse z level as crusher and prea pounce, perhaps rav one works too? not sure right now as I did not test, t3s are supposed to have hard time crossing z levels. The idea for multiz is, that the main combat is supposed to happen on main z level, but you are mainly at risk from being pounced from below when you are not controling the main level, then it is fair punishment for risky play, pouncing top down just feels like normal ambush. When it comes to t3s being able to pounce down I think it should perhaps be added, but I am not fighting over that part. Perhaps some super short windup to pounce up for small xenos could be fine just as an indicator of what is going on and to give chance to look/counteract up but I am speeking 0.5 sec top (as with having to stop moving before clicking I assume it will be effectivly much longer anyway). for comparison the currently tested xeno climb up time on a wall is 1.5 seconds. but I just noticed that you have interupt no need hand so perhaps 1 sec to 1.5 sec pounce up could be fine, none for down for small xenos that is certain

Red-byte3D:
Closing since theres another pr

PR #9572, opened 2025-06-11
[Re] Gardener Drone: Changes and Improvements.

Venuska1117:
Thanks to **Archivator** for comming up with marine-sided fruit description.
Thanks to **Adan Stafford | JAW5** for creating better description for spore fruit that distinguished them from other fruits.

Re-open PR: #9327
_(closed by stale bot)_

# About the pull request

**Improves General Gardener gameplay and understanding by:**

- Gives more information when examinating fruits/hardened weeds, like how much they heal, what they do, how long they last etc. Marines cannot see those infomations anymore, they now get basic descriptions for fruits.
- Gardener now pickup fruits 50% faster than any other caste, this only apply to unrooting and feeding fruit, it does not apply to consuming fruit from ground. (it makes sense, because they specialize in gardening so they do it quicker, like how medics heal themself faster than others)
- Returns gardener fruit counter as status tab info, change made from #8647 broke original counter on button.
- Resprites item fruits to have twitchy roots, so its more easy to recognize when they are uprooted from not mature ones.
- If marine uproots fruit (or find one) they now get slight "hint" for them to maybe bring this to research.
- Ability to place fruits next to resin holes, don't worry, you **cannot** place them on top to hide them.
- Gives ability to instantly mature any fruit with resin surge, in exchange it will apply dynamic cooldown based on fruit growth using previous code maturity + cooldown time.
- Gives visual balloon alert when consuming fruit or when fruit expires, so player knows when effect of fruit ends.
- Gives audiable clue when fruits expires, so player know by intuition that text that popped up is expiration one. (if they read too slow)

Also, don't say giving xenos detailed information is broken, marines have ability to open weapon status tab, and see how much damage it does, how far it shoot, armor penetration etc. Its more QoL change than anything else.

# Explain why it's good for the game

Gardener is one of best strain made, but it was made with its own drawbacks, vague fruit descriptions, recent cooldown addition from #8647 to buttons broke fruit counter, and marines having access to general description what fruit does without even playing xenomorph should not take place. (also they should not know by lore reasons)

Thanks to new changes, xeno players will have access to more detailed information about fruits and what they do when examinating them.

Also makes Gardener more capable at its job of growing fruits, instead of it be side character that just supply them.

# Testing Photographs and Procedure

<details>
<summary>> Click here <</summary>

Rooted and unrooted sprites (they have animations):
image

Xeno-sided information when examine fruits:
image
image
image
image
image

Only works when examine node:
image

</details>

# Changelog

:cl: Venuska1117
[...]

PR #9877, opened 2025-07-01
Prevent Fruits balloon alert spam for every not related Destroy() proc call.

Venuska1117:
# About the pull request

This is a fix for issue caused by #9572, i did make tests for shield fruit, becouse there many xenos using shield abilities, but i forgot about xenos that do healing abilities (i forgor)

also i forgot to credit fixing issue on #9572:

Fixes: #8660

Less balloon alert spam caused by healing, like healing drones, queen heal, warrior slash healing, some bullets effects triggering balloons, it should be triggered only by fruits.

# Changelog
:cl: Venuska1117
fix: Fruit balloon alert now only trigger after consuming fruit, instead for every not related Destroy() proc call.
/:cl:

### DIFF
--- a/code/datums/effects/heal_over_time.dm
+++ b/code/datums/effects/heal_over_time.dm
@@ -6,14 +6,14 @@
 	var/ticks_between_heals = 1
 	var/heal_each_process = 0
 
-/datum/effects/heal_over_time/New(atom/A, heal_amount = 0, healing_time = 5, time_between_heals = 1, limb_name = null, from_fruit = FALSE)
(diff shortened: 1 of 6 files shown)


code/datums/effects/heal_over_time.dm
```diff
@@ -36,7 +38,7 @@
 /datum/effects/heal_over_time/Destroy()
 	if(affected_atom)
 		var/mob/living/carbon/xenomorph/xeno = affected_atom
-		if(istype(xeno))
+		if(from_fruit)
```
Drulikar:
```suggestion
		if(istype(xeno) && from_fruit)
```

Git-Nivrak (changes requested):
Change `from_fruit` to something like `show_baloon_alert`

PR #9907, opened 2025-07-05
Lurker Pounce Rework Revived

private-tristan:
# About the pull request
Revival of #9034, which previously staled out and died.

This PR makes quite a few changes to lurker's pounce, most notably allowing lurkers to **do a runner level pounce when uncloaked** alongside having their cloak be interruptible/disrupted by taking damage. Here's a full list of the changes:

<details>
<summary> Full list of changes </summary>
Buff - Lurker pounce now acts like runner pounce when not cloaked, stunning for 1s.

> This pounce will be identical to runner pounce, giving the lurker enough time for 1 slash before leaving (due to lurker's slower attack speed), and a 2nd slash right as the marine gets up (assuming endurance 1). Lurker pounce stun/freeze unchanged if you're cloaked

Buff - Lurkers heal a bit faster when resting, from 1.5 to 2

> Previously they healed (relatively) slower when resting compared to runner's 1.75

Nerf - Lurker pounce CD increased to 7s (from 6)

> gives the lurker a little delay so that they can't immediately pounce the marine after they get back up, assuming the marine has endurance 1, there will be a roughly 1.5s~ delay before the lurker can pounce again.

Nerf - Landing a cloaked pounce or slashing a marine while invisible will **massively** increase your pull multiplier

> Kills pouncedragging, get a friend to help you.

Nerf - Lurker pounce now decreases invis.

> punishes lurkers for missing their pounce more, especially if the marine they attempted to pounce didn't know a lurker was trying to pounce them. Lurkers can still run away, but they'll get, yaknow, shot at as they do, instead of invisibly scampering off into the darkness.
> Regular cloaked pounces still fully uncloak you.

Nerf - Lurkers taking damage while cloaked will cause their visibility to increase, if visibility increases too much, your invisibility is cancelled with a 50% cloak refund. the speed boost from being cloaked will linger for 4 seconds

Nerf - Decreases the slow-per-slash on floored targets from 3.5s to 2.5s per slash.

> with a cloak pounce, the marine will have roughly 8s of slowdown when they get up (from about 13s of slowdown)

QoL - Bumping no longer causes the generic "invis lost" message alongside the "you bumped and lost invis" message.

</details>

It came up in CM discussion and I felt like throwing this PR together, I've generally felt that lurker hasn't been the most engaging to play, as you're forced to play around the 20s cooldown for invis so that you can access your stunning pounce.

With lurkers always uncloaking when pouncing, missing a pounce means that the lurker loses access to their escape tools while also probably being directly adjacent to the marine that they tried to pounce. Meaning that the marine is immediately made aware of the lurker (if they didn't notice/have a MD) and can fight back during the 8/10 seconds that the lurker has no escape tools outside of "crippling slash and run away"

Strob (Before leaving CM) put it as:

> adds skill expression to lurker / opens more opportunities to punish them for missing pounce ( while also allowing them more room to outplay without cloak )

As for the slowdown-per-slash change, it's a bit absurd that getting lurker pounce gave you 13 seconds of slowdown, so I toned it down a bit.
<details>
<summary>Demo Videos</summary>

Damaging a lurker will disrupt their cloak

https://github.com/user-attachments/assets/444f83b8-3619-4494-8d0f-eb909c00c76e

Uncloaked pounce (ignore the "We use our invisibility to empower our pounce!" msg, fixed).

https://github.com/user-attachments/assets/9805b259-82f1-4d7a-88cd-337ea3bbdfbb

Full Combo Demo

https://github.com/user-attachments/assets/b71c3e0c-36c6-4607-ba84-02e90695b2aa

Pouncing while your cloak is too damaged will cause it to be ended.

https://github.com/user-attachments/assets/34bc3794-9cbf-4b6a-ae99-68b4d2ac5096

</details>

# Changelog
:cl:
balance: Lurker's Pounce now acts like a runner pounce if the lurker isn't cloaked, stunning for 1s. Cloak pounce stays at 2.5s
balance: Lurker's Pounce now makes you more visible when missed
balance: Lurker's Pounce now massively decreases pull speed for a few seconds, nerfing pouncedragging
balance: Lurkers taking damage while cloaked will have their visibility increased, becoming too visible will cause cloak to be cancelled. The speed boost linger for 4 seconds after the cloak is disrupted.
balance: Lurker's Pounce CD increased to 7s (from 6s)
balance: Decreased Lurker's slow-per-slash on floored targets from 3.5s per slash to 2.5s per slash. A full cloak combo will generally leave the marine with 8s of slowdown
balance: Increased lurker's health regen mult while resting from 1.5 to 2 (for reference, runner has 1.75). This translates to about 13 hp per tick while resting.
qol: Lurker's no longer get 2 different "invis loss" texts when bumping into a visible marine.
code: Replaced single letter vars
/:cl:


Drulikar:
stale exempt until reviewed and tested (again I guess)

PR #9915, opened 2025-07-05
Fire axes can open unpowered doors after a wind up

DarkLordCabbage:
# About the pull request

A fire axe can be used on an unpowered door to force it open/shut after a wind up, similar to using a bayonet on a gun to open/shut a door. Has the exact same wind up time

Fire axes right now are kind of useless, despite the fact that we have cool "break in case of emergency" glass cases scattered around the almayer with them in it. This should solve that, restore the fire axe's original purpose as a door opener, and give people another chance at escaping an emergency such as hijack. Now a nurse with no crowbar isn't trapped, they can escape via breaking the glass with a scalpel! 

<details>
I tested it but was too lazy to film it

</details>

# Changelog

:cl:
add: Fire axes can now open unpowered doors after a 3 sec wind up, same as a bayonet attached to a gun
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -634,6 +634,25 @@ GLOBAL_LIST_INIT(airlock_wire_descriptions, list(
 	else if(HAS_TRAIT(attacking_item, TRAIT_TOOL_MULTITOOL))
 		return attack_hand(user)
 
+	else if(istype(attacking_item, /obj/item/weapon/twohanded/fireaxe))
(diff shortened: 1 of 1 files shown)


JOHNBALD123:
fire axes could already open up unpowered doors though? why add a cooldown now

DarkLordCabbage:
because i'm an evil xenodev who wants to add arbitrary cooldowns to everything... no, I'm just stupid and failed to see that fire axes have IS_PRY_CAPABLE_SIMPLE when _wielded_

PR #9920, opened 2025-07-05
Removes z level requirement from markers

Red-byte3D:
# About the pull request

Removes z level check from xenomorph info markers.

This wont change much other then adding the ability of being able to mark stuff that arent exactly in the same z level, so marking a lower z level from solaris ridge middle level to the upper floors.

I dont think theres any other way to do this without adding 200 lines of code, atleast not that i know of

<details>

</details>

# Changelog

:cl:
code: Xenomorph markers can now be put into different z levels
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
@@ -275,10 +275,8 @@
 		return FALSE
 	var/turf/target_turf = get_turf(A)
 
-	if(target_turf.z != X.z)
-		to_chat(X, SPAN_XENOWARNING("This area is too far away to affect!"))
-		return
-	if(!X.hive.living_xeno_queen || X.hive.living_xeno_queen.z != X.z)
+
+	if(!X.hive.living_xeno_queen)
 		to_chat(X, SPAN_XENOWARNING("We have no queen, the psychic link is gone!"))
 		return
 


cuberound:
you can possibly add check if the two z levels have the same traits, this is in place so that you can not mark from planet to ship and back but ship and ground have different traits

code/modules/mob/living/carbon/xenomorph/Powers.dm
```diff
@@ -197,9 +197,6 @@
 		to_chat(src, SPAN_NOTICE("We must have a meaning for the mark before you can make it."))
 		hive.mark_ui.open_mark_menu(src)
 		return FALSE
-	if(target_turf.z != src.z)
-		to_chat(src, SPAN_NOTICE("We have no psychic presence on that world."))
-		return FALSE
```
Git-Nivrak, line 202: Do the same here instead of removing

code/modules/mob/living/carbon/xenomorph/abilities/general_powers.dm
```diff
@@ -275,10 +275,11 @@
...
 	var/turf/target_turf = get_turf(A)
 
-	if(target_turf.z != X.z)
-		to_chat(X, SPAN_XENOWARNING("This area is too far away to affect!"))
+	if(!SSmapping.same_z_map(X.loc.z, target_turf.loc.z))
+		to_chat(X, SPAN_XENOWARNING("Our mind cannot reach that far."))
 		return
-	if(!X.hive.living_xeno_queen || X.hive.living_xeno_queen.z != X.z)
+
+	if(!X.hive.living_xeno_queen)
```
Git-Nivrak: You removed the check for Queen being on the same map as well

Red-byte3D: This isnt really important though, it was on purpose, people should still be able to mark if they get on almayer while queen is groundside etc

PR #9930, opened 2025-07-06
MK30 Pilot Visor - New flippable visors for PO's

Zenith00000:
# About the pull request

Pretty simple PR that adds six new visors attachments, the black one being the default one which comes pre-installed with every MK30 helmet. You can choose one extra one from your clothing vendor. They serve no real function other then to look drippy, oh I guess they give you flash protection as well, but otherwise just visual.

May add more if there's a demand for it...

For some strange reason the PO helmets never came with visors, even though in the original "Aliens" movie they had them, basically it's good for the game for two reason - One is "Drip" and Two is it's lore-accurate and makes sense.

<details>

PO_visors_small
vendor
line_up

image

</details>

# Changelog
:cl: Zenith
add: Added six new MK30 flight visors to the PO's clothing vendor.
add: Added black MK30 flight visors to all M30 helmets by default.

/:cl:


nauticall (approved):
pixels 👍

Nanu308 (approved):
mapping fine, but gonne need to wait for MultiZ Almayer

code/modules/clothing/head/helmet.dm
```diff
@@ -1034,9 +1039,11 @@ GLOBAL_LIST_INIT(allowed_helmet_items, list(
 	flags_inventory = BLOCKSHARPOBJ
 	flags_inv_hide = HIDEEARS|HIDETOPHAIR
 	specialty = "MK30 tactical"
-	built_in_visors = list(new /obj/item/device/helmet_visor, new /obj/item/device/helmet_visor/medical/advanced)
+	built_in_visors = list(new /obj/item/device/helmet_visor)
+	inserted_visors = list(new /obj/item/device/helmet_visor/po_visor)
```
Red-byte3D, line 1043: Code is fine but do we really need to be able to insert two of them? switching through them makes it kinda silly

Zenith00000: It's because I removed the built in med-hud, and added it as an option to purchase on the PO wardrobe instead, basically  just so you could remove them at will, if you wanted to also just wear the glasses version and not have both that + the med hud on the visor, having to switch between them. I hope you get what I'm trying to say, but yeah I mean it seems more complicated but it's kind of making it more simple this way.

If this doesn't work I can just change it back so the med-hud is built in again, and it only accepts 1 inserted visor or so, but yeah I just thought this might be better.

Zenith00000: Ah never mind, I've just reverted the changes, you can only insert one visor now, might be less confusing. Although I've always found it strange how you can pick the "Medical HUD Glasses" on top of already having an inbuilt HUD, but I suppose it means you can wear the visor down with the glasses and still have a medical HUD up at all times this way.

PR #9933, opened 2025-07-06
Item associated cursors

Red-byte3D:
# About the pull request
Pretty simple, the objects you pick up become your cursor.

These will require sprites i just lay down the framework for blundir, should be pretty easy to just assign the path to the item and it will directly do the rest.

Requested by blundir and i also think its neat, it can be disabled by choosing to not have custom cursors through the preferences tab.

<details>

</details>

# Changelog

:cl:
add: New cursors that mimick the item you're holding, you can enable or disable this through preferences enable custom cursors
/:cl:


nauticall (changes requested):
This is neat, but also is an accessibility problem. I'd recommend finding a way to overlay a more traditional cursor on top of it so people still know where they're clicking. Otherwise you're gonna spend 30 seconds fumbling a C4 because a newscaster or something is on the wall.

code/game/objects/items.dm
```diff
@@ -407,6 +409,9 @@
 	if(drop_sound && (src.loc?.z))
 		playsound(src, drop_sound, dropvol, drop_vary)
 	src.do_drop_animation(user)
+	if(src.mouse_pointer)
```
Detective-Google, line 412:
```suggestion
	if(mouse_pointer)
```

code/game/objects/items.dm
```diff
@@ -407,6 +409,9 @@
...
+		if(user.client?.prefs.custom_cursors)
```
Detective-Google, line 413: a user cannot be assumed to have a client associated, user?.client?.etc

code/game/objects/items.dm
```diff
@@ -423,8 +428,20 @@
 	if(pickup_sound && !silent && src.loc?.z)
 		playsound(src, pickup_sound, pickupvol, pickup_vary)
 	do_pickup_animation(user)
+	if(src.mouse_pointer)
```
Detective-Google, line 431:
```suggestion
	if(mouse_pointer)
```

code/game/objects/items.dm
```diff
@@ -423,8 +428,20 @@
...
+		if(user.client?.prefs.custom_cursors)
```
Detective-Google, line 432: mob/user can't be assumed to have a client

code/game/objects/items.dm
```diff
@@ -423,8 +428,20 @@
...
+			update_mouse_pointer(user, TRUE)
 	return TRUE
 
+
+///Turns the mouse cursor into a crosshair if new_cursor is set to TRUE. If set to FALSE, returns the cursor to its initial icon.
+/obj/item/proc/update_mouse_pointer(mob/user, new_cursor)
+	if(!user.client?.prefs.custom_cursors)
```
Detective-Google, line 439: user can't be assumed to have a client

Detective-Google (changes requested):
seems okay besides this but will probably need some testing-- otherwise as sammy said there needs to be an indication of where the specific pixel the cursor is clicking is, something like this for the C4
image
(though this is more spritainer territory so I'd ask them for specifics)

PR #9934, opened 2025-07-07
Xeno melee targeting changed to random organs

aceroyuno:
# About the pull request

The purpose of this change is to make Xeno melee attacks randomly target a limb zone

Quite simply with the new found strength of double burst being merged, the need for the foot fracture meta to overcome the adversity of the Marine's side's strength no longer exists. And therefore Xenos, whom do not understand human biology, should not be able to specifically target a human's weakest, or least armored, body part. (This does not affect taking off masks from mobs like Predators)

<details>

https://github.com/user-attachments/assets/7d00a205-298a-4deb-ae41-cbcf13689d14

</details>

# Changelog
:cl: aceroy
balance: Aliens can no longer target specific limb zones for melee when attacking
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
+++ b/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
@@ -92,9 +92,7 @@
 
 			attacking_xeno.flick_attack_overlay(src, "slash")
 			var/obj/limb/affecting
-			affecting = get_limb(rand_zone(attacking_xeno.zone_selected, 70))
-			if(!affecting) //No organ, just get a random one
-				affecting = get_limb(rand_zone(null, 0))
+			affecting = get_limb(rand_zone(null, 0)) //Grab a random organ
 			if(!affecting) //Still nothing??
 				affecting = get_limb("chest") //Gotta have a torso?!
 


Drulikar:
You all are obsessed with double burst yet marines are still winning most games right now. This is not necessary.

aceroyuno:
> You all are obsessed with double burst yet marines are still winning most games right now. This is not necessary.

Considering the foot frac meta has been around for over a year now, I don't think you can label it an obsession. and could you please elaborate on how this PR is not necessary?

Drulikar:
You are focused entirely on an incorrect assumption that double burst has tipped balance so bad that you need to drastically increase every xeno's time to kill, remove all autonomy in where their attacks land (negating nearly every reason to use the paperdoll in any offensive capacity) unlike their opposing force, and leave it to RNGesus for their attacks to ever have a lasting effect that will make that marine consider if they are capable enough to still stay in the fight or if they should return to lick their wounds before they go down entirely.

How about you explore breaking the light armor meta instead.

aceroyuno:
A couple things:

1. Do marines actually get head shot bonus damage? Cause otherwise I am completely unaware of how the marine paperdoll plays any role in shooting or meleeing xenos.

2. You are making an incorrect assumption that this has to do with double burst and not my year+ long crusade against the foot fracturing meta.

3. Genuinely, how would you want me to go about fixing the light armor meta?

Bonus. Would it not be possible to buff slash damage to accommodate the change to random attacks that could also then increase overall xeno strength?

Drulikar:
> Do marines actually get head shot bonus damage? Cause otherwise I am completely unaware of how the marine paperdoll plays any role in shooting or meleeing xenos.

- I'm not aware of head shot bonus damage against xenos
- Marines generally prefer to target chest to reduce FF damage since even with light armor, it is the most armored for marines, but this is counteracted by base miss chance by limb so marines aiming feet can be a valid strategy too to mitigate FF damage
- Additionally for non-xenos (so pertains to FF in this case), all zones except for chest have a base miss chance
- When you hit chest on a xeno, there is a slight higher chance for blood splatter damage
- When you don't direct click, your targeted zone for projectiles is randomized (up until 4788 base miss chance also applied for the targeted zone on xenos)
- Even if you direct click a xeno, there is still a chance for the targeted zone to be randomized
- If the zone wasn't randomized this is deemed a direct hit (e.g. lever action has a streak modifier using that signal)
- When not deemed a direct hit the base miss chance applies for the randomized zone

So in summary, its complicated, but marines paperdoll selection against xenos is primarily concerning FF both via splash and with bullets actually hitting friendlies.

> You are making an incorrect assumption that this has to do with double burst and not my year+ long crusade against the foot fracturing meta.

You opened your explanation of why its good for the game with "Quite simply with the new found strength of double burst being merged"

Regardless this PR is boiling down to change for the sake of change. As proposed it would be a large nerf to xenos, and as I stated previously, this is not necessary.

> Genuinely, how would you want me to go about fixing the light armor meta?

I'm not all knowing arbiter of ideas. The statement was to get you to think of what to focus your efforts on instead. The most compelling way IMO I have been proposed regarding this though is regarding modular armor. But ultimately it'd likely have some semblance of risk vs reward.

> Bonus. Would it not be possible to buff slash damage to accommodate the change to random attacks that could also then increase overall xeno strength?

Not really worth the change of trying to balance it. I would much rather us move away from RNG for combat as much as possible, not lean in to it more.

I am not going to keep entertaining this back and forth here. I would encourage you to discuss with other contributors a direction to take that more people can agree on - or at the very least to see the discussions you opening the PR made. But as far as I know you are not in the CM discord assuming you still use Aceroy.

PR #9938, opened 2025-07-07
Minor GAU fixes

cuberound:
# About the pull request

this is a fix as it sets values to how I inteanded and how I said I am changing them in the GAU rework. it makes spread of both ammo types 3 instead of 4 on AP ~~and ammo hits random bodypart to prevent broken organs on every shot~~ (should already be the case)

makes GAU ammo work as inteanded ~~and spreads out its damage to otehr bodyparts~~

<details>

</details>

# Changelog
:cl:
fix: reduces AP GAU ammo spread to 3 tile from 4 as intended
/:cl:

### DIFF
--- a/code/modules/cm_marines/dropship_ammo.dm
+++ b/code/modules/cm_marines/dropship_ammo.dm
@@ -187,7 +187,7 @@
 			if(iscarbon(explosion_effect))
 				var/mob/living/carbon/bullet_effect = explosion_effect
 				explosion_effect.ex_act(EXPLOSION_THRESHOLD_VLOW, null, cause_data)
-				bullet_effect.apply_armoured_damage(directhit_damage,ARMOR_BULLET,BRUTE,pick(ALL_LIMBS),penetration)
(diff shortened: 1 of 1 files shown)


code/modules/cm_marines/dropship_ammo.dm
```diff
@@ -187,7 +187,7 @@
 			if(iscarbon(explosion_effect))
 				var/mob/living/carbon/bullet_effect = explosion_effect
 				explosion_effect.ex_act(EXPLOSION_THRESHOLD_VLOW, null, cause_data)
-				bullet_effect.apply_armoured_damage(directhit_damage,ARMOR_BULLET,BRUTE,null,penetration)
+				bullet_effect.apply_armoured_damage(directhit_damage,ARMOR_BULLET,BRUTE,pick(ALL_LIMBS),penetration)
```
Drulikar: Is there actually any benefit to this - can you point to where this wouldn't already result in random limbs? If so it likely needs to be fixed for that proc not just in this situation.

Drulikar (approved):

PR #9987, opened 2025-07-14
Temporary maturity for early game queens

Drulikar:
# About the pull request

This PR adds temporary maturity for early game queens (first 20 minutes of play). Basically when in the area that is the hive_location (hive_core) or within 10 tiles of it, the queen will have a 1 minute timer that refreshes every Life tick when either taking damage (bumps the timer by 15s currently) or when remaining near the hive_location. The maturity timer appears in status panel and also has a balloon alert at about 7s remaining.

For clarity, being shot won't ever mature a queen - it can only *extend* an *existing* timer. Even has an assertion that will call CRASH ending the proc if it somehow could try to do that. 

This is to mitigate the issue where an immature queen at the start of the game cannot deal with an early game rush from survivors.

I don't want it where people aren't free to try to do things e.g. seek out xenos. The risk of rushes in itself should combat xenos thinking predrop is boring and they shouldn't be on alert. There should be a threat. HOWEVER the aspect that doesn't sit well with me is that an immature queen would literally be better de-evolved as a drone than to be fighting anything. I didn’t want to just give early queen free maturity though so we don’t have the opposite problem of queen rushing survivors.

<details>

Main testing (though I forgot to turn off second view to see status panel)
https://youtu.be/FpDwqYGDeEY

Also balloon alerts:
alert

</details>

# Changelog

:cl: Drathek
balance: Early game queen is now temporarily mature when near the hive core (1 minute timer that refreshes while close or when getting shot)
code: Added missing undefs for queen and removed a few redundant checks for refreshing the choose resin ability 
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/Queen.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Queen.dm
@@ -437,9 +437,8 @@
 		// Also update the choose_resin icon since it resets
 		if(istype(action, /datum/action/xeno_action/onclick/choose_resin))
 			var/datum/action/xeno_action/onclick/choose_resin/choose_resin_ability = action
-			if(choose_resin_ability)
-				choose_resin_ability.update_button_icon(selected_resin)
-				break // Don't need to keep looking
+			choose_resin_ability.update_button_icon(selected_resin)
+			break // Don't need to keep looking
 
 	if(hive.dynamic_evolution && !queen_aged)
 		queen_age_timer_id = addtimer(CALLBACK(src, PROC_REF(make_combat_effective)), XENO_QUEEN_AGE_TIME, TIMER_UNIQUE|TIMER_STOPPABLE)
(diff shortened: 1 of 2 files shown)


foob1:
They're gonna plop the hive core right where they predict the survs are gonna try to hold, cap all the survs, and then move the hive core once the survs are dealt with. Kind of ridiculous survs are getting nerfed because one round in the past 90 days on lowpop they managed to effect the round by wiping a hive.

Ammoniacres:
I'm curious as to how this wouldn't be abused by placing hive cores near loot or hold-spots pre-drop to harass survivors, given a mature queen attacking a hold-spot at any state of the survivor gameplay loop is generally enough to secure the maximum possible captures.

Red-byte3D:
> I'm curious as to how this wouldn't be abused by placing hive cores near loot or hold-spots pre-drop to harass survivors, given a mature queen attacking a hold-spot at any state of the survivor gameplay loop is generally enough to secure the maximum possible captures.

Because you cant put cores outside caves, The only map this should realistically be a problem on is solaris ridge, and thats not really a "problem" because if people are diving into what would be usual "hive locations" thats a risk they are taking for better loot.

PR #10021, opened 2025-07-18
Spec priority preference for initial spawn

Drulikar:
# About the pull request

This PR adds a preference to character setup where you can optionally specify what order of specs you prefer to redeem if possible. This choice currently only affects initial spawn, not late join or ERT. You specify the order of specs depending on the order you select them in the checkmark tgui window. When the options are exhausted it behaves as normal (for example if you pick only sniper, and sniper isn't available, no choice will be made and if you need to instead go to the vendor or open your spec kit to choose). For presets that don't normally spawn a spec kit item, you still must go to the vendor which will automatically redeem your spec kit on first interaction.

I did make some minor changes to the checkmark input making it more like tgui_input_list to handle associated lists instead of erroring (though the list it returns will only ever be the keys, not key value pairs or values). I also corrected some `usr` usage in cm_vending (all `ui_act` should use the `ui.user` not `usr`).

The order role authority equips characters is now randomized since the order players equip now matters for specs, and it doesn't just need to be who connected first getting priority.

Eliminates the need for rushing to open a kit or rush to a vendor to choose a kit because now roundstart will choose the specs immediately on spawn as far as they have their priority preference set. The order ultimately is the order of `GLOB.player_list` (or `GLOB.human_mob_list` for non-distress signal/WO game modes) which now is shuffled for more fairness.

<details>

Distress signal/Extended:
https://youtu.be/4gbqcbt1g2A

WO:
https://youtu.be/5vU_MpIKo3c

</details>

# Changelog
:cl: Drathek
add: Added a preference for spec priority which will attempt to automatically select a specialization on round start (does not apply to ERT or late join currently). The order choices are made will be the priority order.
code: Corrected some incorrect usr usage in cm_vending
code: tgui_input_checkboxes can now handle associated lists (ignores values currently)
/:cl:


Drulikar:
> Looks good to me but I'd like to get someone with regex knowledge to check the expression in checkboxes.dm out.

Its not a new regex. See https://github.com/search?q=repo%3Acmss13-devs%2Fcmss13%20%40%7B%22(%5B%5E%5Cu0020-%5Cu8000%5D%2B)%22%7D&type=code

https://github.com/search?q=repo%3Atgstation%2Ftgstation%20%40%7B%22(%5B%5E%5Cu0020-%5Cu8000%5D%2B)%22%7D&type=code

image

PR #10035, opened 2025-07-21
Xenomorph Banned Allies

realforest2001:
# About the pull request

Adds a variable list that allows for restrictions on what factions each hive can ally with. There is also an admin toggle variable to override the list.

Helps to avoid confusion with certain things. Currently the only hives affected are Tutorial and Tamed hives. (No one can ally with them, and vice versa.

<details>

</details>

# Changelog

:cl:
code: Changed each xeno hive name into a define.
add: Added restriction variables on what factions each Xeno hive can ally with.
/:cl:


code/modules/mob/living/carbon/xenomorph/hive_status.dm
```diff
@@ -1134,7 +1138,7 @@
 	return FALSE
 
 /datum/hive_status/alpha
-	name = "Alpha Hive"
+	name = FACTION_XENOMORPH_ALPHA
```
Red-byte3D, line 1163: Maybe not allowing alpha & base ally since we dont allow it by rules is a good idea by default?

realforest2001: Alpha and Base aren't strictly prohibited from working together, they're just strictly prohibited from working with humans.

Red-byte3D: Huh. I was told they arent supposed to ally together, since they are by default atlast in lore at a civil war

hry-gh (approved):

PR #10314, opened 2025-08-10
Adds a button to secure storage from the inside

### LINKED ISSUE #6944: Locked into Engineering Secure Storage
### Testmerges

 Local commit: f32eda41db9dfec7ce2a2b118b2587ba641f7ade TGS version: 6.6.0 DMAPI version: 5.9.0

### Round ID

Round ID: 23491 

### Description of the bug

Basically, as CE you can unlock Secure Storage to get the mats out using a button in the CE office. If you are in Secure Storage when the Code Level goes from Red to Blue or Green, the shutters shut and lock you in.

image

Shouldn't get locked in with no way out except level changes if you can let yourself in.

1. Unlock Secure Storage.
2. Stand in it.
3. Have someone reset the Code Level.
...

### Issue Bingo

ItsVyzo:
# About the pull request
Adds a button to secure storage from the inside

no longer getting stuck in the case of code changes fixes #6944

<details>

</details>

# Changelog

:cl: LTNTS
maptweak: Adds a button to secure storage from the inside
/:cl:


Steelpoint:
The mapping change does what it states, I can't see any issues.

I might suggest changing the buttons access to ENGINEERING (7) instead of CE (6). My rational is that the outer door has ENGINEERING access assuming the CE has opened the shutters blocking the outer door. So its possible a MT can get themselves trapped in a similar issue to the bug report and have no way to escape the room. I would not think there is a balance issue with this proposed change.

ItsVyzo:
> The mapping change does what it states, I can't see any issues.
> 
> I might suggest changing the buttons access to ENGINEERING (7) instead of CE (6). My rational is that the outer door has ENGINEERING access assuming the CE has opened the shutters blocking the outer door. So its possible a MT can get themselves trapped in a similar issue to the bug report and have no way to escape the room. I would not think there is a balance issue with this proposed change.

resolved

PR #10358, opened 2025-08-14
Makes Trijent locks C4-able and acidable

TheGamerdk:
# About the pull request

All (non-elevator) locks on Trijent have been made destructable by both sides

Fixes inconsistent behaviour. Some of these were already destructable.

Should especially help with this choke, which was previously barely openable:

image

<details>

</details>

# Changelog

:cl:
add: Added something
del: Removed old things
qol: made something easier to use
balance: Trijent locks can now be blown and melted open.
/:cl:

### DIFF
--- a/maps/map_files/DesertDam/Desert_Dam.dmm
+++ b/maps/map_files/DesertDam/Desert_Dam.dmm
@@ -3153,7 +3153,8 @@
 /obj/structure/machinery/door/poddoor/almayer/locked{
 	dir = 4;
 	id = "dam_checkpoint_northeast";
-	name = "\improper Checkpoint Lock"
+	name = "\improper Checkpoint Lock";
+	unacidable = 0
 	},
 /turf/open/floor/warning/north,
 /area/desert_dam/interior/lab_northeast/east_lab_east_entrance)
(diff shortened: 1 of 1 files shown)


Wintermote:
Can the padlocks be opened with crowbars? That was something inconsistent with them on this map vs most other maps with the same kind of doors.

Bibby0110:
> Can the padlocks be opened with crowbars? That was something inconsistent with them on this map vs most other maps with the same kind of doors.

No, they're a subtype that can't get crowbarred for some reason

Steelpoint:
From my look over. This change also now allows xenomorphs to pry open the doors if the APC powering the doors is depowered or destroyed. 

Otherwise the change does what is stated. I would suggest adding the prying information to the changelog.

Detective-Google (commented):
I think adding a /desert_dam (or /trijent) subtype of the locks with the desired attributes is going to be better than instance editing the locks-- which is a big part of why they're so inconsistent in the first place(the same type being used but instanced in many different applications)

Nanu308 (approved):
Subject to be changed back, but approved for now map wise

See Google's comment tho.

PR #10408, opened 2025-08-17
reverse bursting out of a xeno now gibs it again

byondsucks:
# About the pull request

reverse bursting out of a xeno gibs it like it used to before double larva

Xenos can easily circumvent this by dropping the hauled marine and tackling them down to make them drop their knife, i believe that such a level of handholding is very unnecessary when it can be avoided by taking 5 seconds to tackle the marine down again. Usually even after reverse bursting out of a xeno you get tackled down really fast, making it not punish the xeno at all. 

# Changelog

:cl:
balance: Xenos now gib when reverse bursted out of.
/:cl:


MeH0y:
mechanic literally only kills new players that don't know it exists until their round is spontaneously over

Drulikar:
Don't undo intended changes. https://github.com/cmss13-devs/cmss13/pull/9412

PR #10491, opened 2025-08-25
Adds Non-Binary As A Gender Option

Ediblebomb:
# About the pull request
Adds Non-Binary as a gender option in character creation, it can also roll on randomly generated humans with a 5% (Subject to maint approval, I guess. PvE was fine with that number.) chance.
This should probably work, but be on the lookout for weird stuff with naming and pronouns in messages from objs and such.
(Sort of a port of [#1167](https://github.com/cmss13-devs/cmss13-pve/pull/1167), does it count if it never got merged and got closed so I could open this instead?)
Character stuff and diversity good.
# Changelog

:cl:
add: Adds Non-Binary as gender option in character creation and random generation
ui: Adds a body presentation button to the PredPicker:tm:
/:cl:

### DIFF
--- a/code/modules/mob/new_player/preferences_setup.dm
+++ b/code/modules/mob/new_player/preferences_setup.dm
@@ -1,10 +1,15 @@
 //The mob should have a gender you want before running this proc. Will run fine without H
 /datum/preferences/proc/randomize_appearance(mob/living/carbon/human/H)
 	if(H)
-		if(H.gender == MALE)
-			gender = MALE
-		else
-			gender = FEMALE
+		switch(H.gender)
+			if(MALE)
+				gender = MALE
+			if(FEMALE)
+				gender = FEMALE
+			if(PLURAL)
+				gender = PLURAL
+			else
+				gender = pick(47.5;MALE, 47.5;FEMALE, 5;PLURAL)
 
 	skin_color = random_skin_color()
 	body_type = random_body_type()
--- a/tgui/packages/tgui/interfaces/PredPicker.tsx
+++ b/tgui/packages/tgui/interfaces/PredPicker.tsx
@@ -24,6 +24,7 @@ import { HairPickerElement } from './HairPicker';
 type PredData = {
   name: string;
   gender: string;
+  body_presentation: string;
   age: number;
   hair_style: string;
   skin_color: string;
@@ -88,6 +89,7 @@ export const PredPicker = () => {
   const {
     name,
     gender,
+    body_presentation,
     age,
     hair_icon,
     hair_style,
@@ -121,7 +123,9 @@ export const PredPicker = () => {
                 </LabeledList.Item>
                 <LabeledList.Item label="Gender">
                   <Button onClick={() => act('gender')}>
-                    {capitalizeFirst(gender)}
+                    {gender === 'plural'
+                      ? 'Non-Binary'
+                      : capitalizeFirst(gender)}
                   </Button>
                 </LabeledList.Item>
                 <LabeledList.Item label="Age">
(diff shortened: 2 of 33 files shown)


funky-monke:
wtf is this woke shit dude. speed close now

Ediblebomb:
> wtf is this woke shit dude. speed close now

Funky you're going to get gitbanned because they think you're serious, shush

Also oh boy failed checks I love failed checks so much

Ediblebomb:
> i'd deny 5% purely out of naming weirdness and player not giving consent to that

The random gens just pick between the name list and pick from there, which probably shouldn't be an issue, for the latter point it could be a preference that's off by default? I'm not entirely sure that's feasible but with enough bashing my head into a brick wall (How I code) I might be able to make it work.

> also i'm really not sure about pred edits, you really should talk to pred senator about it first

I would, but I'm permabanned from the discord, if they don't want it then that can be relayed. I wanted to cover everything, if things need to be shaved off from there it's less work than needing to add something new.

Blundir:
> The random gens just pick between the name list and pick from there, which probably shouldn't be an issue, for the latter point it could be a preference that's off by default? I'm not entirely sure that's feasible but with enough bashing my head into a brick wall (How I code) I might be able to make it work.

I don't really think it's worth a whole preference entry

realforest2001:
I'm all for equality and such, but this is not necessary and we've managed just fine for the last decade.

sippykot:
what a grim reasoning to close this pr

hry-gh:
reopened pending wider discussion

Drulikar:
I see little reason why this can't be on PVP. It would just be on PVE otherwise. They were encouraged to PR it upstream by me for this reason. Shouldn't be something forced on someone, byond has always had plural as a gender and this is just making it selectable for humans.

Jesters-Ballad:
I can't speak to anything regarding the overall wish of the CM maintainers and those who make the big decisions about what gets merged and added to the game. I can say however, that this addition would make one of our very hard working mentors happy to have the option. And the kind of people this update would make upset, I do not particularly wish to play alongside. For the sole reason of it making that one person happier, and who knows how many other people who want their digital avatars to match themselves better, this getting in would be really wonderful.

Willzadl:
non binary predators feel very un-yautja-like to me

Lukas-Seltz:
I look at the gender option like you would when you're signing up for the military, which you have to put your biological gender due to there are gender speciffic issues that can come up. While I like having diversity, the need for a whole option is not needed. If you really want to do this, add an option for gender identity and change one to biological gender. But overall, I do not really think this is needed... Because ultimately you can RP as non-binary anyway, even without having the Gender

L0pz:
> I look at the gender option like you would when you're signing up for the military, which you have to put your biological gender due to there are gender speciffic issues that can come up.

From my understanding, the examine text purely states IC information one would know about the person and has nothing to do with official uscm records or anything of the sort

code/modules/mob/new_player/preferences_setup.dm
```diff
@@ -1,10 +1,15 @@
...
-			gender = FEMALE
+		switch(H.gender)
+			if(MALE)
+				gender = MALE
+			if(FEMALE)
+				gender = FEMALE
+			if(PLURAL)
+				gender = PLURAL
+			else
+				gender = pick(47.5;MALE, 47.5;FEMALE, 5;PLURAL)
```
hry-gh: i hate this syntax please use pick_weight

```suggestion
				gender = pick_weight(list(
						MALE = 47.5, FEMALE = 47.5, PLURAL = 5)
				)
```

L0pz:
I think you gotta set the pr to open

Ediblebomb:
> I think you gotta set the pr to open

Slightly confusing way to go about things when there is a "Re-request Review" button, but there.

tgui/packages/tgui/interfaces/PredPicker.tsx
hry-gh, line 1: this needs reverting if the rest has been reverted

hry-gh (changes requested):

PR #10500, opened 2025-08-26
Fixes curtain code to actually support transparent curtains

### LINKED ISSUE #10498: Curtains Bugs
### Testmerges

none

### Round ID

_No response_

### Description of the bug

in maps that have curtains, specifically Hybrisa. 

1. You can see through closed curtains when you haven't clicked on them
2. Once you click on them. Their sprite are still closed, but they obscure the view
3. Clicking on them finally opens the curtain and unobscures the view

Closed curtains should obscure the view

you're suppose to open the curtains when clicking on closed curtains

1. Find a curtain in any map (forgot if they apply to all maps, but specifically in Hybrisa)
2. Click on them to "open" the curtain, but they now obscure the view
3. Click on them again to finally open the curtain and seeing through the window
...

### Issue Bingo

FloranOtten:
# About the pull request

Fixes #10498 

Curtain code works based on opacity. A closed curtain spawns with an opacity of 0, and a closed sprite. An open curtain spawns the exact same way, but applies a toggle() immediately. 

In adding transparent curtains, the original author attempted to define an alpha value for them, and then make the curtains transparent if this value is present. Unfortunately. all curtains inherit an alpha value, which meant all curtains were being made transparent. This is fixed by clicking the curtain, which cycles the opacity, and thus sets the correct sprite for the opacity. However, this also cycles the opacity for transparent curtains, which makes them opaque. 

This PR changes curtain code to enable curtains to open/close without relying on opacity to store that value.

Open curtains shouldn't start off closed, and closed curtains shouldn't start off open. And transparent curtains should probably be transparent, too.

<details>

Before:
image
image

After:
image
image
colorable_transparent:
image

</details>

# Changelog
:cl:
fix: Curtains block view without prior encouragement.
fix: Transparent curtains remain transparent, even when opened and closed.
/:cl:

### DIFF
--- a/code/game/objects/structures/curtains.dm
+++ b/code/game/objects/structures/curtains.dm
@@ -5,6 +5,8 @@
 	layer = ABOVE_MOB_LAYER
 	opacity = TRUE
 	density = FALSE
+	var/open = FALSE
+	var/transparent = FALSE
 
 /obj/structure/curtain/open/New()
 	..()
@@ -29,13 +31,15 @@
 	return XENO_ATTACK_ACTION
 
 /obj/structure/curtain/proc/toggle()
-	opacity = !opacity
-	if(opacity)
-		icon_state = "[initial(icon_state)]"
-		layer = ABOVE_MOB_LAYER
-	else
+	open = !open
+	if(!transparent)
+		opacity = !opacity
+	if(open)
 		icon_state = "[initial(icon_state)]-o"
 		layer = OBJ_LAYER
+	else
+		icon_state = "[initial(icon_state)]"
+		layer = ABOVE_MOB_LAYER
 
 /obj/structure/curtain/shower
 	name = "shower curtain"
@@ -96,6 +100,7 @@
 	name = "blinds"
 	icon_state = "colorable_transparent"
 	alpha = 200
+	transparent = TRUE
 
 // Open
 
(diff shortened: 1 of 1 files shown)


Zenith00000:
This will break the "transparent" curtains which you're "supposed to see through". Just so you know.

Like I said on the bug report, there's two types of curtains -

Opaque curtains in which you can never see-through

&

Semi or fully transparent  Curtains, which you should always be able to see through, they're purely cosmetic really.

FloranOtten:
> This will break the "transparent" curtains which you're "supposed to see through". Just so you know.


Transparent curtains were already broken in that case; Without this PR, clicking on them reverts them to opaque/closed curtains.

Is the transparent behaviour intended, and if so, is this described anywhere?

FloranOtten:
I think I see the intent of the original PR; if(alpha) was supposed to check for an alpha value which is only defined on transparent curtains, then set them to have no opacity. 

This does not work, because curtain opening/closing is entirely based on opacity code, but the original bug that made all curtains start off transparent probably obscured this.

I will turn this into a refactor of curtains to support transparency.

Zenith00000:
> > This will break the "transparent" curtains which you're "supposed to see through". Just so you know.
> 
> Transparent curtains were already broken in that case; Without this PR, clicking on them reverts them to opaque/closed curtains.
> 
> Is the transparent behaviour intended, and if so, is this described anywhere?

Yeah they were, I couldn't figure out a way to get them to be transparent and open/close while still being a curtain, but yeah probably an easy fix. Thanks.

PR #10502, opened 2025-08-26
Survivor lighting tweak

DonCycon:
# About the pull request
Replaces flashlights in survival pouches with lanterns.
Replaces flashlights in firesuit closets with a flare packet.

Visibility for survivors has felt awful ever since the lighting has been updated to the new system.

Lanterns are essentially a bit more dim version of the marine combat flashlight, a tile of difference in radius from the regular flashlight.

Flare packets replacing flashlights in firesuit closets is a good way to introduce a decentralized and fairly limited supply of flares on all the maps without mapping them in individually, while this affects marines on the almayer i believe the impact of this wont be noticable  as marines already have ample supply of both flares and flashlights shipside even during hijack.

This gives survivors a robust but temporary lighting source alongside a dilemma between picking another magazine for your weapon or a packet of flares.

This PR should also adress some cases of lowpop survivor rounds where you are doomed to darkness because restoring lights is not possible due to nobody spawning as an engineer, while also making exploring the colony a bit more rewarding and more fun.

I don't think this will affect survivors restoring power, as flares are limited, can be melted and lose their brightness with time, while a powered colony lets you use the utility of shocked doors and vendors alongside the lights.
<details>
<summary>Regular flashlight first picture, Lantern second picture.obraz
obraz</summary>

</details>

# Changelog
:cl: Don Cycon
balance: Replaces flashlights in firesuit closets with M94 flare packets.
balance: Replaces flashlights in survival pouches with lanterns.
/:cl:


Red-byte3D:
Why does the role that has limited supplies to survive need better gear

ItsVyzo:
You should probably create a subtype for the fire closet instead of replacing all current closet's flashlights with flares

DonCycon:
> Why does the role that has limited supplies to survive need better gear

Flares in firesuit lockers create new natural points of intrest on all maps for both marines and survivors.
(even new maps!)
A lantern with a single tile more of visbility gives survivors a slightly bigger window to react, survivor always has been notoriously hard and unfair, and i don't think this takes away from it at all other than slightly putting the bar higher for cases where you are being hit from complete darkness unable to really do anything.
While it isn't necessary, it is something that is nice and adds depth across the board, think of it like tailstab for an example.
I believe this small change, while niche, would make the game just a bit more fun for everyone with the opportunities it creates.

sirjantkot1337:
Disguised nerf. Flashlight is somewhat better because it can be put on a gun.

Mister-moon1:
Survivor intentionally has not great items (or should do), including lighting, if you cant find a location to hold that has some form of light by the time you're getting attacked then something has gone wrong somewhere, each map has atleast 1 somewhat good location like this

realforest2001 (approved):

PR #10523, opened 2025-08-28
Nuclear bomb no longer explodes after a last-second disarm.

zzzmike:
# About the pull request

Currently, it's possible to disarm the bomb at the last second (after the music starts but before the bomb explodes) - the game throws the usual announcements about the bomb being disarmed, but the bomb still explodes as there isn't a last-minute check in the code. Happened in round 29620 for example. This will stop it exploding.

If someone disarms the bomb at the last second, they shouldn't explode. Or don't allow them to disarm it at the last second at all. I'm going with the first option.

<details>

</details>

# Changelog
:cl:
fix: nuclear bomb no longer explodes if disarmed at the last second
/:cl:

### DIFF
--- a/code/game/machinery/nuclearbomb.dm
+++ b/code/game/machinery/nuclearbomb.dm
@@ -390,7 +390,7 @@ GLOBAL_VAR_INIT(bomb_set, FALSE)
 	timeleft = initial(timeleft)
 	explosion_time = null
 	for(var/mob/player in world)
-    	player << sound(null, channel=5) // stops the explosion sound in case bomb is disabled while it's playing
+		player << sound(null, channel=5) // stops the explosion sound in case bomb is disabled while it's playing
 	announce_to_players()
 
 /obj/structure/machinery/nuclearbomb/proc/explode()
--- a/code/game/machinery/nuclearbomb.dm
+++ b/code/game/machinery/nuclearbomb.dm
@@ -411,6 +411,11 @@ GLOBAL_VAR_INIT(bomb_set, FALSE)
 			shake_camera(current_mob, 110, 4)
 
 	sleep(10 SECONDS)
+	if(GLOB.bomb_set == FALSE)
+		update_minimap_icon()
+		stop_processing()
+		update_icon()
+		return FALSE
 
 	var/list/mob/alive_mobs = list() //Everyone who will be destroyed on the zlevel(s).
 	var/list/mob/dead_mobs = list() //Everyone that needs embryos cleared
--- a/code/game/machinery/nuclearbomb.dm
+++ b/code/game/machinery/nuclearbomb.dm
@@ -391,6 +391,10 @@ GLOBAL_VAR_INIT(bomb_set, FALSE)
 	explosion_time = null
 	for(var/mob/player in world)
 		player << sound(null, channel=5) // stops the explosion sound in case bomb is disabled while it's playing
+	for(var/mob/current_mob as anything in GLOB.mob_list)
+		var/turf/current_turf = get_turf(current_mob)
+		if(current_turf?.z == z && current_mob.stat != DEAD && current_mob.shakecamera > world.time)
+			stop_camera_shake(current_mob) // stops explosion camera shake in case bomb is disabled while it's shaking
 	announce_to_players()
 
 /obj/structure/machinery/nuclearbomb/proc/explode()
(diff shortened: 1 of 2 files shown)


matttheficus:
Would it not be better to just make it so after the music starts its GG?

zzzmike:
> Would it not be better to just make it so after the music starts its GG?

If you're talking about preventing any disarm after the music starts, I already addressed this option in the "Explain why it's good for the game" section.

Edit: Basically I think it's more intuitive to do it this way.

ihatethisengine:
Honestly I think it would make a lot more sense to make it undisarmable after music starts. Imagine playing marine, hearing the nuke music and then nothing happens. That would be a lot more confusing and feel more "buggy" than not being able to disable the nuke the very last second. Like you can imagine some irreversible chemical reaction already started that would lead to the explosion, so it's too late to disarm it.

zzzmike:
> Imagine playing marine, hearing the nuke music and then nothing happens. That would be a lot more confusing and feel more "buggy" than not being able to disable the nuke the very last second.

I don't see it as confusing, because the "NUCLEAR EXPLOSIVE DECRYPTION HALTED." message exists. And the music and screenshake is getting halted midway if a disarm happens now anyways.

ihatethisengine:
> I don't see it as confusing, because the "NUCLEAR EXPLOSIVE DECRYPTION HALTED." message exists. And the music and screenshake is getting halted midway if a disarm happens now anyways.

Ok, look on it from another perspective. Player is told there is a minute until nuke goes off. But with your changes it is a minute + like 5 seconds (or how long is the sequence) when the music and screenshake goes on, because your change allows you to stop the explosion AFTER the time is up. It is in fact confusing and misleading. 

I dunno why are you so set on choosing "allow player to disarm the bomb AFTER the timer is up" over "forbid player from disarming the bomb after the timer is up". Just logically speaking the second option is more sound.

zzzmike:
> It is in fact confusing and misleading.

I feel like it's more confusing and misleading to throw some lame error message at the queen during the extremely rare situation where she manages to make it to the bomb and disarm it during the sleep(10) process (which seems to only exist so that the explosion is timed near the end of the sound clip)

ihatethisengine:
> the extremely rare situation where she manages to make it to the bomb and disarm it during the sleep(10) process (which seems to only exist so that the explosion is timed near the end of the sound clip)

* It's not extremely rare. Because mind you, marines have to evac to not get caught by the blast. Usually they evac when the timer is like less than 30 seconds. They have to abandon cades even sooner. So Queen coming to the bomb right after the dropship has left is actually VERY common. If Queen gets additional free 10 seconds, it becomes a lot more easy to approach and disarm the bomb. 

* So unless you have 10 layers of cades set, nuke gets pretty easy to reach. Especially since most LZs don't have much space you also kinda have to set up nuke close to the dropship, and if dropship is gone, that side is left open. (Moot point I guess but still worth mentioning)

* What this change promotes is the risky very last second evacs and people staying behind to protect the nuke from the last second disarms (_actually, it's not the last second, it's AFTER the last second disarm, but i digress_). While heroic and somewhat fitting the game, this **should not** be the base requirement. Marines should only have to stay behind when the defenses are basically overwhelmed, your change basically forces marines to do this every time because you don't wanna give the Queen a free disarm during the music sequence.

* In the end, what do we gain? I get it somewhat SUCKS to not being able to disable the bomb after the timer is already done. But wouldn't it suck to lose the nuke just because now xenos can disarm it after the timer is done? Wouldn't it suck that you have to re-deploy or suffer through hijack, when the round could have been over already? I feel like there is a very little upsides and many downsides in the approach you decided to take.

zzzmike:
> It's not extremely rare

I said that because I've played regularly since 2023 and I've seen the queen sucessfully disarm during the sound clip only once. I get that it might seem common in theory, but I think it's quite hard to actually pull it off - you have to go through potentially multiple cade layers, while potentially fighting marines at the same time (especially if they realize that the alamo only needs to have finished taking off before the 10 second sound clip finishes). Add in more stationary defenses like tesla and turrets if the marines are competent. Then go through and complete the disarm process... The one time I see it in 2.5 years of play, the bomb explodes anyways. It seemed very lame for the xeno side. By letting them do it, I see it being a fun thing that ends up happening in maybe one round every year or something.

PR #10577, opened 2025-09-06
colony nailgun pinning

byondsucks:
# About the pull request

readds nailgun pinning (when a xeno is pb'd and a wall is behind it, it gets immobilized for 0.5 seconds)
when a xeno is hit by a nailgun and there is a wall behind it, it does slightly more damage and applies a tiny bit of superslow
normal nailgun bullets also apply a bit of slow now
the changes above only apply to engineers using the nailgun (engi skill 2 and above).

it was a cool feature which got removed due to being blatantly too op, I believe it'd be nice to add some uniqueness to a rarely used "gun". This also shouldn't affect balance much due to only engineers having these buffs, and due to the nailgun being uncommon on colonies.

# Changelog

:cl:
balance: Nailgun pinning along with the nailgun slowing on hit have been added back.
/:cl:


code/datums/ammo/bullet/smg.dm
```diff
@@ -69,6 +69,65 @@
...
 
+/datum/ammo/bullet/smg/nail/on_pointblank(mob/living/target, obj/projectile/bullet, mob/living/user, obj/item/weapon/gun/fired_from)
+	if(!target || target == bullet.firer || target.body_position == LYING_DOWN)
+		return
+
+	if(!skillcheck(user, SKILL_ENGINEER, SKILL_ENGINEER_TRAINED))
+		return
+
+	if(iscarbonsizexeno(target))
+		var/mob/living/carbon/xenomorph/X = target
```
Red-byte3D: single letter var

Git-Nivrak:
I don't see the benefit of this, feel free to reopen if you get balance approval.

PR #10592, opened 2025-09-08
Removes dead larva core refunding

BOBAMAx:
# About the pull request

Removes the ability to drag dead larva and revive them at the hive core

A couple of things to note: this was never an intended feature and was only borne out of an oversight. Originally, larva without clients were automatically burrowed and added to the reserve pool to remove AFK larva from the round, but this had the added consequence of allowing DEAD larva (who also usually dont have a client) to be refunded as well. It was semi-formalized as a feature some time later for some god-forsaken reason instead of being actually fixed.

Things larva refunding does not do:
It does not prevent larva suicide, as larva have a base chance of 25 percent to gib and will often be unrecoverable anyway.
It does not prevent survs killing AFK larva roundstart, for the same reason as above.
both of these are ADMIN issues that are NOT solved by larva refunding, and need to be ahelped anyway.

Things larva refunding does do:
Provide a get out of jail free card for xenos that fail to protect the hive. The whole reason why xenos are supposed to build and protect the hive is to prevent captures and burst larva from dying. Failing to do this should have real consequences, including the loss of the young you fail to protect. It may feel bad to die as a larva, but it's not like refunding even let you return to your body anyway, it just gave it to the next guy in queue because it's not a real feature anyway.
Look incredibly stupid. The message is the same regardless of whether or not the larva is dead from being riddled with bullets because it was never an intended mechanic
`visible_message(SPAN_XENODANGER("[worm] quickly burrows into \the [src]."))`
I'd like to see how this guy burrowed into anything after being cut in half by a machine gun.
<details>
image
</details>

Also, because this is just an oversight and not a real feature, there is NO time limit on this, so if a larva dies after bursting shipside, gets sent to research, sits there for an hour, and gets dragged to the core post-hijack, it crawls into the core and gets added to the burrowed counter because ????

I fired up local, if I ghost I still burrow, if I die I don't, and I can't drag dead larva anymore but I can drag living ones.

# Changelog
:cl:
fix: Dead larva no longer get refunded at the hive core as if they are AFK/ghosted
/:cl:

### DIFF
--- a/code/modules/cm_aliens/structures/special/pylon_core.dm
+++ b/code/modules/cm_aliens/structures/special/pylon_core.dm
@@ -263,10 +263,9 @@
 	// Handle spawning larva if core is connected to a hive
 	if(linked_hive)
 		for(var/mob/living/carbon/xenomorph/larva/worm in range(2, src))
-			if((!worm.ckey || worm.stat == DEAD) && worm.burrowable && (worm.hivenumber == linked_hive.hivenumber) && !QDELETED(worm))
+			if((!worm.ckey && worm.stat != DEAD) && worm.burrowable && (worm.hivenumber == linked_hive.hivenumber) && !QDELETED(worm))
 				visible_message(SPAN_XENODANGER("[worm] quickly burrows into \the [src]."))
-				if(!worm.banished)
-					// Goob job bringing her back home, but no doubling please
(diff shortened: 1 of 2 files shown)


Red-byte3D:
Intended or not, this has been a part of the gameplay for a while and it counts as a balance change, please tag the pr correctly

Drulikar:
Not a bug.

PR #10633, opened 2025-09-14
fixes detonating on open_space impact

### LINKED ISSUE #10001: Boiler Glob doesn't travel Z levels
### Testmerges

#9427, #9590, #9872. #8878, #9894, #9175, #9405, #9978

### Round ID

_No response_

### Description of the bug

When I shoot a boiler glob from a z-level above, the glob explodes mid-air and doesn't affect the z-level below it. I can only assume globs shot on ground level don't gas open air above it either.

I'd think a glob would fall down to the nearest solid ground and then burst into gas, somewhat similar to how a spitter's spit can hit a marine on a lower z level.

1. Get on a z-level with a visible z-level below it.
2. Fire a glob into the empty space above the ground.
3. Weep.

### Issue Bingo

cuberound:
# About the pull request

bunch of projectiles were detonating on hitting open turf and not going down fixes #10001 

detonation should not happen on hitting air

<details>

</details>

# Changelog

:cl:
fix: fixes projectile detonations on hitting open space, for stuff like boiler gas, plasma caster, SHARP, SADAR and many more
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -440,7 +440,7 @@
 	var/hit_turf = FALSE
 	// Explosive ammo always explodes on the turf of the clicked target
 	// So does ammo that's flagged to always hit the target
-	if(((ammo_flags & AMMO_EXPLOSIVE) || (ammo_flags & AMMO_HITS_TARGET_TURF)) && T == target_turf)
+	if(!istype(T, /turf/open_space) && (((ammo_flags & AMMO_EXPLOSIVE) || (ammo_flags & AMMO_HITS_TARGET_TURF)) && T == target_turf))
 		hit_turf = TRUE
 
 	for(var/atom/movable/clone/C in T) //Handle clones if there are any


code/modules/projectiles/projectile.dm
```diff
@@ -440,7 +440,7 @@
 	var/hit_turf = FALSE
 	// Explosive ammo always explodes on the turf of the clicked target
 	// So does ammo that's flagged to always hit the target
-	if(((ammo_flags & AMMO_EXPLOSIVE) || (ammo_flags & AMMO_HITS_TARGET_TURF)) && T == target_turf)
+	if(!istype(T, /turf/open_space) && (((ammo_flags & AMMO_EXPLOSIVE) || (ammo_flags & AMMO_HITS_TARGET_TURF)) && T == target_turf))
```
Drulikar, line 443: Why aren't `/turf/open_space` a subtype of `turf/open` anyways ? Wouldn't that solve this?

code/datums/ammo/energy.dm
```diff
@@ -144,21 +144,21 @@
...
 /datum/ammo/energy/yautja/caster/sphere/aoe_stun/do_at_max_range(obj/projectile/stun_projectile)
-	do_area_stun(stun_projectile)
+	do_area_stun(stun_projectile, stun_projectile)
 
-/datum/ammo/energy/yautja/caster/sphere/aoe_stun/proc/do_area_stun(obj/projectile/stun_projectile)
+/datum/ammo/energy/yautja/caster/sphere/aoe_stun/proc/do_area_stun(atom/impact, obj/projectile/stun_projectile)
 	playsound(stun_projectile, 'sound/weapons/wave.ogg', 75, 1, 25)
 
-	for(var/mob/living/carbon/any_target in orange(stun_range, stun_projectile))
+	for(var/mob/living/carbon/any_target in range(stun_range, impact))
```
Drulikar, line 161: This should be mentioned in CL

cuberound: ok

Drulikar (changes requested):
As far as I can tell this doesn't fix the linked issue nor am I seeing a behavior change.

From what I understand of that PR they want it where if you're looking up that you *can* explode it on open space to affect that upper z. But doing so would prevent shooting down a Z - nor could you really understand player intent of what z they want it to go to because we don't have a look down (you're always looking down).

Only solution to their issue I think would be for gas to always spread multiz regardless of where its fired at. E.g. cloud steps also attempt to spread +/- z at cost of 1 distance if it does.

PR #10666, opened 2025-09-19
Fix bun.exe not being found when repo path contains spaces

rattybag:
# About the pull request

This PR fixes an issue where `javascript.bat` was failing to find `bun.exe` when the repository path contained spaces. The problem occurred because the batch file's method of capturing the bun path from PowerShell was being corrupted by console output wrapping.

**Root Cause:** The original code used a `for /f` loop to capture PowerShell output, but when paths contained spaces and were long enough to wrap in the console, the path variable would be truncated at the line break, resulting in an invalid path.

**Solution:** Replaced the problematic approach with a more robust method that:
- Uses a temporary file to avoid console wrapping issues
- Employs PowerShell's `-NoNewline` parameter to prevent trailing newlines
- Uses the full path to `bun.exe` directly instead of relying on PATH environment variable lookup

**Testing:** Verified that `javascript.bat --version` and `javascript.bat --help` work correctly in both PowerShell and cmd environments, even with repository paths containing spaces.

This is a development environment fix that ensures the build tools work correctly for all contributors, regardless of where they place their repository folder.

**Before Fix (Failure):**

```> & cmd /c "`"D:\cmss13 repos\cmss13\tools\bootstrap\javascript.bat`" --version"```
`"javascript.bat: Failed to bootstrap Bun!"`

**After fix:**

```> & cmd /c "`"D:\cmss13 repos\cmss13\tools\bootstrap\javascript.bat`" --version"```
`Using vendored Bun 1.2.18` 
`1.2.18`

:cl: rattybag
fix: Fixed javascript.bat failing to find bun.exe when repository path contains spaces
/:cl:

### DIFF
--- a/code/datums/ammo/ammo.dm
+++ b/code/datums/ammo/ammo.dm
@@ -237,7 +237,8 @@
 	set waitfor = 0
 
 	var/turf/curloc = get_turf(original_P.shot_from)
-	var/initial_angle = Get_Angle(curloc, original_P.target_turf)
+	var/turf/cur_target = get_turf(original_P.target_turf)
+	var/initial_angle = Get_Angle(curloc, cur_target)
 
 	for(var/i in 1 to bonus_projectiles_amount) //Want to run this for the number of bonus projectiles.
 		var/final_angle = initial_angle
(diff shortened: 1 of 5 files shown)


Drulikar:
I currently cannot reproduce this issue as mentioned on the discord. But I doubt using a temp file is necessary regardless.

rattybag:
So this issue was actually due to me following the manual installation guide for vscode shell integration, where I added `if ($env:TERM_PROGRAM -eq "vscode") { . "$(code --locate-shell-integration-path pwsh)" }` to my powershell profile, without actually having `pwsh` available in my system.

PR #179, opened 2022-04-30
dir locking now slows you down

stanalbatross:
Changing directional lock

## About The Pull Request
Changes how directional locking works so it adds a slowdown when you are directional locked.

## Why It's Good For The Game

Direction lock needs a drawback as plenty of people just do it for no apparent reason except to make their sprite smaller and so forth. It also makes little sense you can sprint backwards at full speed without issue and shoot etc. which leads to gameplay immersion being broken.

This PR intends to change it so direction locking no longer is only, or rather, more used to permaface a certain direction by adding a slowdown when you are directional locked.

## Changelog

:cl: nanu308, stan_albatross
balance: being direction locked and not moving in the direction your character is facing will now slow you down a bit.
/:cl:

PR #10679, opened 2025-09-21
removes dirlock slowdown from humans

stanalbatross:
# About the pull request

As directional lighting was added back to humans, the reason for humans being able to dirlock is back, and thus the slowdown should be removed. Let me explain:

2019-06-25
  RavingManiac:
  - rscadd: Made light from humans directional. In front of you, light. Behind you,
      darkness.
  - rscadd: You can now lock your direction with alt-move. This is unset by pressing
      alt-move in the same direction again, or facing another direction via clicking
      or ctrl-move.

directional lighting is added, and direction locking is added in turn to let marines keep their shoulder lamps locked in one direction.

2019-09-29
  Vampmare:
  - rscdel: Disabled directional lightning temporarily.

dir lighting is removed, but dir locking remains

2022
nanu tells me to remove dir lock because dir lighting is gone and its goofy (also bad for powergaming xenos who can lock their sprite to minimise target area), I come to a compromise of adding a slowdown (#179)

2023
harry adds back dir lighting #4229 but dirlock is untouched

THEREFORE dirlock slowdown should be removed from humans as they now have a need to dirlock again. Xenos and everyone else have no need to dirlock so it will still slow them down

adds back complexity to directional lighting and encourages thinking about your direction when patrolling to marine gameplay instead of dirlock being a giga slow and never used

<details>

</details>

# Changelog
:cl:
balance: direction locking no longer slows down humans. Xenos etc unaffected.
/:cl:

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -185,7 +185,7 @@ CLIENT_VERB(drop_item)
 		if(mob.next_move_slowdown)
 			move_delay += mob.next_move_slowdown
 			mob.next_move_slowdown = 0
-		if(!ishuman(mob)) //humans can dirlock with no slowdown
+		if(mob.dirlock_slowdown) //humans can dirlock with no slowdown
 			if((mob.flags_atom & DIRLOCK) && mob.dir != direct)
 				move_delay += MOVE_REDUCTION_DIRECTION_LOCKED // by Geeves
 
--- a/code/modules/mob/living/carbon/human/human_defines.dm
+++ b/code/modules/mob/living/carbon/human/human_defines.dm
@@ -2,6 +2,7 @@
 	light_system = MOVABLE_LIGHT
 	rotate_on_lying = TRUE
 	blocks_emissive = EMISSIVE_BLOCK_UNIQUE
+	dirlock_slowdown = FALSE
 	//Hair color and style
 	var/r_hair = 0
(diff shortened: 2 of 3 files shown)


code/modules/mob/mob_movement.dm
```diff
@@ -185,8 +185,9 @@ CLIENT_VERB(drop_item)
 		if(mob.next_move_slowdown)
 			move_delay += mob.next_move_slowdown
 			mob.next_move_slowdown = 0
-		if((mob.flags_atom & DIRLOCK) && mob.dir != direct)
-			move_delay += MOVE_REDUCTION_DIRECTION_LOCKED // by Geeves
+		if(!ishuman(mob)) //humans can dirlock with no slowdown
+			if((mob.flags_atom & DIRLOCK) && mob.dir != direct)
+				move_delay += MOVE_REDUCTION_DIRECTION_LOCKED // by Geeves
```
realforest2001: Rather than typechecking can you make a var or mob flag for no slowdown

realforest2001 (changes requested):

code/modules/mob/mob_movement.dm
stanalbatross: done as a var

Wintermote:
Really great improvement. Can finally watch my squad's back and not get left behind.

PR #3636, opened 2023-06-15
Xeno Candidate Queue

Drulikar:
# About the pull request

This PR changes it so get_alien_candidates is normally a sorted list based generally on timeofdeath in ascending order and larva spawns now pull candidates out of that list in order. That means that an observer that has yet to play, or was the first to die, and still meets all other criteria to become a larva (not AFK, has preferences set to become larva, has been dead long enough, etc.) will be chosen before others.

Playing as a facehugger (dying or hugging successfully) or dying in an admin z-levels (such as thunderdome) also do not affect the value used for how you are sorted in the queue. So you should be able to freely play in either of those situations without losing your relative spot in the queue. Of course its not going to nab you out of whatever mob you are playing, but when you are a ghost again the same values will be used to sort you in the queue.

Since people may enter and leave the queue each time the queue is checked, your place in the queue may go up or down. It is just a snapshot of that moment where you are. If you missed your queue message, or haven't gotten one yet, the join xeno action will now display the last message for you.

# Explain why it's good for the game

Picking candidates randomly is okay, but it would be more fair to give privilege to those that have been waiting longer - especially those who have yet to play.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Example of queen death code (since that was the most altered):
larva

</details>

# Changelog
:cl: Drathek
add: The selection to become a xeno larva is now based on timeofdeath rather than random and also sends a message to all candidates when the queue moves. Playing as a facehugger or on admin z-levels (thunderdome) will not affect your relative place in queue. The join xeno action will also display the last queue message for you when you allow xeno candidacy.
fix: Fixed gibbed humans not properly setting their timeofdeath when on an admin z-level.
fix: Fixed a bad del on shuttle doors when a shuttle deletes (such as intoTheSunset).
/:cl:

PR #10688, opened 2025-09-22
Rename Larva Queue as Larva Pool + Tweaks

Drulikar:
# About the pull request

This PR is a follow up to #3636 and the many of the subsequent PRs tweaking it:
- Larva queue is now known as the larva pool. This was done because there's way too many preconceptions of entering/leaving a queue when the larva pool here only cares about A) How long you've been dead and B) Whether you are currently eligible (e.g. not afk, special role set for xenomorph, not xeno job banned, not banished, been dead a minimum amount of time, etc).
- Observer status panel now attempts a refresh immediately if a xeno candidate but no cached message is set yet. This was done because players were thinking the join as the xeno button would enter them into queue/pool and incorrectly thinking afk xenos being offered would prevent them from being entered into the queue/pool.
- Staff joining hive via lobby now sets mod admin larva protection off automatically. I don't know if this confused anyone, but we're already doing it for the in game button so may as well do it for the lobby button too.
- Tweaked special role wordings. Namely I removed the "after unrevivable death" part because people might think observing isn't death so think it doesn't apply to that situation.
- Removed agent as a special role. I have no idea what it was for or why it was left around.

Yet another attempt to nip at player confusion and misinterpretations of the larva queue/pool.

<details>

</details>

# Changelog
:cl: Drathek
qol: Larva queue is now known as larva pool and its position message is cached immediately for status panel if xeno candidacy is set
ui: Tweaked some messaging/labels for xeno special role candidacy
code: Renamed larva queue variables and comments to pool and rearranged some larva pool functionality
del: Removed the special role known as agent
admin: Late joining as xeno from menu as a mod now immediately sets larva protection false
/:cl:

### DIFF
--- a/code/__DEFINES/client_prefs.dm
+++ b/code/__DEFINES/client_prefs.dm
@@ -1,6 +1,6 @@
 //=================================================
-#define BE_ALIEN_AFTER_DEATH (1<<0)
-#define BE_AGENT (1<<1)
+#define BE_ALIEN (1<<0)
+// 1<<1 was BE_AGENT
 #define BE_KING (1<<2)
 //=================================================
 
--- a/code/__DEFINES/job.dm
+++ b/code/__DEFINES/job.dm
@@ -511,8 +511,8 @@ GLOBAL_LIST_INIT(job_command_roles, JOB_COMMAND_ROLES_LIST)
 
 //---------- ANTAG ----------//
 #define JOB_PREDATOR "Predator"
-#define JOB_XENOMORPH    "Xenomorph"
-#define JOB_XENOMORPH_QUEEN  "Queen"
+#define JOB_XENOMORPH "Xenomorph"
+#define JOB_XENOMORPH_QUEEN "Queen"
 
 // For coloring the ranks in the statistics menu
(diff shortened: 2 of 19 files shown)


code/__DEFINES/client_prefs.dm
```diff
@@ -1,6 +1,6 @@
 //=================================================
-#define BE_ALIEN_AFTER_DEATH (1<<0)
-#define BE_AGENT (1<<1)
```
Git-Nivrak, line 3: Preferably make a savefile update instead of leaving dead code

Git-Nivrak (changes requested):

code/__DEFINES/client_prefs.dm
Drulikar: No dead code left here. The deadcode was removed. Its just a comment to note that should someone want to use that number, to keep in mind that people have already been able to set it before. I intentionally avoided making any save file change because there's no benefit. (Don't know if future thing wants to retain the existing value, or force false, or force true; or if ever anything would be implemented)

hry-gh (approved):

PR #10690, opened 2025-09-22
Fix blocky lighting

Git-Nivrak:
# About the pull request
Bug from #10427 as we handle multiz slightly differently

Before (with #10427):
image

After (this pr):
image

Makes the game look better

<details>
See above screenshots

</details>

# Changelog

:cl:
fix: Fixes lighting sometimes appearing as blocky instead of fancy lighting
/:cl:

### DIFF
--- a/code/modules/lighting/lighting_static/static_lighting_object.dm
+++ b/code/modules/lighting/lighting_static/static_lighting_object.dm
@@ -8,9 +8,6 @@
 	///the turf that our light is applied to
 	var/turf/affected_turf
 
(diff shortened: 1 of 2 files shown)


Drulikar:
I think there's still other issues:

On your pr: 
image

But not necessarily because of their PR because this is prior to it...
image

But I guess I would need to go a lot further back in time to see if this wasn't just always a thing.

Drulikar (approved):
Code currently seems to be working though as far as I can tell; but feel free to iterate if you need to.

PR #10757, opened 2025-09-29
buildmode basic mode more friendly

kiVts:
# About the pull request

Makes basic buildmode more friendlier and easier to use
Less potential for missclicks, more friendly and easy to learn controls. 
I *probably* going to see about other modes , namely the advanced one, but I dunno what could be exactly changed to make it better. 

Also could add way more things to it without too many buttons on the top right.

~~kinda rewrote the entirety of basic mode~~
https://streamable.com/r7v3v5

# Changelog
:cl:
admin: basic buildmode should be more friendlier now.
/:cl:

### DIFF
--- a/code/modules/buildmode/submodes/basic.dm
+++ b/code/modules/buildmode/submodes/basic.dm
@@ -87,7 +87,6 @@
 						"West" = image('icons/misc/buildmode.dmi', icon_state = "build", dir = WEST),
 					)
 					if(GLOB.radial_menus["buildmode_basic_turf_[key_name(admin_building)]"])
-						to_world("duplicate, cancel")
 						return
 					var/picked_turn_option = show_radial_menu(admin_building.mob, clicked_atom, radial_menu_options, tooltips = TRUE, uniqueid = "buildmode_basic_turf_[key_name(admin_building)]")
 					clicked_atom.setDir(text2dir(capitalize(picked_turn_option)))


code/modules/buildmode/submodes/basic.dm
```diff
@@ -1,60 +1,138 @@
...
+						if(istype(clicked_atom, i))
+							downgrade_version = downgrade_reference[i]
+					if(downgrade_version)
+						clicked_turf.PlaceOnTop(downgrade_version)
+					else
+						clicked_turf.ScrapeAway()
+					log_admin("Build Mode: [key_name(admin_building)] downgraded [clicked_turf] at [AREACOORD(clicked_turf)]")
+				if("Make Indestructible")
+					clicked_atom.unacidable = TRUE
+					clicked_atom.explo_proof = TRUE
```
Drulikar: You might also want health `-1` - but its pretty difficult that there isn't a standardized upon way for something to be deemed indestructible. Always depends on the implementation what they check.

code/modules/buildmode/submodes/basic.dm
```diff
@@ -1,60 +1,138 @@
...
+			log_admin("Build Mode: [key_name(admin_building)] built [upgrade_version] at [AREACOORD(clicked_turf)]")
+			return
+		else if(right_click && !alt_click && !ctrl_click)
+			radial_menu_options = list(
+					"Destroy" = image('icons/misc/buildmode.dmi', icon_state = "buildquit"),
+					"Make Indestructible" = image('icons/misc/buildmode.dmi', icon_state = "build", dir = NORTH),
+					"Downgrade" = image('icons/misc/buildmode.dmi', icon_state = "build"),
+					"Turn" = image('icons/misc/buildmode.dmi', icon_state = "buildmode_edit"),
+				)
+			var/picked_option = show_radial_menu(admin_building.mob, clicked_atom, radial_menu_options, tooltips = TRUE)
```
Drulikar: Are you okay with multiple radial menus being allowed to open simultaneously? (Just keep right clicking things) Probably would make more sense to dismiss any already open option when opening a new one (just use the `uniqueid` option and look up any open menu with `GLOB.radial_menus[uniqueid]` but keep in mind it needs to be unique to the user not a server wide shared radial menu.

code/modules/buildmode/submodes/basic.dm
```diff
@@ -1,60 +1,138 @@
...
+					var/turf/downgrade_version
+					for(var/i in downgrade_reference)
+						if(istype(clicked_atom, i))
+							downgrade_version = downgrade_reference[i]
+					if(downgrade_version)
+						clicked_turf.PlaceOnTop(downgrade_version)
+					else
+						clicked_turf.ScrapeAway()
+					log_admin("Build Mode: [key_name(admin_building)] downgraded [clicked_turf] at [AREACOORD(clicked_turf)]")
+				if("Make Indestructible")
```
Drulikar: Would be nice if we're offering this that its a toggle between indestructible vs not

Drulikar (changes requested):
Some other minor things:
- [ ] Merge with master (so you can see the next problem)
- [ ] Tweak the plane order for build mode (should be over action buttons, but probably under ESC menu - but see below)
- [ ] Currently ESC menu is nonfunctional because of the click intercepts - you maybe should just exit buildmode if you open esc menu?
image

- [ ] I don't know if this is a new radial bug or if its just your click interceptor being weird, but some reason they're opening twice - doesn't seem to matter if I am clicking an option or dismissing it, it'll open again afterwards.

https://github.com/user-attachments/assets/3bc1d54c-3a0e-4294-8783-3464f422ad05

- [ ] It would be nice if copy mode by default used the dir option in the top left of build mode.

code/modules/buildmode/submodes/basic.dm
kiVts: quite sure buildmode is instanced each time buildmode is clicked, so this should be good

Drulikar: Well yes it *currently* is fine (other than the obvious of allowing you to open infinite radials), but if you were to implement it where one user can only open one radial at a time, you'd need to be sure the unique ID is unique to that user.

Drulikar: If you needed a visual of what I'm talking about: 
image

kiVts: alright Im convinced, though I meant it would be easy to fix, done.

code/modules/buildmode/submodes/basic.dm
kiVts: turfs dont have health, going to give it turf hull flag instead. Obj do have health though.

Drulikar (approved):

PR #10800, opened 2025-10-04
Airburst Grenade point cost consistency in Vendors

Mr-Carp:
# About the pull request

All roles barring Specialist and SG's get reduced grenade prices compared to riflemen except airburst which are more expensive
This just lowers them down to be in line with how other grenades are priced.

Airburst grenades which are worse then regular HEDP's and HIDP's shouldn't cost more then them and they shouldn't be more expensive then what a riflemen can buy.

# Changelog
:cl: armsdealer9876
balance: Reduce airburst grenade point cost for consistency amongst roles
/:cl:


HaultyAnonie:
Why the number 13 specifically.

Mr-Carp:
I set it to 13 because it cost 15 for a riflemen and the discount given to other grenades was 2 points so I just followed suit

Wintermote:
> I set it to 13 because it cost 15 for a riflemen and the discount given to other grenades was 2 points so I just followed suit

Might have been better to do as a percentage rather than flat, but I don't know the price of grenades since I never buy them.

Mr-Carp:
If we did percentages then it'd be 14 which I could see, this isn't to encourage buying them its just to make it in line with the state airburst are in. They are not better then regular grenades and therefor shouldn't be more expensive. (Especially when riflemen is able to purchase it for cheaper)

kiVts (approved):

PR #8345, opened 2025-02-03
FOB Defense Overhaul: Power System Framework

Nomoresolvalou:
# About the pull request

This is the first PR for what will eventually be a full rework on how marines construct and defend their FOB. This is only a framework, meaning nothing will be added into the game quite yet, but the code and sprites are there. Once all the other structures are completed, then they will all be added in together.

that being said, in this PR the additions are as follows:

A colony transformer, which when powered on will power the rest of the defensive structures that marines will be able to construct in the FOB.

A service terminal, that will allow marines to link up other structures to the power grid using a multitool. It also displays the status of the FOB's power.

Finally, a backup generator. When the transformer is disabled marines will be able to activate the generator. Once turned on it cannot be turned back off. It will supply 5 minutes worth of power to the FOB power network.

All the sprites are done by Thwomper

# Explain why it's good for the game

This is the first step into completely changing the gameplay around setting up and defending a FOB. This framework should make it easier for others to create the rest of the structures that marines will get to use for their FOB. The entire FOB rework plan can be found here: [https://hackmd.io/@XXHgmxPJTIWgk8o6fbWLEw/By9Nfbwdyg](url)

<details>
<summary>Screenshots & Videos</summary>

demonstration of how the three structures will function once added

https://github.com/user-attachments/assets/f40de23a-6387-49b6-8d54-d4e689666d5e

</details>

# Changelog

:cl: Nomoresolvalou, Thwomper

code: added in the back end code for the HVT-7T high voltage transformer, the UE-09 Service Terminal, and the UE-11 Generator Unit
imageadd: added sprites for the HVT-7T high voltage transformer, the UE-09 Service Terminal, and the UE-11 Generator Unit
/:cl:

PR #10829, opened 2025-10-09
Stealth hijack begone

cuberound:
# About the pull request

Adds the same danger check to allowing hijack as there is to preventing boon purchase, aka 12 + humans on the planet.

Stealth hijack is bad for gameplay and anticlimactic, accept your defeat or die fighting.

<details>

</details>

# Changelog
:cl:
balance: Disables hijack option if there are more then 12 humans still on the planet
/:cl:

### DIFF
--- a/code/modules/shuttle/computers/dropship_computer.dm
+++ b/code/modules/shuttle/computers/dropship_computer.dm
@@ -304,10 +304,6 @@
 		playsound(loc, 'sound/machines/terminal_error.ogg', KEYBOARD_SOUND_VOLUME, TRUE)
 		return XENO_NONCOMBAT_ACTION
 
-	if(check_danger())
-		to_chat(xeno, SPAN_XENONOTICE("Lights flash with warnings. There is still too much danger on this planet for us to leave!"))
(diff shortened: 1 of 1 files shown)


RainbowStalin:
Wouldn't it be better to make it impossible through https://github.com/cmss13-devs/cmss13/pull/8345? Less behind the scenes number fenangling, and more visual and physical reasons this does not happen.

cuberound:
> Wouldn't it be better to make it impossible through #8345? Less behind the scenes number fenangling, and more visual and physical reasons this does not happen.

that will come at some point too perhaps, but even with it this should be a thing, perhaps even more so. hijack is to happen when xenos have won the fight on planet, and you can not easily hide 12 humans on planet from xenos

ihatethisengine:
If this gets merged, what would be the gameplay reason for marines to even defend FOB? If Queen cannot hijack, you can literally give up LZ without losing anything. Just keep pushing hive, or anything else.

Also this implementation wont allow you to launch if you have 12 caps which is really funny.

Git-Nivrak:
I don't think this makes a lot of sense as the dropship is the thing you are defending in the FOB, with #9578 as well this doesn't seem like an issue

PR #10832, opened 2025-10-09
Changes the Commander headset to a senior command one

Googles-Hands:
# About the pull request
Changes the headset in the locker for the Commander of the operation to an actual senior command headset.
The CO, XO and ASO already spawn with this headset. Anyone else that would ever open the locker would be a game assigned Commander on lowpop like a SO, PO or other lower-ranking marine. Considering the Commander assigned by the game is responsible for the entire ship, including Marine Law/MP related matters, it makes sense to ensure they have access to all communications channels. When someone higher ranked than them wakes up they will then be expected IC'ly to pass the senior command headset along to them and go back to their original one.
# Changelog
:cl:
balance: Exchanged the CIC headset for the operations Commander for a senior command headset
/:cl:


realforest2001:
I don't think this is needed. Whilst the aCO has relevance in Marine Law, they have less authority than the CMP, with the exception of the Commanding Officer. There's no need for them to be automatically given MP comms.

Googles-Hands:
> I don't think this is needed. Whilst the aCO has relevance in Marine Law, they have less authority than the CMP, with the exception of the Commanding Officer. There's no need for them to be automatically given MP comms.

The Commander is ultimately responsible for all marines, considering that, it would not make sense to exclude them from any part of the force, in this case the matters of the Military Police. He should be able to aid in coordinating them (especially on lowpop) and give his opinion on matters that are discussed on this rather exclusive frequency.
Further, the Commander is also inevitably going to be the target of any mutiny that occurs.

This honestly does not seem a very radical change, you eventuell agreed the ASO could have this headset and decided your concerns of them not being held to a higher standard like XO's by the rules wasn't enough to object to the change. Now if there is no ASO, this thought simply applies to the next in line. Just because the CMO/CE/SO/and so on aren't held to a higher standard by the OOC rules doesn't mean we shouldn't entrust these frequencies to them on the rare occasion they get promoted.

PR #10834, opened 2025-10-09
Add TGUI dev and build launch targets

rattybag:
# About the pull request

Added new launch targets for running TGUI build / dev (hot reload)

- **Improved Developer Experience**: TGUI tasks are now directly accessible from VS Code's Run and Debug interface instead of requiring manual terminal commands
- **Consistency**: Aligns with existing BYOND launch configurations, providing a unified workflow
- **Convenience**: One-click access to common TGUI development tasks without memorizing command paths
- **Better Integration**: Leverages VS Code's built-in task management system for better IDE integration
- **Reduced Context Switching**: Developers can stay within the VS Code interface instead of switching to external terminals

This streamlines the TGUI development workflow and makes the codebase more accessible to new contributors who may not be familiar with the command-line build tools.

image

:cl: rattybag
code: adding TGUI build and dev launch targets
/:cl:

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -121,6 +121,20 @@
             "preLaunchTask": "Build All (runtime map + quickstart + testing)",
             "dmb": "${workspaceFolder}/${command:CurrentDMB}",
             "dreamDaemon": true
+        },
+        {
+            "name": "TGUI Dev Server",
+            "type": "node-terminal",
+            "request": "launch",
+            "command": ".\\bin\\tgui-dev.cmd",
+            "cwd": "${workspaceFolder}"
+        },
+        {
+            "name": "TGUI Build",
+            "type": "node-terminal",
+            "request": "launch",
+            "command": ".\\bin\\tgui-build.cmd",
+            "cwd": "${workspaceFolder}"
         }
     ]
 }


.vscode/launch.json
```diff
@@ -121,6 +121,20 @@
...
+            "request": "launch",
+            "command": ".\\bin\\tgui-dev.cmd",
+            "cwd": "${workspaceFolder}"
+        },
+        {
+            "name": "TGUI Build",
+            "type": "node-terminal",
+            "request": "launch",
+            "command": ".\\bin\\tgui-build.cmd",
+            "cwd": "${workspaceFolder}"
```
Drulikar, line 137: TBH I don't really think tgui build is useful. Yes, it will tell you about errors that linters would tell you about, but as far as fixing them you'd want tgui-fix.cmd. But also weird to make this be a debugging launch target when you actually want to make a command (so something in the [command palette](https://code.visualstudio.com/api/references/contribution-points#contributes.commands) (ctrl+shift+p)

.vscode/launch.json
```diff
@@ -121,6 +121,20 @@
             "preLaunchTask": "Build All (runtime map + quickstart + testing)",
             "dmb": "${workspaceFolder}/${command:CurrentDMB}",
             "dreamDaemon": true
+        },
+        {
+            "name": "TGUI Dev Server",
+            "type": "node-terminal",
+            "request": "launch",
+            "command": ".\\bin\\tgui-dev.cmd",
+            "cwd": "${workspaceFolder}"
```
Drulikar, line 130: This works, but its kinda odd to be *debugging* the dev server. That sure could have uses, but its not the general usecase of using the tgui dev server. Likely better as a command.

rattybag: I mean it does depend on whether we want to be able to debug the node code, which could be useful, but not something I have had to do yet in the codebase. Honestly it's a much of a muchness to me, ~~I'm happy to convert it into a command if that aligns better with the general use cases.~~ 

I don't think any change is necessary here then - I wasn't aware that tasks could be ran like this via the command palate

.vscode/launch.json
rattybag: The reason why I added this even though it's not that useful is more so for consistency, this approach does allow the debugging of the node code but like I said in the other comment not sure if that's a common or necessary use case. Happy to remove this completely.

.vscode/launch.json
rattybag: Thinking more about this, I kind of like it better as a launch target, even disregarding the aspect of debugging the tgui server, as it has better vscode integration. It's nice to be able to see the running dev server integrated in VS code's debug panel, and allowing for debugging of the dev server in this way even though it's not a common use case is a plus. It's nice to be able to have the long running processes be visible in this way.

rattybag: So I messed around with this a bit and I couldn't really get the debugging to work properly on the tgui server sadly, I think it's probably not ideal to have a debugging target that doesn't actually debug so I'll close this

rattybag:
Closing due to existing task being able to be ran through command palate and unable to get debugger to properly attach to tgui server

Issue #10858, opened 2025-10-12
Medical Items Loose Identifier


troytroy400:
### Testmerges

Multiple rounds, not TM dependent.

### Round ID

Multiple rounds.

### Description of the bug

When you take injectors from medical prep and they are automatically put into your storage they loose their identifying letters, you must then take the item out and put it back in to regain the identifying letters.

<img width="533" height="207" alt="Image" src="https://github.com/user-attachments/assets/10b8ea46-2185-4b60-8421-081a32db860c" />

### What's the difference with what should have happened?

Items should keep their identifiers.

### How do we reproduce this bug?

1. Spawns as a corpsman.
2. Go to medical prep.
3. Take an injector pouch (or other relevant storage)
4. Grab injectors from med.
5. Observe.

### Issue Bingo

- [x] Issue could be reproduced at least once
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #10880, opened 2025-10-15
fix injector labels not appearing when automatically put into a container

ori-disillusion-mirnov:
# About the pull request
the chem label on injectors (and other things that might've had a problem like this) now appear when automatically vended into a container, also fixes: #10858

taking an injector out of a container then putting it back in just for a label is annoying

<details>

</details>

# Changelog

:cl:
fix: labels appear on injectors when automatically vended into a container
/:cl:


ori-disillusion-mirnov (commented):
if you want icons to be updated whenver something is vended instead of just when its put into a container i can do that. if you want me to stop coding forever because this fix is that bad i can do that too

PR #10886, opened 2025-10-16
Yautja Cape Colors Persist When Accessory'd

matttheficus:
# About the pull request

Predator Cape colors now persist when turned into an accessory.

We have the ability to change the color - it should stay instead of getting greyscaled.

<details>

image

image

</details>

# Changelog
:cl:
fix: Pred Cloaks now retain their custom colors when turned to accessories and back.
/:cl:


code/modules/clothing/clothing.dm
```diff
@@ -37,6 +37,8 @@
 	// copies the properties of the clothing item to the accessory, in the future, take literally almost every var from ties.dm parent object and place it in clothing parent
 	var/obj/item/clothing/accessory/new_accessory = new accessory_path(loc)
 	new_accessory.name = name
+	if(istype(src, /obj/item/clothing/yautja_cape)) // checks for Pred Cape for recoloring
+		new_accessory.color = color
```
realforest2001: you may as well allow any coloured accessory to remain so, they're not typically changeable without staff intervention anyway

realforest2001 (changes requested):

code/modules/clothing/clothing.dm
matttheficus: Fixed

realforest2001 (approved):

PR #10892, opened 2025-10-17
moves the plane change up to `/obj/structure/window`

hry-gh:
didn't test on the groundmap award

:cl:
fix: fixes even more bad shadows under windows
/:cl:


Drulikar:
Shutters are a problem as Forest pointed out: 
image
image

LV Groundside may be okay, but I dunno if you're okay with this (essentially door has AO rather than flush - doesn't look bad unlike windows prior just likely unintended): 
image

image
image

Drulikar:
Dropships too I guess 
image

PR #10900, opened 2025-10-18
option to remove RMB menu in the preferences

PoltavskaPraca:
# About the pull request

+ Can now disable RMB menu in preferences, both for tile contents and obj verb menu
+ in order to acess RMB menu with it disabled, press shift + RMB
+ As a side effect, it is now possible to shoot and attack with RMB, as well as do other shit like EVERYTHING (picking up items, interacting with objects, opening your worn bag). It still applies CD and all, doesn't fuck with xeno abilities trigger bound to RMB, so whatever? 

+ old and archaic
+ im using rmb xeno abilities and if I deselect my warrior lunge on accident I dont want no fucking menu popping up as I'm getting PB buckshotted

<details>

https://github.com/user-attachments/assets/55d9f147-7a6f-448b-99e4-171df4b116dd

https://github.com/user-attachments/assets/94839388-2126-49b1-bd54-23a85b40ded0

https://github.com/user-attachments/assets/664c51f4-608a-465f-90ad-b7f6f20b101f

Screenshot 2025-10-18 072600

</details>

# Changelog

:cl:
add: Can now disable right click menu in the preferences -> settings, left upper corner: Right-click Contextual Menu. It is ON by default. Can even shoot and do a lot of stuff with RMB instead of LMB.
/:cl:


rattybag:
Instead of disabling the ability to right click completely, how about making it so you can get the right click menu, but you need to hold down a modifier key (control, alt). That way players can use both the new right click keybinds while still having access to the menu if needed.

PoltavskaPraca:
> Instead of disabling the ability to right click completely, how about making it so you can get the right click menu, but you need to hold down a modifier key (control, alt). That way players can use both the new right click keybinds while still having access to the menu if needed.

i will take a look at the code again, but it's kinda abysmal to find the proc that dictates what happens when we click a mouse button, in general

at the first glance this would be the thing to touch

also @ poltava on CM discord if u got  any more suggestions, I'll respond there quicker and it's easier to yap like dat
UPD: I will have to figure out how to make TGUI stuff...
```
/atom/proc/clicked(mob/user, list/mods)
	if (mods[SHIFT_CLICK] && !mods[MIDDLE_CLICK])
		if(can_examine(user))
			examine(user)
		return TRUE

	if (mods[ALT_CLICK])
		var/turf/T = get_turf(src)
		if(T && user.TurfAdjacent(T) && length(T.contents))
			user.set_listed_turf(T)

		return TRUE
	return FALSE

/atom/movable/clicked(mob/user, list/mods)
	if (..())
		return TRUE

	if (mods[CTRL_CLICK])
		if (Adjacent(user) && user.next_move < world.time)
			user.start_pulling(src)
		return TRUE
	return FALSE
```

PoltavskaPraca:
Upon closer inspection, it's... kind of hard. All of the verbs are stored inside of a mob. They are entirely different procs doing entirely different things. I don't think I can go:
```
/atom/proc/clicked(mob/user, list/mods)
	...
	if(!user.client.prefs.toggle_right_click_menu && mods[ALT_CLICK] && mods[RIGHT_CLICK])
		var/verbselection = input(user, "What to do?", src) as null|anything in user.verbs
		if(verbselection)
			to_chat(world, "monkey")
			do_proc_n_verb_stuff()
```
IDK how to go about the whole thing without having to specify every single 78 proc with a massive fucking switch statement
It'd be much easier just to add a keybind to toggle between the two modes, roguetown does so (keybind only usable if you are an admin, in their case) + adding verbs for snowlake items (item/storage, item/clothing, predator clan bracers, idk what else)

hry-gh:
> Instead of disabling the ability to right click completely, how about making it so you can get the right click menu, but you need to hold down a modifier key (control, alt). That way players can use both the new right click keybinds while still having access to the menu if needed.

@PoltavskaPraca you can do this using a macro in the skin, ie https://github.com/tgstation/tgstation/blob/6ce318c19a891bf05f4d2426c3954c0a0bd7d066/interface/skin.dmf#L1-L10. you can just  conditionally add this macro if they're using this pref. lmk if you get stuck

hry-gh (dismissed):
per above please use a local keybind to enable context menu, not a toggle

PoltavskaPraca:
half a year waiting, no thanks

PoltavskaPraca:
changes have been reflected in the main PR post

fira:
Sounds good, though i'm not sure what you need it for exactly. Can you elaborate on the case of shotgun PB?

PoltavskaPraca:
> Sounds good, though i'm not sure what you need it for exactly. Can you elaborate on the case of shotgun PB?

Sometimes my hand slips and I deselect a xeno ability. Xeno use combos. Instead of doing a xeno ability, I get jumpscared by RMB menu which roots me in place.

Besides that, someone could prolly add stuff down the line to RMB. Roguetown and other servers moved away from RMB menus ages ago (either modifier keys for alt actions or alt RMB to display tile contents).

Nihisohel:
exempt till reviewed

code/game/objects/items/storage/storage.dm
```diff
@@ -73,16 +73,28 @@
...
+		storage_draw_logic(src.name)
+
+	if(mods[ALT_CLICK] && mods[MIDDLE_CLICK])
+		if(istype(src, /obj/item/storage/belt/medical))
+			if(ishuman(user))
+				var/obj/item/storage/belt/medical/med_belt = src
+				med_belt.instant_pill_grabbable = !med_belt.instant_pill_grabbable
+				to_chat(user, SPAN_NOTICE("You will now [med_belt.instant_pill_grabbable ? "take pills directly from bottles": "no longer take pills directly from bottles"]."))
+		else
+			to_chat(user, SPAN_NOTICE("This item isn't a medical belt."))
```
Nihisohel: instant pill grabbable is no longer limited to medical belts

code/modules/client/preferences.dm
```diff
@@ -1198,6 +1181,16 @@ GLOBAL_LIST_INIT(be_special_flags, list(
...
+		if("toggle_right_click_menu")
+			if(user.client.prefs.toggle_right_click_menu)
+				user.client.prefs.toggle_right_click_menu = FALSE
+				user.client.set_right_click_menu_mode(shift_only = TRUE)
+				to_chat(user, "Right click no longer opens a menu, only with SHIFT + Right Click.")
+			else
+				user.client.prefs.toggle_right_click_menu = TRUE
+				user.client.set_right_click_menu_mode(shift_only = FALSE)
+				to_chat(user, "Right click opens a menu.")
+
```
Nihisohel, line 1193: rather than just saying "menu" call it a contextual menu

similarly, add a chat span to the message (use SPAN_NOTICE)

code/game/objects/items/storage/storage.dm
Nihisohel: additionally, alt middle clicking triggers turf content checking (check click.dm at line 245)

code/modules/client/preferences.dm
```diff
@@ -595,6 +593,7 @@ GLOBAL_LIST_INIT(be_special_flags, list(
 			dat += "<h2><b><u>Input Settings:</u></b></h2>"
 			dat += "<b>Mode:</b> <a href='byond://?_src_=prefs;preference=hotkeys'><b>[(hotkeys) ? "Hotkeys Mode" : "Send to Chat"]</b></a><br>"
 			dat += "<b>Keybinds:</b> <a href='byond://?_src_=prefs;preference=viewmacros'><b>View Keybinds</b></a><br>"
+			dat += "<b>Menu on Right-click:</b> <a href='byond://?_src_=prefs;preference=toggle_right_click_menu'><b>[(toggle_right_click_menu) ? "On" : "Off"]</b></a><br>"
```
Nihisohel: call this "Right-click Contextual Menu"

Nihisohel (changes requested):
code looks good for the most part

Nihisohel (approved):
bro

PR #10905, opened 2025-10-18
Fix Automatic Primary LZ Selection

Drulikar:
# About the pull request

This PR does several things:
- Changes the timer for automatic LZ selection from 25mins from world.time to 15mins from round start time
- Fixes LZ selection in the scenario if there is only 1 LZ (even if its only LZ2)
- Unifies the LZ picker to a single gamemode proc
- Forces LZ selection to the first available LZ when attempting to clear resin
- Forces LZ selection to whatever the first drop's dropship's destination is if its LZ1 or LZ2

Fixes https://github.com/cmss13-devs/cmss13/issues/10903

<details>

https://youtu.be/M8Ha_Ru2H-Q

</details>

# Changelog
:cl: Drathek
fix: Launching a dropship without a primary LZ selected will now force that dropship's destination to be the primary LZ if possible
fix: Fixes LZ selection not supporting maps that might only have 1 LZ and changes the automatic LZ selection from 25 mins world time to 15 mins round time
/:cl:

### DIFF
--- a/code/game/gamemodes/colonialmarines/colonialmarines.dm
+++ b/code/game/gamemodes/colonialmarines/colonialmarines.dm
@@ -554,6 +554,13 @@
 // Resource Towers
 
 /datum/game_mode/colonialmarines/ds_first_drop(obj/docking_port/mobile/marine_dropship)
+	if(!active_lz)
+		var/dest_id = marine_dropship.destination?.id
+		if(dest_id == DROPSHIP_LZ1)
+			select_lz(locate(/obj/structure/machinery/computer/shuttle/dropship/flight/lz1))
+		else if (dest_id == DROPSHIP_LZ2)
+			select_lz(locate(/obj/structure/machinery/computer/shuttle/dropship/flight/lz2))
+
 	addtimer(CALLBACK(GLOBAL_PROC, GLOBAL_PROC_REF(show_blurb_uscm)), DROPSHIP_DROP_MSG_DELAY)
 	addtimer(CALLBACK(src, PROC_REF(warn_resin_clear), marine_dropship), DROPSHIP_DROP_FIRE_DELAY)
 	DB_ENTITY(/datum/entity/survivor_survival) // Record surv survival right now


ihatethisengine:
errr, I think you are missing the mark with this change.

You can't launch the dropship before 15 minutes in the round. That means primary LZ will be auto-picked before you launch even if you (unlikely) launch as early as possible. I think it should be left 25 (20 in the round) minutes, but maybe ARES should send a reminder at 20 (15)? Setting primary LZ to whatever dropship lands first doesn't sound like a good solution, because splitdrops exist. I think it would be maybe better to forbid launching at all before primary LZ is selected? As an additional reminder.

Also while you are on it, LZ sentries are sent to the first LZ marines drop, instead of the primary LZ. Which many times screwed marines on splitdrops (or when you just needed to pick up survs on another LZ) and also inconsistent with weed killer. Should only drop sentries to the primary LZ, which would work well with no launch without primary LZ rule?

I think there was also a bug when either sentries or weedkiller was dropped on flyby, but I dont really remember

Drulikar:
> errr, I think you are missing the mark with this change.
> 
> You can't launch the dropship before 15 minutes in the round. That means primary LZ will be auto-picked before you launch even if you (unlikely) launch as early as possible. I think it should be left 25 (20 in the round) minutes, but maybe ARES should send a reminder at 20 (15)? Setting primary LZ to whatever dropship lands first doesn't sound like a good solution, because splitdrops exist. I think it would be maybe better to forbid launching at all before primary LZ is selected? As an additional reminder.
> 
> Also while you are on it, LZ sentries are sent to the first LZ marines drop, instead of the primary LZ. Which many times screwed marines on splitdrops (or when you just needed to pick up survs on another LZ) and also inconsistent with weed killer. Should only drop sentries to the primary LZ, which would work well with no launch without primary LZ rule?
> 
> I think there was also a bug when either sentries or weedkiller was dropped on flyby, but I dont really remember

See cited issue. Both me and Steelpoint were present in the game. Primary LZ was designated AFTER dropship flight. Also w/o this PR the requirement is *25 minutes* from world time. This PR is what makes it 15 minutes from round start, which intentionally matches the dropship console lock time.

Also keep in mind that w/o this PR its possible to designate a LZ that doesn't exist (LZ2) cited issue is on a map w/ a single LZ.

Sentries go where dropship lands. Weedkiller goes where the primary LZ is.

PR #10922, opened 2025-10-21
Disables factional patches from char prefs

Blundir:
# About the pull request

Title, CLF patch wasn't just commented out, it was removed from char prefs entirely

I really don't like the kind of awkward rp it brings, in some cases, as with CLF patch, it can also bring admin headache. Other patches could be used well if we'd actually have a factional character prefs, so these could only be used by members of faction those fit in, until that, it shouldn't be a thing.

# Changelog

:cl:
del: factional patches are commented out in char prefs, CLF patch is removed (none of this affects marines)
/:cl:

### DIFF
--- a/code/modules/client/preferences_gear.dm
+++ b/code/modules/client/preferences_gear.dm
@@ -1253,39 +1253,36 @@ GLOBAL_LIST_EMPTY(roles_with_gear)
 	category = "Civilian only (restricted)"
 	allowed_origins = list(ORIGIN_CIVILIAN)
 
-/datum/gear/civilian/patch
-	display_name = "Weyland-Yutani shoulder patch, black"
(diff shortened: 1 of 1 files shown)


BobBurgersLover73:
L

DonCycon:
if i knew where to apply feedback i would but this is total character death for civilian survivors
it adds a lot of flavor and potential rp to otherwise very generic colonists on a lot of maps.
i would not oppose this if we had factional character preferences or more fleshed out survivors/inserts on colonies,
but we dont, and most likely wont have for a long time.
:(

Detective-Google:
> where give feedback?

[here give feedback](https://forum.cm-ss13.com/t/pr-feedback-rules/1459)

PR #10960, opened 2025-10-25
Grant Field Doctor Medical Records Access

troytroy400:
# About the pull request

Adds missing medical records access to field doctor.

This is just an oversight fix. Grants field doctor access as they should have access on par with a normal doctor.

# Changelog

:cl:
add: Field doctors now have medical records access.
/:cl:


realforest2001:
No. Marine Database is used by more than just medical stuff, not giving it to Nurse at all. Would be better to split it into it's own thing.

realforest2001 (changes requested):
As above

troytroy400:
> No. Marine Database is used by more than just medical stuff, not giving it to Nurse at all. Would be better to split it into it's own thing.

Changed to just grant field doctor database access on-par with the normal doctor, I'll open a separate PR later splitting it for just medical records access.

PR #10970, opened 2025-10-27
Restore popout tacmap

FourkhanDaKhan:
# About the pull request

Re-adds the popout tacmap window for drawable tacmaps.

Also cleans a ton of dead/mostly dead code out of TacticalMap.tsx.

I cleaned up the worst behaviors of it but it still has a few odd behaviors that I'm not sure how to fix.

- On the first open, the bottom of the tacmap will be cropped. Requires a close and re-open to fix, then works forever.
- All others (that I know of) addressed!

Helps CIC's situational awareness and restores a feature that we lost when we moved to the PVE tacmap. Also provides a more ergonomic experience for users with multiple monitors.

<details>

[https://i.imgur.com/U9xIcFF.png] (Default width)
[https://i.imgur.com/UK3Ckey.png] (About the widest you can reasonably get it)

</details>

# Changelog

:cl:
add: Re-adds the tacmap popup window.
/:cl:

### DIFF
--- a/code/controllers/subsystem/minimap.dm
+++ b/code/controllers/subsystem/minimap.dm
@@ -952,8 +952,8 @@ SUBSYSTEM_DEF(minimaps)
 	if(desc)
 		closeToolTip(usr)
 
-/atom/movable/screen/exit_map/clicked(location, list/modifiers)
-	linked_map.on_unset_interaction(usr)
+/atom/movable/screen/exit_map/clicked(mob/user, list/mods)
+	linked_map.on_unset_interaction(user)
 	return TRUE
 
 /atom/movable/screen/minimap_tool
@@ -1010,10 +1010,10 @@ SUBSYSTEM_DEF(minimaps)
 	if(desc)
 		closeToolTip(usr)
 
-/atom/movable/screen/minimap_tool/clicked(location, list/modifiers)
-	if(LAZYACCESS(modifiers, LEFT_CLICK))
-		RegisterSignal(usr, COMSIG_MOB_MOUSEDOWN, PROC_REF(on_mousedown))
-		usr.client.mouse_pointer_icon = active_mouse_icon
+/atom/movable/screen/minimap_tool/clicked(mob/user, list/mods)
+	if(LAZYACCESS(mods, LEFT_CLICK))
+		RegisterSignal(user, COMSIG_MOB_MOUSEDOWN, PROC_REF(on_mousedown))
+		user.client.mouse_pointer_icon = active_mouse_icon
 	return TRUE
 
 /**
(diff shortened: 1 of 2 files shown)


tgui/packages/tgui/styles/components/Section.scss
Drulikar, line 1: You might want to be careful about globally changing this. It might be okay, I couldn't find any interfaces negatively impacted, but just be on the lookout.

FourkhanDaKhan:
Popout tacmap no longer receives tool inputs of any kind

Drulikar:
Does not appear to be fixed. See discord (too large for github)

Drulikar (changes requested):
drawing on the popout now seems correct; but theres still a couple problems:
- Popout map doesn't have the same close behavior as the normal tacmap meaning it can stay open and can be opened in duplicate w/ other tacmaps (see discord)
- Marine ground forces don't seem to have the correct tacmap (meaning they don't see drawings nor any icons now
image

FourkhanDaKhan:
CR should all be done

code/datums/components/tacmap.dm
```diff
@@ -51,21 +59,29 @@
...
 	map = null
 	for(var/mob/interactee in _interactees)
 		show_tacmap(interactee)
+		tgui_interact(interactee)
 
 /datum/component/tacmap/proc/popout()
-	tgui_interact(usr)
+	var/datum/tgui/maybe_ui = SStgui.get_open_ui(usr, src)
+	if (maybe_ui == null)
+		tgui_interact(usr)
```
Drulikar: an argument user should be used instead of `usr`. In general `usr` is only safe to use in verbs. See comment below:

code/controllers/subsystem/minimap.dm
Drulikar, line 1: The `clicked` procs appear to have the wrong/misleading arguments in this file. Once corrected that means you can for example pass the correct `user` to the popout proc.
image
Because https://github.com/cmss13-devs/cmss13/blob/1ea64484f6dad83b1c47d385318fdd875392d2c7/code/_onclick/click.dm#L76

code/datums/components/tacmap.dm
```diff
@@ -111,11 +132,21 @@
...
 	. = ..()
 
-	.["mapRef"] = map_holder?.map_ref
+	if (map_holder != null)
+		.["mapRef"] = map_holder.map_ref
+
+	.["isXeno"] = isxeno(user)
+	.["canChangeZ"] = FALSE
+
+/datum/component/tacmap/proc/close_popout_tacmaps()
```
Drulikar: Some calls to this provide a user (and this should be using that user instead of ever using `usr`)

code/game/machinery/computer/groundside_operations.dm
```diff
@@ -101,6 +101,10 @@
 	concurrent_users += WEAKREF(user)
 	onclose(user, "groundside_operations")
 
+/obj/structure/machinery/computer/groundside_operations/ui_close(mob/user as mob)
```
Drulikar:
```suggestion
/obj/structure/machinery/computer/groundside_operations/ui_close(mob/user)
```

code/controllers/subsystem/minimap.dm
Drulikar, line 1: Walking away from a tacmap table when you had a popout open will close the popout, but does not unset the client's `using_popout_tacmap` var making them unable to use tacmaps anymore w/o admin intervention.

image

PR #10972, opened 2025-10-27
Tweaks to Fun Facts

BasilHerb:
# About the pull request

Changes the calories burnt fun fact to be more 'realistic' (on the upper end of calories burnt for your average fully-encumbered marine in streneous combat conditions as in CM), while increasing the min requirement to compensate, plus changed 'damage taken' to be a bit higher since your average marine gets absolutely shattered every round. 

makes people stop complaining on the forums, also yeah the calories burnt thing was pretty pitifully low and didn't account for combat activities + heavy loads and the fact everyone is sprinting everywhere

# Changelog

:cl:
code: upped the average number of calories burnt
code: upped the minimum hurdle for calories burnt fun fact
/:cl:

### DIFF
--- a/code/datums/statistics/random_facts/damage_fact.dm
+++ b/code/datums/statistics/random_facts/damage_fact.dm
@@ -1,7 +1,7 @@
 /datum/random_fact/damage
 	statistic_name = "damage"
 	statistic_verb = "took"
-	min_required = 600
+	min_required = 400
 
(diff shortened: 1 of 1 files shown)


code/datums/statistics/random_facts/damage_fact.dm
Drulikar, line 1: No please. The point is for it to not become a leaderboard. 400 should be sufficient.

PR #10979, opened 2025-10-28
Prevents IBs/Organ Damage on the Hunting Preserve

matttheficus:
# About the pull request

You can no longer get IB or Organ Damage on the Hunting Grounds.

Getting your HG cut short because you were tapped once at the start is not fun, especially if you win the fight.

Speaking from both an HG fighter perspective and Predator perspective its really really annoying to either **have** the IB/giga brain damage/kidney damage/etc or kidnap the person to take them to the ship to fix the said injuries so I can actually fight someone who's able to actually stand up. 

This does not affect bone breaks at all.

<details>

</details>

# Changelog
:cl:
balance: you can no longer get IB or Organ Damage while on the Hunting Preserve as a FACTION_LIST_HUNTED
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -189,6 +189,9 @@
 	if(!owner)
 		return
 
+	if(owner.faction in FACTION_LIST_HUNTED) //Hunting Grounds
+		return
+
 	var/armor = owner.getarmor_organ(src, ARMOR_INTERNALDAMAGE)
 	if(owner.mind && owner.skills)
 		armor += owner.skills.get_skill_level(SKILL_ENDURANCE)*5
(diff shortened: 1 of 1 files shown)


Mister-moon1:
I would say remove organ damage too to avoid cheesing fractures

matttheficus:
I would be open to expanding it to all internal injuries except for fracs.

matttheficus:
As a note, since there is the chance a HG member is taken as a Thrall, they are given Hero Mob status then - so this change would not do anything different for them, as Hero Mobs get `NO_PERMANENT_DAMAGE`

byondsucks:
> Assuming someone takes a thrall from HG this will render them immune to IB whilst in the main game.

this doesn't matter as the pred would fix the thrall's IB if they live anyways, no?

PR #10988, opened 2025-10-28
Dump out unacidable/explo_proof items before deleting storage items

Nihisohel:
# About the pull request

Title.

If the object is important enough to have an explo_proof or unacidable flag, then we should probably stop it from being deleted when it's in acidable or explodable storage items

While it is technically a fix, it will be labelled as a balance change since theres going to be people up in arms about this

https://youtu.be/v4g2x1ahqOc

# Changelog
:cl:
balance: Explosion proof and unacidable objects may no longer be deleted while it's in a storage item, dumping it out before its deletion.
fix: The CO M1911C is now unacidable and unexplodable now similar to their other pistols
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -304,7 +304,6 @@
 
 /obj/item/weapon/gun/Destroy()
 	QDEL_NULL(in_chamber)
-	in_chamber = null
 	ammo = null
 	QDEL_NULL(current_mag)
 	target = null


Nihisohel:
ill need a maintainer to take a look at the failed checks from hard deletions in create_and_destroy.dm

code/game/objects/items/storage/storage.dm
```diff
@@ -912,14 +912,13 @@ W is always an item. stop_warning prevents messaging. user may be null.**/
 	SIGNAL_HANDLER
 	storage_close(watcher)
 
-/obj/item/storage/proc/dump_objectives()
+/obj/item/storage/proc/dump_important_items()
 	for(var/obj/item/cur_item in src)
-		if(cur_item.is_objective)
+		if(cur_item.is_objective || cur_item.explo_proof || cur_item.unacidable)
 			remove_from_storage(cur_item, loc)
```
kiVts: Second variable in remove_from_storage is the *new* location after removing. This removes from storage back into itself, This proc always harddels.

kiVts: Just remove the loc variable

Nihisohel:
yeah this is someone elses problem

PR #11010, opened 2025-10-30
Fixes issues with Synth Presets

realforest2001:
# About the pull request

Fixes missing rank for generalised synth and re-re-readds MP Synth as an option.

# Changelog
:cl:
fix: Fixed missing rank for Generalised synth.
add: Added MP Synth option, again.
/:cl:


Detective-Google (approved):
I think changing the option in the selector from "MP" to "Military Police" would increase clarity and reduce the need for cm jargon knowledge (anyone with the WL should know anyway but yk) 

otherwise code fine

PR #10950, opened 2025-10-24
Atomizes #9972 Part 3: Reagent Tank and Cart QoL Update

Puckaboo2:
# About the pull request

This is an atomized PR from #9972 that has to do with tanks and restock carts.

1. Adds a fill overlay indicator on reagent tanks.

2. Adds missing overlays for Yautja reagent tanks.

3. Adds proper icons and overlays for reagent and medical supply restock carts.

4. I was forced to repath tanks for no. 1 and no. 2, so there's a bunch of map edits. Sorry.

5. Icon touchups.

6. Adds a dispensing/filling indicator on reagent tanks. 

# Explain why it's good for the game

1. There is a little green fill overlay on top of reagent tanks that tells you how much of a reagent is left in a reagent tank so you no longer need to examine tanks for reagent levels. Now, please stop sending unga tanks with 30% juice left; they're still good to use.

2. After repathing everything, some missing overlays for Yautja tanks that slipped past the linter for years reared their ugly heads. I added them back. 

3. Drathek and Sammy wanted decent-looking restock carts and fill overlays for them, so I thought I'd take on that challenge. Thank you, Drathek, for the box and reagent container sprites. I made the little medical stock fills.

4. Yeah... I am not happy about this, either, but it was either this or adding fill code for every single type of tank, individually. Changing paths made it easier to add fill sprite and overlay codes.

5.They're not fresh coats of paint, but they're nice touch-ups. 

6. Sammy told me to and I think they look nice.

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Tanks now have a fill indicator that rises and falls every 200 units.
dreamseeker_2025-10-23_16-40-08

Works for Predator tanks, too.
dreamseeker_2025-10-23_16-48-56

Predator tanks now have overlays. I couldn't test the 'going boom' but if it works for regular tanks, it will work for this one.
dreamseeker_2025-10-23_16-50-57

Filling (Orange) and Dispensing (Green)
dreamseeker_2025-10-28_17-02-21

</details>

# Changelog

:cl: Puckaboo2
qol: Adds a fill indicator on reagent tanks that fills every 20%.
qol: Added a dispensing and filling indicator to reagent tanks that changes from green to orange.
qol: Adds the following transfer amounts to tanks: 15, 25, 80, 120, 150, and 240.
fix: Added missing overlays to Yautja tanks.
imageadd: Restock carts now have proper icons with fill overlays that add and remove a little box every 5 units of supplies.
[...]

PR #11022, opened 2025-11-01
Fixes flamer tank refueling removed from the Reagent Tank PR

Nihisohel:
# About the pull request

#10950 reverted the change after all the type path renames

We really shouldn't be silently reverting approved PRs immediately after its merging

It works

# Changelog
:cl:
fix: Flamer tanks can refill from other flamer tanks again
/:cl:


Drulikar:
> We really shouldn't be silently reverting approved PRs immediately after its merging

This is what happens when people don't know how to resolve merge conflicts and or use force push. And why atomizing is needed because reviewing code is very difficult.

Puckaboo2:
> > We really shouldn't be silently reverting approved PRs immediately after its merging
> 
> This is what happens when people don't know how to resolve merge conflicts and or use force push. And why atomizing is needed because reviewing code is very difficult.

Sorry... I atomized the PR the best I could, but some things may have gotten lost. That may have been one of them.

PR #11047, opened 2025-11-03
Facehuggers no longer eat your glasses or cigarettes

Nihisohel:
# About the pull request

Title.
This is for pugie.

Facehuggers lacked a check for the mask slot to drop any other item that isn't a mask to the ground, deleting it or putting it in null space instead

This fixes that, now facehuggers no longer has a diet of ray bans and cigarettes.

it works, you really don't need a video for this

# Changelog
:cl:
fix: Facehuggers no longer has a diet of ray bans and cigarettes, they will no longer eat your glasses or cigarettes on your face slot
/:cl:


Pugie:
🥳

PR #11080, opened 2025-11-07
Fixes Marine Sorting in Orbit

larentoun:
# About the pull request

Fixes marine sorting logic in Ghost Orbit. `ObservableSection` was overriding sorting by usual "sort by name" instead of checking if it was already sorted/need special sorting logic. So, it was sorting correctly, then overriding to default sorting.

Check images for comparison.

Fixing bugs is good, easier to find roles in descending priority order.

<details>
<summary>
Images
</summary>

Before (sorted ONLY by name)
Screenshot_30

After (sorted by Ranking and then by name)
Screenshot_29

</details>

# Changelog
:cl:
fix: Fixed Marine Sorting in Orbit. Now it first sorts by rank, then by name.
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/Orbit/index.tsx
+++ b/tgui/packages/tgui/interfaces/Orbit/index.tsx
@@ -402,7 +402,8 @@ const GroupedObservable = (props: {
             <ObservableSection
               color={x.color}
               title={x.title}
-              section={props.sorter ? x.members.sort(props.sorter) : x.members}
+              section={x.members}
+              sorter={props.sorter}
               key={x.title}
             />
           ))}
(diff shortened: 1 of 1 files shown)


Drulikar (approved):
https://github.com/cmss13-devs/cmss13/blob/1be59b9c167ad98b4f254a662339e7c46b4bdede/tgui/packages/tgui/interfaces/Orbit/index.tsx#L338-L346
Leaves a lot to be desired (e.g. a CO will appear after these), and its very awkward to be hardcoded in tgui code in case DM code renames them but future problem I guess.

image

larentoun:
> https://github.com/cmss13-devs/cmss13/blob/1be59b9c167ad98b4f254a662339e7c46b4bdede/tgui/packages/tgui/interfaces/Orbit/index.tsx#L338-L346
> 
> 
> Leaves a lot to be desired (e.g. a CO will appear after these), and its very awkward to be hardcoded in tgui code in case DM code renames them but future problem I guess.

The whole system is making me shook in my boots, but yeah, it's demanding a huge refactor in the future.

The whole "If Alpha is in name, make it Alpha squad" and etc was messing me up.

PR #11090, opened 2025-11-08
Fixes another flamer bug

Nihisohel:
# About the pull request

Title.

Fix, the line deletion allows flamers to fire past corners with non-dense objects again.
I don't know if this breaks anything else, or if it reverts the fix in #2143, but through my testing, I haven't found anything crazy

It works, doesn't seem like you can vault through the briefing thindows either

# Changelog
:cl:
fix: Flamers can now fire past some corners again
/:cl:


Nihisohel:
> What's the worst that could happen

dude

PR #11096, opened 2025-11-09
Fix some of the manufacturer logos being applied twice

Blundir:
# About the pull request

Title, fix of https://github.com/cmss13-devs/cmss13/issues/11081

:cl:
fix: fixed some cases of double manufacturer logo being applied
/:cl:


Git-Nivrak:
Loving this spaghetti implementation that was merged before it could be fixed

PR #11117, opened 2025-11-12
Tweak queen temporary maturity values

byondsucks:
# About the pull request

reduces the temp maturity duration to 20 seconds
reduces the duration increase from being shot to 10 seconds 

Ever since temp maturity was added, I haven’t seen a single queen player use it to actually defend the hive. Instead they walk from the core to the survivors and screech them. Now that the timer is 20 seconds, queen players won't be able to walk from the hive to the survivor hold on the other side of the map and still have maturity left, but they will be able to defend the hive in case of hive rushing. Keep in mind queens can still walk back to the hive to the core to refresh the duration even after it runs out.

# Changelog

:cl:
balance: Queen temp maturity duration is now 20 seconds.
/:cl:


Drulikar:
5 seconds warning is way too inadequate a warning that the maturity is waning.

Regardless I don't want this tweaked right now.

PR #11119, opened 2025-11-12
Allows Two Chefs to spawn at anytime, and changes most comments in job.dm to be autodoc comments

Nihisohel:
# About the pull request

Title. A friend asked me to do this
This is for the poor starving marines of the USS Almayer

It is pretty rare to get two MSTs despite the fact that one is scaled to open at 70 marines. Sometimes, even at 130+ pop, you wouldn't even see two due to how it is coded to only prefer active marines rather than total playercount.

That said, I also think messtechs work better with pairs and having one experienced and one inexperienced messtech would lead to better outcomes, and if both are inexperienced, then the shenanigans would be welcome as well, probably.

Should be relatively harmless for balance

Also, makes job.dm comments into autodoc comments

It works, don't need a video for it

# Changelog
:cl:
balance: Two chefs can now spawn at anytime, regardless of scaling
code: Most Job.dm comments are now autodoc'd for convenience
/:cl:

### DIFF
--- a/code/game/jobs/job/civilians/other/mess_seargent.dm
+++ b/code/game/jobs/job/civilians/other/mess_seargent.dm
@@ -1,30 +1,13 @@
 /datum/job/civilian/chef
 	title = JOB_MESS_SERGEANT
 	total_positions = 2
-	spawn_positions = 1
+	spawn_positions = 2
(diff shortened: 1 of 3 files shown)


Diegoflores31:
🥖

PR #11221, opened 2025-11-26
Remove unnecessary lobby xeno latejoin call

Drulikar:
# About the pull request

This PR removes an unnecessary call for late joining as a xeno from lobby. Some reason Harry thought that `attempt_to_join_as_xeno` returned a mob, and if that returned mob wasn't larva it would need a `transfer_xeno` call.

But there shouldn't be any behavior change from this change because returning FALSE would just trigger the observe message. Returning TRUE would attempt `/datum/game_mode/proc/transfer_xeno(xeno_candidate, mob/living/carbon/xenomorph/new_xeno)` where `new_xeno` is `TRUE` but that then wouldn't pass `if(!xeno_candidate || !isxeno(new_xeno) || QDELETED(new_xeno))` so it would do nothing anyways.

Less dead code.

<details>

https://youtu.be/gx3yR95evIM

</details>

# Changelog

No player facing changes.

### DIFF
--- a/code/modules/mob/new_player/login.dm
+++ b/code/modules/mob/new_player/login.dm
@@ -170,17 +170,13 @@
 				return FALSE
 
 			if(SSticker.mode.check_xeno_late_join(src))
-				var/mob/new_xeno = SSticker.mode.attempt_to_join_as_xeno(src, FALSE)
+				var/new_xeno = SSticker.mode.attempt_to_join_as_xeno(src, FALSE)
 				if(!new_xeno)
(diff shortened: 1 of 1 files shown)


hry-gh:
> Some reason Harry thought that attempt_to_join_as_xeno returned a mob, and if that returned mob wasn't larva it would need a transfer_xeno call.

just copypasta from the old lobby screen, i just get the diff
IMG_1119

Git-Nivrak (approved):

PR #11251, opened 2025-12-01
Add Morpher, Cluster, Recovery pylon and Plasma Tree to The Xeno Locator.

Venuska1117:
# About the pull request

Allows xenomorphs to locate any nearby Morphers, Clusters, Recovery pylon and Plasma Tree (sorted by distance)

This is mostly QoL, even if you use locate funtion to see direction where cluster is, its not as handy as tacmap, it show direction but it dont show you how you need to walk to get to it (courages to explore around to find it)

Also this PR fix runtime caused by tgui, when you are for example corrupted without queen or core and want to access locator, it will crash saying they cannot open empty list (and do it anyway after runtime).

Also don't say its broken, currently Morpher, Cluster, Recovery pylon and Plasma Tree send big fat font on chat when they are destroyed, if you worried this will give out that "some structures dissapeared" tunnels do dissapear too when they collapse.

By allowing xenos to track structures that are usually hard to find without any clues, only queen have access to see tacmap when off ovi, and requests to break clusters, but there is for example player who have 20 hours and dont know where to search (maps for them are maze/new) they dont know where to search for specific structure, with this change it would take less time to localize,

# Check photo below

<details>
<summary>Locator in action</summary>
Arrow points towards morpher on the right (showing it works)

image

</details>

# Changelog

:cl: Venuska1117
balance: Adds Morpher, Cluster, Recovery pylon and Plasma Tree to Xeno Locator.
/:cl:

### DIFF
--- a/code/__DEFINES/mob_hud.dm
+++ b/code/__DEFINES/mob_hud.dm
@@ -80,6 +80,10 @@
 #define TRACKER_HIVE "Hive Core"
 #define TRACKER_LEADER "Leader"
 #define TRACKER_TUNNEL "Tunnel"
+#define TRACKER_MORPHER "Morpher"
+#define TRACKER_CLUSTER "Cluster"
+#define TRACKER_RECOVERY_PYLON "Recovery Pylon"
+#define TRACKER_PLASMA_PYLON "Plasma Pylon"
 
 //These are used to manage the same HUD having multiple sources
 #define HUD_SOURCE_ADMIN "admin"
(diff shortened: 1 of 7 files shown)


Drulikar (changes requested):
If you're going to fill it up like this you're going to need to make a different interface than just a tgui_list (e.g. at a minimum a tgui_list that has icons) so that the kinds of things can be better distinguished. In general though I don't think this is a needed change.

Maybe look at xeno tracker interface for inspiration.

Venuska1117:
dont know how to achive it, its better is someone else do it, or maybe i will do it later if i ever understand how tgui works.

PR #11252, opened 2025-12-02
stop autobanning linux users

hry-gh:
title really

:cl:
server: server operators can now configure a list of CIDs that will not be considered for stickybans
/:cl:


Drulikar (approved):
Looks fine but prob ought to have a test subject.

PR #11261, opened 2025-12-03
Fix/Balance: M10 affected by one-handed stats & M10 + M10 Belt nerf & Attachment attack_verb affect gun

MistChristmas:
# About the pull request

## Fix One Handing:
The M10 previously had the `GUN_ONE_HAND_WIELDED` flag, causing it's wielded stats to take effect even when one handed. This is presumably an unintentional oversight. As it's attachments have one-handed stats specifically.

## Mostly Nerfing the M10, minor buffs
- M10 Belt reduced to 4 slots. M10 + 3 magazines.
  - The M10 Belt has had the highest capacity of any gun + magazine belt in normal use. Being far better than the M39 one that can only hold 2 magazines, and better than the normal pistol belt that holds 6. Given that M10 Magazines already have better capacity than pistol belts, I don't see why this should be the case.

   - M10 Magazines can also still fit in the large pistol magazine pouches already. Which is a big boon.
   - Reducing it's capacity to 3, _(matching with it's sprite, neat)_, still gives it something over the M39, whilst still holding more rounds than a pistol belt filled with 88 ammo in it. Not accounting for Drum or extended mags.
   
- Removed negative damage multi on M10 Gun, apply directly to Ammo damage instead.
  - This changes the damage calculation for the M10 Compensators. So I gave it a tier buff so it's slightly better than before. So it's +10% damage to the bullet.
- Remove AP/AP Extended availability.
  - The M10 Having AP and AP Extended Mags gives it comparably utility and performance close to the M39. This has been 
 stated to be a thing that shouldn't be. As the M10 should embody being a sidearm.
- Reduce Extended availability.
  - Given the kits already give 5 Extended Mags now(1 already in the weapon + 4), and this is the only sidearm that comes with extended mags, their availability has been reduced in any place they can be found.
- Following that the availability of it's attachments that give it that comparable performance have also been lowered.
  - Folding stocks added to places where the Solid stock already existed. Folding stock removed from the Squad weapon vendor to the attachment vendor. Lowered in availability.
  - All M10 Specific attachments (M10 Compensator, Micro Laser Sight, M10 Solid Stock) lowered in availability.
  - M10 Kit no longer comes with AP, replaced with Extended Mags.
  - Increased point-buy cost for M10 to be equal to SU-6/M44 costs in the same vendor.
- Reduce Magazine Box sizes
  - I don't see any reason they need to hold more than pistol ammo boxes. Base brought in line to 16 like other pistol ammo boxes
  - Extended at 10, Drum at 8.
- Removed the delay in extending or collapsing the Folding stock.
- Reduced base ammo speed. Not sure why it ever needed to be the fastest bullet when it's for CQC?
- Added Recoil to both one handed and wielded. Increase by one Tier each.
  - One handing this thing you should feel something, easily mitigated by either the stocks or a compensator.
- Increased the micro laser's affect to one handing by three tiers. 
- Increased the Full Auto scatter maximum and peak variables. Making the build up occur faster.
## Misc
- ~~Also Fixes the hitsound & attack verbs not changing with the attachments. And removes more un-needed stuff regarding just the M10 trying to do that.~~ It totally did work before I'm just dumb apparently. Well the sound did, but not the attack_verbs. But I slimmed it down anyway and moved it to gun_recalculate_attachments anyway. Plus added the override vars incase an attachment needs the sound/verbs, but someone doesn't want them on the gun.

- Also add the M10 and ammo to IOs Vendors

- Fix M10 Folding Stock not changing size when collapsed.

Additionally moves a define that didn't need to be above the M10 Belt.

- Fixes an oversight.
- Somewhat addresses the ammo economy regarding the M10 compared to other sidearms.
- Brings the M10 closer in line with the other side-arms.
  - I feel like main contention will be the removal of the AP. But that is currently one of three things that sit it above the other sidearms. The other two being the Full Auto rate of fire, and high capacity magazines. Removing the AP does lessen it's impact on the bigger targets by a good margin. Even though it's still possible under the right circumstances to use two normal mags to kill a Crusher. Something that comparatively the M39 can do in one mag, and the M4A3/4 can get it into crit with three.
    - Granted I'm not taking in TTK/Movement/Or Reload Time. And this was with 3 tiles of range. And when you move into 4 tile range, it needs three normal mags, performing similar to the M4A3/4.
  - With the AP present it has too much impact, akin to that of a Primary weapon, even with it removed it still stands as one of the best Sidearms to use if you have the attachments for it.
- Add handling for attachments changing hitsound that isn't specific to one attachments. And include attack_verbs. Plus a var to enable it.

<details>
<summary>Videos</summary>

Before flag change:

https://github.com/user-attachments/assets/5ba4052f-bd35-4d92-88b3-403cbf27b90b

HitSound + some of the attack verbage:

https://github.com/user-attachments/assets/c8ee5e62-1f0c-4065-a4cd-df38a8118eb9

</details>

# Changelog

:cl: MistChristmas
balance: M10 Holster Rig capacity reduced to M10 + 3 magazines.
balance: Buff M10 Compensator damage up a tier to 10%. Buff Micro Laser sight one-handed scatter by 3 Tiers
balance: Reduced M10 Attachment(M10 Compensator/Micro Laser/M10 Solid/Folding Stock) & Extended Mag availability.
balance: M10 Custom Kit comes with 4 + 1 Extended Mags.
balance: Removed AP/AP Extended Mag availability.
balance: Removed collapse/extending delay for M10 Folding stock
balance: Reduced ammo speed to in-between M39 and M39 AP
balance: Increased one handed and wielded recoil. Increased Full Auto scatter and build up.
fix: M10 uses it's unwielded stats whilst unwielded. Gun attachments that were supposed to changed the hitsound and attack_verb, now actually do that.
/:cl:


sippykot:
remeber when m10 was almost unusable garbage and then has been buffed to be a pretty ok gun as long as you use it in CQC?
...yeah...

MistChristmas:
> remeber when m10 was almost unusable garbage and then has been buffed to be a pretty ok gun as long as you use it in CQC? ...yeah...

Feel free to directly mention me in the discord rather than adding these sorts of comments to the GitHub.
As the maints look at you funny when you comment with nothing of substance.

code/modules/projectiles/gun_attachables.dm
```diff
@@ -629,7 +635,7 @@ Defined in conflicts.dm of the #defines folder.
 	recoil_mod = -RECOIL_AMOUNT_TIER_4
 	accuracy_unwielded_mod = HIT_ACCURACY_MULT_TIER_4
 	recoil_unwielded_mod = -RECOIL_AMOUNT_TIER_4
-	damage_mod = BULLET_DAMAGE_MULT_TIER_1
+	damage_mod = BULLET_DAMAGE_MULT_TIER_2
```
Diegoflores31, line 641: this attachment is not supposed to give + dmg , from what i am aware the fact that it gives a DMG boost is an oversight. 
RC downside is supposed to be the -10% dmg but this one is only positive buffs to the weapon , i think this should be considered.

code/game/objects/items/storage/belt.dm
```diff
@@ -1845,14 +1845,12 @@
...
-
 /obj/item/storage/belt/gun/m10
 	name = "\improper M276 pattern M10 holster rig"
-	desc = "Special issue variant of the M276 - designed exclusively to securely hold a M10 Auto Pistol and seven spare magazines, allowing quick access in close-quarters situations. Ideal for defending against boarding threats, this belt supports rapid deployment of high-rate sidearms while maintaining stability in zero-G environments."
+	desc = "Special issue variant of the M276 - designed exclusively to securely hold a M10 Auto Pistol and three spare magazines, allowing quick access in close-quarters situations. Ideal for defending against boarding threats, this belt supports rapid deployment of high-rate sidearms while maintaining stability in zero-G environments."
 	icon_state = "m10_armor"
 	flags_atom = FPRINT // has gamemode skin
-	storage_slots = 8
+	storage_slots = 4
 	max_w_class = 5
```
Diegoflores31, line 1848: you should probably put it on pair with the pistol pouch since you can just toss the pistol inside the pistol pouch anyways. 

Unless you make sure that the M10 cannot be used on the pistol pouch when it has the "rifle conversion kit" on it , similar to how the M39 has W_class restrictions if you slap too many attachments on it.

MistChristmas: Yeah, you're right it should be forced to use it's own holster or the shoulder holster/pistol pouch.

code/modules/projectiles/gun_attachables.dm
MistChristmas: Looking at it's description and the fact that it's completely exclusive to the M10. I think it was intended as written, whether it stays that way is another matter. I feel like reducing it's availability is alright enough for now. Until it's proven otherwise.

code/game/objects/items/storage/belt.dm
MistChristmas: Though with the stocks extended or the solid stock it already didn't fit in pistol pouches/holsters

code/modules/projectiles/gun_attachables.dm
Diegoflores31: the problem is that RC downside is suposed to be the damage reduction , it makes no sense to have attachments that are ++ versions

bast-guy:
Please, don't implement this PR. The M10 has barely crossed the line into "a fun gun to use with serious drawbacks that you can pick up every few rounds" territory, ceasing to be a meme.
And all this "don't discuss anything on GitHub" is just an excuse to completely get rid of public feedback that everyone can see. Just like with PR which introduced unreadable announcements.

bast-guy:
- ...causing it's wielded stats to take effect even when one handed. This is presumably an unintentional oversight.
  - You may be wrong, and that makes it all the more interesting to have at least one weapon that isn't a pistol and can be used by a marine without one hand. (Yes I know about the arm brace)

- M10 Magazines can also still fit in the large pistol magazine pouches already. Which is a big boon.
   - why?

- Reducing belt's capacity to 3, (matching with it's sprite, neat)
  -  Many storage sprites, including the large pistol mag pouch, do not correspond to how much they can hold.

- Given the kits already give 5 Extended Mags now(1 already in the weapon + 4)
  - It starts with four, since the one that comes with the gun itself is a non-extended, non-AP magazine. Are you sure you've took the kit at least once, and weren't some thin xeno caste main who was afraid to approach a person with an M10?

- Remove AP/AP Extended availability.
  -   No comments

- The M10 Having AP and AP Extended Mags gives it comparably utility and performance close to the M39. This has been
stated to be a thing that shouldn't be. 

  - Why shouldn't this be another option? A PDW instead of a full-size SMG? And for it to be an option, you'd have to install all the attachemtns on it.

- One handing this thing you should feel something
  - One-handed shooting, even with a compensator and a stock, has a large and unpleasant spread.

MistChristmas:
Oh boy I can't wait to get marked as off-topic in my own PR.

> * ...causing it's wielded stats to take effect even when one handed. This is presumably an unintentional oversight.
>   
>   * You may be wrong, and that makes it all the more interesting to have at least one weapon that isn't a pistol and can be used by a marine without one hand. (Yes I know about the arm brace)

I may be wrong? It literally is the same experience firing it wielded. This isn't something any automatic weapon needs.
You could also actually test if I'm wrong, and find that I'm not.
As for making it more interesting? Sure it's neat that you don't have to wield it. It's neat, because you get to use the highest RPM in one hand, and leave your other hand un occupied and don't have to worry about swapping hands and wielding it.
That doesn't make it balanced though. 
> * M10 Magazines can also still fit in the large pistol magazine pouches already. Which is a big boon.
>   
>   * why?

Because you get 6 ammo slots for the price of what would usually be 3. I'm not sure how to elaborate on why carrying more ammo is good for the weapon that has a 600rpm fire rate.
> * Reducing belt's capacity to 3, (matching with it's sprite, neat)
>   
>   * Many storage sprites, including the large pistol mag pouch, do not correspond to how much they can hold.

This was just a neat thing that happened. And not the actual reason for lowering but Yes.
> * Given the kits already give 5 Extended Mags now(1 already in the weapon + 4)
>   
>   * It starts with four, since the one that comes with the gun itself is a non-extended, non-AP magazine. Are you sure you've took the kit at least once, and weren't some thin xeno caste main who was afraid to approach a person with an M10?

Do I even dignify this one with a response? I guess I'll just tell you to read actually. It's literally in the quote you made.
> and weren't some thin xeno caste main who was afraid to approach a person with an M10?

I don't need to be on the receiving end of a weapon that has been over tuned from it's position as a supposed side-arm option to notice it has the performance better than primaries in some circumstances. 

> * Remove AP/AP Extended availability.
>   
>   * No comments

No please. Dispense your wisdom as you've already done with your other points. I'm not averse to actual criticism. Dispite a lack of substance so far. 
> * The M10 Having AP and AP Extended Mags gives it comparably utility and performance close to the M39. This has been
>   stated to be a thing that shouldn't be.
>   
>   * Why shouldn't this be another option? A PDW instead of a full-size SMG? And for it to be an option, you'd have to install all the attachemtns on it.

Because it out performs said SMG? There's little statistical reason to take the M39 over the M10 when it performs so well. As for the attachies. They are very much not hard to obtain. And you can still use the M10 belt even with all the attachments applied. It doesn't force you to not use the holster because it's outsized it.
> * One handing this thing you should feel something
>   
>   * One-handed shooting, even with a compensator and a stock, has a large and unpleasant spread.

As above in the first quote. It is the same experience as firing it wielded. Which again shouldn't be the case.


Even with all of the changes that this PR currently proposes. Whilst it leaves the M10 in a weaker state, and a much less competitive option to the M39. It is still a viable option. With good DPS. And more ammo convivence than just about any other gun other than maybe the SG. I'd also argue if you weren't going to take AP for the M39, you'd still be better off with the M10.

> And all this "don't discuss anything on GitHub" is just an excuse to completely get rid of public feedback that everyone can see. Just like with PR which introduced unreadable announcements.

Maybe just limit your salt to one PR at a time.
Oh also probably because the responses are usually unhelpful, foregoing discourse, whilst providing little or nothing usable in actual consideration. And the various powers that be decided that maybe they shouldn't allow contribs to be subjected to that.

neeshacark:
this pr kinda feels like a lot and should probably be atomized into smaller portions, it'll end up being processed quicker that way.

bast-guy:
I acknowledge the inaccuracies I made in my previous comment, but I still disagree that the gun truly outperforms the M39. Players would certainly pick up the M10 more often if it were unbalanced. The M10 clearly has issues dealing damage and simply hitting targets at ranges greater than 4 tiles. Giving it such an early nerf is not a very pleasant decision.
Removing the flag that gives the M10 two-handed shooting stats when firing one-handed(I misread this initially), and reducing the number of special magazines in the box would probably be fair, but the rest would simply kill a gun that only sees 5 players per round (even on high-pop) or even less. Removing AP ammo would kill any chance of using the M10 as a primary weapon, which, in my opinion, wasn't a bad thing.

Zenith00000:
> M10 Belt reduced to 4 slots. M10 + 3 magazines.

There's a reason it has so many slots - that reason is the speed in which you burn through the M10 mags is unlike any other gun, you'll run out very very quickly. This change doesn't really fit the theme of the gun in any case - fast firing, very quick ammo loss, you really need the storage.

> M10 Magazines can also still fit in the large pistol magazine pouches already. Which is a big boon.

Yes but you need to sacrifice a pocket slot to take this - a better solution would just not allow the mags in certain slots like the pistol pouch + webbing - instead of nerfing the belt overall. Not sure why you didn't just consider doing that instead. 

> Removed negative damage multi on M10 Gun, apply directly to Ammo damage instead.

I can agree with this change the damage values are bit off, but generally the damage is in a good spot, don't change it.

> Remove AP/AP Extended availability.

Honestly the AP is really good for it, but it's also good for the M39, only difference is the range, you basically have to be right in their face to actually do any damage, that's the difference. Also saying it has higher DPS is true, but that's only "Theoretical"

I did already severely nerf the availability, the standard size mags are pretty useless just like the regular standard size ones, you run out before being able to actually kill anything with both types.

> Reduce Extended availability.

Normal mags suck compared to AP, honestly not sure why you'd even consider nerfing their availability even more so. I can kind of understand AP, but extended normal doesn't really make sense.

> Folding stocks added to places where the Solid stock already existed. Folding stock removed from the Squad weapon vendor to the attachment vendor. Lowered in availability.

Fair enough, I probably added way too many attachments like this generally, not a big deal.

> All M10 Specific attachments (M10 Compensator, Micro Laser Sight, M10 Solid Stock) lowered in availability.

Yeah some nerfing in attachment availability is fine, considering the kits have them too.

> M10 Kit no longer comes with AP, replaced with Extended Mags.

Once again don't agree with this change, the kits are supposed to be special, nerfing the tiny bit of actually good AP mags you can get is not fun.

> Reduce Magazine Box sizes

They need to hold more because you'll be blasting through them way quicker then any pistol, it's just the nature of the insanely fast firing, ammo expending gun, if you understand the lore/stats of the gun. The theme of the gun, I don't think you're understanding what it's supposed to be really.

But yeah not a big deal either way, people never buy them in any case, because they're expensive for the little ammo and niche of the gun in any case.

> Removed the delay in extending or collapsing the Folding stock.

Yeah fair enough, good change.

> Reduced base ammo speed. Not sure why it ever needed to be the fastest bullet when it's for CQC?

I'd have made it shoot even faster really, but yeah. The nature of the M10 is that it literally fires insanely fast, you literally expend the mag in 2 seconds or so.  Or it's supposed to be like that, but yeah gameplay limitations.

> Added Recoil to both one handed and wielded. Increase by one Tier each.

Yeah I mean the stocks perhaps should be a bit more negative to the values, one handing shouldn't be so similar, although there is some difference.

> Increased the Full Auto scatter maximum and peak variables. Making the build up occur faster.

The issue with it before the changes was how horrible the scatter was, I don't know if changing it is really a good idea, I feel like it's in a good spot generally.

But yeah this might be good, it's supposed to actually have recoil when you hold it down in full auto, which isn't possible with how we don't actually have what "full-auto" is actually supposed to be, so emulating that or increasing it might be good, so I might agree with this change.

> Fix M10 Folding Stock not changing size when collapsed.

Yeah the stocks were an oversight, they're probably better if you can't use the solid one with the M10 belt, only the folding one folded, but I never changed that, might be a fair change if you do that.


Overall though you say it's a "minor" nerf, but this is a pretty major nerf generally. You also say the M10 has more DPS then the M39, but you're forgetting that's only theoretical. If you theoretically hit all your shots, can land every single round, which is a lot more difficult then it seems. 

The nature of the M10 is that it's a "CQC" gun, you literally have to be right in their face to actually do any serious damage, nerfing the damage + ammo availability even more +giving it scatter, and basically the majority of your nerfs will just nerf the gun back into obscurity again, perhaps not as terrible as it used to be, but honestly it's in a decent spot right now.

I can agree with the fixes - they're good, and some of the minor changes, but I think you need to reconsider what this gun is supposed to be, and not nerf it so severely.

Zenith00000:
> I acknowledge the inaccuracies I made in my previous comment, but I still disagree that the gun truly outperforms the M39. Players would certainly pick up the M10 more often if it were unbalanced. The M10 clearly has issues dealing damage and simply hitting targets at ranges greater than 4 tiles. Giving it such an early nerf is not a very pleasant decision. Removing the flag that gives the M10 two-handed shooting stats when firing one-handed(I misread this initially), and reducing the number of special magazines in the box would probably be fair, but the rest would simply kill a gun that only sees 5 players per round (even on high-pop) or even less. Removing AP ammo would kill any chance of using the M10 as a primary weapon, which, in my opinion, wasn't a bad thing.

Exactly, it might have "higher-dps" on paper, but in reality the M39 vastly out performs it in most situations - the only situation where the M10 could is in a tight narrow hallway with the Xenos rushing you head on and nowhere to side-step, which is literally what the gun is designed to be for in the lore.

> The M10 Auto Pistol was designed to aid [United Americas Aerospace Force] space crews in defending against boarding parties.

> The M10 was capable of firing at an exceptionally high rate, making it devastating in close quarter combat against soft targets. Its low caliber ammunition made it useful aboard internal space ships and stations, being too weak to puncture internal systems or outer hull; that lack of penetrating power made it ineffective against armor, however. The M10's high rate of fire was also a drawback, making it difficult to control in freefall and depleting entire magazines in a single second on full auto.

Diegoflores31:
> > I acknowledge the inaccuracies I made in my previous comment, but I still disagree that the gun truly outperforms the M39. Players would certainly pick up the M10 more often if it were unbalanced. The M10 clearly has issues dealing damage and simply hitting targets at ranges greater than 4 tiles. Giving it such an early nerf is not a very pleasant decision. Removing the flag that gives the M10 two-handed shooting stats when firing one-handed(I misread this initially), and reducing the number of special magazines in the box would probably be fair, but the rest would simply kill a gun that only sees 5 players per round (even on high-pop) or even less. Removing AP ammo would kill any chance of using the M10 as a primary weapon, which, in my opinion, wasn't a bad thing.
> 
> Exactly, it might have "higher-dps" on paper, but in reality the M39 vastly out performs it in most situations - the only situation where the M10 could is in a tight narrow hallway with the Xenos rushing you head on and nowhere to side-step, which is literally what the gun is designed to be for in the lore.
> 
> > The M10 Auto Pistol was designed to aid [United Americas Aerospace Force] space crews in defending against boarding parties.
> 
> > The M10 was capable of firing at an exceptionally high rate, making it devastating in close quarter combat against soft targets. Its low caliber ammunition made it useful aboard internal space ships and stations, being too weak to puncture internal systems or outer hull; that lack of penetrating power made it ineffective against armor, however. The M10's high rate of fire was also a drawback, making it difficult to control in freefall and depleting entire magazines in a single second on full auto.

except it doesnt have any recoil here , free AP in insane amounts , and the only weapon in the game that has extended+AP mags .

its damage is not on paper , unline m39 you can dualwield and 1 hand this thing and it does damages not fit for a sidearm.

m10 is literally the peak of sidearms right now , there is almost no reason to pick any other.

Zenith00000:
> except it doesnt have any recoil here , free AP in insane amounts , and the only weapon in the game that has extended+AP mags .
> 
> its damage is not on paper , unline m39 you can dualwield and 1 hand this thing and it does damages not fit for a sidearm.
> 
> m10 is literally the peak of sidearms right now , there is almost no reason to pick any other.

I mean nerf the availability, but don't outright remove it, especially not from the kits.

Also I understand nerfing the one-handing it, but yeah make the stocks just perform better as they should, but yeah it's also a very light "pistol" so don't just make it insanely inaccurate when not two handing, as you'd still be able to handle it's recoil up until a certain point - especially without a stock.

MistChristmas:
> I mean nerf the availability, but don't outright remove it, especially not from the kits.

It's a side-arm. The other side arms that have AP aren't full auto. It already dominates this space so much it competes with the M39 in similar duty. I don't really consider it required, when you can still 2 mag a crusher with extended mags in DPS. That is already on par with the mod 88 in magazine efficiency. It does not need to be better at all.
Heaven forbid I actually test the TTK comparison with it.

And sure the comparisons I'm using are ideal circumstances in effective range, but when they're not ideal circumstances, it's the same for the other weapons. Beyond counting for the effective range, They're also going to miss. And suffer from the user being in pain. Given that most xenos need to close the gap anyway to be effective, I think I can mostly ignore the effective range.

You can attach this weapon so it's at the peak of it's ability and still have it fit in the holster it comes with.
What does the M39 do if you have a Stock + V grip? It doesn't fit in the holster that gives it 2 extra mags.

> they're probably better if you can't use the solid one with the M10 belt, only the folding one folded, but I never changed that

I think changing how much the belt holds is fine enough for now. Sacrificing a pocket slot for 6 slots of ammo is something for people to consider. And at this junction I don't want to repath it to submachine gun magazines and have it not fit in the shoulder holster (Or snowflakey make it fit in there). It's supposed to be a side-arm.

> They need to hold more because you'll be blasting through them way quicker then any pistol

They already hold 2x the amount of the highest capacity pistol the marines normally get. You could have adjusted the Ammo Cost to buy for req but even still it is and would have been egregious in both cases.

> The issue with it before the changes was how horrible the scatter was, I don't know if changing it is really a good idea,

Admittedly I'm unsure how much I need to change it. But it certainly needs tweaking from what it was.

> I'd have made it shoot even faster really, but yeah.

This is the bullets speed traveling through tiles and not the fire rate? In CQC the bullets speed really doesn't matter all that much. So I have no idea why you made it faster than everything in the game.

> Overall though you say it's a "minor" nerf

I had previously said it was a minor nerf when I wasn't 24 commits deep. It definitely isn't one now. 

> I can agree with this change the damage values are bit off, but generally the damage is in a good spot, don't change it.

The damage values are unchanged. Apart from the compensator. Which gets slightly buffed.
You set the bullet damage higher than needed, and then had the gun multiplier drop it down. For what reason, I really don't know.

I'll respond to your other points at a later time.
But over-all my intent is still to give this weapon a serious nerf and bring it closer to it's other side-arm counterparts.
In a discussion of sidearm balance we shouldn't have to bring in the primary weapons as comparisons, especially not about them being out performed.

Zenith00000:
Please just keep the Extended AP mags exclusively for Req & the kits - instead of out-right removing them from everything. That makes them much less available, but also makes getting a kit worth it. In fact I'd kind of prefer if the kits pretty much only had the attachments and perhaps extended AP mags now that I think about it, but yeah that would probably make them too exclusive.

As for the regular Extended mags, you can nerf their availability, but honestly they don't seem as powerful overall as AP, and I've never had much luck compared to using AP with the M10. You can probably nerf their availability, but they should probably still be available in small quantities in the squad vendor perhaps.

Also I noticed you removed the empty boxes in req, they're there just so you have the option of stocking it up with them, rather then actually having them supplied there, there probably should be more box types there considering, but yeah they don't actually do anything as far as I know, just provide a slot if you ever choose to stock it.

MistChristmas:
> Please just keep the Extended AP mags exclusively for Req & the kits - instead of out-right removing them from everything. That makes them much less available, but also makes getting a kit worth it. In fact I'd kind of prefer if the kits pretty much only had the attachments and perhaps extended AP mags now that I think about it, but yeah that would probably make them too exclusive.
> 
> As for the regular Extended mags, you can nerf their availability, but honestly they don't seem as powerful overall as AP, and I've never had much luck compared to using AP with the M10. You can probably nerf their availability, but they should probably still be available in small quantities in the squad vendor perhaps.

They aren't as powerful as AP. When you're looking to use this gun on T3s but you shouldn't be looking to use this gun off of it's performance against the biggest targets. The side arms that get AP are limited in their function because they don't get Full Auto. And the one that gets the highest capacity fires (in theory) 3x less RPM than the M10. Whilst they have just over a quarter less damage per bullet. And 2x or 3x in the case of Extended AP the magazine capacity.

In your view you think that this weapon should have more than 2.5x - 4x the rounds in one belt compared to a pistol belt with the mod 88. And again this really isn't a fair comparison at all because the mod 88 doesn't have full auto. You're beating it so badly into obscurity with the M10 being the way better choice for anyone that I doubt anyone considers the M10 in place of a pistol, they consider it in place of a primary, if they just want to scare off runners with a burst? Or actually chase something down. It shouldn't have AP, it shouldn't have AP Primary Performance, and it definitely shouldn't have AP Extended mags on top of all of that. And that's just some of it's numbers compared to the highest capacity side-arm before it came along.

And you could have made it into an exclusive kit weapon which wouldn't adjust all of the issues I currently see with it. But it would have certainly had a better excuse for being over-tuned with the fact that it's highly limited in it's availability.
However I do actually like the idea of a Full Auto Side-arm, that doesn't just out perform the other options. So I'm not going to pursue that myself.

> Also I noticed you removed the empty boxes in req, they're there just so you have the option of stocking it up with them, rather then actually having them supplied there, there probably should be more box types there considering, but yeah they don't actually do anything as far as I know, just provide a slot if you ever choose to stock it.

Fluff that will never be filled because their mags can't be bought. I don't see the point. It's just clutter that all the CTs will gloss over. - Let me actually expand on this. You have to be able to put the magazine in the vendor for it to fill a box automatically or put in a whole box. You can't do the former for drum mags on master. And you'd never do the latter as a CT.
And also XM51 boxes are in there for some reason. Time for a commit.

MistChristmas:
Also for reference. 
Performed whilst moving slightly with my skill issue aim. With Reflex, Solid Stock, micro laser, and compensator and normal extended ammo.
Had I used AP extended ammo I would have crit the crusher and it lying down might have skewed the numbers more than my skill issue had already.

Not really a completely fair and controlled test, but I'm not going to all that trouble.
Anyone missing shots within this range is due to themselves. Not accounting for pain. Or leadership auras.
The M10 has nearly triple the fire rate of the mod88 and the M39 is beat out by it have +50% more fire rate than it.
How you can say it's not competing with the M39 just because you need to be in 3 tiles instead of 4, I don't know. Even with the falloff the AP still performs well at 4 tiles because of it's rate of fire.
At the same range of 4 tiles, The M10 only lags behind in TTC for a Crusher by like ~1 second if even that. 

<details>
<summary>Uncontrolled Full Auto - Just holding M1</summary>

https://github.com/user-attachments/assets/d9f1c1bd-8c41-427d-8397-8ea491d34769

</details>

And to note for this one, it basically takes until you've expended a normal M10 Mag until the bloom really starts to kick in. Which is how it's currently outlined in code that it doesn't reach max full auto scatter until you have already shot a normal mag.

<details>
<summary>Controlled Full Auto - Occasionally stopping</summary>

https://github.com/user-attachments/assets/f8a3915f-c8a2-476b-9f04-1b3189fb2c7f

</details>

Zenith00000:
Once again I implore you to just sit and observe a few rounds, see how many people actually even use the M10, how effective it is. Rather then basing it off limited testing in a stationary position. Personally I've watched probably more then 50+ hours of rounds and people using it since, from what I've seen it's nowhere near as "OP" as you're claiming. Everything you talk about is in the realm of theory, doesn't take into account people skill-level, server ping, random factors and stuff like that in general. Situations are never ideal.

It's a blanket nerf without many considerations. Please just keep it mostly the same - fix the broken aspects and issues, you can nerf some of the ammo availability but nerfing it as much as you've done is just gonna once again bring it back into obscurity and nobody will use it at all again - which would be a shame.

MistChristmas:
The other guns share these non ideal situations. 

Power wise. Sure it's not that amaIngly OP compared to the M39. Despite the availability of these exclusive attachments and like double the AP. Because you stack normal sized and extended.

But it shouldn't be on the M39s level. It is a sidearm. And it far outstrips every other side arms performance. It's overturned in many areas. 
And whilst I might give it leeway in some areas, leaving it with any AP availability isn't one of them.

I'm not sure there's any point in rehashing things I have already said.
If you think this is going to turn it into unusable and obscure, I'd urge you to play with the actual sidearms. And experience the disparity.

Zenith00000:
> But it shouldn't be on the M39s level. It is a sidearm. And it far outstrips every other side arms performance. It's overturned in many areas. And whilst I might give it leeway in some areas, leaving it with any AP availability isn't one of them.
> 

Just leave the AP for the kits at least, that's two mags, you're really saying that's too much? Also a small amount at req too, you can't possibly say that that's unbalanced? for let's say 1-2 people who actually bother to even use the M10, it's hardly gonna make any difference.

Issue #10667, opened 2025-09-19
HPR: un-deploying it's bipod buffs it's fire rate.


MistChristmas:
### Testmerges

N/A

### Description of the bug

Or failing to deploy and undeploy it's bipod nerfs your fire-rate. Pick one I suppose.

HPRs spawn with attached bipod, but that bipod's delay mod doesn't seem to take effect until after the bipod is undeployed.

EDIT 1: Actually it's probably that the nerf for having the undeployed bipod doesn't take affect after you undeploy it rather than whatever I said above. 

### What's the difference with what should have happened?

Bipod's undeployed stats should take affect on attach/spawn. Instead of needing the user to do so beforehand.

EDIT 1: The bipods nerf stats should take effect whilst it's still attached to the weapon and not go away because you undeployed it.

### How do we reproduce this bug?

1. Spawn an HPR
2. See it's stats list 133 rpm for the fire-rate
3. Deploy it's bipod it spawned with and undeploy it.
4. Notice the firerate is now 200 rpm or it's gone from fire_delay = 1 to 3
...

### Issue Bingo

- [x] Issue could be reproduced at least once
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Wintermote:
This sounds like it's a bug with the stat card not updating rather than any actual change to the HPR rate of fire.

MistChristmas:
> This sounds like it's a bug with the stat card not updating rather than any actual change to the HPR rate of fire.

You'd think so, but when I tested it on local it seemed to affect the fire rate.

Issue #8046, opened 2025-01-08
M10 Autopistol Automatic doesnt work in akimbo


TheManWithNoHands:
### Testmerges

#7909, #7850, #7716, #7165, #7946, #8007, #7830, #7607, #8034, #8015, #7511

### Round ID

25791

### Description of the bug

When dual wielding the M10 autopistol to fire them in akimbo, it doesnt let you fire in full-auto.
Even when the guns are set to full-auto, you need to fire them in semi-auto. Needing to click for each shot.

### What's the difference with what should have happened?

Switching to full-auto mode should allow players to fire in full-auto. Holding down the mouse button to keep firing.

### How do we reproduce this bug?

1. Hold one M10 autopistol in each hand
2. Set akimbo mode to switching
3. Set each M10 Autopistol to automatic fire
4. Hold down the trigger

### Issue Bingo

- [X] Issue could be reproduced at least once
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Zenith00000:
I don't think you can akimbo any guns properly now, it was changed a while ago, I could be wrong. The M10 is technically a pistol so that might be the issue perhaps, not 100% sure.

TheManWithNoHands:
> I don't think you can akimbo any guns properly now, it was changed a while ago, I could be wrong. The M10 is technically a pistol so that might be the issue perhaps, not 100% sure.

Oh it works in akimbo. If you set the akimbo mode to fire both at the same time it does that. Just doesnt with the switching akimbo mode. Like, switching changes the gun each shot as you hold down the trigger. But it seems like it no longer registers you pressing down the mouse once it switched guns. 

Honestly, that migth be on purpose. Being that the M10 is the only pistol with full auto. So switching akimbo with every other pistol, who are all semi-auto, is perfect to have to click for each shot and have the gun switch each shot.

I really migth be that switching akimbo is uniqe for pistols.

ItsVyzo:
Might be fixed with #10495

PR #11312, opened 2025-12-13
Fix: Bipods buffing the fire rate after one undeploy + Bipod Forever Slow + Akimbo with M10s

MistChristmas:
# About the pull request
Resolves: #10810, Resolves: #10667, Resolves: #8046

- Adjusted one number to be the same as the other. 

- Bipod is supposed to removes it's signal on detach. But if it was being replaced, the user wasn't passed. So we shoehorn that information in.
- For automatic weapons, using swap hands to fire them has no practical use, 
And apparently some people thought it was a bug that their preference did what they wanted? That or someone didn't clean up the lingering issues.
So it now forces simultaneous.
- Fixes a niche issue I noticed in testing where if you shot on simultaneous, and then on alternate, there wouldn't be any signals for autofire on the gun. By just slapping a somewhat dirty check there and applying them if needed.

Bugfix good.

<details>

</details>

# Changelog

:cl: MistChristmas
fix: Bipods don't erroneously buff the gun after first undeploy. Bipods don't apply forever slow.
fix: Using akimbo simultaneous fire and then switching to alternate fire doesn't break shooting.
qol: Akimbo defaults to simultaneous fire if you use automatic weapons for autofire.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -2091,6 +2091,10 @@ not all weapons use normal magazines etc. load_into_chamber() itself is designed
 
 /obj/item/weapon/gun/proc/set_gun_user(mob/to_set)
 	if(to_set == gun_user)
+		if(!(comp_lookup[COMSIG_MOB_MOUSEDOWN]) && to_set)
+			RegisterSignal(gun_user, COMSIG_MOB_MOUSEDOWN, PROC_REF(start_fire))
+			RegisterSignal(gun_user, COMSIG_MOB_MOUSEDRAG, PROC_REF(change_target))
+			RegisterSignal(gun_user, COMSIG_MOB_MOUSEUP, PROC_REF(stop_fire))
 		return
 	if(gun_user)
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/gun_attachables.dm
```diff
@@ -3974,7 +3977,7 @@ Defined in conflicts.dm of the #defines folder.
 	scatter_mod = SCATTER_AMOUNT_TIER_9
 	recoil_mod = RECOIL_AMOUNT_TIER_5
 	burst_scatter_mod = 0
-	delay_mod = FIRE_DELAY_TIER_12
+	delay_mod = FIRE_DELAY_TIER_11
 	//if we are no longer on full auto, don't bother switching back to the old firemode
```
Git-Nivrak, line 3981: Why?

MistChristmas: It's what it starts with.
You'd prefer I do it the other way?

Issue #11307, opened 2025-12-12
ACID HARNESS large and larger beaker


seanlol092:
### Testmerges

#10999: 'Fix of various plane related issues' by Blundir at commit f85a7153c4 #10954: 'Liquid fuel decals now ignite & Toxic no longer destroys weeds' by Drulikar at commit fd225b3883 #10908: 'Adds additional checks to pouncing up a z level' by Git-Nivrak at commit b3295ad70b #10925: 'Cape River Mining Colony - Ground Map' by Steelpoint at commit 7e02380a3a #10375: 'Corporate Bodyguard' by realforest2001 at commit e96d0616b1 #11136: 'Coms Rolling Encryption & Punchcards' by Drulikar at commit c664f54b60 #10932: 'Lazy load map multiz fix' by cuberound at commit c5638a9bad #10935: 'Desert Moon - Hunting Grounds Map' by Unknownity at commit cbd22be694

### Description of the bug

ACID HARNESS about to use all "/obj/item/reagent_container/glass/beaker"'s children

### What's the difference with what should have happened?

as [#10793](https://github.com/cmss13-devs/cmss13/pull/10793) says it should be small Glass Beakers.

### How do we reproduce this bug?

1. get acid harness
2. put beaker (hi cap) and battery
3. activate it
4. now you have "Medicine administered. 428 units remaining."

<img width="583" height="36" alt="Image" src="https://github.com/user-attachments/assets/42ab5138-3546-495c-b204-c4289f0d713b" />

### Issue Bingo

- [x] Issue could be reproduced at least once
- [ ] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Issue #11306, opened 2025-12-12
ACID HARNESS injection


seanlol092:
### Testmerges

#10999: 'Fix of various plane related issues' by Blundir at commit f85a7153c4 #10954: 'Liquid fuel decals now ignite & Toxic no longer destroys weeds' by Drulikar at commit fd225b3883 #10908: 'Adds additional checks to pouncing up a z level' by Git-Nivrak at commit b3295ad70b #10925: 'Cape River Mining Colony - Ground Map' by Steelpoint at commit 7e02380a3a #10375: 'Corporate Bodyguard' by realforest2001 at commit e96d0616b1 #11136: 'Coms Rolling Encryption & Punchcards' by Drulikar at commit c664f54b60 #10932: 'Lazy load map multiz fix' by cuberound at commit c5638a9bad #10935: 'Desert Moon - Hunting Grounds Map' by Unknownity at commit cbd22be694

### Description of the bug

Acid harness injection goes to 3 seconds probably less

https://github.com/user-attachments/assets/ebafcf86-920e-4c81-b706-56ea2bb7a0f7

### What's the difference with what should have happened?

acid harness injects every 20 second as the says at the [code](https://github.com/cmss13-devs/cmss13/blob/f7b0cedeb10689c0074d7a6f9014697b6fe6c2d3/code/modules/reagents/chemistry_machinery/acid_harness.dm#L479)

addtimer(CALLBACK(src, PROC_REF(recheck_conditions)), 20 SECONDS * inject_amount)
and
addtimer(CALLBACK(src, PROC_REF(recheck_conditions)), 20 SECONDS * inject_amount)

### How do we reproduce this bug?

1. get acid harness with chem and battery on it
2. set vital levels on and set "and" (damage type can brute for faster activation)
3. stab yourself (for early activation)
...

### Issue Bingo

- [x] Issue could be reproduced at least once
- [ ] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

Issue #11208, opened 2025-11-24
A.C.I.D. Harness Won't inject reagents w/ no OD limit.


nuklearcellphoneg:
### Testmerges

1

### Round ID

31073

### Description of the bug

So, let's say you fill a vial with space milk, or Tomato Juice, or Water, and use that inside the ACID Harness. It will refuse to inject due to Overdose danger, but there is no danger, as there is no set unit amount it will OD. It seems the absence of an OD is treated like a possible OD.

### What's the difference with what should have happened?

It should inject Sugar, Tomato Juice, Water, Nutriment, Milk, Orange Juice, Blood, Lime Juice, Tea, so on & so forth.

### How do we reproduce this bug?

1. Used ACID Harness w/ a vial of Sugar.
2. Use Multitool, set injection to 1u & threshold to Burn, Brute, Oxygen, Toxin, and do some damage
3. Observe its refusal to inject.

### Issue Bingo

- [x] Issue could be reproduced at least once
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

PR #11317, opened 2025-12-14
Fix: Acid Harness accepting larger beakers + Interval + Mutli-Tooless Configuring

MistChristmas:
# About the pull request

Resolves: #11307, Resolves: #11306, Resolves: #11208

- ~~Adds some cant_holds for the beakers so we're not shoving 500u beakers in.~~ Adds some `can_be_inserted` logic checking if the volume is above a Maximum volume for the acid harness which is currently 60

- Adds a time interval between scans rather than just firing on every process. Currently 15 seconds.
  - Scan time modified to be +15% or +20% per unit on successful injection or OD cancelled injection
  
- Check for Multi-tool in hands before we allow the user to change the configuration settings.

- Check for OD limit before we protect against ODs

Added RETURN_TYPE(/obj/item) to two more procs instead of just the active hand one.

Buggie Fixes
<details>

</details>

# Changelog
:cl: MistChristmas
fix: Acid Harness no longer accepts larger beakers than normal, + needs a multi tool to change configurations, + doesn't protect against reagents that actually have no OD limit
balance: Acid Harness no longer scans so frequently, instead scans every 15 Seconds further modified for successful injection(+15% per U) or OD Warning(+20% per U)
/:cl:


code/modules/reagents/chemistry_machinery/acid_harness.dm
```diff
@@ -43,13 +43,17 @@
 		/obj/item/reagent_container/glass/beaker,
 		/obj/item/cell,
 	)
-
+	cant_hold = list(
```
Diegoflores31: i fell this should probably be a check for volume instead . god know what new containers players will create in the future

MistChristmas: I feel like just updating the list if that ever happens is easier. The better solution would be to have a storage handling for ispath-ing the given item in the list but I don't really want to do that.

image
Doing the way above, has issues where quick equip will just disappear the object. Because it just presumes that it works.
And for some reason has no attempt_item_insertion
And I really don't want to try to adjust that. Especially not in this PR.

Diegoflores31: isnt there a can_hold proc that where you can add a check ?

MistChristmas: `can_be_inserted`

Which does work for this.
Fair enough

code/modules/reagents/chemistry_machinery/acid_harness.dm
```diff
@@ -48,7 +48,7 @@
 		/obj/item/cell,
 	)
 
-/obj/item/storage/internal/accessory/black_vest/acid_harness/can_be_inserted(obj/item/object_to_insert, mob/living/user)
+/obj/item/storage/internal/accessory/black_vest/acid_harness/can_be_inserted(obj/item/object_to_insert, mob/living/user, stop_messages = FALSE)
 	if(istype(object_to_insert, /obj/item/reagent_container/glass/beaker))
 		var/obj/item/reagent_container/glass/beaker/beaker_check = object_to_insert
 		if(!beaker_check.reagents || beaker_check.reagents.maximum_volume > ACID_CONTAINER_MAXIMUM_VOLUME)
```
Diegoflores31, line 54: why are you checking if its full or empty ? the message just says its too large

MistChristmas: It doesn't care about any reagents actually in it. It's just a nullsafe check.
If for any reason the beaker doesn't have reagents.
tbf I could just shorten it to `beaker_check?.reagents.maximum_volume > ACID_CONTAINER_MAXIMUM_VOLUME)`

And maximum volume is the total it could hold. Not the total it is holding.

MistChristmas: Actually no I couldn't cause I want the early return if we somehow are missing `.reagents`

PR #11318, opened 2025-12-14
Stopgap solution to Polaroid generation lagging the server

Nihisohel:
# About the pull request

'Fix' for an exploit, and some minor additions as well

Title, adds a cooldown between camera shots, as the polaroid picture generation will lag the server to a standstill if spammed multiple times.

I am genuinely surprised nobody has at least tried to do a cooldown for this long.

Additionally, pressing unique-action will now switch its camera focus, unlike the previous iteration of using rightclick context menus, and the old disposable camera can now be purchased in the character loadout screen

Well it works, don't need a video for this

# Changelog
:cl:
add: Adds the old disposable camera as a 3 point loadout purchase, go out and do some Frank West larp I guess
balance: Changed the size of the old disposable camera from small to tiny.
fix: Adds a delay to picture taking as to not lag the server
qol: Pressing Unique-action on polaroid cameras will now switch its focus.
del: Removed the change focus verb for redundancy.
/:cl:

### DIFF
--- a/code/modules/paperwork/photography.dm
+++ b/code/modules/paperwork/photography.dm
@@ -141,6 +141,7 @@
 	var/pictures_left = 10
 	var/size = 7
 	var/cooldown = 0
+	var/static/list/possible_focus = list(1, 3, 5, 7)
 
 /obj/item/device/camera/unique_action(mob/user)
 	change_size(user)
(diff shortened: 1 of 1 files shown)


code/modules/paperwork/photography.dm
```diff
@@ -156,14 +162,18 @@
...
+/obj/item/device/camera/proc/change_size(mob/user)
+	switch(size)
+		if(1)
+			size = 3
+		if(3)
+			size = 5
+		if(5)
+			size = 7
+		if(7)
+			size = 1
```
SpypigDev: might be easier to do this with a next in list sorta system
good PR overall though

Nihisohel: think i did something similar when i was figuring out ways to do the accessory style system PR, lemme figure something out, probably some merit to making it dynamic in that way

Nihisohel: made sizes a static list, and the change_size proc dynamically fetches the size now

SpypigDev:
first on my review list, promise

PR #11321, opened 2025-12-14
Fixes being able to use wielded item offhand 'item' as a weapon and another fix to TGUI strip panels

Nihisohel:
# About the pull request

Confusing title, I know.

Exploit fix, although I'm probably the only one that knew how to reproduce it properly

also fixes items being put into nullspace to non-existent limbs, yep, it existed for a long, long time

It works, yes
No, there won't be a video

# Changelog

:cl:
fix: You will no longer be able to use the offhand 'item' as a deadly weapon.
fix: Items will no longer be inaccessible when attempting to place an item on a limbless slot in the TGUI strip panel
/:cl:

### DIFF
--- a/code/datums/elements/strippable.dm
+++ b/code/datums/elements/strippable.dm
@@ -221,16 +221,17 @@
 		to_chat(user, SPAN_WARNING("[source] is missing the limb for this slot!"))
 		return FALSE
 
-	if(!equipping.mob_can_equip(source, key))
-		to_chat(user, SPAN_WARNING("[equipping] doesn't fit in that place!"))
-		return FALSE
-
 	if(user.action_busy)
 		to_chat(user, SPAN_WARNING("You can't do this right now."))
 		return FALSE
 
 	if(equipping.flags_item & WIELDED)
 		equipping.unwield(user)
+
+	if(!equipping.mob_can_equip(source, key))
+		to_chat(user, SPAN_WARNING("[equipping] doesn't fit in that place!"))
+		return FALSE
+
 	return TRUE
 
 /datum/strippable_item/mob_item_slot/start_equip(atom/source, obj/item/equipping, mob/user)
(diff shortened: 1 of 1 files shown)


code/datums/elements/strippable.dm
```diff
@@ -262,6 +268,10 @@
 		return FALSE
 
 	var/mob/sourcemob = source
+	var/obj/item/weapon/twohanded/offhand/inactive_hand = user.get_inactive_hand()
```
MistChristmas: This will pass things like wielded guns so you don't need to be so specific. You're also not accessing anything limited to twohanded/offhand.

code/datums/elements/strippable.dm
```diff
@@ -209,20 +209,26 @@
...
+	if(!human_source.has_limb_for_slot(key))
+		to_chat(user, SPAN_WARNING("[source] is missing the limb for this slot!"))
 		return FALSE
-	if (!equipping.mob_can_equip(
-		source,
-		key
-	))
-		to_chat(user, SPAN_WARNING("\The [equipping] doesn't fit in that place!"))
+
+	if(!equipping.mob_can_equip(source, key))
```
MistChristmas: `mob_can_equip` does actually try to unwield the reciever, and not the one holding the weapon for something like Guns. Which only causes issues when the giving person walks away, breaking their gun slightly and they need to unwield and swap hands to fix it. 

This also means that the `if(equipping.flags_item & WIELDED)` will never resolve to true at least for Guns and fire axe that I tested.

MistChristmas (changes requested):

code/datums/elements/strippable.dm
Nihisohel: literally just moved that check to after the WIELDED flag, and solved the issue of both, incredible

MistChristmas (approved):

PR #11352, opened 2025-12-20
Lets most Marine roles start with headsets, nukes their spare headset options in the vendors, among other changes

Nihisohel:
# About the pull request

Title. This is for Duke.

Lets majority of the marine roles, typically ones that spawn on the lower deck or are not combat exclusive to spawn starting with a headset, basically every role but Squad roles start with one now.

Really hurts the sense of direction for newer players when they lack a headset, especially when roles such as MT and nurse are one of the common choices for them. 

As for the other roles, its more of a QOL decision to assist late joiners more than anything in particular for balance.

However, majority of the spare headset choices have been nuked for redundancy, and most players, while pretty uncommon, typically give out these spare headsets rather than for personal use. I see you QM players.

This basically means that your headset is much, much more valued in gold, and losing it should mean something.

That said, while some roles such as MPs has had their vendor headset removed, they still have spares in their lockers, and the CMP are now also able to vend extras just in case, or with the command headsets through the CIC armory, same goes with the CT lockers.

Synths still retain their spare vendor headset, as they're the only one that has feasible reasons to have a spare anyhow.

Additionally, departmental headsets now default to their department frequency, less fiddling for them.

Lastly, CE tracking options has been gutted, they certainly have no need to track other squad leaders when they should be confined to the FOB, which is, for the most part, the FOB

If the CE has to build an FOB that is NOT the LZ, then command should be making an effort to provide the CE an escort, not for the CE to track down their escort.

At the very least, the CE still retains LZ access, the POs too, just in case

Works, don't really need a video for this

# Changelog

:cl:
balance: Most non-combatant marine roles start with a headset now.
balance: Rebalance the cost of the researcher spare headset in the CMO vendors
balance: CMPs are now able to buy spare MP headsets in their personal vendors
del: Most spare headset choices are removed; you may still find spares in certain places, such as the MP lockers, or CIC Armory Lockers.
code: Cleans up the paths of the headsets, made them all subtypes of command headsets, cleans up redundant variables
qol: Arguably QOL, but departmental headsets now default to their department frequency
add: Pilots can now track the LZ with their headset
del: CE tracking options has been gutted, they have no need to track anything but the LZ for building, anyhow
/:cl:

### DIFF
--- a/code/game/objects/items/devices/radio/headset.dm
+++ b/code/game/objects/items/devices/radio/headset.dm
@@ -595,7 +595,9 @@
 	locate_setting = TRACKER_LZ // for selene
 
 	inbuilt_tracking_options = list(
-		"Landing Zone" = TRACKER_LZ
+		"Landing Zone" = TRACKER_LZ,
+		"Squad Leader" = TRACKER_SL,
+		"Fireteam Leader" = TRACKER_FTL
 	)
 
 /obj/item/device/radio/headset/almayer/mcom/cmo
@@ -622,9 +624,7 @@
 	locate_setting = TRACKER_LZ // mostly just in case
 
 	inbuilt_tracking_options = list(
-		"Landing Zone" = TRACKER_LZ,
-		"Squad Leader" = TRACKER_SL,
-		"Fireteam Leader" = TRACKER_FTL
+		"Landing Zone" = TRACKER_LZ
 	)
 
 /obj/item/device/radio/headset/almayer/mcom/io


DukeofCagliostro:
Much thanks for you making this.

hry-gh (approved):
we should probably offiialise a way to get a /real/ spare from soewhere when you are missing your original headset

Nihisohel:
> we should probably offiialise a way to get a /real/ spare from soewhere when you are missing your original headset

ill think of something in another PR, probably not requisitions
i did like the idea of having department heads being able to vend spares through points, though

PR #11396, opened 2025-12-28
LV-624 Revamp - Lazarus Landing

Zenith00000:
# About the pull request

The controversial map is here, more to come in terms of info. 

Possibly the last CM project I work on.

<details>

</details>

# Changelog
:cl: Zenith, cuberound, BOBAMA, dimdimich1996, AmoryBlaine
add: Added something
code: changed some code
del: Removed old things
imageadd: added some icons and images
imagetweak: tweaked some icons and images
mapadd: added a new map or section to a map
maptweak: tweaked a map
soundadd: added a new sound thingy
/:cl:

### DIFF
--- a/code/game/objects/structures/flora.dm
+++ b/code/game/objects/structures/flora.dm
@@ -91,10 +91,10 @@ PLANT_CUT_MACHETE = 3 = Needs at least a machete to be cut down
 	qdel(src)
 
 /obj/structure/flora/ex_act(power)
-    if(explo_proof)
-        return
-    if(power >= EXPLOSION_THRESHOLD_VLOW)
-        deconstruct(FALSE)
+	if(explo_proof)
+		return
+	if(power >= EXPLOSION_THRESHOLD_VLOW)
+		deconstruct(FALSE)
 
 /obj/structure/flora/get_projectile_hit_boolean(obj/projectile/P)
 	. = ..()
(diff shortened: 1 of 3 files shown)


Drulikar:
https://discord.com/channels/150315577943130112/746325498896056329/1427775876753391646

Steelpoint:
> If you are going to continue this project, I would ask that you will have to rework the south colony to be structurally and topogohically distinct from the original LV-624 colony. The river itself and everything north of that is fine, but everything south should be unique.
>
> Thematically, envision it as a colony on the same planet of LV-624, albeit further down the same river. 
>
> Personally I would be fine with you reusing the same designs in a similar structural manner (e.g., using a similar dome structure system as it would be logical for a colony owned by the same corporation on the same planet to use a similar dome system).

Me:
> We are not allowing you to turn this into a competition. So both maps need to be able to exist along side each other. What kyo's version is like does not concern you any more. It will be evaluated for its own merits.
> 
> However I will not allow maintainers to bother spending time reviewing yours until you can comply with that requested change. Consider it DNM until then.
> ...
> If you need a visual of the parts of the colony to mix up more (you've already done some) its this. The reason why I'm putting ? on the LZs is that they can stay there but if you want to adjust them thats fine too. Corpdome as a kind of structure is fine to exist, but it needs different rooms and not to be placed the same (e.g. maybe its a west LZ building). Check marked stuff is significantly different already.
image

Drulikar:
Contributor has refused to comply to changes. PR is denied.

Issue #10867, opened 2025-10-13
Generators, SMES, Local area APC- bugged/not charging


nuklearcellphoneg:
### Testmerges

#9894: 'Multiz indicator' by Detective-Google at commit 5dbc10a0d2 #7093: 'creates ipcheck subsystem' by hry-gh at commit 83b2e4a403 #10340: 'New on_mobs for splints, bandages and blood overlays (Other changes to blood included)' by LC4492 at commit 14a671a4e4 #10830: 'cm-live' by hry-gh at commit f24923be15 #8963: 'Ports tacmap port from PVE' by Git-Nivrak at commit 167824715d #10536: 'Tailstabs cross Z & Tailstab behavior for most obj/structures' by Drulikar at commit 70af72e2ac #10845: 'Repeated Tackle Stun Resistance' by Drulikar at commit f0aa4e0542

### Round ID

30378

### Description of the bug

1. Generators fixed, fuel cells full.
2. SMES set up to charge & output
3. Local APC repaired and charged

No power would ouput to the APC, or APCs outside of the generator either. Set up a SUPERPACMAN & PACMAN inside FOB. Both successfully charged the local APC. Moved them to the Generator room- didn't power the room at all. Moved SMES from inside generator room to inside FOB- once rebuilt, successfully charged off of the PACMAN & SUPERPACMAN.

All in all, it was as if the Local Powernet for the Generator Area was completely deleted or something? Adding new charged power cells would properly power all equipment, including comms, but there was simply no way to power the APC without changing cells.

https://youtu.be/0x0x-HUsCvg?si=PDkN6zp5XKUtAUSp

Here's video of me showing up 2 hours into round and troubleshooting, covering what all I tried and others tried.

### What's the difference with what should have happened?

PACMAN/SUPERPACMAN/SMES/Generators any of these should have powered the local area APC in Generators. They did not.

### How do we reproduce this bug?

Absolutely no idea. Video included for evidence, though I got there late.

### Issue Bingo

- [x] Issue could be reproduced at least once
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] Couldn't find an existing issue about this (https://github.com/cmss13-devs/cmss13/issues)

MrSmiley3169:
may be worth noting that the issue happened again on round 30388 on LV-759.
These were the test merges at the time
[...]

PR #11440, opened 2026-01-02
Fix: APC Power not charging after reconstruction. + Remove Terminal power function [MDB Ignore]

MistChristmas:
# About the pull request
Resolves: #10867, Resolves: #11391
<details>
<summary>Old Stuff</summary>
Uh so apparently making a new APC doesn't set it's current_area which for some reason we try to assess when we should be doing the terminal.current_area
So fix that I guess.

Next issue was the power not distributing. After being repaired and the APC being charged, but only the APC.
~Well I'm pretty sure the avail() stuff shouldn't return just because the area needs power... It needs the power?
So should probably be the reverse case we toss power into the void.~
Err nevermind. So APCs didn't actually get the powernet set either so didn't put the add_avail() on.
So make it use terminal which actually gets the powernet set I suppose. This really needs to be cleaned up, ~But I'll leave the powernet stuff just in case I suppose. Eh I hate it. Surely just using the terminal which all these things should have will work fine, totally fine...~
Duh of course just override the proc
</details>
Okay so current_area doesn't get updated so we change that, and nuke all instances of .terminal being used to check for powernet stuff and instead just use the APC itself.
Such that APCs and SMES don't need their own procs for somethings like `add_load` etc.

Other things like iterating on powernet.nodes that checked for terminals now just check for apcs because all two of these were looking for the APC anyway so uh yeah.

Includes an UpdatePaths for changing terminals because they're no longer a subtype of `machinery/power` and are now just `structure`

Power chargy worky?

Did testing with fully deconning an APC and reconning it in the same area, but not the same place, 
Seems to work, Power Monitor shows all the APCs as far as I saw.
PACMAN Generator seems to correctly charge the APC in said area.
SMES seem to charge fine as well when surplus is available.
<details>

</details>

# Changelog
:cl: MistChristmas
fix: Decon-ing an APC and reconstructing it elsewhere no longer stops it from charging
code: Terminals are now somewhat fluff and not a powernet object.
/:cl:

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -364,7 +364,7 @@ GLOBAL_LIST_INIT(apc_wire_descriptions, flatten_numeric_alist(alist(
 /obj/structure/machinery/power/apc/proc/make_terminal()
 	//Create a terminal object at the same position as original turf loc
 	//Wires will attach to this
-	terminal = new/obj/structure/terminal(src.loc)
+	terminal = new/obj/structure/terminal(loc)
 	terminal.setDir(dir)
 	terminal.master = src
 


Drulikar (changes requested):
The solution shouldn't be to merely depend on the terminal's data more. If you have found a way that the APC does not perform `connect_to_network` and it normally is connected to network for maploaded apcs, then either the connected terminal should make sure the apc gets `connect_to_network` performed or you just also have the apc `connect_to_network` here https://github.com/cmss13-devs/cmss13/blob/9e46a4af63818ec5c627c77624a0a9904491f90b/code/modules/power/apc.dm#L724-L736

Drulikar:
If you did want to remove the confusion w/ terminals, they'd likely still need to exist as an `/obj/structure` just not as a `/obj/structure/machinery/power` subtype; but that would probably be a pretty large refactor.

cm13-github:

cm13-github:

code/modules/power/apc.dm
```diff
@@ -364,7 +364,7 @@ GLOBAL_LIST_INIT(apc_wire_descriptions, flatten_numeric_alist(alist(
 /obj/structure/machinery/power/apc/proc/make_terminal()
 	//Create a terminal object at the same position as original turf loc
 	//Wires will attach to this
-	terminal = new/obj/structure/machinery/power/terminal(src.loc)
+	terminal = new/obj/structure/terminal(src.loc)
```
Drulikar:
```suggestion
	terminal = new/obj/structure/terminal(loc)
```

PR #11446, opened 2026-01-03
Fixes being unable to treat severe burns without untreated damage

Puckaboo2:
# I am drafting this PR until we can identify the reason we can have severe burns with no burn damage or treated burn damage.

# About the pull request

Title is self-explanatory. If you somehow have a severe burn on a body part without any burn damage, you could not treat it unless you burn yourself in that area.

"Doc! I have a severe burn on my arm!"
"But there's no damage on your arm?"
"But it hurts! Please, do something!... What're you doing with that cautery--OW! Why'd you burn me!?"
"Hold still... burn pack here... and... done."
"Hey! The pain's gone! Thanks, doc!"

Yeah... That... Shouldn't be happening.

<details>

image

</details>

# Changelog
:cl: Puckaboo2
fix: You can now treat severe burns with ointment or a burn kit without the need for burn wounds to be present.
fix: You can now treat severe burns with a synthgraft without the need for burn wounds to be present.
/:cl:


Drulikar (changes requested):
Merely allowing severe burns to be treated in a situation they shouldn't be in in the first place is not a proper solution. Please provide what situation(s) allow severe burns or eschar to ever allow a person to be under 5 dmg.

Puckaboo2:
> Merely allowing severe burns to be treated in a situation they shouldn't be in in the first place is not a proper solution. Please provide what situation(s) allow severe burns or eschar to ever allow a person to be under 5 dmg.

I don't know how it happened because severe burns prevent chemicals from fully healing burn damage on that limb.

There were several instances where a patient waltzed into the OR, told me they had severe burns, but there wasn't any damage on that limb for me to treat, so they were stuck with severe burns... So...

Sizzle goes the cautery on harm intent, slap a burn kit, good to go.

But it shouldn't have to be that way. I wish I could discern any further knowledge, but I have none.

PR #4122, opened 2023-08-06
Attachment nerfs and removals

morrowwolf:
# About the pull request

This PR:

Removes the barrel charger from vendors

Removes all benefits other than wield delay mod from the angled grip

Adds wield delay to the extended barrel

# Explain why it's good for the game

Barrel charger is a straight damage increase and rather silly to work around given how burst works bypassing real fire rate concerns. If you know, you know. Horrible idea, I am amazed it's been around this long.

Angled grip had zero downside. Now it still has zero downside but isn't also a ton of accuracy buffs on top of the god-tier lower wield delay.

Extended barrel had zero downside. Now it has a downside.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl: Morrow
balance: Removed the barrel charger from vendors
balance: Removed all benefits other than wield delay mod from the angled grip
balance: Added wield delay to extended barrel
/:cl:

PR #11469, opened 2026-01-06
Deus Ex Chargenia: Revengeance II

Nihisohel:
# About the pull request

Readds the barrel charger to marine preps and requisitions.
You won't be able to order more though, so marine late joins are the primary way of resupplying.
Also gives it a facelift.

This is more or less a follow up to #4122, when barrel chargers were removed due to issues with burst fire rate at the time, at least according to that PR.

With the recent PRs that have practically fixed that issue, there should be no issue to return the charger back to the marine loadout, if at least to limited quantities similar to how it was back it was in.

If people are concerned about how this makes it easier to kill xenos, it doesn't- even if it takes slightly less ammunition to kill something.
Essentially, the DPS still remains the same but you have a faster TTK with initial bursts
As confusing as that statement is, you would have to use it to understand what I'm saying
That said, while it can fit on the shotgun, its practically useless on it, don't bother

If jamming & durability was still in, this would have dealt more durability damage to the gun when fired as an additionally downside, but I digress

Entertain me, please fight over this attachment in squad preps, its quite a limited commodity.

No idea why the M10 can fit the BC but whatever
image
These things are gonna look ugly on certain guns, but im not really interested in making any visual exceptions for it

Old and New sprite comparison
image

# Changelog
:cl:
balance: Readds the barrel charger to marine vendors at very limited quantities and low opportunistic restocks
imageadd: Adds a new barrel charger sprite, old one is still retained in its dmi file for legacy
/:cl:


carlarctg:
soulless sprite

Nihisohel:
dude

Git-Nivrak:
What's the worst that could happen

PR #11505, opened 2026-01-10
[SUSPENDED] Attempt to stop OBs landing in unbreakable walls.

realforest2001:
# About the pull request

In theory, this will allow the OB cannon to reroll offsets a little bit to prevent it landing inside a solid wall that it cannot break.

It sucks having the OB just vanish because of offset RNG placing it inside a cavern wall for no reason.

<details>

</details>

# Changelog
:cl:
add: OB cannon now checks if walls are unbreakable and if it cannot locate a better tile, it will fail.
/:cl:

### DIFF
--- a/code/modules/cm_marines/orbital_cannon.dm
+++ b/code/modules/cm_marines/orbital_cannon.dm
@@ -28,6 +28,9 @@ GLOBAL_LIST(ob_type_fuel_requirements)
 	COOLDOWN_DECLARE(ob_chambering_cooldown) //cooldown for chambering the gun
 	var/chamber_cooldown_time = 250 SECONDS
 
+	var/failed_launch = FALSE
+	var/failed_attempts = 0
+
 /obj/structure/orbital_cannon/New()
 	..()
 	if(!GLOB.almayer_orbital_cannon)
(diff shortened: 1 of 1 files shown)


code/modules/cm_marines/orbital_cannon.dm
```diff
@@ -205,11 +208,40 @@ GLOBAL_LIST_EMPTY(orbital_cannon_cancellation)
...
+	var/offset_x = (inaccurate_fuel + 1) * round(rand(-3,3), 1)
+	var/offset_y = (inaccurate_fuel + 1) * round(rand(-3,3), 1)
+	var/target_x = clamp(target_tile.x + offset_x, 1, world.maxx)
+	var/target_y = clamp(target_tile.y + offset_y, 1, world.maxy)
+
+	var/turf/target_turf = locate(target_x, target_y, target_tile.z)
+	if(target_turf.turf_flags & TURF_HULL)
+		failed_attempts++
+		if(failed_attempts < 3)
+			return get_offset_turf(target_tile)
```
Drulikar, line 222: Trying the same turf multiple times isn't going to do anything unless you have misfuel.

If you just want HE OBs reliable, you should have OBs either always work regardless of target (e.g. an explosion that ignores density) or expand the search out here a bit: https://github.com/cmss13-devs/cmss13/blob/b7461b73c45b80bdd44018657ce90e5a5941b242/code/modules/cm_marines/orbital_cannon.dm#L490-L492

Also keep in mind if you're doing a large search you need to use this instead:
https://github.com/cmss13-devs/cmss13/blob/b7461b73c45b80bdd44018657ce90e5a5941b242/code/__DEFINES/turfs.dm#L11-L12
orange because center was already checked.

Drulikar: Also if it's made significantly larger, you're going to need to sort the list by distance else it's going to favor the bottom left

fira:
This is already kind of implemented in HE Shells, see /obj/structure/ob_ammo/warhead/explosive/warhead_impact

PR #11529, opened 2026-01-14
Allows the ability to queue your gun wields

Nihisohel:
# About the pull request

Wanted to commit this on my other PR but I realize that this is somewhat crazy for balance, and I wasn't sure how performant it is if every gun had a unique timer each time a wield is being queued.

Unless the gun can be wielded instantly, you have to spam wield your gun in order for it to actually wield, which can be really annoying, and incentivizes holding your gun 24/7 to bypass that delay entirely

This is more or less a QOL change labelled as a balance change since this does streamline the wielding process in which you don't need to spam hit your interact button to actually wield a gun, which also boosts its responsiveness minorly

Pretty sure TGMC has this, but I couldn't be arsed to port whats essentially a 5 line change anyway
and for performance reasons as well

~~Disclaimer: The balloon alert is for demonstration purposes only~~ It exists now but only if the wield_delay wasnt instant to begin with, and the popup is shown only at higher delays
https://www.youtube.com/watch?v=WdTkTI5GwAA

# Changelog

:cl:
balance: Gun wielding can now be queued, ensuring that you do not need to spam hit the wield button in order to actually wield the thing. This does not apply to twohanded melee weapons as they don't have a delay to begin with, as of yet anyhow.
/:cl:


Venuska1117:
Sounds interesting but what if you just want to pull out without wielding it? Like want to reload empty gun? Will it auto wield and then need to press button to unwield and then reload? Also I suggest make this as preference so person can choose between automato weld or manually.

Nihisohel:
> Sounds interesting but what if you just want to pull out without wielding it? Like want to reload empty gun? Will it auto wield and then need to press button to unwield and then reload? Also I suggest make this as preference so person can choose between automato weld or manually.

you are asking a question that is out of scope of the PR
because it doesnt auto-wield the gun, it QUEUES your wield when you press the interact button

ihatethisengine:
Isn't this an important part of skill expression though, timing your wielding correctly? Delaying the use of a big gun somewhat makes you feel like you are actually operating something heavy, adding something to immersion, replacing it with a soulless timer feels wrong... It's akin to streamlining the reloading process even more, reloading your gun under pressure is the core CM experience, just like fucking up wielding your gun because you panicked...

Nihisohel:
> Isn't this an important part of skill expression though, timing your wielding correctly? 

It makes no difference to time your wield when you can just as easily spam the button until you can wield it, it makes it easier for newer players in that regard, and for everyone that isnt using an angled grip

Arguing skill expression in this regard isn't exactly valid given that if you do unwield immediately after wielding, whether by accident, or through spamming the button, you aren't exactly penalized for it since you can hit the button again and just wield normally anyway.

> Delaying the use of a big gun somewhat makes you feel like you are actually operating something heavy, adding something to immersion, replacing it with a soulless timer feels wrong... 

I genuinely do not understand what you mean by this, since this doesn't even remove the delay, and would actually convey to the user that the player is making an effort to wield it- if the wield_delay is high enough, and it would arguably make it more immersive than just having the same message as the grabbing it normally without a delay

> It's akin to streamlining the reloading process even more, reloading your gun under pressure is the core CM experience, just like fucking up wielding your gun because you panicked...

No comment about this, even if I have my own reservations about interacting with other hand but its out of scope of the PRs intent

code/modules/projectiles/gun.dm
```diff
@@ -772,12 +772,22 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
...
+	// dont want a wield when its not on the user, obviously
+	if(loc != user)
 		return
 
-	var/obj/item/I = user.get_inactive_hand()
-	if(I)
-		if(!user.drop_inv_item_on_ground(I))
+	if(world.time < pull_time) //Need to wait until it's pulled out to aim
+		if(addtimer(CALLBACK(src, PROC_REF(wield), user), pull_time - world.time, TIMER_UNIQUE))
+			if(!wield_delay <= WIELD_DELAY_VERY_FAST) // dont want the message to play when you can instantly wield it anyway
```
Contrabang: This will negate wield_delay, so its almost always zero. YOu should use the opposite of `<=` instead.
```suggestion
			if(wield_delay > WIELD_DELAY_VERY_FAST) // dont want the message to play when you can instantly wield it anyway
```

cm13-github:

cm13-github:

PR #11551, opened 2026-01-17
Some Facehugger changes, and base carrier hugger rebalancing

Nihisohel:
# About the pull request
image

Title

This PR reduces the range of facehuggers from 3 tiles to 1, but in exchange, lets them activate quicker, from a minimum of 5 seconds to a max of 15 seconds, to 3.5 seconds and 7 seconds.

Additionally, dropping facehuggers, or throwing facehuggers normally now imposes a 2.5 second cooldown on the alien to prevent spam

Lets be real, nobody likes it when a carrier stacks ten thousand huggers behind a door while you are breaking down as a marine, only to be hugged maybe almost immediately once those huggers get activated if the carrier isnt completely unlucky

While carriers are somewhat underpowered, that shouldnt change the fact that they do have round ending, if at least round ruining capabilities arguably worse than any caste in comparison, and spamming facehuggers everywhere by bypassing the throw delay that the throw hugger ability had by dropping facehuggers is not good for balance and bypasses the intent of the ability. Even if it meant they still took a long time to activate, it can still effectively change the tide of a round during a queen screech, denying a push with drop spamming in complete darkness, or guaranteeing the carriers survival during an escape (although I have no problems with the latter particularly)

With that said, this PR isn't all bad news, and also increases the viability (if ever) of facehugger throwing ability for carriers, allowing them to throw farther and faster, which should be interesting to play with the value changes on AI huggers.

Check the changelog for a complete overview.

Don't really need a video for this, but a TM would be appreciated

# Changelog
:cl:
balance: Huggers get activated sooner now, from a minimum of 5 seconds to a max of 15 seconds, to 3.5 seconds and 7 seconds. They still have an average lifespan of 30 seconds after activation, though
balance: Huggers can only hug the person next to them rather from 3 tiles away
balance: Huggers now have 5 tries to jump at someone, from only having 2 tries
balance: All xenos now have a cooldown on dropping or throwing facehuggers
balance: Carriers can now throw their stored facehuggers farther with their throw facehugger ability, from a range of 3 tiles to 6 tiles, nearly screenwide
balance: Carrier facehugger throwing is much faster, from 2 seconds to 5 deciseconds, much faster than the standard xeno cooldown
balance: Carrier facehugger retrieval from their storage now has a 5 decisecond cooldown, matching their throw cooldown
code: Facehugger throw range is no longer hardcoded to 1, but its range switches to 1 if the mob is a xenomorph. You can now throw dead facehuggers at your friends, rather than it landing pathetically next to you
code: Kills some single letter variables in facehugger and carrier.dm
refactor: Refactors carrier throw facehugger code a bit slightly
refactor: Some minor reorganization of files for cleanliness
del: Removes some defunct and unused variables
/:cl:


DangerRevolution:
> Increases the viability (if ever) of facehugger throwing ability for carriers, allowing them to throw farther and faster, which should be interesting to play with the value changes on AI huggers.

To make this true: can there be some sort of bonus for direct hitting Marines with a Hugger ala TGMC? 

Otherwise this is just a blanket nerf to any gameplay with Huggers outside of trapping, given if you hit any movement key you're now out of leap range for the Hugger despite the lower hugger timers.

cm13-github:

Solid-Fury:
A few issues with this PR.

1. The "hugger" method used is extremely hugger economy intensive and often rarely works, unless someone over extends or they synergise with a boiler for neuromas to hide the huggers (and even then it's still rather low chance especially at front due to blindfire and SGers), and could be a quick fix to have huggers die in gas
2. The "mass dropping" method mentioned really pretty useless and it used nearly exclusively as "panic dropping" (Using them to try incentivise someone not to shoot at you/let hugger take the hit) or when queen screeches (Which the activation time makes them non-viable due to marines usually shooting the hugger, one SGer for example can clear every hugger in a matter of moments.
3.  The comment on carriers being underpowered, yet have the ability, in extremely niche situations to be "round ending" with the dropped hugger method (and its low probability) can only be achieved IF the marine is overextended and assaulting, or completely alone and chasing. Two very specific scenarios. This is something that requires specific behaviour from the opponent, meaning counterplay is there. Removing this due to its very niche situation is akin to saying lurker pounce, warrior grab, prae hook and queen screech should be removed, despite this being less applicable situations overall. 
4. Removal of carrier feature/skill ceiling with no feature replacement.  Historically, carrier has been nerfed directly and indirectly, with little to no buffs. This follows the same pattern, although this does at least attempt to mitigate the impact.
5. **Potentially one of the most important** - The mitigation of the debuff may cause more frustration to the marines. I believe that the change could result in the carrier becoming a "Screech slave" and just using huggers just before or during a screech, allowing infestation of multiple marines with little to no counterplay. This could make the specific situation much worse, (and carrier less enjoyable) by making them able to capture on mass during a screech. Basically, this may remove the skill ceiling for carriers while also increasing their base strength sizeably with queen in play, similar to the old "SHAMAN" strain of Carrier.
6. There is potential for work arounds with this code, which I won't mention here incase it is merged.

Ultimately, I do think this is largely a non-issue and a reactionary response (other than the boiler cloud hiding it), predators very quickly adapted to this technique to prevent captures after a month or two of it being used, and I've personally recognised marines learning too.

De-D3:
Let's face it, this is getting nerfed because of me. You are nerfing a caste because of how one singular person plays it and maybe a couple others trying to do the same. The precision and timing it takes to properly conceal and prime huggers in the dark (although most marines can see them anyway) is overlooked. Sometimes when marines have a high amount of momentum this is the ONLY way I can manage to capture marines because they are fool-hardy enough to charge without concern. This caste has one of the best ways to punish careless fragout mentality that has been instilled and rewarded in Marine play.

What you're basically saying here is that you want marines to not have to be as concerned about their surroundings. You want to reward blind fool-hardy behavior by removing one of the ONLY ways an experienced Xeno player has to punish that type of gun-ho mentality. It's not the first time. This is a continuous cycle of xenos taking something mediocre and turning it into something great with practice, precision and in this case strategy because yes, letting myself get to low health only to have huggers hidden around a corner while a marine blindly charges after me is a risk vs reward strategy. (It's not even guaranteed the huggers wont just watch me die) You want to remove this to save the marine players feelings. You are concerned about the marine players having fun but have 0 concern whatsoever about carrier players and the sad state the caste was in before I started developing this strategy. 

This is ludicrous - outrageously and obvious bias decision making. You are gutting this caste completely to do nothing but set obvious, rinse and repeat traps in the same locations every round. This is ridiculous. But to be quite frank, I saw it coming as soon as marines refuse to adapt to something and complain about it... you just remove it. It's extremely on brand and is trivially sabotaging the state of the game one update at a time. When you say "nobody likes it" well yeah, obviously if nobody = Marines and what you're referring to "not liking" is being punished for careless behavior. Sure nobody likes it. Remove CAS because when I stand in it, I die. And nobody likes that.

You say carriers are "underpowered" but have "potential to be round changing" is just a reflection of the skill ceiling it takes to even employ this strategy. A SADAR can be round changing. A scout can be round changing. A GL can be round changing. But do you just remove the things that make them round changing? Why do xeno castes get gutted when they match that type of energy?

EDIT: And as another tip, staying stocked enough on huggers to even utilize this as effectively as the OP claims it is- is almost impossible. Running back and forth to egg morphers and even the hive to do an egg run is how a carrier who employs this type of gameplay spends half of their round and as marines have become more savvy to this playstyle, the less effective it becomes and 16 huggers is lucky to result in 2 caps. THIS is what you deem to just be groundbreaking. Marines have several buttons to push to alter the entire tide of a battle but THIS is what gets nerfed. You aren't even trying to hide it anymore, at least that in itself is somewhat admirable. We see you.

Mister-moon1:
PR feedback is around the corner son

Nihisohel:
lol

cm13-github:

Git-Nivrak (approved):

PR #11556, opened 2026-01-18
Stops hijack timer from being reset if second dropship is hijacked.

zzzmike:
# About the pull request

Stops hijack timer from being reset if second dropship is hijacked.

Clearly not intended for this to happen.

<details>

</details>

# Changelog

:cl:
fix: hijack timer will not reset if second dropship is hijacked
/:cl:

### DIFF
--- a/code/modules/shuttles/marine_ferry.dm
+++ b/code/modules/shuttles/marine_ferry.dm
@@ -531,10 +531,9 @@
 	sleep(100)
 	moving_status = SHUTTLE_CRASHED
 
-	if(SSticker.mode)
+	if(SSticker.mode && !SSticker.mode.is_in_endgame)
 		SSticker.mode.is_in_endgame = TRUE
-		if(!(SSticker.mode.force_end_at && SSticker.mode.force_end_at > 0 && SSticker.mode.force_end_at < 15000))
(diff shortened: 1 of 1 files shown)


code/controllers/subsystem/ticker.dm
```diff
@@ -518,4 +518,5 @@ SUBSYSTEM_DEF(ticker)
 /datum/controller/subsystem/ticker/proc/hijack_ocurred()
 	if(mode)
```
FourkhanDaKhan: Instead of doing the check you wrote later on, I'd just change this to be:

```suggestion
	if(mode && !mode.is_in_endgame)
```

FourkhanDaKhan (commented):
I think its simpler to just memoize the proc rather than changing it, that's what I would do.

code/modules/shuttles/marine_ferry.dm
```diff
@@ -533,7 +533,8 @@
 
 	if(SSticker.mode)
```
FourkhanDaKhan: Same here

FourkhanDaKhan (commented):
LGTM

realforest2001 (approved):

PR #11557, opened 2026-01-18
Faster cryo and automatic new aCO

JackieEstegado:
# About the pull request

Previously cryoing took 5 minutes, which was reduced to 2 minutes if you ghosted.
Now the reduction makes it 1 minute and is also applied for anyone that enters by themselves or clicked accept when someone else tried to put them in or if they are the aCO. So effectively now only players that are SSD get the full 5 minutes of safety.

Made the aCO autopromote proc trigger whenever the current aCO gets cryoed.

If a player is in the process of cryoing the autopromote will skip them.

Also ran into a null runtime error, fixed that, an empty list error (cant remember the exact name of it), as well as numerous random edge cases that got exposed.

Removed "Surgeon" from Chain of Command list as its just a variant of "Doctor", which already handles it.

Many players dont know that they need to ghost to get cryoed quickly, which can be bad when its an important role that goes to cryo.
Also the 2 minutes just feels too long still so I made it 1 minute, didnt want to make it even shorter as it could be a bit odd for RP if say a person was talking on radio 10 seconds ago, then their replacement wakes up and says hello to everyone.

Players that dont want to aCO sometimes immediately head to cryo without passing the torch to someone else, on lowpop this can cause a problem where no one has Command Console / ID Console access.

Tested in multiple runs where I joined in as roles, had autopromote trigger, cryoed as the aCO, had autopromote trigger while being in cryo and successfully skipping, joining as XO/CO randomly and cryoing. Everything worked as intended with no runtimes.

# Changelog
:cl:
qol: Cryoing now only takes 1 minute upon entering the cryobed, except for players who were forced inside while SSD unless they are the aCO.
qol: If the current aCO cryoes ARES will automatically pick a new aCO.
qol: Players in the process of cryoing will get skipped during autopromotion by ARES.
/:cl:


JackieEstegado:
Maybe we should also force the quick cryo on the current aCO too, so marines dont have to wait 5 minutes to get a new one. But I will wait for a maintainer to give a say on that one.

Edit: Harry says yes

code/game/machinery/cryopod.dm
```diff
@@ -401,7 +408,7 @@ GLOBAL_LIST_INIT(frozen_items, list(SQUAD_MARINE_1 = list(), SQUAD_MARINE_2 = li
 		if(!isliving(G.grabbed_thing))
 			return FALSE
 
-		var/willing = FALSE //We don't want to allow people to be forced into despawning.
+		willing = FALSE //We don't want to allow people to be forced into despawning.
```
FourkhanDaKhan: Pretty redundant comments through here, could probably be removed

MistChristmas (changes requested):
This doesn't account for the occurrence that is_aco gets set to TRUE whilst the time is less than 10 Minutes, I.E. It's been processed but they weren't the aCO on the first check.

I feel like you should be performing the check outside of the process.
And anyone in the cryopods just shouldn't be a candidate for aCO unless forced for some reason.

Oh and as your touching it `cryopod/attackby` one letter vars.

code/game/gamemodes/colonialmarines/colonialmarines.dm
```diff
@@ -374,39 +374,41 @@
...
+			if(job_by_chain == JOB_SO || job_by_chain == JOB_INTEL || job_by_chain == JOB_DOCTOR)
+				var/list/candidates = deep_copy_list(GLOB.marine_leaders + GLOB.marine_officers)
+				while(candidates[job_by_chain])
+					person_in_charge = pick(candidates[job_by_chain])
+					if(is_mob_cryoing(person_in_charge))
+						candidates[job_by_chain] -= person_in_charge
+						person_in_charge = null
+						//If we emptied the list then lets delete it
+						if(!length(candidates[job_by_chain]))
+							del(candidates[job_by_chain])
```
hry-gh: there's no need for this, just drop the reference to it and let the garbage collector delete it

hry-gh (dismissed):
otherwise looks fine

cm13-github:

cm13-github:

code/game/gamemodes/colonialmarines/colonialmarines.dm
JackieEstegado: Hopefully what I did is what you wanted

cm13-github:

cm13-github:

MistChristmas (approved):

PR #11562, opened 2026-01-18
Trapper window smashing fix

JackieEstegado:
# About the pull request

This is utterly horrible. Please help.

What I am trying to fix here is issue #11510
Basically right now if a trapper tries to fire an acid mine through any N number of windows it will cause all the windows to be silently destroyed, without the ability hitting or getting a cooldown.

With this fix as is right now, what will instead happen is that the acid mine will go through a singular window, if it doesnt encounter any more obstructions it will hit its target and destroy the window, if it does encounter then the ability will be completely cancelled without cooldown and the window will remain unharmed.

However the only way I could make it work in the code is horrible.

I will fix single letter vars after we get better code.

Silent bugs bad.

<details>

</details>

# Changelog
:cl:
fix: Trapper boilers can no longer mass delete windows.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities/xeno_action.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities/xeno_action.dm
@@ -391,12 +391,13 @@
 
 // Helper proc to check if there is anything blocking the way from mob M to the atom A
 // Max distance can be supplied to check some of the way instead of the whole way.
-/proc/check_clear_path_to_target(mob/M, atom/A, smash_windows = TRUE, max_distance = 1000)
+/proc/check_clear_path_to_target(mob/M, atom/A, smash_windows = TRUE, max_distance = 1000, smash_multiple = TRUE)
 	if(A.z != M.z)
 		return FALSE
 
 	var/list/turf/path = get_line(M, A, include_start_atom = FALSE)
 	var/distance = 0
+	var/obj/structure/window/framed/stored_window = null //only used with smash_multiple = FALSE, singular window that will be destroyed if there arent any more obstructions
 	for(var/turf/T in path)
 		if(distance >= max_distance)
 			return FALSE
@@ -412,11 +413,18 @@
 		for(var/obj/structure/S in T)
 			if(istype(S, /obj/structure/window/framed) && smash_windows)
 				var/obj/structure/window/framed/W = S
-				if(!W.unslashable)
-					W.deconstruct(disassembled = FALSE)
+				if(smash_multiple)
+					if(!W.unslashable && smash_multiple)
+						W.health = 0
+						W.healthcheck(TRUE, TRUE, TRUE, M)
+				else if(!stored_window)
+					stored_window = W
+				else
+					return FALSE
 
-			if(S.opacity)
-				return FALSE
+	if(stored_window)
+		stored_window.health = 0
+		stored_window.healthcheck(TRUE, TRUE, TRUE, M)
 
 	return TRUE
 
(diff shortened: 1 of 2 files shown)


FourkhanDaKhan (commented):
You could do this more simply with a list of visited windowframes.

I'll put some suggestions down below but it should overall reduce the branching

code/modules/mob/living/carbon/xenomorph/abilities/xeno_action.dm
```diff
@@ -412,11 +413,18 @@
 		for(var/obj/structure/S in T)
```
FourkhanDaKhan, line 413: Seems odd and buggy that other structures don't obscure this. Maybe just dense structures should block? Idk, there's a lot of janky things to do with structures.

code/modules/mob/living/carbon/xenomorph/abilities/xeno_action.dm
```diff
@@ -412,11 +413,18 @@
...
 			if(istype(S, /obj/structure/window/framed) && smash_windows)
 				var/obj/structure/window/framed/W = S
-				if(!W.unslashable)
-					W.deconstruct(disassembled = FALSE)
-
-			if(S.opacity)
-				return FALSE
+				if(smash_multiple)
```
FourkhanDaKhan, line 416:
```suggestion
               if (!W.unslashable)
			         window_list += window
```

FourkhanDaKhan: replace the entire flow control here

code/modules/mob/living/carbon/xenomorph/abilities/xeno_action.dm
```diff
@@ -412,11 +413,18 @@
...
+					if(!W.unslashable && smash_multiple)
+						W.health = 0
+						W.healthcheck(TRUE, TRUE, TRUE, M)
+				else if(!stored_window)
+					stored_window = W
+				else
+					return FALSE
+
+	if(stored_window)
```
FourkhanDaKhan, line 425:
```suggestion
    /proc/break_all_windows(list/obj/structure/window/windows)
       	     for (var/obj/structure/window/window in windows)
	             window.health = 0
	             window.healthcheck(TRUE, TRUE, TRUE, mob)  
	             
	/proc/break_first_window(list/obj/structure/window/windows)
	       var/obj/structure/window/first_window = windows[0]
	       first_window.health = 0
	       first_window.healthcheck(TRUE, TRUE, TRUE, mob)
	       windows -= first_window

    if (!smash_windows)
         return length(windows) == 0
         
   if (break_multiple)
         break_all_windows(windows)
         return TRUE
    else
         // !break_multiple && break_windows. Break the first window so we can make incremental progress with our spell, then return TRUE iff there was only one window to begin with.
         break_first_window(windows)
         return length(windows) == 0  
```

code/modules/mob/living/carbon/xenomorph/abilities/xeno_action.dm
```diff
@@ -391,12 +391,13 @@
 
 // Helper proc to check if there is anything blocking the way from mob M to the atom A
 // Max distance can be supplied to check some of the way instead of the whole way.
-/proc/check_clear_path_to_target(mob/M, atom/A, smash_windows = TRUE, max_distance = 1000)
+/proc/check_clear_path_to_target(mob/M, atom/A, smash_windows = TRUE, max_distance = 1000, smash_multiple = TRUE)
 	if(A.z != M.z)
 		return FALSE
```
FourkhanDaKhan, line 397:
```suggestion
	if (smash_multiple && !smash_windows)
	     CRASH("smash_multiple => smash_windows")
```

FourkhanDaKhan: One more nit; I'd probably put all the boolean arguments at the end. There's only 3 callers of this method anyway

hry-gh:
closed as stale, please recreate this pr if you want to resume working on it

PR #11574, opened 2026-01-19
Increases the limit of say/picksay/me binds to 60, gives picksay 30 options

Staykeu:
# About the pull request

As stated in title

I felt 20 was too restrictive. Harry gave the green light for 100.

<details>

</details>

# Changelog
:cl: Stakeyng Drathek
qol: added 40 more say/picksay/me bind slots (60 total)
qol: added 10 more options per picksay (30 total)
qol: limited the amount of characters per emote and picksay option to 256
fix: Fixed custom picksays not displaying the right mode when loading existing bindings
/:cl:


code/__DEFINES/keybinding.dm
```diff
@@ -4,8 +4,8 @@
 
 #define COMSIG_KB_CUSTOM_KEYBIND_DOWN "keybinding_custom_down"
 
-#define KEYBIND_CUSTOM_MAX 20
-#define KEYBIND_CUSTOM_PICKSAY_MAX 20
+#define KEYBIND_CUSTOM_MAX 100
+#define KEYBIND_CUSTOM_PICKSAY_MAX 100
```
Drulikar: Uhm you really need 100 options PER picksay too?

Also you need to edit the tsx file.

Drulikar (changes requested):

code/__DEFINES/keybinding.dm
Staykeu: > Uhm you really need 100 options PER picksay?

Do you have a preference for the amount of options per picksay? Let me know and I'll implement it.

Drulikar: I don't think you understand what these defines do is all. Changing `KEYBIND_CUSTOM_MAX` to 100 means you can have say up to 100 pick says, and EACH pick say can have up to `KEYBIND_CUSTOM_PICKSAY_MAX` contained within that single picksay.

By all means [autodoc](https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/AUTODOC.md) `///` this since it confused you though.

So my preference is the existing value.

Drulikar: To give perspective at extremes:

If you allowed 100 custom bindings, and pick says allowed 100 options each, and we kept the existing limit of 1024 characters per entry.

Then PER PLAYER they could have 100 x 100 x 1024 characters (10,240,000 characters or 40.96MB assuming Lummox has each character at a 4 byte size per character) in their custom bindings.

Staykeu: So, just to get it straight, keep the picksay options at 20, while extending the overall say/picksay/me slots to 100? This would limit the max amount of bytes per player to 2.048mb (regarding these keybind preferences specifically, on the odd chance they use every single character/slot/picksay limit slot they can)

Alternatively I can bring down the overall slots (nobody will use all 100, even in my case I only go over the half of that limit by one) to 60 and bring the picksay options up to 30 for a total of 1.8432mb.

Drulikar: So you really came thinking that "rah","raah","raaagh","raaargh","raaaaaahghh","rah","raah","raaagh","raaargh","raaaaaahghh","rah","raah","raaagh","raaargh","raaaaaahghh","rah","raah","raaagh","raaargh","raaaaaahghh" is insufficent for a single picksay?

PR #11576, opened 2026-01-20
Removal of Belt/Pouch/Armor/Uniform interaction hotkeys, restricts other hand interaction

Mister-moon1:
# About the pull request
Removal of interaction hotkeys from numerous inventory options alongside the restriction of interact other hand with ammo types, labelled as a balance change despite the original being QOL as it was improperly labelled

These were added as per QOL which I and many others would strongly disagree, put simply, it was a massive, unneeded buff that required the gutting of the MOU to compensate which can and should be reverted should this go through.

Other than that, it is abhorrent from a roleplay perspective, you are your character, if you fumble a reload, your character fumbles a reload, if you are panicking, your character is panicking, the interaction hotkeys regarding ammo in particular do not reflect this at all.

To specify more on the MOU gutting, people would make simple macros using realistically any software to have one button press interact with an ammo storage, fully reload the MOU and have it ready to shoot within a second round about, which was the same reasoning for the removal and permanent banishment of the boxer strain so I'm not sure why this hasn't got the same treatment sooner.

Also this is officially host approved

<details>

</details>

# Changelog

:cl:
balance: removal of Belt/Pouch/Armor/Uniform interaction hotkeys, restricts other hand interaction to non ammo types
/:cl:

### DIFF
--- a/code/datums/keybinding/human_inventory.dm
+++ b/code/datums/keybinding/human_inventory.dm
@@ -115,6 +115,15 @@
 	keybind_signal = COMSIG_KB_HUMAN_INTERACT_OTHER_HAND
 
 /datum/keybinding/human/inventory/interact_other_hand/down(client/user)
+	var/mob/living/carbon/human/human_mob = user.mob
+	if(!human_mob)
+		return
+
+	var/obj/item/active_item = human_mob.get_active_hand()
+	if(istype(active_item, /obj/item/ammo_magazine) || istype(active_item, /obj/item/explosive/grenade))
+		to_chat(user, SPAN_WARNING("You can't use this hotkey with ammo or explosives!"))
+		return TRUE // Stops the hotkey if there's ammo or a grenade in the active hand, accounts for shotgun shells.
+
 	. = ..()
 	if(.)
 		return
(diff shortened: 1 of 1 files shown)


Drulikar:
Title needs to be descriptive

kiVts:
Mou change is irrelevant here ( could and should be lowered though.)

Mister-moon1:
point of instant reload still stands, just applies to other weapons

Anonymosus:
: /

Viddvik:
So, we literally have all hotkeys for Xeno, we have macroses for them, so they can use entire Prae combo with one button-click, we have macroses, that allowing Xeno to perferctly attack each zone around him, so he can bypass direct-attack restrictions. BUT! The fact, that marine player can used his slots without clicking on it - was branded as a problem?

And is at least someone inside your 'which I and many others would strongly disagree', who are mainly playing marines and not Xeno? I think all changes like that shouldn't be added just because one side found it injust to play versus. 

Many Marines players can add much more PR about Xeno nerfs, than Git can hold, like 'adding GCD for every Xeno abilities', 'removing all macros and fast buttons for every xeno skills', 'removing thermal sight for the scout, as he is easily seen behind the door or wall' and much, much more, this ones just from couple of people I asked, who are playing Marines preferably. 

And If this problem, as I see, only about MOU users with fast reload, then it's part of admin problem, just add this as a rule breaking moment, that's all. It's not a situation, because of which one, should be removed ALL interactions, as I think.

Mister-moon1:
hey dude, PR feedback is around the corner

Tyranicranger4:
>  restricts other hand interaction to non ammo types

Instead of the wholesale removal of inventory interaction hotkeys, why not apply a similar restriction to something like alternative storage quick drawing? That addresses the main stated goal of this PR with far less collateral damage. After all, reload macros are the only real example where inventory interaction hotkeys meaningfully affect balance, and they're otherwise a genuinely useful QoL feature.

Mister-moon1:
Youre free to make another PR, this is just reverting a balance change marketed as a QOL change that has skewed the meta horrifically with extra handholding, and neither did i code this it is a revived PR that went stale

FeelFlows:
This (seemingly) also affects merging magazines, which can be incredibly time consuming to do when your magazine sizes are very large and handful sizes very small, bullet consolidation is not usually not done in combat either

Detective-Google:
I don't find the reasoning for this PR to be particularly convincing; Rather I believe it is, like many before it, a product of bias on the part of players who have developed clicking skills and wanting to gatekeep that advantage over others who haven't played as long. ("Well I had to do it the hard way, so why shouldn't they") 

More productive ways to accomplish this would be making the keybinding have a delay more analogous to doing it manually, retaining the control scheme ability without being a straight upgrade.

Mister-moon1:
I would recommend taking this PR up yourself, it is out of scope for this PR and I am inadequately experienced to code such, I also do not believe what you have said, I stand by my words of the fact it makes weapons such as the MOU too strong, causing it to be nerfed, takes away from the atmosphere, and one of the more interesting sides of CM which is how you interact with your menu and equipment.

Thus I will ask that until someone who is capable of completing your request shows and does said changes, this PR stays under consideration since it was a balance change disguised as a QOL.

Detective-Google:
This was discussed amongst the maintainer team, and is currently not wanted. As stated above allowing the keybind versions to be consistent, but slower (lower skill ceiling lower skill floor) would probably be acceptable instead; outright removing them is not.

PR #11627, opened 2026-01-26
Convert reagent grinder to TGUI

rattybag:
# About the pull request

Convert the reagent grinder interface to TGUI from the old browser system.

I wanted to make it so that the reagent grinder window closes when you step away from it (and also does not reopen even after you have stepped away), but it wasn't possible with the old system, so I have converted it to TGUI to achieve this.

I comprehensively tested by grinding, juicing and bottling when connected and disconnected from a smartfridge.

https://youtu.be/H5Et41APNSU

# Changelog

:cl:
refactor: refactored reagent grinder to use TGUI
ui: reagent grinder interface now uses TGUI
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/ReagentGrinder.tsx
+++ b/tgui/packages/tgui/interfaces/ReagentGrinder.tsx
@@ -39,137 +39,148 @@ export const ReagentGrinder = (props: {}) => {
 
   return (
     <Window width={450} height={500} theme="nologo">
-      <Window.Content scrollable>
-        <Stack fill vertical>
-          <Stack.Item>
-            <Section title="Processing Chamber">
-              {inuse ? (
-                <Box color="yellow">Processing... Please wait.</Box>
-              ) : hasChamberContents ? (
-                <LabeledList>
-                  {chamberContents.map((item, index) => (
-                    <LabeledList.Item
-                      key={`${item.name}-${item.amount}`}
-                      label={item.name}
-                    >
-                      x{item.amount}
-                    </LabeledList.Item>
-                  ))}
-                </LabeledList>
-              ) : (
-                <Box color="label">Chamber is empty.</Box>
-              )}
-            </Section>
-          </Stack.Item>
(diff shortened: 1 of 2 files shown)


tgui/packages/tgui/interfaces/ReagentGrinder.tsx
```diff
@@ -0,0 +1,176 @@
...
+    beakerContents,
+  } = data;
+
+  const hasChamberContents = chamberContents && chamberContents.length > 0;
+  const hasBeakerContents = beakerContents && beakerContents.length > 0;
+  const canProcess = isBeakerLoaded && hasChamberContents && !inuse;
+
+  return (
+    <Window width={450} height={500} theme="nologo">
+      <Window.Content scrollable>
```
MistChristmas: Not an issue. But as a heads up. Window.Conent scrollable is depreciated, IIRC Kashargul said something on it somewhere. Section scrollable is prefered. It'll only cause issues currently on UIs with something like a CRT Theme.

tgui/packages/tgui/interfaces/ReagentGrinder.tsx
```diff
@@ -0,0 +1,176 @@
...
+          {!inuse && (
+            <>
+              <Stack.Item>
+                <Section
+                  title="Beaker"
+                  buttons={
+                    isBeakerLoaded ? (
+                      <Button
+                        icon="eject"
+                        content="Detach"
```
MistChristmas: Button content is depreciated. Otherwise UI looks good.

rattybag:
> Functionality wise, the iteratation breaking when it finds something not allowed should probably be moved to continuing. Seeing as it doesn't stop the grindrer from processing. Especially noticable on the Juicing function.

Thanks for the review, addressed all the points you made:

- Continue instead of break when juicing/grinding non allowed items
- Refactor deprecated  Window.Content scrollable to instead have a wrapped Section with scrollable
- Refactor deprecated button content

PR #11645, opened 2026-01-31
Removes aimed shot from PMC sniper spec

ihatethisengine:
# About the pull request

Removes aimed shot from PMC sniper spec

A sniper spec with with wallhack that can two tap every xeno from complete safety is not really good for the game, even if it's hijack. For what it worth, PMC ERT will still be the strongest ERT even after this change. 

<details>

</details>

# Changelog
:cl: ihatethisengine
balance: removed aimed shot from PMC sniper spec
/:cl:

### DIFF
--- a/code/datums/emergency_calls/pmc.dm
+++ b/code/datums/emergency_calls/pmc.dm
@@ -11,7 +11,7 @@
 
 	max_smartgunners = 1
 	max_medics = 2
-	max_heavies = 0
+	max_heavies = 1
 	var/max_synths = 1
 	var/synths = 0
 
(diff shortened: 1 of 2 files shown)


Therealdoooc213:
This is a pretty big kneejerk reaction to the previous round (32214) where IIRC you died to the sniper. It definitely is too overpowered currently (Being able to stop entire xeno pushes and completely ignore defences is bad). There are alternatives other than completely removing it that I think would be better, because the M42C is a pretty unique item.

An example I could give would be reducing its damage, because wallpen ammo with thermal is definitely more than enough.

Git-Nivrak (changes requested):
Just remove the aimed shot, it was an oversight in the first place due to untying it from ghillie suit

Unknownity:
> Just remove the aimed shot, it was an oversight in the first place due to untying it from ghillie suit

I'd expand on that and suggest adding a subtype with the removed aimed shot. Whiteout Cloakers use the same unchanged sniper and removal of the ability gimps them as they don't have thermals unlike the PMC Sniper

Redbastard10101010:
In other words, you want to remove the sniper from the PMC ert, which has yet to fall out among other erts, which still has to accommodate good players who still have to play as a team, and in addition, this ert still needs to be approved by administrators who may not be online?

PR #11667, opened 2026-02-04
Fix: Bullet cannot bounce off the reflective wall back into the shooter

LiquidPotroh:
# About the pull request

When shooting at a xenomorph's reflective wall, the bullet could not be reflected back at the shooter. This PR fixes that.
Bugfix

<details>

</details>

# Changelog

:cl:

fix: fixed the bullet's inability to reflect into the shooter

/:cl:

### DIFF
--- a/code/game/turfs/walls/wall_types.dm
+++ b/code/game/turfs/walls/wall_types.dm
@@ -1346,7 +1346,7 @@
 	var/angle = Get_Angle(src, P.firer) + rand(30, -30)
 	var/atom/target = get_angle_target_turf(src, angle, get_dist(src, P.firer))
 	new_proj.projectile_flags |= PROJECTILE_SHRAPNEL
-	new_proj.fire_at(target, P.firer, src, reflect_range, speed = P.ammo.shell_speed)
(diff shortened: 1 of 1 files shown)


MistChristmas (changes requested):
This isn't actually a fix. You can still get hit by your own or anyone else's reflected bullets. 

It's just the chance is low as the accuracy is 35 and then gets further reduced by the FF_hit_evade stat because it's coming from a friendly shooter.
https://github.com/cmss13-devs/cmss13/blob/7e2da6e367acba0ffd8a74654f19f28c59e4c54c/code/modules/projectiles/projectile.dm#L883-L884

So effectively all this PR does is increase the chance you get hit by 15. 

I'd rather see you keep the shooter's source intact, and do something like add the flag AMMO_ALWAYS_FF flag to the projectile. Or a unique one so you can:

And ideally for proper logging there should be an indication that the shot was reflected, but replacing the shooter entirely with a wall isn't it. With a unique flag you can add it to the logging that it was reflected whilst keeping the other information intact.

MistChristmas:
No movement on this so closing as stale. Please re-open this PR if you intend to continue working on it.

PR #11718, opened 2026-02-15
Revert "Queens can rest again" (Ovi)

Drulikar:
# About the pull request

This PR reverts #10247 because it was not intended for queen to be able to rest, nor does queen need to benefit from nodes that require resting. I have however made it where unresting is always possible on ovi if you should somehow be resting. If someone knows how to rest while on ovi, let me know how and I will address that.

- Queen's plasma generation on ovi does not need the help of plasma trees at the ovi position. Queen only even feels plasma shortages if quickly spreading weeds, healing, or transferring plasma frequently (at least assuming recovery pheros which an ovi queen nearly always will be).
- Queen always has lost the action button to rest when going ovi (this wasn't even added in the reverted PR)
- Queen doesn't have any different resting sprites for ovi - you could only tell because most of your abilities would disable
- Should you somehow manage to be resting when ovi, it will now still allow the action to unrest you.
- Resist already has a behavior of canceling most do_afters and this includes de-ovi (as in you don't need this rest behavior to cancel de-ovi - you can just resist to cancel)

<details>

https://youtu.be/lEVYPH2a4nA

</details>

# Changelog
:cl: Drathek
del: Queen can no longer rest on ovi again
add: Added metatip to explain that resisting can interrupt most do_after actions such as de-ovi
/:cl:

### DIFF
--- a/strings/metatips.txt
+++ b/strings/metatips.txt
@@ -21,3 +21,4 @@ Control of intelligence is important: be it for retrieving it as marine, or deny
 If the lobby music is too loud or bothering you, you can disable it in Preferences tab at top right.
 Maps can and will be unpredictably modified by the Nightmare system - stay frosty while roaming around!
 You can go to GitHub to view code of the game and propose modifications of your own.
+Resisting can interrupt many different wind-up actions such as de-oving.


Red-byte3D:
This doesnt really cover for the main reason i made the PR which is the ability to cancel the de-ovi

PR #11740, opened 2026-02-18
Re-adds QFA

CapCamIII:
# Foreword
This took me like 30 minutes to do(most of it testing because there was a weird bug that took me a while to figure out), so idc if it gets closed or anything, but I think this is balanced and wouldn't be the meta attachment like old QFA was.

# About the pull request

Re-adds QFA with actual downsides, hopefully balanced hopefully.
Damage per mag is about the same but DPS is a bit higher based on my calculations.

Unlike the original which just had increased fire-rate in exchange for a poor burst and worse accuracy, this reduces damage, accuracy, burst count, and increases scatter in exchange for fire-rate.

Its kinda a reverse barrel charger, which was also recently re-added.
Speaking of: It can technically be comboed with BC but it completely nukes your accuracy and scatter in exchange for being almost the exact same as if you had neither attached.

Allows slightly better DPS in-exchange for worse ammo usage and overall accuracy(even more ammo usage)
Same rarity as the re-added BC, as it serves a similar(but also opposite) purpose.

I forgor to take photos but it attaches to the specified guns, has a working sprite, makes the gun shoot faster, and feels decent to use for all the standard marine guns it was added to

# Changelog
:cl: Riot
add: Re-adds the Quick-Fire Adapter
balance: The M41A, M4RA, 88 Mod, VP78, HPR, and some ERT weapons can fit the Quick-Fire Adapter
/:cl:


Git-Nivrak:
The qfa was just added to the stats of all weapons iirc so this is just doubling up

CapCamIII:
> The qfa was just added to the stats of all weapons iirc so this is just doubling up

bro that was like 4 or 5 years ago @Git-Nivrak

PR #11766, opened 2026-02-21
prevents larvas from getting more evo points then other castes roundstart

cuberound:
# About the pull request

there is a bug that some players abuse where they stay as larva when the queen is about to ovi late as they keep getting evo points even when queen does not enter ovi for over 3 minutes, while for otehrs it stopes at 3 minutes of oviless queen

does not reward players for bugabusing and staying larva roundstart by getting t3 slots first when queen is late to ovi

<details>

</details>

# Changelog
:cl:
fix: larvas stop getting extra evo points over their threshold as every other caste if queen does not enter ovi for over 3 minutes roundstart
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -59,7 +59,7 @@
 	if(SSxevolution)
 		progress_amount = SSxevolution.get_evolution_boost_power(hive.hivenumber)
 	var/ovipositor_check = hive.allow_no_queen_evo || hive.evolution_without_ovipositor || (hive.living_xeno_queen && hive.living_xeno_queen.ovipositor)
-	if(caste.evolution_allowed && (ovipositor_check || caste.evolve_without_queen))
+	if(caste.evolution_allowed && (ovipositor_check || (caste.evolve_without_queen && evolution_stored < evolution_threshold)))
 		if(evolution_stored >= evolution_threshold)
 			if(!got_evolution_message)
 				evolve_message()
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -59,7 +59,7 @@
 	if(SSxevolution)
 		progress_amount = SSxevolution.get_evolution_boost_power(hive.hivenumber)
 	var/ovipositor_check = hive.allow_no_queen_evo || hive.evolution_without_ovipositor || (hive.living_xeno_queen && hive.living_xeno_queen.ovipositor)
-	if(caste.evolution_allowed && (ovipositor_check || (caste.evolve_without_queen && evolution_stored < evolution_threshold)))
+	if(caste.evolution_allowed && (ovipositor_check || (caste.evolve_without_queen && evolution_stored <= evolution_threshold)))
 		if(evolution_stored >= evolution_threshold)
 			if(!got_evolution_message)
 				evolve_message()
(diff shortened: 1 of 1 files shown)


ori-disillusion-mirnov:
not a bug, as you were told multiple times

Git-Nivrak (changes requested):
I fail to see what is the bug here?
The larva still only gets bonus evo up until roundstart free evo is over so how is it any different from any other xenos?

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -59,7 +59,7 @@
 	if(SSxevolution)
 		progress_amount = SSxevolution.get_evolution_boost_power(hive.hivenumber)
 	var/ovipositor_check = hive.allow_no_queen_evo || hive.evolution_without_ovipositor || (hive.living_xeno_queen && hive.living_xeno_queen.ovipositor)
-	if(caste.evolution_allowed && (ovipositor_check || caste.evolve_without_queen))
+	if(caste.evolution_allowed && (ovipositor_check || (caste.evolve_without_queen && evolution_stored < evolution_threshold)))
```
Drulikar:
```suggestion
	if(caste.evolution_allowed && (ovipositor_check || (caste.evolve_without_queen && evolution_stored <= evolution_threshold)))
```
Needs to be <= because otherwise they won't get an evolution message. Alternatively, just make larva override this proc and call parent as necessary.

Drulikar (dismissed):

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -59,7 +59,7 @@
 	if(SSxevolution)
 		progress_amount = SSxevolution.get_evolution_boost_power(hive.hivenumber)
 	var/ovipositor_check = hive.allow_no_queen_evo || hive.evolution_without_ovipositor || (hive.living_xeno_queen && hive.living_xeno_queen.ovipositor)
-	if(caste.evolution_allowed && (ovipositor_check || caste.evolve_without_queen))
+	if(caste.evolution_allowed && (ovipositor_check || (caste.evolve_without_queen && evolution_stored <= evolution_threshold)))
```
Git-Nivrak: Wouldn't this make larvas unable to get bonus evolution at all now? Could add ovipositor check only when the evolution stored is above the threshold or override it like drathek suggested

Git-Nivrak (changes requested):

code/modules/mob/living/carbon/xenomorph/life.dm
cuberound: no it would not, if the larva has less then evo points needed to evolve it still gets points as before

Git-Nivrak (dismissed):

Drulikar:
Stale exempt until reviewed

MistChristmas:
I think doing something like below would be more readable as to what is being achieved. But otherwise LGTM.
```
	if(caste.evolution_allowed && (ovipositor_check || caste.evolve_without_queen))
		if(evolution_stored >= evolution_threshold)
			if(!got_evolution_message)
				evolve_message()
				got_evolution_message = TRUE

			if(ROUND_TIME < XENO_ROUNDSTART_BOOSTED_EVO_TIME)
				if(ovipostor_check)
					evolution_stored += progress_amount
				return
```

Nihisohel (changes requested):
do mists suggestion please

MistChristmas (approved):

Nihisohel (approved):
ye

PR #11773, opened 2026-02-22
Fix various issues with Tyrargo Rift (and Shiva's Snowball)

noelle-lavenza:
# About the pull request
Fixes some incorrect map config values for Tyrargo Rift and adds a unit test to ensure it doesn't happen for any future maps. Apparently Shiva's was also inconsistent and used the highpop map's map, but they both have the same image so the only difference was the name (Ice Colony).

~~Proposes a fix for the tacmap update issue on Tyrargo; it's hardcoded to Z2, but that's not the surface on Tyrargo like it is on every other map. ***I have not tested this yet because I'm not sure how to do so locally.***~~ this change has been removed because it works on live now

Fixes some issues.

I have not tested it yet.

### DIFF
--- a/maps/shivas_snowball.json
+++ b/maps/shivas_snowball.json
@@ -32,7 +32,7 @@
         "Magpies": "maps/map_briefings/commanding_officer/shivas/SHIVASmodi.html",
         "Unaligned": "maps/map_briefings/commanding_officer/shivas/SHIVASunaligned.html"
     },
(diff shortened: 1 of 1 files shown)


cm13-github:

cm13-github:

hry-gh (approved):
first new unit test in 5000 years

PR #11793, opened 2026-02-23
fixes dmapi cron schedule

ZephyrTFA:
LOL

image


ZephyrTFA:
Not the issue, yaml is a fake language

PR #11798, opened 2026-02-24
Experimental: Try without apt update installs from CI

Drulikar:
# About the pull request

This very likely won't work, but https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md notes all these (though maybe not specifically an i386 variant of them) except for zlib1g-dev should be present. And if we can cut back on the long durations for apt update steps in CI things should move faster. I do not currently see TG's CI do this any more.

# Changelog

No player facing changes.


Drulikar:
TG still uses ancient 516 builds so these are necessary. Azure mirrors just need to stop being crap.

PR #11843, opened 2026-02-27
Plasteel cades no longer upgradeable

Red-byte3D:
# About the pull request

Title

Plasteel cades are are already incredibly strong, they do not need the health upgrade from the upgrade.

Before anyone leaves feedback or comment ssaying "but erm, plasteel is more expensive and rare!" yes its why people use it for folding cades, its intended use, barricades are an emergency option.

Folding cades should also be removed- since they have basically made plasteel nothing but a barricade only material, but that will maybe come at a later date depending on how this goes.

Please remember that the amount of health they get is also multiplied when an upgrade hits, not only that but combined witth nailguns (sure rare) but also the fact that you can repair them even with a blowtorch from a tile away- it makes some scenarios absurd as a xeno.

<details>

</details>

# Changelog
:cl:
balance: Plasteel barricades are no longer upgradeable.
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade/non_folding.dm
+++ b/code/game/objects/structures/barricade/non_folding.dm
@@ -19,6 +19,7 @@
 	var/build_state = BARRICADE_BSTATE_SECURED //Look at __game.dm for barricade defines
 	var/upgrade = null
 	var/refund_type = /obj/item/stack/sheet/metal
+	var/upgradable = TRUE
 
 	welder_lower_damage_limit = BARRICADE_DMG_HEAVY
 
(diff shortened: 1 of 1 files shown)


antlersss:
> i also do not care what you have to say

Always the sign of a good contribution.

sippykot:
> > i also do not care what you have to say
> 
> Always the sign of a good contribution.

Following the official policy lol.

SpypigDev:
image

Comments like these are _extremely_ unproductive, if not outright incendiary

In terms of the balance implications of this PR, I dont believe justification for these changes has been adequately provided. If there is an issue of lopsided barricade strength in favor of the marines, I believe there are better ways to address them that dont simply remove a feature outright.

Feedback and teamwork are key elements of PR design, and should not be disregarded.

Closed

Drulikar:
For clarity, public feedback is not encouraged on the forums or discord. Rather the feedback form should be the avenue for feedback not related to just the implementation https://forum.cm-ss13.com/w/pr-feedback/steps/step_1

Balance wise I'm unsure about this. If you or someone wishes to still explore something like this, consider making the upgrades more specialization e.g. they get bonus to that resistance, but get weaker in other resistances.

PR #11856, opened 2026-03-01
Warrior lunge removal

Red-byte3D:
# About the pull request
Title

This pr finally touches upon a very controversial topic in cm, for years warrior has been a prevelant issue in the "balance" of the gameplay cycle, nerfs, buffs, unfair "corner lunging" "shove capping "lunge capping" i mean, it's only understandable that a core concept that relies on a click-stun ability would never be fun to "play against" people keep saying try counters like "slugs" or "don't be alone" or "maybe push the corner with a wide angle" but we all know these are "strategies" made by "people" so lets get down to the core of the issue, lunge is problematic, i mean, a 4 tile dash that stuns its target when it lands? basically impossible to counter, and lets not get started on those pesky prime warriors that keep "wiping" frontlines, a one button win ability is simply not fair.

So here's what we're going to do instead
As you can see in the video there is a warrior surrounded by 3 marines, but fret not.. (make sure the video sound is on for optimal experience)

https://github.com/user-attachments/assets/59322ce6-51a7-49aa-8d3c-797cba7ba3b7

By allowing Warriors acces to cursed energy, we are able to let them cast "Domain Expansion", what does this mean for us?

Well all players in a 5x5 range will be put in your domain, meaning that all of your attacks are a sure hit. But besides this, your enemies will also be hit with the special technique you as a warrior have imbued your Domain with, in this case ; "Limitless" while inside your domain, which means that they will be given so much information that they will not be able to process all of it , so they simply appear "frozen in time" any and all projectiles also shot into the domain will also slow down the more it "progresses into the zone"

How does this affect gameplay?

Being able to access cursed energy is a huge bonus, allowing warriors to develop more interactions with the gameplay world, 
Allowing warriors to cast their domains at will, lets marines "play around it" by simply making them a threat at "close range"
Limitless is a technique that damages the users brain, so warriors will need to play around their "damaged brain" to avoid going braindead as this caste heavily relies on the brain.

What does this mean for the future of the game?

If this test goes well, i dont see why we cant extend upon more implciations of cursed energy, maybe in the future marines can cast their own innate domain to counter the effects of limitless.

<details>

</details>

# Changelog
:cl:
balance: Warriors now have access to "Domain expansion, Infinite Void"
balance: Warriors can now see cursed energy radiating off people
code: Lays the framework for Cursed Energy"
/:cl:


Detective-Google:
hey man you're a month off

Uristthedorf:
I have reviewed your code.

Unfortunately I found some glaring issues, for one, the warriors don't take any brain damage, and there is no "sure hit" component in the code. You also forgot to remove lunge.

Therealdoooc213:
I have also reviewed your code, and it's honestly disappointing.

Adding a singular domain expansion without any framework to make the addition of others is extremely lazy. What if somebody wanted to give Carrier a domain in the future? They would have to make their own strange subtype, or refactor a root. I've also noticed that despite the CL claiming you added the framework for CE, there's nothing in the code you added for it. -1

Mister-moon1:
I would like to request that there is a 5% chance for every punch you land to be a black flash which is a concept i came up with that amplifies your attack potency by the power of 2.5 by channelling your cursed energy on the point of contact, your character should do this automatically though for simplicity sake

Next what if there was another domain like one that automatically slashes everything but in a larger radius and put it on a new strain

Finally what if marines could get some sort of simplified domain pamphlet from their prep to allow them to both see cursed energy themself and defend themself and anyone in their radius from domains, they should also get the punch ability for the chance to black flash 😁😁😁

L0pz:
sonfys

sippykot:
You forgot to add a minigame you need to play to charge up the domain expansion. Perhaps minesweeper of an encryption puzzle?

Uristthedorf:
If every xeno caste had a domain expansion what domain expansion would they have?

What about marine jobs? What does hospital corpsman get?

RainbowStalin:
> If every xeno caste had a domain expansion what domain expansion would they have?
> 
> What about marine jobs? What does hospital corpsman get?

Berserker rav would 100% have malevolent shrine.

Red-byte3D:
image

PR #11862, opened 2026-03-03
Makes the basketball hoops on LV-624 be able to be walked through

CapCamIII:
# About the pull request

makes the basketball hoops have density 0

It is so unbelievably annoying to be in a critical situation and for the hoops to block you from escaping even through there is clearly enough room on the tile to move through them.
Related: See branch name

<details>

I didn't test it, I hope it works but I made it in less than 5 minutes

</details>

# Changelog
:cl: riot
maptweak: The basketball hoops on LV-624's Fitness dome can now be walked through.
/:cl:

### DIFF
--- a/maps/map_files/LV624/LV624.dmm
+++ b/maps/map_files/LV624/LV624.dmm
@@ -4339,9 +4339,7 @@
 /turf/open/floor/warnwhite/northwest,
 /area/lv624/lazarus/fitness)
 "axu" = (
-/obj/structure/holohoop{
-	density = 0
-	},
+/obj/structure/holohoop,
 /turf/open/floor/warnwhite/north,
 /area/lv624/lazarus/fitness)
 "axv" = (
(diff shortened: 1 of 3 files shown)


maps/map_files/LV624/LV624.dmm
```diff
@@ -4339,7 +4339,9 @@
 /turf/open/floor/warnwhite/northwest,
 /area/lv624/lazarus/fitness)
 "axu" = (
-/obj/structure/holohoop,
+/obj/structure/holohoop{
```
dawsonkeyes: why not just edit the hoop obj itself for consistency so all of them behave the same

CapCamIII: I'm not sure if its intended for some of them to have density so its better to keep as is for the parent type, these are the only ones in an active combat area to my knowledge so I feel changing just these ones are fine.

Zenith00000: It should be alright to just give them all no density to be honest, I can't think of any that actually **need** to have density.

Unless anyone can think of a reason why.

Detective-Google (changes requested):
Give them the shimmy around component instead (what hybrisa streetlights use to be semi-dense.)

example;
https://github.com/cmss13-devs/cmss13/blob/master/code/game/machinery/hybrisa_lights.dm#L80-L93

PR #11882, opened 2026-03-05
Actually gives some ERTs a working skilless first-aid kit

CapCamIII:
# About the pull request

3 years ago I did this PR: #4670 to give ERTs a first aid kit that was not skill-blocked(due to the at the time recent change that made putting a first aid kit in your backpack as a non-medic impossible), little did I know but apparently this did not work at all and I just realized that today.

This fixes that by allowing it to be actually stored if you don't have the required skills.

The specific ERTs who were meant to have this(and actively still have it in the code) are:
VAIPO/VAISO, CLF, DD, freelancers, Provosts, and HGs

This was intended and balance approved but never actually worked, so this fixes a bug which is good for the game.
Also see the why its good for the original PR

I actually tested it this time and it WORKED

# Changelog
:cl: Riot
fix: Certain ERTs should now correctly spawn with basic first aid kits, as they were meant to do for the past 3 years.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/backpack.dm
+++ b/code/game/objects/items/storage/backpack.dm
@@ -21,6 +21,7 @@
 	can_hold_skill = list(
 		/obj/item/storage/firstaid = list(SKILL_MEDICAL, SKILL_MEDICAL_MEDIC),
 		/obj/item/storage/toolkit = list(SKILL_ENGINEER, SKILL_ENGINEER_TRAINED),
+		/obj/item/storage/firstaid/regular/response = list(SKILL_MEDICAL, SKILL_MEDICAL_DEFAULT),
 		)
 	drop_sound = "armorequip"
 	var/worn_accessible = FALSE //whether you can access its content while worn on the back


code/game/objects/items/storage/backpack.dm
```diff
@@ -21,6 +21,7 @@
 	can_hold_skill = list(
 		/obj/item/storage/firstaid = list(SKILL_MEDICAL, SKILL_MEDICAL_MEDIC),
 		/obj/item/storage/toolkit = list(SKILL_ENGINEER, SKILL_ENGINEER_TRAINED),
+		/obj/item/storage/firstaid/regular/response = list(SKILL_MEDICAL, SKILL_MEDICAL_DEFAULT),
```
noelle-lavenza, line 24: i'll note that this allows the ERT kit to be placed in ANY backpack by ANYONE. that's probably fine but an important distinction worth noting IMO

CapCamIII: Considering that the box ONLY spawns on ERTs and thus by the time this is accessible to get(either as a spawned ERT or someone looting one) you have a hijack to deal with, I don't think it'll be a problem(hence its approval in the PR before)

PR #11911, opened 2026-03-09
Loadout Pistol Changes

CapCamIII:
# About the pull request

Main thing this does is have the various pistol boxes now hold six magazines(the amount in a pistol belt) in addition to the pistol, instead of the 2 extra it was previous, in exchange these now cannot fit in backpacks, spawning in your hand instead.

This adds a pistol box for the L54, since despite essentially being an M4A3 reskin, it cannot fit M4a3 mags, leaving you with just the one

This adds an M1911 variant(without the damage mod) to the loadout menu, without the extra damage its basically a sidegrade/downgrade to the M4A3 stat-wise, so I don't think this is really a balance change.

Adds the NP92 as an additional option, fits alongside the Type 73 as UPP captured weapons taken as trophies

The VAIPO M1911 has the sleek silencer instead of the normal big one because I like the sprite.

To more easily do the first part's backpack change, loadout pistols are now their own subtype of boxes, with the details applied there(and changed on a case-per-case basis), additionally, some gearsets have changed their right hand to be open, allowing someone to spawn with a loadout pistol.

Also this gives 1 extra mag to both of the loadout holdout pistols, you can just drop the extra mag if you don't want to store it.
These can still fit in backpacks since they only have the pistol and 2 mags

This doesn't actually change maps I just changed a subtype to the new version with VSC.
In order:

Allows the cool special pistols you spent 4 loadout points on to be used with a pistol belt, ammo is still limited because you can't get more besides black market, but you get to play around with your cool guns more, I did the storage size changes so they aren't used as backpack pistol belts, you can just drop or store the extra mags somewhere if you don't have room

L54 is pretty cool, but only has a singular magazine that comes loaded in the gun, this gives you the extra mags that it didn't have before allowing you to use it for more than 12 shots

M1911 is peak soul, this adds the modernized version(M48A4) as it fits a tiny bit more as a modern production, its nothing out of the ordinary too considering the other loadout guns that have been added

I was asked to do this in the comments of this PR and I see no reason why not, it still fits with the other captured weapon, and is balance-wise not a concern.

This was easier to do, and makes it simpler for people to add more weapons in the future.

Just gives a bit more staying power for the pistols, so they aren't something you toss out after 2 uses, kept them with just an extra mag instead of the 6 like the others so they can still be stored in backpacks/shoulder holsters entirely, noone is putting these in a pistol belt anyway

I don't think anything in here is an actual balance concern but I tagged the magazine count changes anyway just in case.
<details>

</details>

:cl: Riot
add: Adds an M1911 variant and the NP92 as loadout point weapons.
add: The L54 now spawns with its own box with spare ammo
add: The M48A4 Tactical(VAIPO M1911) now spawns with the sleek silencer instead of the normal one
balance: Most pistol boxes from loadout menu now hold the pistol plus six magazines, but cannot fit in backpacks
/:cl:

### DIFF
--- a/code/game/objects/items/storage/misc.dm
+++ b/code/game/objects/items/storage/misc.dm
@@ -128,64 +128,99 @@
 	else
 		icon_state = "6_pack_[length(contents)]"
 
-/obj/item/storage/box/clf
-	name = "D18-storing box"
-	desc = "A fairly decorated and ceremonial box containing a CLF D18 and a single additional magazine for it. I guess those CLF folk really care about their craftsmanship and prose rather than practicality, eh?"
+/obj/item/storage/box/fluff_gun
+	name = "Sidearm Storage Box"
+	desc = "Base item for prefrence weapons."
+	w_class = SIZE_LARGE
+	max_w_class = SIZE_MEDIUM
+	storage_slots = 7
 	icon = 'icons/obj/items/storage/kits.dmi'
 	icon_state = "m43case"
-	w_class = SIZE_SMALL
+
+/obj/item/storage/box/fluff_gun/clf
+	name = "D18-storing box"
+	desc = "A fairly decorated and ceremonial box containing a CLF D18 and a few additional magazines for it. I guess those CLF folk really care about their craftsmanship and prose rather than practicality, eh?"
+	w_class = SIZE_SMALL //pocket pistols have smaller boxes because noone is going to put these in a belt
 	max_w_class = SIZE_TINY
-	storage_slots = 2
+	storage_slots = 4
 	can_hold = list(/obj/item/weapon/gun/pistol/clfpistol, /obj/item/ammo_magazine/pistol/clfpistol)
 
-/obj/item/storage/box/clf/fill_preset_inventory()
+/obj/item/storage/box/fluff_gun/clf/fill_preset_inventory()
 	new /obj/item/weapon/gun/pistol/clfpistol(src)
 	new /obj/item/ammo_magazine/pistol/clfpistol(src)
+	new /obj/item/ammo_magazine/pistol/clfpistol(src)
+	new /obj/item/ammo_magazine/pistol/clfpistol(src)
 
-/obj/item/storage/box/upp //war trophy luger
+/obj/item/storage/box/fluff_gun/upp //war trophy luger
 	name = "Type 73 storing case"
-	desc = "A small case containing the once-standard sidearm of the UPP, the Type 73, and two additional magazines. The contained sidearm is probably looted off a dead officer or from a captured stockpile, either way this thing is worth a pretty penny."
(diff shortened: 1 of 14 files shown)


sippykot:
can we get the UPP not!makarov added? right now you can only get the not!tokarev in loadout

CapCamIII:
@sippykot sure.

Steelpoint (dismissed):
Mapping LGTM

code/game/objects/items/storage/misc.dm
```diff
@@ -128,64 +128,113 @@
 	else
 		icon_state = "6_pack_[length(contents)]"
 
-/obj/item/storage/box/clf
-	name = "D18-storing box"
-	desc = "A fairly decorated and ceremonial box containing a CLF D18 and a single additional magazine for it. I guess those CLF folk really care about their craftsmanship and prose rather than practicality, eh?"
+/obj/item/storage/box/fluff_gun
+	name = "Sidearm Storage Box"
+	desc = "Base item for prefrence weapons."
```
Detective-Google:
```suggestion
	desc = "Base item for preference weapons."
```

maps/map_files/LV759_Hybrisa_Prospera/standalone/clfspaceport.dmm
```diff
@@ -764,7 +764,7 @@
 	pixel_x = -1;
 	pixel_y = 5
 	},
-/obj/item/storage/box/clf{
+/obj/item/storage/box/fluff_gun/clf{
```
Detective-Google, line 767: you need an updatepaths script if you're changing item paths like this 
see; https://github.com/cmss13-devs/cmss13/tree/master/tools/UpdatePaths

Detective-Google (changes requested):

maps/map_files/LV759_Hybrisa_Prospera/standalone/clfspaceport.dmm
CapCamIII: I ran the scripts(correctly, using the documentation) and they didn't do anything, so I'm assuming that means I'm fine.

Detective-Google: shouldnt have done anything since you already changed the one instance, this is mostly for downstreams like PVE so they can just run the script when they import this PR instead of manuually find/replacing :)

Detective-Google:
@CapCamIII that updatepaths script goes into /tools/UpdatePaths/Scripts as well please!

CapCamIII:
Oh I understand what this is now, I had no clue what you were asking me to do so I just did the first line of the documentation. I think this is correct, now. @Detective-Google

Detective-Google (approved):

PR #11914, opened 2026-03-09
Shifts Big Red LZ1 FOB Sentry

DangerRevolution:
# About the pull request

Slight Foreword: I assumed that Nightmare Inserts would override sentry landmarks but I guess not, or maybe not in this case. If there is a way to override it I'm happy to shift the insert and not the FOB landmark.

image

In the LZ1 Container Scramble insert, you can get a shit sentry that does nothing. This fixes it.

It's kinda cooked you can get a non-working sentry, especially with how reliant you can be for it during FOB.

Working sentries good, bugged sentries bad

Check MDB Bot

# Changelog

:cl: DangerRevolution
maptweak: Shifted Solaris' LZ1 Sentry Locations to prevent a Sentry spawning on a container, and therefore being bugged inside it.
/:cl:


Steelpoint (approved):
Mapping LGTM

PR #11921, opened 2026-03-10
automagically profile server init

hry-gh:
title, ci only. right now it just commits into the repo, a followup will check the repo version vs a pull request version to monitor for performance regressions


tools/ci/profile_server.sh
```diff
@@ -0,0 +1,42 @@
...
+TRACY_CAPTURE="$(_find_tool tracy-capture)"
+TRACY_CSVEXPORT="$(_find_tool tracy-csvexport)"
+DREAM_DAEMON="$(_find_tool DreamDaemon)"
+
+BYOND_TRACY_LIB="${BYOND_TRACY_LIB:-./libprof.so}"
+[[ ! -f "$BYOND_TRACY_LIB" ]] && { echo "ERROR: byond-tracy library not found at '$BYOND_TRACY_LIB'" >&2; exit 1; }
+
+(
+    "$TRACY_CAPTURE" -o "$OUTPUT_TRACY" -f -a 127.0.0.1 -p "$TRACY_PORT" 2>/dev/null || true
+    "$TRACY_CAPTURE" -o "$OUTPUT_TRACY" -f -a 127.0.0.1 -p "$TRACY_PORT" 2>/dev/null || true
```
noelle-lavenza, line 26: profile so nice we did it twice?

noelle-lavenza:
would also recommend gating the profiler run behind unit tests passing, no use wasting time profiling a failing run IMO

tools/ci/profile_server.sh
hry-gh: as discussed, intended but would like to fix pre-merge. tracy-capture just bails after the first 3 frames for some reason

hry-gh:
> would also recommend gating the profiler run behind unit tests passing, no use wasting time profiling a failing run IMO

drafted so  i do this + some things from my fork

PR #11385, opened 2025-12-26
Xenomorphs get their own per hive HUD, invisible xenos HUD icons are now visible.

Venuska1117:
If someone comes here from future and ask "how i add all huds to one xeno so i can test XvX damage?" you do it this way:
- Admin tab
- Events (category)
- Mob Event Verbs - Show
- Right click on mob to get context menu
- Add HUD to
- Choose Xeno Status HUD (gives all hives HUDs)
- Enjoy!

# Spotted issues or want to give feedback?
click this link: https://discord.com/channels/150315577943130112/1464087612796174501

# About the pull request

- Separates xeno HUDs icons per hivenumber (hive type)
- Make invisible xenos HUDs visible, burrower and lurker.

Originally i wanted invisible xenos huds to be visible when they go invisible like burrower or lurker in #7271 but because it also made other (enemy) hives see their huds, i got request to separate huds.

Expanded continuation of #7271

# Explain why it's good for the game

Everyone who at least played once and had a fight Xeno vs Xeno, knows how unbalanced it is, when xenos can see each other health, if they see someone low in health, they disengage in fight to chase that ONE xeno who is low to kill it, now it makes so you cannot easy tell if enemy xeno is actually low other than its damaged sprites, making so you should approach combat with more calculation behind risk - reward

This PR also makes interaction with xenos that are invisible way better, previously you could not see lurker, even if they were dying in critical next to you on floor... how you are supposed to know? get suddenly bodyblocked out of nowhere maybe it was lurker you didn't notice... want to cooperate with another lurker? good luck following one unless they are leader.

Now their HUD such as health, plasma etc. doesn't become transparent so you can now see and follow lurkers and any other in future transparent caste/strain.

# Testing Videos
<details>
<summary>Hud Separation (demonstration)</summary>

https://github.com/user-attachments/assets/29a26616-1124-475b-876c-ba138e7b131a

</details>

<details>
<summary>Lurker HUD Invisibility change</summary>

https://github.com/user-attachments/assets/e49c3bda-4774-49af-953e-574f6a8fab4d

</details>

<details>
<summary>Made sure it doesn't affect others (example headhunters)</summary>

https://github.com/user-attachments/assets/588e92cd-bb13-4ea0-a7dc-68ad4a6d0a59

</details>

# Changelog
:cl: Venuska1117, Harry
balance: Xenomorphs now have their own separate HUD icons per hive, if xenomorph do not belong to same hive, they cannot see other hive HUD.
[...]

PR #11925, opened 2026-03-10
Fix hivenumber not being passed to handle_xeno_hive_hud by weed food component

noelle-lavenza:
# About the pull request
`hivenumber` was passed as a bespoke argument for only xenomorph mobs (xenomorphs and the queen eye). There was no reason for this, since it always used `src.hivenumber` anyway, so now we just use that.

Should address https://runtimes.cm-ss13.com/cm13/issues/187.

I really should test this but I'm not sure how.

<details>

</details>

# Changelog
:cl: MoondancerPony
fix: fixed an error when weeds covering a xenomorph corpse die
/:cl:

### DIFF
--- a/code/datums/mob_hud.dm
+++ b/code/datums/mob_hud.dm
@@ -280,7 +280,7 @@ GLOBAL_LIST_INIT_TYPED(huds, /datum/mob_hud, flatten_numeric_alist(alist(
 /mob/proc/add_to_all_mob_huds()
 	return
 
-/mob/hologram/queen/add_to_all_mob_huds(hivenumber)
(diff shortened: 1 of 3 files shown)


Drulikar:
https://github.com/cmss13-devs/cmss13/pull/11385 is what was adding hivenumber argument - but looks to have failed to update all places calling that proc.

noelle-lavenza:
> #11385 is what was adding hivenumber argument - but looks to have failed to update all places calling that proc.

it should not have added that argument, is the point. there is no reason to pass it because it is always `src.hivenumber`.

PR #11935, opened 2026-03-12
Extinguishers can now be refilled from more water sources

halgmor:
# About the pull request

Allows extinguishers and underslung extinguishers to be refilled from water bottles, flasks, sinks, water tanks, and water coolers. 
Mixed reagents are rejected. 
Fixes an existing bug where the spray effect was invisible when firing at own or adjacent tile.

QoL update. 

Marines are grabbing 4+ portable fire extinguishers with an underslung extinguisher. Gives players a practical option to maintain their extinguishers in the field. Made at Boonie's request.

<details>

Video of refill in action: 
https://streamable.com/dezyxn

Video showing that you cannot fill it with mixed reagent or non-pure water only mixes: 
https://streamable.com/oowlat
</details>

# Changelog

:cl:
add: Fire extinguishers (portable extinguishers and underslung extinguishers) can now be refilled from water bottles, flasks, sinks, water tanks, and water coolers. 
fix: Fixed existing bug where adjacent extinguisher spray was not appearing visually
/:cl:

### DIFF
--- a/code/game/objects/items/tools/extinguisher.dm
+++ b/code/game/objects/items/tools/extinguisher.dm
@@ -118,7 +118,7 @@
 	if(is_type_in_typecache(target, refill_sources))
 		if(is_type_in_typecache(target, refill_excluded) || target == src)
 			return
-		if(!proximity_flag)
+		if(!target.Adjacent(user))
 			return
 		if(istype(target, /obj/structure/sink))
 			var/obj/structure/sink/target_sink = target
--- a/code/game/objects/structures/sink.dm
+++ b/code/game/objects/structures/sink.dm
@@ -55,7 +55,7 @@
 
 
 /obj/structure/sink/attackby(obj/item/attacking_item, mob/living/user, list/mods)
-	if(istype(attacking_item, /obj/item/tool/extinguisher))
+	if(istype(attacking_item, /obj/item/tool/extinguisher) || istype(attacking_item, /obj/item/attachable/attached_gun/extinguisher))
 		return
 
 	if(busy)
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -3666,7 +3666,7 @@ Defined in conflicts.dm of the #defines folder.
 	w_class = SIZE_MEDIUM
 	attachment_action_type = /datum/action/item_action/toggle/ext
 	slot = "under"
-	flags_attach_features = ATTACH_REMOVABLE|ATTACH_ACTIVATION|ATTACH_WEAPON|ATTACH_MELEE
+	flags_attach_features = ATTACH_REMOVABLE|ATTACH_ACTIVATION|ATTACH_RELOADABLE|ATTACH_WEAPON|ATTACH_MELEE
 	var/obj/item/tool/extinguisher/internal_extinguisher
 	current_rounds = 1 //This has to be done to pass the fire_attachment check.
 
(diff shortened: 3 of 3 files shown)


code/game/objects/items/tools/extinguisher.dm
```diff
@@ -209,6 +263,7 @@
 		if(T == target)
 			break
 		sleep(2)
+	sleep(2)
 	qdel(W)
```
Drulikar:
```suggestion
	QDEL_IN(W, 2 DECISECONDS)
```

This whole `release_liquid` proc in general though is really yucky but I don't really want to work though that right now.

code/game/objects/structures/sink.dm
```diff
@@ -55,6 +55,9 @@
 
 
 /obj/structure/sink/attackby(obj/item/O as obj, mob/user as mob)
```
Drulikar: Please refactor this to something like
`/atom/proc/attackby(obj/item/attacking_item, mob/living/user, list/mods)`

code/game/objects/items/tools/extinguisher.dm
```diff
@@ -81,15 +81,69 @@
...
+		/obj/item/reagent_container/food/drinks/cans/waterbottle,
+		/obj/item/reagent_container/food/drinks/flask,
+	))
+	var/static/list/refill_excluded = typecacheof(list(
+		/obj/structure/sink/puddle,
+		/obj/structure/reagent_dispensers/tank/water/yautja,
+		/obj/item/reagent_container/food/drinks/cans/waterbottle/upp,
+	))
+
+/obj/item/tool/extinguisher/attackby(obj/item/used_item, mob/user, params)
```
Drulikar, line 98: This proc in general would be cleaner if it utilized early returns for the first two if statements per https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#use-early-returns

code/game/objects/items/tools/extinguisher.dm
```diff
@@ -81,15 +81,69 @@
...
+				to_chat(user, SPAN_NOTICE("[src] is now refilled."))
+				playsound(user, 'sound/effects/refill.ogg', 25, TRUE, 3)
+			else if(reagents.total_volume >= max_water)
+				to_chat(user, SPAN_WARNING("[src] is already full."))
+			else
+				to_chat(user, SPAN_WARNING("[used_item] doesn't have enough water."))
+		return
+	return ..()
+
 /obj/item/tool/extinguisher/afterattack(atom/target, mob/user , flag)
```
Drulikar:
```suggestion
/obj/item/tool/extinguisher/afterattack(atom/target, mob/user, proximity_flag, click_parameters)
```

Also, this proc in general would be cleaner if it utilized early returns for the first two if statements per https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#use-early-returns

code/game/objects/structures/sink.dm
halgmor: Updated

code/game/objects/items/tools/extinguisher.dm
halgmor: Updated

code/game/objects/items/tools/extinguisher.dm
halgmor: Updated

code/game/objects/items/tools/extinguisher.dm
```diff
@@ -96,54 +96,55 @@
...
-					to_chat(user, SPAN_NOTICE("[src] is now refilled."))
-					playsound(user, 'sound/effects/refill.ogg', 25, TRUE, 3)
-				else
-					to_chat(user, SPAN_WARNING("[src] is already full."))
-				return
-			if(target.reagents.get_reagent_amount("water") != target.reagents.total_volume)
-				to_chat(user, SPAN_WARNING("You cannot refill [src] with the contents of this."))
+		if(is_type_in_typecache(target, refill_excluded) || target == src)
+			return
+		if(!proximity_flag)
```
Drulikar: If you're wanting to rely on `proxmity_flag`, you then should make sure all calls to it specify it (`/obj/item/attachable/attached_gun/extinguisher/fire_attachment` and `/obj/item/reagent_container/spray/mister/atmos/afterattack` currently fail to - though that latter one you likely won't want it to be refilled); or check it w/ null to know if its specified.

This looks to currently prevent refilling underbarrel extinguishers: 

Image

halgmor: Should work now for the underbarrel extinguishers using the same logic as the extinguishers. You need to activate the underbarrel extinguisher when refilling.

PR #11952, opened 2026-03-14
Option to make Yautja and Synth specific custom binds

BlitzArde:
# About the pull request

All changes tested. There may be a small issue with using the same keybind for different say/picksay's, though, it seems to be an issue not derived from these changes.

This PR adds 2 extra choices to the custom bind screen: Yautja and Synth. This in turn will check if your species is specifically either yautja or synth respectively before making you emote.

If you have a bind set to all (every 4 choice) then it seems to prioritize using that bind over a single-choice keybind, in the event you decide to have two say/picksay bound to the same key.

There has been a request on the forums for this and also I have been wanting it myself: Basically, "Human" is very broad to check for when using picksay, which means any non-xeno mob (Marine, Yautja, Synth, etc.) would be able to use the same picksay/say command. While this can be intentional and useful, it is also useful to have different binds for different whitelist characters to avoid accidentally pressing the wrong bind.

<details>

https://imgur.com/a/zPkm4no

</details>

# Changelog

:cl: BlitzArde
qol: Introduced options to make custom binds yautja or synth specific
/:cl:


code/datums/keybinding/custom.dm
```diff
@@ -29,6 +35,12 @@
 	if(isxeno(user?.mob))
 		return when_xeno
 
+	if(isyautja(user?.mob))
+		return when_yautja
+
+	if(issynth(user?.mob))
+		return when_synth
+
```
MistChristmas: Might as well take the user's mob once and put it in a var for these assessments.

MistChristmas (changes requested):

code/datums/keybinding/custom.dm
BlitzArde: 👍

code/datums/keybinding/custom.dm
```diff
@@ -26,9 +32,17 @@
...
+	if(isxeno(usermob))
 		return when_xeno
 
+	if(isyautja(usermob))
+		return when_yautja
+
+	if(issynth(usermob))
+		return when_synth
+
 	return when_human
```
MistChristmas:
```suggestion
	if(isspecieshuman(usermob))
		return when_human

	if(isspeciesyautja(usermob))
		return when_yautja

	if(isspeciessynth(usermob))
		return when_synth

	return when_xeno
```

Actually let's just use these rather than istyping. Double check I didn't fuck up the suggestion on GitHub.

MistChristmas (changes requested):

code/datums/keybinding/custom.dm
```diff
@@ -26,9 +32,17 @@
...
```
MistChristmas, line 49:
```suggestion
	var/usermob = user?.mob
	if(ishuman(usermob))
		var/mob/living/carbon/human/human_based_mob = usermob

		if(isspecieshuman(human_based_mob))
			return when_human

		if(isspeciesyautja(human_based_mob))
			return when_yautja

		if(isspeciessynth(human_based_mob))
			return when_synth

	return when_xeno

```
Bleh

MistChristmas (changes requested):

MistChristmas (approved):

PR #11953, opened 2026-03-14
Minor CLF buffs

CapCamIII:
# About the pull request

buffs CLF IEDs by giving them normal OT chems for their reactions instead of just water/potassium, and gives the incendiary one some ethanol so it has some decent flames, even if short-lasted.

Gives CLF combat flashlights, you can see like a tile further with these

strongly raises cyanide's oxygen damage levels, adds neurotoxic so it does brain damage, and raises its sleep chemical level.
this is technically a CLF buff because cyanide only spawns on CLF guys, poision apples, and in the liason briefcase

The grenades kinda really suck, xenos can just literally ignore them as they do that little damage, and the only reason they bother humans is the stun from being too close, this makes them actually threatening, which is nice

You can't see shit as CLF in the darkness, marine armor lights are like double your view-range, this is better

Cyanide took like a half a minute to down and a minute to kill you, now its roughly 10 seconds to down 30ish to kill.
This is mostly an admin event tool buff, as cyanide is so rare that I've seen it more often used by admins than the 3 sources possible ingame.

<details>

</details>

# Changelog
:cl: Riot
balance: Buffs the IEDs used by the CLF
balance: CLF have combat flashlights now.
balance: Cyanide kills you faster now
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/grenades/chem_grenade.dm
+++ b/code/game/objects/items/explosives/grenades/chem_grenade.dm
@@ -209,10 +209,11 @@
 	var/obj/item/reagent_container/glass/beaker/B1 = new(src)
 	var/obj/item/reagent_container/glass/beaker/B2 = new(src)
 
-	B1.reagents.add_reagent("potassium", 20)
-	B1.reagents.add_reagent("iron", 40)
-	B2.reagents.add_reagent("water", 20)
-	B2.reagents.add_reagent("iron", 40)
+	B1.reagents.add_reagent("cyclonite", 40)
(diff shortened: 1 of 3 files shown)


Uristthedorf:
Would giving them military flashlights instead of normal flashlights also be a good idea? Whenever I play as them during ERT they have such poor vision.

Forker9000:
Water + potassium nades are so bad you can't even reliably crit yourself with 'em even if you hold them while they explode it's crazy.

Detective-Google (approved):
ok

PR #11954, opened 2026-03-14
Huggers now properly leap from eggs

Red-byte3D:
# About the pull request

The pr that touched xeno huggers forgot that eggs use the same proc
This will fix it so now eggs properly throw the huggers to people that step into the range instead of just shutting off if they move one tile to the left or right.

Eggs are not supposed to be useless, im writing this up because i guess it was an "intended" change
being able to out run an egg is silly and incredibly strong for the marines
something something movielike egg gif

<details>

</details>

# Changelog
:cl:
balance: Huggers now leap up to 2 tiles from the egg it activates from instead of 1
/:cl:


Nihisohel:
balance change since i dont remember calling it a bug
image

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
```diff
@@ -255,7 +255,6 @@
 		go_idle()
 
 
-
 /obj/item/clothing/mask/facehugger/proc/leap_at_nearest_target()
```
Bibby0110: Why not just make the range a parameter for this function?
Code duplication like this is :/

Zenith00000:
I hope it isn't as broken as huggers dropped on the ground are. Where they instantly pounce with no time for you to react if they've been sitting on the ground long enough or so. Perhaps that was fixed, but I remember Carriers would purposefully drop them behind closed resin doors, once a Marine would open them they'd instantly be hugged without warning.

Taxman4889:
> I hope it isn't as broken as huggers dropped on the ground are. Where they instantly pounce with no time for you to react if they've been sitting on the ground long enough or so. Perhaps that was fixed, but I remember Carriers would purposefully drop them behind closed resin doors, once a Marine would open them they'd instantly be hugged without warning.

Huggers have a 1 tile range from the ground now per https://github.com/cmss13-devs/cmss13/pull/11551

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
Red-byte3D: the proc doesnt handle just the range, it also handles the throw speed, in case people want to switch stuff up i left it as is, unless a maint asks me to i think this is better for future editing

Bibby0110: If that became relevant they could make it a parameter too?

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
```diff
@@ -280,6 +279,31 @@
 	throw_atom(target, FACEHUGGER_JUMP_RANGE, SPEED_FAST)
 	return TRUE
 
+
+/obj/item/clothing/mask/facehugger/proc/leap_from_egg()
```
SpypigDev: rather than making a new proc, adapt the pre-existing leap_at_nearest_target proc to use range as an argument

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
```diff
@@ -280,6 +279,31 @@
...
+	if(!isturf(loc))
+		return FALSE
+
+	for(var/mob/living/human in loc)
+		if(can_hug(human, hivenumber))
+			attach(human)
+			return TRUE
+
+	var/mob/living/target
+	for(var/mob/living/human in view(EGG_JUMP_RANGE, src))
```
SpypigDev:
```suggestion
	for(var/mob/living/human in oview(EGG_JUMP_RANGE, loc))
```

SpypigDev (dismissed):
review

SpypigDev:
> Eggs are not supposed to be useless, im writing this up because i guess it was an "intended" change

For reference also, the egg range changes are not considered a bug, so the fix tag is not required

code/modules/cm_aliens/structures/egg.dm
```diff
@@ -223,7 +223,7 @@
 		return
 
 	if(instant_trigger)
-		if(!child.leap_at_nearest_target())
+		if(!child.leap_at_nearest_target(EGG_JUMP_RANGE, src))
```
Detective-Google, line 226: why are you passing src here? I don't see any code in the proc being called that uses this argument

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
```diff
@@ -266,7 +265,7 @@
 			return TRUE
 
 	var/mob/living/target
-	for(var/mob/living/human in view(FACEHUGGER_JUMP_RANGE, src))
+	for(var/mob/living/human in oview(range, loc))
```
Detective-Google, line 268: rationale for the change for src to loc here?

Detective-Google (dismissed):

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
Red-byte3D: your maintainer co-worker asked me to in the review above

Drulikar (approved):

PR #11357, opened 2025-12-21
Pouch, Autoinjector, Smart Refill Tank, Wey-Med Update PR Remake

Puckaboo2:
# About the pull request

**PART ONE: Autoinjectors**

- [x]  1. We now can choose from 5u up to 15u reagent pouch autoinjectors from Weymed Plus machines.
- [x]  2. Updates some icons; reagent pouch autoinjectors are the same color as the little overlay on the pouch.
- [x]  3. Renames autoinjectors in Medilathe.
- [x]  4. Adds the ability to flush a custom autoinjector by right-clicking it and flushing it, right-clicking a reagent canister pouch with the autoinjector inside and flushing, or right-clicking and flushing it inside a reagent canister pouch with a canister, wherein it will dump its contents and flush the rest that cannot fit.
- [x]  5. Changes pain-stop and first-aid autoinjector sprites to look more like and behave like refillable one-use autoinjectors but still keep the 'marine' autoinjector beige style.
- [x]  6. Reorganizes and renames paths of autoinjectors to better categorize them and know their code function from a glance.
- [x]  7. Applies the Dexalin+ in emergency autoinjector suggestion in #10397
- [x]  8. Restores the 4 shades of color autoinjector fills used to have.

**PART TWO: POUCHES** (Was in #11118)

- [x]  1. Oxycodone pouches carry a 5u autoinjector now.
- [x]  2. Updates some icons; the autoinjector overlay matches the color of the custom reagent autoinjectors.
- [x]  3. Fixes a misaligned reagent overlay sprite.
- [x]  4. Adds names, descriptions, icon states, overlays, and limitations to stim pouches.
- [x]  5. Adds Oxycodone Revival Mix, Peridaxon Revival MixSurgery Prep Mix, and Damage Mender Mix to vends.
- [x]  6. Removes Tricordrazine, Bicaridine, and Kelotane pouches.
- [x]  7. Removes revival mix pouches from the Researcher ColMarTech loadout.

**PART THREE: Tank and Canister Update**

- [x] 1. Fixes #10916. Smart Refill tanks can refill autoinjectors that are partially filled, provided the tank has enough of the reagent.
- [x] 2. Smart Refill tanks are even smarter now! It has 6 error dialogues for why it cannot fill an autoinjector.
- [x] 3. Smart Refill tanks will recognize most autoinjectors, including ez, one-use, meralyne, dermaline, and most any future autoinjectors that come pre-filled with medicine.
[...]

PR #11959, opened 2026-03-15
Adds flush option to Reagent Pouch Autoinjectors

halgmor:
# About the pull request
Adds a flush option to the Reagent Pouch Autoinjectors. 

You currently can flush pouch containers, but not the injectors. Meaning, after flushing your container, your injector is STILL filled with reagents that you had wanted cleared out, yet they're stuck in it with no way out other than injecting them into yourself and hope you don't OD. 

<details>

</details>

# Changelog
:cl:
qol: Reagent Pouch Autoinjectors can now be flushed.
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/autoinjectors.dm
+++ b/code/game/objects/items/reagent_containers/autoinjectors.dm
@@ -609,7 +609,7 @@
 	if(!reagents.total_volume)
 		to_chat(usr, SPAN_WARNING("[src] is already empty!"))
 		return
-	playsound(src.loc, 'sound/effects/slosh.ogg', 25, 1, 3)
+	playsound(loc, 'sound/effects/slosh.ogg', 25, 1, 3)
 	to_chat(usr, SPAN_WARNING("You work the flush valve and empty [src]'s contents!"))
 	reagents.clear_reagents()
 	uses_left = 0


Puckaboo2:
I already have this covered in #11357.

code/game/objects/items/reagent_containers/autoinjectors.dm
```diff
@@ -599,6 +599,22 @@
...
+	set category = "Object"
+	set name = "Flush Autoinjector"
+	set src in usr
+
+	if(usr.is_mob_incapacitated())
+		return
+	if(!reagents.total_volume)
+		to_chat(usr, SPAN_WARNING("[src] is already empty!"))
+		return
+	playsound(src.loc, 'sound/effects/slosh.ogg', 25, 1, 3)
```
Drulikar:
```suggestion
	playsound(loc, 'sound/effects/slosh.ogg', 25, 1, 3)
```

Drulikar (changes requested):
Also is there any reason you'd prefer this being done on `/obj/item/reagent_container/hypospray/autoinjector/empty/medic` instead of `/obj/item/reagent_container/hypospray/autoinjector/empty` like in Puck's PR? If not then the PR is probably moot.

halgmor:
> Also is there any reason you'd prefer this being done on `/obj/item/reagent_container/hypospray/autoinjector/empty/medic` instead of `/obj/item/reagent_container/hypospray/autoinjector/empty` like in Puck's PR? If not then the PR is probably moot.

Yes, several reasons:
1) I wanted to address medics' autoinjectors specifically and especially because of accidental cross-contamination during multiple autoinjector juggling in the field. 
Your half-full custom mix autoinjector was accidentally added to the second pouch that contains another custom mix. Your autoinjector is now full with both mixes together and you cannot "examine" the reagents within the autoinjector nor can you know what the marine patient has in their system while marines are dying around you and xenos are coming down on your head. You would need to flush the autoinjector and refill, but currently you can't do that. This PR gives you that option. 
Another scenario would be in the medical prep room, you fill the pouch with a mix, add the wrong autoinjector to it, now you have 15u of unknown chems that you need to offload into someone to refill the autoinjector, and then you still need to refill the pouch back to 480. 

2) I wanted a tighter, targeted, more limited scope with less chance of unintended consequences

3) A smaller change I believed would provide a more digestible PR with a quicker turnaround time and therefore faster relief for medics and their Reagent Pouch Autoinjectors than implementing a wider change

4) Puck's PR, I imagine, will take a good while before it's out. Comparably, this tiny PR will provide immediate relief for medics without having to wait for a larger PR to be tested, reworked, refined, and dragged through weeks/months of revisions and changes

Drulikar:
That doesn't really answer my question in regards to whether there's any problem with how puck has it done. But ok.

I see no point to this PR since it just knowingly conflicts with an existing PR that does the change more generically and I wasn't given a reason why their implementation doesn't just achieve the same. Merging this PR would only slow down their PR with the requested change of removing every line addition from this PR in theirs. Nor is this some critical change.

PR #11988, opened 2026-03-19
moves m85 grenade launcher fire speed in line with m92

Detective-Google:
# About the pull request

increases the fire delay of the break action grenade launcher from 12 to 32 (same as the specialist m92 grenade launcher)

players continually try and ditch their specialist launcher to use the break action one because it shoots faster (even though shooting faster is usually worse because of how stuns work). This makes it so that the break action launcher is no longer an 'upgrade' over the rotary one, because that doesn't make sense.

<details>

image

I didnt bother videoing it but I did test this 
</details>

# Changelog
:cl:
balance: the break action grenade launcher is no longer faster than the rotary one.
/:cl:


Detective-Google:
collect my webedits

PR #11992, opened 2026-03-20
Optimize flamer fire spread

noelle-lavenza:
# About the pull request
The fire spread proc now requires a reagent to be passed in, which avoids creating a new reagent for every recursive call.
`fire_spread_recur()` now uses `CHECK_TICK` instead of `spawn(0)`, so instead of delaying each wave by a single tick, it'll proceed as fast as possible without overloading the server.
`fire_spread_recur()` now only checks the directions it'll try to spread to, rather than all directions.
`fire_spread_recur()` and `fire_spread()` now use `set waitfor = FALSE` to avoid blocking callers. This is preferable to an explicit `spawn(0)` because it will only spawn if the server is overloaded, which avoids giving the scheduler a bunch of spawned procs to handle.

Split out of #11946.

Improves performance a good bit, there's no pause/delay when an incendiary OB hits.

I do have screenshots and videos, I just am too lazy to upload them right now.

<details>

</details>

# Changelog

:cl: MoondancerPony
refactor: made fire spread (from OB, flamers, etc) much faster
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/flamer/flamer.dm
+++ b/code/modules/projectiles/guns/flamer/flamer.dm
@@ -899,38 +899,30 @@ GLOBAL_LIST_EMPTY(flamer_particles)
 	else
 		weather_smothering_strength = 0
 
-/proc/fire_spread_recur(turf/target, datum/cause_data/cause_data, remaining_distance, direction, fire_lvl, burn_lvl, f_color, burn_sprite = "dynamic", aerial_flame_level)
-	var/direction_angle = dir2angle(direction)
+/proc/create_fire_reagent(fire_lvl, burn_lvl, f_color, burn_sprite = "dynamic")
+	var/datum/reagent/fire_reag = new()
+	fire_reag.intensityfire = burn_lvl
+	fire_reag.durationfire = fire_lvl
+	fire_reag.burn_sprite = burn_sprite
+	fire_reag.burncolor = f_color
+	return fire_reag
+
+/proc/fire_spread_recur(turf/target, datum/cause_data/cause_data, remaining_distance, direction, datum/reagent/fire_reagent, aerial_flame_level)
+	set waitfor = FALSE
 	var/obj/flamer_fire/foundflame = locate() in target
 	if(!foundflame)
-		var/datum/reagent/fire_reag = new()
-		fire_reag.intensityfire = burn_lvl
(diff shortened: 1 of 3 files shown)


code/modules/projectiles/guns/flamer/flamer.dm
```diff
@@ -942,37 +934,33 @@ GLOBAL_LIST_EMPTY(flamer_particles)
...
-		var/turf/picked_turf = get_step(target, direction)
 		if(aerial_flame_level)
 			if(picked_turf.get_pylon_protection_level() >= aerial_flame_level)
 				continue
 			var/area/picked_area = get_area(picked_turf)
 			if(CEILING_IS_PROTECTED(picked_area?.ceiling, get_ceiling_protection_level(aerial_flame_level)))
 				continue
-		fire_spread_recur(picked_turf, cause_data, spread_power, direction, fire_lvl, burn_lvl, f_color, burn_sprite, aerial_flame_level)
+		fire_spread_recur(picked_turf, cause_data, spread_power, direction, fire_reagent, aerial_flame_level)
+		CHECK_TICK // don't overrun spreading in just one direction
```
fira, line 963: That's a noop with waitfor = FALSE

fira (changes requested):

code/modules/projectiles/guns/flamer/flamer.dm
noelle-lavenza: > This setting will also dictate what happens if a callee sleeps. Consider an example where proc A calls proc B which calls proc C, and proc B has `waitfor` set to 0. When proc C sleeps, proc B will also sleep, but proc A will continue running as if proc B returned early. The `.` var currently in proc B will be used as its return value to A. When proc C wakes up and finishes, it will wake up proc B, but now B's return value will be ignored since A is no longer waiting for it. 

I'm not sure that's actually the case. In this case proc C is `stoplag()` and proc B is `fire_spread()`. Not blocking the caller of `fire_spread()` is fine, but I could remove `waitfor = FALSE` from the recursive proc specifically, although I'd argue the behavior it has now is desirable where one direction will sleep and it'll be allowed to spread in the other.

hry-gh (approved):

fira (approved):

PR #11993, opened 2026-03-20
fix duplicate wyc6 paygrade definition

Detective-Google:
# About the pull request

title; saw this while looking at why von bandolier shows up as WYC6 as their title

image

far as I could tell this path isnt directly referenced anywhere else that would need to be changed.

# Changelog

:cl:
fix: WYC6 Executive supervisor paygrade fixed
/:cl:


Puckaboo2:
Damn, I thought I fixed that before those changes went through? I swear I did... I know I caught it one day and I fixed it.

PR #12023, opened 2026-03-24
add confirmation for rejuvenate

ori-disillusion-mirnov:
# About the pull request
title

sometimes i press rejuvenate instead of player panel and one day it will happen when it matters and i will have to break 5 bones

<details>

</details>

# Changelog

:cl:
admin: confirmation popup for rejuvenating
/:cl:

### DIFF
--- a/code/modules/admin/verbs/mob_verbs.dm
+++ b/code/modules/admin/verbs/mob_verbs.dm
@@ -123,6 +123,9 @@
 		alert("Cannot revive a ghost")
 		return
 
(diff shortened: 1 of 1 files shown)


code/modules/admin/verbs/mob_verbs.dm
Drulikar, line 1: PlayerPanel is already a confirm button, and it utuilizes this proc. 

Image

Image

So either make the verb form the only version w/ confirmation (e.g. player panel passes a second optional argument to skip confirmation) - or just leave it as is.

ori-disillusion-mirnov:
the right click menu rejuvenate stops working when i add a second argument and i cant be bothered to figure out why so i will just be more careful

PR #12026, opened 2026-03-24
Lets hivelord early evolve into T3 xenomorphs

Ammoniacres:
# About the pull request

Title. Lets hivelord evolve from a Tier-2 support caste into any Tier-3 and Tier-2 xenomorph before marines land. This puts it in line with drones for allowing xenomorph players to contribute to the hive without dedicating an entire round to that purpose. Because of its t2 status, makes contributing as a hivelord (even temporarily) even more alluring. 

Helps encourage people to contribute to hivebuilding without feeling like they need to dedicate an entire round to the purpose of hivebuilding. Encourages more players to learn hivelord and its inns-and-outs. 

<details>

Before marines have landed.
image

After marines have landed.
image

</details>

# Changelog
:cl:
balance: Hivelords can now evolve to T3 castes, and laterally (to prevent issues) into T2 before marines drop. 
balance: Hivelords can now evolve into Queen.

/:cl:

### DIFF
--- a/code/game/objects/items/tools/extinguisher.dm
+++ b/code/game/objects/items/tools/extinguisher.dm
@@ -81,15 +81,70 @@
 	else
 		return ..()
 
-/obj/item/tool/extinguisher/afterattack(atom/target, mob/user , flag)
-	if(istype(target, /obj/structure/reagent_dispensers/tank/water) && get_dist(user,target) <= 1)
-		var/obj/object = target
-		if(object.reagents.contains_harmful_substances())
-			to_chat(user, SPAN_WARNING("You cannot re-fill the extinguisher with the contents of this."))
-			return
-		object.reagents.trans_to(src, 50)
+/obj/item/tool/extinguisher
+	var/static/list/refill_sources = typecacheof(list(
+		/obj/structure/sink,
+		/obj/structure/reagent_dispensers/tank/water,
+		/obj/structure/reagent_dispensers/water_cooler,
+		/obj/item/reagent_container/food/drinks/cans/waterbottle,
+		/obj/item/reagent_container/food/drinks/flask,
+	))
+	var/static/list/refill_excluded = typecacheof(list(
+		/obj/structure/sink/puddle,
+		/obj/structure/reagent_dispensers/tank/water/yautja,
+		/obj/item/reagent_container/food/drinks/cans/waterbottle/upp,
+	))
+
+/obj/item/tool/extinguisher/attackby(obj/item/used_item, mob/user, params)
+	if(!is_type_in_typecache(used_item, refill_sources))
+		return ..()
+	if(is_type_in_typecache(used_item, refill_excluded) || used_item == src)
+		return
+	if(used_item.reagents.get_reagent_amount("water") != used_item.reagents.total_volume)
+		to_chat(user, SPAN_WARNING("You cannot refill [src] with the contents of this."))
+		return
+	var/to_add = min(used_item.reagents.get_reagent_amount("water"), max_water - reagents.total_volume)
+	if(to_add > 0)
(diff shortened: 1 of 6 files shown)


ori-disillusion-mirnov:
this will lead to people afking as hivelord to get their t3 and getting even less done because the hivelord slot is taken up by someone doing nothing

Ammoniacres:
> this will lead to people afking as hivelord to get their t3 and getting even less done because the hivelord slot is taken up by someone doing nothing

Testmerge should prove if people exhibit this behavior or not. I'm only really worried about edge cases where *every* xeno evolves into hivelord and takes up all T2 slots before T2 castes are available. Either way I think the benefits outweigh, and I think most people are comfortable enough with evolving from T1-T2 that the idea of being stuck as a hivelord without having any intention of doing anything adjacent to building is enough disincentivisation. 

I do think this is something the other support T2s could benefit from, also - especially burrower.

dawsonkeyes:
testmerging is a good idea but i share ori's sentiment; this is probably just going to result in a lot of hivelord slot-camping and queen banishments

Bibby0110:
Just to make sure everyone is on the same page:
This is just mirroring what drones can do to T2, i.e. you can evolve from hivelord to ravager AFTER T3s are unlocked for everyone.
The only "advantage" you get here is that there's 1 free hivelord spot, so if you miss your T2 and there's no hivelords you could 'cheat' that way

Also you'll have to change things at https://github.com/cmss13-devs/cmss13/blob/7085fc2ecb710d4d9ae50d892f71a5ccbaf6f493/code/modules/mob/living/carbon/xenomorph/Evolution.dm#L32-L33

dawsonkeyes:
> Just to make sure everyone is on the same page: This is just mirroring what drones can do to T2, i.e. you can evolve from hivelord to ravager AFTER T3s are unlocked for everyone. The only "advantage" you get here is that there's 1 free hivelord spot, so if you miss your T2 and there's no hivelords you could 'cheat' that way

Another issue in my mind is that hivelord is one of the first available T2s, and this might lead to a lot of people picking hivelord to both help and to get first dibs on T3, and then when all T3 slots are filled you have a big swathe of hivelords that didn't get their T3 that have to devolve to become another T2

ihatethisengine:
> Another issue in my mind is that hivelord is one of the first available T2s, and this might lead to a lot of people picking hivelord to both help and to get first dibs on T3, and then when all T3 slots are filled you have a big swathe of hivelords that didn't get their T3 that have to devolve to become another T2

Maybe give adaptability too?

Ammoniacres:
> The only "advantage" you get here is that there's 1 free hivelord spot, so if you miss your T2 and there's no hivelords you could 'cheat' that way

This is about the only thing I can see being an issue.

Venuska1117:
you could also make so, hivelord can de-evolve to any other T1 caste before marines land (if they are unlocked), considering they can evolve to any T3, they should also regress (de-evo) to other T1 castes.

Ammoniacres:
> you could also make so, hivelord can de-evolve to any other T1 caste before marines land (if they are unlocked), considering they can evolve to any T3, they should also regress (de-evo) to other T1 castes.

I'll get an early_devolves_to markup going

Detective-Google:
We're not interested in this presently. If you want to aid building 'for free' you can play a drone and have a hivelord/queen upgrade your constructions at reduced cost. If you want to build hivelord-tier defenses, then that should come with the cost of specializing and potentially missing out on evo or being locked into playing it. I do not believe it would be a benefit for the game as a whole if the entire hive was able to spend a portion of the pre-drop as hivelords setting up defenses, as this would massively increase the quantity of defenses that exist.

To use an analogy, imagine if every marine was given an engi pamphlet and 30 metal and the size of the FOBs you would have to deal with as a result.

PR #12036, opened 2026-03-26
Scouts lose 100% IFF on cloak, but have a lower uncloak gun cooldown

private-tristan:
It would be funnier if the scout's sprite got replaced with a lurker's while cloaked, but i'm too lazy to do that. so just removing the IFF is all I'll do 
# About the pull request
see: https://forum.cm-ss13.com/t/remove-ff-protection-cloaked-scouts-get/19799/7

Scout's friendly bullet evade dropped from 1000 -> 40 (for reference, base Friendly evade is 15). this means that RARELY you'll be able to avoid bullets fired from friendlies that have NON-INSANE accuracy stats (for reference, all mk2 wielded bullets will ALWAYS hit because the mk2 has an accurate wielded range of TWENTY TILES, this will provide partial resistance to... pistols sometimes I guess)

Scout can now fire after 1 second, instead of 1.5s after uncloaking.

Demonstration of how you can, infact, shoot the scout here:

https://github.com/user-attachments/assets/fc116b8a-06ca-4a96-af3e-79107eef8886

image

# Changelog
:cl:
balance: Scouts are no longer completely IFF while cloaked
balance: Scouts can fire 0.5s (1.5s->1s) faster from uncloak
/:cl:

### DIFF
--- a/code/game/objects/items/storage/backpack.dm
+++ b/code/game/objects/items/storage/backpack.dm
@@ -1045,7 +1045,7 @@ GLOBAL_LIST_EMPTY_TYPED(radio_packs, /obj/item/storage/backpack/marine/satchel/r
 	H.unset_interaction()
 
 	H.alpha = camo_alpha
-	H.FF_hit_evade = 30
+	H.FF_hit_evade = 40
 	if(!allowed_stealth_shooting)
 		H.allow_gun_usage = allow_gun_usage
 


Uristthedorf:
If you wanted to get some guidance on making scouts look like lurkers, you could look at the cardborg suit code from tgstation, they make people wearing the suit look like cyborgs to silicons.

ihatethisengine:
let us shoot sniper too

psyendrocronologicalwarfare:
should lowkey be no delay just like the old times...

private-tristan:
> should lowkey be no delay just like the old times...

sorry I don't actualy like scout, but if you make a PR for it, I'll see if I can't change my PR so that they don't conflict

PR #12042, opened 2026-03-28
Stasis bags now accept infected corpses and tell you when they won't accept uninfected ones

Mary-Marilyn:
# About the pull request

Stasis bags will no longer yeet infected corpses out onto the floor. Previously, if a marine was infected and died, the stasis bag would detect them as dead and immediately eject them. This PR makes the bag check whether the corpse is carrying an alien embryo before deciding to reject it, if they're infected, they stay in. Uninfected corpses still get rejected, and now actually tell you why instead of silently doing nothing.

I have personally witnessed, on multiple occasions, brand new players spend upwards of three minutes repeatedly trying to bag a dead infected marine, getting increasingly confused because the bag keeps spitting the body out with zero explanation. Three minutes. Of just. Trying to put a body in a bag. These people are not stupid, the bag is just doing something completely counterintuitive with no feedback whatsoever.

<details>

</details>

# Changelog
:cl:
balance: Stasis bags no longer reject infected corpses, allowing them to be preserved for embryo removal surgery
qol: Stasis bags now display a message when refusing to store an uninfected corpse instead of silently doing nothing
/:cl:

### DIFF
--- a/code/game/objects/items/bodybag.dm
+++ b/code/game/objects/items/bodybag.dm
@@ -307,7 +307,10 @@
 	. = ..()
 	var/list/mobs_can_store = list()
 	for(var/mob/living/carbon/human/human in loc)
-		if(human.buckled || (human.stat == DEAD))
+		if(human.buckled)
+			continue
+		if(human.stat == DEAD && !(locate(/obj/item/alien_embryo) in human))
+			visible_message(SPAN_WARNING("\The [src] cannot store an uninfected corpse."))
 			continue
 		mobs_can_store += human
 	if(length(mobs_can_store))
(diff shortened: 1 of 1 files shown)


neeshacark:
as a medic main i will say you should likely make it display that the corpse is dead, the big reason this behavior exists is to prevent people permaing in stasis bags. even with a display it'll still be less readable so i would be cautious about how you do it

BarteG44:
extending the death timer would honestly be pretty on brand for a stasis bag too

zzzmike:
I don't find it unintuitive as you've stated in your PR description. It works like it does currently for a very good reason, to stop people dying and being forgotten inside a stasis bag. I'm not a fan of this, as it's gonna lead to people going perma inside bags and people will have no idea until it's too late.

zzzmike:
I'd much rather the bag just scream EMERGENCY EJECTION, DEAD BODY DETECTED or something as it spits them out. Regardless of whether they are infected or not. Also, if this PR is implemented as is, it'll let people use stasis bags as ways to detect if someone is hugged or not, which is just weird.

Detective-Google:
We're not interested in this at present, the fact that stasis bags cannot hold corpses both prevents revivable corpses from getting lost in bags, but more importantly serves as a source of decisionmaking for the medic handling a patient, and prevents the stasis bag from being an all usecases benefit without consideration. If you'd like to implement zzzmike's above suggestion to add a message saying *why* it won't accept a body when you attempt to close one, that would be a good separate PR to open.

PR #12047, opened 2026-03-29
Removes brute rockets from the requisitions vendor.

Ammoniacres:
# About the pull request

Removes BRUTE rockets from the requisitions vendors. This means the only source of BRUTE rockets is from comtech-loadout vendors.

Puts BRUTE on the same level as the other special tools comtech has available to it (emplacements) in terms of ammo economy, and reduces the sheer omnipresence of BRUTE weaponry on the battlefield.  

<details>

image

</details>

# Changelog
:cl:
balance: removes guaranteed BRUTE ammunition from the requisitions vendors.
/:cl:


kiVts:
Hi. Please get approval from drathek first, if you do get it, feel free to ask to reopen. And well, yes their omnipresence on the field is unavoidable, same with turrets.

otherwise, this leaves CTs with 3 total rockets before req needs to buy more, which is 6 for 3k$.

And no commtechs don't have brutes for points as an option. their only option is through req.

Drulikar:
I think it unlikely I would approve this; but what all equipment can be bought and in what quantity is going to change up a lot ever I get around to the req/squad vendor changes in https://github.com/cmss13-devs/cm-roadmap/issues/35

PR #12049, opened 2026-03-30
Nerfs forsaken buffs

Mister-moon1:
# About the pull request
Title

Haphazard and overtuned in nature, runners can survive PBs while dealing 32 damage slashes, vamp lurkers are quite literally mini ravs with 40 damage slashes and 35 armour etc etc, ruins any experience if you choose to go to the planet after hijack since youre destined to be round ended regardless of how well you play

I didnt test it

# Changelog

:cl:
balance: severely lowers forsaken buffs
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/hive_status.dm
+++ b/code/modules/mob/living/carbon/xenomorph/hive_status.dm
@@ -1279,22 +1279,22 @@
 
 	hive_stat_modifier_multiplier = list(
 		"damage" = XENO_HIVE_STATMOD_MULT_NONE,
-		"health" = XENO_HIVE_STATMOD_MULT_MED,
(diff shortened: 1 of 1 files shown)


tallfission:
CM's next PR: Buffs nerfed forsaken buffs

ihatethisengine:
imo just remove the buffs completely, let people roll the caste they want (like whiskey outpost) and make the CD after death really small (like 1-2 minutes)

realforest2001:
It's not supposed to be survivable regardless. ICly the planet would be swarming with xenos following evacuation, forsaken exist for the exact purpose of stopping people running to hide on the planet.
Having them buffed is irrelevant aside from making them an actual threat. They are supposed to be a threat.
Playing good might keep you alive for a little bit but you are meant to die and you will die.

PR #12055, opened 2026-03-30
Lets low threat UPP commandos be chosen as a hostile ERT.

Uristthedorf:
# About the pull request
Enables low threat UPP commandos as a hostile ERT. (Same chance to arrive as WY commandos.)

We get low threat WY commandos as an ERT, why not UPP commandos?
They're easily much weaker than WY commandos, they have no smartgunner or sniper, and can't go invisible like the high threat variant.
<details>

</details>

# Changelog
:cl:
add: Low threat UPP commandos can be chosen as hostile ERT.
/:cl:

### DIFF
--- a/code/datums/emergency_calls/upp_commando.dm
+++ b/code/datums/emergency_calls/upp_commando.dm
@@ -54,6 +54,7 @@
 /datum/emergency_call/upp_commando/low_threat
 	name = "UPP Commandos"
 	probability = 5
+	arrival_message = "[MAIN_SHIP_NAME] t*is i* UP* d^sp^*ch`. STr*&e teaM, #*u are cLe*% for a*pr*%^h. Pr*mE a*l wE*p^ns and pR*epr# t% r@nd$r a(tD."
 
 /datum/emergency_call/upp_commando/low_threat/create_member(datum/mind/mind, turf/override_spawn_loc)
 	var/turf/spawn_loc = override_spawn_loc ? override_spawn_loc : get_spawn_point()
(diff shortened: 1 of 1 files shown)


Uristthedorf:
Maybe the backstory for UPP commandos should be differentiated from the normal UPP. But I'm not a very good storyteller.

code/datums/emergency_calls/upp_commando.dm
```diff
@@ -53,6 +53,7 @@
 
 /datum/emergency_call/upp_commando/low_threat
```
Therealdoooc213, line 54: You need to set `arrival_message` for the ERT or else it'll just show up blank.

Therealdoooc213: Works now, feel free to resolve this.

Detective-Google:
We're not interested in this at the moment.

- WY commandos =/= UPP commandos; the WY commandos are specifically a tier between the weyland PMCs and the weyland whiteout, which is their deathsquad. The WY commandos come with specific downsides as opposed to their PMC counterparts.

UPP commandos are the UPP's deathsquad, and they should not be showing up normally on the Almayer-- the 'low threat' preset does exist to bring them *closer* to being balanced for normal use, but even so they are not visually distinct from the actual UPP deathsquad. (an example would be the fact the WY whiteout robots do have a low threat version, but that doesn't mean it would be acceptable for normal use.) 

stuff like the full nvgs and the rifle are not designed or tweaked to be used in normal gameplay, and shouldn't be added without a rework. You may want to revisit this idea once Blundir is able to be active again and do their UPP rework.

PR #12060, opened 2026-04-02
Boxer but not yuji

Mister-moon1:
# About the pull request
Title

Boxer is fun and this is the only times of the year it's justifiable, not the yuji variant so it can overstay its welcome a tad bit without being out of character and place (still april fools pr dnm)
I tested nothing

# Changelog
:cl:
add: Boxer added
/:cl:


dawsonkeyes:
what if we fullmerged boxer. haha

PR #12068, opened 2026-04-03
Vampire Lurker rework. 

Ammoniacres:
# About the pull request

Reworks vampire lurker in a few ways to be closer to what a typical "vampire" archetype functions like in analogous games. 
This means: 
- Vamp's healing has been increased for flurry, with it now healing 60 per person hit instead of 30.
- Headbite is now heavily incentivized, **no longer being interrupted by the victim dying mid-bite**, and now comes with a significant armor boost for the duration of the bite's doafter. 
- Tail jab is (for ease of use) a two-tile line (a la scissor cut) instead of a spriteclick/tileclick, and the stun has been increased on a successful wall-collision. [TODO] 
- Rush's distance has been nerfed from 4 tiles to 3 tiles, unless a hostile is targeted directly, in which case rush becomes 5 tiles. [TODO]

To offset this, typical to what health-stealing self-sustain brings to the archetype, **Vamp-lurker no longer gains (significant) healing from a variety of outside sources, and is heavily reliant on its own ability to secure health.** Healing pheromones are capped at weak (lesser tier), Healer drones can no longer heal Vampires, and Gardener drone fruits only restore half of their original healing rate to Vampires. 

This caste has been in a lurch for a while after its infamous usage had it kneecapped, but now doesn't have a significant place of use anywhere compared to its regular counterpart, and is hard to pilot even in the places where you would consider it specialized. A lot of the caste's problems stem from its ability to avoid significant weaknesses in the design of a self-healing, low-health ambush caste through outside healing and support, and this rework aims to focus on what makes the original design as a guerilla appealing. This should hopefully makes the strain relevant again without fear of it overcentralizing the xeno side of the game.

<details>

image

image

</details>

# Changelog

:cl: Wabberspace
balance: Vampire lurker no longer gains healing pheromones above weak, can no longer be healed by healer drones, and receives only half the total healing from healing fruits. 
balance: Vampire lurker's flurry self-heal has been increased from 30 > 60 per target hit.
balance: Vampire lurker's headbite ability increases armor from 20/20 to 40/20 for the duration of the doafter, significantly increasing its ability to ward off damage while performing a lifesteal. 
balance: Vampire lurker's tail jab is no longer a spriteclick, but is now a two-tile turf-line with the same effects. This makes it significantly easier to perform in combat.
qol: Vampire lurker's headbite ability is no longer interrupted by the victim dying during the doafter.
/:cl:


cm13-github:

cm13-github:

PR #12079, opened 2026-04-06
[EXPERIMENTAL] Increases Guaranteed Hivelord Slot from 1 to 2.

Venuska1117:
# Spotted issues or want to give feedback?
click this link: https://discord.com/channels/150315577943130112/1464087612796174501

# About the pull request

Title

I noticed specific drift in community around hivelord, people are arguing to make hivelord mandatory, having more than 1 hivelord is waste of T2 slot, there are hivelord strains but you will rarely see them in action considering there will be 1 hivelord so it doesn't take up T2 slot, it kinda makes strains inferior to base one, hivelords have the biggest impact on game but no one want to take on hivelord duty *alone* maybe they would like to work together with designer or whisperer?

Hivelord is essential for xenos but people will be discouraged to do it alone or to occupy extra T2 slot, this simple change is mainly experimental to see if hivelords pickrate would increase and if there would be more hivelords variety.

<details>
<summary>Just Image HERE</summary>

image

</details>

# Changelog
:cl: Venuska1117
balance: Hivelords Guaranteed slots are increased from 1 to 2.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/hive_status.dm
+++ b/code/modules/mob/living/carbon/xenomorph/hive_status.dm
@@ -31,7 +31,7 @@
 	/// Assoc list of free slots available to specific castes
 	var/list/free_slots = list(
 		/datum/caste_datum/burrower = 1,
-		/datum/caste_datum/hivelord = 1,
+		/datum/caste_datum/hivelord = 2,
 		/datum/caste_datum/carrier = 1
 	)
 	/// Assoc list of slots currently used by specific castes (for calculating free_slot usage)


Detective-Google:
We're not interested in this at the moment, hivelords have an extremely high game impact and should not just be a 'free' upgrade without any tradeoff. The singular free slot allows a guarantee a hive can get a hivelord if they need one, but they should not have as many hivelords as they want without cost. 

Having a hivelord is in no circumstance a 'waste of a slot'; though its playstyle is one that most players, especially xeno players, are not as interested in, hence it's unpopularity. We don't make spitters etc free because they have a lower kill count than warriors.

PR #12080, opened 2026-04-06
Changes hijack larva surge

Uristthedorf:
# About the pull request
Changes how the number of larva granted at hijack is calculated.

- Makes certain roles more impactful. (CIC/police roles weigh more than they did before for example.) (This is a xeno buff.)
- If xenos have below the desired ratio (for example if meta evac leaves a 1:4 ratio for xenos to marines) xenos get 0.8 of the difference to the desired ratio. (This is a xeno nerf, it used to be 1)

Exact numbers can be changed if needed. (For example changing the desired ratio, or whatever.)

The current system, has some flaws.

- The larva given from larva surge doesn't highly value CIC or police roles, both of which are not weaker than marines in hijack. And values specialists/smartgunners to be equally valuable as a rifleman.
- Giving xenos larva until a 1:2 ratio is reached means that there's nothing gained from killing xenos during fob siege below a 1:2 ratio since xenos are just given that larva for free. (That's why I added a 0.8 multiplier.)

This fixes those flaws. Maybe the numbers I put in are not the most balanced, but that would be easy to fix.
Maybe in the future different things could be given to xenos instead of extra larve, like extra t2/t3 slots, free evilution of ovi or boons.
It compiled, although I'm unsure how to test this as it'd need 20 players or something.
<details>

</details>

# Changelog
:cl:
balance: Reworks hijack larva surge.
/:cl:

### DIFF
--- a/code/game/jobs/role_authority.dm
+++ b/code/game/jobs/role_authority.dm
@@ -325,10 +325,18 @@ I hope it's easier to tell what the heck this proc is even doing, unlike previou
 /datum/authority/branch/role/proc/calculate_role_weight(datum/job/J)
 	if(!J)
 		return 0
+	if(GLOB.ROLES_CIC.Find(J.title))
+		return 1.5	//officer gear is stronger than marine gear.
+	if(GLOB.ROLES_POLICE.Find(J.title))
+		return 1	//I think MPs do actually have slightly better skills/gear than the standard marine but will set to 1.
+	if(J.title == JOB_SQUAD_SPECIALIST)
+		return 2
+	if(J.title == JOB_SQUAD_SMARTGUN)
+		return 1.5
 	if(GLOB.ROLES_MARINES.Find(J.title))
 		return 1
 	if(GLOB.ROLES_XENO.Find(J.title))
-		return 1
+		return 1.5	//Greenos are a huge force multiplier in hijack. Not too sure what to set this to.
 	if(J.title == JOB_SURVIVOR)
 		return 1
 	return SHIPSIDE_ROLE_WEIGHT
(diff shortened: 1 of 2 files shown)


Uristthedorf:
Oh wait, calculate_role_weight() is used for stuff other than calculating hijack surge larva too, should I just make a separate function for hijack surge larva then?

ihatethisengine:
Xenos gaining larva from surge _over time_, not in an instant. Therefore, losing a large number of xenos during a FOB siege means you won't get to the 1:2 ratio until much much later in the round, which means you won't be able to provide enough evo for those who spawn later, as you'll need to de-ovi after the crash.

code/game/jobs/role_authority.dm
```diff
@@ -325,10 +325,18 @@ I hope it's easier to tell what the heck this proc is even doing, unlike previou
...
+		return 1	//I think MPs do actually have slightly better skills/gear than the standard marine but will set to 1.
+	if(J.title == JOB_SQUAD_SPECIALIST)
+		return 2
+	if(J.title == JOB_SQUAD_SMARTGUN)
+		return 1.5
 	if(GLOB.ROLES_MARINES.Find(J.title))
 		return 1
 	if(GLOB.ROLES_XENO.Find(J.title))
-		return 1
+		return 1.5	//Greenos are a huge force multiplier in hijack. Not too sure what to set this to.
```
ihatethisengine, line 339: Larva surge only checks for humans, but this is a bad idea regardless. Greenos might be hostile, or might turn on marines later, etc. It's like scaling the surge with ERT, kinda snowflake. 

Also yes this proc is used for calculating slots roundstart, so touching it is not a good idea.

Detective-Google:
I don't think this PR is going to accomplish the desired goal and I don't think it currently justifies the added complexity to the game, since it's been open so long I'm gonna close it.

That being said, I know Drathek has some ideas for tweaking surge to encourage against metaevacs, so if you are still interested in making tweaks to the system I suggest you get in touch with them and hammer out the details on what they'd like changed.

PR #12090, opened 2026-04-08
Remove comms from most survivors.

Uristthedorf:
# About the pull request
Removes comms from most survivors.
Forecon, army, and survivor inserts keep comms, as they're supposed to be more organized I think.

Do YOU carry on your person a radio device capable of communication even after infrastructure/society has broken down?

This makes survivor gameplay more survivor horror, as you have to find friends instead of immediately getting hit with "lets hit ETA" or whatever when you spawn in.
I spawned various survivors in on a testing server and they didn't have radios.
<details>

</details>

# Changelog
:cl:
balance: Removes comms from most survivors.
/:cl:


ori-disillusion-mirnov (commented):
wrong radio those ones dont even work unless the tower is fixed, you want to remove the radio in their survival pouch

Locardoc:
Methinks they'll just rush to the metaspots as implied somewhere when this was intially suggested. This won't solve the issue, but I guess it'll make normal survivors way harder to communicate and coordinate a hold. This might majorly change survivor gameplay, as everyone will either locker hide or run around the map in hopes of not being captured.

tallfission:
> Do YOU carry on your person a radio device capable of communication even after infrastructure/society has broken down?

yes actually, and you should too, given how fucked up & on the brink the world is right now 

and on an unrelated note, i feel like spare radio communications equipment & emergency supplies would be Very Commonplace on relatively remote extrasolar colonies. places like hybrisa & tyrargo make sense to not have them *as much* because it's a city and implied to be connected to much larger infrastructure, but 624 is just 7 buildings in a jungle, kutjevo is 2 buildings in a desert, same shit different day for 522, fiorina, etc etc.

Uristthedorf:
> wrong radio those ones dont even work unless the tower is fixed, you want to remove the radio in their survival pouch
Dangit, you can tell I've never played survivor.

I made this PR because I made a jokish thread on forums and some people seemed to like it.

Detective-Google:
So, in theory what this would accomplish is decreasing the coordination of survivors, making it 'harder' and 'more immersive' due to the lack of radio callouts.

in practice the outcomes I see are:
1. new players are fucked 
		- newer players rely heavily on the coordination from the radios to figure out where to  go, what to do, and whats going on, they often do not have near encyclopedic memorization of the maps, where xenos will be going and when they evo, or how to fight them, considering the current system requires advance knowledge of item locations to thrive, this would inordinately disadvantage those players over more experienced ones who simply go to the item spawns silently anyway after a quick "hello" on the radio

2. meta calcification
		 - If survivors cannot coordinate their holding spot remotely via radio, they will be heavily incentivized to just all rush the meta spot every round, because it is where they will always know they can find other survivors, this will create a positive reinforcement loop which gradually makes the non meta spots become completely unheard of, and will lead to dealing with solaris garage and LV corpo every round for the rest of all time, until those holds get salt nerfed because "they do it every round", and then a new one is picked, repeat ad infinitum. This is undesirable as unique and creative holds are more interesting for every party involved in the surv loop than doing the same thing every time.
		 
With that in mind, I am closing this PR, I don't think it's something we want for the game, CM is a team game and I don't think kneecapping the teamwork aspect of survivor is a good idea.

PR #12119, opened 2026-04-16
Provides optional sounds for ghost ERT alerts

EmeraldCandy:
# About the pull request

Adds the alert_sound variable to emergency calls. Adding a sound here will allow response team alerts to ghosts to have a special sound, if desired. This allows for certain response teams to have unique pings to ghosts (if so desired) or allows an ert alert sound on otherwise silent pings.

This PR keeps the vast majority of response teams unchanged (since many are already accompanied by sounds ghosts can already hear). The only datum changed by this is the predator alerts, which now has the bracer alert sound. Forsaken are also silent, but due to their frequency during hijack they were left silent.

Extra options for response teams allow for certain pings to feel unique, and allows coverage to normally silent pings. Not knowing alerts are happening because there is no accompanying sound can be frustrating if you're looking for it. Many players are alt tabbed when dead and only tab back in if certain sounds alert the. This should assist that issue a bit.

Changed the youngblood ERT temporarily to clownshoes
Spawned in as a  yautja, called youngbloods with no observers. Heard no sound.
Had a second client join as a ghost and called youngbloods. Heard clownshoes.
Called many ERT, heart no additional sounds.

# Changelog

:cl:
add: Added optional customizable sound alerts for response teams
add: Predator response team calls now produce a bracer noise alert for ghosts.
/:cl:

### DIFF
--- a/code/datums/emergency_calls/emergency_call.dm
+++ b/code/datums/emergency_calls/emergency_call.dm
@@ -71,6 +71,9 @@
 	/// the [/datum/lazy_template] we should attempt to spawn in for the return journey
 	var/home_base = /datum/lazy_template/ert/freelancer_station
 
+	/// What sound plays to ghosts when this rolls? Silent if null
+	var/alert_sound
+
 /datum/game_mode/proc/initialize_emergency_calls()
 	if(length(all_calls)) //It's already been set up.
 		return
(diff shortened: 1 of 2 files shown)


EmeraldCandy:
Alternatively I can give /all/ response team pings a noise, though it may overlap sounds but make it more clear for end users. There is a very nice one paradise uses.

Detective-Google (approved):

PR #12122, opened 2026-04-17
fix: Sensor tower now reacts to explosions as it should

Rocky0558:
# About the pull request
The sensor tower now breaks from explosions as intended, rather than being magically repaired by grenades.

# Changelog

:cl:
fix: Sensor tower now damaging by explosions rather than repairing
/:cl:


MistChristmas:
The changelog is player facing, so please describe the fix in layman's terms.

Rocky0558:
> The changelog is player facing, so please describe the fix in layman's terms.

as you wish

MistChristmas (approved):

PR #12123, opened 2026-04-17
M46C Guncase + Attachment Vendor Tweaks

matttheficus:
# About the pull request

Changes the M46C selection in the CO Equipment Vendor to drop as a guncase with 3x AP mags.
Attachments for COs are now free, you can take one of each type.
Coincidentally, CBs also now get free attachments because when I looked to see how they did attachments and dropping ammo I noticed they use a define for glasses? Not a fan. So they use the same define as the CO's attachment picks.

CO Senator said to make PR. Loyal codemonkey jumps into action.

Also funny:
image

Joking aside, this should make it so COs dont feel as much a need to loot the CIC Armory. And the fact that the **infinite attachments in Req** mean somehow that COs need to spend points on attachments is goofy as hell.

<details>

</details>

# Changelog
:cl:
add: CO attachments are now free, you can select one of each slot of attachment.
add: Corporate Bodyguard can now select one of each slot of attachment.
balance: M46C now spawns in a guncase with 3x Mk1 AP mags.
/:cl:


psyendrocronologicalwarfare:
> Joking aside, this should make it so COs dont feel as much a need to loot the CIC Armory. And the fact that the infinite attachments in Req mean somehow that COs need to spend points on attachments is goofy as hell.


Isn't this intentional? The smartgun being an ammo-safe option, and the M46C being an upgraded MK1 with IFF
The M46C is already good as is, giving it more AP mags feels like an unneeded solution to an intentional problem

Instead of thinking "Hey do I go for the M46C and maybe take some mags away from my marines/officers" it'll become even a bigger no-brainer option

You should still have to interact with requisition, the reason why the RTO backpack got removed from synth is for the same reason. I don't see why giving them everything is good for the game.

matttheficus:
> Isn't this intentional? The smartgun being an ammo-safe option, and the M46C being an upgraded MK1 with IFF The M46C is already good as is, giving it more AP mags feels like an unneeded solution to an intentional problem
> 
> Instead of thinking "Hey do I go for the M46C and maybe take some mags away from my marines/officers" it'll become even a bigger no-brainer option
> 
> You should still have to interact with requisition, the reason why the RTO backpack got removed from synth is for the same reason. I don't see why giving them everything is good for the game.

Synth can literally spawn with an RTO bag. Its one of the fluff loadout options.

And frankly, if you wanna loot Req for all the gear you want, you can. Nothing is stopping you, and Req cannot do anything about it anyway.

Ditto for the mags in CIC. If you take **all** of them yeah you'll probably get smacked. But the other option is just letting them continue to yoink mags (which personally IDC if you do, but apparently some people are pissed about it so it is what it is)

psyendrocronologicalwarfare:
>And frankly, if you wanna loot Req for all the gear you want, you can. Nothing is stopping you, and Req cannot do anything about it anyway. 

Then do it. You're still interacting with the food chain.

cm13-github:

Detective-Google:
I've been going back and forth in my head on this one for a while since initially giving them the ap mags seemed like a good way to solve the goofy armory raiding, but I also agree that there's no reason the CO needs to just get 300 rounds of AP on tap-- they shouldnt really be raiding the armories either IMO but that's more of an issue with how those armories work that there's such a glut of AP ammo just sitting around to be taken by any powergamer (not just COs do this.)

Ditto w/ attachments; You can use your points if you want to bypass req, but if you want to minmax them then you'll have to go talk to them, or ignore them and vend them yourself, both of those things say something about your CO and if you want to silently gearmax then I think it's' good that there's social consequences for that.

PR #12135, opened 2026-04-20
Valkyrie buff package

TheBoondock:
# About the pull request

Does the following:
1 - Increase the healing range from 3->5 (nvm keeping it as 3)
2 - Remove the fury requirements from retrieve
3 - Add a 3 seconds grace period to xeno affected by roar ability which prevents further stuns and slow, the grace period is indicated by the green outline xeno will have

1 - 3 tiles healing range is too short to effectively heal allies in front line unless its very close quarter which people avoids clumping up due to body block and thus less people to heal. The increase in range should allow for more significant aid especially those healing or ally xeno pushing together but doesnt stay too far.
2 - The retrieve ability is normally used in split second decision and having to manage your rage for it sucks and would lead to xeno dying bc you are unable to pull them back, additionally the ability is really janky with its pathfinding and doesnt work if the target is moving so you end up just wasting your fury most of the time. The only situation where ive gotten it to reliably work is if both target and i are in the open and they enter crit. Without the fury it should be less crippling to your fury stash when this ability fails
3 - The  roar ability cancels stun only once and is not effective for chain stuning ie grenade spam, scout impact round, slugs. The ability also cannot be used when you yourself are stun thus it promotes this stay in the back and use roar when your ally is stun playstyle which i dont think is the playstyle the original author had in mind for valk. I think there are 2 solution to promote a more frontlining approach to this which is to 1 - allow roar usage while stunned 2 - add a grace period after it clears stun. I think 1 is a little inconsistent with how all xeno abilities are done while 2 allows for a little more skillful play, with good players being able to predict ahead of time and use roar. Note that this does not decrease the range so if you wanna sit back and save xeno you still can, it just rewards people who wanna push out together and support.

<details>

</details>

# Changelog

:cl:
add: Added grace period to valkyrie praetorian fight or flight ability, 3 seconds of stun immunity 
balance:  valkyrie retrieve ability no long consumes fury to activate
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/strains/castes/praetorian/valkyrie.dm
+++ b/code/modules/mob/living/carbon/xenomorph/strains/castes/praetorian/valkyrie.dm
@@ -36,7 +36,7 @@
 	var/fury_max = 200
 	var/fury_per_attack = 15
 	var/fury_per_life = 5
-	var/heal_range =  3
+	var/heal_range =  5
 	var/raging = FALSE
 
 	// State
@@ -467,9 +467,6 @@
 	if(!check_plasma_owner())
 		return
 
-	if(!behavior.use_internal_fury_ability(retrieve_cost))
-		return
-
 	if(!check_and_use_plasma_owner())
 		return
 
@@ -561,3 +558,21 @@
 	targetXeno.throw_atom(throw_target_turf, throw_dist, SPEED_VERY_FAST, pass_flags = PASS_MOB_THRU)
 	apply_cooldown()
 	return ..()
+
+/datum/status_effect/grace_period
+	status_type = STATUS_EFFECT_REFRESH
+	alert_type = null
+	remove_on_fullheal = TRUE
+	duration = 3 SECONDS
+	var/list/bit_to_remove = list(CANSTUN,CANDAZE,CANSLOW)
+
+/datum/status_effect/grace_period/on_creation(mob/living/new_owner, ...)
+	. = ..()
+	for(var/bit in bit_to_remove)
+		new_owner.status_flags &= bit
+
+/datum/status_effect/grace_period/on_remove()
+	. = ..()
+	for(var/bit in bit_to_remove)
+		owner.status_flags |= bit
+
--- a/code/modules/mob/living/carbon/xenomorph/strains/castes/praetorian/valkyrie.dm
+++ b/code/modules/mob/living/carbon/xenomorph/strains/castes/praetorian/valkyrie.dm
@@ -36,7 +36,7 @@
 	var/fury_max = 200
 	var/fury_per_attack = 15
 	var/fury_per_life = 5
-	var/heal_range =  5
+	var/heal_range =  3
 	var/raging = FALSE
 
 	// State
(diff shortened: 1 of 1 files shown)


MomoBerri:
Just echoing what i said on the discord, I think that 3 seconds is too long, but appreciate the changes as a whole.

TheBoondock:
I don't entirely disagree however, if this is ok I would to see it TM to gauge its effective and fine tune the time.

KeRSedChaplain:
I couldn't find the discussion thread on the discord but I'm putting up the counterpoint that the tantrum ability _also_ only lasts 3 seconds.

Fight or flight seems perfectly fine to be given the 3 second grace period, because In practice it only gives you one extra ignored stun, and no damage mitigation unless a Valkyrie spends all of their fury trying to save a singular xeno from a wave of explosives.

Venuska1117:
3 seconds grace period would be fine, but i would suggest making so, during that 3 second grace period, if you get any negative effect, lets say, stunned by grenade/PB/rocket, then your grace period ends on that 1 instance of "debuff", so it doesn't make you temporary immune to this sources, but give you "one time" saving power.

(you need to also think about marines side on that one, even if you are not marine main)

Red-byte3D:
I made it 3 tiles for good reason, 5 tiles is unreasonably wide, especially when you consider its a 5x5.

code/modules/mob/living/carbon/xenomorph/strains/castes/praetorian/valkyrie.dm
```diff
@@ -36,7 +36,7 @@
 	var/fury_max = 200
 	var/fury_per_attack = 15
 	var/fury_per_life = 5
-	var/heal_range =  3
+	var/heal_range =  5
```
realforest2001: This is too extreme a balance change.

realforest2001 (changes requested):

code/modules/mob/living/carbon/xenomorph/strains/castes/praetorian/valkyrie.dm
TheBoondock: ok changed it back to 3

Red-byte3D: Edit your pr desc

MistChristmas:
If you're doing any visual changes you should probably provide screenshots to preview. Personally I don't like the green outline. And it's supposed to be a cleanse not a heal, so it should be something like blue instead, ditto with the heal overlay.
Screenshot 2026-05-22 013610


Something like:
image

Which is `new_owner.add_filter("grace", 1, list("type" = "outline", "color" = "#00d0f5", "size" = 1, "alpha" = 90))`
But just an example. Defer to Spritetainers

DexterDude:
I think it may be a good idea to use either gold or white for the outline to showcase that they are going to be invincible for the duration. Currently green and blue are sort of the "healing" colors, and I can't recall any ability that uses white or gold. You'll also probably want to add a horizontal loop or bar going up to signify the "cleansing" that the ability does. Think of the shields from Halo when recharging.

shieldstation

Detective-Google:
I don't think valkyrie is underperforming or needs a significant buff package in observed practice.

PR #12142, opened 2026-04-21
Xenomorphs can evolve while injured predrop. 

Ammoniacres:
# About the pull request

Lets xenomorphs evolve while below 100% health before marines have landed on the planet. This restores *only* the health difference between both max health capacities (if applicable), and confers no other advantages. 

No longer punishes players for hunting survivors before marines land. Alleviates awkward bumps in flow between evolution periods predrop, and encourages engaging with survivors.

<details>

https://github.com/user-attachments/assets/c0dce8e1-d55e-497f-b8ed-240597966e3a

</details>

# Changelog

:cl:
balance: Xenomorphs can now evolve below 100% health before marines make their first drop. 
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Evolution.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Evolution.dm
@@ -271,7 +271,7 @@ GLOBAL_LIST_EMPTY(deevolved_ckeys)
 		to_chat(src, SPAN_WARNING("We are already the apex of form and function. Go forth and spread the hive!"))
 		return FALSE
 
-	if(health < maxHealth)
+	if(health < maxHealth && SSobjectives.first_drop_complete) //lets xenomorphs evolve while injured predrop
 		to_chat(src, SPAN_WARNING("We must be at full health to evolve."))
 		return FALSE
 


Drulikar:
It is intentional that you need to be at full health to evolve.

Drulikar:
DaTimeSmog really wants an aspect of this however, so speak with them on something we sort of work-shopped along the lines of:
- Allow evo & devo & transmutation if health > some percentage (atleast ~25% please)
- This process puts you in a cocoon/mutation phase where you're vulnerable
- Duration of cocoon/mutation phase is some static amount (e.g. 15s) + additional time it would take to heal to full
- Can only occur on weeds for T2+
- Evo/devo/transmutaiton always restores to full hp rather than partial hp (current code appears to retain lost hp) if fully completed. This means essentially the cocoon will "heal" them using that additional time penalty; not something external. This is so the xeno isn't penalized if someone takes away their weeds or pheros or something to make it take longer so its not griefable.
- Taking damage cancels mutation (losing weeds or being tackled does not)

The reasoning for this is
> Evo, just like de-evo requires full health because the intention is for this to be a change that is only done for someone not currently engaged in something. Its near instant presently and it should not ever be considered something that can be used in a combat situation either - even if its something simple as just confusing a marine or whoever that they were chasing some other target.

Ultimately even the mutation thing does not solve the problems I see with slots. I would much rather something like slots being more a lottery somehow - but I don't currently have any ideas that I currently like as far as how to go about implementing that.

PR #7194, opened 2024-09-22
WIP: ndatabase Refactor

boskoramen:
# About the pull request

Refactoring ndatabase to be a bit more readable and redefining the interface for defining entities, entity links, and views to reduce boiler plate. Using field and field type datums to validate that appropriate fields and field types are passed.

Also adding unit tests to validate views and links at the very list.

Will try to document what is going on with ndatabase since variable name changes might not be sufficient in that front.

Feedback from those more familiar with how this works is much appreciated.

### Implementation
- [x] Update DB adapters to handle datumized fields and field types
- [x] Update existing DB entities to use new interfaces
- [x] Make foreign key for entity links connect to a field on entity
- [ ] Make sure native functions are correctly handled
- [ ] Extract entity management from entity meta into entity manager itself or entity managing class (separation of concerns)
- [ ] Update `internal_parse_column` and `internal_generate_view_query` so that `meta_to_load` maps entity nodes to table names
### Documentation
- [ ] Add documentation for database interfaces (adapter, connection, etc)
- [ ] Add documentation for database entities (including table definition)
- [ ] Add documentation for views
- [ ] Add documentation for indexes
- [ ] Add documentation for database implementations
- [ ] Rename abbreviated/shortened variables to be fully descriptive
- [ ] Update the `meta_to_load` and `meta_to_table` variable names to more clearly reflect what they represent in adapters
- [ ] Add documentation for `internal_parse_column` and `internal_generate_view_query` since aliasing is not very clear without digging into the code
### Validation
- [ ] Add unit tests to validate links are properly configured
- [ ] Add unit tests to validate links are not duplicated
- [ ] Add unit tests to validate view fields (especially those that are through entity links)
### Misc
- [x] Understand DB indexing update
### Nice to have's (will probably do these changes in separate PRs to avoid scope creep)
- Interface for migrating data between tables
- Non-entity based views
- Interface for junction tables (many-to-many mappings) 

# Explain why it's good for the game

With datumized preferences port I am planning to do would like to try storing all preferences in the database instead of through savefiles. Wanted the DB to be a lot easier to work with before diving into that.
# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

tba

</details>

# Changelog
no player facing changes

PR #12146, opened 2026-04-23
DB Index search, additional int types, db side customization of id (name/value type), adding "as anything" in for var operators in entity manager

The-Last-Shadow:
### If you need any further assistance with the database - whether it’s improving existing functionality, adding new features, or fixing issues I’m unaware of. Just let me know and I’ll be happy to help.

Also, I like #7194 Pr, but it's a shame the author has disappeared. 
If that Idea/PR still has the green light for development, I can look forward into what I can do.

# About the pull request
* Added search support for /datum/db/index. Previously _ndatabase could only create indexes. Now it can also simplify searching by automatically generating filters with make_filters_for_index. You can also create entities via DB_INDEX, which greatly eases working with both the indexes and any entities linked to them, effectively replicating the select_by_key system.

* The key_managed list originally relied on the assumption that nobody would change a secondary key of an already‑loaded entity, because updates are performed only by id. Consequently, changing a key could cause conflicts only if the database already contained the new key. I’ve improved this behavior, the list now updates when keys exist and overwrites them using a cached_keys_managed variable on each entity. It removes old links and writes the new ones. If no keys are present, the list usage remain unchanged.

* Added missing integer types and unsigned support to the DB: TINYINT, SMALLINT, MEDIUMINT, plus unsigned variants for all of them.

* Added **as anything** in some places related to DB where it required.

* Reformatted various pieces of code to follow the standard DB interaction conventions.

* Basic ID data type changed to UNSIGNED BIGINT

* One known issue remains: when you change an key_field or rename /datum/db/index, the old index stays in the DB and a new one is added. I didn’t fix this because I wasn’t sure whether it’s required. Let me know if you need this addressed. I can see more potential issues, lets skip that till another day.

DB is good.

# Testing
Not required.

# Changelog

:cl:
add: control over DB side ID name and type, now you can set different data type (still only number)
add: DB more INT types and unsigned variation for use
code: DB more indexes functional
code: DB key_managed better handling
server: Basic ID type now changed to UNSIGNED
/:cl:

### DIFF
--- a/code/__DEFINES/db_defs.dm
+++ b/code/__DEFINES/db_defs.dm
@@ -21,6 +21,7 @@
 // SOFTWARE.
 #define DB_ENTITY SSentity_manager.select
 #define DB_EKEY SSentity_manager.select_by_key
+#define DB_INDEX SSentity_manager.select_by_index
 #define DB_FILTER SSentity_manager.filter_then
 #define DB_FILTER_LOCAL SSentity_manager.filter_local
 #define DB_META SSentity_manager.tables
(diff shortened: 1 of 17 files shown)


Drulikar:
Up to Harry or author if they want to revive this I guess.

PR #12154, opened 2026-04-25
Fix: Add Senior Command Access to Dropship controls

MistChristmas:
# About the pull request
Let's the aCO if they don't have dropship or leader, neither of which they get from the auto "promotion" use the dropship consoles if they're unlucky enough to be aCO without already having those things.

Nice of ARES to gift them the gift of being able to put the Dropships on Auto
<details>

</details>

# Changelog
:cl:
fix: Dropship controls now recognise Senior Command Access as valid.
/:cl:

### DIFF
--- a/code/modules/shuttle/computers/dropship_computer.dm
+++ b/code/modules/shuttle/computers/dropship_computer.dm
@@ -3,7 +3,7 @@
 	desc = "A flight computer that can be used for autopilot or long-range flights."
 	icon = 'icons/obj/structures/machinery/shuttle-parts.dmi'
 	icon_state = "console"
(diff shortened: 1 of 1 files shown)


realforest2001 (approved):

zzzmike:
side effect of this is that CMO can now always launch dropship even when not aCO - they couldn't before. but who knows if that was a bug or not as well, I guess

MistChristmas:
> side effect of this is that CMO can now always launch dropship even when not aCO - they couldn't before. but who knows if that was a bug or not as well, I guess

The CMO is 5IC. And part of the Senior Command staff. Why should they not have dropship access?

PR #12173, opened 2026-05-05
you can see mobs within 6 tiles

cuberound:
# About the pull request

right now the -15 acc debuff is applied even when the mob is on your screen edge tile aka you can fully see it

ehh projectile code is jank, this does not give you hidden acc debuff on stuff you can still see

<details>

</details>

# Changelog
:cl:
fix: now -15 accuracy debuff from being unable to see your target is applied when target is off screen, rather then from the last tile on screen 
/:cl:


Drulikar:
Doesn't need to change as I've been explaining to you in DMs. You were already going the route that it needs to account for scopes too; but its behaved this way for 9 years.

PR #12215, opened 2026-05-13
Roller Bed safety checks.

realforest2001:
# About the pull request

As title

Roller bed combat buckling is not something I anticipated having to deal with.

<details>

</details>

# Changelog
:cl:
add: Roller bed mob-click buckling no longer works on enemies unless they're dead or unconscious.
/:cl:

### DIFF
--- a/code/game/objects/structures/stool_bed_chair_nest/bed.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/bed.dm
@@ -312,7 +312,7 @@
 	user.temp_drop_inv_item(src)
 	forceMove(roller)
 	SEND_SIGNAL(user, COMSIG_MOB_ITEM_ROLLER_DEPLOYED, roller)
-	if(target_mob)
+	if(target_mob && ((target_mob.faction == user.faction) || target_mob.stat))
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/stool_bed_chair_nest/bed.dm
```diff
@@ -312,7 +312,7 @@
 	user.temp_drop_inv_item(src)
 	forceMove(roller)
 	SEND_SIGNAL(user, COMSIG_MOB_ITEM_ROLLER_DEPLOYED, roller)
-	if(target_mob)
+	if(target_mob && ((target_mob.faction == user.faction) || target_mob.stat))
```
Drulikar, line 315: Checking this here doesn't really solve the issue; only a convenient way. I would recommend something like this instead (though HUMAN_STRIP_DELAY as a base is probably too long so should probably use its own define for hostile buckling) 

[patch.diff.txt](https://github.com/user-attachments/files/27736622/patch.diff.txt)

Drulikar (changes requested):

PR #12229, opened 2026-05-15
smartgun fixes + modularized ability actions

### LINKED ISSUE #8271: Smartgun Frontline Mode doesn't work with IFF Disabled
### Testmerges

#7850 #7607 #7650 #8226 #8217 #8224 #8270 #8264

### Round ID

_No response_

### Description of the bug

You can toggle IFF on the Smartgun to determine whether your bullets hit friendlies or not. 
If you have IFF disabled and turn on frontline mode, you still can't hit friendlies - the Smartgun refuses to fire. 
Found this out during the new Mutiny Round mechanics. 

If IFF is disabled, you should be able to shoot at people regardless if they have IFF or not. 

1. Disable IFF on your Smartgun
2. Enable Frontline Mode on your smartgun
3. Try to shoot at someone with your IFF. Even with smartgun IFF disabled, you cannot shoot at people. 
...

### Issue Bingo

### LINKED ISSUE #11928: USCM Solar Devils smartgunner frontline mode bug
## Testmerges
<a href="https://github.com/cmss13-devs/cmss13/pull/10999">#10999: 'Fix of various plane related issues' by Blundir at commit f85a7153c4</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/9741">#9741: '[TM v0.9] Dancer Praetorian: Buffs and Changes.' by Venuska1117 at commit e9f22468cf</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/11823">#11823: 'Fixes Minimap Bugs and more' by TrollerNoob at commit 5195ebc3e1</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/11075">#11075: 'Ticket Panel (Adminhelps &amp; Mentorhelps)' by GeneriedJenelle at commit 4ad8d0cf87</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/10414">#10414: 'Port: Harm Intent uses Auto-Fire instead of PB' by MistChristmas at commit 8cb88f17ef</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/10823">#10823: 'Abyssal Station - (Lowpop) Ground Map' by Steelpoint at commit 1d6cfafd01</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/11902">#11902: 'implement pii scrubbing for sentry' by hry-gh at commit c216b6135f</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/11901">#11901: 'Reduce overtime in relay_render_to_plane' by noelle-lavenza at commit ae81d26562</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/11885">#11885: 'Various FTL Fixes' by Drulikar at commit d912ea0985</a><br>

## Round ID
32931

## Description of the bug
if you spawn in as the solar devils ert as an sg, the frontline mode toggle doesnt appear on your gun or hud, but if you have it hotkeyed it still toggles on, but you cant toggle it off

i expected for the sg to either have toggleable frontline mode, or prevent frontline mode toggling at all

1) Spawn as Solar Devils Smartgunner
2) Bind Toggle-Frontline-Mode hotkey to any key you want
3) Equip Smartgun
4) toggle frontline mode with hotkey
5) ???

## Attached logs
```
N/A
```

## Additional details
- Author: Despadoru
- Admin: SmellyHippie
- Note: None

kitchenenthusiast:
# About the pull request
fixes #11928 ; SD SGs now get a little warning in chat if they try to toggle frontline on that they can't do that, and they can't toggle it anymore

fixes #8271 ; i did the thing koba suggested; having IFF off is now mutually exclusive with frontline mode. turning IFF off when frontline mode is on will force frontline mode off and turn IFF off. the user gets a little line in the chat saying frontline is being disabled. then, when the user toggles frontline mode back on, it forces IFF back on also. the user gets a notif in the chat saying iff is back on. mechanically this doesn't change how it works/has worked, but it gives more feedback to user and automagically handles toggling between the two.

refactored ability actions for the HUD icons.

bug fixes of first two are good!

to fix a lot of the stuff for the second, i ended up refactoring the ability icons on the hud and the way they're handled for all the abilities. this makes it so if anyone wants to add abilities in the future (or change future interactions between abilities like the IFF/frontline stuff), the code is more modular and they don't have to define the state changes for icons on each action proc.

this is not the fullest refactor i could do as i have discovered a kind of horrible logic bug and some ways to optimize for when multiple SGs are active; however, i will save the first for a different PR since it would fundamentally change the way the SG plays. 😏
solar devils sg can't toggle frontline mode anymore and gets a notif they can't manually change modes.
all buttons toggle as intended and correctly represent the actual state change of the gun/ammo. each action produces a chat message output to the user and balloon alert where appropriate (wherever they were originally, i didn't add any).
IFF off forces frontline mode off; frontline mode on forces IFF on. IFF off forcing ammo back to HP now correctly updates the HUD.
overlays for actual item state changes (feed cover, drum) still work as intended.
Screenshot 2026-05-15 044009

my test video file is too big :(

# Changelog
:cl:
fix: Toggling IFF off on a smartgun now forces Frontline mode off; likewise, toggling Frontline mode on will force IFF back on.
fix: Solar Devils Smartgunners can no longer manually toggle Frontline mode.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/smartgun.dm
+++ b/code/modules/projectiles/guns/smartgun.dm
@@ -166,6 +166,7 @@
 		ammo_primary = ammo_primary_def
 		ammo_secondary = ammo_secondary_def
 	ammo = secondary_toggled ? ammo_secondary : ammo_primary
+	update_action_icons()
 
 /obj/item/weapon/gun/smartgun/set_bullet_traits()
 	LAZYADD(traits_to_give, list(
@@ -249,6 +250,10 @@
 			overlays += "+[base_gun_icon]_cover_closed"
 	else
 		overlays += "+[base_gun_icon]_cover_open"
+	update_action_icons()
+/obj/item/weapon/gun/smartgun/proc/update_action_icons()
+	for (var/datum/action/item_action/smartgun/Action in actions)
+		Action.update_button_icon()
 
 //---ability actions--\\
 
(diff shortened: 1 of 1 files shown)


Drulikar (changes requested):
Briefly looking at this PR, it really needs to be following https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#variables especially for newly added code.

Stuff like `var/datum/action/item_action/smartgun/Action` would be PascalCase when we use snake_case. In addition to that all the abbreviation usage.

kitchenenthusiast:
> Stuff like `var/datum/action/item_action/smartgun/Action` would be PascalCase when we use snake_case. In addition to that all the abbreviation usage.

just to be clear: for the abbreviations, all the `H`, `G`?

Drulikar (changes requested):
You are excessively calling `update_action_icons`

For example, if I press the action button for Toggle Frontline Mode this will cause 21 calls to update_button_icon:
- Every action button is updated when toggle_frontline_mode -> recalculate_attachment_bonuses -> set_gun_config_values -> update_action_icons occurs
- Every action button is updated when toggle_frontline_mode -> set_gun_config_values -> update_action_icons occurs
- Every action button is updated when toggle_frontline_mode -> update_action_icons occurs

1. `set_gun_config_values` will be called when defaulting values during `recalculate_attachment_bonuses`
2. Its unnecessary to have `set_gun_config_values` call update buttons if everything else already updates the icons OR its unnecessary for things that do call `set_gun_config_values` to then also update the icons
3. Its unnecessary to update all 7 actions when you're really only affecting 1-2
4. Changing the mode of an action should just be coded to update itself (meaning if you tell IFF to disable itself, IFF action should handle updating its icon)

Look at removing the need for `update_action_icons` to be added at all. It shouldn't be needed.

kitchenenthusiast:
I know this staled but thank you for taking the time to type everything out and explain.

PR #12241, opened 2026-05-17
Crusher Melee Attacks now Slow for 2 seconds (was 3.5)

larentoun:
# About the pull request

Actually it is a fix, since during a refactor 15 months ago `2 second slow` was not correctly set in the proc argument, so it was using the proc's default value of `3.5 second slow`

Fixes good?

# Changelog
:cl:
balance: Crusher Melee Attacks now Slow for 2 Seconds (was 3.5 Seconds)
/:cl:


MistChristmas:
Considering it's been like that for... a year and 3 months it's probably fine if it stays at 3.5 seconds.

kiVts (approved):

PR #12304, opened 2026-05-25
SHARP dart manual disarming now converts to usable ammo (+fixes)

### LINKED ISSUE #12239: Specialist Sharp, incorrect work. Stacks of incendiary darts.
### Testmerges

Doesn't matter

### Round ID

Doesn't matter

### Description of the bug

When two incendiary darts are stacked, they turn into some kind of strange bomb, capable of killing almost any T3 without any problem. The exception is the Queen and Ravager, as they don't burn. 

These are just incendiary darts that shouldn't do that much damage.

Recording how it happens. (https://drive.google.com/file/d/1LJoBYeUVaTj51nOoVI1v4hvk9V8i8esu/view?usp=sharing)
1. They fire two incendiary darts onto the ground to activate them.
2. I wait for the dart to charge to level 4.
3. I deactivate it with a knife and collect the darts.
4. I place a mine or other explosive object that activates when an enemy approaches.
5. I place two incendiary darts on the tile with the explosive object.
6. I move as a Praetorian or Crusher to the set composition (Death T3 and anything that burns).

### Issue Bingo

larentoun:
# About the pull request

Inactive SHARP darts (5 minute lifetime) can now be converted back into ammo with multitool. The same goes for active SHARP darts when multitooled.

Inactive SHARP darts are now truly safe and not explosive.

Also various fixes.

fixes https://github.com/cmss13-devs/cmss13/issues/12239

Inactive SHARP darts should not act as active darts; a little buff for ammo conservation?

<details>

Breaking inactive as xeno, disarming into reusable

https://github.com/user-attachments/assets/972d183f-2a6e-4226-ba9f-203d1863398c

Now completely safe when disarmed

https://github.com/user-attachments/assets/b792a713-3139-4220-9cd3-ccc902907422

</details>

# Changelog
:cl:
fix: Using any item should now not immediately disarm the dart - disarming is intended to be done with multitool.
fix: Disarmed (inactive) SHARP darts now are truly inactive and safe.
spellcheck: Incendiary dart description now says that's it is incendiary, not explosive.
balance: Disarming SHARP darts with multitool now converts them into usable ammo.
add: Can now use multitool on inactive darts. That means, 5 minute lifespan auto-disarmed darts can now be reused as ammo with multitool.
/:cl:


code/game/objects/items/explosives/mine.dm
```diff
@@ -469,8 +478,12 @@
 	QDEL_NULL(tripwire)
 	disarmed = TRUE
 	deltimer(timer_id)
-	add_to_garbage(src)
 
+/// Converts the mine back into usable ammo
+/obj/item/explosive/mine/sharp/proc/convert_into_ammo()
+	var/obj/item/ammo_magazine/handful/new_handful = new(get_turf(src))
+	new_handful.generate_handful(ammo_path, /obj/item/ammo_magazine/rifle/sharp::caliber, 5, 1, /obj/item/weapon/gun/rifle/sharp::type)
```
larentoun: Not sure how to handle it. Either this way, since there is no other rifles that use sharp, or add new arguments that set caliber and gun type when dart is shot

larentoun:
I have no discord communication, pls help

larentoun:
It will surely be fixed soon

larentoun:
Oh no.

larentoun:
F

larentoun:
Not yet time

Steelpoint (dismissed):
Balanced approved. Pending code review.

code/game/objects/items/explosives/mine.dm
boskoramen: should be able to do ammo_path::caliber and ammo_path::type

With that being said, you should add an assertion on initialize for sharp ammo that ammo_path is defined and is an ammo typepath so if folks add sharp ammo that is not incorrectly set unit tests will fail

code/game/objects/items/explosives/mine.dm
```diff
@@ -469,8 +478,12 @@
 	QDEL_NULL(tripwire)
 	disarmed = TRUE
 	deltimer(timer_id)
-	add_to_garbage(src)
```
boskoramen, line 473: I don't think removing this check and moving the qdel in prime is the way to go, the flow is weird since you're waiting until when mine is primed again to delete it.

Why don't we just call convert_into_ammo() in disarm?

code/game/objects/items/explosives/mine.dm
```diff
@@ -469,8 +478,12 @@
 	QDEL_NULL(tripwire)
 	disarmed = TRUE
 	deltimer(timer_id)
```
boskoramen, line 489: I'm noticing that we don't call deltimer on Destroy for sharp mines. Could you add that line too as part of bug fixes?

boskoramen (changes requested):

code/game/objects/items/explosives/mine.dm
larentoun: Disarm() happens after some time passes, making it an inactive mine. Converting into ammo is called when the person tries to multitool disarmed/armed mine to make it into a usable ammo.

I could make that the inactive-because-timer auto-turn into ammo, but wouldn't it look strange?

larentoun: Also, since sharp ammo can be reused (unlike spent shells), I decided to remove them from garbage, so it won't get deleted. Same if it was a bandage lying around ready to be reused.

boskoramen (approved):

PR #12329, opened 2026-05-28
Medicine Reflects Practiced Skill

Katskan:
# About the pull request

Brings medical doctor support roles in-line with other technical roles like engineers, intel, and command (JTAC) by increasing the medical and surgical skill of roles who would reasonably have lots of practice, and who don't have other varied combat or utility skills. Also modifies some of the flavor messaging around suturing at max skill to reflect that it's the result of being well practiced and this is your job (surgical training is a freaking 5 year long program you get kind of good at it), not robot magic. 

Other roles like the chief engineer, ASO, and even lowly maintenance techs have max engineer skill, roles like IO have max intel, CO max JTAC, etc. Max skill in these fields reflects practice and competency with the skill, rather than adding new functionality not available without that skill level. Roles which have a specialty have maximum skills for those roles, except for doctors. This standardizes the doctors whose primary roles is to practice medicine to be on the same level as other departments, while roles whose primary role is not to practice medicine/are backups (research) are physically capable of doing the same jobs but less practiced and efficient because their expertise is in doing other things, or in the case of ERTs or nightmare survivor inserts they have other roles like fighting.

Tested on local server

# Changelog

:cl:
balance: Increased medical practitioner doctor roles medical and surgical skill to reflect their specialized role and level of practice. Did not increase skills on roles such as ERT medics or nightmare survivors who have other skills reflecting a wider body of knowledge rather than a specialty. 
/:cl:


Detective-Google:
medical/surgical skill isn't really a matter of "oh other jobs have max in their field" but of overall action speed.

at the moment I don't think doctors are doing actions too slowly or need it sped up, so I don't see the need for this.

PR #12343, opened 2026-05-30
putting the 'swag' back into 'scimitar' - reduces clickdelay of wrist scimitars from 1 SECOND to 0.5 SECONDS, implements a temporary bugfix

Ollanius:
# About the pull request

Hello! Just a very brief contribution, nothing too fancy.
A quick thing, first:
* If the balance changes shown here feel a bit too radical, I'm _also_ completely fine with toning them down. No shame there.

With that out of the way, here's what I've changed:
* The Wrist Scimitars now have a click delay of 0.5 SECONDS instead of 1 SECOND - the very same as the Wristblades.
* Sets the Wrist Scimitar's attack speed bonus to 0 SECONDS instead of -0.4 SECONDS, as a temporary band-aided bugfix.
* Adjusts the verbage in the Wrist Scimitar's description from 'huge' _(shared by the Wristblades)_ to 'massive'.

* Wrist Scimitars take up a dedicated weapon slot _(like the Combistick and War Glaive),_ but are fairly weaker than the freely available Wristblades due to their heftsome click delay. This particular issue has been so prevalent that almost _every_ recorded mention of it in within the past year - both in-game and on-the-'cord - has been near-exclusively negative. This fixes that.
_(To note, this is almost complete inverse to what people thought about the Wrist Scimitar's performance, during my first tenure on the Hunter whitelist from a couple years back. You know something has to be wrong, when even a Xenomorph'll use LOOC to politely remark on the wristweapon's ineffectiveness.)_

* Implements a temporary(?) bugfix for a glitch that, to my knowledge, is still ongoing. I've left a note in the code so that the intended function can be easily restored, if-or-when a more permanent solution is found. Now, the Wrist Scimitars shouldn't inexplicably have a differing click delay depending on which gauntlet you've attached them to.

Relatively simple change, composed of a one-liner.

# Changelog

:cl:
balance: The Wrist Scimitars now have a click delay of 0.5 SECONDS instead of 1 SECOND, making them a slight - but direct - upgrade over the Wristblades.
code: Temporarily disables the ATTACK_SPEED_BONUS mechanic for the Wrist Scimitars, due to an ongoing bug. This shouldn't affect the mechanical balance too much, besides a -0.1 SECOND bonus to its click delay.
spellcheck: Adds a new verb to the Wrist Scimitar's description to better differentiate it from the Wristblades.
/:cl:

### DIFF
--- a/code/modules/cm_preds/yaut_bracers.dm
+++ b/code/modules/cm_preds/yaut_bracers.dm
@@ -502,7 +502,7 @@
 
 /obj/item/bracer_attachments/scimitars
 	name = "scimitar bracer attachment"
-	desc = "A pair of huge, serrated blades."
+	desc = "A pair of massive, serrated blades."
 	icon_state = "scim"
 	item_state = "scim"
 	attached_weapon_type = /obj/item/weapon/bracer_attachment/scimitar
@@ -511,7 +511,7 @@
 
 /obj/item/bracer_attachments/scimitars_alt
 	name = "scimitar bracer attachment"
-	desc = "A pair of huge, serrated blades."
+	desc = "A pair of massive, serrated blades."
 	icon_state = "scim_alt"
 	item_state = "scim_alt"
 	attached_weapon_type = /obj/item/weapon/bracer_attachment/scimitar/alt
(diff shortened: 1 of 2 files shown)


Ollanius:
On second thought, I may have kerfuffled the implementation.
I'll play it safe - this might be something better handled by someone with more coding _(and balancing)_ knowledge than me.

PR #12347, opened 2026-05-31
Move Admin Sleeping to use status effects

MistChristmas:
# About the pull request
Title.
Move it from the sleeping var to just applying status effects with a flag that stops it being removed by other sources and also stops humans whispering in their sleep.
Works a bit faster than sleeping.

Also stop rejuvenate from making them conscious or getting rid of the effect.

Does the thing faster, and stops people whispering rule breaks
<details>

</details>

# Changelog
:cl:
admin: Admin Sleeping works faster, Rejuvenate doesn't get rid of Admin Sleeping
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -120,6 +120,8 @@
 #define NO_PERMANENT_DAMAGE (1<<20)
 #define CORRUPTED_ALLY (1<<21)
 #define FAKESOUL (1<<22) // Lets things without souls pretend like they do
+#define ASLEPT (1<<23)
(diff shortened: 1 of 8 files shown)


fira:
You can just make a /datum/status_effect/incapacitating subtype 
 * that removes the need for 999999 hackery
 * you don't need an extra flag (especially given this is the 24th)
 * it has remove_on_fullheal builtin to handle rejuvenate

MistChristmas:
> * it has remove_on_fullheal builtin to handle rejuvenate

Yeah but I think it's better for it to only be affected by the Sleeping, lest someone actually tries to rejuv everyone in view and suddenly the griefer is at it again.

... Nevermind I get what you're pointing out

fira (approved):

PR #12355, opened 2026-06-01
riding element changes number 1 million

ori-disillusion-mirnov:
# About the pull request

moves proc definitions for mounting a mob from /mob/living/carbon/xenomorph to /mob/living

also changes the default component type to /datum/component/riding/creature so you can add it through VV's add element prompt without overriding the argument, theres no reason to ever be using /datum/component/riding anyways

sometimes i want to ride a cow and i cant do it with var editing because non xeno mobs dont have the procs

mobs are not ridable by default, xenos are not ridable by default. i checked :)

# Changelog
:cl:
code: all living mobs now have procs for riding
code: adding the riding element is easier
/:cl:


code/modules/mob/living/living.dm
```diff
@@ -720,3 +720,53 @@
...
+		return
+	visible_message(SPAN_NOTICE("[target_mounting ? "[target] starts to mount [src]" : "[src] starts hoisting [target] onto [p_their()] back..."]"),
+	SPAN_NOTICE("[target_mounting ? "[target] starts to mount on your back" : "You start to lift [target] onto your back..."]"))
+	if(!do_after(target_mounting ? target : src, 5 SECONDS, NONE, BUSY_ICON_HOSTILE, target_mounting ? src : target))
+		visible_message(SPAN_WARNING("[target_mounting ? "[target] fails to mount on [src]" : "[src] fails to carry [target]!"]"))
+		return
+	//Second check to make sure they're still valid to be carried
+	if(target.is_mob_incapacitated())
+		return
+	buckle_mob(target, usr, target_hands_needed = 1)
```
fira:
```suggestion
	buckle_mob(target, src, target_hands_needed = 1)
```

fira (changes requested):
LGTM otherwise

fira (approved):

code/modules/mob/living/living.dm
```diff
@@ -720,3 +720,53 @@
...
+	//Second check to make sure they're still valid to be carried
+	if(target.is_mob_incapacitated())
+		return
+	buckle_mob(target, src, target_hands_needed = 1)
+
+/mob/living/MouseDrop_T(atom/dropping, mob/user)
+	. = ..()
+	if(!can_mount(user, TRUE))
+		return
+	INVOKE_ASYNC(src, PROC_REF(carry_target), dropping, TRUE)
```
fira, line 772: I think you might want to remove the /mob/living/carbon/xenomorph one since it's going to be the same one doubled up.

Also in the future this might conflict with the handler on human, but this will need refactoring later

fira (changes requested):
Just an extra thing

fira (approved):

PR #12363, opened 2026-06-03
"""fixes""" xenos not being stunned by impact ammo while resting

### LINKED ISSUE #1570: M4RA Impact Scout Ammo not stunning resting Xenos
## Description
When using the M4RA Impact ammo on a resting Xenos, they simply shrug it off and can standup anytime they wish to. Unlike when they're standing where they get knocked down & pushed away one tile into the direction the shot was going.

Doesn't affect the Xeno at all, no stun or knockback.
What it should be doing: Knockback & Knockdown/Stun

## Steps to reproduce
1. Have the scout kit. 
2. Load Impact ammo.
3. Find a resting Xeno.
4. Shoot the resting Xeno.
5. Get dunked or lose the Xeno due to no stun/knockback.

ori-disillusion-mirnov:
# About the pull request
not a bug since it was very intentionally being checked that whoever was hit was resting, but the report is open 4 years later so here we are

fixes: #1570 (as much as you can fix a non-bug)

"bugs" bad

<details>

</details>

# Changelog
:cl:
balance: resting xenos are now stunned by impact bullets
code: change some one letter vars since i was originally going to do something there but changed my mind
/:cl:

### DIFF
--- a/code/datums/ammo/ammo.dm
+++ b/code/datums/ammo/ammo.dm
@@ -153,7 +153,7 @@
 			return
 	if(!living_mob || living_mob == fired_projectile.firer)
 		return
-	if(fired_projectile.distance_travelled > max_range || living_mob.GetKnockDownDuration() != 0)
+	if(fired_projectile.distance_travelled > max_range || living_mob.IsKnockDown())
 		return //Two tiles away or more, basically.
 
 	if(living_mob.mob_size >= MOB_SIZE_BIG)
--- a/code/datums/ammo/ammo.dm
+++ b/code/datums/ammo/ammo.dm
@@ -153,7 +153,7 @@
 			return
 	if(!living_mob || living_mob == fired_projectile.firer)
 		return
-	if(fired_projectile.distance_travelled > max_range)
+	if(fired_projectile.distance_travelled > max_range || living_mob.GetKnockDownDuration() != 0)
 		return //Two tiles away or more, basically.
 
 	if(living_mob.mob_size >= MOB_SIZE_BIG)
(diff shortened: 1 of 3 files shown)


fira:
Won't that mean two persons with Slugs or such can now completely stunlock?

ori-disillusion-mirnov:
> Won't that mean two persons with Slugs or such can now completely stunlock?

yeah thats probably what the check was for

fira:
`IsKnockDown` / `GetKnockDownDuration` are legacy crutches, here are some ways I could suggest handling it (if any is satisfactory) : 

 * body_position == LYING_DOWN (current in master)
    Can be cheesed by rest, can only be chained with precise timing
 
 * Nothing (previous)
    Can be chained without regard for timing, but at least it won't stack, only refresh

 * Bypass lying down check if resting
    Bit convoluted, but you could argue resting makes you more vulnerable
    Fixes cheesing it, and that's it 

 * Check for either KnockDown or Stun effect (currently in PR)
    Drawback is you essentially lose the Slow if the target is KnockedDown (or Stunned) by another source, it doesn't apply. Uses legacy logic but it's not sooo bad i guess

 * Check either for `TRAIT_IMMOBILIZED` or `MOBILITY_MOVE` instead
    Same, but also doesn't apply if there is an explicit Immobilization source (KD and Stun both apply immobilization). But there's like 3 in the game

 * Check `TRAIT_FLOORED` 
    Not viable, you can be floored for other reasons

fira (commented):
Anyway code OK just need someone to check balance properly about this

Drulikar:
Not a bug. They still take damage, and a something laying on the ground is going to be harder to move than something standing.

PR #12369, opened 2026-06-03
replace verb hidden with invisibility

### LINKED ISSUE #1674: Hidden verbs are visible

ori-disillusion-mirnov:
# About the pull request

fixes: #1674, set hidden doesn't work and ignores whatever you try to set it to, instead giving you nothing. invisibility does work though so i use that to do what hidden was supposed to, it doesnt look like it was used anywhere except 2 Life() procs but i think thats a different kind of invisibility and shouldn't matter anyways

apparently you can only use the verbs if you have "`see_invisibility` higher". i dont know what that means. you can use the verbs through keybindings and the chatbar, and the only mention of any "see_invisibility" is in the comment that says that so its probably nothing to worry about

/datum/action has its own hidden var that probably works maybe so some checks for hidden are still left in

verbs meant to be hidden from stat panel are hidden

<details>

</details>

# Changelog

:cl:
fix: hidden verbs no longer appear in the stat panel, as they use invisibility now
code: rip limb is hidden
spellcheck: fixes typo in rip limb description
/:cl:

### DIFF
--- a/code/_onclick/hud/hud.dm
+++ b/code/_onclick/hud/hud.dm
@@ -282,7 +282,7 @@
 //Triggered when F12 is pressed (Unless someone changed something in the DMF)
 /mob/verb/button_pressed_F12()
 	set name = "F12"
-	set hidden = TRUE
+	set invisibility = TRUE
 
 	if(hud_used && client)
 		hud_used.show_hud()
--- a/code/_onclick/hud/map_popups.dm
+++ b/code/_onclick/hud/map_popups.dm
@@ -195,5 +195,5 @@
  * When the popup closes in any way (player or proc call) it calls this.
  */
 CLIENT_VERB(handle_popup_close, window_id as text)
-	set hidden = TRUE
+	set invisibility = TRUE
 	clear_map("[window_id]_map")
--- a/code/controllers/subsystem/minimap.dm
+++ b/code/controllers/subsystem/minimap.dm
@@ -1627,7 +1627,7 @@ SUBSYSTEM_DEF(minimaps)
 /client/proc/handle_draw(mouse_x as num, mouse_y as num, size_x as num, size_y as num, view_size_x as num, view_size_y as num)
 	set instant = TRUE
 	set category = null
-	set hidden = TRUE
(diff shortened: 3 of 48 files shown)


Detective-Google:
> apparently you can only use the verbs if you have "see_invisibility higher"

https://ref.dm-lang.org/verb/set/invisibility  | https://ref.dm-lang.org/mob/var/see_invisible 
tldr you can see anything that your see_invis var >= invis var of the thing

Detective-Google:
considering invis is expecting a # from like 0-100 and you're providing TRUE which just evaluates to 1, and, a ghost, for example, has a see_invis of 60, Im not sure this PR would do what you're intending 
image

Detective-Google:
image
 some additional context from the blue book

ori-disillusion-mirnov:
> considering invis is expecting a # from like 0-100 and you're providing TRUE which just evaluates to 1, and, a ghost, for example, has a see_invis of 60, Im not sure this PR would do what you're intending

it hides it from stat panel and thats all i really want to do

ori-disillusion-mirnov:
here is .hidden not working on a clean branchl , the xeno_primary_action verbs should be hidden and the rest should be defaulting to 0 because thats what i set the hidden var to default to in procpath.dm

image

fira:
`see_invisible` is a vision thing i'm pretty sure, i don't think that's what it's intended for

We can probably homebrew it somehow for the statpanel since it's entirely under our control, but `hidden` also hides them in the native command bar

fira (changes requested):
As per comments above

PR #10696, opened 2025-09-22
[code b***ty] Sentinel rework

cuberound:
# About the pull request

Sentinel is in a bad spot and even that is an understatement with the role currently in a limbo between being utterly worthless or a stun duration increase of 1 second away of being the best capper in the game, with the marine on the receiving side being absolutely unable to fight it in any way possible. This makes sentinel perhaps THE most difficult T1 to balance as the alternative will always be the caste being utterly useless and unfun to play, or extremely overpowered and unfun to play against.

TGCM solved this issue by having sentinel work around a stacking neurotoxin mechanic that dictates the stun duration as well as rewarding the sentinel for engaging in close range.

Goals

The idea is to add a unique gameplay loop to sentinel without infringing its core identity of being a stun based capture xenomorph that focuses on debuffing enemies from afar before going in for the final sweep.

Non-goals

Turn sentinel back into a nightmare caste where it flipflops between overpowered or worthless based on small numerical changes.

Content

Porting over the TGCM sentinel kit with some adjustments to fit the current CM playstyle.

Passive: Neurotoxin

Neurotoxin is applied by the sentinel’s “Toxic Spit” and “Toxic Slash”. The maximum amount of stacks a target can have is 30. 
Neurotoxin stacks apply OXY damage to a maximum of 50

Toxic Spit | 1 Cooldown | Costs: 50 plasma | Applies 5 Stacks

Slows a marine as well as applying neurotoxin stacks. Passes other xenomorphs as if they have IFF.

Toxic Slash | 10 Seconds Cooldown | Costs: 100 plasma | Applies 10 Stacks 

Inflicts large amounts of Neurotoxin stacks per slash. Only lasts for either 3 slashes or if the duration of 5 seconds has passed.

Catalytic Shock Tailstab | 10 Second Cooldown | Windup Time: 1 Second

Same behavior as a regular tailstab, with no damage modifier however. Applies following effects

Stab your victim, draining some Intoxicated stacks and gaining benefits from it. It removes 50% of the current stack amount on use.
If the victim is above 20 stacks, this ability will increase armor by 10 and grants the sentinel a movement buff.

Deals a proportional amount of damage. The formula for this damage is: ( Amount of stacks * 6 ) / 5. This translates to a minimum of 1.2 damage and a maximum of 36 damage NON-AP

Draining Bite | 12 Second Cooldown | Costs: 100 plasma

Bite your victim, draining some Intoxicated stacks and gaining benefits from it. It removes 30% of the current stack amount on use.

[...]

PR #12378, opened 2026-06-05
Reworks Sentinel neurospit, removes conditional fullstun

Ammoniacres:
# About the pull request

Reworks the effects of Sentinel's neurotoxic spit to no longer fully knockdown targets without armor, instead increasing the length slow state, enforcing a stutter, daze and screenshake. On top of this, unarmored targets are burned for 5-15 points of damage, to still encourage the caste to be used against armorless targets. 

Sentinel has been in a terrible spot for a long time now, as a caste entirely built around ranged stuns and debuffs. While I think a lot of the caste's problems require substantial positive tuning on the part of its effectiveness, I don't think those problems justify a T1 having access to a ranged stunlock, no matter how conditional, as it encourages skewed use cases that are entirely uninteractive for the receiving player. Currently, as it stands, the caste is in dire need of a rework that addresses its wider issues, such as #10696. For this particular PR, it addresses this specific balance aspect that rework doesn't. 

<details>

</details>

# Changelog
:cl:
balance: removed Sentinel's neurospit stunlock condition for unarmored targets, replaced it with greater slow, daze, screenshake and burn effects.
/:cl:

### DIFF
--- a/code/datums/ammo/xeno.dm
+++ b/code/datums/ammo/xeno.dm
@@ -38,71 +38,73 @@
 
 	neuro_callback = CALLBACK(GLOBAL_PROC, GLOBAL_PROC_REF(apply_neuro))
 
-/proc/apply_neuro(mob/living/M, power, drain, insta_neuro = FALSE, drain_stims = FALSE, drain_medchems = FALSE, apply_effect = TRUE)
-	if(skillcheck(M, SKILL_ENDURANCE, SKILL_ENDURANCE_MAX) && !insta_neuro)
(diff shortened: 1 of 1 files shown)


Drulikar:
In general sentinel is a capturing focused caste, and the mechanic is to easily catch survivors that didn't bother or couldn't find armor in time. If there's maps with little to no armor available, that's a mapping problem. I don't see this change as an improvement for a caste that tends to be underpowered, nor do I want to encourage survivors to go armorless for speed, nor do I want survivors to have another situation they can just suicide to avoid capture when it's already too late.

PR #12383, opened 2026-06-06
fixes bypassing engi skill upgrade limit

Detective-Google:
# About the pull request

jobs like medics and certain specs, etc are not meant to have skills equivalent to squad engis, and are not meant to be able to do things like fix APCs. This is in an attempt to not have people powergame and defeat the teamwork aspect of the game.

in the upgradable pamphlet PR, the cap for the machine gun pamphlet was mistakenly set at 2, which allowed users to use an engineering pamphlet and then a HMG pamphlet to obtain the ability to repair APCs etc as jobs which are not intended to have that ability, such as scout specialist and medic.

this pr resolves that by lowering the cap

fixes an oversight

# Changelog

:cl:
fix: pamphlets can no longer be used to bypass intended engi skill caps
/:cl:


Nihisohel:
it was fun until it lasted

fira (approved):

realforest2001 (approved):

PR #12396, opened 2026-06-07
Eliminates hunted faction lists (fix for Hunted PMCs)

Unknownity:
# About the pull request

Removes Hunted Factions from their respective faction lists as it was clutter and it didn't do anything as of now + it broke some things (Hunted W-Y not showing up in the Hunted observer list).

Added Hunted PMC faction to the list of allowed factions that have W-Y gun restrictions so the hunted PMCs can shoot their guns.

**Disclaimer for anyone who wants to bring them back to the lists:** You will have to re-order FACTION_LIST_HUNTED above FACTION_LIST_WY in https://github.com/cmss13-devs/cmss13/blob/a4bb96a37b75529698dbf3d01820c7a71b494ffc/code/modules/mob/dead/observer/orbit.dm#L201-L217 so the Hunted PMCs still show up in the Hunted orbit menu.

Unecessary clutter bad? Also bug fix good

tested every faction still being in the hunted observer list by spawning them in, controlling them then aghosting and checking the observer list.

tested Hunted PMCs can still shoot their guns by spawning them all and controlling them.

Also tested IFF with Hunted factions and it works properly (regarding smartguns, NOT turrets)

<details>

</details>

# Changelog

:cl: Unknownity
fix: Hunted PMCs and Goons show up in the Hunted observer list properly now.
/:cl:


code/modules/projectiles/gun_helpers.dm
MistChristmas: I dunno why we're checking if there's a mind here of all places, but we shouldn't need to. So kill that.

The `faction in FACTION_LIST_WY` checks should be first.
The job check should die apart from Ivan I guess.
And `switch(user.faction)` should also die.

Why elite mercs get WY Restricted weapons at all is dumb. They can have the no_lock version if existing or created. Otherwise just shove them a different gun.
`/datum/equipment_preset/proc/spawn_merc_elite_weapon`

MistChristmas (changes requested):

MistChristmas (approved):

PR #11008, opened 2025-10-30
Civilian-nization of Research, Complete Remap of Research Wing/Containment

LC4492:
# About the pull request

This PR focuses on trying to expand the RP opportunities and quality of life of the researchers by taking opportunity of the middle deck added by the multi-Z conversion and by improving their "civilian" aspect through a series of changes.

The entirety of upper medical have been re-mapped, and research received an expansion that involves the addition of new rooms, such as a conference room, and changes for the quality of life of the researchers themselves (Players that like to play as researchers have been consulted on what could be done to improve their gameplay). Containment has been transferred down to the middle deck and is now connected to the upper via a staircase localized in the center of the Research Wing.

Researchers now have access to a much larger array of attires, including new shoes, labcoats, uniforms, ties and etcetera, everything to allow for a much bigger aesthetical freedom, without going against the idea of the role itself. Now researchers fall onto the same flags as the Corporate Liaison, and for this reason will not be receiving service medals according to playtime tier anymore (Playtime ranks still apply).

It has been communicated to me that the use of the BIOSUIT is vital for survival in the Research Laboratories, the reason being that it's the only apparel that grants biological protection, being true given that a labcoat, for example, gives a mere 15 points of protection, with the max being 100 (The BIOSUIT has 100 points of protection). For that reason, I raised the protection of labcoats to 50 points (Half of what a BIOSUIT would give) and made a smaller increase of protection in other items that would contribute (latex gloves give an extra 15 points, sterile masks give an extra 15 points). In the meanwhile, the BIOSUIT has received a slowdown, to make it's use less appealing and to incline researchers to use the equipment that has been modified.

Also fixes issue #11380 by making the grass patches inside the Psychatric Care Unit fake.

# Explain why it's good for the game

Must be, not sure, people will tell.

# Testing Photographs and Procedure
<details>
<summary>SCREENSHOTS OF THE MAP CHANGES</summary>

UPPER DECK MEDICAL
image

MIDDLE DECK CONTAINMENT
image

</details>

# Changelog

:cl: Blundir. Forest2001. LCMS1.
add: Adds a second heavy-duty stretcher to containment cells to compensate for the longer travel time.
maptweak: The entirety of upper medical have been re-mapped, and research received an expansion that involves the addition of new rooms, such as a conference room, and changes for the quality of life of the researchers themselves (Players that like to play as researchers have been consulted on what could be done to improve their gameplay). Containment have been transferred down to the middle deck and is now connected to the upper via a staircase localized in the center of the Research Wing.
[...]

PR #12407, opened 2026-06-09
Adds Update Path Script for Containment Tile Changes Made In PR #11008

LC4492:
# About the pull request

Adds Update Path Script for Containment Tile Changes Made In PR #11008.

Was pending in the original PR.

# Changelog
No player facing changes.


tools/UpdatePaths/Scripts/11008-updates-containment-tiles-paths.txt
```diff
@@ -0,0 +1,3 @@
+/turf/open/floor/almayer/research/containment/floor1 : /turf/open/floor/almayer/research/containment/floor2/west{@OLD}
```
Drulikar: `/turf/open/floor/almayer/research/containment/floor1` is still a path that exists and is used in some places, so running this script will make all usages of this path now become `/turf/open/floor/almayer/research/containment/floor2/west` which I doubt you want. 

In order for this update path line to exist, you either need to for certain want no map to use `/turf/open/floor/almayer/research/containment/floor1` or to manually edit one of the two situations in maps & code to be a unique path so that they remain distinct. So unless its properly done, this line should be removed.

Running this script right now would affect all these maps in ways you don't intend: 

Image

So be sure to test your script to see if its doing what you want (on current master it likely should do nothing unless there's maps you forgot about and got past Steelpoint).

E.g. on the Almayer:

Image
becomes
Image

If you notice the center tile is now wrong.

Drulikar (changes requested):
Presently running this script on say sulaco would change the broken map from 

Image
into 
Image

If you notice the top right corner is still wrong (but the missing tiles fixed) - and maybe there's no way to perfectly correct it with how it was originally done - but pretty sure if the new changes were distinct paths from the old you could - but its whatever I guess

Drulikar (approved):

PR #12410, opened 2026-06-09
Rename wield_delay define to more generic weapon_delay

MistChristmas:
# About the pull request
Split off from https://github.com/cmss13-devs/cmss13/pull/10414
Cause otherwise it would conflict with some PRs.

It'll be functioning for two things eventually.
<details>

</details>

# Changelog
:cl:
code: Rename some defines from Wield_delay to Weapon_delay
/:cl:

### DIFF
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -135,15 +135,14 @@
 #define SLOWDOWN_ADS_MINISCOPE_DYNAMIC 1.8
 #define SLOWDOWN_ADS_SUPERWEAPON 2.75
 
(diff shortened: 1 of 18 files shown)


Pugie:
This breaks wield speeds. They're now the same for every single gun.
At least I think so, I haven't looked any further but this seems the most likely suspect.

MistChristmas:
> This breaks wield speeds. They're now the same for every single gun. At least I think so, I haven't looked any further but this seems the most likely suspect.

This PR isn't even tested, so no it's not this one, and the answer is in maint-announcements on the discord

realforest2001 (approved):

PR #12414, opened 2026-06-10
Fix: Bitfield2list limiting the flags available to be seen on the minimap.

MistChristmas:
# About the pull request
Uhhhh it was capped at 1<<16 and I've raised it to 1<<23.
~~...This kinda breaks your local if you go over it for reasons that are probably very concerning?~~
It can't go up to 1<<24 due to byond limitations. Though I have no idea how I encountered issues below that previously.

...Well uhhhhh it covers all the flags we currently have + pathogen. So uhhhhh yay?
Maybe we should kill off some of these flags.
<details>

</details>

# Changelog
:cl:
fix: Some of the larger minimap bitflags can now be seen on the minimap. This just means Hives that have the bigger number in the code can be viewed on the minimap like normal.
/:cl:


code/__DEFINES/minimap.dm
```diff
@@ -20,6 +20,8 @@
 #define MINIMAP_FLAG_XENO_K_SERIES (1<<19)
 #define MINIMAP_FLAG_ALL (1<<20) - 1
 #define MINIMAP_FLAG_ALL_XENOS (MINIMAP_FLAG_XENO|MINIMAP_FLAG_XENO_CORRUPTED|MINIMAP_FLAG_XENO_ALPHA|MINIMAP_FLAG_XENO_BRAVO|MINIMAP_FLAG_XENO_CHARLIE|MINIMAP_FLAG_XENO_DELTA|MINIMAP_FLAG_XENO_FERAL|MINIMAP_FLAG_XENO_TAMED|MINIMAP_FLAG_XENO_MUTATED|MINIMAP_FLAG_XENO_FORSAKEN|MINIMAP_FLAG_XENO_RENEGADE|MINIMAP_FLAG_XENO_HUNTED)
+//Due to limitations specifically in bitfield2list don't exceed 1<<22
```
Drulikar:
```suggestion
// NOTE: Don't add flags past 1<<23, it'll break things due to BYOND limitations.
```

code/__HELPERS/lists.dm
```diff
@@ -408,7 +408,7 @@
 				r += wordlist[i]
 			bit = bit << 1
 	else
-		for(var/bit=1, bit<=65535, bit = bit << 1)
+		for(var/bit=1, bit<=4194303, bit = bit << 1)
 			if(bitfield & bit)
 				r += bit
```
Drulikar:
```js
///Converts a bitfield to a list of numbers (up to the byond limit of 1 << 23)
/proc/bitfield2list(bitfield = 0)
	var/list/result = list()
	for(var/i = 0, i < 24, i++)
		var/bit = 1 << i
		if(bitfield & bit)
			result += bit
	return result
```
I dunno whats going on with the wordlist stuff, but that was still limited to 16, don't understand the point of it, and nothing uses it...

Drulikar (changes requested):

code/__HELPERS/lists.dm
Drulikar: Also description should be updated up to 23

code/__HELPERS/lists.dm
```diff
@@ -397,22 +397,14 @@
...
-			if(bitfield & bit)
-				r += wordlist[i]
-			bit = bit << 1
-	else
-		for(var/bit=1, bit<=65535, bit = bit << 1)
-			if(bitfield & bit)
-				r += bit
-
-	return r
+//Converts a bitfield to a list of numbers
```
Drulikar:
```suggestion
///Converts a bitfield to a list of numbers of the bits that are set
```

Drulikar (approved):

PR #12417, opened 2026-06-10
maybe fix pharm phys showing as misc dept

### LINKED ISSUE #12405: Pharmaceutical Physician appears under Misc tab on the Manifest
## Testmerges
<a href="https://github.com/cmss13-devs/cmss13/pull/10999">#10999: '[DNM] Fix of various plane related issues' by Blundir at commit f85a7153c4</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/11075">#11075: 'Ticket Panel (Adminhelps &amp; Mentorhelps)' by GeneriedJenelle at commit 4ad8d0cf87</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/11902">#11902: 'implement pii scrubbing for sentry' by hry-gh at commit 759efb7bd1</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/12174">#12174: 'Yautja Armor Sets' by realforest2001 at commit 90c66798ed</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/10414">#10414: 'Port: Harm Intent uses Auto-Fire instead of PB' by MistChristmas at commit c70b0ced36</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/10541">#10541: '[TM v.1.2] Warrior &#34;Bulwark&#34; Strain. (Feedback link in PR description)' by Venuska1117 at commit 5eb1fcac03</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/11161">#11161: 'Gas trap buff &amp; neuro smoke tweaks' by Diegoflores31 at commit 92b29e6eb0</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/12361">#12361: 'pride scarfs in loadout' by Detective-Google at commit 6ddcf88cb8</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/10376">#10376: 'CIA Overhaul' by realforest2001 at commit cb35aca713</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/12322">#12322: 'Fort McNeil - (Lowpop) Ground Map' by Steelpoint at commit ce3292659b</a><br>

## Round ID
34642

## Description of the bug
Pharmaceutical Physician appears under Misc tab on the Manifest

Should appear under medical tab

1) Have someone join with Pharmaceutical physician doctor title
2) They appear under misc tab instead of Medical

## Attached logs
```
N/A
```

## Additional details
- Author: dangerrevolution
- Admin: the lalipar
- Note: None

kitchenenthusiast:
# About the pull request

Probably fixes #12405 by forcing the job presets to explicitly pass job_doctor to the datacore first before the equip presents will call manfest_modify and force update it to the correct job title

**this is exactly how the surgeon vanity title also does it**.  some thoughts on that below

Well, well, well.

So actually what i found out in the course of this is that the way surgeon passes the vanity title into the manifest specifically sets up great circumstances for a race condition. Did I duplicate that logic? Yes. Here's why:
1. I don't want to override completely the way the roles populate into the manifest so every doctor appears as "doctor" because having pharmacists be stripped of their vanity titles on manifest and surgeons not seems...weird. happy to force both to appear as doctor though if we want to move away from vanity titles. Maybe its more intelligible for people to see there's 4 doctors versus 2 doctors and a surgeon and a pharm
2. I also didnt want to overwrite the actual assignment in the datacore; on the backend, I thought it's probably best for doctors, surgeons, and pharms to still all appear as just doctors.
In short, there were better ways to do this; this is the only one I came up with preserving current behavior across both roles.
I did this on my phone smh. Pls provide feedback if we'd rather just cut the potential for race conditions here, it feels hack-y.

# Changelog

:cl:
fix: fixed pharmacists not showing in the medical department
/:cl:

### DIFF
--- a/code/modules/gear_presets/uscm_medical.dm
+++ b/code/modules/gear_presets/uscm_medical.dm
@@ -108,6 +108,7 @@
 /datum/equipment_preset/uscm_ship/uscm_medical/doctor/pharmacist
 	name = "USCM Pharmaceutical Physician"
 	assignment = JOB_PHARMACIST
+	job_title = JOB_DOCTOR
 
 /datum/equipment_preset/uscm_ship/uscm_medical/doctor/pharmacist/load_gear(mob/living/carbon/human/new_human)
 	var/back_item = /obj/item/storage/backpack/marine/satchel


fira:
Are you sure about that?
`uscm_medical/doctor` already has `job_title = JOB_DOCTOR` - that's why it applies to surgeon.
What does this change?

kitchenenthusiast:
to the best of my knowledge:
parent inheritance doesn't catch this since the issue in spawning pharmacists is later on during role mapping eval.
looking at `RoleAuthority` and `manifest_modify`, surgeon inherits from doctor to group correctly, then passes string lookup from equipment presets because the string macro Surgeon = the job assignment SURGEON. on some other init (id load since manifest modify is hooked to it), it gets retitled correctly on the manifest from "Doctor". it survives to this point though because the job string is the same as the assignment so it's able to pass through all the logic to get there.
the way `#define GET_MAPPED_ROLE(title) (GLOB.RoleAuthority?.role_mappings[title] ? ...` works, though, i think Pharmaceutical Physician is only implicitly inherited because the string macro is not 1:1 with job assignment = PHARMACIST and it gets dropped before the manifest mod can fix it. equipment loading on spawn independently then fired the manifest mod, so pharm got put in on the manifest, but they were no longer mapped to being a doctor. defining doctor would override the lookup and force the compiler to just call them a doctor; then on manifest mod, the name would update to be the vanity title, like it does for surgeon (hence race condition).

the role could always be renamed pharmacist but then i don't know if that gives the impression they should never do surgery. or rename the job assignment to the full title but it's kind of clunky for backend.

also i think without forcing a doc define onto the role it might be possible for it to not get bundled correctly in job bans because that's RoleAuthority based and depends on the string lookup

MistChristmas:
Surgeons don't actually get their title showing up on the Crew Manifest either. 
This PR doesn't actually do anything. It inherits the job_title from the parent just fine.

The issue you're trying to fix is when someone has their ID modified and then gets sorted to Misc cause none of the titles line up with what the Crew Manifest is expecting anymore. Because they go from Doctor to whatever custom title/Phys Pharm/Surgeon.

Additionally you didn't actually test anything in this PR. This is closed, re open when you actually address and test that you have addressed the issue you have set out to fix.

PR #12433, opened 2026-06-12
Removes rangefinders from CO survivors + adds rangefinders to CC and Tyrargo

GoodWheatley:
# About the pull request

Removes the rangefinder binoculars that two military CO survivors spawn with, and adds them back into the map at set locations. I also removed the rangefinder the Tyrargo SL survivor gets for consistency's sake.

On Chance's Claim, the rangefinder spawns in the alleyway north of the basketball court in the fitness building.

On Tyrargo, the rangefinder spawns in the northwest apartment on the upper level of the apartment complex.

Right now there's an unspoken expectation that if a CO survivor spawns with a rangefinder, they *have* to call in a supply drop for the survivors. This ends up skewing balance due to the possibility of requisitions being able to send a full crate of supplies to the survivors just because a CO decided to play, and can also lead to the interaction being gamified on all sides to the point where if you don't do it, you're throwing the round.

Removing the rangefinder from the CO survivors and adding it back as map loot solves this problem. There is still the possibility of the survivors being able to contact the Almayer and call in a supply drop, but they have to put in effort to be able to achieve that instead of gambling on a CO survivor existing. This also means that there's more emphasis placed on teamwork by having someone run to get the rangefinder.

<details>

</details>

# Changelog
:cl:
del: Removed rangefinders that spawn with survivors on Chance's Claim and Tyrargo Rift
maptweak: Added rangefinders on Chance's Claim and Tyrargo Rift. The rangefinder spawns north of Fitness on CC, and in the northwest corner of upper Apartments on Tyrargo.
/:cl:


realforest2001 (dismissed):

Detective-Google (changes requested):
the one on chances should be in a different place from the RTO, and ideally somewhere that is a bit of a loot dead zone otherwise.

Detective-Google (approved):
code, mapping

Drulikar (approved):

PR #12444, opened 2026-06-14
Fix gun_user bad index runtimes: comp_lookup is a lazy list

Drulikar:
# About the pull request

This PR does is primarily to fix a runtime where a lazy list is accessed. Rather than simply using the lazy macros to access comp_lookup from #11312 we only set the gun_user for the primary hand weapon. Also display_ammo requires a user in arguments and won't bother doing anything if the user is null.

Fixes 
image

<details>
Dropped and shot a gun a bunch; also grabbed a second gun and tinkered with alt fire settings (though I'm just ultimately looking for runtimes or something that feels off): 
image

Some debug testing to double check signal registrations:
image
</details>

# Changelog
:cl: Drathek MistChristmas
fix: Fixed a runtime with set_gun_user causing a bad index
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -2191,10 +2191,10 @@ not all weapons use normal magazines etc. load_into_chamber() itself is designed
 
 /obj/item/weapon/gun/proc/set_gun_user(mob/to_set)
 	if(to_set == gun_user)
-		if(gun_user)
-			RegisterSignal(gun_user, COMSIG_MOB_MOUSEDOWN, PROC_REF(start_fire), override=TRUE)
-			RegisterSignal(gun_user, COMSIG_MOB_MOUSEDRAG, PROC_REF(change_target), override=TRUE)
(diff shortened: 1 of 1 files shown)


fira:
This shouldn't use comp_lookup at all, it's DCS internals

Isn't override = TRUE on the signal registration an option here instead?

fira (approved):
I checked code in more detail, i don't really get why this is required to begin with, but this should be a cleaner solution at least

fira (dismissed):
Need more investigation

fira:
So according to what I understood from Mist this isn't a bandaid for bypassing the stack_trace from re-registration of signal (that you'd avoid with override = TRUE), but rather a bandaid to prevent registering handlers of another gun during alternating akimbo when one already is

Maybe this can simply be fixed with checking active guns when the signals fire. It'd be great to reproduce the original issue to know how to fix this properly, but otherwise I think it can be merged as is for now, it's Less Worse (TM) i guess...

MistChristmas (approved):

PR #12452, opened 2026-06-15
Clean up the priming subroutine for explosives

LatoDiafol:
# About the pull request

Made the prime() function in explosive.dm more streamlined and robust, this involved:
The removal of the "i" variable and replacing it with that simply counted up as the containers were iterated across the first time and counted down when the second loop then transferred the contents of those containers to the "meta-container" that is used to calculate the actual reaction. This was replaced by a call to find the length of the container array instead of counting up in the first loop, this makes the code more robust by keeping each loop independent while also making the code easier to read and comprehend.

Introduction of an error message should an invalid explosive with 0 or less containers that somehow has reagents contained within, theoretically this code should never fire, but it should catch any edge cases I've missed and aid in debugging.

The change was tested using all of the partticular test cases I could think of, the results of which can be found below.

Removes fragility from the procedure which could cause it to break in unexpected ways if modified. Removes an ambiguous variable "i" and replaces it with a usefully named variable "container_count"

Test Cases:
    No containers: Properly makes the dud sound and does not trigger volatiles.
    One container of an explosive substance: Sets the volatiles flag properly and detonates. (expected 60 units welding fuel present at the point of explosion and the "Trigger Volatiles" flag set as true, Test successful)
    Two containers of explosives: Properly combines both canisters and triggers the detonation as it should. (Two containers of 60 units welding fuel each, saw 120 units of welding fuel at the point of explosion and the Trigger Volatiles flag set as expected.)
    One container of non explosives: Produces a whiff of steam as expected, but no actual explosion (Explosive contained 60 units of Watermelon Juice)
    One Container of Explosives and one container of non explosives (One bottle of 60u Welding Fuel, One bottle of 60u Watermelon Juice, explosion triggered with identical stats to the grenade containing just 60u welding fuel as expected)
<details>

Not Applicable

</details>

# Changelog

:cl:
code: Streamlined the prime() function for explosives in explosives.dm
/:cl:


code/game/objects/items/explosives/explosive.dm
```diff
@@ -205,28 +205,28 @@
...
 	var/mob/cause_mob = cause_data?.resolve_mob()
 	if(cause_mob) //so we don't message for simulations
-		reagents.source_mob = WEAKREF(cause_mob)
+		reagents.source_mob = WEAKREF(cause_mob) //Generate the Admin log for the custom grenade detonation
 		msg_admin_niche("[key_name(cause_mob)] detonated custom explosive by [key_name(creator)]: [name] (REAGENTS: [reagent_list_text]) in [get_area(src)] [ADMIN_JMP(loc)]", loc.x, loc.y, loc.z)
 
-	if(length(containers) < 2)
-		reagents.trigger_volatiles = TRUE //Explode on the first transfer
-
+	var/container_count = length(containers) ///Find the number of containers, this will count down as the containers are added to the "meta-container" for the explosive
```
fira, line 221:
```suggestion
	var/container_count = length(containers) // Find the number of containers, this will count down as the containers are added to the "meta-container" for the explosive
```

/// is only for documenting specific procs/vars

code/game/objects/items/explosives/explosive.dm
```diff
@@ -205,28 +205,28 @@
...
 	for(var/obj/item/reagent_container/glass/G in containers)
-		G.reagents.trans_to(src, G.reagents.total_volume)
-		i--
-		if(reagents && i <= 1)
+		if (container_count <= 1) //Iff this is the last container set the trigger volatiles flag to set off the explosion when the last container is transferred, if somehow there is less then one container left, detonate anyway, but log an error as this should not be possible
 			reagents.trigger_volatiles = TRUE //So it doesn't explode before transferring the last container
+			if(container_count < 1) // This should never trigger, but if for some ungodly reason an explosive is primed with less then one container but somehow has reagents this will create an error log
+				msg_admin_niche("ERROR: [key_name(cause_mob)] threw an invalid explosive, which did did not contain a proper number of chemical containers. [key_name(creator)]: [name] (REAGENTS: [reagent_list_text]) in [get_area(src)] [ADMIN_JMP(loc)]", loc.x, loc.y, loc.z)
```
fira, line 226: Mental note this breaks after #12483

code/game/objects/items/explosives/explosive.dm
LatoDiafol: Ah I understand, thanks for letting me know

code/game/objects/items/explosives/explosive.dm
LatoDiafol: I'll take a look and see what needs  fixing.

cm13-github:

PR #12465, opened 2026-06-17
Fix: Rocket backblast direction

### LINKED ISSUE #12342: sadar backblast location is chosen before turning
### Testmerges

no

### Round ID

tested locally

### Description of the bug

backblast is done before the mob turns, so you could face south, shoot north, and have backblast come out north in the direction you are shooting

backblast comes out the back not front

1. face south
2. shoot north

### Issue Bingo

antlersss:
# About the pull request

Fixes #12342

Also fixes a small null error when checking if an area is heavily ventilated

<details>

</details>

# Changelog

:cl: Antlers
fix: Fixed a bug where rocket backblast would not be applied in the correct direction if the user turned while firing.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -1351,7 +1351,7 @@ and you're good to go.
 		if(before_fire_cancel & COMPONENT_HARD_CANCEL_GUN_BEFORE_FIRE)
 			return NONE
 
-	apply_bullet_effects(projectile_to_fire, user, reflex, dual_wield) //User can be passed as null.
+	apply_bullet_effects(projectile_to_fire, user, target, reflex, dual_wield) //User can be passed as null.
 	SEND_SIGNAL(projectile_to_fire, COMSIG_BULLET_USER_EFFECTS, user)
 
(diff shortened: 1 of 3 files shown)


code/modules/projectiles/gun.dm
```diff
@@ -1351,7 +1351,7 @@ and you're good to go.
 		if(before_fire_cancel & COMPONENT_HARD_CANCEL_GUN_BEFORE_FIRE)
 			return NONE
 
-	apply_bullet_effects(projectile_to_fire, user, reflex, dual_wield) //User can be passed as null.
+	apply_bullet_effects(projectile_to_fire, user, target, reflex, dual_wield) //User can be passed as null.
```
fira, line 1354: You need to also update the other `apply_bullet_effects` call!

--> https://github.com/antlersss/ant-cmss13/blob/9ce571b9c43306318dc1753c214b43643d640e03/code/modules/projectiles/gun.dm#L1526

fira (changes requested):

fira (approved):

PR #12476, opened 2026-06-20
Long List Lint Changes

Drulikar:
# About the pull request

This PR reworks the regex used for long list linting doing a few things:
- The closing ) no longer requires specific indentation - just for it to be on its own line
- Long list linting now applies to all multi-line list definitions (rather than avoiding lists with spaces in them??)
- Indentation now is required to be consistent for every item
- **NEW** List entries over 200 characters long are now linted to be in long list form.

However there are two things I would want to accommodate for, but I haven't been able to craft a solution for (atleast not without the use of variable length backtracing that most regex don't allow):
- A comment that ends with `)` will incorrectly be treated as the end of the list and be flagged by the lint for not having the closing paren on its own line
- ~~Any comment on the same line after the final `,` will incorrectly be flagged by the lint as missing a `,`~~
E.g. 
image

This was originally started because of issues with false positives with something like:
```js
/obj/whatever
    steps = list(
        PCWJ_ADD_ITEM(/datum/rawr),
        PCWJ_ADD_ITEM(/datum/rawr),
        PCWJ_ADD_ITEM(/datum/rawr),
    )
```

But is now supported.

See lints & changed files for files I needed to address (though I did iterate on this so not necessarily everything changed would be flagged by the current version of the regex)

# Changelog
:cl: Drathek
code: Various long lists formatted differently for new long list linting
/:cl:


hry-gh (dismissed):

fira (dismissed):
I don't like the fact you can't have a comment on the last line entry, but I spent a bit looking at it and was unable to come up with a satisfactory solution

Drulikar:
> I don't like the fact you can't have a comment on the last line entry, but I spent a bit looking at it and was unable to come up with a satisfactory solution

If this regex supported variable length backtracing (most don't) I think its accomplish-able but I dunno otherwise.
**Edit** Resolved.

Drulikar:
Stale exempt until reviewed

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

tools/ci/check_grep.sh
```diff
@@ -250,28 +250,24 @@ if [ "$pcre2_support" -eq 1 ]; then
 		st=1
 	fi;
 	part "long list formatting"
-	if $grep -PU '^(\t)[\w_]+ = list\(\n\1\t{2,}' code/**/*.dm; then
-		echo -e "${RED}ERROR: Long list overindented, should be two tabs.${NC}"
-		st=1
-	fi;
-	if $grep -PU '^(\t)[\w_]+ = list\(\n\1\S' code/**/*.dm; then
-		echo -e "${RED}ERROR: Long list underindented, should be two tabs.${NC}"
+	if $grep -PU '^(\t+)[\w_\/]+\s*=\s*list\([^\s)]+(\s*=\s*[\w\d]+)?,\n' $code_files; then
```
antlersss: This does not match if the author decides to instantiate something on the first line like so:
```
    var/list/new_list = list("first" = new thing(foo, bar),
    )
```
It also does not match if there was any whitespace between the comma and newline.

The capturing group is unused, so should be converted to non-capturing groups.
```suggestion
	if $grep -PU '^\t+[\w_\/]+\s*=\s*list\([^\s)]+(?:\s*=\s*\S+)?,\s*\n' $code_files; then
```

tools/ci/check_grep.sh
```diff
@@ -250,28 +250,24 @@ if [ "$pcre2_support" -eq 1 ]; then
...
+		echo -e "${RED}ERROR: First item in a long list should be on the next line.${NC}"
 		st=1
 	fi;
-	if $grep -PU '^(\t)[\w_]+ = list\([^\s)]+( ?= ?[\w\d]+)?,\n' code/**/*.dm; then
-		echo -e "${RED}ERROR: First item in a long list should be on the next line.${NC}"
+	if $grep -PU '^(\t+)[\w_\/]+\s*=\s*list\(.{200,}' $code_files; then
```
antlersss: Tabs do not need to be captured here.
```suggestion
	if $grep -PU '^\t+[\w_\/]+\s*=\s*list\(.{200,}' $code_files; then
```

tools/ci/check_grep.sh
```diff
@@ -250,28 +250,24 @@ if [ "$pcre2_support" -eq 1 ]; then
...
+		echo -e "${RED}ERROR: Long lists should be in long list form where the first item is on the next line.${NC}"
 		st=1
 	fi;
-	if $grep -PU '^(\t)[\w_]+ = list\(\n(\1\t\S+( ?= ?[\w\d]+)?,\n)*\1\t[^\s,)]+( ?= ?[\w\d]+)?\n' code/**/*.dm; then
-		echo -e "${RED}ERROR: Last item in a long list should still have a comma.${NC}"
+	if $grep -PU '^(\t+)[\w_\/]+\s*=\s*list\((\n\1\t(\(.*\)|[^\n\t()])+)*(\n(?!\1\t)(\(.*\)|[^\n\t()])+|\n(?!\1\t\S)\t*(\(.*\)|[^\n\t()])+)+(\n\t*(\(.*\)|[^\n\t()])+)*\s*\)\s*$' $code_files; then
```
antlersss: Unused groups should be converted to non-capturing groups for small performance gains.
```suggestion
	if $grep -PU '^(\t+)[\w_\/]+\s*=\s*list\((?:\n\1\t(?:\(.*\)|[^\n\t()])+)*(?:\n(?!\1\t)(?:\(.*\)|[^\n\t()])+|\n(?!\1\t\S)\t*(?:\(.*\)|[^\n\t()])+)+(?:\n\t*(?:\(.*\)|[^\n\t()])+)*\s*\)\s*$' $code_files; then
```
I'm having a hard time parsing this expression as well... but it seems to work properly when testing. One thing to note is that it does not check for the proper amount of tabs before the final closing parenthesis.

tools/ci/check_grep.sh
```diff
@@ -250,28 +250,24 @@ if [ "$pcre2_support" -eq 1 ]; then
...
+		echo -e "${RED}ERROR: All items in a long list should be indented one more than the opening list line.${NC}"
 		st=1
 	fi;
-	if $grep -PU '^(\t)[\w_]+ = list\(\n(\1\t[^\s)]+( ?= ?[\w\d]+)?,\n)*\1\t[^\s)]+( ?= ?[\w\d]+)?\)' code/**/*.dm; then
-		echo -e "${RED}ERROR: The ) in a long list should be on a new line.${NC}"
+	if $grep -PU '/^(\t+)[\w_\/]+\s*=\s*list\((?:\n\1\t.+)*\n\1\t(?![^\n]*,(?:\s*(?:\/\/.*|\/\*(?:[^*]|\*[^\/])*\*\/))*\s*$)[^\n]+(\n\s*)\)' $code_files; then
```
antlersss: Character before the BEGIN character was causing this to never match on my tester. Additionally, requiring the end of the string to be a part of the negative lookahead means that valid comma-having last entries were being discarded, causing false positives.
```suggestion
	if $grep -PU '^(\t+)[\w_\/]+\s*=\s*list\((?:\n\1\t.+)*\n\1\t(?![^\n]*,(?:\s*(?:\/\/.*|\/\*(?:[^*]|\*[^\/])*\*\/))*)[^\n]+\n\s*\)' $code_files; then
```

tools/ci/check_grep.sh
```diff
@@ -250,28 +250,24 @@ if [ "$pcre2_support" -eq 1 ]; then
...
+		echo -e "${RED}ERROR: Last item in a long list should end with a comma.${NC}"
 		st=1
 	fi;
-	if $grep -PU '^(\t)[\w_]+ = list\(\n(\1\t[^\s)]+( ?= ?[\w\d]+)?,\n)+\1\t\)' code/**/*.dm; then
-		echo -e "${RED}ERROR: The ) in a long list should match identation of the opening list line.${NC}"
+	if $grep -PU '^(\t+)[\w_\/]+\s*=\s*list\((\n\1\t.+)+\)\s*$' $code_files; then
```
antlersss: Fails to match when a list is formatted with linebreaks in between like so:
```
	var/list/new_list = list(
		foo,
		bar,
		
		third,
		fourth,) 
```
```suggestion
	if $grep -PU '^(\t+)[\w_\/]+\s*=\s*list\((?:\n\1\t(?:.+|(?!\))))+\)\s*$' $code_files; then
```

tools/ci/check_grep.sh
Drulikar: Thanks for pointing it out, tho your suggestion doesn't match it either because theres whitespace and you are matching only non-whitespace after the equals. But maybe `^\t+[\w_\/]+\s*=\s*list\([^\s)]+(?:\s*=.+)?,\n`

tools/ci/check_grep.sh
Drulikar: Each regex is a pass; no single regex is trying to accomplish everything. I don't think it necessary that this one checks what you're noting. This pass is only supposed to care about the last item having a comma or not.

tools/ci/check_grep.sh
Drulikar: Good catch however your version then doesn't catch your code w/ no tabs on that line and normally prettier rules or whatever in VSC will remove that whitespace so it will generally go to case that it doesn't catch.

But maybe `^(\t+)[\w_\/]+\s*=\s*list\((?:\n\1\t.+|\n\s*$)+\)\s*(?:\/[\/*].*)?$`

tools/ci/check_grep.sh
Drulikar: I have no idea how that `/` got in there before `^` but meh.

Could you provide what false positive examples you were getting however?

tools/ci/check_grep.sh
Drulikar: ATM I dunno how I can catch something like (of note we explicitly are looking for tabs since spaces will be checked in another pass)
```js
    var/list/new_list = list("first" = new thing(foo, bar),
    )
```
but not catch something like
```js
    var/static/list/icons = list(
        FA_ICON_LEMON = list("value" = 2, "colour" = "yellow"),
        FA_ICON_DOLLAR_SIGN = list("value" = 2, "colour" = "green"),
    )
```

antlersss: I'll see if I can throw something together. I'll get back to you on this.

PR #12481, opened 2026-06-20
Turns the Grenade Satchel into a Belt, plus related changes

TheManWithNoHands:
# About the pull request

* Adds a grenade belt that can be ID locked with the same stats as the Grenade Satchel
* Moves the code for id lock from backpack to its parent storage, allowing storage items to easily gain an ID lock using a 1 var change.
* Adds a new belt sprite for the lockable belt.
* Cleand up some redundant var's from the UPP grenade belt.

The Grenade Satchel has no real reason to use the backpack slot. It holds about the same in grenades, and while it can hold bigger grenades we just dont have bigger grenades. Its only real diffrence is the ability to lock it. Wich is not worth the backpack slot. So on the belt it goes like its non lockable cousin. This should make the satchel into more then a box sitting at FOB to store your private stash.

How i did that i might have wasted my time by doing a small refactor kinda. (Idunno if that counts)
To not repeat code by having the same id lock code in bags and belts, i put it on their parent. /obj/item/storage.
I think this is the better option, being now you can slap the ID lock function on any storage with 1 var change, but i will undo that and just put it on belts if thats preferd.
I dont think its breaking any item either, i couldnt find any storage item that is messed up because of it.

<details>

grafik

grafik

</details>

# Changelog
:cl:NHC
add: Added an ID lockable Grenade Belt
code: Moved the ID lock code from backpacks to its parent, storage
del: Removed the Grenade Satchel
imageadd: added Item and on_mob sprites for the new grenade belt
/:cl:


kooarbiter:
does the grenade qol also work for the regular grenade belts riflemen can buy?

TheManWithNoHands:
> does the grenade qol also work for the regular grenade belts riflemen can buy?

Its for ALL grenade belts. All.. like 3 or so.

code/game/objects/items/storage/belt.dm
```diff
@@ -1336,11 +1329,27 @@
 	new /obj/item/explosive/grenade/high_explosive/upp(src)
 
 /obj/item/storage/belt/grenade/upp/attackby(obj/item/attacked_item, mob/user)
-	if(istype(attacked_item, /obj/item/storage/box/nade_box) || istype(attacked_item, /obj/item/storage/backpack/marine/grenadepack))
+	if(istype(attacked_item, /obj/item/storage/box/nade_box) || istype(attacked_item, /obj/item/storage/box/packet) ||istype(attacked_item, /obj/item/storage/belt/grenade/grenadepack))
 		dump_into(attacked_item, user)
 	else
 		return ..()
```
fira: This all looks redundant to the parent proc unless i'm missing something

code/game/objects/items/storage/belt.dm
```diff
@@ -1336,11 +1329,27 @@
...
+	max_storage_space = 36 //12 grenades
+	storage_slots = 12
+	is_id_lockable = TRUE
+	flags_atom = FPRINT|NO_GAMEMODE_SKIN // same sprite for all gamemodes
+
+/obj/item/storage/belt/grenade/grenadepack/attackby(obj/item/W, mob/user)
+	if(istype(W, /obj/item/storage/box/nade_box) || istype(W, /obj/item/storage/box/packet) || istype(W, /obj/item/storage/belt/grenade/grenadepack) || istype(W, /obj/item/storage/belt/grenade))
+		dump_into(W,user)
+	else
+		return ..()
```
fira: same

code/game/objects/items/storage/belt.dm
```diff
@@ -1270,7 +1270,7 @@
 	new /obj/item/explosive/grenade/high_explosive/airburst(src)
 
 /obj/item/storage/belt/grenade/attackby(obj/item/W, mob/user)
-	if(istype(W, /obj/item/storage/box/nade_box) || istype(W, /obj/item/storage/backpack/marine/grenadepack))
+	if(istype(W, /obj/item/storage/box/nade_box) || istype(W, /obj/item/storage/box/packet) || istype(W, /obj/item/storage/belt/grenade/grenadepack))
```
fira: Why just the /grenadepack and not all belts? Just a question

code/game/objects/items/storage/storage.dm
```diff
@@ -1028,3 +1046,46 @@ Returns FALSE if no top level turf (a loc was null somewhere, or a non-turf atom
 	if(src && ishuman(usr))
 		instant_pill_grab_mode = !instant_pill_grab_mode
 		to_chat(usr, SPAN_NOTICE("You will now [instant_pill_grab_mode ? "take pills directly from bottles": "no longer take pills directly from bottles"]."))
+
+/obj/item/storage/proc/toggle_lock(obj/item/card/id/card, mob/living/carbon/human/H)
+	if(QDELETED(locking_id))
+		to_chat(H, SPAN_NOTICE("You lock \the [src]!"))
+		locking_id = card
```
fira, line 1054: This needs to clear the `locking_id` reference properly

Either store a `WEAKREF()` or add it to `Destroy` on `/obj/item/storage` then attach a `COMSIG_PARENT_QDELETED` handler to it

fira (changes requested):

code/game/objects/items/storage/belt.dm
TheManWithNoHands: parent doesnt allow the belt. Was before like this too and i wasnt sure if i should change that bit. I take this as its fine to change.

Do am noticing there is grenade belt twice in these though. Wupps.

code/game/objects/items/storage/belt.dm
TheManWithNoHands: Can make it to allow all grenade belts, sure.

code/game/objects/items/storage/storage.dm
TheManWithNoHands: Aight. Then i'll sit down and learn how those things are used.
I think i did it right?

fira: Here's what i suggest specifically: Use the WEAKREF() method. Since this is core storage object code, using signals can conflict with other storages use of it.
A weakref allows you to keep track of something without holding it up from being deleted. 

 * Instead of defining `locking_id` as /obj/item/card/id above, you use a /datum/weakref
 * When you store it, you use WEAKREF : `locking_id = WEAKREF(card)` 
 * When you want to check it, you "resolve" it into the real item: `var/obj/item/card/id/card = locking_id?.resolve()`
 * If the result is null, that means the id is gone. That can also replace your `QDELETED` check just above this.
 * In Destroy() you only need to `= null` the weakref, not `QDEL_NULL`

If you have more questions about it don't hesitate to reply here or hit me up on discord

TheManWithNoHands:
removed qol changes with grenade packs for atomization

Detective-Google:
What is the incentive to take the normal old grenade belt instead of this one? If the only difference is that one can be locked and the other can't I don't think that really justifies it's inclusion as a separate item.

TheManWithNoHands:
> What is the incentive to take the normal old grenade belt instead of this one? If the only difference is that one can be locked and the other can't I don't think that really justifies it's inclusion as a separate item.

There aint no incentive really. Because there is not supposed to be. 
This is simply a fix for the grenade satchels problem of being a grenade belt stat wise, but letting the player pay for the ID lock with a backpackslot. Thats real bad.

Grenade satchel doesnt need buffs, is already the grenade belt stat wise, so turning it into a belt is the solution that makes the most sense.
I also dont intent to make this a viable sidegrade to the belt here.
All this PR is really, is to make the grenade satchel no longer a trap for new players. And thats it.

code/game/objects/items/storage/storage.dm
TheManWithNoHands: Turns out harry did this already 3 years ago, under code>datums>components>id_lock.dm
It even seems to even add it one parent up in Items.dm 

Will have to do this diffrently and use the ID lock already in code.

TheManWithNoHands:
Converting to draft for now.
Will need to remove the ID lock i added and further integrade the ID lock already present.

TheManWithNoHands:
Closing this for now. Will take a bit for me to work on redoing this. No point keeping it open even as draft.

PR #12485, opened 2026-06-21
Bone Spur buff and fixes

TheManWithNoHands:
# About the pull request

* Increased runners bonechip projectile speed from tier 1 to tier 2. Still slower then most xeno projectiles that sit at tier 3.
* Gave Bone Spur the SFX from Hedghogs spike. Now both their shoot spikes ability share the same SFX. (it had non before)
* Cleaned up code a bit by making runners bone chips a subtype of bonechips, from being a subtype of bonechips subtype. Also made the subtypes naming cleaner.

Burn Spur sucks to use simply by how slow the projectiles speed is. It feels unresponsive, and kinda uselss causing it to easily get ignored. Having a slow moving skill shot is fine, but not to the point where its annoying to use.
So lets bump its speed up by 1 tier to feel finally nice to use. This should have little impact on actual game balance, more so just make the ability get used more.

Besides the speed buff the ability now has actual SFX so you got better feedback that you used an ability. It simply reuses the SFX Hedghogs Fire Spikes ability has, giving the evo line some more nice consitency.
Also the code was a bit weird. 
The bonechips for this ability were a subtype of a subtybe of normal bonechips, specificly the aditional ones meant to scatter, gaining nothing through it. Thats nonsense, and will only cause problems when someone wants to adjust the above subtype. The subtypes naming was also cleaned up to be less confusing.

<details>

</details>

# Changelog
:cl:NHC
balance: Doubled the projectile speed from Tier 1 to 2
code: Cleaned up Bone Spur code
fix: Added a missing SFX to Bone Spur
/:cl:


TheManWithNoHands:
removed dmg debuf and shrapnel imbed chance buff, in response to player feedback

PR #12532, opened 2026-06-26
Remove SSmove_manager references from STANDARDS.md

fira:
We just don't have it


Drulikar (approved):
Our code is using `walk_away` and `walk_to` however.

PR #11826, opened 2026-02-26
Port the area contents system from TGstation

noelle-lavenza:
# About the pull request
This is an experimental optimization to things that loop over areas. We evidently have a lot of those. This is probably also the cause of the occasional overtime in `add_base_lighting()`.

This should have no user-facing changes.

# Explain why it's good for the game

So, a little rundown on how `area.contents` works in BYOND:

**It doesn't exist.**

No, really. It's not like, tracked anywhere. If you access `area.contents`, it's created on the fly. How? By looping over ***every turf in the world*** and checking its area, then adding it to the list. This is faster than if we did that manually, because engine magic, but it's still not ideal.

One workaround for this is to make the `in world` loop explicit. If you're checking the contents of multiple areas, it would be more efficient to do one `in world` loop and check if the area is in a list; this goes from many `in world` loops to just one. Good, but annoying, and still not ideal.

The ideal way to do this would be for BYOND to just track area turfs internally. It doesn't, but TGstation coders wrote their own system for handling that, which has some optimizations already like being split up by Z-level and doing batched removals rather than removing them as they come in.

# Testing Photographs and Procedure
I tested vehicle interiors and the tutorial. They were pitch-black or otherwise broken initially, and then I fixed that and they worked. Hooray. I also played through the entirety of the marine tutorials, as well as 8 waves of the corpsman sandbox and 16 of the req line. There's probably a lot that needs to be tested (like hijack, etc) but wow that's going to suck.
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl: MoondancerPony
code: made the game use TGstation's area contents system, which is faster
/:cl:

PR #12567, opened 2026-07-02
Optimize /datum/datacore/proc/manifest

noelle-lavenza:
# About the pull request
/datum/datacore/proc/manifest now uses an assoc lookup for valid job titles to make the looping step much faster.

Removes the sleep from manifest injection. It's entirely unnecessary, see here: https://github.com/tgstation/tgstation/pull/20038

No longer will there be an arbitrary 4 second pause before the manifest loads.

Will take screenshots once I'm done PRing all the commits I've split out of #11826. I believe it was testmerged and no one reported issues with the manifest, though.

<details>

</details>

# Changelog

:cl:
code: removed an unnecessary pause before crew manifest generation
/:cl:

### DIFF
--- a/code/datums/datacore.dm
+++ b/code/datums/datacore.dm
@@ -9,22 +9,20 @@ GLOBAL_DATUM_INIT(data_core, /datum/datacore, new)
 	var/leveled_riflemen = 0
 	var/leveled_riflemen_max = 7
 
-/datum/datacore/proc/manifest(nosleep = 0)
-	spawn()
-		if(!nosleep)
-			sleep(40)
+/datum/datacore/proc/manifest()
+	var/alist/lookup = alist()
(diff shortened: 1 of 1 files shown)


fira:
I'm pretty sure last time i tried to shoot down a sleep involving datacore on roundstart, it resulted in a race condition. Job/Spawn code is an absolute mess and can't be relied upon.
I'd advocate for live testing this

noelle-lavenza:
> I'm pretty sure last time i tried to shoot down a sleep involving datacore on roundstart, it resulted in a race condition. Job/Spawn code is an absolute mess and can't be relied upon. I'd advocate for live testing this

theres a very real possibility that the whole thing needs a once-over, i think the biggest risk is that it works without the sleep on other codebases because they removed even more ones, or made something faster, or dont have as much pop, etc

PR #12587, opened 2026-07-05
Changes some logging from key -> ckey

fira:
# About the pull request

Switches around key for ckey in fingerprints and some logging interactions.

We need ckey in logs for two reasons:
 - It's what's used in 90% of places, and you can't properly filter if there are several forms of it
 - It's easier and more consistent for parsing logs

Right now it turns out SAY logs use key not ckey so you can't search them by ckey for example

Doesn't touch adminhelp logs because they're not actual logs (just sent to redis/discord)

Also fixes an invalid emote log and one runtimey fingerprint usage
Removes a weird logging clause in whisper. There's no need for spawn, and the person can't be dead when using it anyway. So now it always logs.

Untested, no way i can manage to check every single one

# Changelog
:cl:
admin: Most logging (and fingerprints) should now properly use ckey, instead of key
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -300,25 +300,25 @@ directive is properly returned.
 	return
 
 /atom/proc/add_hiddenprint(mob/living/M)
-	if(!M || QDELETED(M) || !M.key || !(flags_atom  & FPRINT) || fingerprintslast == M.key)
+	if(!M || QDELETED(M) || !M.logging_ckey || !(flags_atom  & FPRINT) || fingerprintslast == M.logging_ckey)
 		return
 	if (ishuman(M))
 		var/mob/living/carbon/human/H = M
 		if (H.gloves)
-			fingerprintshidden += "\[[time_stamp()]\] (Wearing gloves). Real name: [H.real_name], Key: [H.key]"
+			fingerprintshidden += "\[[time_stamp()]\] (Wearing gloves). Real name: [H.real_name], CKEY: [H.logging_ckey]"
 		else
-			fingerprintshidden += "\[[time_stamp()]\] Real name: [H.real_name], Key: [H.key]"
+			fingerprintshidden += "\[[time_stamp()]\] Real name: [H.real_name], CKEY: [H.logging_ckey]"
 	else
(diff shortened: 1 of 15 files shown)


hry-gh (dismissed):

Drulikar:
You may want to also update the key usage from https://github.com/cmss13-devs/cmss13/pull/11075

fira:
Looks like pretty much all ckey in there

Drulikar:
> Looks like pretty much all ckey in there

Every single use of log_mhelp uses .key for example https://github.com/search?q=repo%3Acmss13-devs%2Fcmss13%20log_mhelp&type=code and yes `author_key` is itself a `.key` too.

fira:
Oh damnit. I'll have to fix that yeah
~~I'll just PR it separately i think, simpler~~

Drulikar:
> Oh damnit. I'll have to fix that yeah ~I'll just PR it separately i think, simpler~

Just poke me on discord if you add it in this pr and I can look.

cm13-github:

cm13-github:

code/game/atoms.dm
Drulikar, line 1: Why is this file using `logging_ckey` now instead of `ckey` or `persistent_ckey`

code/modules/projectiles/guns/specialist/launcher/grenade_launcher.dm
```diff
@@ -200,7 +200,7 @@
 			if(ishuman(user))
 				var/mob/living/carbon/human/human_user = user
 				human_user.remember_dropped_object(fired)
-				fired.fingerprintslast = key_name(user)
+				fired.fingerprintslast = user.logging_ckey
```
Drulikar: Why is this using `logging_ckey` now instead of `ckey` or `persistent_ckey`

Drulikar (changes requested):

code/game/atoms.dm
fira: I just wanted fingerprints to work even if the user goes offline but maybe that's not useful

Drulikar: Then use `persistent_ckey`. As I showed you `logging_ckey` is cleared on logout.

Drulikar (approved):

PR #12601, opened 2026-07-07
[DNR] Xeno content pack: Buffs, Nerfs, Fixes and QoL (Check PR for Atomization progress)

Venuska1117:
# This PR will be closed when most atomization will be done, i need it to stay in pull request to keep track on changes that i could miss.

Atomization in process:

- [x] #12607
- [ ] #12613
- [ ] #12616
- [x] #12617
- [ ] #12618
- [ ] #12628
- [ ] #12630
- [x] #12631
- [ ] #12632
- [x] #12634
- [ ] #12636
- [x] #12737
- [ ] #12684
- [ ] #12733
- [ ] #12734
- [x] #12735
- [ ] #12736
- [ ] #12872
- [x] #12873
- [x] #12874
- [x] #12875
- [x] #12876
- [ ] #12877
- [x] #12878

PR #12617, opened 2026-07-10
Praetorian Acid Grenade changes

Venuska1117:
# Feedback: 
https://discord.com/channels/150315577943130112/1464087612796174501

Detached from: https://github.com/cmss13-devs/cmss13/pull/12601

# About the pull request

Increases acid grenade speed by one tier and changes launch_type to "HIGH_LAUNCH"

Prae acid grenades are at wierd spot, they are partially dependant on RNG mechanic, but this PR is not about RNG issue, it is more about "time dilation" that you can experience in game, when throwing grenade it sometimes can travel slower than observed on private instance (most of time when server start to lag or slow down by some code issues), and in general this skill is not as fast as it should be.

With this change acid grenades travel *slightly* faster and can travel over obstacles such as handrails, racks or closets, of course in exchange for speed increase and better throwing of acid grenades your fuze time of acid grenade is increased from 1s to 1.25s

# Testing Videos

<details>
<summary>▶️ Acid grenade BEFORE changes</summary>

https://github.com/user-attachments/assets/b2030d2d-e885-4223-9935-3d1ffb8f9750

</details>

<details>
<summary>▶️ Acid grenade AFTER changes</summary>

https://github.com/user-attachments/assets/2b3b90b9-4b6d-4e6f-929e-dd725b2124f5

</details>

# Changelog

:cl: Venuska1117
balance: Praetorian acid grenade got increased throwing speed and launch_type changed to "HIGH_LAUNCH", allowing it to throw over more obstacles such as railings or closets, but in exchange "explosion" time is increased from 1s to 1.25s
fix: Fixed cades that got unwired not getting back their pass_flags.
/:cl:


Drulikar (approved):

Drulikar (changes requested):
Needs to not go over blocking cades

cm13-github:

cm13-github:

code/game/objects/structures/barricade/barricade.dm
```diff
@@ -268,8 +268,8 @@
 				update_health(-50)
 				can_wire = FALSE
 				is_wired = TRUE
-				flags_can_pass_front_temp &= ~PASS_OVER_THROW_MOB
-				flags_can_pass_behind_temp &= ~PASS_OVER_THROW_MOB
+				flags_can_pass_front_temp &= ~PASS_OVER_THROW_MOB|PASS_ACID_GRENADE
+				flags_can_pass_behind_temp &= ~PASS_OVER_THROW_MOB|PASS_ACID_GRENADE
```
Drulikar: You need to be careful about operator order

~A|B is different than
~(A|B)

See https://ref.dm-lang.org/operator/
> Each line has higher order of operations than the next. Operators within a line have equal precedence and therefore are processed from left to right as they occur in an expression. (Assignment, or operate-and-assign, are processed from right to left.)

Image

```suggestion
				flags_can_pass_front_temp &= ~(PASS_OVER_THROW_MOB|PASS_ACID_GRENADE)
				flags_can_pass_behind_temp &= ~(PASS_OVER_THROW_MOB|PASS_ACID_GRENADE)
```

The suggested code above would be: Disable both `PASS_OVER_THROW_MOB` and `PASS_ACID_GRENADE` for both `flags_can_pass_front_temp` and `flags_can_pass_behind_temp`

code/game/objects/structures/barricade/barricade.dm
```diff
@@ -289,8 +289,8 @@
 				update_health(50)
 				can_wire = TRUE
 				is_wired = FALSE
-				flags_can_pass_front_temp &= ~PASS_OVER_THROW_MOB
-				flags_can_pass_behind_temp &= ~PASS_OVER_THROW_MOB
+				flags_can_pass_front_temp |= ~PASS_OVER_THROW_MOB|PASS_ACID_GRENADE
+				flags_can_pass_behind_temp |= ~PASS_OVER_THROW_MOB|PASS_ACID_GRENADE
```
Drulikar: Maybe you meant this?

```suggestion
				flags_can_pass_front_temp |= PASS_OVER_THROW_MOB|PASS_ACID_GRENADE
				flags_can_pass_behind_temp |= PASS_OVER_THROW_MOB|PASS_ACID_GRENADE
```
But if so, you probably intended to also change `/obj/structure/barricade/initialize_pass_flags` (it lacks having PASS_ACID_GRENADE there by default?)

As you have it coded it means to both `flags_can_pass_front_temp` and `flags_can_pass_behind_temp`:
- Enable every flag that is not `PASS_OVER_THROW_MOB`
- Also enable `PASS_ACID_GRENADE` (already included in above)

Suggested code above would be: Enable the flag `PASS_OVER_THROW_MOB` and `PASS_ACID_GRENADE` for both `flags_can_pass_front_temp` and `flags_can_pass_behind_temp`

Drulikar (changes requested):

code/game/objects/structures/barricade/barricade.dm
Venuska1117: i put PASS_ACID_GRENADE inside PASS_OVER, PASS_OVER Is already in initialize_pass_flags so no need to add it twice, unless adding PASS_ACID_GRENADE inside PASS_OVER was a *bad* idea.

code/game/objects/structures/barricade/barricade.dm
```diff
@@ -289,8 +289,8 @@
 				update_health(50)
 				can_wire = TRUE
 				is_wired = FALSE
-				flags_can_pass_front_temp &= ~PASS_OVER_THROW_MOB
-				flags_can_pass_behind_temp &= ~PASS_OVER_THROW_MOB
+				flags_can_pass_front_temp |= PASS_OVER_THROW_MOB|PASS_ACID_GRENADE
+				flags_can_pass_behind_temp |= PASS_OVER_THROW_MOB|PASS_ACID_GRENADE
```
Drulikar, line 293: Still funky IMO that temps will have only `PASS_OVER_THROW_MOB`
https://github.com/cmss13-devs/cmss13/blob/c890d89baad0e871af2bfc9729a599bbc04a427f/code/game/objects/structures/barricade/barricade.dm#L80-L87

But after unwiring the temps will have `PASS_OVER_THROW_MOB|PASS_ACID_GRENADE` but it ultimately doesn't change behavior because yes you have it in PASS_OVER

Drulikar (approved):

PR #12628, opened 2026-07-12
[REFACTOR] Updates Lurker code to use new defines and traits.

Venuska1117:
Detached from: #12601

# About the pull request

Replaced one var with trait for easier references.

Easier to utilize code in future.

# Testing Procedure

Made new code, used invisibility, ability works correctly.
Used assasinate, assasinate works correctly.

Code works as intended.

# Changelog

:cl: Venuska1117
refactor: Updates Lurker code with new define and replaced some vars with traits.
/:cl:


Venuska1117:
I have text file for pre-changes and current-changes for lurker for better code visibility and easier review, ping me on discord if you are interested in files that can be used on "text compare" website

fira (changes requested):
It's AssasSinate, with 4 S

code/modules/mob/living/carbon/xenomorph/castes/Lurker.dm
```diff
@@ -71,28 +71,18 @@
 /datum/behavior_delegate/lurker_base
 	name = "Base Lurker Behavior Delegate"
 
-	// Config
-	var/invis_recharge_time = 20 SECONDS
```
Red-byte3D, line 75: How is this meant to makea nything easier, when it makes var editing harder?

Venuska1117: If you plan making caste in future that uses invisibility, you dont need to define same vars in new behavior delegate, instead its now bound to ability, so it you give it to different caste, it will work, could also be used by admins to add abilities to different xeno without need to make a PR.

fira (dismissed):

code/modules/mob/living/carbon/xenomorph/castes/Lurker.dm
```diff
@@ -111,85 +93,66 @@
...
-		lurker_pounce_action.freeze_self = FALSE
-	bound_xeno.stealth = FALSE
-	REMOVE_TRAIT(bound_xeno, TRAIT_CLOAKED, TRAIT_SOURCE_ABILITY("cloak"))
-	UnregisterSignal(bound_xeno, COMSIG_MOB_EFFECT_CLOAK_CANCEL)
-	invis_start_time = -1
-
 /datum/behavior_delegate/lurker_base/append_to_stat()
 	. = list()
 
+	var/datum/action/xeno_action/onclick/lurker_invisibility/lurker_inv = get_action(bound_xeno, /datum/action/xeno_action/onclick/lurker_invisibility)
```
Drulikar: Searching through every action, to grab an action, every 0.4s is especially gross. I know old code still did this, but it really needs not to. This information should be available to the behavior (and just keeping a reference to the action is likely not a good idea)

code/modules/mob/living/carbon/xenomorph/castes/Lurker.dm
```diff
@@ -67,40 +68,21 @@
...
-	item_state = "heart_t2"
-	research_value = 2000
-
-	xeno_organ_flags = XENO_ORGAN_STRONG|XENO_ORGAN_TACHYCARDIA
-
-
 /datum/behavior_delegate/lurker_base
 	name = "Base Lurker Behavior Delegate"
 
-	// Config
```
Drulikar, line 82: Moving things off of behaviors onto actions is generally not an improvement. But I dunno; having to get_action everything instead just feels a lot worse to me.

Drulikar (changes requested):

PR #9575, opened 2025-06-12
Oppressor Praetorian: Fixes abduction ability interaction with objects.

Venuska1117:
I am not sure if this should be QoL or Balance change, so i keep it balance because it allows you to do thing that was not possible before because of wonky code.

# About the pull request
Allows Abduction ability to be casted at same tile as barricade, flipped tables and glass window panes (as long as they don't face ability casting direction), and also allows abduction to be cast trought low obstacles that contain pass_flags, such as PASS_HIGH_OVER

#  Explain why it's good for the game
As person who enjoyed playing oppressor i found it frustrating when cade was destroyed at corner, so cade facing towards you was open, but at same tile, there was another cade facing 90 degree in other direction, so when you cast ability it gets blocked, same follows for other small obstacles, its so annoying when you want to cast through tables, trays or crates and your ability get blocked by them, this change allows for abductor main ability to be more usefull.

_With this change now you can cast ability at tiles that have barricade in them as long as they dont face casting direction, same follows for flipped tables and window panes._

# Testing Photographs and Procedure
<details>
<summary>>Click Here<</summary>

Some testing done with abduct:

https://github.com/user-attachments/assets/1b0fc84d-517b-426f-b279-e6adde07abaf

https://github.com/user-attachments/assets/d2080ed4-0a05-4b7a-8d0c-851753dcbea7

https://github.com/user-attachments/assets/26c5acae-eecf-4050-bf83-d8b8d6f2ed54

https://github.com/user-attachments/assets/af0e2779-222d-41dd-a3b0-9a59be3b91f2

</details>

# Details for Reviewers.

**Need to re-type**

<details>
<summary>opressor.dm</summary>

line 117 check if tile contains any structures, if one of checks is **FALSE** it will continue to next check.

line 162 check if structure is barricade then go to 164 and attach cade direction to "cade_facing" var.

line 165 check if **cade_facing** direction faces 180^ in opposite direction of casting, for example if you cast from west to east, if cade on east is facing west (towards you) it will then trigger line 166, where it sets **"blocked = TRUE"**, preventing ability continuation.

[...]

PR #12633, opened 2026-07-12
Oppressor's abduct goes through xeno structures

### LINKED ISSUE #12997: Oppressor Abduct blocked by Hive structure
## Testmerges
<a href="https://github.com/cmss13-devs/cmss13/pull/12824">#12824: 'Bundles Wound Overlay Updates' by fira at commit 6b48f2d5f1</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/12537">#12537: 'Atlan Station - Groundmap' by oceany-a11y at commit ed9be01957</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/12721">#12721: 'Assorted Sound Fixes' by fira at commit 8b5bf0b649</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/12618">#12618: 'Vanguard Passive Shield no longer regenerates when UNCONSCIOUS or INCAPACITATED, FLOORED, updated strain description.' by Venuska1117 at commit 813c96e27b</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/12627">#12627: 'Tyrargo Rift (2) - Various Alterations' by Steelpoint at commit dbc4496523</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/12626">#12626: 'Refurbished Combat Information Centre (2) - USS Almayer' by Steelpoint at commit 86a9778d04</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/12947">#12947: 'Revert &#34;Working multiz tacmap, round two, extra buttons&#34;' by Rain7x at commit ea384007a4</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/12989">#12989: 'Fix: Transmute' by MistChristmas at commit 5d01c7e31b</a><br><a href="https://github.com/cmss13-devs/cmss13/pull/12992">#12992: 'Fix: Not being able to cut open xenos for harvesting' by MistChristmas at commit b0e91071a7</a><br>

## Round ID
35981

## Description of the bug
Oppressors abduct gets blocked by hive clusters and possibly egg morphers/recovery nodes

The abduct should pass straight through the hive clustert

Place a Hive cluster, get an oppressor, have them abdcut diagonally so that it should pass over the cluster.

## Attached logs
```
N/A
```

## Additional details
- Author: Jester_Devotion
- Admin: Detective Google
- Note: None

JHINgle4ce:
# About the pull request
Adds a flag to all advanced xeno structures in order to satisfy a tile check in abduct ability of prae's oppressor strain.

Tbh, I don't really know full consequences of adding such a flag, so if there are some meaningful, please, tell me I'll fix.
So the problem is that hook from abduct ability isn't going through advanced xeno structures such as clusters, morphers, pylons etc. which is strange, 'cause grenades, bullets and all other stuff goes through. Also xeno structures don't look that big, high or smth else to interfere with the tail. They are not walls, barricades, etc. 

If we think about balance of the game, then only eggmorhers will create a bit of a problem, and yet its too little in the face of a hook stopping before clusters and pylons.

<details>

https://github.com/user-attachments/assets/dbe495cf-957a-4127-84cf-e587d1a1ffaa

https://github.com/user-attachments/assets/524c853a-ccc4-4230-af2f-b1ee8aaaf137

</details>

# Changelog
:cl:
fix: oppressor's hook can't go through pane glass and flipped tables diagonally the same way it did with cades
fix: oppressor's hook goes through structures that you can toss items over. Main examples: clusters, morphers
fix: oppressor's hook now checks diagonal passability fairly
/:cl:


Venuska1117:
Do not worry, before changes were added to oppressor abduct ability, it could pass through special structures such as egg morpher, cluster etc. Code changes in #9575 caused it to consider special structures as solids, even when they are not intended to be.

I will leave this to maintainers to decide.

fira:
Well, I don't know. I guess that's going to mean you can toss grenades over pylons too for example.

Venuska1117:
> Well, I don't know. I guess that's going to mean you can toss grenades over pylons too for example.

They already can do that

https://github.com/user-attachments/assets/15d0548a-3614-4a72-8a42-b68b2967249f

Unless you are worried they will be unable to do that after change.

fira:
Yeah i misunderstood how PASS_HIGH_OVER worked

fira:
I see the condition in oppressor code and it's kind of halfbaked since there's also can_pass_front and can_pass_behind 
Might be worth implementing this more thoroughly

Otherwise i can't see any problem with this but i kinda suck with the pass flags usage

JHINgle4ce:
Please!!!

boskoramen (changes requested):
Think a more durable fix here is to add a check on prae_abduct to ignore atoms that do not block movement, since that is the reason why resin structures do not block grenades/movement in general.

Code reference from movement code (https://github.com/cmss13-devs/cmss13/blob/master/code/game/turfs/turf.dm#L292):
```
if (!istype(A) || !A.can_block_movement)
```

Will prevent other issues in the future where something blocks prae abduct but not any other movement.

JHINgle4ce:
So there were the same problem as with cades. Diagonal hook went through some obj, so I added or changed a check for girders, tables and windows in addition to can_block_movement check

code/modules/mob/living/carbon/xenomorph/strains/castes/praetorian/oppressor.dm
```diff
@@ -106,15 +106,14 @@
 			var/reverse_face = GLOB.reverse_dir[facing]
 			var/turf/back_left = get_step(temp, turn(reverse_face, 45))
 			var/turf/back_right = get_step(temp, turn(reverse_face, -45))
-			if((!back_left || back_left.density) && (!back_right || back_right.density))
+			if((!back_left ||  back_left.density) && (!back_right || back_right.density) || (locate(/obj/structure/girder) in back_left && locate(/obj/structure/girder) in back_right))
```
boskoramen: The girder checks do not work because "in" has lower precedence than all other operations, so you end up doing something like
```dm
locate(/obj/structure/girder) in (back_left && locate(/obj/structure/girder)) in back_right
```

Also, why do we only check for girders here instead of all other dense structures? Think this needs a more durable solution like using `LinkBlocked()`

boskoramen (changes requested):

code/modules/mob/living/carbon/xenomorph/strains/castes/praetorian/oppressor.dm
JHINgle4ce: dont really know why its only girders myself. There were checks only for girders, border windows, flipped tables and cades

JHINgle4ce: my guess is it was intended to hook people like above computers, consoles - all that stuff, but like girders are to tall for that

JHINgle4ce: Its not like I refuse to implement  LinkBlocked(). Just a question whether it wont prohibit some interactions. For example two consoles - diagonally hook won't go, yet it will if we do it straight through the same console

boskoramen: here's the issue: non-diagonal movement is blocked for non-girders if they block movement AND they do not have the PASS_HIGH_OVER flag

we don't have this same parity with the checks for diagonal movement

so for example you can do oppressor hook between two fog tiles diagonally

since your PR is not introducing more bugs and is fixing a pretty annoying quirk with oppressor hook, I will not block approval on this diagonal stuff

also linkblocked as it is won't really meet your needs without a refactor

boskoramen: https://github.com/user-attachments/assets/38a13a5b-4452-4145-9fe8-3796852a8946

boskoramen:
Update changelog with all the fixes (mainly referring to fixes for diagonals). Will put this on TM in case there is stuff I'm missing.

Code looks good otherwise.

cm13-github:

<details><summary>Test Merge Deployed @ 09/03/2026 23:48:30 +00:00:</summary>

##### Server Instance

##### Revision
Origin: 10f5ecd4c005831e9ea313b30f743e035b4d070e
Pull Request: becbc41ebd7dab8e4560fa074708c983f340ec63
Server: 132855d9ab06031dd0fd1b6e964728c05b9d80ab
</details>
<details><summary>Test Merge Updated @ 09/04/2026 03:49:56 +00:00:</summary>

##### Server Instance
CM13 Live

##### Revision
Origin: 4a44e84a71104e2bab07805d113928d3ee8a24d8
Pull Request: becbc41ebd7dab8e4560fa074708c983f340ec63
Server: 026317c10321d089af5851c3533994db0315f4a9
</details>
<details><summary>Test Merge Removed @ 09/04/2026 18:23:30 +00:00:</summary>

##### Server Instance
CM13 Live
</details>

code/modules/mob/living/carbon/xenomorph/strains/castes/praetorian/oppressor.dm
JHINgle4ce: get it

JHINgle4ce:
So I did fair diagonal checking:
image


Also fixed another diagonal problem with cades: now it stops before cades instead of going on red line.
image

boskoramen:
Will take off TM until this is reviewed.

code/modules/mob/living/carbon/xenomorph/strains/castes/praetorian/oppressor.dm
```diff
@@ -245,6 +212,55 @@
 
 	return ..()
 
+/datum/action/xeno_action/activable/prae_abduct/proc/abduct_blocked_tile(turf/T, direction, next_direction = direction)
```
boskoramen, line 215: Instead of this proc, we should really just be refactoring LinkBlocked to support oppressor, which encompasses:
- configurable check for can_block_movement
- not requiring an atom to be passed to calculate whether link is blocked (i.e. should be able to pass in pass flags and the proc handles sending some sort of abstract moving atom)
- setting denylist for which blockers we care about (if this is necessary)

The reason why I say this is because: this proc and the way it's used follow the same premises as LinkBlocked but with checks for very specific blockers. LinkBlocked should natively support handling checks for directionals (such as barricades), diagonal-handling, etc. If there are specific things that we don't want to be checked for, that is why I suggest a specific denylist. If this and LinkBlocked, at least in the abstract, do the same thing, we should just stick with the proc already used throughout the game and extend it.

Gonna leave feedback about the code-as-is just for your future reference but I think we can do away with this and just update LinkBlocked

code/modules/mob/living/carbon/xenomorph/strains/castes/praetorian/oppressor.dm
```diff
@@ -245,6 +212,55 @@
...
+	if(!istype(T))
```
boskoramen, line 216: We shouldn't be using single letter vars, especially for new procs

code/modules/mob/living/carbon/xenomorph/strains/castes/praetorian/oppressor.dm
```diff
@@ -245,6 +212,55 @@
...
+		return 1
+
+	//0 - the way is clear, nothing blocks
+	//1 - there is an obstacle in the way
+	//2 - there will be an obstacle in the way
```
boskoramen, line 221: Instead of hard values that have comments explaining what they mean, we should be using defines for this.

code/modules/mob/living/carbon/xenomorph/abilities/praetorian/praetorian_abilities.dm
```diff
@@ -122,6 +122,7 @@
 	xeno_cooldown = 15 SECONDS
 	plasma_cost = 180
 
+	var/allow_one_more_step = FALSE
```
boskoramen, line 125: don't need to make this a state variable if we follow LinkBlocked suggestion

code/modules/mob/living/carbon/xenomorph/strains/castes/praetorian/oppressor.dm
boskoramen: The way things are setup now is a little inconsistent too, see using abduct to the NW when there is a barricade facing south and a barricade facing west

https://github.com/user-attachments/assets/1a7f0bc6-590e-4cf2-a457-add453e3c51b

boskoramen (changes requested):

PR #12654, opened 2026-07-15
Explosive Decompression: Backend

antlersss:
# About the pull request

New functionality for explosive decompressions!
Adds the following:
 - A new area subtype for handling decompression events
 - New pressure shutters that react automatically to window breaches
 - New "decompression" window subtypes that trigger an explosive decompression in their area when destroyed
 - A new testing map in `generic/decompression_test.dmm` designed to experiment with decompression events.

Getting some of this backend code stuff atomized so that the future map implementing them can be its own PR.

<details>

https://github.com/user-attachments/assets/7b3e5a50-a054-47fd-ba9c-674fe790b0bb

https://github.com/user-attachments/assets/fd9274c2-0dfa-4a42-98f1-eeded5185731

</details>

# Changelog
:cl: Antlers
code: Added backend support for explosive decompression events
/:cl:


code/game/area/decompressible.dm
```diff
@@ -0,0 +1,213 @@
+/area/decompressible
```
boskoramen, line 1: Instead of a decompressible subtype, would it be possible to tie this to a component or (if possible) a bespoke element?

Think we want to use composition over inheritance whenever possible since it decouples behavior from area's type and, if we want to ever make an existing area decompressible, makes the subsequent code change a lot smaller diff-wise since we just need to add a component.

antlersss: I thought about it, but ultimately I went with inheritance because mappers using this tool should be extremely deliberate when adding areas that have this functionality. I could see the argument for making it a component, though.

fira: I agree, this is component work or at least, flags behavior. Shuttles are currently locked to /area/shuttle in various part of the code and they would be prime candidates for decompression. Obviously the two are incompatible.

code/game/area/decompressible.dm
```diff
@@ -0,0 +1,213 @@
...
+	var/recompression_time = 1 MINUTES
+
+	/// Is this room currently decompressed?
+	var/decompressed = FALSE
+
+	/// When in gametime this room was decompressed.
+	var/time_decompressed = null
+
+	var/obj/effect/decompressed/decompression_overlay = null
+	var/list/datum/looping_sound/recompression_loop = list()
```
fira, line 15: It's unlikely to matter outside of Nightmare inserts but these need to be cleaned up in Destroy

fira: Also it would be _loops here

code/game/area/decompressible.dm
```diff
@@ -0,0 +1,213 @@
...
+
+	/// How long it takes for this room to repressurize and become habitable.
+	/// Set to -1 for rooms that are permanently decompressed.
+	var/recompression_time = 1 MINUTES
+
+	/// Is this room currently decompressed?
+	var/decompressed = FALSE
+
+	/// When in gametime this room was decompressed.
+	var/time_decompressed = null
```
fira, line 12: This doesn't seem to actually be used. Also gametime is world.time, not REALTIMEOFDAY

code/game/objects/structures/window.dm
```diff
@@ -574,6 +574,51 @@
...
+
+	. = ..()
+
+	if (!area_decompressible)
+		return
+
+	for (var/obj/structure/machinery/door/poddoor/shutters/decompression/pressure_shutters in breach_location)
+		// Shutters are already closed, this isn't an actual breach
+		if (pressure_shutters.density)
+			return
```
fira, line 606: This logic should go in your decompression code

code/game/area/decompressible.dm
```diff
@@ -0,0 +1,213 @@
...
+				throw_speed = SPEED_AVERAGE
+
+		// Throw everything not bolted down towards the breach location, and then some
+		var/list/end_throw_callbacks = list(CALLBACK(src, PROC_REF(space_object), obj_in_area, get_turf(obj_in_area), breach_location))
+		if (throw_range >= 1)
+			INVOKE_ASYNC(src, PROC_REF(fling), obj_in_area, breach_location, throw_range, throw_speed, end_throw_callbacks, null)
+
+	for (var/mob/living/carbon/victim in src)
+		// Immobilizing the victim so decompression flinging works properly
+		ADD_TRAIT(victim, TRAIT_IMMOBILIZED, TRAIT_SOURCE_DECOMPRESSION)
```
fira, line 68: Worth noting, that only prevents them from moving intentionally, it doesn't prevent say, a collision from tossing them aside

fira (changes requested):
In addition i have big concerns about the whole concept and think it's an horrible idea. Even in base SS13 if you get spaced, your body is recoverable, not straight up deleted. But i'll leave the gameplay gurus to judge on that.

fira:
I'm tagging this balance for approval due to the sheer abuse/derailing potential of it.

antlersss:
Closing this to refine it later

PR #11024, opened 2025-11-01
Announcement garble w/o coms and stars_decode_html fix

Drulikar:
# About the pull request

This PR does several things:
- Makes it so coms are required for clear announcements and overwatch from command and the garble strength depends on message length (<= 20 good, 120+ bad) and in the case of announcements the duration between announcements (30s bad, 3min+ good).  Max clarity 95, min clarity 45.
- For overwatch objectives are saved as both a garbled and ungarbled version and as soon as the marine has been with active coms will they have their objective sticky as ungarbled.
- Fixes some logic in `stars_decode_html`. 
- Removed an unused signal that seemed not useful anyways.
- Ghosts also receive garbled announcements if the Z they're observing would be garbled, but this can be opted-out using a ghost ears preference.
- Fixes remind objectives not working (wrong ui_act)
- Some misc refactoring as usual.

# Explain why it's good for the game

- Command announcements shouldn't just ignore coms are down entirely.
- More soul (IMO I guess)
- Should make coms a higher priority objective for both sides: Marines because now their command is hampered by it, and xenos because they know it hampers command.
- May encourage more usage of radio packs.

I may do something in the future to garble tacmaps too (e.g. distorted filter if coms down?).

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Wake up in cryo (this is technically off Z but now is handled): 
image

Coms down while groundside (transmitter location doesn't matter currently only receiver):
image

Coms up and freqs set:
image

Human/Ghost viewing paper: 
image

Xenos viewing paper: 
image

With recent changes to overwatch and scaling garble based on message length:
https://youtu.be/RsYdbief9Ck

With latest changes to how clarity is calculated: 
image

Logging
image

</details>

# Changelog
:cl: Drathek
balance: Command announcements and overwatch are now garbled if coms are down (applies to ghosts too based on their position). The garble strength depends on message length (<= 20 good, 120+ bad) without signature and in the case of announcements the duration between them (30s bad <-> 3min+ good). Whichever variable is better wins.
fix: Fixed some logic for marine announcements not handling people in cryopods
fix: Fixed some logic for stars_decode_html (xeno paper viewing)
fix: Fixed remind objectives in overwatch not working
add: Ghosts now have a preference to force full clarity for announcements
config: Added config for garbling settings (see game_options.txt example) and log_garble flag
/:cl:

PR #12663, opened 2026-07-16
Removes announcement garble and associated mechanics

Mary-Marilyn:
# About the pull request

This PR removes the announcement garble mechanic that was introduced in #11024 ("Announcement garble w/o coms and starsdecodehtml fix")

Comms garble was introduced under the premise that marines were ignoring the importance of holding the comms tower. This was never a problem at a scale that justified a mechanic this punitive, and a mechanic built on a problem that didn't exist has no legitimate basis for staying in the game.

Announcement garble has had a profoundly negative effect on lowpop server population. It's already a coin flip whether there's a single Squad Leader or command player online at those hours, let alone both. Garble takes that already fragile situation and makes it actively worse by punishing the few people willing to play those roles on top of everything else they're already managing alone.

Feedback has been consistently negative from everyone who actually plays with this mechanic. Staff, CO whitelist holders, (including several councilors), and long-time command players have independently reached the same conclusion that the current implementation doesn't work and no one has found a version of it that meaningfully improves gameplay. 

Multiple long-time command players and CO whitelist holders have directly said this mechanic is why they stopped playing command entirely. A system that measurably drives people away from an entire category of gameplay is actively harming the game.

No one advocating to keep garble has been able to point to a concrete gameplay improvement it delivers, the strongest defense offered has been that it's "here to stay," which isn't a valid justification. A mechanic that is universally disliked by the people forced to use it, that is driving away players, and that produces no measurable benefit has no business staying in the game out of stubbornness.

Tested marine announcements both on-ship and off-ship with no comms tower. Received full text with no scrambling.

<details>

</details>

# Changelog
:cl:
del: Removed announcement garble and all associated mechanics.
/:cl:


code/__DEFINES/__game.dm
```diff
@@ -134,8 +134,6 @@
 #define CHAT_GHOSTHIVEMIND (1<<12)
 #define CHAT_NICHELOGS (1<<13)
 #define CHAT_LISTENINGBUG (1<<14)
-#define CHAT_GHOSTANNOUNCECLARITY (1<<15)
```
fira, line 137: Keep it commented (or defined with a comment) in some fashion. This is important because these stay in savfiles. The next person that uses 1<<15 will have to do a migration to clear the flags.

fira (changes requested):
What are the hat icon changes? They don't show up on Icon DiffBot

Mary-Marilyn:
> What are the hat icon changes? They don't show up on Icon DiffBot

Dirty commit. It's already merged into master from #12602

Should have no effect whatsoever.

code/__DEFINES/__game.dm
```diff
@@ -134,8 +134,7 @@
...
-
+// (1<<15) is reserved: CHAT_GHOSTANNOUNCECLARITY was saved in player preference flags.
```
fira:
```suggestion
// (1<<15) was previously used: CHAT_GHOSTANNOUNCECLARITY was saved in player preference flags.
```

fira: People can reuse it. They just need to know so they can reset preference files.

fira (changes requested):

code/__DEFINES/__game.dm
antlersss: It might be better to keep the define but just label it as unused; that way if a mechanic like it is reintroduced for whatever reason the original bit flag isn't lost. Plus, coders are more likely to notice when they accidentally use a bitflag already used by a pre-existing define.

Drulikar:
Not that I imagine you'll believe me - but it was a mechanic that was worked out with the CO council. You can work with me on improving or at a minimum isolating the problems with it, but the mechanic is there to give SOs more to do to impact the round, and to give more importance to marines restoring comms. Consider properly giving feedback to PRs at https://forum.cm-ss13.com/w/pr-feedback/steps/step_1 but I would say generally people did use the feedback form while this was being tested - I currently see 5 feedback posts regarding it, all of which I personally discussed with those people and adjusted things to try and meet the problems they brought up with it at that time.

PR #12571, opened 2026-07-02
Stops hardened (crash proof) APCs from shorting out.

realforest2001:
# About the pull request

Mainly an issue with ARES, but having every single APC break with 5% chance each tick renders them pointless really as they all will break within a minute.

# Explain why it's good for the game

This will at least mean the critical areas are more resistant, and will stop the unreachable APC in ARES from breaking.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
balance: APCs that do not break on crash will not short circuit either.
/:cl:

PR #12522, opened 2026-06-25
More Punishing Ground Crash

Drulikar:
# About the pull request

This PR is a follow up to #11368 making ground crash more punishing:
- Pipe explosions are doubled in quantity
- Ship powernet loses all available power
- All ship apcs will randomly short out over time (repairable but you still ultimately only have its battery)

# Explain why it's good for the game

In general it doesn't make sense why a ship cracked in half would still be electrically working, but in addition feedback has been that groundcrash has not felt meaningful enough for xenos to achieve or marines to avoid.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Hijack:
image

After ground crash has hit the ground:
image

After random fault for this APC:
image

</details>

# Changelog
:cl: Drathek
balance: Almayer ground crash now breaks the powernet and APCs will randomly short out over time
/:cl:

PR #12674, opened 2026-07-18
Ground crash now causes xenomorph major

Drulikar:
# About the pull request

This PR does a couple things:
- Ground crash now triggers xenomorph major once the crashing sequence is complete
- A game mode modifier can be enabled by admins to not trigger round end on ground crash
- Modifier panel no longer generates an admin message if no setting is changed (from the modifier refusing the change)

Ultimately this came about because https://github.com/cmss13-devs/cmss13/pull/12571 and I'm not really getting a decision from Forest where it would be undoing some of the APC shorting from https://github.com/cmss13-devs/cmss13/pull/12522 but generally feedback has been that ground crash harms xeno's advantage more than it helps even though they completed their hijack objectives early. So now it will just by default end the round as a xeno major for smashing the ship into the ground.

<details>

New modifier: 
image

Modifier panel testing: 
image

Test as default modifier setting (round ends once ground crash completed): 
image

</details>

# Changelog
:cl: Drathek
balance: Hijack ground crash now causes xenomorph major once the ground crash sequence is complete
/:cl:

### DIFF
--- a/code/controllers/subsystem/hijack.dm
+++ b/code/controllers/subsystem/hijack.dm
@@ -802,7 +802,7 @@ SUBSYSTEM_DEF(hijack)
 	log_debug("crack_open_ship took [(world.timeofday - time) / 10]s")
 	explode_apcs(50)
 
-	if(!admin_sd_blocked)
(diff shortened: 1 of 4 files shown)


Detective-Google (approved):
code

PR #12678, opened 2026-07-18
HEDP no longer pushes bodies back

Red-byte3D:
# About the pull request

HEDP no longer pushes bodies back.

There's too much HEDP availability, and it used to bring bodies back is something i'd consider needs balancing.

The options are either this, or every time you get HEDP you get dealt a shit ton of damage.
I think people would prefer this as you can still clear around bodies without completetly aninhilating the body you want to retrieve

<details>

</details>

# Changelog
:cl:
balance: HEDP no longer pushes bodies around when they're dead for humans.
/:cl:


MomoBerri:
I'd rather shit ton of damage and reduce the explosion resistance you get when prone.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -201,7 +201,8 @@
 		var/knockout_minus_armor = min(knockout_value * bomb_armor_mult * 0.5, 0.5 SECONDS) // the KO time is halved from the knockdown timer. basically same stun time, you just spend less time KO'd.
 		apply_effect(floor(knockout_minus_armor), PARALYZE)
 		apply_effect(floor(knockout_minus_armor) * 2, DAZE)
-	explosion_throw(severity, direction)
+		if(!(stat == DEAD && cause_data?.body_throw))
```
fira: Wow uh no? cause_data is meant to be a container for logging. Not gameplay logic. On top of that, `body_throw` being set exclusively when an HEDP is fired is completely backwards. Find another way please.

fira (dismissed):

code/modules/mob/living/carbon/human/human.dm
Drulikar: I would either have more of a explosion power cutoff for throwing a mob thats lying down/dead, or you could have a `falloff_shape` that specifically is no dead mob throw.

fira: I think the most straightforward way would just be to pass an "explosion mods" alist to ex_act, if keeping the current PR binary logic of throw/don't throw. 

Components would be ideal but are unpractical here because we don't have any information about the explosion besides the arguments in ex_act ; and adding them in cellauto is convoluted due to explosion merging. 

Either way doing it properly requires some thinking and refactoring. Alternatively if this is acceptable in behavior i think having a cutoff on explosion strength is a great idea too yeah.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -201,7 +201,8 @@
 		var/knockout_minus_armor = min(knockout_value * bomb_armor_mult * 0.5, 0.5 SECONDS) // the KO time is halved from the knockdown timer. basically same stun time, you just spend less time KO'd.
 		apply_effect(floor(knockout_minus_armor), PARALYZE)
 		apply_effect(floor(knockout_minus_armor) * 2, DAZE)
-	explosion_throw(severity, direction)
+	if(!(stat == DEAD && severity <= EXPLOSION_THRESHOLD_LOW))
```
Drulikar:
```suggestion
	if(stat != DEAD || severity > EXPLOSION_THRESHOLD_LOW)
```

Drulikar (dismissed):

Git-Nivrak (approved):
Code

PR #12278, opened 2026-05-22
comms signal clarity meter on OW consoles

kitchenenthusiast:
# About the pull request
this was a good idea: https://forum.cm-ss13.com/t/overwatch-console-radio-clarity-meter/20423. so i DID it

adds a meter to the overwatch console UI to show the announcement signal strength based on the groundside comms garble. accounts for decoder sequences, drops dynamically in real time, and unobtrusively doesn't extend the OW console interface to make it any longer.

44% clarity or less = CRITICAL BLACKOUT / red
45% to 79% clarity = DEGRADED / orange
80% clarity or greater = STABLE / green

# Explain why it's good for the game
more feedback for butterbars that no one can hear them
 groundside overwatch console still needs this added but that UI is a jenga block stack i fear

uhhhhhh feel free to lmk if there's something different for how i should handle ftl crashes; i left as-is because i feel like survivors running out to the comms towers to set them up is very sovlful

# Testing Photographs and Procedure
Screenshot 2026-05-26 143147
Screenshot 2026-05-26 143201
Screenshot 2026-05-26 143307
Screenshot 2026-05-26 143720

not broken, updates dynamically, matches tgui, doesn't extend past the console window

only edge case is based on the radio garble itself. you can TECHNICALLY have your radio tower set up but if you don't have fac frequency added, but do have decrypt updated, it'll max at 95% and slowly degrade over time to follow with the announcement clarity degradation. once you add fac freq, it goes back to 100%.

# Changelog

:cl:
add: added an announcement clarity meter to Overwatch consoles so you no longer scream into the void, Lieutenant
ui: added an announcement clarity meter to Overwatch consoles
/:cl:

PR #11136, opened 2025-11-15
Coms Rolling Encryption & Punchcards

Drulikar:
# About the pull request

This PR is a follow up to #11024 reworking how clarity is calculated heavily. Namely now when comms are down several computers in CIC need to be used to keep a rolling authentication code current. (This replaces the mechanic of message length and delay for clarity)

---
**A mentor guide explaining this PR can be found here:**
https://forum.cm-ss13.com/t/guide-to-decoding-encrypting-encoding/17991

--- 

The process for completing a challenge goes like this:
- Get a random series of numbers from the decoder computer.
- Optionally print a punch card for the next step If the decoder has punch cards stocked (starts with 10 and 20 more are on the table - low chance for a single failed punch so double check the resulting number).
- Enter that number into the cipher computer (keep in mind its intentionally hex so if you want to enter the value directly you need to convert to decimal - though you can scroll the input to iterate numbers).
- Shift an offset until it creates a word.
- Optionally print a punch card for the next step again.
- Enter the required offset using the knob as well as the resulting number (punch card entry does not alter offset knob).
- Submit the encoding: If you are incorrect the PONG result will be ****. If you are correct but not timely it will be missing some number of letters directly proportional to the resulting clarity. If correct and timely it'll read PONG.

Challenges have a configured grace time (currently 1 minute) but otherwise degrade at the same rate that comms clarity degrades over time (currently 2.5% every 30s). That means if you're very timely you could technically resubmit the same answer again and get a bit of clarity back again, but after that grace time its directly proportional to the age of the challenge for how it would degrade normally.

Challenge phrases are a config so technically they could change at any point. The UIs are designed around 7 letter words, but I gave them enough height to support another row of numbers. The first configured word determines the length of words for all. The configured value will be forced uppercase, and anything not `A`-`Z` will be forced to be `-`.

Most art assets (card bgs, card logos, and item icons) were done with the help of Blundir. Inkscape SVG of the punch card is also in the assets. The font used is https://www.1001fonts.com/special-elite-font.html#license

Some other changes:
- `stars_decode_html()` is massively faster now using a regex (it performed abysmally on these punch cards without the change).
- USCM and WY paper item icons are now properly set on Initialize. For example if you spawned USCM paper it would revert to the paper icon because `update_icon` explicitly set anything without data to "paper"
- The CRT theme now sets the round gauge ring color.
- Fax windows now open with the same window size as papers.
- The photocopier now has the correct amount when copying bundles.
[...]

PR #12686, opened 2026-07-20
No comms decay and only garble on xeno comms destruction

Drulikar:
# About the pull request

This PR is a follow up to #11136 and #12278 doing the following:
- Signal quality no longer decays (other than challenges themselves decay so taking a long time on a challenge can still have a not-perfect result).
- Signal quality is set to minimum whenever a comms tower is destroyed that had marine frequencies encoded (if for example marines hold both comms, technically their faction clarity will drop to 45 but all communication remains in perfect clarity while the other tower remains online. If the other tower also goes down this would trigger a garble event again and then communication would be garbled until CIC resolved it or marines bring the tower back online).
- Whenever this occurs, it has a slightly delayed reaction to alert CIC with a balloon alert on the encoder computer, sfx from the encoder computer, and an ARES message on command channel. If you happen to have the encoding computer panel open when it occurs it even triggers the ping progress bar as if it was an automatic ping that failed.
- Fixed incorrect usage of `set_interaction` for the comms computers.
- The data packing from the OW meter PR is tweaked to no longer trample the theme on ground operations console and to correctly display the right numbers in the case of fallbacks for other factions ever they used the system.
- I also removed a useless argument `bypass_cooldown` from `ai_silent_announcement` and updated all calls that were trying to use it.

There was a sudden large influx of complaints about garble, and this is a solution I think still sort of achieves the aspect of xenos having the power to disrupt communications for marines - atleast temporarily - but drastically reduce the tedium of maintaining signal quality. Now its only on the event of xenos destroying the comms tower, and even then a successful decryption will remain in place until theres a tower again for xenos to destroy.

1. Run in dream daemon
2. Have one client connect while logged in to byond
3. Close pager and connect again
4. Start game with disabled end round checks on distress signal
5. Test various scenarios of bringing comms online and xenos destroying comms while the other client is in CIC checking encoder computer & OW computer

<details>

Data packing for OW consoles still works fine: 
image

</details>

# Changelog
:cl: Drathek
balance: Comms signal clarity no longer decays and instead drops to minimum when xenos destroy a comms tower that had marine frequencies. If a second tower happens to be online at the same time, there will still be perfect communication clarity until it goes down too.
fix: Fixed some logic with OW signal clarity trampling ui theme and incorrect fallback behavior
fix: Fixed incorrect set_interaction usage on comms computers
code: Removed useless argument bypass_cooldown from ai_silent_announcement
/:cl:


cm13-github:

cm13-github:

fira (approved):

PR #12690, opened 2026-07-21
Alters despoiler's unique screech sound effect, lowers volume

Mister-moon1:
# About the pull request
the screech sound unique to despoiler has been replaced with a unique roar sound effect at a much quieter volume

The screech sound is the same one used in alien isolation, the difference being it both wasn't ontop of you and wasnt spamming the sound, it is very harsh on the ears and i have to actually turn my volume down if there is a persistent despoiler in the game.

Other than that it feels very out of place and not in line with the other roar sound effects already present.

I tested it and I am not deaf

# Changelog

:cl:
soundtweak: despoiler screech replaced with unique roar that is in line with others, volume lowered from 75 to 30
/:cl:


SpypigDev:
While lowering the volume is a perfectly understandable change, I dont understand the rationale behind replacing the sound with a generic one.

a good chunk of features in the game get their definition from sound (mostly weapons), and i feel that replacing the sound with a generic roar would take away 

If it is indeed the case that the sound doesnt fit within the theme of the caste, would it not be better to replace it with an equally distinct sound that does?

DaGitSmog:
I'd rather be more in favor of lowering the sound rather than outright removing it.

SpypigDev (approved):

PR #12539, opened 2026-06-27
A.S.S again, Assault, Support, Security

ori-disillusion-mirnov:
# About the pull request

image

may or may not work atm, original A.S.S pr hasn't been updated in months and i have no experience with taking other prs and opening them as my own

Changes up squads to have designated roles and purposes which also affect their role composition and size.

The three "types" of squads are as follows:

Assault (>=50% of squad marines):
Alpha and Delta, these will be the squads as you've known until today and will not be limited, for people looking to just play without extra responsibility. The only difference is that they will have two smartgunners now instead of one.

Support (<=30% of squad marines):
Charlie, Kilo and Oscar, these squads will serve as filling the niche for players looking for small, coordinated squad gameplay and serve as an unknown variable up to command's discretion. They have no smartgunner or team leader.

Security (<= 20 of squad marines):
Bravo, tasked with, you guessed it - FOB Security, This includes FOB, Comms and possibly other duties down the line with logistics changes. The specialist and smartgunner are removed from bravo.

image

# Explain why it's good for the game
ASS system introduces an asymmetrical squad structure that allows players to head into rounds under the pretense that they will be able to engage in a specific type of CM experience, while leaning into squad archetypes and giving more flexibility to command roles.

The Assault squads represent the classic, large group CM gameplay experience, where individual player agency is at a maximum and cohesion is a function of herd behavior.

Support squads introduce a more refined, guided experience where 3 small squads function at the behest of command (similar to a PvE style experience, and officers are giving more tactile control over a subset of players on the ground.

Security is the traditional slower-paced backline and defense group, streamlined to only contain roles relevant, while issuing specialist and smartgunner roles to other squads.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl: Thwomper, Git-Nivrak
refactor: All headsets are now self setting.
add: Kilo and Oscar squads.
balance: Increase SG slots on Alpha and Delta to 2, Reduced SG slots on rest of the squads to 0, Reduced specialist slots on Bravo to 0, Reduced FTL slots on Charlie, Oscar and Kilo to 0.
balance: Riflemen slots on Kilo, Oscar and Charlie will be limited to 10% of total squad marines each, 20% on Bravo
[...]

PR #10623, opened 2025-09-13
[MDB IGNORE] A.S.S - Assault, Support, Security

Git-Nivrak:
# About the pull request
Changes up squads to have designated roles and purposes which also affect their role composition and size.

The three "types" of squads are as follows:

Assault (>=50% of squad marines):
Alpha and Delta, these will be the squads as you've known until today and will not be limited, for people looking to just play without extra responsibility. The only difference is that they will have two smartgunners now instead of one.

Support (<=30% of squad marines):
Charlie, Kilo and Oscar, these squads will serve as filling the niche for players looking for small, coordinated squad gameplay and serve as an unknown variable up to command's discretion. They have no smartgunner or team leader.

Security (<= 20 of squad marines):
Bravo, tasked with, you guessed it - ~~FOB~~ Security, This includes FOB, Comms and possibly other duties down the line with logistics changes. The specialist and smartgunner are removed from bravo.

IMG_5456

# Explain why it's good for the game
ASS system introduces an asymmetrical squad structure that allows players to head into rounds under the pretense that they will be able to engage in a specific type of CM experience, while leaning into squad archetypes and giving more flexibility to command roles.

The Assault squads represent the classic, large group CM gameplay experience, where individual player agency is at a maximum and cohesion is a function of herd behavior.

Support squads introduce a more refined, guided experience where 3 small squads function at the behest of command (similar to a PvE style experience, and officers are giving more tactile control over a subset of players on the ground.

Security is the traditional slower-paced backline and defense group, streamlined to only contain roles relevant, while issuing specialist and smartgunner roles to other squads.
# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
refactor: All headsets are now self setting.
add: Kilo and Oscar squads.
balance: Increase SG slots on Alpha and Delta to 2, Reduced SG slots on rest of the squads to 0, Reduced specialist slots on Bravo to 0, Reduced FTL slots on Charlie, Oscar and Kilo to 0.
balance: Riflemen slots on Kilo, Oscar and Charlie will be limited to 10% of total squad marines each, 20% on Bravo
maptweak: Changed up squad preps into platoon preps, Platoon 1 is Alpha And Delta, Platoon 2 is Charlie, Oscar, Kilo and Bravo. Credit to thwomper.
/:cl:

PR #12545, opened 2026-06-28
Repair Station Resprite + Directionals for WJ Storage.

Ordosian:
# About the pull request

Repair station needed a resprite, however after discussing with Blundir, it will remain ingame with an updated ALT sprite. ALT sprite has directionals.

New Repair Station Maps:
- Solaris
- Hybrisa
- Rift ship insert
- Chinook
- USS Almayer
- Desert Dam
- Whiskey Outpost
- White Antre

All the other maps use the old repair station as it seems more fitting.

Fixes #12544

All original WJ station sprites by Zenith & Blundir.

New repair station and WJ station directionals by DimDim

# Explain why it's good for the game

Fits better and looks better. 

WJ storage has directionals now as well so mappers can go crazy.

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

image

</details>

# Changelog
:cl: Ordosian, DimDim
fix: Fixed WJ station invisibility bug
imageadd: New synth repair station. WJ station directionals added.
maptweak: Implemented new repair stations.
/:cl:

PR #11934, opened 2026-03-12
Whiskey Outpost Bunker Improvements

CapCamIII:
# About the pull request

We had a WO round yesterday, and it was pretty fun but some improvements could definitely be made, so I'm making this PR to do those.

<details>
<summary>Map Changes</summary>

Expands the western bunker(CIC/Synth prep) to include an additional Corpsman prep(and spawners) as well as a decently sized armory attached to the CIC

Expands the command hallway in the bunker to 3 tiles wide, making it less annoying to traverse when crowded

Replaces the old corpsman hovel in the bunker offshoot with an outside triage area, so marines can grab additional meds without blocking each other in a 3x1 box to access the two marinemed vendors, additionally the room from moving this also allowed me to replace the the old marinemed vendors with some recreation/food vendors, easier to reach just outside medical.

Moves a few medic spawns from Beer to the main bunker so they spawn closer to the chem line

Moves SL spawns all to bunker, so Command can manually assign which squad is going to which bunker without having to shuffle squads already at the front.

Adds Medical Supply links to the medvendors at triage.

Adds rangefinders at the forward positions, 1 scattered at all 3

Adds motion detectors scattered on some racks throughout the map, allowing marines to more easily tell if they're being flanked.

Adds phone-lines to the map, phones are located at CIC, OW consoles, Beer, Beer Command Center, Medical, Triage, Logistics, Engineering, Mortar Pit, all 4 pillboxes

Changes the groundside ops console on WO to the new TGUI version, which is better.

</details>
<details>
<summary>Code Changes</summary>

Honor Guards can fireman carry

Mortar Crew have normal engineering skills, and are a bit better at lasing

Updates the uniform and weapon vendors with modern things that were not added, as well as updating existing things to match the Almayer vendor's amounts(if WO version was lower)

Adds RTO packs to the uniform vendor, half the amount as a normal game.

Re-phrases the Advanced Designator to make it clear that this is not a normal rangefinder, renaming the mortar strike to arty to differentiate it.

Makes the Advanced Designator's cooldowns 3 minutes instead of 5.

Makes the default glove for WO marines the normal marine gloves, instead of generic combat gloves

Makes the default pouch for WO marines a first aid bandages pouch, instead of a flare pouch

Honor Guard members should spawn with a full loadout now, with both this and the new armory meaning they don't have to run to squad prep and req to make sure they actually have combat gear besides a gun.

CO/XO spawn with gloves and boots now, as well as their own specific berets, in addition the XO now spawns with a helmet in their pack.

[...]

PR #12697, opened 2026-07-21
A.S.S. Redistributed: Assault, Support, Security

SunriseOverYourHead:
# About the pull request

This is a re-opening of the original 'A.S.S.: Assault, Support, Security' PR with some minor changes to squad pop distribution. All credit goes to Thwomper and Nivrak. I've only really made a few minor tweaks.

Original PR: https://github.com/cmss13-devs/cmss13/pull/10623
Code most recently forked from: https://github.com/cmss13-devs/cmss13/pull/12539

I've decided to try my hand at addressing some of the concerns that the marine playerbase initially expressed towards A.S.S. I hope this is a solution that ends up pleasing everyone, or at least compromises on some of the issues.

The changes I've made are as follows:

* BRAVO regains one non-roundstart slot for a Weapon Spec and a SG

This isn't me sneakily adding  a fifth spec and a fifth SG in. This just means that those roles can be TRANSFERRED into Bravo, but they will never spawn in as bravo.

* CHARLIE has double the rifleman pop cap now. CHARLIE also regains its roundstart spec & SG slot.

This means that in a round where all 4 specs and SGs can spawn, one is guaranteed to roll into Charlie. The intent here is giving Charlie a niche as a 'flanking' squad for more major objectives, while Alpha and Delta serve as frontline units.

* ALPHA and DELTA share THREE roundstart Spec & SG slots, and have a FOURTH slot for transfers only.

This means that in a round where all 4 specs and SGs can spawn, 3 are GUARANTEED to be distributed across Alpha and Delta in accordance to their preferences, with one extra slot for transfers if CIC wants to transfer Charlie's spec in, or Foxtrot spec in if it is bought. Etc.

The rest of A.S.S. remains unchanged. Hopefully those changes to pop & spec distribution will serve as a compromise that generally pleases most people who had issues with it. The intent here is to still allow a more fluid distribution of resources, maintaining A.S.S.' more tightly-knit playstyle, while preserving squad identity. If you have any concerns, feel free to reach me over Discord at BugWasabi and I'll try to address them.

Cheers!

Compromises between A.S.S.'s original vision & maintaining Charlie & Bravo's 'squad identities' with a simple and easy pop adjustment. Overall amount of specs and SGs remains the same. The only thing that really changes is their distribution.

<details>

</details>

# Changelog
I've altered the changelog to reflect those new changes. I am not adding myself in because this is really such a minor alteration that it wouldn't sit right with me.

🆑 Thwomper, Git-Nivrak, bugwasabi, Steelpoint
refactor: All headsets are now self setting.
add: Kilo and Oscar squads.
balance: Marine squads function differently. There are now three squad types, these are Assault Squads, Support Squads and the Security Squad.
balance: Assault Squads comprise of Alpha and Delta, this squad type functions identically to how squads work currently and are intended for players to play without extra responsibilities. They are composed of 50% of all active Marines. Each Assault squad has two smartgunners.
balance: Support Squads include Charlie, Kilo and Oscar squads. These fulfill a niche for players looking for smaller scale, coordinated, squad gameplay and serve as a unknown variable up to commands discretion. They are composed of 30% of active Marines. Each Support squad has no smartgunner or specialist. 
balance: The Security Squad consists of Bravo. They are tasked with security of USCM groundside assets. This includes the FOB, communications and possibly other duties down the line. The specialist and smartgunner are removed from Bravo squad. This squad will be composed of around 20% of active marines.
balance: Riflemen slots on Charlie, Kilo and Oscar will be limited to 10% of total squad marines each, 20% on Bravo
balance: Assault squads have one default Specialist slot each, and a 'joint cap' of two additional Specialists, for a total of four specialists between Alpha and Delta.
maptweak: Changed up squad preps into platoon preps, Platoon 1 is Alpha And Delta, Platoon 2 is Charlie, Oscar, Kilo and Bravo.
/:cl:


Unknownity:
Discard the Whiskey Outpost mapping changes as you would be reverting this PR #11934 if it would merge at this current state

cm13-github:

cm13-github:

code/game/machinery/doors/airlock_types.dm
SunriseOverYourHead, line 1: This is a patchwork fix and is largely a hacky and bad way to solve the unit tests failing.

A better way would be to remove those references from Whiskey Outpost, but this would probably involve opening a can of worms: A.S.S. removed references to Charlie airlocks and vendors but not Bravo, Alpha or Delta's. And the extra support squads do not currently exist in WO. 

So, instead of reworking WO to take in account support squads, it makes more sense to just add this patchwork fix.

cm13-github:

cm13-github:

Steelpoint (dismissed):
Mapping is fine at this time. 

We will need to revamp the layout of the marine prep areas to better account for high pop. I am aware this is a work in progress as the prep areas are being reworked by another contrib. The current prep areas are serviceable for possible testing.

SunriseOverYourHead:
Per Steelpoint's request, removing Charlie's roundstart SG and Spec slots. They still keep their CIC transferrable slots, as does Bravo. Max RFN back down to 4 from 8, as was in A.S.S.'s original PR. Alpha/Delta roundstart SPEC and SGs raised to 4 again.

code/game/jobs/job/marine/squad/tl.dm
```diff
@@ -11,6 +11,19 @@
 	. = ..()
 	spawning_human.important_radio_channels += JTAC_FREQ
 
+/datum/job/marine/tl/get_total_positions(latejoin=0)
+	var/extra_slots = 0
+
+	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
+		if(!target_squad)
+			continue
```
hry-gh: this will never happen, the loop above typechecks

code/game/jobs/job/marine/squad/leader.dm
```diff
@@ -7,6 +7,18 @@
 	gear_preset = /datum/equipment_preset/uscm/leader
 	entry_message_body = "<a href='"+WIKI_PLACEHOLDER+"'>You are responsible for the men and women of your squad.</a> Make sure they are on task, working together, and communicating. You are also in charge of communicating with command and letting them know about the situation first hand. Keep out of harm's way."
 
+/datum/job/marine/leader/get_total_positions(latejoin=0)
+	var/extra_slots = 0
+
+	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
+		if(!target_squad)
+			continue
```
hry-gh: this will never happen

code/game/jobs/job/marine/marine.dm
```diff
@@ -15,6 +15,24 @@
 	if(!Check_WO())
 		current_human.nutrition = rand(NUTRITION_VERYLOW, NUTRITION_LOW) //Start hungry for the default marine.
 
+/datum/job/marine/proc/calculate_extra_slots(latejoin, slots)
+	var/extra_slots = 0
+
+	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
+		if(!target_squad)
+			continue
```
hry-gh: this will never happen

code/game/jobs/role_authority.dm
```diff
@@ -279,6 +281,41 @@ I hope it's easier to tell what the heck this proc is even doing, unlike previou
...
 
+// Assigns players to squads to check if they should be returned to lobby
+/datum/authority/branch/role/proc/test_squads(list/random_players)
+	var/list/mob/living/carbon/human/test_humans = list()
+	var/list/mob/new_player/player_refs = alist()
+	for(var/mob/new_player/player in random_players)
+		if(!player || !player.ready || !player.mind || !player.job)
+			continue
+
+		var/mob/living/carbon/human/test_human = new
```
hry-gh: i still absolutely hate this, we have to figure out a way to handle this without spawning dummy mobs

code/game/machinery/vending/vendor_types/squad_prep/squad_specialist.dm
```diff
@@ -89,7 +89,7 @@ GLOBAL_LIST_INIT(cm_vending_gear_spec_heavy, list(
 		return
 	if(!user.skills.get_skill_level(SKILL_SPEC_WEAPONS) == SKILL_SPEC_TRAINED)
 		return
-	if((!human_user.assigned_squad && squad_tag) || (!human_user.assigned_squad?.omni_squad_vendor && (squad_tag && human_user.assigned_squad.name != squad_tag)))
+	if((!human_user.assigned_squad && squad_tag) || (!human_user.assigned_squad?.omni_squad_vendor && (squad_tag && human_user.assigned_squad.name != squad_tag)) && squad_tag != "null")
```
hry-gh, line 92: what is this "null" magic string???

code/game/machinery/vending/cm_vending.dm
```diff
@@ -575,7 +575,7 @@ GLOBAL_LIST_EMPTY(vending_products)
 				vendor_successful_vend(itemspec, user)
 				add_fingerprint(user)
 				return TRUE
-			if((!human_user.assigned_squad && squad_tag) || (!human_user.assigned_squad?.omni_squad_vendor && (squad_tag && human_user.assigned_squad.name != squad_tag)))
+			if((!human_user.assigned_squad && squad_tag) || (!human_user.assigned_squad?.omni_squad_vendor && (squad_tag && human_user.assigned_squad.name != squad_tag)) && squad_tag != "null")
```
hry-gh, line 592: what is this "null" magic string???

code/game/objects/structures/hybrisa_props.dm
```diff
@@ -1787,6 +1787,52 @@
...
+
+	. = ..()
+
+/obj/structure/mug_rack/update_icon()
+	overlays.Cut()
+	if(!amount)
+		return
+
+	if(amount < initial(amount) / 3)
+		overlays += image(icon=icon, icon_state="mugs-3")
```
hry-gh: these are equivalent, the icon is the first argument

```suggestion
		overlays += image(icon, icon_state="mugs-3")
```

code/modules/mob/new_player/new_player.dm
```diff
@@ -132,6 +132,22 @@
 		to_chat(src, SPAN_WARNING("[rank] is not available. Please try another."))
 		return
 
+	var/mob/living/carbon/human/test_human = new
```
hry-gh: this also sucks

hry-gh (changes requested):
all being said fine for testmerge

cm13-github:

<details><summary>Test Merge Deployed @ 07/28/2026 15:11:59 +00:00:</summary>

##### Server Instance

##### Revision
Origin: b5609dc6ad5195b287538a4db7a6ee127d662176
Pull Request: 9eb9aef54439192ee7288be921a344c174bdace8
Server: 891b422a0a5eb8131afbc65ab807cd8f1e766313
</details>
<details><summary>Test Merge Removed @ 07/28/2026 17:14:06 +00:00:</summary>

##### Server Instance
CM13 Live
</details>
<details><summary>Test Merge Deployed @ 09/06/2026 04:55:22 +00:00:</summary>

##### Server Instance

##### Revision
Origin: 63e9abc9a9c74ffcb4276cd8a007a4bcee74a50d
Pull Request: 537517e7c7ecfc1678d0b218e2a674c518c9f24f
Server: 1075ff5e68024fab68729ca295f117dcf5038fe6
</details>
<details><summary>Test Merge Removed @ 09/06/2026 05:38:46 +00:00:</summary>

##### Server Instance
CM13 Live
</details>

cm13-github:

code/game/jobs/role_authority.dm
SunriseOverYourHead: I'll take a gander at it later. Remind me if I forget please

code/game/machinery/vending/vendor_types/squad_prep/squad_specialist.dm
SunriseOverYourHead: Did a quick control+f. This is used in other spots of the codebase.

'Other spots' being, but not limited to:

code\modules\mapping\reader.dm at line 1104
code\modules\admin\view_variables\modify_variables.dm at line 130
code\game\machinery\telecomms\telecommunications.dm at line 22

Which isn't saying that it is correct, just that I suppose thwomper thought it was ok or copied it from somewhere else.

I agree, though, I have no idea why this is a string "null" instead of just null or void or an .exists? check. It seems mostly harmless, though. Let me know if you still want it changed.

code/game/machinery/vending/cm_vending.dm
SunriseOverYourHead: See my comment on the squad_specialist.dm conversation

code/modules/mob/new_player/new_player.dm
SunriseOverYourHead: I'll take a gander at it later. Remind me if I forget please

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

cm13-github:

code/__DEFINES/mode.dm
```diff
@@ -121,7 +121,7 @@ GLOBAL_LIST_INIT(ROLES_MEDICAL, list(JOB_CMO, JOB_RESEARCHER, JOB_DOCTOR, JOB_FI
 GLOBAL_LIST_INIT(ROLES_MEDICAL_ANTAG, list(JOB_UPP_LT_DOKTOR))
 GLOBAL_LIST_INIT(ROLES_MARINES,  list(JOB_SQUAD_LEADER, JOB_SQUAD_TEAM_LEADER, JOB_SQUAD_SPECIALIST, JOB_SQUAD_SMARTGUN, JOB_SQUAD_MEDIC, JOB_SQUAD_ENGI, JOB_SQUAD_MARINE))
 GLOBAL_LIST_INIT(ROLES_MARINES_ANTAG, list(JOB_UPP, JOB_UPP_ENGI, JOB_UPP_MEDIC, JOB_UPP_SPECIALIST, JOB_UPP_LEADER))
-GLOBAL_LIST_INIT(ROLES_SQUAD_ALL, list(SQUAD_MARINE_1, SQUAD_MARINE_2, SQUAD_MARINE_3, SQUAD_MARINE_4, SQUAD_MARINE_5, SQUAD_MARINE_CRYO, SQUAD_MARINE_INTEL))
+GLOBAL_LIST_INIT(ROLES_SQUAD_ALL, list())
```
Drulikar:
```suggestion
GLOBAL_LIST_EMPTY(ROLES_SQUAD_ALL) // Now filled in /datum/squad/New()
```

code/game/jobs/role_authority.dm
Drulikar: This is still unresolved (from Harry review); and seems to be relying on the BlackCrystalic logic here https://github.com/cmss13-devs/cmss13/pull/7911/changes#diff-5ef6549932190ff7ad0a8ea24861884626ce984ccb4f8af68c7459f020bc72acR565-R567

You would need to reintroduce similar logic, or pull it out into a proc if possible without a mob reference (just whatever values needed from a mob instance as arguments).

Alternatively you just have all job assignments also care about squad assignment too which probably would be more sane.

code/modules/mob/new_player/new_player.dm
Drulikar: " (above comment & Harry review)

code/modules/mob/new_player/new_player.dm
```diff
@@ -132,6 +132,22 @@
...
+	var/datum/equipment_preset/preset = new player_rank.gear_preset
+	test_human.faction = preset.faction
+	test_human.job = player_rank.title
+	if(player_rank.flags_startup_parameters & ROLE_ADD_TO_SQUAD && test_human.job != JOB_INTEL)
+		GLOB.RoleAuthority.randomize_squad(test_human, force_client=client)
+		if(istype(test_human.assigned_squad, /datum/squad/marine/cryo))
+			test_human.assigned_squad.forget_marine_in_squad(test_human)
+			GLOB.RoleAuthority.free_role(player_rank, force = TRUE)
```
Drulikar: Of note its yucky to have them put into a squad and then force them out of it just because you allowed BlackCrystalic's changes to the proc fall back to cryo squad. If logic like this stays around at all, you'd be better off handing `randomize_squad` an argument telling it to not fall back to cryo squad and having it return TRUE/FALSE if successful.

code/modules/mob/new_player/new_player.dm
```diff
@@ -151,10 +167,12 @@
 	SSticker.mode.update_energy_scale()
 
 	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
-		if(target_squad)
+
+		if(target_squad.dynamic_scaling)
 			target_squad.roles_cap[JOB_SQUAD_ENGI] = engi_slot_formula(length(GLOB.clients))
 			target_squad.roles_cap[JOB_SQUAD_MEDIC] = medic_slot_formula(length(GLOB.clients))
 
+
```
Drulikar, line 167: Of note this is being removed in https://github.com/cmss13-devs/cmss13/pull/13032 because this would be done every `check_role_entry` call to the job's `get_total_positions` BUT you removed that in this PR so at that point you can consider keeping this (But you need to use `get_total_marines()` instead not just count of clients - ghosts scaling engi/medic slots is stupid) OR consider restoring it to `get_total_positions`.

code/game/gamemodes/cm_initialize.dm
```diff
@@ -134,10 +134,14 @@ Additional game mode variables.
...
 	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
-		if(target_squad)
+
+		if(target_squad.dynamic_scaling)
 			target_squad.roles_cap[JOB_SQUAD_ENGI] = engi_slot_formula(marine_starting_num)
 			target_squad.roles_cap[JOB_SQUAD_MEDIC] = medic_slot_formula(marine_starting_num)
 
+		if(target_squad.pop_lock && target_squad.pop_lock < length(GLOB.clients))
+			target_squad.roles_cap = target_squad.initial_roles_cap
+
```
Drulikar: This proc is being removed in https://github.com/cmss13-devs/cmss13/pull/13032 which instead has calls to set_spawn_positions to perform the roles_cap. You probably should consider setting up whatever this `target_squad.roles_cap = target_squad.initial_roles_cap` logic is (if it still is necessary) in `/datum/authority/branch/role/proc/setup_candidates_and_roles`

code/game/jobs/role_authority.dm
```diff
@@ -413,8 +450,8 @@ I hope it's easier to tell what the heck this proc is even doing, unlike previou
 
 	return TRUE
 
-/datum/authority/branch/role/proc/free_role(datum/job/J, latejoin = 1) //Want to make sure it's a job, and nothing like a MODE or special role.
-	if(istype(J) && J.total_positions != -1 && J.get_total_positions(latejoin) >= J.current_positions)
+/datum/authority/branch/role/proc/free_role(datum/job/J, latejoin = 1, force = FALSE) //Want to make sure it's a job, and nothing like a MODE or special role.
+	if((istype(J) && J.total_positions != -1 && J.get_total_positions(latejoin) >= J.current_positions) || force)
```
Drulikar, line 452: forcing will be moot to have once earlier issues are resolved (no longer needing to remove someone from cryo squad).

code/game/jobs/role_authority.dm
```diff
@@ -558,7 +596,7 @@ I hope it's easier to tell what the heck this proc is even doing, unlike previou
 	SSround_recording.recorder.track_player(new_human)
 
 //This proc is a bit of a misnomer, since there's no actual randomization going on.
-/datum/authority/branch/role/proc/randomize_squad(mob/living/carbon/human/human, skip_limit = FALSE)
+/datum/authority/branch/role/proc/randomize_squad(mob/living/carbon/human/human, skip_limit = FALSE, client/force_client = null)
```
Drulikar: `force_client` will be moot to use once earlier issues are resolved (no longer having placeholder mobs).

Drulikar (changes requested):
Not a full review, but a preemptive review because of pending conflicts this PR is going to have.

code/modules/mob/new_player/new_player.dm
SunriseOverYourHead: Ended up solving this issue by extracting the squad selection algorithm out of the randomize_squad proc and into get_eligible_squad(). It now takes primitives instead of a mob and uses an accumulator for bookkeeping. Let me know if you'd like it handled in a different way. Regarding squads, this is still checking whether or not a job position is available and then seeing if there's a squad match.

code/modules/mob/new_player/new_player.dm
SunriseOverYourHead: done, length(GLOB.clients) is swapped for get_total_marines()

code/game/gamemodes/cm_initialize.dm
SunriseOverYourHead: I removed it from ass5 following what #13032 did. Still need pop_lock / initial_roles_cap since we still have Kilo and Oscar as poplocked squads, but I moved those into setup_candidates_and_roles after the engi/medic calls.

code/game/jobs/role_authority.dm
SunriseOverYourHead: force_client is gone from randomize_squad now that I removed the placeholder mobs, but I was a bit wary about removing the force parameter from free_role since I control+f'd it and saw that it's still used in cryopod.dm and in the admin panel, also test_squads still needs it because freeing an already granted slot is still necessary there since round starts commits clients to jobs and then validates squads.

code/game/jobs/role_authority.dm
SunriseOverYourHead: ditto, see above

cm13-github:

code/modules/cm_marines/overwatch.dm
```diff
@@ -558,22 +570,21 @@ GLOBAL_LIST_EMPTY_TYPED(active_overwatch_consoles, /obj/structure/machinery/comp
 
 
 /obj/structure/machinery/computer/overwatch/groundside_operations/ui_data(mob/user)
-	var/list/data = list()
-
-	pack_radio_data(data)
+	var/list/data = pack_radio_data()
+	data["executive"] = isnull(role)
```
Drulikar, line 575: You're unintentionally reverting changes from https://github.com/cmss13-devs/cmss13/pull/12686/changes#diff-4116d64978e3f746aaf9272545bcf45b211104e12ef3d5da826a1987d9eee6ecR516
which now causes https://runtimes.cm-ss13.com/cm13/issues/3852 & https://runtimes.cm-ss13.com/cm13/issues/3853

code/game/jobs/role_authority.dm
Drulikar: Don't use ctrl+f to find references. Use VSC and right click it to find references (in this case you would right click free_role and see what (if any) specify a third/named argument for force). Ctrl+f will find things that happen to just be named that. Of note if you're on 516.1687 you need to manually set the aux tools dll since ss13 peeps been slow about updating the extension. See https://discord.com/channels/150315577943130112/508399944986787850/1539484110504988762 and https://discord.com/channels/150315577943130112/964684928161808384/1539757412012138607

maps/map_files/USS_Almayer/USS_Almayer.dmm
Drulikar, line 1: Also I believe you're unintentionally reverting the change to the synth closet from https://github.com/cmss13-devs/cmss13/pull/12545 / https://github.com/cmss13-devs/cmss13/pull/12902

Image

SunriseOverYourHead: Fixed. It now looks like this:

image

cm13-github:

code/modules/cm_marines/overwatch.dm
SunriseOverYourHead: Solved

code/game/jobs/role_authority.dm
SunriseOverYourHead: That's fair, I looked over at it quickly and didn't notice that cryopod and the admin panel didn't actually use the force parameter.
I went ahead and took it out.

code/game/jobs/job/marine/squad/leader.dm
```diff
@@ -7,6 +7,15 @@
 	gear_preset = /datum/equipment_preset/uscm/leader
 	entry_message_body = "<a href='"+WIKI_PLACEHOLDER+"'>You are responsible for the men and women of your squad.</a> Make sure they are on task, working together, and communicating. You are also in charge of communicating with command and letting them know about the situation first hand. Keep out of harm's way."
 
+/datum/job/marine/leader/get_total_positions(latejoin=0)
+	var/extra_slots = 0
+
+	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
+		if(target_squad.pop_lock && target_squad.pop_lock < length(GLOB.clients))
```
Drulikar, line 14: Why does this scale using clients? Should be marine count or GLOB.players_preassigned or atleast player count.

code/game/jobs/job/marine/squad/tl.dm
```diff
@@ -11,6 +11,16 @@
 	. = ..()
 	spawning_human.important_radio_channels += JTAC_FREQ
 
+/datum/job/marine/tl/get_total_positions(latejoin=0)
+	var/extra_slots = 0
+
+	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
+		if(target_squad.pop_lock && target_squad.pop_lock < length(GLOB.clients))
```
Drulikar, line 18: Why does this scale using clients? Should be marine count or GLOB.players_preassigned or atleast player count.

code/game/jobs/job/marine/marine.dm
```diff
@@ -15,6 +15,21 @@
 	if(!Check_WO())
 		current_human.nutrition = rand(NUTRITION_VERYLOW, NUTRITION_LOW) //Start hungry for the default marine.
 
+/datum/job/marine/proc/calculate_extra_slots(latejoin, slots)
+	var/extra_slots = 0
+
+	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
+		if(target_squad.pop_lock && target_squad.pop_lock < length(GLOB.clients))
```
Drulikar, line 22: Why does this scale using clients? Should be marine count or GLOB.players_preassigned or atleast player count.

code/game/jobs/job/marine/marine.dm
```diff
@@ -15,6 +15,21 @@
...
+			target_squad.roles_cap = target_squad.initial_roles_cap
+
+		if(target_squad.dynamic_scaling)
+			if(latejoin)
+				target_squad.roles_cap[title] = slots
```
Drulikar, line 27: Why does only this one change the caps for late join but the others don't?

code/game/jobs/role_authority.dm
```diff
@@ -195,6 +195,11 @@ I hope it's easier to tell what the heck this proc is even doing, unlike previou
 	if(istype(medic_job))
 		medic_job.set_spawn_positions(GLOB.players_preassigned)
 
+	// Restore kilo and oscar to their caps once we have enough clients
+	for(var/datum/squad/target_squad in squads)
+		if(target_squad.pop_lock && target_squad.pop_lock < length(GLOB.clients))
```
Drulikar, line 200: Why does this scale using clients? Should be marine count or GLOB.players_preassigned or atleast player count.

code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
```diff
@@ -309,6 +306,65 @@
...
+		list("M41A AP Magazine (10x24mm)", floor(scale * 3), /obj/item/ammo_magazine/rifle/ap, VENDOR_ITEM_REGULAR),
+
+		list("EXTENDED AMMUNITION", -1, null, null),
+		list("M4RA Extended Magazine (10x24mm)", 1.8, /obj/item/ammo_magazine/rifle/m4ra/extended, null, VENDOR_ITEM_REGULAR),
+		list("M39 Extended Magazine (10x20mm)", 1.8, /obj/item/ammo_magazine/smg/m39/extended, VENDOR_ITEM_REGULAR),
+		list("M41A Extended Magazine (10x24mm)", 1.9, /obj/item/ammo_magazine/rifle/extended, VENDOR_ITEM_REGULAR),
+		list("M10 HV extended magazine (10x20mm)", floor(scale * 5), /obj/item/ammo_magazine/pistol/m10/extended , VENDOR_ITEM_REGULAR),
+
+		list("SPECIAL AMMUNITION", -1, null, null),
+		list("M56 Smartgun Drum", 2, /obj/item/ammo_magazine/smartgun, VENDOR_ITEM_REGULAR),
```
Drulikar, line 323: Is it intended only this squad has sg drums?

code/game/machinery/vending/vendor_types/squad_prep/squad_specialist.dm
Drulikar: `squad_tag` is never assigned so not only would it never be `"null"` it will never be anything other than `""`

Whatever `squad_tag` was/is for its not present and this is dead code.

Right click `squad_tag` in VSC and view all references. Its only *read* never written to.

code/game/machinery/vending/cm_vending.dm
Drulikar: `squad_tag` is never assigned so not only would it never be `"null"` it will never be anything other than `""`

Whatever `squad_tag` was/is for its not present and this is dead code.

Right click `squad_tag` in VSC and view all references. Its only *read* never written to.

code/game/machinery/vending/cm_vending.dm
```diff
@@ -1496,6 +1496,11 @@ GLOBAL_LIST_INIT(cm_vending_gear_corresponding_types_list, list(
 				user.equip_to_appropriate_slot(new_item)
 				new_item.update_icon()
 
+			if(istype(new_item, /obj/item/device/radio/headset/almayer/marine/self_setting))
+				var/obj/item/device/radio/headset/almayer/marine/self_setting/new_headset = new_item
+				new_headset.self_set()
```
Drulikar, line 1501: `self_set()` should be done in `/obj/item/device/radio/headset/almayer/marine/self_setting/Initialize(mapload, ...)` or `/obj/item/device/radio/headset/almayer/marine/self_setting/equipped(mob/living/carbon/human/user, slot)` instead.

code/modules/almayer/machinery.dm
```diff
@@ -79,6 +79,9 @@
 	var/minimap_flag = MINIMAP_FLAG_USCM
 	var/drawing = TRUE
 
+/obj/structure/machinery/prop/almayer/CICmap/dumb
```
Drulikar, line 82: This really could use a better typepath name e.g. `no_drawing` instead of `dumb`

code/modules/cm_marines/overwatch.dm
```diff
@@ -559,21 +571,21 @@ GLOBAL_LIST_EMPTY_TYPED(active_overwatch_consoles, /obj/structure/machinery/comp
...
-
 	pack_radio_data(data)
+	data["executive"] = isnull(role)
 	data["theme"] = ui_theme
 
 	if(!current_squad)
 		data["squad_list"] = list()
 		for(var/datum/squad/current_squad in GLOB.RoleAuthority.squads)
-			if(current_squad.active && !current_squad.overwatch_officer && current_squad.faction == faction && current_squad.name != "Root")
+			if(current_squad.active && (role ? current_squad.squad_role == role : TRUE) &&!current_squad.overwatch_officer && current_squad.faction == faction && current_squad.name != "Root")
```
Drulikar, line 581:
```suggestion
			if(current_squad.active && (!role || current_squad.squad_role == role) && !current_squad.overwatch_officer && current_squad.faction == faction && current_squad.name != "Root")
```

code/modules/cm_marines/overwatch.dm
```diff
@@ -559,21 +571,21 @@ GLOBAL_LIST_EMPTY_TYPED(active_overwatch_consoles, /obj/structure/machinery/comp
...
 				data["squad_list"] += current_squad.name
 		return data
 
 	data["current_squad"] = current_squad.name
 
 	for(var/datum/squad/index_squad in GLOB.RoleAuthority.squads)
-		if(index_squad.active && index_squad.faction == faction && index_squad.name != "Root")
+		if(index_squad.active && (role ? index_squad.squad_role == role : TRUE) && index_squad.faction == faction && index_squad.name != "Root")
```
Drulikar, line 588:
```suggestion
		if(index_squad.active && (!role || index_squad.squad_role == role) && index_squad.faction == faction && index_squad.name != "Root")
```

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -121,8 +121,6 @@
 		. += "Operation Name: [GLOB.round_statistics.round_name]"
 
 	if(assigned_squad)
-		if(assigned_squad.overwatch_officer)
-			. += "Overwatch Officer: [assigned_squad.overwatch_officer.get_paygrade()][assigned_squad.overwatch_officer.name]"
```
Drulikar, line 125: How come we're removing this info from their status tab?

code/modules/tgui_input/priority.dm
Drulikar, line 1: This ought to have an optional argument `default` where you pass it the existing `preferred_squad` list so that their current enabled preferences are present in their sorted order (but you still offer the other `items` as disabled items so they can still choose them if they want)

tgui/packages/tgui/interfaces/CentralOverwatchConsole.tsx
Drulikar, line 1: Kilo and Oscar colors should be used rather than red in normal overwatch console

Image

Drulikar, line 1: There is delayed updating for squad monitor & command personnel lists. 

Image

What this should mean is that its only updating because of an automatic refresh in tgui, rather than the react state being set correctly.

See https://react.dev/learn/updating-objects-in-state
https://react.dev/reference/react/useState

But this is a very complex panel so I don't think I'm going to easily identify what is done wrong as far as state. But its not a huge deal atm since the regular refresh fixes it eventually.

code/modules/cm_marines/overwatch.dm
Drulikar, line 1: Currently has some false assumptions about what is capable per squad. For example not all squads have a pad: 

Image
Image

Drulikar (changes requested):

tgui/packages/tgui/interfaces/CentralOverwatchConsole.tsx
Drulikar: This isn't exactly an ideal fix (ideally the UI is doing squad selection with a useSharedState and it just can locally change what is rendered rather than waiting on the server to tell it to render something else differently), but an adequate solution to this is to ensure `if("gather_index_squad_data")` performs a `return TRUE` at the end indicating to tgui that it should send an update now (rather than waiting on auto update to occur some time later).

Please look through the rest of these `switch(action)` and have the case `return TRUE` if changes were made to the data that the UI needs to reflect. Many already do, but some like this one don't.

From what I can tell the ones that should are:
- `if("dropbomb")` (cooldown goes to ui_data)
- `if("dropsupply")` (cooldown goes to ui_data)
- `if("gather_index_squad_data")`
- `if("evacuation_start")` (evac_status goes to ui_data)
- `if("evacuation_cancel")` (evac_status goes to ui_data)

SunriseOverYourHead: image

Fixed

code/modules/cm_marines/overwatch.dm
SunriseOverYourHead: Fixed

image

Intel, Oscar and Kilo don't have a supply drop entry anymore.

code/game/jobs/job/marine/squad/leader.dm
SunriseOverYourHead: Fixed, using get_total_marines() now

code/game/jobs/job/marine/squad/tl.dm
SunriseOverYourHead: Fixed, using get_total_marines() now

code/game/jobs/job/marine/marine.dm
SunriseOverYourHead: Fixed, using get_total_marines() now

code/game/jobs/role_authority.dm
SunriseOverYourHead: Fixed, using players_preassigned now

code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
SunriseOverYourHead: Added to platoon2 aswell (Alpha + Delta)

code/modules/cm_marines/overwatch.dm
SunriseOverYourHead: Applied

code/modules/cm_marines/overwatch.dm
SunriseOverYourHead: Applied

code/modules/almayer/machinery.dm
SunriseOverYourHead: Switched to no_drawing

code/game/jobs/job/marine/marine.dm
SunriseOverYourHead: Engineer.dm and Medic.dm do the same thing. Also, worth it to mention that the latejoin slots only get computed for Alpha and Delta, and they're static for Bravo Charlie Kilo and Oscar because those are unlocked wholesale once their pop limit is reached.

code/modules/mob/living/carbon/human/human.dm
SunriseOverYourHead: I believe this was removed by Thwomper. I've re-added it, since I also don't think there's any reason for it to be gone.

code/modules/tgui_input/priority.dm
SunriseOverYourHead: Added

code/game/machinery/vending/cm_vending.dm
SunriseOverYourHead: Done

code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
SunriseOverYourHead: Upon further examination, it is intended.
Platoon1 is Alpha+Delta
Platoon2 is support and security.
Platoon2 has no SGs, so, it shouldn't vend SG drums

tgui/packages/tgui/interfaces/CentralOverwatchConsole.tsx
SunriseOverYourHead: Done, I applied your suggestion to the issue. I noticed some lag when opening those tabs and switching them for the first time, but, after that, it is instant. Try it out again and let me know if it is still an issue. I can spend some more time diagnosing it if it remains a problem.

code/modules/cm_marines/overwatch.dm
SunriseOverYourHead: It would be interesting to rework the REQ supply pads to match this new system in the future. Kilo and Oscar not having a pad but Bravo and Charlie having one, as support squads, is inconsistent.

code/game/machinery/vending/cm_vending.dm
SunriseOverYourHead: I've also removed squad_tag, regarding the other two comments above.

code/game/jobs/job/marine/squad/leader.dm
```diff
@@ -7,6 +7,15 @@
 	gear_preset = /datum/equipment_preset/uscm/leader
 	entry_message_body = "<a href='"+WIKI_PLACEHOLDER+"'>You are responsible for the men and women of your squad.</a> Make sure they are on task, working together, and communicating. You are also in charge of communicating with command and letting them know about the situation first hand. Keep out of harm's way."
 
+/datum/job/marine/leader/get_total_positions(latejoin=0)
+	var/extra_slots = 0
+
+	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
+		if(target_squad.pop_lock && target_squad.pop_lock < get_total_marines())
```
Drulikar:
```suggestion
	var/marine_count = get_total_marines()

	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
		if(target_squad.pop_lock && target_squad.pop_lock < marine_count)
```
Don't need to re-count every squad

code/game/jobs/job/marine/squad/tl.dm
```diff
@@ -11,6 +11,16 @@
 	. = ..()
 	spawning_human.important_radio_channels += JTAC_FREQ
 
+/datum/job/marine/tl/get_total_positions(latejoin=0)
+	var/extra_slots = 0
+
+	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
+		if(target_squad.pop_lock && target_squad.pop_lock < get_total_marines())
```
Drulikar:
```suggestion
	var/marine_count = get_total_marines()

	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
		if(target_squad.pop_lock && target_squad.pop_lock < marine_count)
```

code/game/jobs/job/marine/marine.dm
```diff
@@ -15,6 +15,21 @@
 	if(!Check_WO())
 		current_human.nutrition = rand(NUTRITION_VERYLOW, NUTRITION_LOW) //Start hungry for the default marine.
 
+/datum/job/marine/proc/calculate_extra_slots(latejoin, slots)
+	var/extra_slots = 0
+
+	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
+		if(target_squad.pop_lock && target_squad.pop_lock < get_total_marines())
```
Drulikar:
```suggestion
	var/marine_count = get_total_marines()

	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
		if(target_squad.pop_lock && target_squad.pop_lock < marine_count)
```

Drulikar, line 22:
```suggestion
/datum/job/marine/proc/calculate_extra_slots(latejoin, slots, total_marines)
	var/extra_slots = 0

	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
		if(target_squad.pop_lock && target_squad.pop_lock < total_marines)
```
Bit redundant to keep counting many times. The caller has already performed this count.

code/game/jobs/job/marine/squad/leader.dm
Drulikar, line 14:
```suggestion
/datum/job/marine/leader/get_total_positions(latejoin)
	var/extra_slots = 0
	var/total_marines = get_total_marines()

	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
		if(target_squad.pop_lock && target_squad.pop_lock < total_marines)
```

code/game/jobs/job/marine/squad/tl.dm
Drulikar, line 18:
```suggestion
/datum/job/marine/tl/get_total_positions(latejoin)
	var/extra_slots = 0
	var/total_marines = get_total_marines()

	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
		if(target_squad.pop_lock && target_squad.pop_lock < total_marines)
```

code/game/jobs/job/marine/squad/medic.dm
```diff
@@ -11,7 +11,8 @@
 
 /datum/job/marine/medic/set_spawn_positions(count)
 	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
-		target_squad.roles_cap[title] = medic_slot_formula(count * MARINE_TO_TOTAL_SPAWN_RATIO)
+		if(target_squad.dynamic_scaling)
+			target_squad.roles_cap[title] = medic_slot_formula(count * MARINE_TO_TOTAL_SPAWN_RATIO)
 
 /datum/job/marine/medic/get_total_positions(latejoin=0)
 	var/slots = medic_slot_formula(get_total_marines())
```
Drulikar, line 18:
```suggestion
	var/total_marines = get_total_marines()
	var/slots = medic_slot_formula(total_marines)
```

code/game/jobs/job/marine/squad/medic.dm
```diff
@@ -21,11 +22,7 @@
 	else
 		total_positions_so_far = slots
 
-	if(latejoin)
-		for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
-			target_squad.roles_cap[title] = slots
-
-	return (slots*4)
+	return slots * 2 + calculate_extra_slots(latejoin, slots)
```
Drulikar, line 25:
```suggestion
	return slots * 2 + calculate_extra_slots(latejoin, slots, total_marines)
```

code/game/jobs/job/marine/squad/engineer.dm
```diff
@@ -11,7 +11,8 @@
 
 /datum/job/marine/engineer/set_spawn_positions(count)
 	for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
-		target_squad.roles_cap[title] = engi_slot_formula(count * MARINE_TO_TOTAL_SPAWN_RATIO)
+		if(target_squad.dynamic_scaling)
+			target_squad.roles_cap[title] = engi_slot_formula(count * MARINE_TO_TOTAL_SPAWN_RATIO)
 
 /datum/job/marine/engineer/get_total_positions(latejoin=0)
 	var/slots = engi_slot_formula(get_total_marines())
```
Drulikar, line 18:
```suggestion
	var/total_marines = get_total_marines()
	var/slots = engi_slot_formula(total_marines)
```

code/game/jobs/job/marine/squad/engineer.dm
```diff
@@ -21,11 +22,7 @@
 	else
 		total_positions_so_far = slots
 
-	if(latejoin)
-		for(var/datum/squad/target_squad in GLOB.RoleAuthority.squads)
-			target_squad.roles_cap[title] = slots
-
-	return (slots*4)
+	return slots * 2 + calculate_extra_slots(latejoin, slots)
```
Drulikar, line 25:
```suggestion
	return slots * 2 + calculate_extra_slots(latejoin, slots, total_marines)
```

code/game/objects/items/devices/radio/headset.dm
```diff
@@ -1072,12 +1072,23 @@
 //*************************************
 //-----SELF SETTING MARINE HEADSET-----
 //*************************************/
-//For events. Currently used for WO only. After equipping it, self_set() will adapt headset to marine.
+//Adapts itself to the wearer's squad and role on equip. Whiskey Outpost assigns squads after equipping, so it calls self_set() again once the squad is set.
```
Drulikar, line 1075: Calling self_set again after equipping tho isn't going to do anything if you have it prevented from self_setting a duplicate time.

So either both need to have squads figured out prior to equipping, or both need to wait for squad assignment before performing self_set.

code/modules/client/preferences.dm
```diff
@@ -1687,9 +1694,22 @@ GLOBAL_LIST_INIT(be_special_flags, list(
...
+						var/list/existing_squad_prefs = preferred_squad
+						default_selected = existing_squad_prefs.Copy()
+						choices = existing_squad_prefs.Copy()
+
+						for(var/squad in list("Alpha", "Bravo", "Charlie", "Delta", "Oscar", "Kilo"))
+							if(!(squad in choices))
+								choices += squad
+					else
+						choices = list("Alpha", "Bravo", "Charlie", "Delta", "Oscar", "Kilo")
+					var/new_pref_squad = tgui_priority_input(user, "Choose your preferred squads in order of priority or none for 'No Preference'.", "Squad Preference", choices, default_selected)
```
Drulikar, line 1709:
```suggestion
					var/list/choices = GLOB.squad_preference_options
					if(islist(preferred_squad) && length(preferred_squad))
						choices = preferred_squad.Copy()
						choices |= GLOB.squad_preference_options // Slot in any missing
					var/new_pref_squad = tgui_priority_input(user, "Choose your preferred squads in order of priority or none for 'No Preference'.", "Squad Preference", choices, preferred_squad)
```

So a few things being done here to make it cleaner:
- Just uses the master list as the default list if we happened to somehow not pass the list checks.
- Little redundant to islist check and also truthy check (one can argue truthy check may be faster but most situations we expect it to be truthy anyways.
- Redundant to copy existing_squad_prefs when tgui_priority_input is just going to make a new list and add each entry manually anyways (so its already copying)
- Although we could have just used a static list for the options, since we want that list also for sanitization its better as a GLOB; either way we only want to make that list once.
- Uses the `|` operator on lists to add unique entries rather than doing it manually.
- Redundant to typecast preferred_squad to a list if you just type it correctly in the definition in the first place (see next suggestion)

I would say its redundant to be checking preferred_squad is a list (because preferences should just sanitize it) but its whatever.

code/modules/client/preferences.dm
```diff
@@ -208,7 +208,7 @@ GLOBAL_LIST_INIT(be_special_flags, list(
 	var/body_size = BODY_SIZE_AVERAGE // Body Size
 	var/body_type = BODY_TYPE_LEAN // Body Type
 	var/language = "None" //Secondary language
-	var/preferred_squad = "None"
+	var/preferred_squad = list()
 	var/preferred_spec = list()
```
Drulikar, line 212:
```suggestion
	var/list/preferred_squad = list()
	var/list/preferred_spec = list()
```

code/modules/client/preferences.dm
Drulikar, line 1: Please also change https://github.com/cmss13-devs/cmss13/blob/090a692ce28563775eb087083d36cc51847448c2/code/modules/client/preferences_savefile.dm#L916-L918
to instead be:
```js
	preferred_squad = sanitize_list(preferred_squad, allow=GLOB.squad_preference_options)
	preferred_spec = sanitize_list(preferred_spec, allow=GLOB.specialist_set_name_dict)
```
It'll make sure the items are in the list and its atleast an empty list as default rather than sometimes sticking "None" in there.

Drulikar, line 1: Please also add
```js
GLOBAL_LIST_INIT(squad_preference_options, list("Alpha", "Bravo", "Charlie", "Delta", "Oscar", "Kilo"))
```
Near the top of this file or wherever makes sense.

code/modules/client/preferences.dm
```diff
@@ -1687,9 +1694,22 @@ GLOBAL_LIST_INIT(be_special_flags, list(
...
+								choices += squad
+					else
+						choices = list("Alpha", "Bravo", "Charlie", "Delta", "Oscar", "Kilo")
+					var/new_pref_squad = tgui_priority_input(user, "Choose your preferred squads in order of priority or none for 'No Preference'.", "Squad Preference", choices, default_selected)
+					if(isnull(new_pref_squad))
+						return // Canceled
+					preferred_squad = new_pref_squad
 
 				if("prefspec")
 					var/new_pref_spec = tgui_input_checkboxes(user, "Choose your preferred spec in order of priority or none for 'No Preference'.", "Specialist Preference", GLOB.specialist_set_name_dict, min_checked=0)
```
Drulikar, line 1715:
```suggestion
					var/list/choices = GLOB.specialist_set_name_dict
					if(islist(preferred_spec) && length(preferred_spec))
						choices = preferred_spec.Copy()
						choices |= GLOB.specialist_set_name_dict // Slot in any missing
					var/new_pref_spec = tgui_priority_input(user, "Choose your preferred spec in order of priority or none for 'No Preference'.", "Specialist Preference", choices, preferred_spec)
```

code/modules/tgui_input/priority.dm
```diff
@@ -0,0 +1,240 @@
...
+ * # async tgui_priority_input
+ *
+ * An asynchronous version of tgui_priority_input to be used with callbacks instead of waiting on user responses.
+ */
+/datum/tgui_priority_input/async
+	/// The callback to be invoked by the tgui_modal upon having a choice made.
+	var/datum/callback/callback
+
+/datum/tgui_priority_input/async/New(mob/user, message, title, list/items, callback, timeout, theme = null, ui_state)
+	..(user, message, title, items, callback, timeout, theme, ui_state)
```
Drulikar, line 223:
```suggestion
/datum/tgui_priority_input/async/New(mob/user, message, title, list/items, list/default_selections, callback, timeout, theme = null, ui_state)
	..(user, message, title, items, default_selections, callback, timeout, theme, ui_state)
```
Same needed to be done for the async version.

code/modules/tgui_input/priority.dm
```diff
@@ -0,0 +1,240 @@
...
+/**
+ * ### tgui_priority_input
+ * Opens a window with a list of items and returns a list of selected choices in order of priority.
+ *
+ * * Arguments:
+ * * user - The mob to display the window to
+ * * message - The message inside the window
+ * * title - The title of the window
+ * * list/items - The list of items to display
+ * * timeout - The timeout for the input (optional)
```
Drulikar, line 49:
```suggestion
 * * list/items - The list of items to display
 * * list/default_selections - Which items should start checked/ordered, in that order (optional, defaults to all of items)
 * * timeout - The timeout for the input (optional)
```

code/modules/tgui_input/priority.dm
```diff
@@ -0,0 +1,240 @@
...
+ * * theme - The ui theme to use for the TGUI window (optional).
+ * * ui_state - The TGUI UI state that will be returned in ui_state(). Default: always_state
+ */
+/proc/tgui_priority_input_async(mob/user, message, title = "Select", list/items, datum/callback/callback, timeout = 0, theme = null, ui_state = GLOB.always_state)
```
Drulikar, line 53:
```suggestion
/proc/tgui_priority_input_async(mob/user, message, title = "Select", list/items, list/default_selections, datum/callback/callback, timeout = 0, theme = null, ui_state = GLOB.always_state)
```

code/modules/tgui_input/priority.dm
```diff
@@ -0,0 +1,240 @@
...
+		if (istype(user, /client))
+			var/client/client = user
+			user = client.mob
+		else
+			return null
+
+	if(isnull(user.client))
+		return null
+
+	var/datum/tgui_priority_input/async/input = new(user, message, title, items, callback, timeout, theme, ui_state)
```
Drulikar, line 68:
```suggestion
	var/datum/tgui_priority_input/async/input = new(user, message, title, items, default_selections, callback, timeout, theme, ui_state)
```

code/modules/cm_marines/overwatch.dm
Drulikar: I kinda intend to remove squad droppods in the future actually and make it something that is very restricted (weight limit + long cooldown)

Anyways - theres a sort of minor issue now: If you swap from a squad that does have it, be on that page, and then select a squad that doesn't have it, you just get a 0.

Image

Maybe on squad selection, check if the new squad can_launch_crates and if not perform `setsecondaryCategory` to something it can?

Same thing can happen on the normal overwatch computers too (where you pick a squad that has it, use that category, log out, then log in for a squad without it) But in this case I'd probably just have logging out always set the tab to squad monitor or whatever.

Image

tgui/packages/tgui/interfaces/PriorityInput.tsx
Drulikar, line 1: IMO it would probably be nicer looking and more consistent if you had the sorting buttons on the left (since they're a fixed size) and then just have the entries fill the space.

E.g. instead of these
Image

Image

---

You instead have 

Image

Image

E.g. 

[patch.diff.txt](https://github.com/user-attachments/files/32224840/patch.diff.txt)

Though it probably doesn't even need to be a table and could instead be a flex but its whatever. The buttons don't need to be this small, but I was fighting with getting them to center the icon...

Drulikar (changes requested):
Don't forget you can batch apply changes on the files changed tab. Also you'll need to merge master in re kutjevo checks.

SunriseOverYourHead:
Sending in some radio fixes.

Confirmed to be working in-game and in WO

image

tgui/packages/tgui/interfaces/PriorityInput.tsx
SunriseOverYourHead: Addressed, as per your suggestion

image

code/game/objects/items/devices/radio/headset.dm
SunriseOverYourHead: Fixed

code/modules/client/preferences.dm
SunriseOverYourHead: Done

SunriseOverYourHead: Done

cm13-github:

code/game/objects/items/devices/radio/headset.dm
```diff
@@ -1069,66 +1086,96 @@
 //*************************************
 //-----SELF SETTING MARINE HEADSET-----
 //*************************************/
-//For events. Currently used for WO only. After equipping it, self_set() will adapt headset to marine.
+//Adapts itself to the wearer's squad and role. Gear is equipped before squads are assigned at round start,
+//but a vendor-bought headset arrives after, so we listen at both ends and let whichever happens last
+//configure us. Re-running is also ssafe, so squad transfers rebrand and rekey correctly.
```
Drulikar, line 1091:
```suggestion
//configure us. Re-running is also safe, so squad transfers rebrand and rekey correctly.
```

code/game/objects/items/devices/radio/headset.dm
Drulikar, line 1: So since headset keys are a hardref, and these are keys that are removable, and that set squad can be done at a terminal, this means that you can say extract the key, put it into another headset, then change squads, that key does (try) to get qdeleted from the other headset but it has a ref to it in keys for that other headset (so its going to eventually hard delete) but in the meantime you can just keep pumping out keys no?

I do not think we need to ever re-grant the key (also likely doesn't need to be a hard ref) because the key doesn't really care about *squad* it only cares about job.

So unless you want to do something else (like just have the encryption keys in their vendors or on their person to slot in) I would have a TRUE/FALSE var like var/key_granted instead of `granted_key`, don't ever need to qdel it or remove it from the list, and you just check that var to prevent re-granting a key. But you then are in the situation again where you need to be sure self_set only occurs when the marine is finished spawning. So alternatively you could just make these keys not be removable.

But up to you what you think makes the most sense. This is just some really bad, inflexible techdebt.

Drulikar (changes requested):

Drulikar:
Also there are a ton of sprites added in this PR - see https://github.com/cmss13-devs/cmss13/blob/master/.github/CONTRIBUTING.md#a-note-on-prs-altering-sprites re getting sprite approval. Doesn't look like any prior iteration of this PR ever complied.

PR #12704, opened 2026-07-23
Manual LZ Selection Required

RadChandelier:
# About the pull request

Requires manual designation of primary LZ before either dropship can take off; disables auto random LZ selection.

This is more or less an edge-case fix for early hijack. Instead of an LZ automatically being picked if none manually are - thus allowing Xenos to call down and hijack a dropship from a potentially empty LZ - this PR requires CO/aCO to select a primary LZ before flyby's, autopilot or manual departure can occur.

<details>
Video attached is local host test, showing successful implementation. Additionally, I did not record, but tested UPP ERT shuttle that was able to take off and land successfully, again completed on local.

https://github.com/user-attachments/assets/948b835f-414e-4709-b778-28c601c1015b

</details>

# Changelog

:cl: Rad Chandelier
add: Error messages to dropship consoles and remote consoles if dropship action is selected without LZ designated.
code: Requires primary LZ selection before dropships can do dropship stuff.
del: Removed random auto LZ selection.
/:cl:


code/game/gamemodes/cm_process.dm
```diff
@@ -180,10 +180,6 @@ GLOBAL_VAR_INIT(next_admin_bioscan, 30 MINUTES)
 			select_lz(lz2)
 		return
 
-	if(lz1 || lz2)
-		select_lz(lz1 || lz2)
-		return
-
```
Drulikar, line 186: Please look at all references to this proc `pick_a_lz` and the above check of `if(lz1 && lz2 && user)`. You've now made it impossible to select *any* lz if a map doesn't have two lzs.

Nor have you removed any automated calls to this, and you're just triggering a CRASH.

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -449,7 +449,8 @@
 		.["is_disabled"] = TRUE
 	.["locked_down"] = FALSE
 	.["can_fly_by"] = !is_remote
-	.["can_set_automated"] = is_remote
+	if(SSticker.mode.active_lz == null) .["can_set_automated"] = FALSE //Disables auto mode if default LZ not set
+	else {	.["can_set_automated"] = is_remote }
```
Drulikar: https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#control-statements

code/modules/shuttle/computers/dropship_computer.dm
```diff
@@ -612,6 +621,10 @@
 			shuttle.automated_lz_id = ground_lz
 			shuttle.automated_delay = delay
 			playsound(loc, get_sfx("terminal_button"), KEYBOARD_SOUND_VOLUME, 1)
+			if(SSticker.mode.active_lz == null)
+				to_chat(user, SPAN_WARNING("A primary landing zone must be designated prior to launch."))
+				playsound(loc, 'sound/machines/terminal_error.ogg', KEYBOARD_SOUND_VOLUME, 1)
+				return FALSE
```
Drulikar: If you wanted to disable auto, you're too late putting the code here. You are just preventing logging of it...

Drulikar (changes requested):
1. `select_lz(lz1 || lz2)` is not random. This means use lz1 if its truthy, else use lz2.
2. Code is still attempting to call `pick_a_lz` after 15 minutes - and it is intentional that this occurs after some time. If you want to increase the time that's fine - but I'm not going to allow marines to indefinitely stall the round.
3. If you simply wanted to align the `skip_time_lock` check lockout with the automatic selection in `/datum/game_mode/colonialmarines/process()` then you could have both check say `ROUND_TIME > SHUTTLE_TIME_LOCK` which would make it moot to have your new `SSticker.mode.active_lz == null` checks.
4. If your concern is marines flying to not the primary LZ first drop and they didn't pick one, which `/datum/game_mode/colonialmarines/proc/clear_proximity_resin()` would then cause lz1 to be picked, then instead of falling back to lz1 with `select_lz(lz1 || lz2)` have `/datum/game_mode/colonialmarines/proc/clear_proximity_resin()` specify which LZ they flew to. But this is moot if you just do 3 above unless you're concerned about admins force sending a shuttle to lz2 with the shuttle manipulator.
5. Please test your changes.

code/modules/shuttle/computers/dropship_computer.dm
RadChandelier: Adjusted to fit the rest of the PR!

code/modules/shuttle/computers/dropship_computer.dm
RadChandelier: Adjusted to fit style guide!

code/game/gamemodes/cm_process.dm
RadChandelier: Removed requirements of needing both LZ's, and simply requiring one and the user. I think that should eliminate the crash, so long as either LZ is selected. I'll do some testing on the time proc, as I think that's the only time it can be called with my changes.

code/modules/shuttle/computers/dropship_computer.dm
RadChandelier: Also just realized the video I uploaded is semi-useless, I'll try to get a better one and upload it with new commit after testing.

RadChandelier:
All fixes current; changes tested on local branch and all went through well for me.

Tried two other recording software and neither allowed the TGUI to be recorded - I can happily take screenshots or such as proof if needed.

code/game/gamemodes/cm_process.dm
```diff
@@ -161,15 +161,14 @@ GLOBAL_VAR_INIT(next_predator_bioscan, 5 MINUTES)
...
-/// If only one, or no user, it will pick the first available
 /datum/game_mode/proc/pick_a_lz(mob/user)
 	if(active_lz)
 		return
 
 	var/lz1 = locate(/obj/structure/machinery/computer/shuttle/dropship/flight/lz1)
 	var/lz2 = locate(/obj/structure/machinery/computer/shuttle/dropship/flight/lz2)
 
-	if(lz1 && lz2 && user)
+	if(user && lz1 || lz2)
```
Drulikar: This doesn't need to change. Revert your changes.

code/game/gamemodes/cm_process.dm
```diff
@@ -180,10 +179,6 @@ GLOBAL_VAR_INIT(next_admin_bioscan, 30 MINUTES)
 			select_lz(lz2)
 		return
 
-	if(lz1 || lz2)
-		select_lz(lz1 || lz2)
-		return
-
```
Drulikar, line 186: This doesn't need to change. Revert your changes.

code/game/gamemodes/cm_process.dm
```diff
@@ -161,15 +161,14 @@ GLOBAL_VAR_INIT(next_predator_bioscan, 5 MINUTES)
 GLOBAL_VAR_INIT(next_admin_bioscan, 30 MINUTES)
 
 /// Asks the user (optional) to pick the primary LZ if both LZ1 and LZ2 exist and it hasn't been set yet
-/// If only one, or no user, it will pick the first available
```
Drulikar, line 164: This doesn't need to change. Revert your changes.

code/game/gamemodes/colonialmarines/colonialmarines.dm
```diff
@@ -319,9 +319,6 @@
 /datum/game_mode/colonialmarines/proc/clear_proximity_resin()
 	var/datum/cause_data/cause_data = create_cause_data(/obj/effect/particle_effect/smoke/weedkiller::name)
 
-	if(!active_lz)
-		pick_a_lz()
-
```
Drulikar, line 324: This doesn't need to change. Revert your changes.

code/game/gamemodes/colonialmarines/colonialmarines.dm
```diff
@@ -535,8 +532,17 @@
 					xeno_message("The Hive is ready for a new Queen to evolve. The Hive can only survive for a limited time without a Queen!", 3, hive.hivenumber)
 
 
-		if(!active_lz && ROUND_TIME > lz_selection_timer)
+		if(active_lz == null && ROUND_TIME > lz_selection_timer)
```
Drulikar: This doesn't need to change. Revert your changes.

code/game/gamemodes/colonialmarines/colonialmarines.dm
```diff
@@ -535,8 +532,17 @@
...
 			pick_a_lz()
+		//Since forced LZ selection is no longer required, marines failing to select an LZ prior to the 30 minute mark will generate a warning, and end the game at the 35 minute mark, with leeways for server ticks.
+		if(active_lz == null && ROUND_TIME > 1800 SECONDS && ROUND_TIME < 1803 SECONDS)
+			var/name = "OPERATION WARNING"
+			var/input = "Operation will be terminated unless Primary Landing Zone is designated."
+			marine_announcement(input, name, 'sound/AI/commandreport.ogg')
+
+		if(active_lz == null && ROUND_TIME > 2100 SECONDS && ROUND_TIME < 2103 SECONDS)
+			round_finished = MODE_INFESTATION_X_MINOR
+			log_game("Marines failed to deploy!")
```
Drulikar, line 545: No. This is not at all what I was suggesting to be done.

Drulikar (changes requested):
In general I'm really lost at this point what you're even trying to do.

- Leave automatic LZ selection as it is.
- Again, if you want to adjust the timers to afford marines more time to pick, or standardize the automatic LZ selection time with the existing (on master) lockout thats fine.
- If you standardize the times, then all your changes to dropship_computer.dm are moot too.

code/game/gamemodes/colonialmarines/colonialmarines.dm
RadChandelier: In response to "I'm not going to allow marines to indefinitely stall the round." If you could provide more actionable information about why this doesn't work, or other options, I'd be happy to take a look. This seems parallel to me with xenos losing if no Queen joins on roundstart.

code/game/gamemodes/colonialmarines/colonialmarines.dm
RadChandelier: Reverted.

code/game/gamemodes/colonialmarines/colonialmarines.dm
RadChandelier: In response to "Nor have you removed any automated calls to this"

This is now redundant anyways, as the dropships cannot leave the hangar bay without an active LZ

code/game/gamemodes/cm_process.dm
RadChandelier: As this has changed and is no longer accurate, it is justified to be removed.

code/game/gamemodes/cm_process.dm
RadChandelier: This is a portion of the automatic LZ selection that I missed, and should be removed as the point of the PR.

code/game/gamemodes/cm_process.dm
RadChandelier: Left in response to " Please look at all references to this proc `pick_a_lz` and the above check of `if(lz1 && lz2 && user)`. You've now made it impossible to select _any_ lz if a map doesn't have two lzs.

Nor have you removed any automated calls to this, and you're just triggering a CRASH."

I don't understand what else you could mean outside of contradicting yourself here.

RadChandelier:
I am equally lost as to understanding what your interpretation of what my intention is. The entire purpose of the PR is to eliminate the automatic LZ selection. 

This necessitates a manual LZ selection through the console of a primary LZ. This has been clearly stated numerous times through the PR, commit details, attached (albeit shitty) video, and replies in here. Thus the new timers are a failsafe for an edge case scenario where marine CO/aCO that feel asleep at their desk doesn't AFK and prevent the round from occurring; and the cleanup of auto code and references to it were intentional and missed as this is my first time working with this codebase.

I'd kindly ask that you take another look at the PR from the top, and to please provide more explicit and actionable feedback as to what issues the PR would have from making it to being TM'd.

code/game/gamemodes/cm_process.dm
Drulikar, line 1: Again, this proc (`pick_a_lz`) doesn't need to change at all. If you wanted to remove automatic lz selection, you'd remove the references that call `pick_a_lz` without a user argument.

The existing, untouched by this PR version of this proc would do the following:
- Do nothing if a LZ is already selected
- Attempt to get a reference to LZ1 and LZ2
- If user is provided, LZ1 exists, and LZ2 exists, ask that user to pick a LZ
- If only one LZ exists, or no user provided, pick the first LZ that is truthy
- If no LZ exists error

code/game/gamemodes/colonialmarines/colonialmarines.dm
Drulikar: I would much rather prefer xenos just have a timer where they can call down the dropship; but w/o automatic LZ selection that would mean logic would have to change to allow them to approach either LZ's terminal in the situation of no LZ has been selected and bring down the dropship.

Its not so much that this should be occurring in a round, but the possibility of it can should make this not something marines stall to do.

code/game/gamemodes/colonialmarines/colonialmarines.dm
```diff
@@ -537,6 +534,15 @@
 
 		if(!active_lz && ROUND_TIME > lz_selection_timer)
 			pick_a_lz()
```
Drulikar, line 536: This for example is a place where automatic LZ selection is attempted. The solution to disable automatic LZ selection wouldn't be to break the `pick_a_lz` proc. It would be remove this call, and others like it that are performing automatic LZ selection. If you are using VSC like https://cm-ss13.com/wiki/Guide_to_Git would walk you through setting up, you can right click the `pick_a_lz()` symbol there and Find All References: 

Image

OF NOTE: `/datum/game_mode/colonialmarines/proc/clear_proximity_resin()` is also a place that automatic LZ selection can occur. This is because this proc *needs* the `active_lz` set so it knows where to spray the weed killer. This proc is only called once ever on first drop. So I would prefer it remain as is in the event that an admin causes first drop to occur manually. (And potentially on how xenos pulling a dropship down is handled I guess)

Drulikar (changes requested):

code/game/gamemodes/cm_process.dm
RadChandelier: I think I'm doing the same thing, just in a different way. Edited, it does the same except for picking an LZ automatically if there is no user, an instead of requiring both LZ1 and LZ2, it now requires LZ1 or LZ2. Thus, there needs to be a user input to pick an LZ, which is the effect of the PR.

code/game/gamemodes/colonialmarines/colonialmarines.dm
RadChandelier: I can have a look at replacing the end game scenario, but it feels more of a stylistic choice; I think it makes less sense to initiate a hijack in this case than it does to end the game - again in a very edge-case scenario. I feel this is a parallel ending to which the Xeno don't have a Queen and the round ends with a timer as a marine minor - marines don't get the message to run to the hive and have a TDM, the game just ends.

As far as "breaking" the proc, all I changed was - again - the automatic selection process within the proc, and the resolution of a potential crash if there is only one LZ. Calls to pick an LZ if there isn't actively one should stay, and with user input will resolve. This is the case for each of the references to the proc; in colonialmarines.dm, it is how the proc is handled. In communications.dm, groundside_operations.dm, and overwatch.dm, it allows the respective consoles to run the proc. Unless I'm completely misunderstanding this, nothing needs to change here.

The weedkiller proc - again - will not fail to fire unless something breaks as - again - neither dropship can leave the hangar bay without a primary LZ selected. Which is - again - the whole point of the PR. I'm not sure of what scenario an admin would cause first drop to happen outside of utilizing in-game systems, or simply taking the few seconds to interact with the console, but I can implement that if you can point me to where the admin proc would happen outside of these systems.

code/game/gamemodes/cm_process.dm
Drulikar: `if(user && lz1 || lz2)`

Means if there's a user and a lz1, do this
Or there is lz2, do this

A) You're once again breaking handling for maps with only one lz (they exist and I won't necessarily be lz1 or lz2)
B) Even if you used parens to fix the operator order, in the scenario of just one LZ you are making the user select from a list of one. That's unnecessary.
C) Without how the code already is, you can't accommodate code like the the clear promixity resin which won't have a user. 

See http://play.dm-lang.org/#1:eF5T0i8oyk_Wz03MzNPQjMlTUCjPL8pJ0cvJT1ewsVHQMFBQU1MwUKipUTDEJWsIkjXAImsI1wuSVQIAWIIdFw

code/game/gamemodes/colonialmarines/colonialmarines.dm
Drulikar: A) Marines already see xeno minor as a win (that's what they get when they wipe out all xenos in hijack)
B) Hijack haters are eager to have ways to skip hijack when the point is to not deny xenos the gameplay, and to bring the danger to shipside personnel.

See other comment re the unnecessary changes that you for whatever reason are insiting on ignoring requested changes.

code/game/gamemodes/cm_process.dm
RadChandelier: Oh, I misunderstood how that works. Defaulted if both LZs to: if (user && lz1 && lz2).

Added precursor check of:
 if((lz1 && !lz2) || (lz2 && !lz1))
		select_lz(lz1 || lz2)
In order to prevent need to select an lz if there is only one on the map, while only being able to execute if there explicitly is only one without the other. I couldn't find an example in code for a map that only has a single lz and not labeled lz1; Fort McNeil comes to mind, but I'm not sure where to find the label for its lz (unless it's just lz, which I can slap in). This should then fix the clear proxy resin, which is based off of there being an active lz selected.

code/game/gamemodes/colonialmarines/colonialmarines.dm
RadChandelier: I saw it more as if there isn't anyone competent enough on the marines side of things to be able to pick an LZ - or tell aCO how to - after half an hour (which I also think could come down to warn at 25, effect at 30), a mulligan would be better with a fresh round anyways, as neither side is getting the gameplay they want. 

However, I can see your perspective of the hijack solution providing something to do in the round; but I see the it as a potential incentive to bait out hijack from the marine side, thus having full manpower to fight the hive without risk of anything groundside, hence my reference to it as an impromptu TDM; and why in my eyes it'd be better to just end it, to disincentivize that behavior - majority of marines wouldn't get the gameplay they want! I can see that this change might be outside the scope of the PR though, and can revert it if that's still desired.

code/game/gamemodes/colonialmarines/colonialmarines.dm
Drulikar: Dropships can be sent via shuttle manipulator so it should remain as I've been explaining.

RadChandelier: I tested this on local as it stands, by sending Alamo via Shuttle Manipulator, and it correctly set the primary LZ to the selected LZ without any other interference; both sending to LZ1 and LZ2 set the primary LZ to whichever it was sent to.

code/game/gamemodes/cm_process.dm
Drulikar: The old version of the file would already achieve this and again your changes are just unnecessary.

Drulikar:
I'm not against something like this being done, but whoever is going to do this needs to heed the requested changes not just try again and again to skirt around them.

PR #12709, opened 2026-07-23
FIX: Roller Beds and Surgical Beds are now always collapsed when spawned in

Puckaboo2:
# About the pull request
This PR fixes an unreported bug that caused roller beds and surgical beds, in some instances, to be vended or spawned on characters already deployed, so you could not interact with them unless you rolled them back up.

Will test later

This was caused by (myself) and other people using the `/obj/structure/bed/portable_surgery` and `/obj/structure/bed/roller` item paths instead of `/obj/item/roller` and `/obj/item/roller/surgical_bed` paths. 
`

For obvious reasons, structures should not be vended or spawn in inventories. In this case, bended Roller Beds and Surgical Beds needed to be rolled up before being picked up. For those who *spawn* with said bed in their inventory, well... That item slot is forever broken because they cannot drop or roll up that bed. 

<details>

</details>

# Changelog
:cl:
fix: CLF Soldiers and CLF Survivor Medics no longer spawn with deployed Surgical Beds in their inventories.
fix: Roller Beds and Surgical Beds longer vend from vending machines and Doctor/Field Doctor Essentials already deployed, so you no longer need to roll them up. 
/:cl:

### DIFF
--- a/code/game/machinery/vending/vendor_types/crew/medical.dm
+++ b/code/game/machinery/vending/vendor_types/crew/medical.dm
@@ -391,8 +391,8 @@ GLOBAL_LIST_INIT(cm_vending_clothing_field_doctor, list(
 		/obj/item/storage/surgical_case/regular,
 		/obj/item/clothing/accessory/stethoscope,
 		/obj/item/device/flashlight/pen,
-		/obj/structure/bed/portable_surgery,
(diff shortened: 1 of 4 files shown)


fira (approved):

fira:
nice catch

PR #12711, opened 2026-07-24
Engineering survivors and synths no longer spawn with plasteel.

Uristthedorf:
# About the pull request
Removes plasteel from the inventory of all survivors. If they did not have any metal, gave them an equivalent amount of metal.

rwall2
R-walls allow for degenerate strategies that are uninteractive and boring to fight against or hold. If survivors are to get any plasteel. in my opinion they should have to risk going out into the colony to gather it, instead of getting it for free.

If this makes engineering survivors weak or something, we could make nailguns standard for engineer survivors or something.

<details>
Made test server and spawned in various engineering survivors.

</details>

# Changelog
:cl:
balance: Survivors no longer spawn with plasteel.
/:cl:


Steelpoint (approved):
I am inclined to agree with this proposal. For the utility that plasteel provides, survivors should be required to scavenge the colony to find plasteel resources. 

This may need a followup revision of some maps loot tables to ensure there is a fair distribution of plasteel. I also would note that there is a PR (Cape River) which adds RNG plasteel spawners which could also be used to help add more plasteel acquisition opportunities.

Ammoniacres:
> I am inclined to agree with this proposal. For the utility that plasteel provides, survivors should be required to scavenge the colony to find plasteel resources.
> 
> This may need a followup revision of some maps loot tables to ensure there is a fair distribution of plasteel. I also would note that there is a PR (Cape River) which adds RNG plasteel spawners which could also be used to help add more plasteel acquisition opportunities.

Lot of places to work on for this. RNG plasteel spawners would be nice but, realistically unneeded. The focus should be on making sure inserts have access to plasteel so that they aren't suddenly in far worse positions under this PR.

PR #12706, opened 2026-07-23
MapGrids: QuadTrees Replacement

fira:
# About the pull request

This replaces our antique QuadTrees with a much more boring structure that I call "MapGrid", basically simply a grid with configurable dimension (by default 5x5) cells. Also comes with granular updating instead of SS-backed scanning.

## TODO

- [x] Implement
- [x] Validate basic functionality
- [x] Validate thoroughly nested contents tracking
- [x] Review / Robustify for live testing
- [x] Live Test and observe performance and edge case bugs
- [x] Implement Grid boundaries balancing 
- [x] Live Test again
- [x] Get rid of legacy query proc for something more direct
- [x] Retest
- [ ] Re-review
- [ ] Merge

## Why would you do that?

QuadTrees in our codebase have a number of problems, such as: 
 * They are only created every 0.5s, meaning you can have up to 1s of lag on ranging things like Motion Detectors and sound ; this causes inaccuracies in sound play and generally jank
 * They are created from scratch every 0.5s then discarded entirely
 * QuadTrees make an absurd amount of proc calls which make up for the bulk of its performance impact - in good part due to the fact you have /datum/shape-s and it goes through many hoops to use them 
 * QuadTrees largely predate Multi-Z and handle each of them separately - this requires multiple updates each SS fire, and things like sound to do multiple scanning runs (remember, with all the proc overhead) each time
 * QuadTrees are built to track /client-s specifically. If something had no client, it can't be tracked. This causes inconsistent behavior with dead xenos that stayed in body and connected triggering Motion Detectors while others don't
 * This also means sentries cannot fire at simple_animal-s for example because they can't detect them

## What's better about this?

 * It can track any movable, not limited to online clients
 * There is no delay, updates are granular
 * The structure is reused and can be balance for optimization (that's WIP)
 * Bulk of search is done in a single proc, avoiding proccall overhead
 * MapGrids are common for several levels of a single Multi-Z, requiring only one structure and query
 
## Performance?

Testing still ongoing but so far:
 * Query performance seems 2-3x better than QuadTrees
 * Upfront performance can't easily be quantified because it happens on every mob move, but should be better too
 * Rebalancing is done with very loose scheduling, so it should never really be a factor

Overall as per current projections I think this can improve the game overall performance by 2-4%. Which sounds like nothing, but is really pretty huge.

## What's the catch?

[...]

PR #12717, opened 2026-07-25
MapGrids TM Holder with Rebalancing

fira:
This is just #12706 with the added rebalancing feature, so we can try both in TestMerge and evaluate performance impact.

More Tech info about the rebalancing? Sure.
By default QuadTrees and MapGrids both have fixed boundaries. MapGrids cut space in a 4x4 grid, and QuadTree cuts it in recursive 2x2 grids (hence the name).

This suffers greatly in a case we're well acquainted with at CM, which is when many people are in the same places. This typically occurs during hive pushes, FOB sieges, and hijack. 
When this happens, the entire structure becomes a liability, it's just a mega list with extra handling overhead.
QuadTrees should alleviate this issue, but ours are too crude to compensate - in fact, they made this worse.

Balancing the grids is much easier, because they're not recursive so it can be done in one pass. 
We do this by sorting all of the objects, then setting bounds to the appropriate percentiles, so each vertical/horizontal delimiter has the same amount of objects. It's not perfect in that balancing X and Y separately isn't correct, but i believe it's already 90% of the benefits for our usecase.

The downside is that, just like QuadTrees, this requires upfront work. This work in fact is pretty expensive. 
SSmapgrids is therefore configured to only run every 20 seconds at very low priority, such that it's not a burden in load.
The problem is that during highpop, it may still cause overtime, which is why this is in a separate PR for testing.


cm13-github:

<details><summary>Test Merge Deployed @ 07/25/2026 11:32:53 +00:00:</summary>

##### Server Instance

##### Revision
Origin: 1cb954c660dc70efe3fc5b07e74cd70afb6c6c5c
Pull Request: e19c39239ee454abf4159c8e471b4904fd2d9826
Server: 859d76158f878b267449412e22f8dfbba77d7cea
</details>
<details><summary>Test Merge Updated @ 07/25/2026 13:07:52 +00:00:</summary>

##### Server Instance

##### Revision
Origin: 1cb954c660dc70efe3fc5b07e74cd70afb6c6c5c
Pull Request: 81ae3c627457c377e491f162926de5642ef35ea8
Server: e34b1d5031c32090bf5ea4e18450bfcde505057d
</details>
<details><summary>Test Merge Updated @ 07/25/2026 13:29:23 +00:00:</summary>

##### Server Instance

##### Revision
Origin: 1cb954c660dc70efe3fc5b07e74cd70afb6c6c5c
Pull Request: 1a87327138b01ce8ce6f06971a2036260035eba0
Server: 9d35e7b34496f727bfcb8b01e0a95548e6941d42
</details>
<details><summary>Test Merge Removed @ 07/25/2026 14:25:42 +00:00:</summary>

##### Server Instance
CM13 Live
</details>

PR #12731, opened 2026-07-26
HPR holotargetting box magazine capacity normalization

sippykot:
# About the pull request

Normalizes the size of all HPR magazines to follow the base 300 rounds - the only in-game change is the holotargetting magazine being upped from 200 to 300

1) code maintenance - there's no reason for all the subtype magazines to have their own max_rounds
2) consistency - no other marine weapon has special ammo mags being smaller
3) balance and utility - the big one, a lot of people don't even take holo-targetting boxes in the first place because they consider them a waste of space; bringing them in line with other magazines would make them at least provide similar value to the normal ones, they are rarer and harder to get but less desirable at the moment

very simple value change

# Changelog

:cl:
balance: M41AE2 holo-targetting box magazine holds 300 rounds (previously 200)
/:cl:

### DIFF
--- a/code/modules/projectiles/magazines/rifles.dm
+++ b/code/modules/projectiles/magazines/rifles.dm
@@ -319,21 +319,18 @@
 	name = "\improper M41AE2 ammo box (10x24mm holo-target)"
 	desc = "A semi-rectangular holo-targeting box magazine for the M41AE2 Heavy Pulse Rifle."
 	default_ammo = /datum/ammo/bullet/rifle/holo_target
-	max_rounds = 200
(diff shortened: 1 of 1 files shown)


Detective-Google (approved):
the holotargeting rounds already are mid and can't be refilled so I don't see an issue with it being made not also a space loss over normal ammo.

Steelpoint (approved):
This is a fair change. Holo ammo is strictly inferior to standard ammo, and all HPR kits come with a single magazine for free, that is usually discarded or used as a emergency backup when the standard magazine is empty.

PR #12736, opened 2026-07-26
[REFACTOR] Updates Hedgehog code to use new define and removed spaces.

Venuska1117:
Detached from: #12601

# About the pull request

Updates code to use new define.

Shorter code.

# Testing Procedure

Made new code, run code, no errors.
spawn ravager, use abilities, abilities work fine.

Code works as intended.

# Changelog

:cl: Venuska1117
refactor: Updates Hedge Ravager code to use new define and removed spaces in code.
/:cl:


cm13-github:

cm13-github:

Drulikar (changes requested):
I'm generally not okay with this refactor. I don't see it as an improvement. The concept of shards should be a hedgehog resource, not any individual ability.

If say you're trying to make `/datum/action/xeno_action/onclick/spike_shed` be a generic ability - it needs a way that you provide it some arbitrary resource. But then with it being a generic ability, its still hardcoded with the intention of it locking you out from gaining that resource anymore. It only promises that via a to_chat and sets a trait but it itself isn't in control of that resource.

The resource is a hedgehog ability thats its gimmick that it tries to interweave into all of its abilities.

Also the traits - though they're appropriate in the perspective of the ability - aren't anywhere as clear why they're affecting the shards how they are in the delegate. What was once a concept of shards being locked is now just the absence of a `TRAIT_ABILITY_SHED_SPIKES` trait.

PR #12747, opened 2026-07-27
Minor optimization to hot minimap procs

fira:
# About the pull request

Minor optimization to `/image/proc/minimap_on_move`

The proc is both hot and extremely costly for some reason.
During recent 200-pop round it hit about 82,000 ms over 510,000 calls. 

The only real highlight is it reuses the result of minimap lookup instead of doing it twice, because I don't see what else could be expensive here unless it's literally setting pixel_x/y.

-- 

Same treatment for `/datum/controller/subsystem/minimaps/proc/remove_marker` which is vastly less expensive, but still expensive

:cl:
code: Improved minimap_on_move and remove_marker performance
/:cl:

### DIFF
--- a/code/controllers/subsystem/minimap.dm
+++ b/code/controllers/subsystem/minimap.dm
@@ -624,18 +624,20 @@ SUBSYSTEM_DEF(minimaps)
  */
 /image/proc/minimap_on_move(atom/movable/source, oldloc)
 	SIGNAL_HANDLER
-	if(isturf(source.loc))
-		pixel_x = MINIMAP_PIXEL_FROM_WORLD(source.x) + SSminimaps.minimaps_by_z["[source.z]"].x_offset
-		pixel_y = MINIMAP_PIXEL_FROM_WORLD(source.y) + SSminimaps.minimaps_by_z["[source.z]"].y_offset
+	if(source.z)
+		var/datum/hud_displays/minimap = SSminimaps.minimaps_by_z["[source.z]"]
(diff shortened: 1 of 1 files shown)


code/controllers/subsystem/minimap.dm
```diff
@@ -624,18 +624,20 @@ SUBSYSTEM_DEF(minimaps)
  */
 /image/proc/minimap_on_move(atom/movable/source, oldloc)
 	SIGNAL_HANDLER
-	if(isturf(source.loc))
-		pixel_x = MINIMAP_PIXEL_FROM_WORLD(source.x) + SSminimaps.minimaps_by_z["[source.z]"].x_offset
-		pixel_y = MINIMAP_PIXEL_FROM_WORLD(source.y) + SSminimaps.minimaps_by_z["[source.z]"].y_offset
+	if(source.z)
```
boskoramen, line 627: this is a neat way to check if an atom is on a turf, might be worth exporting this check to a helper define

boskoramen (approved):

code/controllers/subsystem/minimap.dm
fira: Yeah i use that everywhere, figure it's simpler and faster.
Technically it can fail in one way, if you nullspace it, you can VV z only to a value and it'll still be nullspaced since x/y are 0. But that's obviously not supposed to happen, just like you aren't supposed to change loc.

PR #12749, opened 2026-07-27
Ferret Heavy Industries Mk4-L Crate Loader

Mary-Marilyn:
# About the pull request

This PR adds the Ferret Heavy Industries Mk4-L Crate Loader to requisitions and the construction area south of the hangar.

The Mk4-L is a faster variant of the standard powerloader (being exactly one move_delay faster), which is unable to haul anything but crates. 

https://github.com/user-attachments/assets/20761dae-0cfb-4461-ae7a-814c0583d53d

If you try to pick up anything else, the powerloader will blow apart. 

https://github.com/user-attachments/assets/11470a6e-ebf7-499c-a0e6-d1dbb10e047a

It will also blow apart if you try to transport it on a dropship, to prevent marines from using it groundside.

https://github.com/user-attachments/assets/1f15153d-4542-4877-8686-e06ccfdb977d

grafik

Also adds a flag that allows non-mob movables to traverse multi-Z stairs. (modularity)

One of my least favorite shipside activities is hauling crates around the ship at a snails pace. This should hopefully make the job of requisitions and the ASO not as tedious.

grafik
grafik

# Changelog
:cl:
add: Added the Ferret Heavy Industries Mk4-L Crate Loader. A faster powerloader which can only haul crates.
code: Added a flag allowing non-mob movables to traverse multi-Z stairs.
maptweak: Added an Mk4-L Crate Loader to the construction area south of the hangar.
/:cl:

### DIFF
--- a/code/__DEFINES/equipment.dm
+++ b/code/__DEFINES/equipment.dm
@@ -52,6 +52,8 @@
 #define MAP_COLOR_INDEX (1<<21)
 /// If an object will fall through open space, use this when dashing \ jumping for example
 #define NO_ZFALL (1<<22)
+/// Allows a non-mob movable to use multi-Z stairs
+#define CAN_USE_MULTIZ_STAIRS (1<<23)
 
 //==========================================================================================
 
--- a/code/game/objects/structures/multiz_stairs.dm
+++ b/code/game/objects/structures/multiz_stairs.dm
@@ -25,7 +25,7 @@
 
 /obj/structure/stairs/multiz/proc/on_stairs_moved(turf/source, atom/movable/enterer)
 	SIGNAL_HANDLER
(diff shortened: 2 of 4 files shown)


Uristthedorf:
Make it so when it blows apart from trying to lift a non-crate item, it tears off a limb. Or maybe break EVERY bone in the body.

Ferret heavy industries must have gone bankrupt for a reason.

sippykot:
you know the cargo train is supposed to be used for that but the functionality got disabled and then not enabled because of the "logistics rework" that's now no longer a thing due to thwomper quitting

Blundir (changes requested):
For source material and aesthetics reasons, keep the power loaders the same as old ones, rather give some ds only loaders to pilots

PR #12758, opened 2026-07-28
Prevents placing eggs near ladders

Mary-Marilyn:
# About the pull request

This prevents xenos from planting eggs within two tiles of ladders.

Placing eggs near ladders is unfair because players arriving through them have little opportunity to see or avoid the eggs before triggering them. 

grafik

Tested ingame, and it worked.

# Changelog
:cl:
balance: Xenos can no longer plant eggs within two tiles of ladders.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/egg_item.dm
+++ b/code/modules/mob/living/carbon/xenomorph/egg_item.dm
@@ -99,6 +99,9 @@
 		return
 	if(!user.check_alien_construction(T, ignore_nest = TRUE))
 		return
+	if(locate(/obj/structure/ladder) in range(2, T))
+		to_chat(user, SPAN_XENOWARNING("We cannot plant [src] so close to a ladder."))
+		return
 	if(!user.check_plasma(30))
 		return
 


DaGitSmog:
If this is the case, then the same should be done with playing mines near ladders. Right now you can prevent any kind of ladder trapping as marine by confirming the eggs with a flare and just tossing a nade into it if you want to move through that route. As Xeno you kinda have to facetank the mine, god forbid there's a fueltank nearby as well.

zzzmike:
> Right now you can prevent any kind of ladder trapping as marine by confirming the eggs with a flare

Yup. The whole 'look at above floor' function was designed for this very situation, right? It feels like this PR will just end up nerfing interesting situations. There won't be moments of 'remember that crazy round where like 3 people got hugged because they went up the ladder without checking first' etc.

ori-disillusion-mirnov:
> Yup. The whole 'look at above floor' function was designed for this very situation, right? It feels like this PR will just end up nerfing interesting situations. There won't be moments of 'remember that crazy round where like 3 people got hugged because they went up the ladder without checking first' etc.

youre so right lets remove the egg placement restriction near stairs too for even more memorable moments of round removal

TheManWithNoHands:
> Placing eggs near ladders is unfair because players arriving through them have little opportunity to see or avoid the eggs before triggering them.

This is just factually wrong. You can look up and down ladders, iirc it was click drag.
Maybe rewrite the pr body a bit.

neeshacark:
as one of the people who was hugged using this in the round that incited this pr, i don't think it's needed. i didn't look up the ladder (even though i heard xenos up there) and got burned. i had multiple other options and i chose to be reckless.

1 flare and one looking up of the ladder is all it takes to not have this happen to you

Detective-Google:
https://forum.cm-ss13.com/w/pr-feedback you should place your arguments here instead.

Git-Nivrak (approved):
Consistent with multiz stairs

PR #12759, opened 2026-07-28
Fix: Onehanding Lever Action chambering runtime

MistChristmas:
# About the pull request
For some reason I glossed over that this is bottled to a signal handler proc. And that it really doesn't need to be in this proc at all.
Not really any player facing effects as the R4T isn't really seen anywhere.
But it now spins in the opposite direction, and uses `animation_spin` instead. For one handed lever actioning.

Runtimes bad. Code nonsensically in the code for when you direct hit something, when it's about chambering a weapon... also bad.
<details>

</details>

# Changelog
:cl:
add: R4T now spins anti clockwise. As spinning it clockwise makes no sense...
code: Clean up some code related to lever action, onehanded chambering
fix: A runtime related to the above.
fix: If your hand is already broken, failing the r4t levering breaks your arm and removes splints from both your hand and arm. As well as interrupting knitting time. Otherwise just does it only for the hand.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/lever_action.dm
+++ b/code/modules/projectiles/guns/lever_action.dm
@@ -278,12 +278,13 @@ their unique feature is that a direct hit will buff your damage and firerate
 		cur_onehand_chance = cur_onehand_chance - 20 //gets steadily worse if you spam it
 	else
 		to_chat(user, SPAN_DANGER("Augh! Your hand catches on the [lever_name]!!"))
+		user.drop_held_item()
 		var/obj/limb/holding_hand = user.get_limb(user.hand ? "l_hand" : "r_hand")
-		if(holding_hand.status & LIMB_BROKEN)
-			holding_hand = user.get_limb(user.hand ? "l_arm" : "r_arm")
-			user.drop_held_item()
-		else
-			holding_hand.fracture()
+		if(holding_hand.status & LIMB_BROKEN) //Hand already broken? Break their arm next
+			var/obj/limb/holding_arm = user.get_limb(user.hand ? "l_arm" : "r_arm")
+			holding_arm.fracture()
+			holding_arm.status &= ~LIMB_SPLINTED
+		holding_hand.fracture()
 		holding_hand.status &= ~LIMB_SPLINTED
 		user.pain.recalculate_pain()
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/guns/lever_action.dm
```diff
@@ -289,8 +269,24 @@ their unique feature is that a direct hit will buff your damage and firerate
...
+		cur_onehand_chance = cur_onehand_chance - 20 //gets steadily worse if you spam it
+	else
+		to_chat(user, SPAN_DANGER("Augh! Your hand catches on the [lever_name]!!"))
+		var/obj/limb/holding_hand = user.get_limb(user.hand ? "l_hand" : "r_hand")
+		if(holding_hand.status & LIMB_BROKEN)
+			holding_hand = user.get_limb(user.hand ? "l_arm" : "r_arm")
+			user.drop_held_item()
+		else
+			holding_hand.fracture()
+		holding_hand.status &= ~LIMB_SPLINTED
```
boskoramen, line 288: way this is setup now, if hand is already broken by this point the splint will be removed from the arm instead of the hand

is this intentional?

code/modules/projectiles/guns/lever_action.dm
```diff
@@ -289,8 +269,24 @@ their unique feature is that a direct hit will buff your damage and firerate
 		return
 	if(flags_gun_lever_action & MOVES_WHEN_LEVERING)
 		to_chat(user, SPAN_WARNING("<i>You spin \the [src] one-handed! Fuck yeah!<i>"))
-		animation_wrist_flick(src)
-	direct_hit_buff(user, projectile_source=src, one_hand_lever=TRUE)
```
boskoramen, line 293: why did this even fail? I'm confused since runtime is about a bad arg name that exists in the parent proc and all child overrides of the proc

boskoramen: asked about this on discord, runtime is strange and doesn't make any sense

don't think this is worth investing time into understanding at this point unless we see a lot more similar runtimes

boskoramen:
lmk when changelog is updated and good to go once analyzers pass

boskoramen (approved):

Git-Nivrak (approved):

PR #12761, opened 2026-07-29
Fixes stack runtimes

Nihisohel:
# About the pull request

oops

glitchtip stop yelling at me

hitting the stack with a non-stack item no longer runtimes it now

# Changelog
:cl:
fix: Fixes a runtime related to attempting to stack a non-stackable item
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -482,7 +482,7 @@ Also change the icon to reflect the amount of sheets, if possible.*/
 	return TRUE
 
 /obj/item/stack/attackby(obj/item/stack/used_stack, mob/user)
-	if(used_stack.stack_id != stack_id) //not the same stack type :)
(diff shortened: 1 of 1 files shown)


code/game/objects/items/stacks/stack.dm
```diff
@@ -482,7 +482,7 @@ Also change the icon to reflect the amount of sheets, if possible.*/
 	return TRUE
 
 /obj/item/stack/attackby(obj/item/stack/used_stack, mob/user)
```
boskoramen, line 484: we really need to normalize the proc headers here so they're not different per override, not relevant for this PR just general comment

boskoramen (approved):

code/game/objects/items/stacks/stack.dm
fira: It kind of happens auto if you use kwargs

fira (approved):

Git-Nivrak (approved):

PR #6650, opened 2024-07-07
Comtech Stuff

Git-Nivrak:
# About the pull request
A bunch of qol and new items for combat technicians

# Explain why it's good for the game
Gives combat techs some new tools, toys and skills to make them feel like an actual role rather than another guy with engineering skill.
# Sprite screenshots
305680653-a9a4818a-83c9-43cf-8a12-a21cc95de856
305680657-5f5f6226-448e-4196-8a20-983da8f718be

</details>

# Changelog
:cl: Nivrak, NervanCatos
balance: Modified engineering skill levels, Combat technicians now do everything engineering related slightly faster. This does not affect any other roles or skillchecks.
add: Adds the tactical compact nailgun to the Combat Technician's vendor essential engineering kit, It cannot fire and uses 2 metal instead of 1 to repair. Credit to NervanCatos.
add: Added the M277 pattern construction rig, It comes with 6 slots instead of 10 in the M276 but can carry metal and plasteel stacks. Available in the CT vendor.
add: Added the Engineer Kit pouch, basically and engineer kit - in a pouch. Restricted by engineering skill. Available in the CT vendor.
add: Added the Small Tool Webbing, A smaller variant of the tool webbing with 6 slots instead of 7. Available in the CT vendor.
qol: Screwdriver, Crowbar, Blowtorch, Multitool, Wrench and Wirecutters will now prioritize the tool webbing when quick equipping.
code: Added a new variable for items, preferred_storage which allows to replicate the above behavior for other items and storages.
add: Added a new sentry upgrade, The omni-sentry. As the name suggests it is omni-directional but has 30% reduces damage and a shorter range by 1.
qol: Made APC examine-text a bit more clear about what you should be doing next.
/:cl:

PR #12764, opened 2026-07-29
Adds Tool Webbing as options for SL, FTL, MT, SO, Crewman and adds some to cargo

sippykot:
# About the pull request
As in the title, adds both of the tool webbings as an option for some marine roles.
Also adds a new tool webbing variant which swaps the normal welder for the MCT and uses that above for the roles except MT and Crewman

Better options for marines to create their loadout. Previously the webbing was extremely exclusive, with the only source being the CT vendor and i think one map spawn? (but it spawns the synth version anyways) The roles affected are either dedicated engineering roles or roles that are frequently expected to provide engineering support.

image

# Changelog

:cl:
add: Adds Tool Webbing as option for SL, FTL, MT, SO, Crewman
add: Adds Tool Webbings to cargo vendor
/:cl:


Git-Nivrak:
It is CT exclusive on purpose, see #6650

PR #12781, opened 2026-07-30
Acid hole direction handling

JackieEstegado:
# About the pull request

Made acid holes have their direction properly set no matter how the mapper decided to varedit them.
At first I thought about just fixing each instance by hand and leaving a couple Crash() lines to make sure mappers dont make the mistakes again, but I rather went with this approach.

Originally the hole's direction was intended to be set by the acided_hole_dir variable on the wall. However mappers didnt use this (probably because it didnt cause the hole to change direction in the map editor).
Expanded it so that if either the acided_hole_dir, dir or icon_state is varedited to anything other than default (which is vertical) then it will set the hole up as being horizontal. 
Hopefully mappers dont invent any more new and exciting ways to try to set the direction.

Also removed two instances of invalid hole placements (which got deleted anyways after init).

Vertical holes in vertical walls bad

Checked one of each varedit variant on each map (nightmares not included) and crawled through them as a drone.
Also melted a horizontal and vertical hole for myself as a boiler, which also worked properly.

<details>
<summary>Before</summary>
image
</details>
<details>
<summary>After</summary>
Screenshot_20260730_130031
</details>

# Changelog
:cl:
fix: Fixed some acid holes on maps having the wrong orientation.
/:cl:

### DIFF
--- a/code/game/objects/effects/acid_hole.dm
+++ b/code/game/objects/effects/acid_hole.dm
@@ -16,10 +16,10 @@
 		holed_wall = acided_wall
 		holed_wall.opacity = FALSE
 		// Original intended way for mappers to set the direction was to set wall.acided_hole_dir, but mappers didnt know that, so we will just handle all ways they could have set it.
-		// Since holes can only have two states (vertical/horizontal) we assume no varedits to be vertical and any edits horizontal.
+		// Since holes can only have two states (vertical/horizontal), default is vertical.
 		if(icon_state != initial(icon_state)) // Mapper varedited icon_state to horizontal, so we update code to reflect it.
 			holed_wall.acided_hole_dir = EAST
-		if(dir != EAST) // Mapper didnt varedit dir, so we will use acided_hole_dir. If he did thats fine that will work.
+		if(dir != EAST && dir != WEST) // Mapper didnt varedit dir to horizontal, so we will use acided_hole_dir.
 			setDir(acided_wall.acided_hole_dir)
 
 /obj/effect/acid_hole/Destroy()
--- a/code/game/objects/effects/acid_hole.dm
+++ b/code/game/objects/effects/acid_hole.dm
@@ -16,8 +16,8 @@
 		holed_wall = W
 		holed_wall.opacity = FALSE
 		// Original intended way for mappers to set the direction was to set wall.acided_hole_dir, but mappers didnt know that, so we will just handle all ways they could have set it.
-		// Since holes can only have two states (S-N, E-W) we assume no varedits to be S-N and any edits W-E.
-		if(icon_state != initial(icon_state)) // Mapper varedited icon_state to east/west, so we update code to reflect it.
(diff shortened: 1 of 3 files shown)


boskoramen (dismissed):
code approved, pending mapper approval

code/game/objects/effects/acid_hole.dm
```diff
@@ -15,7 +15,12 @@
 		W.acided_hole = src
 		holed_wall = W
 		holed_wall.opacity = FALSE
-		setDir(W.acided_hole_dir)
+		// Original intended way for mappers to set the direction was to set wall.acided_hole_dir, but mappers didnt know that, so we will just handle all ways they could have set it.
+		// Since holes can only have two states (vertical/horizontal) we assume no varedits to be vertical and any edits horizontal.
+		if(icon_state != initial(icon_state)) // Mapper varedited icon_state to horizontal, so we update code to reflect it.
+			holed_wall.acided_hole_dir = EAST
+		if(dir != EAST) // Mapper didnt varedit dir, so we will use acided_hole_dir. If he did thats fine that will work.
+			setDir(W.acided_hole_dir)
```
boskoramen: 1 letter var

boskoramen (changes requested):

boskoramen:
Marking as stale exempt til I or someone else reviews.

There is a potential issue I noticed in the code after glancing diff again, need to dedicate some time to test and articulate issue (if it exists).

code/game/objects/effects/acid_hole.dm
```diff
@@ -11,11 +11,16 @@
...
+		var/turf/closed/wall/acided_wall = loc
+		acided_wall.acided_hole = src
+		holed_wall = acided_wall
 		holed_wall.opacity = FALSE
-		setDir(W.acided_hole_dir)
+		// Original intended way for mappers to set the direction was to set wall.acided_hole_dir, but mappers didnt know that, so we will just handle all ways they could have set it.
+		// Since holes can only have two states (vertical/horizontal) we assume no varedits to be vertical and any edits horizontal.
+		if(icon_state != initial(icon_state)) // Mapper varedited icon_state to horizontal, so we update code to reflect it.
+			holed_wall.acided_hole_dir = EAST
+		if(dir != EAST) // Mapper didnt varedit dir, so we will use acided_hole_dir. If he did thats fine that will work.
```
boskoramen: Why are we checking for the EAST direction specifically here? Would mappers (for some reason) ever set dir to WEST via varedit?

boskoramen (changes requested):

code/game/objects/effects/acid_hole.dm
JackieEstegado: Handling for west added. No need for handling north as by default it should work correctly. And it may be hopeful but I trust mappers not to set diagonal directions.

boskoramen (approved):

PR #12788, opened 2026-07-31
Balance: Smart scope gets its scope back, Alt-IFF lockout removal

antlersss:
# About the pull request

Re-buffs the B8 smart scope by:
 - Re-adding its mini-scope component
 - Removing its damage penalty
 - Removing firing delay lockout from alt-iff

NOTE: Removing the firing delay lockout *does not* mean that it acts like regular IFF now. It only means that it will continue to reattempt to fire at the next available shot, instead of applying its own delay ontop of firing delay.

Additionally buffs the mini-scope to give an always-active accuracy buff slightly lower than the red dot, as to not make it objectively worse than the B8.

The B8 was double-nerfed a long time ago by both reducing its damage and introducing alt-iff instead of its former full iff. Both of these nerfs sought to decrease the total effective DPS of a squad of B8's. However, I believe the alt-IFF component was more than enough to make this attachment both unique and non-oppressive. 

As it stands now, the double nerf makes it holistically unappealing as an alternative to options such as the reflex sight and red dot, which offer buffs with very little downside. This PR aims to buff it to be a competitive alternative to the other scope options in the game.

<details>

</details>

# Changelog

:cl:
balance: The B8 Smart Scope no longer applies a damage penalty, and now has its mini-scope component back.
balance: The S4 2x Telescopic Mini-scope now has an accuracy buff when unscoped, slightly lower than the red dot.
balance: Alt-IFF weapons, such as the B8 or M56 Smartguns on frontline mode, no longer apply a half-second lockout when encountering an IFF target in the line of fire, and instead resume shooting at the next firing interval.
/:cl:


ihatethisengine:
Just to clarify, there was never a "double-nerf". Old IFF scope being a subtype of "scope" always had a huge firerate debuff. Replacing it with a simple damage debuff actually increased the DPS. Making it a subtype of "scope" again re-adds the firerate debuff, reducing the DPS drastically (which is basically what made the full IFF scope a notorious noob trap outside of huge squads and some cheese strats like bipod + smartscope). If you really want it to be a scope this much, consider making it a subtype of "scope/mini" (though better still double check other stats it might affect).

10% damage debuff is meant as a trade-off you get for the ability to deal no FF (which kinda indirectly increases your ability to deal damage in a combat situation even more as you don't hesitate to shoot). I don't think the opportunity cost (inability to put RDS, RS or MH) alone is going to cut it really. There are other attachments to get accuracy buffs from (like stock or vertical grip if we are talking m41a).

The extra delay when an ally crosses your firing line is intended to encourage you to still be mindful of your firing line and discourage you from firing blindly.

Sunless2:
Removing both the fire delay and the damage penalty means there is basically no tradeoff to running this, and it will quickly become a meta-pick for the Mk2, especially with aciding not destroying your weapon in one go.

I do like the idea of just removing the damage penalty though.

Though if we were speaking for just the M4RA the buffs you suggested would probably be fine.

sippykot:
> Removing both the fire delay and the damage penalty means there is basically no tradeoff to running this, and it will quickly become a meta-pick for the Mk2, especially with aciding not destroying your weapon in one go.
> 
> I do like the idea of just removing the damage penalty though.
> 
> Though if we were speaking for just the M4RA the buffs you suggested would probably be fine.

there are several attachies that are just straight upgrades
what is the tradeoff for extended barrel? reflex sight? red dot?

as for restricting it to m4ra only, hell no
if anything you should be able to put it on more guns

sippykot:
if you want a downside to the scope, make it add 1-handed scatter and make it a bit slower to 2-hand the gun i guess?

Uristthedorf:
> Though if we were speaking for just the M4RA the buffs you suggested would probably be fine.

Definitely shouldn't exclusive to m4ra gun.

B8 is helpful for privates to play without FFing every other burst.

Ammoniacres:
> > Though if we were speaking for just the M4RA the buffs you suggested would probably be fine.
> 
> Definitely shouldn't exclusive to m4ra gun.
> 
> B8 is helpful for privates to play without FFing every other burst.

The reduction of B8 as a private-only weapon is why it landed in its initial, almost entirely unused state in the first place. The MK2 and its variations are already the training wheels for privates, and the ability to not friendly fire should be acknowledged as the mixed bag as it is rather than the outright benefit it isn't.

sippykot:
> The MK2 and its variations are already the training wheels for privates

this is what plat delta riflemen who die 10 minutes after drop actually believe...

MistChristmas:
Please use other methods to discuss balance implications or anything. Because as we can see it quickly devolves into fruitless comments.

The GitHub comments should be restricted to that of which is actually relevant regarding the code thank you.

cm13-github:

<details><summary>Test Merge Deployed @ 08/11/2026 00:35:06 +00:00:</summary>

##### Server Instance

##### Revision
Origin: 2bc00b4dd22144e6da17752d2058c7b81d535cbc
Pull Request: ce3bec7d952ad48e3ab8c42902fabef18fe2a57e
Server: 657673d82d5618b163b9eabbda12ff45961398d7
</details>
<details><summary>Test Merge Updated @ 08/11/2026 02:06:09 +00:00:</summary>

##### Server Instance
CM13 Live

##### Revision
Origin: 74c04bdc8466a5d03608ce994ce77ce461072443
Pull Request: ce3bec7d952ad48e3ab8c42902fabef18fe2a57e
Server: b539256471b7e3753be0c7d4cecbbfa51d560af0
</details>
<details><summary>Test Merge Removed @ 08/11/2026 07:38:02 +00:00:</summary>

##### Server Instance
CM13 Live
</details>

PR #12480, opened 2026-06-20
Sneaky Hijack Edgecase + Transmute Fixes

Drulikar:
# About the pull request

This PR does several things:
- Makes it where a hostile queen (an allied queen at the start of transit is ignored) that is on a shuttle that's arriving normally to the almayer will trigger a hijack crash (with little warning)
- Fixes some logic errors in #10238 where the loop would end on the first xeno that is a king or already has the ability (rather than continuing on to the next xeno)
- Transmute is no longer given to queen on purchase
- Moves some hijack effects around (namely evilution changes, ovi reset, queen maturity into `/datum/hive_status/proc/bless_on_hijack()`; and hijack round status+ stat tracking into `/datum/controller/subsystem/hijack/proc/on_call_shuttle()`)
- Standardizes the defines for game modes (some did it this way, many not)
- Transmute needs to be a proc verb (or if a verb verb it needs additional ability checks to see if the ability is granted). Ultimately I messed with transmute because of the weird issue with the loop early returning incorrectly where I wanted to move logic to
- **NEW** Queen pry for dropship airlocks is increased from 3s to 5s
- Mapped in dropships using `/obj/structure/machinery/door/airlock/dropship_hatch` now check if the queen is busy - generally to keep consistency with `/obj/structure/machinery/door/airlock/multi_tile/almayer/dropshiprear` and `/obj/structure/machinery/door/airlock/multi_tile/almayer/dropshiprear/dropshipside`

Technically a xeno evolving as queen on the Almayer could still be in this scenario, but I think it even less likely to occur and unlikely for them to succeed enough that warrants needing to put the game into a hijack state.

# Explain why it's good for the game

Presently admins have been sort of having to intervene whenever a queen sneaks onto the dropship, but since hijack effects are kinda all over the place its difficult for them to do all changes nor in a way that really thematically can work. Now its as if the dropship has its autopilot disabled shortly before docking procedures and ends up crashing into the ship (because the queen is a bad driver).

Queen pry increase is from feedback that thinks DP needs more time to react to a queen trying to enter.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Normal hijack + transmute testing: https://youtu.be/CuSb9vs2NiQ

Stealth hijack as marine: https://youtu.be/mL4UUED1mqA

Stealth hijack as xeno: https://youtu.be/u05M-Y6q32E

</details>

# Changelog
:cl: Drathek
[...]

PR #12791, opened 2026-07-31
Fix xeno transmute

SmokedCyprinid:
# About the pull request

-Stops guaranteed slot T2 xenos (hivelord, burrower, carrier)) going over T2 slot capacity by using transmute.

You can still transmute into another available guaranteed slot or if there is an available general slot . You can transmute If you are not a guaranteed slot holder hivelord, burrower, carrier (meaning there are more than 2 of same caste)

This should still work if the free slot numbers are edited or other castes are added.

-Also, stops players from losing transmute by selecting their current caste.

Bug bad, fix good.

<details>

https://github.com/user-attachments/assets/5ff6f057-8cba-461c-9da9-0d377b0441b6

</details>

# Changelog

:cl:labeo0
fix: Guaranteed slot T2 xenos cannot transmute over the T2 slot capacity
fix: Xenos don't lose transmute by selecting their current caste.
/:cl:


fira:
Does this interact with #12480 ?

SmokedCyprinid:
> Does this interact with #12480 ?

I don't see any issues but Drathek could say for sure

code/modules/mob/living/carbon/xenomorph/Evolution.dm
```diff
@@ -509,3 +512,16 @@ GLOBAL_LIST_EMPTY(deevolved_ckeys)
...
 	return TRUE
+
+//checks if transmuting T2 is guaranteed slot holder and prevents them from going over general T2 slot cap
+/mob/living/carbon/xenomorph/proc/can_transmute(caste_pick)
+	if(caste_type == caste_pick)
+		to_chat(src, SPAN_WARNING("You are already [caste_pick]!"))
+		return FALSE
+
+	var/slots = hive.get_tier_slots()
+	if(tier == 2 && (hive.get_caste_count(caste_type) <= hive.free_slots[caste.type]) && !slots[TIER_2][OPEN_SLOTS] && !slots[TIER_2][GUARANTEED_SLOTS][caste_pick])
```
fira, line 530: Did you mean to write caste.type rather than caste_type here?

code/modules/mob/living/carbon/xenomorph/Evolution.dm
```diff
@@ -509,3 +512,16 @@ GLOBAL_LIST_EMPTY(deevolved_ckeys)
 		return FALSE
 
 	return TRUE
+
+//checks if transmuting T2 is guaranteed slot holder and prevents them from going over general T2 slot cap
+/mob/living/carbon/xenomorph/proc/can_transmute(caste_pick)
+	if(caste_type == caste_pick)
+		to_chat(src, SPAN_WARNING("You are already [caste_pick]!"))
```
fira: Xeno messages use plural for these. We are already....

code/modules/mob/living/carbon/xenomorph/Evolution.dm
```diff
@@ -509,3 +512,16 @@ GLOBAL_LIST_EMPTY(deevolved_ckeys)
...
+		return FALSE
+
+	var/slots = hive.get_tier_slots()
+	if(tier == 2 && (hive.get_caste_count(caste_type) <= hive.free_slots[caste.type]) && !slots[TIER_2][OPEN_SLOTS] && !slots[TIER_2][GUARANTEED_SLOTS][caste_pick])
+		to_chat(src, SPAN_WARNING("The hive cannot support another Tier 2 of this caste, wait for either more aliens to be born or someone to die."))
```
fira: Same here, could make this message a little less gamey

fira (dismissed):

code/modules/mob/living/carbon/xenomorph/Evolution.dm
SmokedCyprinid: done

code/modules/mob/living/carbon/xenomorph/Evolution.dm
SmokedCyprinid: nah, "caste_type" is a string with caste name and that's what get_caste_count wants. "Caste" is a datum holding all caste stats and I need it's type to find if there are guaranteed slots in the list

code/modules/mob/living/carbon/xenomorph/Evolution.dm
SmokedCyprinid: any suggestions? I just copied it from can_evolve pretty much. Something like "We cannot support another Tier 2 of this caste, wait for more sisters to be born or someone to die."?

code/modules/mob/living/carbon/xenomorph/Evolution.dm
SmokedCyprinid: to elaborate on it. I am getting the count of the current caste and and by checking free_slots list I find if and how many free slots it has. 

if current caste number is less or equal the guaranteed slots number, it means they are taking the guaranteed free slot and we need to check if there are general slots open.

if current caste num is more than guaranteed slots number, it means that xeno takes up a general slot and we don't need a check. It can transform into anything.

if it's not in the free_slot list, it means it takes regular slot and can transform into anything.

code/modules/mob/living/carbon/xenomorph/Evolution.dm
fira: I have no idea honestly :clueless:
Good enough probably

Drulikar (changes requested):
Should interact fine with my changes (not tested it) but I don't think your logic is accounting for what slot they currently may be using. It looks like this would prevent a t2 from transmuting to another t2 because their current slot isn't being accounted for as being free yet.

SmokedCyprinid:
> Should interact fine with my changes (not tested it) but I don't think your logic is accounting for what slot they currently may be using. It looks like this would prevent a t2 from transmuting to another t2 because their current slot isn't being accounted for as being free yet.

It only checks for open slots if t2 specifically using the guaranteed free slot. otherwise t2 can transform. 
Checkout the test video at 00:44 lurker transforming into warrior even if remaining slots show 0

SmokedCyprinid:
T2 fails to transform by occupying guaranteed slot and having no empty slot or other empty guaranteed slot.
T2 considered occupying guaranteed slot if the current caste is in free_slot list and number of T2`s caste in the hive is equals or less the number on the list (default 1)

So if there are two hivelords and there are 0 T2 slots then one of them can transform since the slot is a part of general T2 slot capacity. But after the first transforms the second hivelord needs an empty slot. Otherwise queen could keep buying transmute, guaranteed T2s transform into other T2s, T1 evolve into T2 with guaranteed slots and repeat until all T1s become T2s

Drulikar (approved):

PR #12792, opened 2026-07-31
Fixes message

zzzmike:
# About the pull request

Fixes a message telling everyone nearby that they pried open a control valve. I guess the argument is in the wrong order and just needs to be reversed, but I think the code is more easily understood the way I did it now, so I did it this way. And also it's already done this way in some places anyways.

Edit: fira sold me on doing it the other way 🫡

Messages should make sense

<details>

</details>

# Changelog
:cl:
fix: fixed a message telling everyone nearby that they pried a valve
/:cl:

### DIFF
--- a/code/game/objects/structures/props.dm
+++ b/code/game/objects/structures/props.dm
@@ -302,7 +302,8 @@
 		return
 	else if (ishuman(user) && HAS_TRAIT(W, TRAIT_TOOL_CROWBAR))
 		on = !on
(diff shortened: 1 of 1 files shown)


fira:
Well the problem with this vs using only visible_message is that the user gets two messages

zzzmike:
> Well the problem with this vs using only visible_message is that the user gets two messages

ok im sold, changed it 🫡

fira (approved):

PR #12805, opened 2026-08-01
Revert "Hairs adjustments"

Red-byte3D:
image

reverts #12755 at the request of the pr maker and blundir.

🆑
del: Revert 'Hairs adjustments'
/:cl:


fira:
That's not a lot of context

PoltavskaPraca:
> That's not a lot of context

i thought that some of the adjusted sprites looked fine in the end, but no.

there's also fuckery with some hairs
like east/west is 1px shorter than north/south and the upper pixels don't align, so like with bedhead hair 3s/2s when you rotate, your sprite gets squished by 1px

after this is merged, i'm gonna push the pr that touches 3 hairstyles where i'm 100% sure it needs fixing, although i'd say some just need to be resptited.

Steelpoint (approved):

PR #12807, opened 2026-08-01
Hijack: Allows lifeboats and dropships to depart while crashing

antlersss:
# About the pull request

Now allows lifeboats and dropships depart while the mainship is in the process of crashing to the ground. This means that they can now launch in the period between the "jump drive deactivated due to insufficient fueling" announcement to the "altitude rapidly decreasing" whereas they could not previously.

Holding the lifeboats in the current iteration of hijack is undesireable, as it leads to a game of chicken between the marines and the xenomorphs. Marines want to wait until the last possible objective is up to maximize their chances of evacuating, but once that final objective is gone, they are no longer are able to evacuate.

With this change, holding lifeboats will be more viable as there is now a clear window where it is optimal for marines to begin evacuating.

<details>

image

</details>

# Changelog

:cl:
balance: Dropships and lifeboats can now depart while the main ship is still falling to the ground.
add: Added new ARES announcement text to indicate when lifeboats are disabled.
/:cl:


Drulikar (changes requested):
In the condition of `SShijack.hijack_status == HIJACK_OBJECTIVES_GROUND_CRASH` and that the escape pod/lifeboat also crash to the ground, you should avoid the coordinates between the ground origin just picked and the template's bounds. Otherwise you can have a lifeboat/pod crashing under the almayer or worse into the already crashed almayer somehow carving it like cake.

antlersss:
I believe the Almayer smushing escape pods was something already present, but it couldn't hurt to add that as a oversight fix. Will do.

Drulikar:
> I believe the Almayer smushing escape pods was something already present, but it couldn't hurt to add that as a oversight fix. Will do.

It's sort of inevitable for the ones that drop prior. But here you now know where it will be landing. And I especially do not want a lifeboat carving out the almayer.

code/modules/shuttle/shuttles/crashable/crashable.dm
```diff
@@ -112,6 +112,17 @@
 		if(istype(found_turf, /turf/closed/shuttle))
 			return FALSE
 
+		// Prevent the shuttle from being smushed by the shipmap and vice-versa
+		if (SShijack.hijack_status == HIJACK_OBJECTIVES_GROUND_CRASH && SShijack.crashed_offset_x != null && SShijack.crashed_offset_y != null)
```
Drulikar: `crashed_offset_x` and `crashed_offset_y` *never* can be null currently; and 0 != null

Image

https://github.com/cmss13-devs/cmss13/blob/ee396c5ad7ad43c2aece267ad084fbf6ca61dc6f/code/controllers/subsystem/hijack.dm#L101-L105

You would need to set their defaults to null for this to even be worth checking.
```suggestion
		if(SShijack.hijack_status == HIJACK_OBJECTIVES_GROUND_CRASH && SShijack.crashed_offset_x != null && SShijack.crashed_offset_y != null)
```

code/modules/shuttle/shuttles/crashable/crashable.dm
```diff
@@ -112,6 +112,17 @@
...
+			var/above_min_x = found_turf.x >= SShijack.crashed_offset_x + SShijack.open_space_bounds[OPEN_SPACE_BOUNDS_MINX]
+			var/below_max_x = found_turf.x <= SShijack.crashed_offset_x + SShijack.open_space_bounds[OPEN_SPACE_BOUNDS_MAXX]
+			var/above_min_y = found_turf.y >= SShijack.crashed_offset_y + SShijack.open_space_bounds[OPEN_SPACE_BOUNDS_MINY]
+			var/below_max_y = found_turf.y <= SShijack.crashed_offset_y + SShijack.open_space_bounds[OPEN_SPACE_BOUNDS_MAXY]
+
+			// Shuttle would land inside the shipmap or vice versa
+			if (above_min_x && below_max_x && above_min_y && below_max_y)
```
Drulikar:
```suggestion
			if(above_min_x && below_max_x && above_min_y && below_max_y)
```

code/modules/shuttle/shuttles/crashable/crashable.dm
```diff
@@ -112,6 +112,17 @@
 		if(istype(found_turf, /turf/closed/shuttle))
 			return FALSE
 
+		// Prevent the shuttle from being smushed by the shipmap and vice-versa
+		if (SShijack.hijack_status == HIJACK_OBJECTIVES_GROUND_CRASH && SShijack.crashed_offset_x != null && SShijack.crashed_offset_y != null)
+			var/above_min_x = found_turf.x >= SShijack.crashed_offset_x + SShijack.open_space_bounds[OPEN_SPACE_BOUNDS_MINX]
+			var/below_max_x = found_turf.x <= SShijack.crashed_offset_x + SShijack.open_space_bounds[OPEN_SPACE_BOUNDS_MAXX]
+			var/above_min_y = found_turf.y >= SShijack.crashed_offset_y + SShijack.open_space_bounds[OPEN_SPACE_BOUNDS_MINY]
+			var/below_max_y = found_turf.y <= SShijack.crashed_offset_y + SShijack.open_space_bounds[OPEN_SPACE_BOUNDS_MAXY]
```
Drulikar: `open_space_bounds` are drastically larger than the crash site (they're bounds to incorporate binocular distance out of any almayer window) so drastically larger than what you're wanting here (you'd essentially be requiring a shuttle to crash nearly completely outside of even binocular range of the crash site). The template width/height (if it even exists - such as the case of people testing new shipmaps) will be closer though not perfect. A perfect calculation would require the template to be placed and then you just check the area type (since it will no-op the area of the tiles it is leaving alone).

Drulikar (changes requested):

code/modules/shuttle/shuttles/crashable/crashable.dm
```diff
@@ -112,6 +122,25 @@
 		if(istype(found_turf, /turf/closed/shuttle))
 			return FALSE
 
+		// Prevent the shuttle from being smushed by the shipmap and vice-versa
+		if (SShijack.hijack_status == HIJACK_OBJECTIVES_GROUND_CRASH)
```
Drulikar:
```suggestion
		if(SShijack.hijack_status == HIJACK_OBJECTIVES_GROUND_CRASH)
```

code/modules/shuttle/shuttles/crashable/crashable.dm
```diff
@@ -112,6 +122,25 @@
...
+			if (!template)
```
Drulikar:
```suggestion
			if(!template)
```

code/modules/shuttle/shuttles/crashable/crashable.dm
```diff
@@ -112,6 +122,25 @@
...
+				// Template failed to load; to prevent this from blocking all shuttles from crashing, land anyway
+				. = TRUE
+				stack_trace("Shipmap template failed to load when determining a safe position for a shuttle to crash at.")
+			if (!ground_origin)
```
Drulikar:
```suggestion
			if(!ground_origin)
```

code/modules/shuttle/shuttles/crashable/crashable.dm
```diff
@@ -112,6 +122,25 @@
 		if(istype(found_turf, /turf/closed/shuttle))
 			return FALSE
 
+		// Prevent the shuttle from being smushed by the shipmap and vice-versa
+		if (SShijack.hijack_status == HIJACK_OBJECTIVES_GROUND_CRASH)
+			if (!template)
+				// Template failed to load; to prevent this from blocking all shuttles from crashing, land anyway
+				. = TRUE
+				stack_trace("Shipmap template failed to load when determining a safe position for a shuttle to crash at.")
```
Drulikar: No point stack_tracing when you know its going to runtime on line 136.
```suggestion
				CRASH("Shipmap template failed to load when determining a safe position for a shuttle to crash at.")
```

code/modules/shuttle/shuttles/crashable/crashable.dm
```diff
@@ -112,6 +122,25 @@
...
+			if (!ground_origin)
+				. = TRUE
+				stack_trace("Crash origin landmark failed to load when determining a safe position for a shuttle to crash at.")
```
Drulikar: No point stack_tracing when you know its going to runtime on line 135. Also its message is not correct (see below).
```suggestion
				CRASH("Crash origin landmark failed to load when determining a safe position for a shuttle to crash at.")
```

code/modules/shuttle/shuttles/crashable/crashable.dm
```diff
@@ -96,6 +96,16 @@
...
 /obj/docking_port/mobile/crashable/proc/check_crash_point(obj/docking_port/stationary/crashable/checked_crashable_port)
+	// Pre-laod shipmap template and origin if the shipmap has crashed
+	var/datum/map_template/template
+	var/turf/ground_origin
+	if (SShijack.hijack_status == HIJACK_OBJECTIVES_GROUND_CRASH)
+		var/datum/map_config/ship_map_config = SSmapping.configs[SHIP_MAP]
+		template = SSmapping.map_templates[ship_map_config?.ground_crash_template_name]
+
+		var/obj/effect/landmark/mainship_crashsite/origin_landmark = locate() in GLOB.landmarks_list
+		ground_origin = get_turf(origin_landmark)
```
Drulikar: This landmark is not a guaranteed to exist on maps (in fact its only on a couple). Its a landmark to override the automatic logic. But this logic is unnecessary because when `initiate_ground_crash` is called, thats when `hijack_status = HIJACK_OBJECTIVES_GROUND_CRASH` occurs so this position is already calculated at that point IF it didn't abort (it only iterates 50x and 50y automatically w/o a landmark)

Just store the origin in SShijack instead I guess and then access that

Drulikar (changes requested):

code/modules/shuttle/shuttles/crashable/crashable.dm
```diff
@@ -96,6 +96,12 @@
 
 /// Checks for anything that may get in the way of a crash, returns FALSE if there is something in the way or is out of bounds
 /obj/docking_port/mobile/crashable/proc/check_crash_point(obj/docking_port/stationary/crashable/checked_crashable_port)
+	// Pre-laod shipmap template and origin if the shipmap has crashed
```
Drulikar:
```suggestion
	// Pre-load shipmap template if the shipmap has crashed
```

code/modules/shuttle/shuttles/crashable/crashable.dm
```diff
@@ -96,6 +96,12 @@
...
+	var/datum/map_template/template
+	if (SShijack.hijack_status == HIJACK_OBJECTIVES_GROUND_CRASH)
```
Drulikar:
```suggestion
	if(SShijack.hijack_status == HIJACK_OBJECTIVES_GROUND_CRASH)
```

code/modules/shuttle/shuttles/crashable/crashable.dm
```diff
@@ -112,6 +118,22 @@
 		if(istype(found_turf, /turf/closed/shuttle))
 			return FALSE
 
+		// Prevent the shuttle from being smushed by the shipmap and vice-versa
+		if(SShijack.hijack_status == HIJACK_OBJECTIVES_GROUND_CRASH)
+			if(!template)
```
Drulikar:
```suggestion
			if(!template || !SShijack.ground_origin)
```
Comment/message should still be true because in the event of no ground_origin SShiack will call CRASH too instead of loading a template.

code/modules/shuttle/computer.dm
```diff
@@ -503,7 +503,7 @@
 						to_chat(user, SPAN_NOTICE("[src]'s screen says \"Unable to launch, hyperdrive active\"."))
 						return
 
-					if(SShijack.crashed || SShijack.hijack_status == HIJACK_OBJECTIVES_GROUND_CRASH)
+					if(SShijack.crashed)
```
Drulikar: This now only occurs when its FULLY crashed.

code/controllers/subsystem/hijack.dm
```diff
@@ -760,6 +758,12 @@ SUBSYSTEM_DEF(hijack)
 		drop = FALSE,
 	)
 
+	// Break all shipside ships and disable all non-pod/elevator pads
+	unlock_all_dropship_doors() // Unlock doors because they'll be uninteractable
+	disallow_dropship_launching()
+	disallow_dropship_pad_landing()
+	shipwide_ai_announcement("ALERT: Lifeboat telemetry equipment destroyed. Cause: Atmospheric reentry.\n\nEvacuation via port and starboard lifeboats is no longer possible.", HIJACK_ANNOUNCE, sound('sound/effects/creak1.ogg'))
```
Drulikar, line 770: So these changes have no effect on lifeboats currently making the timing of this message incorrect. (and its not simply that I kept the window open)

Image

Drulikar (changes requested):

antlersss:
Surely this will be the final time I forget to change something. :clueless:

Drulikar (approved):

PR #12769, opened 2026-07-29
Possible fix for bloody feet flaky test

Drulikar:
# About the pull request

This PR does a few things:
- The clear_blood timer now hands a weakref to the mob instead of a hardref (I think normally things would handle a hardref in a timer because of the `src` of the timer getting deleted, but I guess elements don't get deleted soo)
- Adds a conscious check when a human is crossing blood and moving
- Variable cleanup

# Explain why it's good for the game

Should fix #12722 and should fix #12770 or we'll see it come up again.

> REF SEARCH Found /mob/living/carbon/human [0x300008b] in list Datums -> /datum/callback [0x210310a4] (obj: /datum/element/bloody_feet proc: clear_blood args: ["the unknown"] -> arguments (list).

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Murder 
image

</details>

# Changelog
:cl: Drathek
fix: You now need to be conscious for bloody footprints to apply
code: Cleaned up code regarding bloody footprints and hopefully eliminated the cause of flaky tests
/:cl:

PR #12813, opened 2026-08-02
Sleep Linter fixes - Chapter 1

fira:
# About the pull request

Last update to SpacemanDMM / Dreamchecker expands on the sleep detection from SHOULD_NOT_SLEEP linter directive.

Unfortunately it is not perfect, and by nature of our broken game static analysis is bound to pick on false positives. By expanding the detection, it uncovers things that have been missed, but there are also more false positives.

This solves the low hanging fruits in Linter warnings with the new version, from 43 errors down to 21. 

The vast majority of what remains will require a refactor of attachables and actions. They almost all amount to the following call chain:  item dropped --> actions/attachies removed --> actions/attachies disabled --> action/attachies activated as a way to toggle it off. Because the action of activating it usually has other side effects such as a windup or user input, this causes sleeping errors.

Untested, too much random stuff. TM it.

# Changelog

:cl:
del: Xenos cannot remove books from bookcases to put them on the floor anymore.
/:cl:

### DIFF
--- a/code/game/objects/effects/step_triggers.dm
+++ b/code/game/objects/effects/step_triggers.dm
@@ -20,6 +20,9 @@
 		return
 	if(istype(crossed_by, /mob/dead/observer) && !affect_ghosts)
 		return
+	INVOKE_ASYNC(src, PROC_REF(PreTrigger), crossed_by)
+
+/obj/effect/step_trigger/proc/PreTrigger(atom/movable/crossed_by)
 	/// If we have a proc to validate trigger, call it to see whether trigger can occur
 	if(can_trigger_proc_ref && !call(src, can_trigger_proc_ref)(crossed_by))
 		return
--- a/code/game/objects/effects/step_triggers.dm
+++ b/code/game/objects/effects/step_triggers.dm
@@ -20,9 +20,9 @@
 		return
 	if(istype(crossed_by, /mob/dead/observer) && !affect_ghosts)
 		return
-	INVOKE_ASYNC(src, PROC_REF(PreTrigger), crossed_by)
+	INVOKE_ASYNC(src, PROC_REF(pretrigger), crossed_by)
 
-/obj/effect/step_trigger/proc/PreTrigger(atom/movable/crossed_by)
+/obj/effect/step_trigger/proc/pretrigger(atom/movable/crossed_by)
 	/// If we have a proc to validate trigger, call it to see whether trigger can occur
 	if(can_trigger_proc_ref && !call(src, can_trigger_proc_ref)(crossed_by))
 		return
--- a/code/game/objects/items/backpack_sprayers.dm
+++ b/code/game/objects/items/backpack_sprayers.dm
@@ -365,10 +365,11 @@
 			var/obj/effect/resin_container/ball = new (get_turf(src))
 			playsound(src,'sound/items/syringeproj.ogg',40,TRUE)
 			//projectile movement
+			var/step_delay = 0
 			for(var/i in 1 to 5)
-				step_towards(ball, target)
-				sleep(2)
-			ball.Smoke()
+				addtimer(CALLBACK(src, PROC_REF(move_ball), ball, target), 0.2 SECONDS * step_delay)
+				step_delay++
+			addtimer(CALLBACK(src, PROC_REF(smoke_ball), ball), 0.2 SECONDS * step_delay)
 			return
 
 		if(METAL_FOAM)
@@ -393,6 +394,11 @@
 			foam.set_up(0, target, null, metal_foam = FOAM_METAL_TYPE_ALUMINIUM)
 			foam.start()
 			return
+
+/obj/item/reagent_container/spray/mister/atmos/proc/move_ball(obj/effect/resin_container/ball, target)
+	step_towards(ball, target)
+/obj/item/reagent_container/spray/mister/atmos/proc/smoke_ball(obj/effect/resin_container/ball)
+	ball.Smoke()
 
 /obj/effect/resin_container //the projectile that the launcher fires
 	name = "metal foam ball"
--- a/code/game/objects/items/implants/implant.dm
+++ b/code/game/objects/items/implants/implant.dm
@@ -184,7 +184,6 @@ Implant Specifics:<BR>"}
 					var/malf_msg = "Something beeps inside [imp_in][part ? "'s [part.display_name]" : ""]!"
 					imp_in.visible_message(SPAN_DANGER(malf_msg))
 					playsound(loc, 'sound/items/countdown.ogg', 25, 1, 6)
-					sleep(25)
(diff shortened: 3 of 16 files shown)


code/game/objects/effects/step_triggers.dm
```diff
@@ -20,6 +20,9 @@
 		return
 	if(istype(crossed_by, /mob/dead/observer) && !affect_ghosts)
 		return
+	INVOKE_ASYNC(src, PROC_REF(PreTrigger), crossed_by)
+
+/obj/effect/step_trigger/proc/PreTrigger(atom/movable/crossed_by)
```
boskoramen:
```suggestion
/obj/effect/step_trigger/proc/pretrigger(atom/movable/crossed_by)
```
Don't need to update other proc names for consistency, but we shouldn't be introducing new procs that break standards.

code/game/objects/items/backpack_sprayers.dm
```diff
@@ -365,10 +365,11 @@
 			var/obj/effect/resin_container/ball = new (get_turf(src))
 			playsound(src,'sound/items/syringeproj.ogg',40,TRUE)
 			//projectile movement
+			var/step_delay = 0
 			for(var/i in 1 to 5)
-				step_towards(ball, target)
-				sleep(2)
-			ball.Smoke()
+				addtimer(CALLBACK(src, PROC_REF(move_ball), ball, target), 0.2 SECONDS * step_delay)
+				step_delay++
```
boskoramen, line 371: Why do we keep incrementing step delay?

If intention is to maintain existing behavior think it would make more sense to have move_ball contain the addtimer call and include iter count as a parameter in that proc

e.g,

```dm
/proc/move_ball(ball, target, iter)
    // do existing behavior
    if (iter <= 5)
        // addtimer for move_ball callback with it +=1
    else if (iter == 6)
        // addtimer for smoke
```

code/game/objects/items/implants/implant.dm
```diff
@@ -185,16 +185,8 @@ Implant Specifics:<BR>"}
 					imp_in.visible_message(SPAN_DANGER(malf_msg))
 					playsound(loc, 'sound/items/countdown.ogg', 25, 1, 6)
 					sleep(25)
```
boskoramen: Remove this?

code/modules/defenses/defenses.dm
```diff
@@ -433,10 +433,11 @@
 	visible_message("[icon2html(src, viewers(src))] [SPAN_WARNING("The [name] starts to blink rapidly!")]")
 	playsound(loc, 'sound/mecha/critdestrsyndi.ogg', 25, 1)
 
-	sleep(5)
+	addtimer(CALLBACK(src, PROC_REF(final_destroyed_action)), 0.5 SECONDS)
 
-	cell_explosion(loc, 10, 10, EXPLOSION_FALLOFF_SHAPE_LINEAR, null, create_cause_data("defense explosion"))
+/obj/structure/machinery/defenses/proc/final_destroyed_action()
```
boskoramen, line 438: Think the cause data needs to be passed as an argument, since for some the procs that use this have "sentry explosion" as the reason rather than "defense explosion" or pass other args to the create cause data proc

boskoramen (changes requested):

code/game/objects/items/implants/implant.dm
fira: Woops

code/game/objects/items/backpack_sprayers.dm
fira: Well it's the same thing no?
step_delay is just incremented separately so it can reach 6

code/modules/defenses/defenses.dm
fira: I saw that but frankly I don't think it being written defense/sentry explosion will be even noticed by anyone. I can change it but i think it's just adding complexity for no real reason.

boskoramen: Only thing of concern is the sentry explosions passing their owner as the causing mob, but that's funky any way since they didn't necessarily cause the explosion itself. I'm OK with this.

code/game/objects/items/backpack_sprayers.dm
boskoramen: fair enough, and clarified how timing works with SStimer so no real concerns with lag and such on discord: https://discord.com/channels/150315577943130112/746346213443633232/1533463485764206712

code/game/objects/items/implants/implant.dm
```diff
@@ -184,17 +184,8 @@ Implant Specifics:<BR>"}
...
-						istype(part,/obj/limb/groin) || \
-						istype(part,/obj/limb/head))
-						part.createwound(BRUISE, 60) //mangle them instead
-						explosion(get_turf(imp_in), -1, -1, 2, 3)
-						qdel(src)
-					else
-						explosion(get_turf(imp_in), -1, -1, 2, 3)
-						part.droplimb(0, 0, "dismemberment")
-						qdel(src)
+					addtimer(CALLBACK(PROC_REF(kaboom), part), 2.5 SECONDS)
```
boskoramen, line 187: oops missed this in first pass thru, but could we use a weakref here instead of a hardref? I'm not super well versed with how GC works any more, so feel free to educate me, but wondering if this would prevent any potential harddels similar to solution from this PR: https://github.com/cmss13-devs/cmss13/pull/12769/changes

boskoramen (changes requested):

code/game/objects/items/backpack_sprayers.dm
```diff
@@ -365,10 +365,11 @@
 			var/obj/effect/resin_container/ball = new (get_turf(src))
 			playsound(src,'sound/items/syringeproj.ogg',40,TRUE)
 			//projectile movement
+			var/step_delay = 0
 			for(var/i in 1 to 5)
-				step_towards(ball, target)
-				sleep(2)
-			ball.Smoke()
+				addtimer(CALLBACK(src, PROC_REF(move_ball), ball, target), 0.2 SECONDS * step_delay)
```
boskoramen, line 370: see comment above about using hard refs vs weak refs for args passed in timer callbacks

code/game/objects/items/backpack_sprayers.dm
```diff
@@ -365,10 +365,11 @@
...
+				step_delay++
+			addtimer(CALLBACK(src, PROC_REF(smoke_ball), ball), 0.2 SECONDS * step_delay)
```
boskoramen, line 372: see comment above about using hard refs vs weak refs for args passed in timer callbacks

boskoramen (changes requested):

fira:
Passing hard refs in short timers is fine but code needs to guard against using them if the object turn out to be deleted, which i didn't add, so need to fix that

PR #12820, opened 2026-08-03
Adds sound to FTL ERT

zzzmike:
# About the pull request

Adds a warcry sound to when an FTL ert is called.

Having seen the pred hunt ert sound effect in game (from https://github.com/cmss13-devs/cmss13/pull/12119 ), I think it added a really nice touch, and I remember that FTL ert calls (other than maybe the first time?), were completely silent, which results in less people knowing that the ert call even fired. 

Edit: And it's a warcry based on it being a human ERT to repel xenos.

<details>

</details>

# Changelog
:cl:
soundadd: ftl ert calls have a sound
/:cl:

### DIFF
--- a/code/controllers/subsystem/hijack.dm
+++ b/code/controllers/subsystem/hijack.dm
@@ -175,7 +175,7 @@ SUBSYSTEM_DEF(hijack)
 		var/datum/emergency_call/emergency_call = new to_spawn
 		emergency_call.name_of_spawn = /obj/effect/landmark/ert_spawns/umbilical
 		emergency_call.shuttle_id = null
-		emergency_call.alert_sound = 'sound/voice/warcry/male_wryy.ogg'
+		emergency_call.alert_sound = 'sound/items/pred_bracer.ogg'
 		emergency_call.activate(TRUE, FALSE)
 
 		TIMER_COOLDOWN_START(src, COOLDOWN_POSTHIJACK_ERT, 5 MINUTES)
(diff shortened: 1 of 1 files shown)


Detective-Google:
I see no issue with adding a sound but I'm not a fan of the warcry for it, ERT sounds should ideally be fairly unobtrusive little notifications, the current pred ERT with a sound is a good example of a fitting sound.

zzzmike:
I figured since the hijack ERT is so rare that it would be fine for the sound to be a warcry but sure changed

Nihisohel:
dunno if i particularly agree with reusing the pred bracer sfx for distress ERTs, but i dont currently have any bright ideas for another ATM

Detective-Google:
> dunno if i particularly agree with reusing the pred bracer sfx for distress ERTs, but i dont currently have any bright ideas for another ATM

Agreed, the pred bracer sound isn't terrible for it but ideally we could get something else in that'd be better; I was just using it as an example of a sound that worked well.

zzzmike:
fire_crackle to let people know that the ert FIRED!!!!11

zzzmike:
also kind of passes as gunfire which makes more sense for a FTL ERT

zzzmike:
please stale bot

MistChristmas (changes requested):
Not a giant fan of the sound to be honest. But it's not used anywhere else so It would be unique at least.

But anything you pick should also be ~~normalised to at least -8db.~~ - Eh you can ignore that.
This sound is peaking. Multiple times. - Don't ignore this though.

Nihisohel:
stale exempt until you figure out a good sound for it
preferably you grab opinions and suggestions in the cm discords sound dev channel

PR #10411, opened 2025-08-17
Adds a anti macro mou cooldown

kiVts:
# About the pull request
Adds a minor cooldown(1.5 seconds in closing) to opening and closing the breech, most players shouldnt notice it besides not being able to spam it for funny sounds. This will affect some of mou mains and macro(the bad kind) users. opening the breech takes 2 seconds after opening.

# Explain why it's good for the game

# Testing Photographs and Procedure

<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog

:cl:
balance: mou given a minor cooldown to closing and opening the breech. most players shouldn't notice it.
/:cl:

PR #12836, opened 2026-08-03
Tweaks to the MOU anticheat cooldown

Tired-Medic:
# About the pull request

Its a simple change of cooldown from 1.1 to 0.8 seconds.

The MOU shotgun currently feels unrewarding and punishing to use for wrong reasons. The cooldown on it closing and opening is too long which leads to players getting frustrated as they reload faster than the anticheat enables them too even though they are using their mouse and keyboard. It frankly makes the gun much more harder to reload and I myself personally stopped using it because 75% of my reload would not be able to go through because of the cooldown.

The original PRs goal was to simply make it so hotkey users would have to wait 1.1 seconds though it heavily and unexpectedly impacted a lot of noncheaters which was not the point of the PR. https://github.com/cmss13-devs/cmss13/pull/10411

Changing the cooldown to 0.8 seconds would still leave a blockade against cheaters but also not punish the players for simply reloading faster than the original PR creator thought they could.

https://youtu.be/-1Snyqy8pQs

The video shows how much difference simple change of 0,3 seconds makes. The gun feels more smooth in use and you are able to close it after every reload compared to the 1.1 second cooldown where almost every reload was met with the 1.1 second cooldown stopping it from coming through.

# Changelog

:cl: szczypak
balance: Changed the AntiCheat cooldown from 1.1 seconds to 0.8

/:cl:

### DIFF
--- a/code/__DEFINES/guns.dm
+++ b/code/__DEFINES/guns.dm
@@ -3,7 +3,7 @@
 #define TASER_MODE_P "precision"
 #define TASER_MODE_F "free"
 
-#define MOU_ACTION_COOLDOWN 1.1 SECONDS
+#define MOU_ACTION_COOLDOWN 0.8 SECONDS
 
 // Common
 #define AMMO_BAND_COLOR_AP "#1F951F"


kiVts:
The OG pr was done to stop *actual* cheating that would get you banned, it was never meant to curb legitimate gameplay

Detective-Google:
I think this somewhat misrepresents one of the critical reasons some of us supported that original solution; which is that having a gun who's fire rate is effectively as fast as a user can perform inputs is simply not something that we can control the game impact of; the anti-macro aspect was the more important part as the proliferation of reload macros was what caused enough momentum to change it; but fundamentally a gun who's DPS output scales directly with player skill can never be balanced.

The cooldown as it stands provides a limit to the weapons DPS output and at the moment the gun performs if not overperforms at that limit.

To use an analogy, imagine if the M4RA fired as fast as you clicked rather than at a set fire rate, and you would see the issue here-- we don't say the M4RA is clunky because it doesnt fire fast enough, we accept that is how the gun works and part of mastering the weapon is learning and mastering the timings.

That is to say, we've discussed this internally and we don't think it's necessary.

PR #12839, opened 2026-08-04
Fix shooting target

### LINKED ISSUE #11601: Shooting targets don't account for non sprite click hit calc + fall off damage.
### Testmerges

N/A

### Round ID

N/A

### Description of the bug

Title

They should do

1. Spawn is USS Runtime and go to 7 tiles away from them
2. Notice your shots still do full damage even with things that have fall off before 7 tiles.
3. Then move right next to them and notice if you don't click directly on them, they don't get hit.

### Issue Bingo

SmokedCyprinid:
# About the pull request

Fix #11601 

-Shooting target now displays damage correctly. 
-Bullets hit target without spriteclick. As a result you can't walk through a target but you can unanchor it to pass and xeno can break it.

Bug bad, fix good.

<details>
<summary>Testing video inside</summary>

https://github.com/user-attachments/assets/bafe0e96-e126-4134-97e6-9b19c147a7f1

</details>

# Changelog
:cl:labeo0
fix: Shooting target damage display now accounts for damage falloff.
fix: Bullets hit shooting target without spriteclick too. As a result you can't walk through a target but you can unanchor it to pass and xeno can break it.

/:cl:

### DIFF
--- a/code/game/objects/structures/misc.dm
+++ b/code/game/objects/structures/misc.dm
@@ -172,7 +172,7 @@
 
 /obj/structure/target/toggle_anchored(obj/item/wrench, mob/user)
 	//no need for movable,unbreakable bullet sponge
-	. - ..()
+	. = ..()
 	if(.)
 		density = !density
 	return .
--- a/code/game/objects/structures/misc.dm
+++ b/code/game/objects/structures/misc.dm
@@ -157,8 +157,8 @@
 		deconstruct(FALSE)
 		return XENO_ATTACK_ACTION
 	else
-		attack_hand(xeno)
-		return XENO_NONCOMBAT_ACTION
+		to_chat(xeno, SPAN_WARNING("We stare at \the [src] cluelessly."))
+		return XENO_NO_DELAY_ACTION
 
 /obj/structure/target/handle_tail_stab(mob/living/carbon/xenomorph/xeno, blunt_stab)
 	if(unslashable)
--- a/code/game/objects/structures/misc.dm
+++ b/code/game/objects/structures/misc.dm
@@ -175,6 +175,7 @@
 	. - ..()
 	if(.)
 		density = !density
+	return .
 
 
 /obj/structure/target/syndicate
--- a/code/game/objects/structures/misc.dm
+++ b/code/game/objects/structures/misc.dm
@@ -146,6 +146,35 @@
 	langchat_speech("[src] raises back into position!", get_mobs_in_view(7, src) , GLOB.all_languages, skip_language_check = TRUE, additional_styles = list("langchat_small"))
 	practice_health = practice_mode[1]
 
+/obj/structure/target/attack_alien(mob/living/carbon/xenomorph/xeno)
+	if(xeno.a_intent == INTENT_HARM)
+		if(unslashable)
+			return
+		xeno.animation_attack_on(src)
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/misc.dm
```diff
@@ -146,6 +146,35 @@
...
+		if(unslashable)
+			return
+		xeno.animation_attack_on(src)
+		playsound(loc, 'sound/effects/metalhit.ogg', 25, 1)
+		xeno.visible_message(SPAN_DANGER("[xeno] slices [src] apart!"),
+		SPAN_DANGER("We slice [src] apart!"), null, 5, CHAT_TYPE_XENO_COMBAT)
+		deconstruct(FALSE)
+		return XENO_ATTACK_ACTION
+	else
+		to_chat(xeno, SPAN_WARNING("We stare at \the [src] cluelessly."))
```
kiVts:
```suggestion
		to_chat(xeno, SPAN_WARNING("We stare at [src] cluelessly."))
```

code/game/objects/structures/misc.dm
```diff
@@ -146,6 +146,35 @@
...
+	if(unslashable)
+		return TAILSTAB_COOLDOWN_NONE
+	playsound(src, 'sound/effects/metalhit.ogg', 25, 1)
+	deconstruct(FALSE)
+	xeno.visible_message(SPAN_DANGER("[xeno] destroys [src] with its tail!"),
+	SPAN_DANGER("We destroy [src] with our tail!"), null, 5, CHAT_TYPE_XENO_COMBAT)
+	xeno.tail_stab_animation(src, blunt_stab)
+	return TAILSTAB_COOLDOWN_NORMAL
+
+/obj/structure/target/toggle_anchored(obj/item/W, mob/user)
```
kiVts:
```suggestion
/obj/structure/target/toggle_anchored(obj/item/wrench, mob/user)
```

code/game/objects/structures/misc.dm
```diff
@@ -146,6 +146,35 @@
...
+	//no need for immovable,unbreakable bullet sponge
+	density = !density
+	. = ..()
```
kiVts, line 175: you should be respecting what the parent returns and not proceed with the density flip if it returns false.

kiVts (changes requested):

code/game/objects/structures/misc.dm
SmokedCyprinid: done

code/game/objects/structures/misc.dm
SmokedCyprinid: done

code/game/objects/structures/misc.dm
SmokedCyprinid: something like this?

kiVts: Nope, thats just plain wrong and the game doesn't even compile anymore
e.g.: 
```
/obj/structure/machinery/hybrisa/coffee_machine/toggle_anchored(obj/item/W, mob/user)
	. = ..()
	if(.)
		update_icon()
```

kiVts: also, ping on discord for things like these.

kiVts (approved):

PR #10802, opened 2025-10-04
removes f1 for adminhelp hardcoded keybind

PoltavskaPraca:
# About the pull request
IRRELEVANT - OLD PR CHANGES
<details>
+ Title + ahelp has a default F1 keybind now
+ Ahelp F1 hardcode and TAB switching between chatbox entry field and the actual game is gone now
+ TAB is now recognized as an actual fucking key by the game to which you can bind stuff
+ In addition to that, it also disables... options & messages, which is... not good? IDK if it's ever actually fucking used by anyone
</details>
### RELVANT
+ removes f1 hardcoded keybind
+ removes f2 hardcoded keybind, but disables the options & messages menu

# Explain why it's good for the game

nihisohel asked me to

# Testing Photographs and Procedure

yes it works

https://github.com/user-attachments/assets/34793e69-e845-44dd-9ee8-46a7d7fd284f

https://github.com/user-attachments/assets/62fc388d-5a0d-45b0-950a-b1cc71386874

# Changelog

:cl:
del: F1 is no longer hardcoded to open adminhelp
/:cl:

PR #12847, opened 2026-08-05
Revert "removes f1 for adminhelp hardcoded keybind"

PoltavskaPraca:
# About the pull request

Reverts cmss13-devs/cmss13#10802 via working around control freak but otherwise the same aa messages and options still works

<details>

[Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.](https://discord.com/channels/150315577943130112/964684928161808384/1534552769115983962)

</details>

# Changelog

:cl:
code: no more freaky control freaks, f1 and f2 keys are now unhardcoded
/:cl:

### DIFF
--- a/code/controllers/subsystem/input.dm
+++ b/code/controllers/subsystem/input.dm
@@ -24,15 +24,10 @@ SUBSYSTEM_DEF(input)
 		"Any" = "\"KeyDown \[\[*\]\]\"",
 		"Any+UP" = "\"KeyUp \[\[*\]\]\"",
 		"Back" = "\".winset \\\"input.text=\\\"\\\"\\\"\"",
-		"F1" = "dummy", //overriding .options menu being triggered by assigning F1 to proc a verb that actually does nothing
+		"F1" = "adminhelp", // Need to unbind F1 because by default, it is bound to .options
 		"CTRL+SHIFT+F1+REP" = ".options",
 		"Escape" = "Open-Escape-Menu",
 		)
-
-CLIENT_VERB(dummy) // idk where else to put it
-	set category = null
-	set name = "dummy"
-	set hidden = TRUE
 
 // Badmins just wanna have fun ♪
 /datum/controller/subsystem/input/proc/refresh_client_macro_sets()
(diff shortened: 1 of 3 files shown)


kugamo:
F1 is usually a controls help page in games conventionally anyways 👍

code/controllers/subsystem/input.dm
```diff
@@ -25,6 +25,7 @@ SUBSYSTEM_DEF(input)
 		"Any+UP" = "\"KeyUp \[\[*\]\]\"",
 		"Back" = "\".winset \\\"input.text=\\\"\\\"\\\"\"",
 		"F1" = "dummy", //overriding .options menu being triggered by assigning F1 to proc a verb that actually does nothing
```
boskoramen, line 27: could we set this to adminhelp again while achieving what you wanted to do originally?

See: https://discord.com/channels/150315577943130112/964684928161808384/1534622044598042764

> F1 being hardcoded was on purpose it's so you can always ask for help even if you have a keybinds problem...

> also let's it work before preferences loading

boskoramen (changes requested):

code/controllers/subsystem/input.dm
Drulikar: Just keep in mind it will likely no longer necessary to have it even present in Classic keys returning to how https://github.com/cmss13-devs/cmss13/pull/4749 had things.

(Of note I'm not reproducing any jank atm so the PR is currently working)

code/modules/client/client_defines.dm
```diff
@@ -49,7 +49,7 @@
 		//SECURITY//
 		////////////
 	var/next_allowed_topic_time = 10
-	control_freak = CONTROL_FREAK_ALL | CONTROL_FREAK_SKIN
+	control_freak = CONTROL_FREAK_MACROS
```
fira, line 52: Why are you changing control_freak ?

fira (changes requested):

code/modules/client/client_defines.dm
PoltavskaPraca: To reenable options and messages menu. There's a workaround now, achieved by using hardcoded dummy verb again so that we don't ever use F2 to screenshot even though it's a native byond keybind.

Drulikar: This is the main reason why this pr is reverting it back

code/controllers/subsystem/input.dm
Drulikar: This is still unresolved. But again if it introduces jank then we can just go forward with it as is.

PR #12884, opened 2026-08-08
Adds a sound preference to mute the round end music

Jone62:
# About the pull request

Adds a sound preference button to mute the round end music in the preferences tab.

For people who don't want to hear the end round music. More sound accessibility good.

<details>

</details>

# Changelog

:cl:
code: Adds a preference to toggle the end round sound theme
/:cl:


code/__DEFINES/sounds.dm
```diff
@@ -30,7 +30,7 @@
 //#define SOUND_CHANNEL_ADMIN_MIDI 1022
 #define SOUND_CHANNEL_LOBBY 1023
 #define SOUND_CHANNEL_Z 1024
-
+#define SOUND_CHANNEL_END_ROUND_MUSIC 1025
```
fira: You can't do that, it's capped at 1024

fira (changes requested):
Also this can't be TMed due to Prefs change

fira (changes requested):
Technically this will break lobby music for people still in lobby, but I don't think it's a big deal

However you also need to include a savfile migration so it enables by default also for existing users, otherwise they'll keep their current `toggles_sound` value which is without it

Look in `/datum/preferences/proc/savefile_update` - in fact one third of the examples there are sound 🤣 

Edit: You'll also need to bump SAVEFILE_VERSION_MAX to match

Jone62:
Using the lobby channel is technically consistent to how declare_completion() handled end round music prior to this PR. At worst it will cut off the lobby music

fira (approved):

Drulikar (approved):

PR #12886, opened 2026-08-08
Forbids reviving DNR'd corpses

Mary-Marilyn:
# About the pull request

Prevents defibrillators, SMART keys, and synth maintenance stations from reviving DNR'd people

DNR is an explicit choice by a player not to be revived and should therefore be respected.

This also fixes a cheesy exploit where a DNR'd corpse could be revived without its player, placed into cryosleep, and used to free the occupied role slot.

Tested locally and confirmed all three to work.

# Changelog

:cl:
fix: Defibrillators, SMART  keys, and synth maintenance stations now refuse to revive DNR'd people.
/:cl:

### DIFF
--- a/code/game/machinery/rechargestation.dm
+++ b/code/game/machinery/rechargestation.dm
@@ -322,6 +322,10 @@
 	if(revive_target.stat != DEAD)
 		return FALSE
 
+	if(!revive_target.client && !(revive_target.status_flags & FAKESOUL) && !revive_target.get_ghost(FALSE, TRUE))
+		visible_message(SPAN_WARNING("[icon2html(src, viewers(src))] \The [src] buzzes: DNR order detected. Reactivation aborted."))
+		return FALSE
+
 	if(!revive_target.is_revivable())
 		visible_message(SPAN_WARNING("[icon2html(src, viewers(src))] \The [src] buzzes: Synthetic's general condition does not allow reviving."))
 		return FALSE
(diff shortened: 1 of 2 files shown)


Zenith00000:
> This also fixes a cheesy exploit where a DNR'd corpse could be revived without its player, placed into cryosleep, and used to free the occupied role slot.

This is an exploit? I'm pretty sure this is intended, because someone can literally just join and go DNR, then you've lost a Spec role permanently.

This seems unnecessary.

georgie-28:
Why is cyroing DNR people to free up slots an exploit? DNR just means you don't want to play the game anymore it's not a "instant perma" button

kiVts:
There is a reason hardcore mode is only available on rifleman and no one else. Making DNR essentially hardcore that is avaliable for everyone not only defeats hardcore purpose entirely, it will also lead to admin problems and get specs in trouble because they decided they don't want to play anymore after their first death and medics cant recover the slot. It has happened before, but the slot was recovered.

kiVts:
To add onto the above:
Yes, hardcore triggers on death regardless while DNR is a player desired switch. By clicking DNR you essentially say "I won't play this round/character anymore" whether medics revive and cryo you (thus allowing a diffrent person enjoy the game with the slot you took) or leave to rot is already irrelevant to you, you're out of the round anyway. 

If the situation warrants you (by rules) to prevent yourself from being revived, then you can use the in-game mechanics like targetting your mouth and shooting the gun.

PR #12900, opened 2026-08-10
Add a Hear only encryption key variant and give the POs & IOs hear only command

wgtjunior743:
# About the pull request

Title

While discussing in the https://forum.cm-ss13.com/t/removing-comms-from-officers-actively-being-part-of-the-operation/21331/86 forum post, there were some ideas and this was one of those, i don't believe that reverting a PR in that case is the optimal decision so i think this change will help the POs and IOs who are left out (besides the IO SL) and of course it can be expanded into other stuff if people wish.

<details>
a XO talking:

XO_TALKING

a PO hearing and trying to talk on it:

DP_HEARING

How the headset looks with the hear only key, you can disable it if you want:

HEAR_ONLY_HEADSET

</details>

# Changelog
:cl:
add: an Encryption key variant that you can only hear a channel and not talk on it.
add: Give the IO and POs a command key variant which they can only hear it.
/:cl:


code/game/objects/items/devices/radio/encryptionkey.dm
```diff
@@ -38,6 +39,10 @@
 	icon_state = "cap_key"
 	channels = list(RADIO_CHANNEL_COMMAND = TRUE)
 
+/obj/item/device/encryptionkey/command/hear
+	name = "Command Encryption Key (Hear only)"
```
Detective-Google: I think this'd make more sense if the IC mentions were something akin to "Receive Only" instead; "hear only" sounds a bit weird grammatically (I think listen only would be proper?) but receive makes more sense in the context of a radio.

wgtjunior743: I will change to receive, my first language don't really has two words for hear/listening so i just went with the first i thought would fit

matttheficus:
As a small note, the issue of IOs not having Command Comms due to a bug with IO aSL is now fixed.

fira (dismissed):

Detective-Google (approved):
fira approved this so should be fine

PR #12911, opened 2026-08-12
fixes minimap crash when traversing z level inside of an object

cuberound:
# About the pull request

fixes the crash when traversing z level inside of bodybag or multiz pipe. I could not find any inconsistency caused by removing this part of code. checked that moving people in bags still updates them on minimap

fixes crash on pipes and bags

<details>

</details>

# Changelog
:cl:
fix: fixes runtime caused by crossing z levels using pipes and bodybags
/:cl:


fira (commented):
I think this should just not be called at all and checked for in on_z_change instead. it doesn't make sense that it calls oldloc = null when it's incorrect.

PR #12929, opened 2026-08-14
Reinforced Walls are now more expensive to make

Unknownity:
# About the pull request

Increases the plasteel costs to make a reinforced wall from 5 to 8.

nukes tons of single letter vars in the file (not all of them, I don't know what to name the W single letter var *shrug*)

Adds a marine tip regarding a very niche and underrated feature about reinforcing fuel and reagent tanks, lets make people more aware that they can do this!

WHY: As it currently stands out, the costs to make a reinforced wall is cheaper than it is to make a plasteel barricade, foldable or not. This makes reinforced walls more of an attractive option to choose since you are able to make more of them. If we compare them to regular walls, those ones are already balanced enough as they are more expensive to create than a wired metal barricade and take the same amount of metal as a metal foldable barricade.

I believe the design philosophy of reinforced walls should stay consistent with the rest of the buildable structures, they should be more expensive to create compared to their barricade counterparts so one should think carefully with how they want their defenses set up. There are pros and cons to each and currently a reinforced wall has way more pros than cons, especially compared to a regular wall, it being an upgrade to it and still cheaper.

By increasing the costs slightly, it will be on par of a foldable plasteel barricade in terms of costs and it will make the non foldable plasteel barricades more of an attractive option as they would be cheaper (currently I've only ever seen them built when the engineers have an abundant of plasteel)

Under the right hands, a reinforced wall is superior if the person knows what they're doing. I expect there will be less walls if this gets merged and it will eventually even benefit the marines so they stop building them in a way that griefs them lol.

Example: currently, from the default 40 sheets stacks, you can make 8 rwalls. With the nerf, they would only be able to make 5 rwalls. Good enough to seal one 3-4 wide path, but not that good to be able to seal multiple of them or make a rimworld style killbox.

This might affect survivors a little, heavily depending on the map and where they will hold of course.

The main impact this is aimed for is preventing a single person being able to create powerful defenses with just their default plasteel stack they get for free. Less so against multiple engineers cooperating or getting more plasteel from Requisitions or people spending their vendor points.

And also to match the consistency that a wall should always be more expensive to create than a barricade.

<details>

</details>

# Changelog

:cl:
balance: Reinforced Walls now take 8 plasteel sheets to construct instead of 5.
add: Added one marine tip about being able to reinforce reagent and fuel tanks with plasteel.
refactor: Nuked some single letter vars in the girders file.
/:cl:

### DIFF
--- a/code/game/objects/structures/girders.dm
+++ b/code/game/objects/structures/girders.dm
@@ -12,6 +12,7 @@
 #define STATE_RODS 5
 
 #define GIRDER_UPGRADE_MATERIAL_COST 5
+#define GIRDER_PLASTEEL_UPGRADE_MATERIAL_COST 8
 
 /obj/structure/girder
 	icon_state = "girder"
@@ -74,8 +75,8 @@
 		icon_state = "girder"
 
 /obj/structure/girder/attackby(obj/item/W, mob/user)
-	for(var/obj/effect/xenomorph/acid/A in src.loc)
-		if(A.acid_t == src)
+	for(var/obj/effect/xenomorph/acid/acid in src.loc)
+		if(acid.acid_t == src)
 			to_chat(user, "You can't get near that, it's melting!")
 			return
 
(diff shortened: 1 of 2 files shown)


Drulikar (approved):

fira:
Too late now but please don't change letter vars you're not touching in your code, it's extra noise in review for no reason

Unknownity:
> Too late now but please don't change letter vars you're not touching in your code, it's extra noise in review for no reason

damn...i thought we were supposed to nuke single letter vars when we see them...

Drulikar:
> > Too late now but please don't change letter vars you're not touching in your code, it's extra noise in review for no reason
> 
> damn...i thought we were supposed to nuke single letter vars when we see them...

Just should be for code you need to change not just whole files. But it's whatever.

PR #10485, opened 2025-08-24
VERY Limited Nest Resisting

realforest2001:
# ONLY SURVIVORS (BEFORE 15 MINUTES) AND UNINFECTED PEOPLE CAN ESCAPE

# About the pull request

This PR changes breaking out of nests. It is quite impactful to balance and therefore would require thorough testing and consideration. I have raised this discussion before and I think this PR brings the best balanced approach to it.

# Explain why it's good for the game

Permanent nesting is not fun. I understand the reason for it was to mean xenos don't have to camp the nests, but that's kind of what sentinels exist for. Further, this will make it so resisting out of nests (in a normal round) can only happen if the target is either A. Not infected or B. A survivor before first drop is mechanically possible.

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
balance: Reintroduced escaping from xeno nests.
balance: Escaping takes 30 seconds and is obvious when being attempted. It cannot be cancelled by the person attempting it, so no stopping and starting waiting for the best moment. It can be stopped by stuns.
balance: Survivors can escape prior to the first drop being mechanically possible (currently 15 minutes after round start).
balance: Anyone can escape if they're not infected.
/:cl:

PR #12943, opened 2026-08-16
Allows escaping nests when unhugged.

Mary-Marilyn:
# About the pull request

Allows nested who have not been facehugged to break free by resisting for one uninterrupted minute. Being stunned, incapacitated, moved, facehugged, or otherwise freed interrupts the attempt.

This also removes the unused remnants of the old nest breaking out system.

It prevents xenos from permanently securing an unhugged capture simply by nesting them, even if they abandon them and make no effort to actually guard or infect the victim.

Tested in localhosted. Worked as intended

<details>

No screenshots or videos; this change does not modify sprites or UI.

</details>

# Changelog

:cl:
balance: Unhugged nested can now break free by resisting for one uninterrupted minute.
/:cl:

### DIFF
--- a/code/game/objects/structures/stool_bed_chair_nest/xeno_nest.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/xeno_nest.dm
@@ -13,9 +13,6 @@
 	buckle_lying = 0
 	flags_obj = OBJ_ORGANIC
 	var/on_fire = 0
-	var/resisting = 0
-	var/resisting_ready = 0
-	var/nest_resist_time = 1200
 	var/mob/dead/observer/ghost_of_buckled_mob =  null
 	var/hivenumber = XENO_HIVE_NORMAL
 	var/force_nest = FALSE
@@ -45,8 +42,6 @@
 	. = ..()
 	if(. && current_mob.pulledby)
 		current_mob.pulledby.stop_pulling()
-		resisting = FALSE //just in case
-		resisting_ready = FALSE
 
 	if(buckled_mob == current_mob)
 		current_mob.pixel_y = buckling_y["[dir]"]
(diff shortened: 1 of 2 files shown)


Kilmannan:
Was done before, and was removed because it's a thoroughly unenjoyable and grindy mechanic for everyone. Forced xeno drones to just tackle and re-nest, forced marine players to keep resisting just to be re-nested. Canonically no-one's breaking out of a xeno nest.

Mary-Marilyn:
> Was done before, and was removed because it's a thoroughly unenjoyable and grindy mechanic for everyone. Forced xeno drones to just tackle and re-nest, forced marine players to keep resisting just to be re-nested. Canonically no-one's breaking out of a xeno nest.

Thanks for the feedback! Fortunately, the mechanic you're arguing against isn't actually in the PR.

The old system let hugged Marines repeatedly break out. This PR very deliberately does not do that. It only applies to **unhugged** victims.

ori-disillusion-mirnov:
i think they are talking about https://github.com/cmss13-devs/cmss13/pull/10485 and the problem in that one was sitting guarding survivors for 20 minutes isnt fun, which is not in this pr

zzzmike:
This will encourage xenos to just kill if there isn't easy eggs available. Xenos killing easy caps is already a problem as-is. I don't see this change as a benefit to xenos or to marines.

ihatethisengine:
Antiparasitic property is going to turn this into a headache for xenos. It really happened last round, when a guy got capped after injecting antipara, so he got unhugged.

Steelpoint:
The issue brought up by @ihatethisengine is a very valid reason to be wary of this balance change in of itself. I am also concerned that this attempt to change nesting balance was attempted a whiles ago and it also unearthed issues that necessitated it not being proceeded with. 

I would be interested in a discussion about marine nesting and changing how it works, such as encouraging Xenomorphs to nest Marines closer to the frontline instead of the hive, but that is a separate discussion. 

This change as it is is not something we will proceed with at this time.

PR #12948, opened 2026-08-16
USCM late join menu redesign

atebite:
# About the pull request

Removes the old HTML-string-building late join menu for the USCM roles and replaces it with TGUI. The menu is still standalone rather than a lobby menu modal because I think it's still useful to be able to open character setup while looking at the join menu.

UX-wise there's a few changes. First off, role categories are now explicitly ordered:

1. Marines (most players will likely want a role from this category)
2. Command (important for strategic decisions)
3. Aux roles (important for fire support, relatively popular)
4. Req/cargo (important for logistics)
5. Shipside medbay (important for non-urgent (and some urgent) surgeries)
6. MPs (LRP police)
7. Shipside engineering (honestly idk what these roles are for anymore)
8. Miscellaneous (CL/bodyguard/synths/mess/correspondent) (RP roles)
9. Other (jobs that are uncategorized in code)

For another thing, players are now only shown the *active* players for any given role, as well as how many slots are open. The total amount of players for the role, including inactive players, is no longer shown.

This is all very easy to change and I don't mind doing it if this is controversial.

Role icons are generated at runtime using the exact same code as the role HUD overlay rendering to make icon updates zero-maintenance. They will automatically reflect in the late join UI. It's very hacky, but I couldn't think of a better way that didn't involve hundreds of lines of refactoring since overlay rendering is tightly coupled with mob state while jobs are only weakly linked to factions.

Brings the UI up to modern standards and hopefully makes it slightly more intuitive to use for most people

also just looks neat 👍

The role icon sprite sheet initialization overhead is ~128ms over 20 calls on my system. This occurs only once per round, in the lobby during SSasset initialization. In other words keep in mind this perf experiment benefits slightly from warm execution where it will always run cold in real use. The overtime in this profile comes from the profiling setup:
```dm
/datum/asset/spritesheet/role_icons/New()
	GLOB.asset_datums[type] = src
	for(var/i = 1 to 20)
		sizes.Cut()
		sprites.Cut()
		register()
```
```
                                                        Profile results (total time)
Proc Name                                                                                     Self CPU    Total CPU    Real Time     Overtime        Calls
-----------------------------------------------------------------------------------------    ---------    ---------    ---------    ---------    ---------
/datum/asset/spritesheet/role_icons/register                                                     0.068        2.554        2.562        0.068           20

                                                       Profile results (average time)
Proc Name                                                                                     Self CPU    Total CPU    Real Time     Overtime        Calls
-----------------------------------------------------------------------------------------    ---------    ---------    ---------    ---------    ---------
/datum/asset/spritesheet/role_icons/register                                                     0.003        0.128        0.128        0.003           20
```

<details>

Standard CRT green theme:
bilde

CRT disabled (WeyYu theme):
bilde

UPP theme:
bilde

Hijack warning:
bilde

</details>

# Changelog

:cl: Atebite
ui: Redesigned the USCM late join menu
/:cl:

Not user facing:
- fix CRT effect clipping on scrollable TGUIs with the CRT themes
- fix a few runtimes that would break role icon rendering


atebite:
my bad on lints, fixed

code/datums/late_join.dm
```diff
@@ -0,0 +1,90 @@
...
+		else if(GLOB.ROLES_MARINES.Find(J.title))
+			role_category = "Marines"
+		else
+			role_category = "Other"
+
+		if(!(role_category in categorized_roles))
+			categorized_roles[role_category] = list()
+
+		// Why is DM so insistent that you don't make multidimensional arrays?
+		categorized_roles[role_category][++categorized_roles[role_category].len] = list(
```
Drulikar, line 49: use https://ref.dm-lang.org/proc/length instead

code/datums/late_join.dm
```diff
@@ -0,0 +1,90 @@
...
+		ui.close()
+
+/datum/late_join/ui_state(mob/user, datum/ui_state/state)
+	if(isnewplayer(user))
+		return GLOB.new_player_state
+	return GLOB.default_state
+
+/datum/late_join/tgui_interact(mob/user, datum/tgui/ui)
+	. = ..()
+
```
Drulikar:
```suggestion
```
Parent implementation is just a stub

code/datums/late_join.dm
```diff
@@ -0,0 +1,90 @@
...
+		if(!GLOB.RoleAuthority.check_role_entry(user, J, latejoin = TRUE, faction = FACTION_NEUTRAL))
+			continue
+		var/active = 0
+		// Only players with the job assigned and AFK for less than 10 minutes count as active
+		for(var/mob/M in GLOB.player_list)
+			if(M.client && M.job == J.title)
+				active++
+
+		var/role_category
+		if(GLOB.ROLES_CIC.Find(J.title))
```
Drulikar: See https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#dont-use-abbreviations
and https://github.com/cmss13-devs/cmss13/blob/master/.github/guides/STYLES.md#exceptions-to-variable-names

code/datums/late_join.dm
```diff
@@ -0,0 +1,90 @@
...
+	user = new_user
+
+/datum/late_join/ui_data()
+	. = ..()
+	var/list/data = list()
+
+	if(SShijack)
+		switch(SShijack.evac_status)
+			if(EVACUATION_STATUS_INITIATED)
+				LAZYADD(data["HijackInitiated"], TRUE)
```
Drulikar: Unless you foresee where more kinds of statuses matter, this is really verbose for just
`if(SShijack?.evac_status == EVACUATION_STATUS_INITIATED)`

code/datums/late_join.dm
```diff
@@ -0,0 +1,90 @@
...
+	// Build a list of named categories of roles, each containing a list of information on individual roles
+	var/list/categorized_roles = list()
+	for(var/i in GLOB.RoleAuthority.roles_for_mode)
+		var/datum/job/J = GLOB.RoleAuthority.roles_for_mode[i]
+		if(!GLOB.RoleAuthority.check_role_entry(user, J, latejoin = TRUE, faction = FACTION_NEUTRAL))
+			continue
+		var/active = 0
+		// Only players with the job assigned and AFK for less than 10 minutes count as active
+		for(var/mob/M in GLOB.player_list)
+			if(M.client && M.job == J.title)
```
Drulikar: player_list should already be only cliented mobs but its not a huge deal

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -62,7 +62,8 @@
 	stamina = new /datum/stamina(src)
 
 /mob/living/carbon/human/Destroy()
-	SSround_recording.recorder.stop_tracking(src)
+	if(!isnull(SSround_recording.recorder))
+		SSround_recording.recorder.stop_tracking(src)
```
Drulikar:
```suggestion
	SSround_recording?.recorder.stop_tracking(src)
```

code/datums/late_join.dm
```diff
@@ -0,0 +1,90 @@
+GLOBAL_DATUM_INIT(late_join_tgui, /datum/late_join, new)
+
+/datum/late_join
+	var/datum/tgui/ui = null
+	var/mob/new_player/user = null
```
Drulikar: I see no need for you to hold a reference to `user`

code/datums/late_join.dm
```diff
@@ -0,0 +1,90 @@
...
+
+/datum/late_join/New(mob/new_player/new_user)
+	. = ..()
+	user = new_user
+
+/datum/late_join/ui_data()
```
Drulikar: This proc normally has an argument `mob/user`

code/datums/late_join.dm
```diff
@@ -0,0 +1,90 @@
...
+
+	return data
+
+/datum/late_join/ui_assets(mob/user)
+	. = ..()
+	. += get_asset_datum(/datum/asset/spritesheet/role_icons)
+
+/datum/late_join/ui_act(action, list/params, datum/tgui/ui, datum/ui_state/state)
+	. = ..()
+	var/success = user.AttemptLateSpawn(action)
```
Drulikar:
```suggestion
	var/success = ui.user.AttemptLateSpawn(action)
```

code/modules/asset_cache/asset_list_items.dm
```diff
@@ -544,3 +544,46 @@
...
+		// set the fake human's faction by loading the job's ID onto it
+		var/datum/equipment_preset/job_preset = GLOB.equipment_presets.gear_path_presets_list[job.gear_preset]
+		if(!isnull(job_preset))
+			job_preset.load_id(fake_human)
+
+			// Using the faction from the job's gear preset, render the fake human's HUD into our holder
+			var/image/holder = new()
+			var/datum/faction/faction = GLOB.faction_datums[fake_human.faction]
+			if(!isnull(faction))
+				faction.modify_hud_holder(holder, fake_human)
```
Drulikar, line 575: This all very convoluted for what should just be making `modify_hud_holder` called with rank argument instead of a mob so its more flexible.

code/__DEFINES/job.dm
```diff
@@ -1,6 +1,6 @@
 #define get_job_playtime(client, job) (client.player_data? LAZYACCESS(client.player_data.playtimes, job)? client.player_data.playtimes[job].total_minutes MINUTES_TO_DECISECOND : 0 : 0)
 #define GET_MAPPED_ROLE(title) (GLOB.RoleAuthority?.role_mappings[title] ? GLOB.RoleAuthority.role_mappings[title] : GLOB.RoleAuthority.roles_by_name[title])
-#define GET_DEFAULT_ROLE(title) (GLOB.RoleAuthority?.default_roles[title] ? GLOB.RoleAuthority.default_roles[title] : title)
+#define GET_DEFAULT_ROLE(title) (GLOB.RoleAuthority?.default_roles?[title] ? GLOB.RoleAuthority.default_roles[title] : title)
```
Drulikar, line 3: Any particular reason why you're doing this? Wouldn't it be better for it to error than to fail silently?

Drulikar (changes requested):

Drulikar:
Also ideally the width of the UI should be long enough that you can keep the layout of the interface consistent (rather than things getting shoved to the right sometimes)

code/datums/late_join.dm
atebite: a lot of this is copypaste from existing latejoin code over to the tgui data pipe, but can fix

code/datums/late_join.dm
atebite: same as above

code/datums/late_join.dm
atebite: same as above

code/modules/asset_cache/asset_list_items.dm
atebite: I agree, I know it's a nasty hack and thought this would not be a popular approach. I stuck with this intentionally to avoid refactoring `modify_hud_holder` since it's a bigger change than it seems at face. A lot of the rendering is strongly tied to mob vars and you still have the issue of jobs being very weakly linked to those procs.

I'm happy to split this out to a debug function that renders sprite sheets and commit those directly instead, otherwise I'd rather rebase on top of a separate refactor, which is out of scope for this PR.

code/__DEFINES/job.dm
atebite: I had a runtime which broke out of the icon rendering callstack and break the whole icon spritesheet.

This was a fix that seemed in spirit of the actual purpose of the define, which (to my understanding) is to use `GLOB.RoleAuthority.default_roles[title]` if it exists, and `title` otherwise. `default_roles` seems to be for mapping "standard" role names to gamemode-specific roles (e.g. WO), so it makes sense that not all roles have a mapping.

code/datums/late_join.dm
atebite: yup good catch

code/datums/late_join.dm
atebite: `list[++list.len]` extends the list ahead of the access, `list[++length(list)]` is just an OOB access

atebite: can see the linter isn't happy about it though, will see what I can do

atebite:
> Also ideally the width of the UI should be long enough that you can keep the layout of the interface consistent (rather than things getting shoved to the right sometimes)

Increased the window size a little bit and fixed a bunch of widths to get a more consistent layout:

<details>
bilde

with large amounts of players:

bilde

bilde
</details>

code/datums/late_join.dm
atebite: https://github.com/cmss13-devs/cmss13/pull/12948/changes/c385d8c39ba8f477a0b7805ee27ba8ba53d20291#diff-4770dc67ec9c2154cf9af62c8ea3376a3a7dfdd3da64f8f74640b050a7087585R13 makes the linter happy without breaking logic.

changed the comment to be more descriptive

code/datums/late_join.dm
```diff
@@ -0,0 +1,81 @@
...
+
+	return data
+
+/datum/late_join/ui_assets(mob/user)
+	. = ..()
+	. += get_asset_datum(/datum/asset/spritesheet/role_icons)
+
+/datum/late_join/ui_act(action, list/params, datum/tgui/ui, datum/ui_state/state)
+	. = ..()
+	var/success = ui.user?:AttemptLateSpawn(action)
```
Drulikar:
```suggestion
	var/mob/user = ui.user
	var/success = user?.AttemptLateSpawn(action)
```
We don't want `:` usage as an accessor

tgui/packages/tgui/interfaces/LateJoin.tsx
Drulikar, line 1: The [linter failure](https://github.com/cmss13-devs/cmss13/actions/runs/33974646089/job/101630780079?pr=12948#step:9:623) is that prettier would want to format this file differently. Normally just using VSC and saving this file would apply it. But you maybe disabled the extension, edited it without vsc, or maybe this branch is too old to not include https://github.com/cmss13-devs/cmss13/pull/13066 (so in that case you can merge upstream/master in and then try saving this file, or use bin/tgui-fix.cmd

tgui/packages/tgui/interfaces/LateJoin.tsx
```diff
@@ -0,0 +1,201 @@
...
+                    <Stack align="center">
+                      <Stack.Item>
+                        <RoleIcon role={role.Title} />
+                      </Stack.Item>
+                      <Stack.Item>
+                        <h3 className="big-text">{role.DisplayTitle}</h3>
+                      </Stack.Item>
+                    </Stack>
+                  </Flex.Item>
+                  <Flex.Item basis="40%" className="RoleSlotInfo">
```
Drulikar: You somereason removed this class from the scss so its not doing anything

tgui/packages/tgui/interfaces/LateJoin.tsx
```diff
@@ -186,14 +176,19 @@ const RoleCategory = (props: RoleCategoryProps) => {
...
-                  <Flex.Item grow={1}>
+                  <Flex.Item basis="40%" className="RoleSlotInfo">
                     <RoleSlotInfo role={role} />
                   </Flex.Item>
-                  <Flex.Item>
-                    <Box className="medium-text">
-                      Join
-                      <Icon name="caret-right" />
-                    </Box>
+                  <Flex.Item basis="15%" className="RoleJoin">
```
Drulikar: This class isn't defined anywhere

tgui/packages/tgui/styles/themes/crt.scss
```diff
@@ -136,11 +136,12 @@ $scrollbar-color-multiplier: 0.5 !default;
   .Layout__content::before {
     content: ' ';
     display: block;
-    position: absolute;
+    position: fixed;
     top: 0;
     left: 0;
     bottom: 0;
     right: 0;
+    // background: red;
```
Drulikar:
```suggestion
```

tgui/packages/tgui/styles/interfaces/LateJoin.scss
```diff
@@ -0,0 +1,38 @@
...
+}
+
+.HijackIndicator h1 {
+  font-size: 2rem;
+}
+
+.HijackIcon {
+  margin-left: 4rem;
+  margin-right: 4rem;
+}
```
Drulikar: This isn't used anywhere

tgui/packages/tgui/interfaces/LateJoin.tsx
```diff
@@ -0,0 +1,201 @@
...
+  }, []);
+
+  const theme = themeDisabled
+    ? 'weyland_yutani'
+    : data.UPPEnabled
+      ? 'crtred'
+      : 'crtgreen';
+  return (
+    <Window theme={theme} width={650} height={750}>
+      <Window.Content className="LateJoin" scrollable>
```
Drulikar: This class is defined anywhere

tgui/packages/tgui/interfaces/LateJoin.tsx
```diff
@@ -0,0 +1,201 @@
...
+          <Flex.Item basis="40%" className="SlotInfoRow big-text">x{role.Active}</Flex.Item>
+        </Flex>
+      </Flex.Item>
+      <Flex.Item basis="30%">
+        <Flex direction="row" align="center">
+          <Flex.Item basis="60%" className="SlotInfoRow big-text">
+            <Icon name="door-open" />
+          </Flex.Item>
+          <Flex.Item basis="40%" className="SlotInfoRow big-text">
+            {infSlots ? <Icon name="infinity" /> : roleSlots}
```
Drulikar, line 154: SlotInfoRow is a class not defined anywhere

tgui/packages/tgui/interfaces/LateJoin.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const { act } = useBackend<LateJoinData>();
+
+  return (
+    <Section title={category} className="RoleCategory">
+      <Stack vertical>
+        {roles.map((role) => {
+          return (
+            <Stack.Item key={role.Title}>
+              <Button
+                className="JoinButton"
```
Drulikar: JoinButton is a class not defined anywhere

tgui/packages/tgui/interfaces/LateJoin.tsx
```diff
@@ -0,0 +1,201 @@
...
+                fluid
+                onClick={() => act(role.Title)}
+              >
+                <Flex direction="row" align="center">
+                  <Flex.Item basis="55%" className="RoleTitle">
```
Drulikar: RoleTitle is a class not defined anywhere

tgui/packages/tgui/interfaces/LateJoin.tsx
```diff
@@ -0,0 +1,201 @@
...
+                      </Stack.Item>
+                      <Stack.Item>
+                        <h3 className="big-text">{role.DisplayTitle}</h3>
+                      </Stack.Item>
+                    </Stack>
+                  </Flex.Item>
+                  <Flex.Item basis="40%" className="RoleSlotInfo">
+                    <RoleSlotInfo role={role} />
+                  </Flex.Item>
+                  <Flex.Item basis="15%" className="RoleJoin">
```
Drulikar: RoleJoin is a class not defined anywhere

tgui/packages/tgui/interfaces/LateJoin.tsx
```diff
@@ -0,0 +1,201 @@
...
+  readonly DisplayTitle: string;
+  readonly Slots: number;
+  readonly Players: number;
+  readonly Active: number;
+};
+
+type LateJoinData = {
+  HijackInitiated?: boolean;
+  Categories: Object;
+  UPPEnabled: boolean;
```
Drulikar: For these `boolean`s you should use `BooleanLike` instead since byond is going to hand it a `1`/`0` not `true`/`false`. You'll just then need to do say `(!!data.HijackInitiated)` instead of `(data.HijackInitiated === true)` to satisfy typescripts issue with BooleanLike being potentially undefined

tgui/packages/tgui/interfaces/LateJoin.tsx
```diff
@@ -0,0 +1,201 @@
...
+            <Stack
+              className="HijackIndicator"
+              align="center"
+              justify="space-around"
+            >
+              <Stack.Item>
+                <Icon name="warning" size={2} />
+              </Stack.Item>
+              <Stack.Item>
+                <h1>HIJACK IN PROGRESS</h1>
```
Drulikar: You're checking evacuation status not hijack status so this isn't correct.
```suggestion
                <h1>EVACUATION IN PROGRESS</h1>
```

code/modules/asset_cache/asset_list_items.dm
Drulikar: The entire reason why it has to change in the first place is to accommodate this PR, so its not out of scope at all. It should also make things a lot more flexible for you.

[patch.diff.txt](https://github.com/user-attachments/files/31906432/patch.diff.txt)

You can use `git apply path/to/the/patch.diff.txt` or whatever fancies you.

Image

Image

Drulikar (changes requested):
### Important

If you're going to use the patch provided below regarding the refactor, you're going to need to use it first before applying suggestions or making other changes.

tgui/packages/tgui/styles/themes/crt.scss
```diff
@@ -136,11 +136,12 @@ $scrollbar-color-multiplier: 0.5 !default;
   .Layout__content::before {
     content: ' ';
     display: block;
-    position: absolute;
+    position: fixed;
```
Drulikar, line 139: I'm very skeptical this should be done across the board for this style. But I need to see how this change affects all other interfaces that use it. But it would probably be better to just style this unique to your interface rather than change it globally.

tgui/packages/tgui/interfaces/LateJoin.tsx
atebite: yeah I just don't have the extension and I'm terminally bad at running linters before committing

tgui/packages/tgui/interfaces/LateJoin.tsx
atebite: not needed anymore

tgui/packages/tgui/styles/themes/crt.scss
atebite: good catch

tgui/packages/tgui/styles/themes/crt.scss
atebite: This styling is for the scanline effect. When `position: absolute`, the effect cuts off at the height/width border (depending on scroll direction) for scrollable content because the positioning of the effect stays absolute relative to some block in the document rather than the viewport itself. This change [doesn't affect document layout](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/position#values), it just ensures the positioning is fixed to the viewport itself rather than a document object.

`position: absolute`:
bilde

`position: fixed`:
bilde

What specifically are you worried would break with this?

Drulikar: Its just a lot of interfaces for me to go through and check if they now behave wrongly.

Nothing stands out.

code/datums/late_join.dm
```diff
@@ -0,0 +1,84 @@
...
+	. += get_asset_datum(/datum/asset/spritesheet/role_icons)
+
+/datum/late_join/ui_act(action, list/params, datum/tgui/ui, datum/ui_state/state)
+	. = ..()
+	var/mob/new_player/user = ui.user
+	if(!istype(user))
+		return
+	var/success = user.AttemptLateSpawn(action)
+	if(success)
+		ui.close()
```
Drulikar:
```suggestion
	if(user.AttemptLateSpawn(action))
		ui.close()
```
I wouldn't worry about typechecking the mob; I don't think anywhere does in this proc. If this even is possible its whatever if it errors (I'd rather see it error because I'd rather a ui_act not happening at all if its somehow possible to perform w/o a user).

tgui/packages/tgui/interfaces/LateJoin.tsx
```diff
@@ -0,0 +1,198 @@
...
+];
+
+export const LateJoin = (props, context) => {
+  const { act, data } = useBackend<LateJoinData>();
+
+  const [themeDisabled, setThemeDisabled] = useState<boolean | undefined>();
+
+  useEffect(() => {
+    storage.get('lobby-theme-disabled').then((val) => setThemeDisabled(!!val));
+  }, []);
```
Drulikar, line 53: This being done as an effect basically makes it so opening the panel will flashbang you with a green interface every time. But I'm not sure at the moment how the same doesn't occur to the lobby menu even with `Fix-Interfaces` or `Refresh-TGUI` reloading it (UPP mode doesn't do this since its from ui_data)

Image

Not something to block merging but...

tgui/packages/tgui/interfaces/LateJoin.tsx
```diff
@@ -0,0 +1,198 @@
...
+};
+
+type RoleSlotInfoProps = ComponentProps<typeof Box> & {
+  readonly role: RoleInformation;
+};
+const RoleSlotInfo = (props: RoleSlotInfoProps) => {
+  const { role } = props;
+
+  const infSlots = role.Slots === -1;
+  const roleSlots = 'x' + role.Slots;
```
Drulikar:
```suggestion
  const roleSlots = 'x' + (role.Slots - role.Players);
```
Else you're not indicating how many slots are available (and you weren't using the Players data anyways). Num `Active` will decrease as people ghost (but they won't necessarily refund a slot unless they cryod)

Drulikar (changes requested):
Otherwise there isn't anything else standing out to me that needs fixing.

I did ping for someone else to review my code I provided in patch since its not ideal I approve my own code.

Only other note is that another maint brought up is that they'd like colors by department - but then it'd be hard to follow the menu's appearance and would have to just be a normal theme.

tgui/packages/tgui/interfaces/LateJoin.tsx
Drulikar: (Harry suggests await the ui loading (return the rotating toolbox thing) on reading the pref)

tgui/packages/tgui/styles/themes/crt.scss
atebite: Would maybe be easier to testmerge and see if anything sticks out over the course of a round.

atebite:
> Only other note is that another maint brought up is that they'd like colors by department - but then it'd be hard to follow the menu's appearance and would have to just be a normal theme.

Yeah this is a difficult ask unless we define per-theme department color schemes. I thought about doing department icons which would've been a theme-neutral alternative but I don't have a shred of artistic talent in me.

tgui/packages/tgui/interfaces/LateJoin.tsx
atebite: I suppose it's still a bit of a flashbang considering how fast the promise resolves but it's more neutral now and can be changed down the line if it's annoying

code/datums/late_join.dm
```diff
@@ -0,0 +1,82 @@
...
+	. = ..()
+	var/mob/new_player/user = ui.user
+	var/success = user.AttemptLateSpawn(action)
+	if(success)
+		ui.close()
+
+/datum/late_join/ui_state(mob/user, datum/ui_state/state)
+	if(isnewplayer(user))
+		return GLOB.new_player_state
+	return GLOB.default_state
```
Drulikar: Currently its possible for the late join menu to stay open if you say observe or something (and probably not just observe - probably UPP join, xeno observe, tutorial too)

Attempting to click anything just runtimes atleast. You probably should do `return GLOB.never_state` if they're not a new player. (I would say just return only `GLOB.new_player_state` w/o checking but it for whatever reason allows admins to override that...)

Image

Image

Drulikar (changes requested):
The loading screen seems fine. Still waiting on someone else to ok my code but I'll eventually just okay it regardless. Sorry to have another RC because I thought of something else to try.

Drulikar (approved):
Doesn't seem like its regularly calling ui_state again, but the only reason why it stayed open before was because of the admin exception so meh - fine enough. If someone cared it could on observe/tutorial/whatever just explicitly close that UI since its a bit slow about closing.

Ideally someone else also reviews [`ebfa151` (this PR)](https://github.com/cmss13-devs/cmss13/pull/12948/commits/ebfa151902207cc593886b3ef9806be98aaf166d) but I'll eventually merge as is.

Nihisohel:
https://github.com/user-attachments/assets/26d9e6e4-70ba-40eb-9519-d847241b838d

not particularly an issue, but just noting an inconsistency

Nihisohel (approved):
looks good, but we really ought to move the chat color elsewhere

PR #12966, opened 2026-08-19
changes rolling pin and spoon to not contraband

aquariofury:
# why is a spoon contraband

I think mess techs should be allowed to use spoons. and rolling pins to prevent interlopers in the kitchen.

<details>

image

</details>

# Changelog

:cl:
code: moved rolling pin and spoon out of the contraband list
/:cl:


kiVts (approved):
Products list really ought to be formated like the prison one below, but you're just adding two new products to it I won't make you change them

PR #12977, opened 2026-08-20
Bugfix Rebalance: Crusher melee attacks slow for 3.5 seconds (was 2 seconds)

antlersss:
# About the pull request

Un-nerfs the crusher's melee slow from 2 seconds to 3.5 seconds, after bugfix #12241 nerfed them.

Crusher is currently the weakest T3 caste in the game. We should not have nerfed it in the first place.

Additionally, bugfixes should not seek to actively rebalance the game. If returning code to intended behavior has the side effect of numbers or stats being changed, the fix should adjust their values accordingly to return balance to how it was beforehand.

<details>

</details>

# Changelog
:cl: Antlers
balance: Un-nerfs the crusher's melee attack slow after its bugfix, returning it to 3.5 seconds
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/Crusher.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/Crusher.dm
@@ -237,7 +237,7 @@
 	if (!isxeno_human(target))
 		return
 
-	new /datum/effects/xeno_slow(target, bound_xeno, ttl = 2 SECONDS)
+	new /datum/effects/xeno_slow(target, bound_xeno, ttl = 3.5 SECONDS)
 
 	var/damage = bound_xeno.melee_damage_upper * aoe_slash_damage_reduction
 


Red-byte3D:
it being a bugfix does not change the ffact that it has the balance approval tag, so that is a moot point to make here.
i'd also jsut suggest looking at other ways as its really not a fun experience to get slowed for a whole 4 seconds.

antlersss:
The "why is this PR good for the game" section states no balance reasoning as to why the value was altered, purely stating it's a bugfix. 

The inclusion of a balance approved tag does not change the fact that there was no reason provided as to *why* the changes were made, from a balance perspective. Seeing as crusher is already one of the weakest t3 castes in the current game and nothing was given to supplement the nerf, the change should not have been made. 

Passing it under the guise of a bugfix is bad practice. I encourage anyone in the future to atomize their PR's into a fix section and a balance section in the future.

Steelpoint (approved):
I am inclined to accept this. The previous PR was labelled as a bug fix and was likely merged due to that, this PR will return the Crusher to its original intended 3.5 second slowdown following the bug fix. 

I would be happy to consider a separate PR to directly nerf the Crusher slowdown, however such a PR will be required to justify why a nerf is warranted.

PR #12984, opened 2026-08-21
Implements Abstract Datums

fira:
# About the pull request

Implements Abstract Datums, those that we mark as only existing for inheritance or static checking. 

The point of this is mainly twofold: 
 * Allow code using subtypesof for example to instantiate things to skip abstract ones
 * Allow unit tests to skip checking things that won't normally exist in the game anyway and might lack critical props

The first is exposed through a new proc `subtypesof_real` which filters abstract types.
The second in particular is used for the icon checking unit test. 

I also killed a few non-warranted empty subtypes in ammo.
**This also re-elects ITEM_ABSTRACT items for icon unit tests, but might require reverting if it turns out to allow too many false positives.**

# Testing

Did basic testing only, but the failure surface is pretty low. Short of a blatant mistake with the first case, there shouldn't be any issues.

### DIFF
--- a/code/__HELPERS/_types.dm
+++ b/code/__HELPERS/_types.dm
@@ -1,0 +1,7 @@
+/// Like [subtypesof] but returns only datum types that haven't been marked as abstract
+/proc/subtypesof_real(input_path)
+	var/list/paths = subtypesof(input_path)
+	. = list()
+	for(var/datum/path as anything in paths)
+		if(path.abstract_type != path)
+			. += path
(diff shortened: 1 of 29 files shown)


code/__HELPERS/_types.dm
```diff
@@ -0,0 +1,7 @@
+/// Like [subtypesof] but returns only datum types that haven't been marked as abstract
+/proc/subtypesof_real(input_path)
+	var/list/paths = subtypesof(input_path)
+	. = list()
+	for(var/datum/path as anything in paths)
+		if(path.abstract_type != path)
+			. += path
```
Watermelon914, line 7: it would probably be cheaper for a lot of code to just do this check inline, but for general code it's probably fine

fira: Yeah I figured if this is going to be a recurring check it's better to inline it and cache the results, but we only use all of this for game setup / init once right now

fira:
Drafted due to the ITEM_ABSTRACT thing that needs to be removed

PR #12998, opened 2026-08-22
Sentinel Oxy Damage Nerf

lilacrhododendron:
# About the pull request

Yes this is very similar to a prior PR, it was suggested in the Discord that the prior PR was not something that the maintainers were wholly disinterested in and the original author closed it simply because she realized it was intended to have the current values. 

Sentinel post rework is highly oppressive and many people are complaining about it regularly, 50 oxyloss is too much damage for the lowest risk offensive caste. I have seen many people say that they believe Sentinel is stronger than Spitter now, and I'm inclined to agree based on my experience. This needs adjusting. Not particularly attached to the value of 20 if a maintainer wants it somewhere else. 

# Changelog

:cl: Orchidfox
balance: Sentinel max neuro oxyloss reduced from 50 to 20
/:cl:

### DIFF
--- a/code/datums/effects/xeno_strains/sentinel_neuro_stacks.dm
+++ b/code/datums/effects/xeno_strains/sentinel_neuro_stacks.dm
@@ -16,7 +16,7 @@
 	///how much oxy damage should be given per process
 	var/proc_damage_per_stack = 0.2
 	///stopgrap for oxy damage on mob when this stops causing harm
-	var/max_oxyloss = 50
+	var/max_oxyloss = 20
 	///particles
 	var/obj/effect/abstract/particle_holder/particle_holder
 


Detective-Google:
considering
- prior sent did 0 damage whatsoever 
- new sent does the slowdown of old sent with better range and IFF
- new sent has dramatically higher fire rate than old sent in addition to above

seems fine

fira (approved):
Code OK

cuberound:
I would rather suggest to make the cap progressive based on the number of stacks you have with max being 30 aka 1 per stack, would be more interesting

Detective-Google:
this has tested very well

Detective-Google:
I tried to merge the wrong one oops

image
so true github

PR #13001, opened 2026-08-22
Update Docker config

fira:
# About the pull request

Updates the Docker tooling that has been left untouched for almost half a decade to work with our modern game. 

git clone 
docker compose up 
connect to localhost:51400

On any x86 platform (or whatever you want to emulate it with). You'll also want to do this if you're going to run serverside on Linux because it includes rustg building and the repo doesnt. It's also useful if you want to do ndatabase development to test the SQL driver.

Additionally, docker compose allows to spin additional services. Right now it's just mysql, but we can easily spin and configure redis, or influxdb aswell to mirror production environment. Unfortunately, grafana and opensearch are a bit more involved and hard to squeeze in a dev environment.

# Changelog (kinda)
 * Bumped base OS image from Ubuntu 20.04 to Debian 13 -- BYOND simply does not support older versions of glibc anymore
 * Bumped BYOND version (NodeJS will have to go because Bun)
 * This now builds rustg from repository automatically
 * Fixed mysql startup sequence due to a crutch apparently having changed flags
 * Use --chown for multistage copy. Directly doing chown causes heavy CoW operations which can take up to 5 minutes to complete...

capture

# Caveats
 * Sometimes unpacking util-linux will take several minutes, I have no idea why
 * First startup can sometimes fail because the game boots up before the mysql database had time to run its horrendous init scripts


fira:
This fix deployment mode but not standalone running mode that still has issues, actually

Drulikar (changes requested):
Linux no likey windows assets it seems. I still think this is worth having around working - even if its just something in git history someone can revert to since you already got it this far.

Just as a reminder, the bind for map_config and maps likely unnecessary so it can be default in standalone mode.

Drulikar:
https://github.com/cmss13-devs/cmss13/pull/13066 is relevant to this as well.

PR #13011, opened 2026-08-23
Valkyrie and Oppressor hook throw and a misleading line fix

JHINgle4ce:
# About the pull request

Added a new arg to throw proc. Now throw proc calculates a closest linear way from A to B, my change will allow you to give it your own movement line so you can throw objects the way you prefer.

As I was fixing Valkyrie and Oppressor hooks, I also refactored some code in order for it to do less calculations but its a minor change.

Now Valkirie and Oppressor will pull targets via the line they draw.

The problem is that hook line that we see is not the same as the one host or xeno will fly by. Both lines are constructed via use of get_dir and therefore are different cause of different perspectives. To say nerdly this throw proc is not commutative, or simply a path from A to B is not the same as from B to A, which is a problem in some cases.

It is a problem not that big, yet irritating and unsatisfying. You'll clearly see it in a vid below

<details>
The problems:

https://github.com/user-attachments/assets/5072ee37-6939-49b9-a915-af5043bf6660

https://github.com/user-attachments/assets/497fed09-779d-42df-8445-9a4cca4b788b

After my changes:

https://github.com/user-attachments/assets/53538534-ce01-442e-a5f4-ab604cc4e9ae

https://github.com/user-attachments/assets/da63a208-3140-4c80-bb49-12ec42fe9cc1

</details>

# Changelog

:cl: JHINgle4ce
code: added a new arg in throw_atom proc; it allows you to do a custom throw trajectory
refactor: little optimization in Oppressor and Valkyrie hook code. A little less calculations and checks
/:cl:


cm13-github:

cm13-github:

PR #13013, opened 2026-08-23
Makes Lunge, Pounces and its subtypes more consistent during movement.

Venuska1117:
# Spotted issues or want to give feedback?
click this link: https://discord.com/channels/150315577943130112/1464087612796174501

# About the pull request

Fixes warrior lunge and pounce ability inconsistency when casting abilities while moving.

This was also requested on my Feedback channel:
image

Missing a lunge or pounce when you are running around, making code think you missed lunge only because you were holding directional key during throw_atom(), making it think that lunge landed quicker than it should (or it just simply fails because of you jumping from tile B instead of tile A), also affects people with high ping making them miss more often.

This PR makes so player is immobilized before throwing occurs, fixing issue of player moving during execution of throw.

# Testing Videos

<details>
<summary>Lunges during movement</summary>

https://github.com/user-attachments/assets/e6572221-2e32-4c10-b895-f2d1e8121bbd

</details>

<details>
<summary>Pounces during movement</summary>

Runner example:

https://github.com/user-attachments/assets/be3f0ac0-a7b3-4aa3-a214-3c9c87bede79

</details>

# Possible Side-effects (not guaranteed)
<details>
<summary>Click HERE to show</summary>
- Crusher can no longer cancel its charge by holding directional key during windup.
- If server slows down to crawl for any reason (like once there were 10000 sounds in queue), using this abilities could get players stuck in "immobilized" state (experienced that when i resisted during lag and got stuck in that state for 10 minutes) until server catch up, meaning you can get yourself killed without a way to move and possible future issues similar to #9303.
</details>

# Changelog

:cl: Venuska1117
balance: Warrior Lunge can now be used more consistently during movement, preventing player from moving during lunge so it is no longer registered as "failed" lunge.
balance: Pounce ability and its subtypes used by Runners, Lurkers, Ravagers, Praetorians, Vanguards, Despoiler and Vampires can be used more consistently during movement.
/:cl:


code/modules/mob/living/carbon/xenomorph/castes/Warrior.dm
```diff
@@ -226,7 +226,9 @@
 
 	lunge_user.visible_message(SPAN_XENOWARNING("[lunge_user] lunges towards [carbon]!"), SPAN_XENOWARNING("We lunge at [carbon]!"))
 
+	ADD_TRAIT(lunge_user, TRAIT_IMMOBILIZED, TRAIT_SOURCE_ABILITY("lunge"))
```
Red-byte3D, line 229: I think this should be a preference in my opinion. i'd like more control over my lunge and when i want to cancel it instead of being forced to take bad lunges.

Venuska1117: Did you even bother checking attached videos to PR?

Also, bad lunges are not even possible in first place. (they are sprite clicks)

boskoramen:
Think we should make this a pref instead of removing this entirely, and if folks opt to toggle movement cancelling off they should be able to cancel lunge/pounce using a verb like "Resist".

Chatted with the team and think bare minimum is to set things up so "Resist" cancels lunge/pounce. Making movement cancel lunge/pounce as a pref is a nice-to-have, but can be picked up by someone else if they feel super strongly about it.

neeshacark:
hero

fira:
Talked with people about this recently during the throw refactor PR test, and apparently it was wholly intended to let you cancel it as needed. Can't say I agree, but there's that.

Venuska1117:
I will wait for TM feedback to flood in first before doing preferences.

fira:
Yeah if you do prefs it can't be TMed so...

ihatethisengine:
> - Crusher can no longer cancel its charge by holding directional key during windup.

This is pretty bad, probably should let crusher simply cancel the charge.

Don't think it has to be a pref, I think getting fucked over by pounce is more common that needing to cancel the pounce, so this is a net positive. You can just make it so hitting **resist** stops the pounce/lounge early, just like it cancels do_afters already. Very consistent. Though many people have resist unbinded but for a different reason, that's on them.

code/modules/mob/living/carbon/xenomorph/castes/Warrior.dm
Red-byte3D: > Did you even bother checking attached videos to PR?
> 
> Also, bad lunges are not even possible in first place. (they are sprite clicks)

bad lunges imply that the situation has changed and you no logner want to grab somebody.

Venuska1117:
> > * Crusher can no longer cancel its charge by holding directional key during windup.
> 
> This is pretty bad, probably should let crusher simply cancel the charge.
> 
> Don't think it has to be a pref, I think getting fucked over by pounce is more common that needing to cancel the pounce, so this is a net positive. You can just make it so hitting **resist** stops the pounce/lounge early, just like it cancels do_afters already. Very consistent. Though many people have resist unbinded but for a different reason, that's on them.

I will try do it this way, its reasonable to do it on resist, but i am worried it could cause bad side effects, like you charge, get ignited, and you resist (get stunned on floor), should i do warrior approach of "unique action button" ?

ihatethisengine:
> I will try do it this way, its reasonable to do it on resist, but i am worried it could cause bad side effects, like you charge, get ignited, and you resist (get stunned on floor), should i do warrior approach of "unique action button" ?

Fair concern. That's what I meant by a "different reason". Honestly the best way to handle this IMO would be some kind of "cancel" verb that works like current resist but without resisting, meaning it would cancel do_after and pounce when called. Probably can keep the resist functionality as is (given it is a button on your screen, so can be more intuitive for a new player).

code/modules/mob/living/carbon/xenomorph/castes/Warrior.dm
BlitzArde: > bad lunges imply that the situation has changed and you no logner want to grab somebody.

If within the timeframe that it would take for your sprite to reach the target, you realize that the situation is worse than you thought? You shouldn't be lunging in the first place. That's a game sense issue not a code issue imo. Not against a pref though if that's what people want.

Drulikar:
This really needs to retain a way to cancel since its been heavily discussed regarding this already. I do not think it needs to be a pref.

Allowing resist to cancel is okay, and regarding the concen of resisting fire, you can just have it not resist fire when its canceling a immobilize trait from a pounce source.

But ideally this would be implemented where attempting to move away from your pounce target cancels it, and moving towards it doesn't cancel it (this would not be implemented via immobilize but instead in whatever in the throw code that cares about source still being at a location)

PR #13029, opened 2026-08-25
Makes water bottles that appear in a MRE unfillable

kiVts:
# About the pull request

Locks all water bottles that appear in MREs from transfering from/to them.
exploit bad

<details>

</details>

# Changelog
:cl:
fix: you cannot move reagents from/to water bottles that appear in MREs.
/:cl:


antlersss:
Does this prevent people from pouring out water bottles into beakers or portable extinguishers? If so, I don't think that part should be removed.

If I were to code a solution for this, I would make it so MRE's can't store more than one reagent container. It's more the MRE's problem than the water bottle's.

PR #13051, opened 2026-08-29
Adds Acid Rain to most maps

Red-byte3D:
# About the pull request

This adds an acidic rain weather event to most maps 

This adds an acidic weather event to the game that randomly has a chance to begin, i might just change it so its a %100 chance in rounds but

The general idea is that

1.) Weather events are cool, i think things should be dangerous in certain times
2.) Having weather events where you have to be in safety (caves, fob, etc) allows for rounds where people have a chance to get back up on their feet.
 
I think dangerous weather events are cool and i plan on expanding on these with maybe a meteor shower and even typhoon / hurricanes.

This does 20 damage per tick as to make it so you cannot ignore the damage.

I believe that this will be beneficial for both sides and maybe even allow a bit of more roleplay when the 10 minute timer is moving down and more respite at times.

^^
Things i wanna try here but i havent received an asnwer from maint heads is that i want this to be %100 a thing that happens in a round instead of a chance, i want the game to slow donw randomly and hopefulyl its includal in the game (preferably before the 1.30h mark) will make things last longer and have a bit mroe impact.
^^
<details>

</details>

# Changelog
:cl:
add: Adds acidic rain, damages you with 20 damage per tick.
/:cl:


ihatethisengine:
Acid rain (along with other dangerous hazards, sand/snow storms used to deal damage too IIRC) was removed exactly because people didn't like afking for 10 minutes before they could continue playing. It's cool on paper, but in practice it's a huge unnecessary downtime. Maybe if it was very rare (like once per 10 rounds rare) it would be ok though.

Red-byte3D:
> Acid rain (along with other dangerous hazards, sand/snow storms used to deal damage too IIRC) was removed exactly because people didn't like afking for 10 minutes before they could continue playing. It's cool on paper, but in practice it's a huge unnecessary downtime. Maybe if it was very rare (like once per 10 rounds rare) it would be ok though.

10 minutes is not really an issue, talk to the person next to you, alt tab, roleplay, do ltierally anything.

Nihisohel:
this is awesome dude

kugamo:
Alot of maps have interiors, itd be nice if acid rain weren't just straight damage but a debuff in one way or another. xenos should be immune for lore reasons B) 

slowdown, tiny amounts of damage of everytype, or vision restriction etc idk. So you you could still fight in it at a disadvantage instead of it being a complete momentum killer for outdoor pushes

Maybe restricted to heavy industrial maps so its more fitting, why is it raining acid on mars?

just my thoughts

PR #13054, opened 2026-08-30
Fix Rocket/LTB/Caster ground firing and retire AMMO_EXPLOSIVE

fira:
# About the pull request

For about 4 years, Rocket Launchers and LTB have been unable to fire at the ground. Apparently nobody realized.

I actually fixed that bug for Flare Guns way back in 2022 in #2031 but in my grand ignorance I did not realize that AMMO_EXPLOSIVE implicitely means AMMO_HITS_TARGET_TURF because people were too lazy to type both.

## Why does it break ?

As said in #2031 - essentially projectiles subjected to AMMO_HITS_TARGET_TURF must ignore scatter calculation. Scatter re-targets the projectile by raycasting it, so you're now aiming somewhere else than you clicked, far away. It forgets what turf you wanted to hit.

## Won't removing scatter from LTB / M5 RPG / Caster be a big difference?

I don't think so, they're spec'ed to zero scatter anyway.
Technically the HPC and Dual Plasma Cannons have a small amount of scatter, but I don't think that's really the biggest worry if HPCs are in play.

## Wait, you're removing AMMO_EXPLOSIVE entirely.

Yes because it actually only did two things:
 * Act as AMMO_HITS_TARGET_TURF
 * Cause BE/Suicide to be explosive, which only matters for Mateba HE, which already deals >80 damage to trigger the check, so why even bother?

The first obviously doesn't warrant a flag. The second... I don't know, I guess maybe if you were to somehow try to suicide point blank with a WP rocket it'd make a difference, but i'm pretty sure it's impossible too.

## What about HE Mateba?

Well fixing without removing AMMO_EXPLOSIVE would give it ground targeting, which you don't want on small arms, so I didn't give it AMMO_HITS_TARGET_TURF. Happy ending.

## HPC/RPG warning

This means HPC and RPG will explode where you click, even if you click 2 tiles from you. So be careful. Point at what you want to blow up, don't shoot in the general direction.

## I can't believe you found a holy projectile flag in the land of terrible 24-flags BYOND drought!

Yeah well this goes to show how RELEVANT our flags are and why this is a false problem.

# Changelog
:cl:
fix: Rocket/LTB/HPC now explode on clicked turf properly again rather than flying by. Make SURE you click on what you want to explode, not the general direction.
del: Removed Explosive Ammo properties. It only mattered for BE/Suicide with HE Mateba, and it does that anyway.
balance: Caster Eradicator (HPC) / Dual Plasma Cannons had scatter disabled to target ground properly. Little difference that makes if that thing is pointed at you.
/:cl:

### DIFF
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -12,7 +12,7 @@
 #define HEADSHOT_OVERLAY_HEAVY "heavy_headshot"
 
 // flags_ammo_behaviour
-#define AMMO_EXPLOSIVE (1<<0)
+// (1<<0) was formerly AMMO_EXPLOSIVE. Reuse it only if you must. I encourage you to use components.
 #define AMMO_ACIDIC (1<<1)
 #define AMMO_XENO (1<<2)
 #define AMMO_LASER (1<<3)
--- a/code/_globalvars/bitfields.dm
+++ b/code/_globalvars/bitfields.dm
@@ -83,7 +83,6 @@ DEFINE_BITFIELD(reaction_type, list(
 ))
 
 DEFINE_BITFIELD(flags_ammo_behaviour, list(
-	"AMMO_EXPLOSIVE" = AMMO_EXPLOSIVE,
 	"AMMO_ACIDIC" = AMMO_ACIDIC,
 	"AMMO_XENO" = AMMO_XENO,
 	"AMMO_LASER" = AMMO_LASER,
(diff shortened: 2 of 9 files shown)


fira:
Slapping Balance just for the hell of it because it turns out HPC has low scatter which is removed as a byproduct, that's really the only change i see though

antlersss:
Code LGTM

MistChristmas (approved):

PR #13055, opened 2026-08-30
Queen screech needs roof to rezonate

cuberound:
# About the pull request

Scales down queen screech impact based on the roof above queen head and adds variations of screech icon to indicate that to queen. Roof that does not block fulton gives 0.3 multiplier and roof that does not block cas gives 0.5 multiplier with 1 (the same as now) multiplier for roofs that block cas.

Goals: move combat queen more in the direction of emergency mesures and give drawback to combat queening when the xenos are winning, adding autobalancing factor for combat queen power to reduce snowballing stomps

Nongoals: allow marines to rush hive, remove the ability of queen to fight at all

Gameplay: Gives marines another factor to think about before they push the hive. Scales queen power down if xenos are already winning to keep the rounds more autobalanced. Makes early combat queening harder thus allowing for normal round behavior to occure and for everyone to to have a say in the round outcome and idealy extend the round duration as result. Makes one side stomps  

Lore: queen do not enter combat for no reason and generaly only do that once humans are already assaulting her and her eggs. Sound resonates in tight spaces and spreads out in open air, giving sound based ability more power under thick roof.

image
<details>

</details>

# Changelog
:cl:
add: Added something
balance: queen screech stun gets 0.3 multiplier in areas that do not block mortar (sub tier 2)
balance: queen screech stun gets 0.5 multiplier in areas that are not cas proof (sub tier 3)
imageadd: screech icons indicating the power of the screech based on the area roof
/:cl:


Detective-Google:
we want less cave camping from the xenos not more, I think this would just make xenos consider it less viable to fight in non cave giga choke areas

DarkLordCabbage:
Content aside, I think the screech icons are too vague to tell at a moments glance. The difference in shades is pretty slight! Maybe make it three parentheses instead of two, and have portions greyed out. For example, full strength would be three rings, medium strength would be the regular two, and the lightest would just be the smallest inside one

Steelpoint:
As Google states. We do want to see Xenomorphs taking the fight to Marines instead of hiding within the caves. I also agree that this change will just encourage more cave hugging. 

I personally would not mind a change to how impactful the Queens screech is, but it must be paired with changes to how Xenomorphs engage with Marines. I know that there are changes being considered in that regard such as letting Xenos construct more durable defensive structures.

PR #13061, opened 2026-08-31
fixes reinforced wall behavior

cuberound:
# About the pull request

okey we did not have working reinforced wall behavior for about 8 years now. what did I change? moved the behavior of reinforced wall to be based on the damage_cap of the wall rather then it being suptype of r_wall, fixed loads of bugs and stuff along the way. allowing dimond drill to be used on all nonhull walls not just stubtypes of r_wall (we have like one in the game idk can be removed)

REINFORCED WALL BEHAVIOR LINKED TO damage_cap >= HEALTH_WALL_REINFORCED

reinforced walls will finaly behave like reinforced walls in regards to acid, claws, repairs, nesting

<details>

</details>

# Changelog
:cl:
fix: all reinforced walls need strong acid to be melted
fix: you can nest host on all reinforced walls both by clicking the weed and the wall
fix: you need very sharp claws for damaging all reinforced walls
fix: you can no longer repair any reinforced wall just by using metal on them
/:cl:


Bibby0110:
Seems counterproductive to make it based on health and not the subtype? Worse code at least, since this is what subtypes are meant for

cuberound:
> Seems counterproductive to make it based on health and not the subtype? Worse code at least, since this is what subtypes are meant for

 walls are made for mappers not developers, mappers want to inherit the icon from the basic to reinforced and to hull subtypes for GOOD REASON when you are mapping in SDMM and you have wall/hybrisa/marshalls, wall/hybrisa/marshalls/reinforced, wall/hybrisa/marshalls/hull searching for wall/hybrisa/marshalls shows all three, wheras if it was subtype you would need to search for wall/reinforced/hybrisa/marshalls and for wall/hybrisa/marshalls and so on. I hope that makes this design decison clear.

Bibby0110:
> > Seems counterproductive to make it based on health and not the subtype? Worse code at least, since this is what subtypes are meant for
> 
> walls are made for mappers not developers, mappers want to inherit the icon from the basic to reinforced and to hull subtypes for GOOD REASON when you are mapping in SDMM and you have wall/hybrisa/marshalls, wall/hybrisa/marshalls/reinforced, wall/hybrisa/marshalls/hull searching for wall/hybrisa/marshalls shows all three, wheras if it was subtype you would need to search for wall/reinforced/hybrisa/marshalls and for wall/hybrisa/marshalls and so on. I hope that makes this design decison clear.

This sounds like you want to make some fluff walls instead? Should be pretty easy to do with a subtype of normal walls. Don't most map tools have a palette feature also?

The problem here is that it makes it very unclear what an r-wall is and what and r-wall isn't.
What if I want to map in a wall with r-wall health, but without the r-wall properties? I can't do that anymore, but I could before.

EDIT: Also why does this make it better for mappers? You still need to set the damage_cap via an edit in your mapping tool for this to matter, don't you? And you shouldn't be doing that, as it means we can't balance wall damage without editing every map

ori-disillusion-mirnov:
this should really be tagged balance and not fix

MistChristmas:
I feel like you probably should have just stuck to it being health based rather than adding the variables, that would now change the strength of a bunch of walls.

I don't think we really have much need to be granular about wall health/acidability/clawability seperately at the moment. 
And this is just going to cause a need for a bunch more follow up.

Whilst the first option needs less. Searching for `wall/.*/reinforced` And just ensuring that they have the right damage_cap sounds like a lot less work than setting 3 different variables across how ever many walls would now need to be modified.

PR #13064, opened 2026-09-01
Predaliens and Kings can pry open dropships and lock them + ERT shuttles can be interacted by Queen/Predalien/King

Unknownity:
# About the pull request

Predaliens and Kings are now smarter, they can pry open locked dropships and containment doors. King can now also disable lifeboats along with the Predalien (who already could).

Predaliens and Kings can also lock the dropships but they are unable to start hijack themselves.

ERT Shuttles have proper reinforced airlocks now but they can be forced open by Queens, Predaliens and Kings and their console can also be disabled by the aforementioned castes.

**Queen is still the only one who can start hijack.**

Believe it or not but Predaliens were USED to be able to do this already, it was a niche undocumented feature that not many people knew. However at some point the code was refactored around hijack and dropships and their aforementioned consoles and the predalien lost the ability to do it anymore. I believe whoever did it forgot about the predalien there since they were the only other xeno who had the mechanic (is_intelligent, the proc that handled hijacking before. Now it only handles lifeboats, yes predalien can disable lifeboats RIGHT NOW) apart from the queen.

Well I've returned it back now in a working state. Technically this should have been a bug/oversight fix but i'll keep it as a balance change since its been so long and it was a very niche feature that almost nobody knew about.

As for ERT shuttles, I now made their airlocks much more reinforced, since they're supposed to act like dropship hatches (but we dont have unique hatches for shuttles), no longer you can manipulate them via wiring or welding nor slash them as a xeno but now Queens and Predaliens can force them open. They can also now disable the ERT shuttle entirely by interacting with the console. A good way to prevent people from delaying or fleeing now.

NOTE: disabling the shuttle does not cancel it from already spooling if someone managed to make it start flying before you disabled it, you should probably still leave while you can.

**edit: The new direction of the PR is to provide additional xenos who can pry open locked down dropships without allowing them to start hijack as that should be the Queen's authority.**

Made sure things work normally, predalien and king still cant hijack while queen can.

<details>

</details>

# Changelog

:cl:
balance: Predaliens and Kings are now able to pry open locked dropships and containment airlocks.
balance: Predaliens and Kings are now able to lock the dropship but cannot start hijack on their own.
balance: King can now disable the lifeboats along with the Queen and Predalien.
balance: ERT Shuttles' airlocks are now reinforced but can be forced open by Queens/Predaliens/Kings and their consoles can also now be disabled by them.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/hive_status.dm
+++ b/code/modules/mob/living/carbon/xenomorph/hive_status.dm
@@ -9,7 +9,6 @@
 
 	var/hivenumber = XENO_HIVE_NORMAL
 	var/mob/living/carbon/xenomorph/queen/living_xeno_queen
-	var/mob/living/carbon/xenomorph/predalien/living_xeno_predalien
 	var/egg_planting_range = 15
 
 	/// Toggles for the hive that are reset on queen death unless hive_flags_locked
@@ -828,7 +827,7 @@
 	return length(hive_structures[name_ref])
 
 /datum/hive_status/proc/abandon_on_hijack()
-	var/area/hijacked_dropship = get_area(living_xeno_queen && living_xeno_predalien)
+	var/area/hijacked_dropship = get_area(living_xeno_queen)
 	if(!hijacked_dropship)
 		return FALSE
 	var/shipside_humans_weighted_count = 0
--- a/code/modules/mob/living/carbon/xenomorph/hive_status.dm
+++ b/code/modules/mob/living/carbon/xenomorph/hive_status.dm
@@ -9,6 +9,7 @@
 
 	var/hivenumber = XENO_HIVE_NORMAL
 	var/mob/living/carbon/xenomorph/queen/living_xeno_queen
+	var/mob/living/carbon/xenomorph/predalien/living_xeno_predalien
 	var/egg_planting_range = 15
 
 	/// Toggles for the hive that are reset on queen death unless hive_flags_locked
@@ -827,7 +828,7 @@
 	return length(hive_structures[name_ref])
 
 /datum/hive_status/proc/abandon_on_hijack()
-	var/area/hijacked_dropship = get_area(living_xeno_queen)
+	var/area/hijacked_dropship = get_area(living_xeno_queen && living_xeno_predalien)
 	if(!hijacked_dropship)
 		return FALSE
 	var/shipside_humans_weighted_count = 0
--- a/code/__DEFINES/xeno.dm
+++ b/code/__DEFINES/xeno.dm
@@ -107,8 +107,8 @@
 
 #define XENO_STARTING_CRYSTAL 100 //How much building resource the queen gets to start with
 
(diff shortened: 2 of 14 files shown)


code/modules/mob/living/carbon/xenomorph/hive_status.dm
```diff
@@ -9,6 +9,7 @@
 
 	var/hivenumber = XENO_HIVE_NORMAL
 	var/mob/living/carbon/xenomorph/queen/living_xeno_queen
+	var/mob/living/carbon/xenomorph/predalien/living_xeno_predalien
```
Drulikar: Nothing is ever assigned to this variable (so its always null). Also its kinda nonsensical because its not like a predalien there can only be one of. There can be multiple queens technically (but not in normal play), but only one is THE queen of the hive.

---

I would just remove this.

code/modules/mob/living/carbon/xenomorph/hive_status.dm
```diff
@@ -827,7 +828,7 @@
 	return length(hive_structures[name_ref])
 
 /datum/hive_status/proc/abandon_on_hijack()
-	var/area/hijacked_dropship = get_area(living_xeno_queen)
+	var/area/hijacked_dropship = get_area(living_xeno_queen && living_xeno_predalien)
```
Drulikar: This doesn't make sense. You don't boolean AND two different things and then expect get_area to handle that. You COULD however boolean OR (`||`) them to leverage how byond handles falsey items and short circuiting (so it will use `living_xeno_predalien` if `living_xeno_queen` is falsey). But its not exactly the most intuitive thing.

Right now you're sort of requiring both to be set - be sure to test your changes:
Image

---

It'd be better at this point to just have abandon_on_hijack called with an argument of what area so you don't have to determine it (you would easily have that information at both places that call this proc)

Drulikar (dismissed):
In addition:
- Any reason you aren't doing the same for King?
- You should also look at `/obj/docking_port/mobile/marine_dropship/proc/sneak_hijack_check()` if you goal is to have predalien capable of hijack; **but really I dunno if I am keen on more than Queen hijacking because it would be really jank for queen to be stuck groundside or be sent to brazil because of a predalien wanted to hijack. Things just work out a lot nicer when a single xeno (queen) causes hijack.**

This PR is DNM until you sort out the issue of how to handle the situation that there's multiple hijack capable xenos present in a hive. I am okay with a queen causing a predalien to be sent to brazil, but I am not okay with a queen being sent to brazil or stranded. I highly recommend you just leave it to queen to trigger hijack.

Unknownity:
Changed the direction of the PR.

King has been added along the smart xenos.
King and Predalien are still able to pry open dropship/shuttle doors and disable lifeboats and ERT shuttles.
King and Predalien can lock the dropship but not start hijack.

**Only the queen can initiate hijack.**

Nihisohel:
stale exempt until balance reviewed, this is kind of a slippery slope even for me to chime in for

Nihisohel (approved):

PR #12586, opened 2026-07-04
Tank Desant: Hardpoint Rebalances, Additions & Reworks

SunriseOverYourHead:
# About the pull request

**Depends upon #10580 AND should be PR'd onto it, not onto master, otherwise Desant's commits will show up.**
**This is a separate, atomized PR to make it easier for maintainers to review it.**
**Commits are ATOMIZED. Please review the PR through them, starting from c657a8915602f391a515e7d2063545d8a614ee1a**

**Extensively tested in a private server. Don't take my word for it. Download the branch and try to find a major bug/glitch.**

**Full list of changes:**

* Switches the Longstreet Tank's dir-based aiming into a more realistic approach where the turret swivels around the hull in real time, in accordance with the Gunner's mouse position.

* Inertia-based turning: Longer turns are faster, can overshoot. Etc. Just like a real turret drive

* Enables **secondary-hardpoint slaving** to the Driver, who can now fire in a limited (120º) degrees arc based on wherever the Gunner is aiming.

* **LZR-N & DRG-N;** AKA, Primary and Secondary flamers: They now both support 'fire mode' switching (GLOB and STREAM) just like a nozzle'd M240 flamer.

* **LZR-N & DRG-N:** Now support custom fuel mixtures. CLF3. HCN. Sticky Napalm. Naphta. Research Chems. Welding Fuel. This also comes with a more realistic fuel consumption modelling. The fuel economy for the primary wasn't tweaked: DRG-N can still only take 60 GLOB shots; LZR-N has the exact same fuel economy as the M240.

* **LTB**: The dreaded hardpoint that some believe to be 'unbalanceable'. I disagree. Here is a solution that is both balanced and SOVL: Utilizes a sniper-like aiming with a 5 second wind-up. Can not stun queen or King anymore. Can still stun T3s. 300 damage with two separate fall-offs. Slightly more powerful than a SADAR HE ROCKET in practice. If a target runs around it too fast, it can break the lock by out-swiveling the turret. Aiming gets blocked by smoke and walls, just like a normal sniper aimed shot. See the video to understand how it works. **Unlike the previous LTB,** this is now atrocious at close-range as any xeno can just spin around you to lose the lock. **No way to speed up the lock**, but you can fire on the move, provided you can continue tracking your target.

* **LTB Single-Shells**: Does away with LTB magazines and instead uses single shells for a more cumbersome, yet satisfying reloading. Sprites are a courtesy of **holdemann**.

[...]

PR #13071, opened 2026-09-02
Reduces Tank's server population requirement from 200 to 160. 

Ammoniacres:
# About the pull request

Reduces the requirement for Tank Crew roles from >= 200 to 160. This means that when 160 or more players are playing a given round, tank crew roles will be available to play and the Longstreet Tank will be available to play roundstart. 

I have played CMSS13 for roughly 2000 hours, give or take, with a lot of observer hours on top. I have seen Tank feature in less than a handful of rounds over that period of time, and I hear about it or experience it progressively less frequently as the game ages. 

A lot of the power in Tank as a gameplay feature is its rarity and presence because of it, as well as the mechanical power of the vehicle as it presents itself, but it never features, and has become a bit of a running joke as far as content is concerned. 

I value the authenticity and gravitas of gameplay events and think it's important to not overrepresent features, but I also think it's worth evaluating where the game is currently at, and recognizing that CM isn't pulling 200 people even semi-consistently, and doesn't function well even far below those player counts in terms of server performance. 

Tank is fun. Fun should feature more. Even rare fun should be represented.

I think 160 is a good upper limit for what highpop looks like on this server, though I usually see the server hover between 120-140 on peak hours. Someone with access to the numbers being able to corroborate/object to this would be appreciated. 

Also, a PR like [[[12586]]](https://github.com/cmss13-devs/cmss13/pull/12586) shouldn't feature with pop requirements the way they are. 

<details>

</details>

# Changelog

:cl:
balance: reduced Tank Crew's server population requirements from 200 players to 160!
/:cl:


frozentsbgg:
No, The Tank is to be reserved for very rare special events and with a population that supports combined arms.
If you want a 160 pop vehicle the requested fixes can be made to the APC instead.

tallfission:
^ ^ ^ not a maintainer ^ ^ ^

frozentsbgg:
> ^ ^ ^ not a maintainer ^ ^ ^

Screenshot 2023-03-24 171437

Detective-Google (changes requested):
I've discussed this back and forth a little with steel and drath and the TLDR is that we don't think the tank is in a state for the spotlight yet, and is better served in it's current semi-easter egg state. 

The most important thing we would want to see before the tank can be increased in availability is some mechanism of xenos moving past the vehicle whether it is destroyed or functional, similar to how a xeno can climb over a ledge or window frame. This would prevent the tank from being rammed into chokes and used as a 3x3 mobile barricade.

The "Tank Desant" PR, I believe, implements some features for this, though considering the breadth of it's content it is unlikely to be reviewed or merged in the immediate future until it's atomized, *ideally* we'd see it merged before we increase availability, but at the very least the above issue must be handled before we're interested in this.

Ammoniacres:
I'll work on this on the weekend, I think.

PR #13088, opened 2026-09-04
Lets Xenos track the King in the queen/leaders compass HUD thing

kugamo:
# About the pull request

When the hive gets a king, the king will now show up in the options for tracking in the HUD compass alongside the existing options of queen, leaders, hive core, and tunnels

Its a bit of functionality thats been lacking for xenos since the addition of the king. Its on par with the Queen in leadership surely, bizarre its not in the options for the HUD compass

# Changelog
:cl: Kugamo
add:  xenos can now track their KING in their HUD compass
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/King.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/King.dm
@@ -135,6 +135,7 @@
 
 /mob/living/carbon/xenomorph/king/death(cause, gibbed)
 	. = ..()
+	hive.living_xeno_king = null
 	if(hive)
 		hive.setup_banned_allies()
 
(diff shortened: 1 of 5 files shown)


code/modules/mob/living/carbon/xenomorph/castes/King.dm
```diff
@@ -77,7 +77,7 @@
 
 /mob/living/carbon/xenomorph/king/Destroy()
 	UnregisterSignal(src, COMSIG_MOVABLE_PRE_MOVE)
-
+	hive.living_xeno_king = null
```
antlersss, line 80: The king can still be tracked after they die. This is only called when the corpse of the king is gibbed or deleted for whatever reason, which I assume is not intentional. Consider adding another `hive.living_xeno_king = null` to the death proc for the king.

kugamo: lol I only nullified it there initially, which caused the first commit to fail linters. you're right though, not intentional, Ill readd it back keeping the one in destroy

Nihisohel (approved):

PR #13092, opened 2026-09-05
add Update Paths.sh for linux

ori-disillusion-mirnov:
# About the pull request
title

yay for linux, still cant drag and drop files but i like this better than reading the .bash file to figure out how the update paths works

image

# Changelog
:cl:
code: add .sh file for update paths
/:cl:


tools/UpdatePaths/Update_Paths.sh
Drulikar, line 1: Rename the ` ` to `_` or remove it in the filename so you don't need to quote it in terminal. Generally linux avoids whitespace in folders/files where possible.

Drulikar (changes requested):

Drulikar (approved):

PR #13104, opened 2026-09-06
xeno fire resistance buff cost 1 -> 2

Detective-Google:
# About the pull request

this pr increases the cost of fire resistance boon from 1 to 2 resin

the fire resistance boon was buffed to confer total fire immunity as opposed to before when it was only immunity to being *set on fire* and not the contact damage, which drastically increased its utility, with the cost remaining the same. By bumping the cost from 1 to 2 it's still cheap and xenos can easily afford it to suit their needs; but they can't keep it up forever with just one pylon active, they'll have to have periods of vulnerability in between.

# Changelog
:cl:
balance: fire resistance boon cost increased from 1 to 2.
/:cl:

### DIFF
--- a/code/modules/cm_aliens/hivebuffs/hivebuff.dm
+++ b/code/modules/cm_aliens/hivebuffs/hivebuff.dm
@@ -464,6 +464,7 @@
 	name = "Boon of Fire Resistance"
 	desc = "Makes all xenomorphs immune to fire for 5 minutes."
 	tier = HIVEBUFF_TIER_MINOR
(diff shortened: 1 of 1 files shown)


Drulikar (approved):

ihatethisengine:
I'd suggest making it a major buff, so you only get it when it's just the FOB left to deal with.

Drulikar:
>I'd suggest making it a major buff, so you only get it when it's just the FOB left to deal with.

I imagine more will be changing before long.

PR #13107, opened 2026-09-06
JSON Save Tree for Preferences

Drulikar:
# About the pull request

This PR sort of ports https://github.com/tgstation/tgstation/pull/70492 but does a fair bit of things different.

I probably would not TM this unless the path is decided on (not going to change during TMing), and a decision is made about the export verb.

Some things of note generally:
- ~~I currently have the verb available to users to export JSON.~~ **Let me know if the verb should be given another way, or it needs a way to be disabled, or if we want it at all.** There is currently no way for a client to import (TG doesn't currently have an import either; though we'd need to move name bans if we do that and do a lot of sanitization)
- The old byond save file is left intact
- Save file verison upgrade now triggers a save (so it doesn't keep reattempting each load if they don't happen to save manually)
- I removed two situations that sanitization would trigger saving since its niche and if that should be done, any value changing should trigger a save
- I rearranged `save_character` to be before `save_preferences` in `process_link` because only `save_preferences` by default writes to disk
- The logic in preferences datum creation is changed so that things like tgui say loading, quick start job selection, etc occur even if a preference file was loaded
- Unlike the TG implementation, we use an alist and allow direct access to the alist (but I left the getters/setters for now - **but I can remove them**), and use file2text instead of rustg_file_read. 
- Export JSON currently would leave lingering files around in data because ftp/DIRECT_OUTPUT are async so it can't delete the file. I made it not have a timestamp so atleast it'd be one per ckey, disabled the verb to use it, and it probably in the future should perform `fdel("data/preferences_export_working_directory/")` every boot ever we re-enable this functionality

The ultimate reason for this change is to reduce preference overtiming. See Screenshots & Videos below for more info (basic profiler tests - but Harry was also performing some larger N tests in contributors channel). For a while I was doing this as a byond savefile still (to avoid overhead of json_encode and json_decode) but then I was seeing byond save files stuggle with saving larger files...

<details>

Sequencing of events for a byond save file upgrade (that wasn't a tree) as well as a new player connect (Drulikar)
image

Another round (meaning save file already converted)
image

Json export
image

Load slot window: 
image

Master profile (save slot, open role choices, then reload slot)
image

Early json profile (save slot, open role choices, then reload slot)
image

Early byond tree profile (save slot, open role choices, then reload slot)
image

However then I started to see byond saves to be slow with more data: 
image

But then with `dba7c62` going back to json saves are fast again:
image

Using file2text instead of rustg_file_read
image

</details>

# Changelog
:cl: Drathek
qol: Character slot selection now displays the current slot differently
fix: Preference version upgrades now trigger a save immediately
fix: Debugger option quick start will now set CO job selection regardless of save data existing
server: Preferences now use json
/:cl:


code/__HELPERS/logging.dm
```diff
@@ -19,10 +19,14 @@
...
 
+//wrapper macros for easier grepping
 #define DIRECT_OUTPUT(A, B) A << B
-#define SEND_TEXT(target, text) DIRECT_OUTPUT(target, text)
+#define DIRECT_INPUT(A, B) A >> B
+#define SEND_IMAGE(target, image) DIRECT_OUTPUT(target, image)
 #define SEND_SOUND(target, sound) DIRECT_OUTPUT(target, sound)
+#define SEND_TEXT(target, text) DIRECT_OUTPUT(target, text)
 #define WRITE_FILE(file, text) DIRECT_OUTPUT(file, text)
+#define READ_FILE(file, text) DIRECT_INPUT(file, text)
```
hry-gh, line 29: i kinda hate this silly hiding of byond internals but we have so much of it already

hry-gh (dismissed):

hry-gh:
as an aside, letting users export with no import functionality seems totally pointless

code/datums/tree_savefile.dm
```diff
@@ -0,0 +1,127 @@
...
+
+/// Wipes the entire tree
+/datum/json_savefile/proc/wipe()
+	tree?.Cut()
+
+/datum/json_savefile/proc/load()
+	if(!path || !fexists(path))
+		return FALSE
+	try
+		tree = json_decode(rustg_file_read(path))
```
hry-gh: rustg_file_read is just slower than file2text, fyi

benching (under wine, so...?):
<details>
=== 100 B (_bench_100.txt) ===
<~>
n (samples) = 5
tick_lag = 1 (~10 Hz)

file2text:
   ~711 (± 201) runs/tick (95% confidence: 535 .. 887)
rustg_file_read:
   ~208 (± 6) runs/tick (95% confidence: 203 .. 214)
<~>
=== 1 KB (_bench_1000.txt) ===
<~>
n (samples) = 5
tick_lag = 1 (~10 Hz)

file2text:
   ~739 (± 46) runs/tick (95% confidence: 698 .. 779)
rustg_file_read:
   ~158 (± 30) runs/tick (95% confidence: 131 .. 185)
<~>
=== 5 KB (_bench_5000.txt) ===
<~>
n (samples) = 5
tick_lag = 1 (~10 Hz)

file2text:
   ~586 (± 64) runs/tick (95% confidence: 530 .. 642)
rustg_file_read:
   ~166 (± 24) runs/tick (95% confidence: 144 .. 187)
<~>
=== 23 KB (prefs-sized) (_bench_23000.txt) ===
<~>
n (samples) = 5
tick_lag = 1 (~10 Hz)

file2text:
   ~510 (± 41) runs/tick (95% confidence: 474 .. 547)
rustg_file_read:
   ~160 (± 28) runs/tick (95% confidence: 135 .. 185)
<~>
=== 50 KB (_bench_50000.txt) ===
<~>
n (samples) = 5
tick_lag = 1 (~10 Hz)

file2text:
   ~344 (± 56) runs/tick (95% confidence: 295 .. 393)
rustg_file_read:
   ~123 (± 54) runs/tick (95% confidence: 76 .. 171)
<~>
=== 100 KB (_bench_100000.txt) ===
<~>
n (samples) = 5
tick_lag = 1 (~10 Hz)

file2text:
   ~253 (± 9) runs/tick (95% confidence: 245 .. 262)
rustg_file_read:
   ~137 (± 9) runs/tick (95% confidence: 129 .. 145)
<~>
=== 500 KB (_bench_500000.txt) ===
<~>
n (samples) = 5
tick_lag = 1 (~10 Hz)

file2text:
   ~67 (± 4) runs/tick (95% confidence: 63 .. 71)
rustg_file_read:
   ~56 (± 5) runs/tick (95% confidence: 52 .. 61)
<~>
=== 1 MB (_bench_1000000.txt) ===
<~>
n (samples) = 5
tick_lag = 1 (~10 Hz)

file2text:
   ~34 (± 0) runs/tick (95% confidence: 33 .. 35)
rustg_file_read:
   ~36 (± 0) runs/tick (95% confidence: 35 .. 37)
<~>
</details>

Drulikar:
> as an aside, letting users export with no import functionality seems totally pointless

This is their reasoning: https://github.com/tgstation/tgstation/pull/75014

But I discovered its leaving all the temp json files around, so changed it to not be time stamped and disabled the verb for now.

code/datums/tree_savefile.dm
Drulikar: Changed to it and a single new profile result for posterity 
image

hry-gh (dismissed):

cm13-github:

cm13-github:

PR #13139, opened 2026-09-10
Optimization: icon2base64 rust-g call optimization

antlersss:
# About the pull request

Optimizes `icon2base64` by replacing slow, native savefile calls with a call to `rust-g`'s `rustg_iconforge_generate_headless`, making icon conversion 2.5x as fast.
Legacy icon conversion is kept in the codebase as a fallback option.

This should make server initialization faster, as well as mitigate lag spikes related to opening the inventory stripping panel on individuals.

Currently `icon2base64` takes up quite a significant amount of self time when reviewing aggregated data from the most recent 100 rounds. It's no surprise why, considering it writes to a temporary save file and then uses that to return a base64 representation of an icon. Using rust-g's library calls to bypass the native savefile implementation in its entirety promises better performance.

Benchmarked using [wrench-extra](https://github.com/Lohikar/wrench-extra/blob/master/util.dm):

<details>
<summary>Original implementation: (60 runs/tick)</summary>

Benchmark verb:
```
CLIENT_VERB(bench_icon_conversion)
	set name = "Bench icon2base64"
	set category = "Admin"

	var/icon/icon = icon('icons/UI_icons/map_blips.dmi')
	var/list/icon/icon_states = list()
	for(var/icon_state in icon_states(icon))
		icon_states += icon('icons/UI_icons/map_blips.dmi', icon_state = icon_state)

	BEGIN_BENCH_WITH_SAMPLES(1, 1000)
		BENCH_EXPR("init_minimap_icons", icon2base64(pick(icon_states)))
	END_BENCH
	to_chat(mob, SPAN_INFO("Benchmark complete."))
```
Benchmark output:
```
<~>
n (samples) = 1000
tick_lag = 0.5 (~20 Hz)

init_minimap_icons:
	~60 (± 4) runs/tick (95% confidence: 59 .. 60)
<~>
```
</details>

<details>
<summary>Rust-g implementation: (163 runs/tick)</summary>

Benchmark verb:
```
CLIENT_VERB(bench_icon_conversion)
	set name = "Bench icon2base64"
	set category = "Admin"

	var/icon/icon = icon('icons/UI_icons/map_blips.dmi')
	var/list/icon_states = list()
	for(var/icon_state in icon_states(icon))
		icon_states += icon_state

	BEGIN_BENCH_WITH_SAMPLES(1, 1000)
		BENCH_EXPR("init_minimap_icons", icon2base64('icons/UI_icons/map_blips.dmi', pick(icon_states)))
	END_BENCH
	to_chat(mob, SPAN_INFO("Benchmark complete."))
```
Benchmark output:
```
<~>
n (samples) = 1000
tick_lag = 0.5 (~20 Hz)

init_minimap_icons:
	~163 (± 14) runs/tick (95% confidence: 162 .. 164)
<~>
```
</details>

<details>
image

image
</details>

# Changelog

:cl:
code: icon2base64 now uses external rust-g calls instead of slow native /savefile calls.
/:cl:

### DIFF
--- a/code/__HELPERS/icons.dm
+++ b/code/__HELPERS/icons.dm
@@ -844,9 +844,13 @@ world
 	)
 	var/extern_result = rustg_iconforge_generate_headless("tmp/forged.png", json_encode(list("sprite" = sprite_object)), TRUE)
 
-	if (extern_result["file_path"] != "tmp/forged.png" || !fexists("tmp/forged.png"))
+	var/icon64 = ""
+	if (fexists("tmp/forged.png"))
+		icon64 = rustg_hash_file(RUSTG_RNG_FORMAT_BASE64, "tmp/forged.png")
+
+	if (extern_result["file_path"] != "tmp/forged.png" || icon64 == "")
 		// Rust-g errored out, fall back to old implementation
-		log_debug("External rust-g library call for icon2base64 errored out! Reverting to legacy fallback implementation.")
+		log_debug("External rust-g library call for icon2base64 failed! Reverting to legacy fallback implementation.")
 
 		var/savefile/save_buffer = new /savefile("tmp/forged.sav")
 		save_buffer["icon"] << icon(icon_file, icon_state = icon_state)
@@ -861,7 +865,7 @@ world
 		return
 
 	// No need to delete temp file, it'll be overridden by next proc call
-	return rustg_encode_base64(rustg_file_read("tmp/forged.png"))
+	return icon64
 
 /**
  * Center's an image.
--- a/code/__HELPERS/icons.dm
+++ b/code/__HELPERS/icons.dm
@@ -873,6 +873,7 @@ world
 	fdel("tmp/forged.sav")
 
 /// Applies rustg TRANSFORM_OBJECT calls in-house, allowing icons to be manipulated even if rustg fails during the iconforge process.
+/// This function will continue execution even if an individual transform object fails. In the case that it does, it will return FALSE. Otherwise, it will return TRUE.
 ///
 /// YOU SHOULD NOT BE USING THIS UNLESS YOU KNOW WHAT YOU'RE DOING!
 /// To learn, read the documentation for iconforge at:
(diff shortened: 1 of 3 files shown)


boskoramen:
have you used tracy or the profiler to compare execution time before and after? I'd recommend setting up wrench macros locally to compare performance also over a sample: https://github.com/Lohikar/wrench-extra/blob/master/util.dm

e.g.
```dm
BEGIN_BENCH(1)
  BENCH_BLOCK("")
    . = .
END_BENCH
```

see coderbus for more examples or to play around with wrench macros

---

having these numbers to compare would be really helpful to understand the impact of this PR.

also, iirc, there are issues with trying to offload icon generation to rustg. not up-to-date with this stuff though, coderbus might have more insight

Drulikar:
Please provide some comparisons even if is just profiler. We've not found save files to be bad performance. But modifying the same save file multiple times can't be batched without providing byond a single object to work on per operation (e.g. a whole alist tree, or a whole datum to serialize)

antlersss:
Benchmarking completed and added to the PR description.

code/__HELPERS/icons.dm
Drulikar, line 1: You should probably rename the old proc to costly_icon2base64 or something and have your fallback in icon2base64 just call that since this is not a perfect replacement as rust is only handed a dmi and icon_state instead of an icon.

For example on master: 
```js
/client/verb/testing()
	set name = "testing verb"

	var/icon/ash = icon('icons/effects/effects.dmi', "ash")
	var/icon/byond = icon('icons/effects/effects.dmi', "byondlogo")
	ash.Blend(byond, ICON_MULTIPLY)
	ash.Scale(500,500)
	var/result = icon2base64(ash)
	var/data = HTML_FILE("<img src='data:image/png;base64, [result]' />")
	show_browser(src, data, "Testing", "testpopup")
```

Image

But you *should* be able to achieve the same with rust/iconforce, but you currently have no means in your proc to provide this data e.g. https://github.com/cmss13-devs/cmss13/blob/e712b291f42186e3ce18f0155bd51ffe17367e53/code/__DEFINES/__rust_g.dm#L339
 so maybe also consider a third optional argument for a list of extra operations with a comment pointing to the iconforge info above (maybe it has to use rustg_iconforge_generate instead of rustg_iconforge_generate_headless in that situation tho.

I don't really understand the rustg hash source enough to know what are possible returns to know if your failure checks are even correct/necessary.

Drulikar (changes requested):

PR #13140, opened 2026-09-11
Being flayed makes you unrecognizable

kugamo:
# About the pull request

If someone gets flayed, their corpse becomes unrecognizable (loses its name). Unless you're intimately familiar with every single marines bone structure/facial musculature it makes no sense why they'd be recognizable after having their face removed 

Game atmosphere, adds to the horror by correctly simulating how observing a flayed corpse would work.

dreamseeker_OHoc59yYvm
dreamseeker_L0nDH693E6

# Changelog

:cl: Kugamo
add: Flayed corpses no longer identifiable by name.
spellcheck: Wording of examined humans' age, size, and build streamlined.
/:cl:

### DIFF
--- a/code/modules/cm_preds/yaut_weapons.dm
+++ b/code/modules/cm_preds/yaut_weapons.dm
@@ -916,7 +916,9 @@
 				victim.apply_damage(22, BRUTE, limb, sharp = TRUE)
 			for(var/obj/item/item in victim)
 				victim.drop_inv_item_to_loc(item, victim.loc, FALSE, TRUE)
-				victim.status_flags |= PERMANENTLY_DEAD
+			victim.status_flags |= PERMANENTLY_DEAD
+			ADD_TRAIT(victim, TRAIT_FLAYED, user)
+			victim.name = "???"
 			victim.add_flay_overlay(stage = 3)
 
 			//End the loop and remove all references to the datum.
--- a/code/modules/cm_preds/yaut_weapons.dm
+++ b/code/modules/cm_preds/yaut_weapons.dm
@@ -917,7 +917,7 @@
 			for(var/obj/item/item in victim)
 				victim.drop_inv_item_to_loc(item, victim.loc, FALSE, TRUE)
 			victim.status_flags |= PERMANENTLY_DEAD
-			ADD_TRAIT(victim, TRAIT_FLAYED, user)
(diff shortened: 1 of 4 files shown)


code/modules/cm_preds/yaut_weapons.dm
```diff
@@ -916,7 +916,9 @@
 				victim.apply_damage(22, BRUTE, limb, sharp = TRUE)
 			for(var/obj/item/item in victim)
 				victim.drop_inv_item_to_loc(item, victim.loc, FALSE, TRUE)
-				victim.status_flags |= PERMANENTLY_DEAD
+			victim.status_flags |= PERMANENTLY_DEAD
+			ADD_TRAIT(victim, TRAIT_FLAYED, user)
```
fira, line 920: That's not how source works, you typically want a define-string that describe what action caused this

But in all honesty this doesn't use any of the upsides of traits, so it could just have been a mob_flags

fira (dismissed):

code/modules/cm_preds/yaut_weapons.dm
kugamo: I was originally thinking I'd use a status_flag, but those are all used :( 

fix pushed :o)

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -102,7 +107,14 @@
 
 	if(id_paygrade)
 		msg += "<EM>[rank_display] </EM>"
-	msg += "<EM>[src]</EM>!\n"
+
+	msg += "<EM>[src]</EM>"
+
+	if(mob_flags & MOB_FLAYED)
+		msg += " You cannot tell anymore, [t_He] doesn't have a face!\n"
```
antlersss: Add a newline before this message; the first line of examining humans is reserved for the name, and not extra flavor. It looks strange directly after the '???'. I like the text, though.

kugamo: ~~I disagree, its directly relevant to the first line of examining humans. I wouldnt be against adding more spaces or a semi-colon or something. I just think it'd read worse on a newline~~

eh youre probably right, no need to break convention for a flayed desc

Nihisohel (changes requested):
Image

Nihisohel (approved):

PR #13141, opened 2026-09-11
No shuttle mines

Drulikar:
# About the pull request

This PR makes it so marines cannot deploy antigrief checking mines in shuttle turfs; as well as just generally in non-open turfs.

Leme know if there's other turf types that need to be included or excluded; or if I shouldn't even check antigrief mines at all.

Mechanical enforcement against rule breaks.

<details>

image

</details>

# Changelog
:cl: Drathek
balance: Marines can no longer deploy mines in shuttle turfs
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/mine.dm
+++ b/code/game/objects/items/explosives/mine.dm
@@ -90,7 +90,12 @@
 	if(active || user.action_busy)
 		return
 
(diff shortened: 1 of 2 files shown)


Detective-Google (approved):
code

fira (approved):
This could've been a good candidate for generalizing it as turf_flags rather than putting an hyperspecific boolean, but i see no problem with that

PR #13159, opened 2026-09-12
Research Biomass and "Botany" Adjustments

Ammoniacres:
# About the pull request

Rebalances biomass escalation and initial purchase costs across the board to adjust for the increased demand on organs from outside sources, adds **experimental first-aid kits** and **compact advanced defibrillators** for purchase at clearance 4 and 5 respectively. 

Increases the plasma reagent storage cap from 120 > 350.
Increases the reagent production of dedicated organ farms from 10 per tic to 15 per tic.
Lets xenomorph organs be planted twice, either to produce weeds or a chosen gland, and removed the second drone heart in Research's containment
Added egg plasma to the Royal Gland extractor.

And updates descriptions across the board.

Biomass is a very contested resource now that a lot of research's resource gain and best options for stimulants are locked behind engaging with the systems, but the actual mathematics of using organs to produce reagents and stimulants pales in comparison to what can be done using non-plasma involving synthesis papers. Reducing biomass costs and adding more expensive and powerful options keeps the economy in flux and the increases to the rate of resource growth makes developing stimulants using botany systems appealing compared to simply working with advanced papers. 

<details>

image
</details>

# Changelog
:cl:
balance: increased research's plasma storage per reagent from 120u to 350u!
balance: increased the rate of dedicated plasma organs from 10u to 15u per tick!
balance: increased the amount of times an organ can be planted from once to twice!
balance: reduced the costs of biomass purchases in research's analyzer across the board!
balance: added egg plasma to the Royal Gland extractor!
add: added advanced first-aid kits and compact advanced defibrillators to the biomass analyzer!
/:cl:


Watermelon914:
It's producing 10u per tick, not 4u. I don't think it needs to be increased, 10u per tick is already plenty and the point is that if you want to ramp up production, you need multiple glands. I would honestly prefer if xeno hearts could be planted twice instead (if you go with this approach, remove 1 drone heart from research roundstart0.

Frog6645:
The reagent storage increase is great but the glands honestly produce enough if you plant a few. Granted this requires the marines to actually click bodies with fultons, but that isn't something you can control.

Ammoniacres:
> It's producing 10u per tick, not 4u. I don't think it needs to be increased, 10u per tick is already plenty and the point is that if you want to ramp up production, you need multiple glands. I would honestly prefer if xeno hearts could be planted twice instead (if you go with this approach, remove 1 drone heart from research roundstart0.

I'll go with this route, then.

PR #13161, opened 2026-09-13
fix failing unit test; sentries marked not to need power

### LINKED ISSUE #13150: Flaky test areas_unpowered: Kutjevo - North Colony Caves (/area/kutjevo/interior/colony_north) lacks an APC but requires power for 1 machine!
Flaky tests were detected in [this test run](https://github.com/cmss13-devs/cmss13/actions/runs/34693789090/attempts/1). This means that there was a failure that was cleared when the tests were simply restarted.
Failures:
```
areas_unpowered: Kutjevo - North Colony Caves (/area/kutjevo/interior/colony_north) lacks an APC but requires power for 1 machine!
Machine types: /obj/structure/machinery/defenses/tesla_coil at code/modules/unit_tests/areas_unpowered.dm:91
```

Detective-Google:
# About the pull request

CI would fail if a handheld sentry loot spawner was placed in an APCless area, because they contain the deployed machinery version in their contents, sentries are abstracted as using internal power in our sandbox so they can be marked needs power = FALSE so as not to trigger this.

deployers which in some cases do need power have their own logic and are untouched, so shouldn't affect the power-requiring deployers on for example trijent or hybrisa

Fixes #13150 - a CI failure triggering on desired behavior

# Changelog
should be no player changes


Detective-Google:
image
tested on and off on a powered area( ship prep) and an unpowered one(west lv banks)

Drulikar (approved):

Drulikar:
image
I couldn't find a behavior change either messing with power, being powered, breaking the apc, using laptop, etc.

PR #13163, opened 2026-09-14
Shaking hugged does not reduce stun while hugger is still on

ihatethisengine:
# About the pull request
Shakes don't reduce knockout from getting facehugged while the hugger is still attached to the face (random value between 10 and 15 seconds before falling off, didn't change it, though I could if you are interested...)

It looks goofy that you can shake someone who got hugged and they will stand up with hugger still attached to their face. This essentially adds a minimum stun timer from getting hugged, so hugged person can't instantly get up when shaken by several people, and marines need to actually protect and move the hugged person to safety before spamming shakes. Makes getting hugged with friendlies nearby a bit more scary, makes witnessing a teammate being hugged a lot more scary.

Moot point, but in the movie they had to wait for several hours before waking the infected guy.

https://github.com/user-attachments/assets/d6d0b118-027c-4996-bf83-503dae1fda92

# Changelog
:cl: ihatethisengine
balance: humans can't be shaken up with facehugger attached
/:cl:


zzzmike:
Maybe a custom flavor error message alongside it, like "You shake (person), but it seems to have no effect while a facehugger is on their face"?

PR #13046, opened 2026-08-28
Changes valkyrie rage usage style

Red-byte3D:
# About the pull request

Basically changes rage to XENO_ACTION_QUEUE so that instead of having to switch abilities, you can just click it and use it, this was originally ohw it was meant to be but i couldnt get it to work and now discovered this
# Explain why it's good for the game

This makes it more seamless i think

# Testing Photographs and Procedure
<details>
<summary>Screenshots & Videos</summary>

Put screenshots and videos here with an empty line between the screenshots and the `<details>` tags.

</details>

# Changelog
:cl:
code: Switched Valkyrie's rage ability from XENO_ACTION_CLICK to XENO_ACTION_QUEUE this basically means that the ability no longer has to be chosen and will use on your next clicking, meaning you can indirectly use it without clicking the ability.
/:cl:

PR #13174, opened 2026-09-15
Reverts Valkyrie Tantrum Activation

deadspacemkii:
# About the pull request

Partially reverts #13046 which changed valkyrie's tantrum ability to use a different activation method. This PR reverts tantrum to how it used to be, but leaves the rest of the code for stuff like new strains or castes.

The new ability activation, quite frankly, sucks. I've asked almost every valk I've seen if they like the new activation changes and not a single person said yes. It just feels absolutely terrible to use the ability now because if you have to jump through so many hoops just to use or turn the ability off. Clicking more than once with tantrum will often cause you to waste whatever other ability you have selected. Having to turn off the ability instead of switching it like every other ability in the game is also just bad design, there exists a toggle in the preferences menu for this exact reason.

Works like how it used to

# Changelog
:cl: deadspacemkii
qol: Reverts valkyrie's tantrum ability to use XENO_ACTION_CLICK instead of XENO_ACTION_QUEUE. It now remains selected even after it's used and can only be deselected if the appropriate preference is enabled.
/:cl:


Red-byte3D:
The ability design has nothing to do with selecting or unselecting, its meant to be so you can use it on the go without having to switch abilities.  Because as a valkyrie youre meant to be in there slashing, using your knockdown etc while buffing, this just skips the step of changing abilities and missing out on a stomp or being distracted.

It's a quick ability usage since you mainly want to be on your middle ability anyway.

I think people will appreciate the current way more when their muscle memory gets used to it, i dont really care if this goes through or not thats just my standpoint

fira:
Hi, please make the changelog entry more descriptive

PR #13176, opened 2026-09-15
Gives the M20A the M16 reload and unload sound

Willzadl:
# About the pull request
see title, see changes, intuitive

The L42 reload sound is already used for marine standard issue equipment, that being both the L42 and the ubiquitous M4RA. Instead of reusing this sound, which marines will already be hearing very often and will most likely recognize when using the M20A, why don't we use the M16s, which is being replaced and won't be available anymore? There's no reason to deprecate unique sound effects for sprite changes. The PR probably just forgot the M16 has cool unique reload and unload sound effects already.

1 line change

# Changelog
:cl:
soundtweak: The M20A now uses M16 reload and unload sound effects.
/:cl:


Steelpoint:
Personally I much prefer the current audio for the M20A as it feels far more thematically appropriate for a M41A predecessor, as it is a similar pulse-rifle sounding audio.

Willzadl:
It's a magazine. The mag well of the mk1 was obviously a new invention given that it had 99 bullets of capacity, and the mk2 was descended from that. The M20A magazine should sound like a box magazine because it is a box magazine. None of this has anything to do with pulse rifles

PR #13177, opened 2026-09-15
Makes the Addictive property Boosting.

Ammoniacres:
# About the pull request

Addictive as a property now boosts the effectiveness of chemical properties within subjects.

Fun risk and reward. Addictive is probably one of the most harmful properties despite not being an active deterrent when the stimulant is in system, and we should move towards stimulants being designed with more significant drawbacks to the users.

<details>

</details>

# Changelog
:cl:
balance: [RESEARCH] addictive (ADT) now boosts for 1.5 times the potency of property!
/:cl:


code/modules/reagents/chemistry_reagents/other.dm
antlersss, line 168: You'll want to lower the painkilling property to keep its original level of painkilling as well.

antlersss (commented):
Boosting is an incredibly powerful property that's usually limited to one level. You should limit addictive similarly.

Overall, I don't think addictive itself should have positive properties. If you were to introduce a new property that had the properties of addiction and boosting, then I think that would be better.

code/modules/reagents/chemistry_reagents/other.dm
antlersss, line 205: Not sure why this and other property additions are relevant to the PR. Be sure to add them to the description and changelog.

Ammoniacres: Weird, I was working off a branch I thought already had its changes merged. Let me fix it up.

Ammoniacres:
> Boosting is an incredibly powerful property that's usually limited to one level. You should limit addictive similarly.

This isn't the case, instead, boosting is equivalent to 0.5 per level, and is incredibly expensive to upgrade.

In comparison, INV is 1 per level, but its inclusion of duration reduction balances that out. For ADT, the potency of the chemical decides the strength of the addiction, it's usually not a property you can sustain at any level on a stimulant due to the drawbacks. I could lower it to 1.0, but I imagine even with that it'll be tough to use.

Willzadl:
so what happens when you get a parasitic stim? virtually no downside in practice and no need for stims to be buffed whatsoever

Ammoniacres:
> so what happens when you get a parasitic stim? virtually no downside in practice and no need for stims to be buffed whatsoever

HST actually requires a specific combination of HST and hemogenic properties to be safe, so it being 1.5 makes it pretty impossible to get the math right on that relationship. 

It's the same reason INV isn't used for HSTstims, despite being an obvious.

PR #13178, opened 2026-09-16
Block topics prior to config load

Drulikar:
# About the pull request

I'm not certain that this is possible, but if it is I don't want it happening. Admins.txt seemingly loading during init (because cmdb isn't configured maybe because config hasn't even loaded yet) is what prompted this. But regardless seems abuseable if possible.

# Changelog
:cl: Drathek
fix: Topics that occur prior to config load are no longer possible
/:cl:


Drulikar:
Doesn't currently seem possible w/o config loading introducing a sleep or some other sleep in world/new.

PR #13188, opened 2026-09-18
Decreases L23 recoil

Mary-Marilyn:
# About the pull request

Roughly halfs the amount of recoil the L23 has.
The TWE L23 is pretty much universally regarded as being completely unusable due to the insurmountable amount of recoil (despite the description stating that it's supposed to have "improved" handling over the MK2)

</details>

# Changelog
:cl:
balance: Reduced the recoil of the TWE L23 rifle
/:cl:


Detective-Google:
generally opposed to this; I think ppl are just too used to every gun having no recoil and therefore clash against guns where the mechanic actually is visible, and I'm able to use these guns to great effect despite it so I think them being 'unusable' is an exaggeration.

PR #13194, opened 2026-09-19
Makes GL armor require spec skill to block pounces

Ammoniacres:
# About the pull request

Makes GL armor require spec skill to block pounces. Changes nothing else about how the M3G armor-set works for specialist themselves.

The ability to ignore runners and lurkers is incredibly powerful for a regular marine, and similar to spotter sights on other specialists, it's best to lock the benefits of specialist gear to the specialist (and their helpers) it's sourced from. 

<details>

https://github.com/user-attachments/assets/5372dfb6-3c9c-4e14-945f-2a3c225b7a82

</details>

# Changelog
:cl:
balance: makes GL armor anti-pounce limited to GL specialist skill. 
/:cl:


fira:
That doesn't give the privileges back to GL if they reequip it. 

My take, don't do that, just prevent equipping it without skill entirely. Don't overengineer it.

fira (changes requested):
Doesn't preserve state properly

Red-byte3D:
a lot of people give away the GL amror because they prefer speed, this just makes it so it just stays  as dead content unless the player is new

wgtjunior743:
Unlike the Pyro armor which has an active skill or Sniper which then are locked behind a spec skill, the pounce back is a trade off of the armor since it has the biggest slowdown even bigger than the marine heavy armor.

For a GL speed is essential  most experienced GLs know that and won't take the HUGE slowdown which the armor gives, as Red byte said it will just ended up as dead content and eventually be replaced by something better/different in the future.

PR #13197, opened 2026-09-19
Fixes nested players receiving access to Aghost through another player's moderator permissions.

Mary-Marilyn:
# About the pull request

When a player is nested, they receive a prompt allowing them to ghost. Because the nesting player triggers this prompt, the permission check wrongly checks the person doing the nesting instead of the person being nested.

Aghost itself now also checks against the src client rather than relying on usr so something like this can't happen anywhere else. 

# Changelog

:cl:
fix: Fixed players receiving access to Aghost through another player's moderator permissions.
/:cl:


neeshacark:
LOL? I'VE BEEN GIVING PEOPLE FREE AGHOSTS THIS WHOLE TIME?

Mary-Marilyn:
> LOL? I'VE BEEN GIVING PEOPLE FREE AGHOSTS THIS WHOLE TIME?

Only works on mentors, since they're the only people who the game counts as staff but lack moderator perms

code/modules/admin/holder2.dm
fira: Gotta fix the `usr` here too

fira (changes requested):

fira (approved):

PR #13200, opened 2026-09-20
Removes latejoin spec/SG unlocks

Mary-Marilyn:
# About the pull request

Removes the pop-based latejoin slot unlocks for Specialists and Smartgunners.

Their slots are now fixed to whatever roundstart population scaling gave them.

Latejoin Spec and SG slots encourage people to lobbysit to wait for one to open instead of just joining the round.

This is especially bad on lowpop, where you can end up with a ridiculous amount of the server, sometimes close to two thirds, sitting in the lobby waiting for a Spec or SG slot while the actual round is severely understaffed. Keeping the slots fixed removes that incentive.

There are obviously some balance implications to having fewer Specs and SGs available later in the round, but I think that's worth it if the alternative is having a huge chunk of the server simply not playing.

<details>

</details>

# Changelog

:cl:
balance: Specialist and Smartgunner slots no longer increase mid-round based on latejoining marine population.
/:cl:


Ibessas:
I was thinkin if maybe is there a way to see the time they're at lobby, so the ones sit waiting dont get the slots. I feel block this will really affect the late round balance

sippykot:
waiter, waiter, another marine nerf

Mary-Marilyn:
I'm actually no longer in favor of this

